Issue #1, opened 2014-12-20
Interface redesign


tobyzerner:
Since revealing Flarum in the Kickstarter campaign, I've been thinking more about the interface design from three equally important aspects:
- **Usability.** How simple and intuitive the interface is to use for the average joe.
- **Theming.** How simple and intuitive the interface is to customise for forum admins — both in terms of basic customisation (changing colours) and advanced customisation (changing CSS). 
- **Extensions.** How easy it is for extensions to add functionality to the interface in a consistent manner. 

I don't think I quite hit the sweet spot with Flarum's current design:
- Most importantly, there are instances where the sidebar layout doesn't make complete sense. For example, when the two-pane interface is activated, the discussion scrollbar ends up being separated from the discussion itself, which is really counterintuitive.
- It involves some reasonably complex LESS styling, which wouldn't be ideal for themers.
- Overall, the design doesn't convey the level of simplicity and openness that I want it to. 

Thus, I'm working on a redesign. The goals for the redesign are:
- To make complete paradigmic sense in terms of the content (which the sidebar concept didn't quite)
- To be very open, clean, and easy to style in an aesthetic sense
- For the CSS behind it to be simpler for ease of customization

Here are some initial mockups:
- [Discussion list](https://cloud.githubusercontent.com/assets/128862/5514713/44262268-8897-11e4-849d-f4ca723a1b9f.png)
- [Discussion](https://cloud.githubusercontent.com/assets/128862/5514714/4617eb7e-8897-11e4-850e-861dd59f14b8.png)
- [Discussion with multi-pane (Flarum theme)](https://cloud.githubusercontent.com/assets/128862/5514715/47dae0a6-8897-11e4-82d9-db922422fe03.png)

Some notes:
- Two colors will be configurable by a color picker: a dark primary color and a light secondary color. By default, both of these will be blue. As an example in the Flarum theme, the primary color is orange and the secondary color is blue.
- The header will be able to be switched between two modes: white and colored. By default, it will be in white mode. As an example in the Flarum theme, it is in the colored mode.
- The forum title will be able to be replaced with an image, as in the Flarum theme.
- The links in the header, just to the right of the forum title, won’t display by default when there is only one (Discussions). Extensions will be able to add links (e.g. Members, Website, FAQ).
- The “Welcome” area on the discussion list will be the primary color if the header is in white mode, or the secondary color if the header is in colored mode.
- The title area on a discussion page changes background color according to the channel the discussion is in.
[...]

PR #28, opened 2015-01-01
Interface redesign

tobscure:
WIP for #1


tobscure:
I'm just going to work in `master` for now, these changes are large scale and involve other issues as well, not just the interface redesign.

PR #33, opened 2015-01-04
PUT-PATCH route's methods for the res. update

Allineer:
@see http://restful-api-design.readthedocs.org/en/latest/methods.html


tobscure:
We're following the [JSON-API spec](http://jsonapi.org/format/#patch) which specifies to use PUT for partial updating and PATCH for JSON Patch format operations (which we're not supporting for now.)

PR #37, opened 2015-01-17
Update Forum.php

tdsbaaron:
Removed the unnecessary spaces, and corrected some mistakes.


randsonjs:
This change `public static function boot();`. Why you use semicolon? If you can explain to me. I'll be grateful!

tobscure:
Yeah not quite understanding the semicolon?

PR #40, opened 2015-01-21
[proposal] cameCase in the test names

Allineer:


tobscure:
I'm not sure how I feel about this. The camelCase names are less readable, whereas the lengthy snake_case ones describe the exact behaviour the object is expected to have (like in PHPSpec.) What's your rationale?

qrokodial:
I don't feel like the camelCase here is any less descriptive than your snake_case, and it follows best practices for PHP.

Allineer:
> Method names MUST be declared in camelCase.
[PSR-1](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-1-basic-coding-standard.md) and common practice for PHP, as @qrokodial says.

tobscure:
Right, fair enough. Perhaps we should look into integrating the Codeception Specify extension to make tests more readable: http://codeception.com/10-04-2013/specification-phpunit.html

tobscure:
Oh, why did you delete? Was about to merge.

Allineer:
Sorry O_o.
One moment plz...

PR #49, opened 2015-03-24
Always show vertical scrollbar

martindilling:
Show the vertical scrollbar as inactive when content of the site
don't require any scrolling, instead of not showing anything.
This avoids the annoying "jumps" when you switch between
pages that require scrolling vs don't require scrolling.

### DIFF
--- a/ember/app/styles/flarum/layout.less
+++ b/ember/app/styles/flarum/layout.less
@@ -1,6 +1,7 @@
 body {
   background: @fl-body-bg;
   color: @fl-body-color;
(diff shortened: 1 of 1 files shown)


jayanratna:
Agreed. What you explained above was evident when you clicked the login or sign up links on the demo. 👍

tobscure:
Nice one. I hadn't noticed because I'm developing on Mac OS where scrollbars are invisible.

martindilling:
That's a fair excuse :P

PR #50, opened 2015-03-25
API for uploading avatars

franzliedke:
The backend implementation for uploading user avatars. I like this architecture! =)

This is still work in progress.

### DIFF
--- a/ember/app/components/user/avatar-editor.js
+++ b/ember/app/components/user/avatar-editor.js
@@ -1,0 +1,45 @@
+import Ember from 'ember';
+
+import config from 'flarum/config/environment';
+
+var $ = Ember.$;
+
+export default Ember.Component.extend({
+  layoutName: 'components/user/avatar-editor',
+  classNames: ['avatar-editor', 'dropdown'],
+  classNameBindings: ['loading'],
+
+  click: function(e) {
+    if (! this.get('user.avatarUrl')) {
+      e.preventDefault();
+      e.stopPropagation();
+      this.send('upload');
+    }
+  },
+
+  actions: {
+    upload: function() {
+      if (this.get('loading')) { return; }
+
+      var $input = $('<input type="file">');
+      var userId = this.get('user.id');
+      var component = this;
+      $input.appendTo('body').hide().click().on('change', function() {
+        var formData = new FormData();
+        formData.append('avatar', $(this)[0].files[0]);
+        component.set('loading', true);
+        $.ajax({
+          type: 'POST',
+          url: config.apiURL+'/users/'+userId+'/avatar',
+          data: formData,
+          cache: false,
+          contentType: false,
+          processData: false,
+          complete: function() {
+            component.set('loading', false);
+          }
+        });
+      });
+    }
+  }
+});
--- a/ember/app/styles/flarum/user.less
+++ b/ember/app/styles/flarum/user.less
@@ -47,10 +47,17 @@
     display: inline;
     vertical-align: middle;
   }
-  & .avatar {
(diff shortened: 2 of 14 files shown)


thangngoc89:
I think avatar serve should be a Laravel route like:

```
/user/{size}/file_name.jpg
```

This will improve website speed (Flarum homepage avatar size is 48px x 48px)

franzliedke:
@thangngoc89 I basically agree, but I think that this URL should just redirect to that file (served e.g. as a static file by the webserver, or directly from a CDN).

thangngoc89:
@franzliedke I agree. We can simply redirect that URL to the real image. (and its different size)

tobscure:
This is awesome, great work!

A few thoughts (mostly nit-picking):

- I think the `avatar` column in the DB be called `avatar_url`. Just to be explicit and consistent.
- The name of the `$actor` property of the UploadAvatarCommand is a bit confusing because it’s not actually an Actor instance? I agree that `$user` (as in other commands) is also a bit confusing when you’re performing an action on a user... Either way we should decide on something and make it consistent across all commands.
- Store avatars in the skeleton app’s public directory, not the ember app’s one.
- I don’t think avatars should be served via a Laravel route, because then you have a heap of non-static requests on every page load. Rather, a static avatar URL is included in the API for each user.
- I hadn’t thought about having different avatar sizes, but it’s probably a good idea. I think the best way would be to do something like Twitter, where you just modify a suffix in the URL on the consumer-end: https://dev.twitter.com/overview/general/user-profile-images-and-banners

franzliedke:
Thanks.

Thoughts on the thoughts:
- I actually prefer `avatar_path`, since the absolute URL should be generated based on the storage that's used (e.g. local vs. a CDN). This would then be the path relative to that storage location (makes moving easier).
- Yeah, I didn't want to use `$user`, because we also have `$userId` (very confusing in some other command classes). I wasn't referring to that class, it's just a good description (the person executing the command). Any better names?
- Cheers, that's good to know. I'll probably add a configuration for the "disk" (see [docs](http://laravel.com/docs/5.0/filesystem)) to the base repo instead of configuring it directly in the service provider.
- Agreed. Maybe we should still add that redirection route to the API to have a never-changing URL for other applications. (Then again, that should probably be a hypermedia link from the user endpoint.)
- Different sizes sound good. For later. ;)

tobscure:
Thoughts on the thoughts on the thoughts:

- Great thinking. `avatar_path` it is.
- Other ideas: `$agent`, `$operator`, `$endUser`, `$client`... it's a shame I didn't call users "members" like in esoTalk!
- Good idea.
- Yeah, maybe. Facebook provides a redirection URL like this but Twitter requires you parse an API response. I kind of like the purity of the latter, and whenever you need to display a user avatar you'll probably need an API response anyway. We'll see.
- Cool :)

franzliedke:
Quick update:
- I settled on `avatar_path`.
- I [created an avatars directory](https://github.com/flarum/flarum/commit/c8568df40fec2687e192c98bec8625681e75a1ff) in the main repository, and [configured the filesystem](https://github.com/flarum/flarum/commit/8ee8a0a9b9ac5ad9c9530eec3e9073ee21495c39) there.
- I just uploaded my first avatar, yay.

franzliedke:
If you're fine with this, it can be merged, I'd say.

I'll work on the resizing (and other file types) over the weekend.

tobscure:
Just realised we will need to make the UploadAvatarAction return a JSON-API response — probably a refreshed user model — so that the UI has a new avatarUrl to show. I might tackle that today and polish up the UI.

PR #91, opened 2015-05-22
Refactoring: PSR-7, decouple from Laravel, ...

### LINKED ISSUE #75: Decouple from Laravel
From tobyzerner.com/mithril:
> Instead of using the whole framework, we will explicitly pull in components and tie everything together with a custom bootstrapping process. This will allow us to optimize performance and provide better integration with existing applications.

I thought I'd create an issue so we can (a) work out exactly what needs to be done and (b) keep track of it.

[fluxbb/fluxbb#2.0](https://github.com/fluxbb/fluxbb/tree/2.0) and [fluxbb/core](https://github.com/fluxbb/core) are obviously good inspiration :)

- [x] Implement a custom [router](https://github.com/nikic/FastRoute)
- [x] Explicitly pull in only the Illuminate components that we need
- [x] Pull in a PSR-7 implementation (probably the one by @phly, which has now moved to Zend)
- [x] Pull in a PSR-7 middleware stack ([zend-stratigility](https://github.com/zendframework/zend-stratigility)) to the skeleton repository. This will replace Laravel's HTTP kernel.
- [x] Convert the different parts (frontend, admin and API) of the app to separate middlewares that can be mounted independently. This should also speed up routing because only the routes for the proper middleware need to be parsed. (Special attention to route generation, because we do need all routes there.)
- [x] Hook this up with a container instance (an application class?) and load our service providers
- [x] Implement a config system: `config.php` for database details, everything else in a ConfigRepository (database)
- [ ] Provide integration with Laravel (possibly via a separate package), basically bridging the Symfony HTTP kernel to PSR-7.
- [x] Review and update the Composer requirements (Laravel 5.1, PHP 5.5, ...)

@franzliedke I'm dying for your input here :D I'm happy to do anything you want to delegate, I really want to get this moving :)

franzliedke:
An ongoing effort to decouple from Laravel. Currently WIP.

Here's the plan:
- [x] Implement a custom [router](https://github.com/nikic/FastRoute)
- [ ] Explicitly pull in only the Illuminate components that we need
- [x] Pull in a PSR-7 implementation (probably the one by @phly, which has now moved to Zend)
- [x] Pull in a PSR-7 middleware stack ([zend-stratigility](https://github.com/zendframework/zend-stratigility)) to the skeleton repository. This will replace Laravel's HTTP kernel.
- [ ] Convert the different parts (frontend, admin and API) of the app to separate middlewares that can be mounted independently. This should also speed up routing because only the routes for the proper middleware need to be parsed. (Special attention to route generation, because we do need all routes there.)
- [ ] Hook this up with a container instance (an application class?) and load our service providers
- [ ] Implement a config system: `config.php` for database details, everything else in a ConfigRepository (database)
- [ ] Provide integration with Laravel (possibly via a separate package), basically bridging the Symfony HTTP kernel to PSR-7.

This will fix #75.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,9 @@
         "tobscure/permissible": "dev-master",
         "misd/linkify": "1.1.*",
         "oyejorge/less.php": "dev-master",
-        "intervention/image": "dev-master"
+        "intervention/image": "dev-master",
+        "psr/http-message": "^1.0@dev",
+        "nikic/fast-route": "dev-master"
     },
     "require-dev": {
         "fzaninotto/faker": "1.4.0",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "hash": "2c72cfaf1f4ffecca91460d1e6f57f5d",
+    "hash": "8b517a2c5b1d000a443c5ad91c49b2c1",
     "packages": [
         {
             "name": "danielstjules/stringy",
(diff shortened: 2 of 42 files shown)


franzliedke:
Current status: API and admin are completely converted (though not working in all cases). Forum is a bit trickier (dealing with Laravel's views etc.), but almost done, too.

Once I finish that (tomorrow), I'll throw out Laravel and include components only. I will basically copy my work from FluxBB 2.0 for that part.

Oh, and I also rebased this onto Toby's latest changes, so merging should work fine whenever we do it.

franzliedke:
List will be kept in the [corresponding issue](https://github.com/flarum/core/issues/75).

PR #278, opened 2015-08-28
Make Google Web Font work on both HTTP and HTTPS

huytd:
Just a little hack to make the Google Web Font work on both HTTP and HTTPS
The solution is to remove `http:` in Google Web Font import, see change log for more detail.


tobscure:
:+1:

PR #299, opened 2015-08-28
Fix text overlap in search result with long content

huytd:
When text content in search result has a very long single line, it makes the text overlap the page, as in the below image:

Just add word-wrap property to `PostPreview-excerpt` class to make it force wrap the words


tobscure:
:+1:

PR #387, opened 2015-08-31
Align select input properly

rodymol123:
I was annoyed by the select input on the index not being aligned properly.
Set the line-height to 1 for select inputs only.


franzliedke:
Thank you, that looks better indeed!

PR #401, opened 2015-09-01
Add a configuration data provider to fetch variables from environment

jubianchi:
This PR will help automating the Flarum install process. Actually, to bootstrap the initial configuration file the user has to fire a CLI command (`php flarum install [--defaults]`) which either asks the user to type values or fetch them from a default provider.

With this patch, we have a new provider: `DataFromEnvironment`. It will read the configuration from environment variables:

```sh
export FLARUM_DATABASE_HOST=mysql.local
export FLARUM_DATABASE_NAME=flarum
export FLARUM_DATABASE_USER=flarum
export FLARUM_DATABASE_PASSWORD=flarum
export FLARUM_ADMIN_USERNAME=admin
export FLARUM_ADMIN_PASSWORD=admin
export FLARUM_ADMIN_EMAIL=admin@docker-fl
export FLARUM_FORUM_TITLE=Docker Flarum
export FLARUM_BASE_URL=http://flarum.local

php flarum install --env
```

This is really useful to provide an automated install for docker for example.

### DIFF
--- a/src/Install/Console/DataFromEnvironment.php
+++ b/src/Install/Console/DataFromEnvironment.php
@@ -1,0 +1,73 @@
+<?php
+/*
+ * This file is part of Flarum.
+ *
+ * (c) Toby Zerner <toby.zerner@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Flarum\Install\Console;
+
+class DataFromEnvironment implements ProvidesData
+{
+    protected $baseUrl;
+
+    public function getDatabaseConfiguration()
+    {
+        return [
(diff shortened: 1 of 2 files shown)


jubianchi:
I'm thinking of a better solution right now, let me implement it ;)

jubianchi:
This was a bad idea. sorry for the noise!

franzliedke:
That's actually a very interesting idea. :)

jubianchi:
@franzliedke in fact it is but the way I did it was not clean enough to me. I'm experimenting it with an other method and it seems to be ok and does not require any modification in the core ;)

Mecvak:
This would be extremely useful for installing by script. Hopefully this new idea still utilizes environmental variables!

PR #402, opened 2015-09-01
Fixes enable extension in EntensionManager

johannsa:
Enabling an extension was causing previously enabled extensions to get disabled


franzliedke:
:+1:

tobscure:
Thanks!

Issue #246, opened 2015-08-27
Review and improve Extension API


tobyzerner:
I would love to get some feedback on Flarum's Extension API. If you haven't seen much of it, check out the [documentation](http://flarum.org/docs), as well as the source code for all of [Flarum's packaged extensions](https://github.com/flarum).
### Scoping

Currently we don't distinguish very well between what's part of the public API and what's not – and if we want to do SemVer, we need to do this much better. Given how the API is set up, how can we effectively do this? Some ideas:
- Prefix all private JavaScript variables and methods with `_`. Make sure docblocks reflect that too.
- Even if something is public, we may not want to guarantee that we won't break it. Should we explicitly use the `@api` docblock tag to denote our public API?
- On the backend, we need to more methods `private` instead of `protected`.
- Should we make any classes/methods `final`?
### Monkey Patching

Extending the JavaScript application uses monkey patching as its primary mechanism. I chose this because it was the simplest, most straightforward option – however, I'm wondering about a few potential pitfalls:
- There's no way for extensions to specify the priority of their monkey patch. This is not a huge problem because most monkey patching is done to add items to an ItemList, which allows for prioritization.
- Functions directly exported by modules cannot be monkey patched – i.e. helpers and some utils. Workaround is to make helpers into components, and make utils into objects containing functions.
- Class `constructor`s cannot be reliably monkey patched because sometimes the constructor property is used to access a static method (e.g. `this.constructor.whatever()`). We could work around this by making our monkeypatch methods copy over the original object's properties.

The obvious alternative is to have an event-based API, like on the backend. This would entail firing events in all methods that were previously subject to monkey-patching, and then changing all instances of `extend` and `override` into event listener registrations. (This would help to better define the scope of the API too.) While a big change for this stage of development, it's certainly not too late. I would just like to get some second opinions.
### Naming

I think I want to tighten up the naming of Flarum\Event classes, as they're a bit inconsistent at the moment. That and elsewhere, are there any names that you find confusing, and do you have any suggestions for replacements?
### General Feedback

If you have experience with public APIs, can you tell us: What could be better about Flarum's API? What parts are good, what parts are bad?

tobyzerner:
Some random ideas I've come up with while developing a Reports extension:

### PHP
- Use our own dispatcher abstraction
- Change Action $include ['relationship' ⇒ true/false] to $includeDefault and $includeOptional. Update BuildApiAction event too.
- Make more properties and methods private
[...]

PR #403, opened 2015-09-01
Allow support for passing extra signup data to API

mtotheikle:
This is required for Google reCAPTCHA plugin to work


franzliedke:
For overwriting that extracted method, I suppose? :+1:

tobscure:
Yep this is good. We will need to review naming at some stage to try and make it consistent between these kind of methods on various components (e.g. ReplyComposer, DiscussionComposer). ref #246

PR #406, opened 2015-09-01
Turkish translation added...

hidonet:
Turkish translation added...


franzliedke:
Thanks for contributing! :+1:

We're not adding other locales to the core, though. Please create an extension instead. The [French translation](https://github.com/maelsoucaze/flarum) may be a good inspiration.

hidonet:
Ok. I'm going to fork it. Thanks

PR #407, opened 2015-09-01
add traditional Chinese locale

ericcire:
add traditional Chinese locale


rodrigoargumedo:
I would write this as a extension but I might think this is a unnecessary pull request. Though, it works great out of the box and all. However, I suggestively advise to go against it.

This is just my two cent opinion and thoughts.

franzliedke:
Thanks for contributing!

I have to close this, though, for the same reason as #406.

PR #408, opened 2015-09-01
fixes issue of not saving extension to setting table

Luceos:
Fixes the issue of not being able to enable extensions. Before the fix enabling would disable all previous extensions.


Luceos:
already fixed

PR #413, opened 2015-09-02
Update RouteCollection::getPath

WinterSilence:
This version work faster - old code create closure at every calling getPath

### DIFF
--- a/src/Http/RouteCollection.php
+++ b/src/Http/RouteCollection.php
@@ -80,16 +80,19 @@ class RouteCollection
         return $this->dataGenerator->getData();
     }
 
-    public function getPath($name, $parameters = [])
+    protected function fixPathPart(&$part, $key, array $parameters)
+    {
+        if (is_array($part) && array_key_exists($part[0], $parameters)) {
+            $part = $parameters[$part[0]];
+        }
+    }
+
+    public function getPath($name, array $parameters = [])
     {
         $parts = $this->reverse[$name][0];
 
-        $path = implode('', array_map(function ($part) use ($parameters) {
-            if (is_array($part)) {
-                $part = $parameters[$part[0]];
-            }
-            return $part;
-        }, $parts));
+        array_walk($parts, [$this, 'fixPathPart'], $parameters);
+        $path = implode('', $parts);
 
         $path = '/' . ltrim($path, '/');
         return $path;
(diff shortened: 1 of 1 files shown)


src/Http/RouteCollection.php
```diff
@@ -80,16 +80,19 @@ public function getRouteData()
         return $this->dataGenerator->getData();
     }
 
-    public function getPath($name, $parameters = [])
+    protected function fixPathPart(&$part, $key, array $parameters)
+    {
+        if (is_array($part) && array_key_exists($part[0], $parameters)) {
```
franzliedke: Are you sure the `array_key_exists()` call is actually necessary?

src/Http/RouteCollection.php
```diff
@@ -80,16 +80,19 @@ public function getRouteData()
...
-        $path = implode('', array_map(function ($part) use ($parameters) {
-            if (is_array($part)) {
-                $part = $parameters[$part[0]];
-            }
-            return $part;
-        }, $parts));
+        array_walk($parts, [$this, 'fixPathPart'], $parameters);
+        $path = implode('', $parts);
 
         $path = '/' . ltrim($path, '/');
```
franzliedke: You can combine these two lines, I'd say. :)

src/Http/RouteCollection.php
WinterSilence: Validate incoming parameters guarantee correct execution method.

src/Http/RouteCollection.php
WinterSilence: array_walk returns boolean value

franzliedke: I meant the two below that:

~~~php
$path = '/'.ltrim(implode('', $parts), '/');
~~~

WinterSilence: ups)) updated

src/Http/RouteCollection.php
```diff
@@ -80,18 +80,18 @@ public function getRouteData()
...
+            $part = $parameters[$part[0]];
+        }
+    }
 
-        $path = '/' . ltrim($path, '/');
+    public function getPath($name, array $parameters = [])
+    {
+        $parts = $this->reverse[$name][0];
+        array_walk($parts, [$this, 'fixPathPart'], $parameters);
+        $path = '/' . ltrim(implode('', $parts), '/');
```
franzliedke: Just return it directly ;)

franzliedke: Will do that for you.

PR #418, opened 2015-09-03
call to $this-> assertValidPassword from static context

Luceos:
assertValidPassword called using `$this` in static context


tobscure:
Thanks!

PR #438, opened 2015-09-06
Add a default build script for extensions

mtotheikle:


franzliedke:
Very cool, I like this. Thank you! :+1: 

I'll let @tobscure merge this, so he can have a look.

tobscure:
Yeah this is nice, thanks. I think down the track, for both the extension stub and flarum/flarum, we should separate into two scripts: compile.sh (which runs npm/bower install and gulp --production in relevant places), and build.sh (which runs compile.sh and then packages it up into a zip).

PR #488, opened 2015-09-15
Changed default Admin password

billmn:
Default Admin password doesn't pass the new validation rule (min 8 chars)

See: https://github.com/flarum/core/commit/cbcad27679faaaeb10206d7c25e21952231b5542#diff-2e6d4ed85cd06d3e11f7f8428746214eR126


franzliedke:
Hah, fun, thanks!

Issue #266, opened 2015-08-27
Find and fix hardcoded translations


tobscure:
Although most of the client is translatable, there are still a few instances of hardcoded text which need to be fixed.

What needs to be done:

* Make `utils/punctuate` and `utils/abbreviateNumber` translatable. See http://discuss.flarum.org/d/336-potential-l10n-stumbling-blocks/5
* Find any other hardcoded translations by temporarily emptying en.yml and searching far and wide for text that is still in English.
* Do the same for all default extensions.
* Make the whole administration section translatable (it is all hardcoded at the moment).

franzliedke:
@dcsjapan Can we close this?

dcsjapan:
I think it's safe to close it now.

As you say, we've caught enough of them that instead of actively looking for them, we can simply wait for localizers to mention any they find. Hopefully there won't be any! :smile:

PR #504, opened 2015-09-16
Make punctuate function translatable

rodymol123:
Add an 'and' translation key to the core, and let punctuate.js make use of it. #266


franzliedke:
This is looking good, but can you please squash this to one commit? :)

Also, @dcsjapan, can you confirm this is the right way to go about this? Or will the hardcoded spaces cause any problems with other languages?

dcsjapan:
@franzliedke Thanks for pinging me.

Yes, the hardcoded spaces will cause trouble with languages that don't use them.

In addition, it looks like it's tripping over the same issue that I brought up with @tobscure in regard to esoTalk: not all languages put the "and" between the last pair. In Japanese, for example,

- John, Paul, George, and Ringo

would be written as

- John and Paul, George, Ringo

So the whole thing really needs to be handled as a phrase.

In esoTalk, @tobscure was originally using phrase with the structure

- "%s and %s"

and imploding all of the array except the last item into the first %s. We managed to make it work in Japanese by adding a second phrase

- "%s, %s, and %s"

for use in cases when the array was longer than two.

Would it be possible to do something similar here?

PR #507, opened 2015-09-16
adds heavier validation, also limits to 100 as per database column, a…

Luceos:
…dded some blacklist values for username

Username validation was far too simple. Added some basic values and restrictions based on database column.

### DIFF
--- a/src/Core/Users/User.php
+++ b/src/Core/Users/User.php
@@ -55,7 +55,13 @@ class User extends Model
      * @var array
      */
     protected $rules = [
-        'username'          => 'required|alpha_dash|unique',
+        'username'          => [
+            'required',
+            'alpha_dash',
+            'unique',
+            'max:30',
+            'min:3'
+        ],
         'email'             => 'required|email|unique',
         'password'          => 'required',
         'join_time'         => 'date',


tobscure:
- We should limit usernames to maximum 20-30 characters. 3 minimum is good.

- I don't think this is the right way to go about preventing people from using certain usernames. For one, reserving `admin` will break the vagrant installer. But more importantly, the idea of a blacklist is flawed; you can just work around it by using a similar username (e.g. `mod`, `admin_`, `root-`). If a forum admin has a problem with someone taking a username like that, they can just delete or rename them. There's nothing inherently wrong with using those usernames, so we'll leave blacklisting up to an extension.

franzliedke:
I agree, 30 characters should be good.

Luceos:
Updated:
- removed blacklist items
- limited to between 3 and 30

tobscure:
Please squash the commits and it'll be good to go :D

Luceos:
i guess my rebase failed immensily?

franzliedke:
Yep, please try again :)

franzliedke:
[Did it for you.](https://github.com/flarum/core/commit/68369ac5bb6a1729b34ad8b0a2851c5c9d170626)

PR #508, opened 2015-09-16
Make punctuate function translatable

rodymol123:
Add an 'and' translation key to the core, and let punctuate.js make use of it. #266


locale/en.yml
```diff
@@ -104,3 +104,4 @@ core:
   welcome_user: "Welcome, {username}!"
   write_a_post: Write a Post...
   write_a_reply: Write a Reply...
+  and:  and
```
franzliedke: Is there a trailing space after the "and"? I'm not sure that works or is a good idea; as it's so easy to omit or forget...

rodymol123: What would you think is suitable then?

dcsjapan:
I'll drop a note here because @rodymol123 added this PR right before I commented on the earlier one.

Please see my comments here:

https://github.com/flarum/core/pull/504

Adding the "and" all by itself just won't work for some languages.

Issue #351, opened 2015-08-29
Sort menu


Petrouchka:
**_1 Upvote**_ In the sort menu, it's difficult to understand the difference between "Latest" and "Newest" because it concern one time the Replies and the other time the Discussions. Perhaps using this ?
Latest −> Recent replies (or Last replies)
Top -> Top discussions
Newest -> New discussions
Oldest -> Old discussions

maelsoucaze:
+1, some users brought me to this issue as well.

maelsoucaze:
I just changed "Latest" by "Activity" and it's really clearer in French, according to feedbacks. I don't know if it's the case in English. What do you think?

Petrouchka:
Oui, dans diaspora*, il y a aussi Mon activité et Flux.
J'ai laissé tombé Flarum car presque la moitié de mes rapports de bogues ont été fermés sans aucune discussion et maintenant, il n'y a plus de paquet PHP/MySQL téléchargeable. Ça m'a gavé de voir le manque d'intérêt pour les utilisateurs, du coup, j'utilise à la place diaspora* et je suis bien content d'avoir fait l'effort de le découvrir.

maelsoucaze:
@Petrouchka Flarum is powered by only two developers at this time, not available all the time, and they're focused to fix major bugs in priority to release a stable version as soon as possible. I'm sure they're enjoying all contributions, they're just not enough for now to take the time and the energy for all of them, so they often prefer to let a feature to be discussed. Concerning the download package, it's just a temporary solution, it will be easier later. I'm sure everything will be better in the future. Flarum is still a young piece of software, and users are often too excited to implement features. That's a major deal as developer to slow down the hype and keep a cool head to keep the right way. I'm sad you gave up on it, but glad if you found something better for you elsewhere. :)

franzliedke:
I'll leave this to you, @tobscure, to improve when implementing the discussion list revamp.

PR #511, opened 2015-09-17
Differentiations between Discussions and Replies in the sort menu, in…

Petrouchka:
… order to avoid confusion between "Latest" and "Newest"

https://github.com/flarum/core/issues/351


tobscure:
This kind of change needs to be discussed on the forum before we accept a PR.

PR #518, opened 2015-09-18
Routes fix

### LINKED ISSUE #123: Fix URL generation in various places
Splitting up our front controllers is a great move, but URL generation is problematic in some cases:
- Flarum\Core\Handlers\Commands\RequestPasswordResetCommandHandler
- Flarum\Core\Handlers\Events\EmailConfirmationMailer
- Flarum\Api\Actions\Discussions\IndexAction

The first two are run through api.php, but need to generate forum routes.
The third can be run through index.php (when data is preloaded), but needs to generate an API route for the pagination links.
The Mentions extension also needs to generate forum routes when run via api.php.

@franzliedke how can we handle this?

Luceos:
This fixes #123 especially for flarum/mentions ; other extensions need to be checked!


Luceos:
build failed due to a missing newline after namespace declaration, but there is one :-1:

tobscure:
Failing tests are failing tests... look harder! There are *two* new lines after [`namespace Flarum\Forum\Http;`](https://github.com/flarum/core/pull/518/files#diff-296e9611fb920d8a873edc16e6aa4e35R12).

Not sure that three separate (empty) interfaces are the right way to go about this? I would've thought they should all extend `Flarum\Http\UrlGeneratorInterface`, not the other way around... @franzliedke what do you think?

Why was [this line](https://github.com/flarum/core/pull/518/files#diff-d3de9927352a4cf697464e79e9012763L30) removed? (Sorry to be pedantic, but this is a regression in terms of code formatting! :P)

Luceos:
@tobscure [that](https://github.com/flarum/core/pull/518/files#diff-d3de9927352a4cf697464e79e9012763L30) line was removed by mistake, it held my var_dumps. It seems I have mistakenly committed that. I'll re add it. And I hadn't spotted the newlines no, it was past midnight and I _had_ to finish this up because I had been at this for days already :disappointed:

Luceos:
> Not sure that three separate (empty) interfaces are the right way to go about this? I would've thought they should all extend Flarum\Http\UrlGeneratorInterface, not the other way around...

I tried that. That was my first idea too. But a class cannot implement multiple interfaces and due to the fact that all routes are using the one UrlGenerator.. This was the solution without generating completely new UrlGenerators ánd UrlGeneratorInterfaces per package (forum, api, admin).

Luceos:
Btw I recommend [style-ci](https://styleci.io/) for code formatting, it can generate a pull request with fixes automatically.. Lot easier and less work.

Luceos:
Ps if you want to make this backward compatible, we can alias `Flarum\Http\UrlGeneratorInterface` to the current interface of the service provider.

tobscure:
Backwards compat not an issue, but I'd still like to get @franzliedke's input since this was his code originally :)

Luceos:
@tobscure I'd like to hear his opinion as well, maybe he has an alternative solution which fits more to his coding style.

franzliedke:
Hmm, no, I don't want three different interfaces.

When binding them, we probably want to bind not as singleton, but simply with `->bind()`. And as name for the binding, we'll use something like `flarum.api.routes` and not the interface name.

Meh, I'm too tired now, will explain in detail tomorrow.

Luceos:
@franzliedke i looked at this again.. The routes collections are already defined as `flarum.N.routes`, for instance see [admin](https://github.com/flarum/core/blob/master/src/Admin/AdminServiceProvider.php#L56). We could bind them as `flarum.N.url_generator`? Or do you have another proposition?

I'd like to close this so the routes are fixed from the api.

tobscure:
Closing given @franzliedke's comment in #123: https://github.com/flarum/core/issues/123#issuecomment-143175250

PR #538, opened 2015-09-23
make cookie name configurable

Luceos:
The current cookie set for remembering and restoring session is called `flarum_remember`. On domains that have multiple installations this will conflict and users are automatically logged out from another installation if logging in elsewhere on the domain.

This requires also a change in the `config` ioc binding in `flarum/bootstrap.php`, see the references PR: https://github.com/flarum/flarum/issues/17

> Side info, these pr's allow me to stay logged in on all flarum.today installations.

### DIFF
--- a/src/Forum/Actions/WritesRememberCookie.php
+++ b/src/Forum/Actions/WritesRememberCookie.php
@@ -21,7 +21,7 @@ trait WritesRememberCookie
         // Set a long-living cookie (two weeks) with the remember token
         return FigResponseCookies::set(
             $response,
-            SetCookie::create('flarum_remember', $token)
+            SetCookie::create(app('config')->get('session.cookie'), $token)
                 ->withMaxAge(14 * 24 * 60 * 60)
                 ->withPath('/')
                 ->withHttpOnly(true)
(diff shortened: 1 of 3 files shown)


tobscure:
I'm thinking we can probably pull a value straight out of our SettingsRepository (say `cookie_name`), no need to go through the Illuminate config repo? @franzliedke?

Luceos:
I'm not sure this should be configurable from the admin ui, if you specify the database config as being meant for the ui. The reason I added it within the illuminate config is because also the cache driver is configured there.

tobscure:
Database config isn't necessarily exposed to the UI.

Difference is that we use Illuminate's cache component so we have to use its config repository. But we're not using Illuminate's session/cookie stuff so it doesn't make sense to use it here.

franzliedke:
I agree that this shouldn't be in the IoC container.
If anything, I would have probably just added it to config.php. For things like multitenancy, storing it with the other settings (database / cache) may actually be better, though, so I agree with Toby.

kirkbushell:
For multi tenancy, it may be worth thinking a little bigger than the cookie name being configurable. 

Then you can have strategies to manage that aspect - aka, is the forum supporting multiple accounts across domains, or folders.etc?

PR #553, opened 2015-09-28
Tests for core settings code

kirkbushell:
This is the start of something beautiful - TESTS!

@franzliedke had already added some tests using phpspec, but due to some of the complexity and use-cases (such as the repository object chaining), said tool wasn't going to work for all unit/behavioural tests. As a result, and after a discussion with Franz, have added phpunit, mockery, and converted the existing specs to unit tests.

This should act as a pattern/approach for the rest of the project, and I'll begin by implementing this for the settings handlers and other associated code before moving onto other modules.


composer.json
```diff
@@ -36,19 +36,21 @@
...
+        "mockery/mockery": "^0.9.4",
         "squizlabs/php_codesniffer": "2.*",
-        "phpspec/phpspec": "^2.2"
+        "phpunit/phpunit": "^4.8"
     },
     "autoload": {
         "psr-4": {
-            "Flarum\\": "src/"
+            "Flarum\\": "src/",
+            "tests\\": "tests/"
```
franzliedke: Please move this line to an `autoload-dev` block just like this one (it has the same structure)...

kirkbushell: Ah! That's what you meant :)

Fixed.

franzliedke:
Thank you so much!

tobscure:
> This is the start of something beautiful - TESTS!

:sparkling_heart:

PR #554, opened 2015-09-28
Admin tests

kirkbushell:
Added some base admin tests. There are some problems with the middleware currently that places some implicit knowledge (AccessToken) that can't be unit tested currently, so that's been skipped. Happy to discuss further.

May be worth injecting a repository or service here to manage the access token searching, as well as managing the dependencies deeper inside the logIn method.

Also, zend-stragility was missing from the core library.

### DIFF
--- a/tests/Flarum/Admin/Middleware/LoginWithCookieAndCheckAdminTest.php
+++ b/tests/Flarum/Admin/Middleware/LoginWithCookieAndCheckAdminTest.php
@@ -1,0 +1,23 @@
+<?php
+namespace tests\Flarum\Admin\Middleware;
+
+use Flarum\Admin\Middleware\LoginWithCookieAndCheckAdmin;
+use Flarum\Core\Exceptions\PermissionDeniedException;
+use Mockery as m;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use tests\Test\TestCase;
+
+class LoginWithCookieAndCheckAdminTest extends TestCase
+{
+    public function test_it_should_not_allow_invalid_logins()
+    {
+        $this->setExpectedException(PermissionDeniedException::class);
+
+        $request = m::mock(ServerRequestInterface::class);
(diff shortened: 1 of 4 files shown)


tests/Flarum/Admin/Middleware/LoginWithCookieAndCheckAdminTest.php
```diff
@@ -0,0 +1,27 @@
...
+{
+    public function test_it_should_not_allow_invalid_logins()
+    {
+        $this->setExpectedException(PermissionDeniedException::class);
+
+        $container = m::mock(Container::class);
+        $request = m::mock(ServerRequestInterface::class);
+        $response = m::mock(ResponseInterface::class);
+
+        $request->shouldReceive('getCookieParams')->andReturn([]);
```
franzliedke: Are we really testing that the request should receive the method call? Shouldn't this just be a stub?

kirkbushell: Yeah I could set it up as a spy, but the reason I've done it this way is to pave the way for tests once I can test the class in full. Currently there is too much embedded code that is untestable.

kirkbushell: And tbh I should check the return value, as well.

tobscure:
> May be worth injecting a repository or service here to manage the access token searching, as well as managing the dependencies deeper inside the logIn method.

Absolutely. Sorry about the lack of DI/testability in some places... was rushing to get the beta out! Please feel free to take initiative here and extract as you see fit – and let me know if I can be of any assistance :)

kirkbushell:
@tobscure not a problem at all man - well aware of deadlines and that "rush" to get it all out :)

I'll keep these PRs coming, but any major refactoring I'll discuss with you and franz first, as I don't want to step on toes. It's also important to understand design decisions or direction that may have had a lot of effort put into them.

tobscure:
Sure thing, sounds good :D

PR #558, opened 2015-09-29
Feature/json api error handling

kirkbushell:
API exception handling improvement - allows for individual handlers to respond to specific exceptions, with the long-term goal of developers being able to define their own exception handlers for their code, thus generating custom responses.

### DIFF
--- a/src/Api/ApiServiceProvider.php
+++ b/src/Api/ApiServiceProvider.php
@@ -19,6 +19,10 @@ use Flarum\Events\RegisterActivityTypes;
 use Flarum\Events\RegisterNotificationTypes;
 use Flarum\Http\RouteCollection;
 use Flarum\Api\UrlGenerator;
+use Flarum\Support\Json\FallbackExceptionHandler;
+use Flarum\Support\Json\ModelNotFoundExceptionHandler;
+use Flarum\Support\Json\SerializableErrorHandler;
+use Flarum\Support\Json\ValidationExceptionHandler;
 use Illuminate\Support\ServiceProvider;
 use Psr\Http\Message\ServerRequestInterface;
 
@@ -44,7 +48,10 @@ class ApiServiceProvider extends ServiceProvider
 
         $this->app->singleton(ErrorHandler::class, function () {
             $handler = new ErrorHandler;
-            $handler->registerHandler();
+            $handler->registerHandler(new SerializableErrorHandler);
+            $handler->registerHandler(new ValidationExceptionHandler);
+            $handler->registerHandler(new ModelNotFoundExceptionHandler);
+            $handler->registerHandler(new FallbackExceptionHandler);
         });
     }
 
--- a/src/Api/ApiServiceProvider.php
+++ b/src/Api/ApiServiceProvider.php
@@ -42,7 +42,7 @@ class ApiServiceProvider extends ServiceProvider
             }
         );
 
-        $this->app->singleton(ErrorHandler::class, function() {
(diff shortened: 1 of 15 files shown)


tobscure:
If I remember correctly, I think @franzliedke and I were actually talking a while ago about scrapping the specific handling of `ValidationException` and `ModelNotFoundException`. So only if it implements `JsonApiSerializable`, then we output that information; otherwise, it's an internal server error which the consumer shouldn't know anything about.

The responsibility would be on the throwing code to throw a JsonApiSerializable exception if it wants the consumer to see specific information. e.g. by catching an Illuminate ModelNotFoundException and throwing a Flarum ResourceNotFoundException or whatever.

What do you/@franzliedke think?

kirkbushell:
Yeah makes sense to me. 

Shall I add some custom exceptions that implement that interface and replace the relevant thrown exceptions?

kirkbushell:
tbh I can do that as a separate piece of work anyway.

tobscure:
> Shall I add some custom exceptions that implement that interface and replace the relevant thrown exceptions?

Yep that'd be fantastic.

I'll leave this for @franzliedke to merge :)

franzliedke:
Why the error code 599? I think I don't like that. 

Also, can you typehint the ErrorMiddlewareInterface instead of the concrete middleware in the Client class? 

Once you guys come to an agreement about these (I might be wrong), @tobscure can merge... 

Great work, Kirk!

kirkbushell:
Will sort this out next few days - just been busy with work. Intend to smash this and a heap more tests this weekend.

tobscure:
@kirkbushell So in light of only needing `FallbackExceptionHandler` and `SerializableExceptionHandler`, would this be over-engineering somewhat? (Since it could easily be written as an `if ($e instanceof JsonApiSerializable) { ... } else { ... }`) Probably ignore me, I know I haven't really learned to think in terms of testability yet :P but just thought I'd mention it because simplicity is :+1: 

The status code doesn't matter in a test, does it? Nothing wrong with 599.

Type hinting the ErrorMiddlewareInterface sounds like a good idea to me.

kirkbushell:
There is a case to be made for the error handler, particularly if the user wants more data to be returned as part of the error, or a different format. In most cases, however - the interface implementation would suffice.

kirkbushell:
@tobscure @franzliedke when you get a chance would like to talk more about this, thinking about it some more, handling those exceptions is a pretty nice way to do it, and I think removing the handling of validation/model not found exceptions is regressing the quality of the code.

tobscure:
@kirkbushell I think I agree with you, having separate handlers is more powerful than implementing the JsonApiSerializableInterface (which can't be done for external exceptions).

I would like to move all of the generic JSON-API error handling stuff from this PR into [tobscure/json-api#0.2](https://github.com/tobscure/json-api/tree/0.2). 

* ErrorHandler → Tobscure\JsonApi\ErrorHandler
* ExceptionHandler → Tobscure\JsonApi\Exception\Handler\ExceptionHandlerInterface
* FallbackExceptionHandler → Tobscure\JsonApi\Exception\Handler\FallbackExceptionHandler

The other ExceptionHandlers are Flarum-specific and can stay here.

tobscure:
@kirkbushell any ETA on this / time to have a Skype discussion about it?

kirkbushell:
End of this week should be good.

kirkbushell:
@tobscure moving the json handling into the json api package - however there's a line of code that depends on debug setting. How do you want to handle that in the json api package?

PR #563, opened 2015-10-02
Add "forum" namespacing to previously renamed core keys

dcsjapan:
- Does not affect "core.deleted_user" global string.
- Corresponding YAML will be sent later w/ more extracted strings.


tobscure:
We'll need to add a `core.forum.deleted_user` string, don't we? The non-namespaced keys won't be available in the JavaScript client.

PR #575, opened 2015-10-05
Extract admin strings

dcsjapan:
Adds app.trans calls for strings used by the admin UI.
- Strings for AddExtensionModal.js not included.
- Corresponding YAML will be sent later w/ more extracted strings.


kirkbushell:
Nice :)

dcsjapan:
Thanks! Together with #577 this should take care of the core for a while.

At least until the extension installer is ready. :smile:

PR #577, opened 2015-10-05
Extract leftover core strings

dcsjapan:
Adds app.trans calls for a couple strings in core:
- The "there are no discussions" message in DiscussionList.js
- The user deletion confirmation message in UserControls.js
- Also adds new HTML-style tags to LogInModal.js and SignUpModal.js

### DIFF
--- a/js/forum/src/components/DiscussionList.js
+++ b/js/forum/src/components/DiscussionList.js
@@ -55,7 +55,7 @@ export default class DiscussionList extends Component {
     }
 
     if (this.discussions.length === 0 && !this.loading) {
-      const text = app.trans('core.admin.discussion_list_empty_text');
+      const text = app.trans('core.forum.discussion_list_empty_text');
       return (
         <div className="DiscussionList">
           {Placeholder.component({text})}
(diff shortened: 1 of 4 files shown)


js/forum/src/components/DiscussionList.js
```diff
@@ -55,7 +55,7 @@ export default class DiscussionList extends Component {
     }
 
     if (this.discussions.length === 0 && !this.loading) {
-      const text = 'Looks like there are no discussions here. Why don\'t you create a new one?';
+      const text = app.trans('core.admin.discussion_list_empty_text');
```
tobscure: This should be core.forum?

js/forum/src/utils/UserControls.js
```diff
@@ -95,7 +95,7 @@ export default {
    * Delete the user.
    */
   deleteAction() {
-    if (confirm('Are you sure you want to delete this user? All of the user\'s posts will be deleted.')) {
+    if (confirm(app.trans('core.admin.user_controls_delete_confirmation'))) {
```
tobscure: This should be core.forum?

dcsjapan:
Sorry, yes to both of those. I'll revise and resend.

Issue #262, opened 2015-08-27
Extract some low-level code into independent open-source packages


tobyzerner:
To reduce Flarum's scope and make it more maintainable, we should consider extracting some application-agnostic code into external packages. Specifically:
- [ ] Our component paradigm that sits on top of Mithril (i.e. the base `Component` class).
- [ ] Some of the Mithril utilities, like `classList` and `SubtreeRetainer`.
- [ ] The whole JavaScript data layer: `Store` and `Model`.
- [x] More of the JSON-API layer, specifically the Actions and Request/Response stuff may be within the scope of the tobscure/json-api package.

Keen to get some thoughts on whether or not this is a good idea, and more than happy for anyone to have a go at extracting some stuff.

PR #594, opened 2015-10-20
Add third tier to key namespacing

dcsjapan:
- Changes all `app.trans` calls to `app.translator.trans` calls.
- Changes existing keys to [three-tier namespace structure](https://github.com/flarum/english/pull/12).
- Extracts additional strings for `lib:` namespace.
- Extracts two previously missed strings for EditGroupModal.js.

### DIFF
--- a/js/forum/src/utils/DiscussionControls.js
+++ b/js/forum/src/utils/DiscussionControls.js
@@ -55,14 +55,14 @@ export default {
         !app.session.user || discussion.canReply()
           ? Button.component({
             icon: 'reply',
-            children: app.trans(app.session.user ? 'core.forum.discussion_controls_reply_button' : 'core.forum.discussion_controls_log_in_to_reply_button'),
+            children: app.translator.translator.trans(app.session.user ? 'core.forum.discussion_controls.reply_button' : 'core.forum.discussion_controls.log_in_to_reply_button'),
             onclick: this.replyAction.bind(discussion, true, false)
           })
           : Button.component({
             icon: 'reply',
-            children: app.trans('core.forum.discussion_controls_cannot_reply_button'),
+            children: app.translator.trans('core.forum.discussion_controls.cannot_reply_button'),
             className: 'disabled',
-            title: app.trans('core.forum.discussion_controls_cannot_reply_text')
+            title: app.translator.trans('core.forum.discussion_controls.cannot_reply_text')
           })
       );
     }
@@ -85,7 +85,7 @@ export default {
     if (discussion.canRename()) {
       items.add('rename', Button.component({
         icon: 'pencil',
-        children: app.trans('core.forum.discussion_controls_rename_button'),
+        children: app.translator.trans('core.forum.discussion_controls.rename_button'),
         onclick: this.renameAction.bind(discussion)
       }));
     }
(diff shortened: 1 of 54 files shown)


js/forum/src/utils/DiscussionControls.js
```diff
@@ -55,14 +55,14 @@ export default {
         !app.session.user || discussion.canReply()
           ? Button.component({
             icon: 'reply',
-            children: app.trans(app.session.user ? 'core.forum.discussion_controls_reply_button' : 'core.forum.discussion_controls_log_in_to_reply_button'),
+            children: app.translator.translator.trans(app.session.user ? 'core.forum.discussion_controls.reply_button' : 'core.forum.discussion_controls.log_in_to_reply_button'),
```
franzliedke: Whoopsie...

js/lib/utils/abbreviateNumber.js
```diff
@@ -11,9 +11,9 @@
 export default function abbreviateNumber(number) {
   // TODO: translation
   if (number >= 1000000) {
-    return Math.floor(number / 1000000) + 'M';
+    return Math.floor(number / 1000000) + app.translator.trans('core.lib.number_suffix.mega_text');
```
franzliedke: Can we be sure this is always a suffix?

franzliedke:
Syntax looks good, except for that one line I commented on...

I did not cross-check all the keys, because I trust @dcsjapan. :)

js/lib/utils/abbreviateNumber.js
tobscure: Good point.

js/forum/src/utils/DiscussionControls.js
dcsjapan: Whoopsie indeed! Thanks for catching that.

I started out by searching for "app.trans" instead of "app.trans(" and ended up applying the same change twice in a couple places. I thought I had got all of them. Sorry about that!

js/lib/utils/abbreviateNumber.js
dcsjapan: That's a good question. I guess we *could* apply `app.translator` to the whole thing...

But I get the feeling that the computer age is forcing its own syntax on the world, so expressions such as "1KB" end up getting used as-is regardless of how things were originally done in any given language. My guess is that we're safe enough leaving this one as a prefix for now.

tobscure: OK. I'll merge these PRs in tomorrow :+1:

franzliedke: You're the boss. ;)

dcsjapan:
Thanks for the trust ... I'm not sure I trust me that much. :wink:

Nearly all of the keys are practically unchanged, with the exception of one underbar changed to a period. The only exceptions are the ones noted in the PR description. Oh, and I changed the namespacing for the extensions loading modal string from `extensions_loading_` to simply `loading:` because it seemed possible that someone might want to reuse this modal at some point.

I wouldn't rule out the possibility that I screwed something up, since I was doing these before my morning coffee, but I've tried to keep screw-ups to a minimum.

kirkbushell:
Why not have the translator promoted to its own object?

tobscure:
@kirkbushell Assuming you mean the Translator class – good thought, see #262. It's basically just a JS implementation of Symfony's Translator anyway, so would be a good fit for extraction.

PR #606, opened 2015-10-26
Use exception handlers instead of JsonApiSerializableInterface

tobscure:
@kirkbushell Just wanted to see what you thought of this (basically copied from your PR) before I merge it in.

### DIFF
--- a/src/Api/Handler/FloodingExceptionHandler.php
+++ b/src/Api/Handler/FloodingExceptionHandler.php
@@ -1,0 +1,38 @@
+<?php
+/*
+ * This file is part of Flarum.
+ *
+ * (c) Toby Zerner <toby.zerner@gmail.com>
+ *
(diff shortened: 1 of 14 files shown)


src/Api/Handler/FloodingExceptionHandler.php
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+namespace Flarum\Api\Handler;
+
+use Exception;
+use Flarum\Core\Exception\FloodingException;
+use Tobscure\JsonApi\Exception\Handler\ExceptionHandlerInterface;
+use Tobscure\JsonApi\Exception\Handler\ResponseBag;
+
+class FloodingExceptionHandler implements ExceptionHandlerInterface
```
kirkbushell: This is cool - didn't see this before. Nice way to handle it.

PR #628, opened 2015-11-06
[beta4]Fixed login input fields have different style

binaryoung:
In beta4,login input fields have different width , font size and text-align.This PR fixes the problem.


franzliedke:
Couldn't we use `<input type="email">`?

binaryoung:
Actually we can't use `<input type="email">`,because users maybe use username or email to log in.If we add `type="email"` to input tag,we can't use username to log in any more.

franzliedke:
Whoopsie, true, I forgot that. Thanks for the PR!

PR #642, opened 2015-11-25
Fixed #627

### LINKED ISSUE #627: Sorting "Oldest" does not work
For now (01.0 Beta 4), it's impossible to sort discussions with the "Oldest" option. An error occurs, and it prevents to select another option. It can be tested on the Flarum community: https://discuss.flarum.org/?sort=oldest

binaryoung:
In tobscure/json-api we use "startTime" not "+startTime" to represent asc.So this PR fixs the bug #627.

### DIFF
--- a/js/forum/src/components/DiscussionList.js
+++ b/js/forum/src/components/DiscussionList.js
@@ -115,7 +115,7 @@ export default class DiscussionList extends Component {
     map.latest = '-lastTime';
     map.top = '-commentsCount';
     map.newest = '-startTime';
-    map.oldest = '+startTime';
+    map.oldest = 'startTime';
 
     return map;
   }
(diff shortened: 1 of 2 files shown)


franzliedke:
This should be fixed at the JSON-API repo instead. Please comment on the PR you sent there if you need help with the correct fix. :)

I believe [this commit](https://github.com/tobscure/json-api/commit/fba08d85ee935279e6a33bdb93d272ef95dff582#diff-f2d22b3177b18140dda397efe4e82770) introduced the issue - see the previous implementation in Criteria.php (that's where that link goes).

tobscure:
Actually, this PR is correct. The [relevant part of the JSON-API spec](http://jsonapi.org/format/#fetching-sorting) has nothing about a `+` prefix (an older version did, but 1.0 final doesn't). The tobscure/json-api repo is working as expected.

PR #647, opened 2015-11-30
Add third-level namespacing to deleted_user_text

dcsjapan:
Just a small change for consistent namespacing in `lib`.


dcsjapan:
Whoops ... brain's not working right this morning. I'll try that again.

PR #663, opened 2015-12-05
Add base OAuth2 controller API

tobscure:
This PR adds an abstract controller which makes it much simpler for extensions to implement external authentication using the league/oauth2-client package. Previously all of this code has been duplicated between the Facebook/GitHub extensions.

Technically this doesn't belong in core, because it's only useful for pretty specific cases – Twitter auth can't use it, for example, because that uses league/oauth1-client. In that respect it's more like a utility than an API, and should be in its own package (flarum/oauth2-controller) which extensions could then require.

My hesitation in doing this is that as we start adding "utility" packages like these to the flarum organisation, they become mixed in with our extension repos. So I'm wondering if it'd be best to prefix our extension repos with `flarum-`, just like we recommend other vendors do. We'd end up with repos like:

* flarum/core
* flarum/gulp
* flarum/oauth2-controller
* flarum/flarum-tags
* flarum/flarum-sticky
* flarum/flarum-auth-facebook

This might also be good because it's consistent with what we recommend for other vendors in terms of naming Flarum extensions. 

Any thoughts about all of this?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,6 +35,7 @@
         "illuminate/validation": "5.1.*",
         "illuminate/view": "5.1.*",
         "league/flysystem": "^1.0.11",
+        "league/oauth2-client": "~1.0",
         "tobscure/json-api": "^0.2.0",
         "oyejorge/less.php": "~1.5",
         "intervention/image": "^2.3.0",
--- a/src/Forum/Controller/AbstractOAuth2Controller.php
+++ b/src/Forum/Controller/AbstractOAuth2Controller.php
@@ -1,0 +1,94 @@
+<?php
+/*
+ * This file is part of Flarum.
+ *
+ * (c) Toby Zerner <toby.zerner@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Flarum\Forum\Controller;
+
+use Flarum\Forum\AuthenticationResponseFactory;
+use Flarum\Http\Controller\ControllerInterface;
+use League\OAuth2\Client\Provider\ResourceOwnerInterface;
+use Psr\Http\Message\ServerRequestInterface as Request;
+use Zend\Diactoros\Response\HtmlResponse;
+use Zend\Diactoros\Response\RedirectResponse;
+
+abstract class AbstractOAuth2Controller implements ControllerInterface
+{
+    /**
+     * @var AuthenticationResponseFactory
+     */
+    protected $authResponse;
+
+    /**
+     * @param AuthenticationResponseFactory $authResponse
+     */
+    public function __construct(AuthenticationResponseFactory $authResponse)
+    {
+        $this->authResponse = $authResponse;
+    }
+
+    /**
+     * @param Request $request
+     * @return \Psr\Http\Message\ResponseInterface|RedirectResponse
+     */
+    public function handle(Request $request)
+    {
+        $redirectUri = (string) $request->getUri()->withQuery('');
+
+        $provider = $this->getProvider($redirectUri);
+
+        $session = $request->getAttribute('session');
+
+        $queryParams = $request->getQueryParams();
+        $code = array_get($queryParams, 'code');
+        $state = array_get($queryParams, 'state');
+
+        if (! $code) {
+            $authUrl = $provider->getAuthorizationUrl($this->getAuthorizationUrlOptions());
+            $session->set('oauth2state', $provider->getState());
+
+            return new RedirectResponse($authUrl.'&display=popup');
+        } elseif (! $state || $state !== $session->get('oauth2state')) {
+            $session->forget('oauth2state');
+            echo 'Invalid state. Please close the window and try again.';
+            exit;
+        }
+
+        $token = $provider->getAccessToken('authorization_code', compact('code'));
+
+        $owner = $provider->getResourceOwner($token);
+
+        $identification = $this->getIdentification($owner);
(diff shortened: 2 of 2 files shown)


Ernest0x:
What about prefixing everything with flarum like this:

* flarum/flarum
* flarum/flarum-core
* flarum/flarum-gulp
* flarum/flarum-oauth2-controller
* flarum/flarum-ext-tags
* flarum/flarum-ext-sticky
* flarum/flarum-ext-auth-facebook

Luceos:
Now I understand your preference @tobscure for prefixing all packages with flarum-. I couldn't understand why you would any developer to force to that convention, because - in my opinion - it's truly a matter of taste. The fact that you support that prefixing (must be said without breaking the installation wow!) is really appreciated, because it mainly helps keep the repositories nicely seperated from non-flarum repositories. This in fact is the only reason I have been prefixing the flarum extensions.

Returning to the point. I do not feel strongly about your suggested approach, because for me the package name should closely relate the repository name for convenience purposes. Also I'm not sure how obvious it is in the Flarum namespace to prefix extensions with Flarum.. As @Ernest0x says maybe you can use `ext-` for extension on the Flarum extension. Maybe a seperate vendor might help with keeping extensions divided from core functionality?

franzliedke:
I like the idea of a separate "FlarumExtensions" organization. Polymer does the same with "PolymerElements" if I'm not mistaken.

That way, we have a very clear separation, and can avoid these awkward prefixes.

Ernest0x:
> I like the idea of a separate "FlarumExtensions" organization. Polymer does the same with "PolymerElements" if I'm not mistaken.
> 

Will that organization really exist? Will there be a separate team working on official extensions perhaps in a different way than the team working on core functionality?

Luceos:
> Will there be a separate team working on official extensions perhaps in a different way than the team working on core functionality?

That's also an interesting benefit. Even though any organization can create teams and seperate access based on those already.

tobscure:
@franzliedke I see where you're coming from with the separate extension organisation. [phpBB do it too.](https://github.com/phpbb-extensions) But I'm not sold on it for Flarum, given that our extensions are not just little extra add-ons, but they're really a core part of the project – as much as flarum/core itself. I feel like code that is that integral to the project should live under the main organisation. Not to mention all our issues are centralised in flarum/core – does cross-organisation issue tracking even work?

I think prefixes will be fine if we can devise a good system.

> @Luceos it [the flarum- prefix] mainly helps keep the repositories nicely seperated from non-flarum repositories. This in fact is the only reason I have been prefixing the flarum extensions.

Exactly :)

I think to reduce confusion for third-party devs, we shouldn't give the flarum organisation any special treatment when it comes to repo prefixes. In other words, whatever we choose to prefix extensions with, it should be consistent with what third-party developers can (and should) prefix their Flarum extension repos with.

So it should really contain `flarum-`, even though our vendor name is already flarum. I wouldn't be opposed to making the extension part explicit with a `flarum-ext-` prefix. In fact, @JoshyPHP is already using this prefix for his [s9e/flarum-ext-mediaembed](https://github.com/s9e/flarum-ext-mediaembed) extension.

Getting back to the crux of this PR, perhaps it might also be beneficial to have a `util-` prefix for things that aren't standalone packages? So to refresh my original examples:

* flarum/core
* flarum/gulp
* flarum/util-oauth2-controller
* flarum/flarum-ext-tags
* flarum/flarum-ext-sticky
* flarum/flarum-ext-auth-facebook

Here's another project which organises repos like this, and I think it works quite nicely: https://github.com/getgrav

Ernest0x:
> Here's another project which organises repos like this, and I think it works quite nicely: https://github.com/getgrav

Yeap. They are also prefixing everything with 'grav', as I suggested for flarum too. So, one can fork/clone 'flarum/flarum-core' and have their own, pretty self-describing, 'flarum-core' repository.
Unprefixed 'core' may make sense when it is under the 'flarum' organization, but doesn't look right when it is cloned under another organization or personal account. I know it is possible to rename it afterwards (or clone it in a different path), and some people may be actually doing that, but that's another step (or thing to think of) that can be avoided.

Also, unprefixed 'gulp' is a bit misleading. It's just gulp tasks and configuration for compiling flarum extensions, not a fork of gulp itself maintained by the flarum organization for some reason.

franzliedke:
Hmm, the issue argument convinced me.

Let's do it like Grav. :)

tobscure:
Decided in the end to just merge this into core, as I think OAuth2 is going to be a pretty common requirement for auth extensions. We can always extract into another repo later if necessary.

Still yet to give extensions the `flarum-ext-` prefix and rename `flarum/gulp`.

tobscure:
Repo renaming is done! Installation of beta 4 still works fine; for beta 5 upgrade we'll need to tell everyone to update their `composer.json` files.

tarunmarkose:
At this stage, can we use this controller to authenticate against a wordpress installation with a OAuth2.0 Server plugin like [this one](https://wordpress.org/plugins/oauth2-provider/)? If not, what else is required to be done?

tobscure:
@tarunmarkose I believe so, yes. You should be able to extend the AbstractOAuth2Controller and implement the abstract methods – specifically, you'll want to return a [GenericProvider](https://github.com/thephpleague/oauth2-client#usage) from the getProvider method. Feel free to comment if you need any help.

tarunmarkose:
Thanks @tobscure. I'll work with the flarum-ext-auth-facebook extension as an example, and comment if I get stuck anywhere.

PR #664, opened 2015-12-05
Applied fixes from StyleCI

tobscure:
What rules should we get StyleCI to check? The easiest thing to do would be to use their "recommended" preset, but that means a few alterations to our current style, as shown in this PR.

Otherwise, if we want to keep our current style, we can add the appropriate rules to StyleCI config.

What do we think?


franzliedke:
I'm probably not too fussed on mot of these things. Two things I noticed from a quick glance:

- I like spaces between strings and the concatenation operator.
- Same for boolean negation.

Luceos:
I'd either use Laravel or the default which adheres to psr. Do not go for any custom combination, that feels like bad practice.

Luceos:
I think we should pick one and stick to it, I also think we should pick a generic one. I know this has little priority, but it would be a good feeling to decide on this and move on to more pressing matters :smile:

tobscure:
@franzliedke and I decided that we'll stick with what we've currently got and add the necessary rules to StyleCI config. It's on my todo list.

Issue #254, opened 2015-08-27
Add user online indicators


tobscure:
* Similar to esoTalk, if a user is online (seen in the last 5 mins), we want to show a little green circle just to the left of the username.
* This should be added to the PostHeaderUser component.
* Probably use an fa-circle icon, with a style to make it a bit smaller just like in the UserCard component.

dcsjapan:
In connection with this, consider matching the color and style of the online indicator on the user page and profile popup to the above, as per this suggestion:

http://discuss.flarum.org/d/902-green-circle-to-display-user-is-online

Luceos:
See PR: https://github.com/flarum/core/pull/676

PR #676, opened 2015-12-16
User online indicators

petermein:
My first contribution to Flarum. I just picked one of the issues good for new contributors.

The issue:
https://github.com/flarum/core/issues/254

I added a helper to display the icon when the condition is met. Added styling from the UserCard.


franzliedke:
This looks great, thank you very much! I will pull this down to test it tonight.

(The failing StyleCI analysis is not your fault. Toby is working on fixing that...)

Luceos:
Referencing: https://github.com/flarum/core/issues/254

tobscure:
Looks good, thanks!

Nitpicking, but it might be better to consolidate this directly into the PostUser component since it's a unique context and unlikely to be reused anywhere else... i.e.

```js
<div className="PostUser">
  {user.lastSeenTime() && user.isOnline() ? <span className="PostUser-online">{icon('circle')}</span> : ''}
```

Luceos:
@tobscure I'm not sure about that. I would like to be able to use such a state icon seperately from the whole user component, for instance is table listing. Not sure whether you meant that seperation.

One thing I saw: `useronline` should be according to the naming conventions `UserOnline` as a component name in the helpers directory.

petermein:
@Luceos The naming is the same as the username helper class. But indeed maybe it should be userOnline to match the rest of the helper classes. 

Just tell what option you think is best, the helper approach or put the logic inside the component. Il update the PR accordingly.

franzliedke:
Let's go with a small helper, that's what they're there for.

@petermein Please rename to `userOnline` and I'll merge. :)

Luceos:
Changed to camelCasing, let's merge this @flarum/core

tobscure:
:beers:

Issue #197, opened 2015-07-28
Make PostStreamScrubber display numbers relating to only "comment" posts


Fastidious:
Setting a post as sticky is adding to the count of posts. For example, if a post (a single one) is set to sticky, the sticky "action" will count as a comment or reply. See the [corresponding thread](http://discuss.flarum.org/d/407/stickies-count-as-a-post/1).

Fastidious:
Adding to this. It seems any action taken on the original post (adding or removing a label, for example), will add more "post" count. [This post](http://discuss.flarum.org/d/392/frequently-asked-questions-please-read-here-before-posting/1), for example, now has a post count of three, even though it is only one.

tobscure:
Steps to reproduce:

1. Go to http://discuss.flarum.org/d/392/frequently-asked-questions-please-read-here-before-posting/1
2. Scroll down to the bottom
3. Observe that the scrollbar says "1 of 3 posts"

Expected behaviour:
It should say "1 of 1 post"

Technical info:
In Flarum, there are different "types" of posts. Typical posts that users create are of the `comment` type. Other post types may reflect events within the discussion, such as changing the title (`discussionRenamed`) or stickying the discussion (`discussionStickied`).

The PostStreamScrubber, the scrollbar showing the current position within the discussion, displays the index of the last post visible in the viewport, as well as the total number of posts (e.g. "1 of 3 posts"). However, it doesn't take post type into consideration, and displays the index/count of all posts. It would be better if it only displayed the numbers relating to `comment` posts.

What needs to be done:

* In the PostStream component, add a data-type attribute to stream items which contains the post's type.
* In the PostStreamScrubber component's `update` method, when looping through stream items, check on the item's data-type attribute and work out the index of the last visible comment post.
* Change the PostStreamScrubber component's `view` to display that index instead of `this.visibleIndex()`, and the discussion's `commentsCount` instead of `this.count()`.

PR #687, opened 2015-12-20
Make PostStreamScrubber display numbers relating to only "comment" posts

ahsanity:
PR for #197.

This needs review as my expertise in JS / ES6 / JSX / Mithril is very limited. This is more an attempt to explore the frontend codebase. 

I would be glad to make improvements and changes if required.

### DIFF
--- a/js/forum/src/components/PostStream.js
+++ b/js/forum/src/components/PostStream.js
@@ -212,6 +212,7 @@ class PostStream extends Component {
         attrs['data-time'] = time.toISOString();
         attrs['data-number'] = post.number();
         attrs['data-id'] = post.id();
+        attrs['data-type'] = post.contentType();
 
         // If the post before this one was more than 4 hours ago, we will
         // display a 'time gap' indicating how long it has been in between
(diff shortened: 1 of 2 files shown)


tobscure:
Thanks for taking this on!

Just tried it out – it's a good start but not quite there. The issue is that we only want to change the numbers that are displayed on the scrubber, without having an impact on the calculations of the scrubber's positioning. So instead of changing how the `count` is retrieved and skipping over non-`comment` items in the loop altogether, try:

* Changing the `.Scrubber-count` in the `view` to be the discussion's `commentsCount` instead of the stream's count
* Add a new instance variable (something like `commentsIndex`). In the `update` loop, check on the item data-type attributes and increment the `commentsIndex` for every comment post until the last visible one.
* Then render `commentsIndex` into `.Scrubber-index` in the `view` instead of `visibleIndex`, and also in `renderScrollbar`. This means we can also get rid of the `visibleIndex` method completely.

Does that make sense? Please let me know if you have any questions!

ahsanity:
Yup, makes sense! I will take a bash at it later in the week. Thanks!

ahsanity:
@tobscure have a look now. Thanks for such clear directions, you made it too easy ;)

ahsanity:
Squashed my commits.

Luceos:
@tobscure any feedback on this?

tobscure:
Looks good, just need to test when I get a chance.

tobscure:
Perfect, thanks!

ahsanity:
Coolio!

Issue #256, opened 2015-08-27
Implement more gambits


tobscure:
Gambits are little pieces of search criteria that can be specified in the search string. Just like GitHub's [search syntax](https://help.github.com/articles/search-syntax/).

Right now we've only implemented a few: `is:unread` and `author:username` in core, and `tag:foo,bar`, `is:followed`, `is:ignored`, `is:sticky`, and `is:locked` in their respective extensions.

We probably want to implement the following gambits:

* Multiple author search, separated by commas (`author:username1,username2`)
* `created:date` where `date` is the same syntax as on GitHub
* `updated:date`
* `contributor:username1,username2`
* `replies:number` where `number` is the same syntax as on GitHub

Pending #XX, we will also want to have an `in:title` gambit which scopes fulltext search to only look in discussion titles.

This is a precursor to an extensible advanced search interface which will come later.

Albert221:
See #703.

BartVB:
Can I put in a request for searching by IP address? We use that to track down spammers and their posts. Obviously this should be behind some kind of permission.

franzliedke:
@BartVB Good idea. If @tobscure agrees, we can add that to the list at the top of this issue.

luceos:
Contributor or participant is related to the author of a post, contrary to the current Author Gambit for Discussions.

askvortsov1:
@jordanjay29 I'm assuming that the replies gambit is just for sorting, but if it's for filtering, I would propose splitting it into replies_gte and replies_lte (although not sure what the use case for that would be).

askvortsov1:
This post was made a while ago, and hasn't been updated recently. The following have already been updated:

- Multiple author search
- Multiple contributor search (PR)
- Time Created
- Time Last Updated (PR)
- Improvements to fulltext title searching

There are a few more things to consider:
- Should the created and uploaded gambits be updated offer better support for "before" and "after"? eg: `created:before:2020-01-04`
- Do we even want a replies # gambit? This feels like something that would definitely need to be greater than or less than, I can't imagine searching for posts with exactly n replies
- Do we want to consider permissions in gambits? While it would be nice to search by IP address, that would mean hardwiring in a big chunk of new logic which, frankly, feels like overkill.

luceos:
[...]

PR #703, opened 2016-01-03
#256 Added multiple author search gambit

Albert221:
It's my first contribution BTW.


Albert221:
As of [this](http://stackoverflow.com/questions/9476332/sql-statements-with-equals-vs-in) `WHERE id IN (x)` is equal to `WHERE id = x` in terms of speed, so there's no need to handle the case when there's only 1 author independently.

tobscure:
Fantastic, thanks! :beers:

PR #705, opened 2016-01-04
#256 Added created gambit

Albert221:
Pattern matches ISO8601 which is used on GitHub, but passes with i.e. 25th hour and this need to be fixed later.

### DIFF
--- a/src/Core/Command/EditUserHandler.php
+++ b/src/Core/Command/EditUserHandler.php
@@ -83,6 +83,10 @@ class EditUserHandler
             }
         }
 
+        if ($actor->isAdmin() && ! empty($attributes['isActivated'])) {
+            $user->activate();
+        }
+
         if (isset($attributes['password'])) {
             $this->assertPermission($canEdit);
             $user->changePassword($attributes['password']);


Albert221:
Sorry for this commit mess, I'm not familiar with sending PRs.

src/Core/Search/Discussion/Gambit/CreatedGambit.php
```diff
@@ -0,0 +1,40 @@
...
+use Flarum\Core\Search\AbstractSearch;
+use Flarum\Core\Search\Discussion\DiscussionSearch;
+use LogicException;
+
+class CreatedGambit extends AbstractRegexGambit
+{
+    /**
+     * http://stackoverflow.com/a/8270148/3158312
+     *
+     * @FIXME: Find better regular expression
```
franzliedke: What isn't good about this one?

franzliedke:
You're on a roll. Thanks a lot!

Some notes:
- You should always create a separate branch off master before sending a pull request - that way, your master branch can stay in sync with ours, whether or not your PR gets merged.
- Please [squash](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html) your four commits into one and then force push your master branch. This PR will be updated automatically.

src/Core/Search/Discussion/Gambit/CreatedGambit.php
Albert221: It passess for values like 25th hour or 70th minute. Regexes are not my good side so I only tested a few regexes and this will work fine too, it will just generate query which returns nothing (in case like 25th hour), but I think that there should be more strict pattern.

franzliedke: Okay, cool. Which formats do you want to support? I can probably help you.

Albert221: I firstly wanted [this](http://stackoverflow.com/questions/28020805/regex-validate-correct-iso8601-date-string-with-time), but with optional *parameters*. In that regex every part is required to pass.

tobscure:
Thanks for contributing!

Do we really want to use ISO8601 for gambit dates? It's not very human-friendly – gambits should be easy to type. What about a simple `YYYY-MM-DD` format like [GitHub search tokens](https://help.github.com/articles/searching-issues/#search-based-on-commit-status)? We also should support date ranges joined by `..` because there's not much utility in searching for an exact creation date/time.

Albert221:
OK, I tested locally everything and I've changed regex pattern. With previous pattern when we'd type date with time (i.e. 2015-07-08T14:25) `whereDate` method would return nothing since it compares date to that datetime string. I've just leaved date regex and added this range support as you said.

src/Core/Search/Discussion/Gambit/CreatedGambit.php
```diff
@@ -0,0 +1,42 @@
...
+use LogicException;
+
+class CreatedGambit extends AbstractRegexGambit
+{
+    /**
+     * http://stackoverflow.com/a/8270148/3158312
+     *
+     * {@inheritdoc}
+     */
+    protected $pattern = 'created:(\d{4}\-\d\d\-\d\d)(..(\d{4}\-\d\d\-\d\d))?';
```
franzliedke: Please escape the periods.

Issue #696, opened 2015-12-29
UrlGenerator without prefix


LzIManD:
> core/src/Api/UrlGenerator.php

    class UrlGenerator extends AbstractUrlGenerator
    {
        protected $prefix = 'api';
    }
> core/src/Api/Controller/ListDiscussionsController.php

    87    $document->addPaginationLinks(
            $this->url->toRoute('discussions.index'),
            $request->getQueryParams(),
            $offset,
            $limit,
            $results->areMoreResults() ? null : 0
    93  );

but,
get http://localhost:88/api/discussions?include=startUser&page%5Boffset%5D=0,
not prefix , just  

    ...
    "links": {
        "first": "https://discuss.flarum.org/discussions?include=startUser",
        "next": "https://discuss.flarum.org/discussions?include=startUser&page%5Boffset%5D=20"
     },
    ...

Albert221:
I have reproduced this issue and it actually occurs.

tobscure:
Merged #706

PR #706, opened 2016-01-04
#696 Added support for prefixes in AbstractUrlGenerator.

Albert221:


franzliedke:
Hmm, do we even need `toPrefix()`? Shouldn't `toBase()` work just like your `toPrefix()` implementation?

Albert221:
Fixed as you said

PR #713, opened 2016-01-05
Applied fixes from StyleCI

tobscure:
This pull request applies code style fixes from an analysis carried out by [StyleCI](https://styleci.io).

---

For more information, click [here](https://styleci.io/analyses/zGrMlq).


tobscure:
@franzliedke How's this?

franzliedke:
Looks good.

The only thing I don't like, I think, is the alignment of array assignments:

~~~php
[
  'code'  => 503,
  'title' => 'Service Unavailable',
]
~~~

Not aligning them is easier to type, easier to change and easier on diffs.

Luceos:
@franzliedke but readability is hugely improved by aligning arrays, especially if there are many entries (like a config file).

franzliedke:
I think it's fine for such cases, but for normal code, most arrays look like the one I quoted above. I'd rather have them easy to edit (and diff) than slightly more beautiful for the eye. But I'm willing to bend to democracy ;)

tobscure:
I agree with @franzliedke, let's not *enforce* it either way. So arrays such as those in Api\Serializers can be aligned, but other, simpler ones don't have to be. I'll update the StyleCI config soon then merge a new PR.

maelsoucaze:
This should be solve the alignment of array assignments:

```yaml
preset: recommended

risky: false

linting: true

enabled:
  - unalign_double_arrow

disabled:
  - align_double_arrow
```

Plus, I think it's better to put this configuration into a _.styleci.yml_ and to untouch the online configuration. That's always be useful, for extension developer for example, to use the same coding guidelines as yours. The online configuration is not available for people who're not in the Flarum organization.

franzliedke:
> Plus, I think it's better to put this configuration into a .styleci.yml file and to untouch the online configuration. 

I totally agree. :+1:

maelsoucaze:
Just a quick note concerning:

```yaml
enabled:
  - unalign_double_arrow
```

It will unalign **all** alignment of array assignments **everywhere**, but I saw you want to keep some of them (concerning Api\Serializers). On my side, I like when everything is standardized, it's easier to check, but you can of course remove that if you want some exceptions. But that's a risk to not see when someone will align or unalign some array assignments in wrong locations.

Luceos:
I'm still not convinced why we should not simply apply a preset like psr-2.. We all have a different taste, that's why such a standardization is so welcome. By applying some custom setup you will only create difficulty with other devs to contribute. I hope you'll take that into consideration.

franzliedke:
But we *are* doing PSR-2, no? We just have some additional rules, because PSR-2 doesn't specify everything...

maelsoucaze:
Of course, **psr2** preset [is available](https://styleci.readme.io/docs/presets#psr2). However, fixers are really few. But I totally agree with you @Luceos, it's better to use a common preset well known (or correctly display our coding guidelines with a file and a proper documentation).

Luceos:
There you have it @maelsoucaze 

> coding guidelines and proper documentation

Someone has to write and maintain, how easy is it to simply refer to PSR-2 documentation? That's why I mention "personal taste". Why force a personal taste (or additions) over the project as a whole, it will only generate more workload by explaining it all or documenting it. That's why I initiate the discussion.

@franzliedke ; additions to a preset still have to be documented.. Would you really want to bother with that? Don't tell me you can't get used to PSR-2 or any other preset in the long run. As a very seasoned dev - and someone I highly respect - don't you agree that conventions should follow the goals of a project? In that sense, wouldn't a generalized convention be more helpful in attracting more contributors in the long run?

If you truly want to perfect formatted code, shouldn't you also require all multi-line phpdoc comments to lose 3 characters in each line like [Taylor does](https://www.reddit.com/r/laravel/comments/3bifw5/every_doc_block_comment_is_3_characters_less_than/) :sweat_smile:

franzliedke:
Well, we document them in a StyleCI file.

And any additional rule is simply on top of PSR-2, for situations that aren't fully clarified by it.

Luceos:
Would love to see a decision about this :+1: and get it merged so that the checks will give a proper state in the future.

tobscure:
Merged in #831

Issue #719, opened 2016-01-07
Opening the admin panel while logged out causes a error.


MrPowerGamerBR:
![http://i.imgur.com/q8d8833.png](http://i.imgur.com/q8d8833.png)

Maybe I'm wrong, but shouldn't trying to open the admin panel while logged out display a error message instead of showing a "PrettyPageHandler"?

If debug is disabled, this message is shown:

![http://i.imgur.com/9An0Oy8.png](http://i.imgur.com/9An0Oy8.png)

Forum thread: https://discuss.flarum.org/d/1784-opening-the-admin-panel-while-logged-out-causes-a-error/6

franzliedke:
Thanks a lot! This one is indeed not intended.

franzliedke:
What needs to be done:

- Adapt the `PermissionDeniedException` to have a default code of `403`. This can be done similar to the current [`RouteNotFoundException` implementation](https://github.com/flarum/core/blob/b83c81c06e78aba5f4f5528e960ddff456826368/src/Http/Exception/RouteNotFoundException.php).
- Add a simple template for this error to the [error directory](https://github.com/flarum/core/tree/b83c81c06e78aba5f4f5528e960ddff456826368/error).
- Update the [`HandleErrors` middleware](https://github.com/flarum/core/blob/417b7f7972352f7bcc1afb75948b8e4877a80d82/src/Http/Middleware/HandleErrors.php#L55): exceptions with status code 403 should always return a pretty page, too.

Albert221:
I'm taking this. What shall happen on debug? Whoops or this HTML page?

PR #720, opened 2016-01-07
#719 Fixed PermissionDeniedException

Albert221:
...causing Whoops on debug and 500 HTTP error
instead of 403 Forbidden error page.

### DIFF
--- a/src/Core/Exception/PermissionDeniedException.php
+++ b/src/Core/Exception/PermissionDeniedException.php
@@ -14,4 +14,8 @@ use Exception;
 
 class PermissionDeniedException extends Exception
 {
+    public function __construct($message = null, $code = 403, Exception $previous = null)
+    {
+        parent::__construct($message, $code, $previous);
+    }
 }
--- a/src/Http/Middleware/HandleErrors.php
+++ b/src/Http/Middleware/HandleErrors.php
@@ -52,7 +52,7 @@ class HandleErrors implements ErrorMiddlewareInterface
             $status = $errorCode;
         }
 
-        if ($this->debug && $errorCode !== 404) {
+        if ($this->debug && ! in_array($errorCode, [403, 404])) {
             $whoops = new WhoopsMiddleware;
 
             return $whoops($error, $request, $response, $out);
--- a/error/403.html
+++ b/error/403.html
@@ -1,0 +1,13 @@
+<!DOCTYPE html>
+<html>
+<head lang="en">
+    <meta charset="UTF-8">
+    <title></title>
+</head>
+<body>
+
+<h1>403 Forbidden</h1>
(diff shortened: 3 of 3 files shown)


src/Core/Exception/PermissionDeniedException.php
```diff
@@ -1,4 +1,5 @@
 <?php
+
```
franzliedke: Uh oh.

Please fix and squash. :P

franzliedke:
Wonderful!

Please fix the empty line, then I'll merge. :)

src/Http/Middleware/HandleErrors.php
```diff
@@ -52,7 +52,7 @@ public function __invoke($error, Request $request, Response $response, callable
             $status = $errorCode;
         }
 
-        if ($this->debug && $errorCode !== 404) {
+        if ($this->debug && ! in_array($errorCode, [403, 404])) {
```
tobscure: Hmm I don't like this change... if debug mode is on, it's handy to have Whoops to see where the PermissionDeniedException has been thrown from.

franzliedke: But isn't this the expected behaviour? When a user tries to access a site they can not access, they should be told so, no matter whether we're in debug mode or not.

franzliedke: If need be, we can include a (pretty-ish) stack traces on any error pages where it might make sense, if debug mode is enabled.

tobscure: I think one of the main points of debug mode is that Exceptions are exposed, with all debug information, and Whoops is the perfect way to do that. 404 is an exception (pun intended) to the rule because they almost always original from the same place – the route dispatcher. Users shouldn't have debug mode on unless they are developing or debugging an issue, in which case showing as much information as possible is necessary and pretty error pages are not.

franzliedke: To me, this behavior is a feature. When somebody tries to access a page that they are not allowed to see, they should get a pretty and helpful error page saying so. 

In debug mode, you want to quickly find information about unexpected errors. This one is not unexpected. We are just using exceptions as an implementation detail. 

If a developer needs a more detailed stack trace, they can always throw a random exception...

franzliedke: One more thing: If we decide not to merge this, we are bound to get more reports like this...

tobscure: Alright, you've convinced me :)

tobscure:
@franzliedke is this good to merge then?

franzliedke:
Yep, good to go.

franzliedke:
Thanks, @Albert221!

Issue #679, opened 2015-12-16
Ask for confirmation before "Mark all as Read"


pwFoo:
I'm logged in an able to click and write replies, but get an password prompt if I click at "Mark All as Read" button. I'm also logged in after page reload
Is it a bug?

franzliedke:
Nope, it's the new "sudo mode". It asks you for your password for certain sensitive actions (e.g. visiting the admin panel), but only if you haven't authenticated in the last 30 minutes.

pwFoo:
Ok, is the sudo mode configurable? Because I would disable it for mark read feature ;)
But it's great to secure the admin center ;)

franzliedke:
Let's wait for @tobscure's opinion...

tobscure:
Sudo mode is currently required to edit any user attributes as there are sensitive things like password, email, and groups. However, it's true that it's not ideal to require it for *everything*, like the mark-as-read time. Perhaps we should make [the check](https://github.com/flarum/core/blob/master/src/Api/Controller/UpdateUserController.php#L55) more specific depending on the contents of the `$data` array.

franzliedke:
Okay, makes sense. It does make sense to have at least an alert popping up to confirm that users really want to mark all posts as read. Happened to me on accident once.

tobscure:
Agreed.

dcsjapan:
I agree that an "Are you sure you want to mark all threads as read?" sort of confirmation dialog would make sense, but asking for password confirmation seems a bit overboard. It interferes with the normal flow of forum use.

bogdanteodoru:
Hello guys. I started implementing this.. Is not a hard thing to do but just want to get your opinion on something. 

First, I've added the text to the English locale file (core.yml), but doesn't seem to get updated in my JS code. Any clue why? Do I need to run another command or something? 

Second, do we want a modal popup like.. for example when changing the password? I have implemented a simple alert popup with Cancel/OK buttons which from my point of view does the job.

franzliedke:
@bogdanteodoru Did you recompile the JS code? You can also try removing the JS files in the assets directory. 

I don't think we need a full modal. Using JavaScript's `confirm()` method should be enough.

sijad:
[...]

PR #733, opened 2016-01-11
#679 Ask for confirmation before "Mark all as Read"

bogdanteodoru:


bogdanteodoru:
So this is what I came up for #679. Should I fix what @Nitpick-CI provided as comments?

Issue #349, opened 2015-08-29
"Start a discussion" button is cutted


Petrouchka:
When you show tags on 3 colomns instead of two (on a large screen), the "Start a discussion" button is cut at its right side.
![capture-forum de discussion logiciels libres - mozilla firefox](https://cloud.githubusercontent.com/assets/3955499/9563770/93e2da80-4e8f-11e5-8609-76ee55decd46.png)

Petrouchka:
The size of the button is correct when I uncheck "width 100%" :
![capture-forum de discussion logiciels libres - mozilla firefox](https://cloud.githubusercontent.com/assets/3955499/9567337/30ee27f2-4f27-11e5-8314-d8c08b85f0e2.png)

dcsjapan:
The "Log In to Reply" button seems to be affected as well:

http://discuss.flarum.org/d/672-polish-language-extension/11

maelsoucaze:
For information, the *"Send Password Reset Email"* button is also affected (especially when it's translated *"Envoyer le courriel de réinitialisation du mot de passe""*, I should reduce it!).

PR #734, opened 2016-01-11
Fix primary issue discussed in #349

HLFH:
Fix primary issue discussed in https://github.com/flarum/core/issues/349
IMHO, it should fix the issue related to the cut button "Start a discussion".

### DIFF
--- a/less/lib/sideNav.less
+++ b/less/lib/sideNav.less
@@ -95,7 +95,7 @@
     float: left;
 
     &, > ul {
-      width: 198px;
+      width: 200px;
     }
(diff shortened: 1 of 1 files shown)


tobscure:
Why 198? Let's go a nice round 200 :)

HLFH:
@tobscure I just liked 198px and I didn't want too much lonely pixels, but that's ok. Done in the last commit.

tobscure:
I've implemented this in dfef3c1 with a few additional alterations. Thanks :)

Issue #674, opened 2015-12-12
Remove sudo mode and add password confirmation when changing email address


jordanjay29:
When the new password modal pops up (changing user account settings), it's impossible for someone using a SSO method (Facebook, Twitter, Github, etc) to successfully pass through due to the password being auto-filled by the SSO method and thus unknown to the user.

tobscure:
@franzliedke and I discussed this the other day and came to the conclusion that a user who logged in via SSO on their current session should not be presented with the sudo mode dialog.

jordanjay29:
What about asking the user to create/use an application (Flarum) specific password for this circumstance? Or having a security question for this purpose? Something to give Flarum an extra layer of security in this circumstance. 

I'm not sure it's appropriate for an SSO-authed user to have less security than a normal user/pass combo. For example, I log in to discuss.flarum using Facebook but I have mod permissions, seems like a vulnerability.

tobscure:
Hmm what do you think @franzliedke?

franzliedke:
It's complicated.

For comparison, I only know GitHub that does this. And they don't allow logins from other SSO providers. Do any of you know another place where sudo mode is implemented where we could check for a sensible implementation? After all, we also cannot enforce sudo mode for things like external API access, right? Or can we?

tobscure:
> After all, we also cannot enforce sudo mode for things like external API access, right?

Technically we could (add a sudo_expiry attribute to the access_tokens table or something), but I don't think it's really common practice / necessary

tobscure:
@franzliedke and I discussed this and we have decided to get rid of sudo mode for now, and only require the user to enter their current password in order to change their email address. We came to the conclusion that the only way to implement sudo mode is to require the user to come up with a password, which defeats the purpose of SSO. We talked about protecting admin-only functions, but given the dynamic nature of Flarum's permissions system and the fact that admin functions are spread across both the forum and the admin clients (via the common API), we decided that it's too complicated to pursue. And we can think of no other applications that allow SSO that also have sudo mode, so there is no "best practice" implementation to copy.

We can potentially revisit this at a later time, but for now we're getting rid of it altogether.

PR #735, opened 2016-01-11
#679 Ask for confirmation before "Mark all as Read"

bogdanteodoru:


franzliedke:
Why the change to the `UpdateUserController`?

tobscure:
@franzliedke The change to `UpdateUserController` looks like it skips the sudo mode requirement if it's only the `readTime` attribute being updated. Which is good, except we actually plan to get rid of sudo mode before beta 5 drops due to some unforeseen complications (see #674). @bogdanteodoru, could you please revert that addition? Otherwise it looks good to merge! :+1:

bogdanteodoru:
@tobscure, done. Please remember to merge the language pull request as well.

tobscure:
Beautiful, thanks!

Issue #667, opened 2015-12-06
Locking Thread Then Sticky Thread Shows Locked For Both Badges


pxwee5:
![](http://i.imgur.com/0vfQd3g.png)

To replicate
1. Lock thread
2. Sticky thread
3. Mouseover to confirm. 

```html
<ul class="DiscussionListItem-badges badges">
  <li class="item-sticky"><span class="Badge Badge--sticky " title="Sticky" data-original-title="Locked"><i class="icon fa fa-fw fa-thumb-tack Badge-icon"></i></span>
  </li> 
  <li class="item-locked"><span class="Badge Badge--locked " title="" data-original-title="Locked" aria-describedby="tooltip219726"><i class="icon fa fa-fw fa-lock Badge-icon"></i></span>
  </li> 
</ul>
```
Check out the html code. 

A refresh fixed the issue. Just that its abit annoying to get it in the beginning :P

Discussion Thread:
https://discuss.flarum.org/d/1584-wrong-text-for-sticky-lock-thread-popup/4

bogdanteodoru:
I can do this one as well. Should we keep the sticky or lock icon? Or both?

franzliedke:
The icons should stay the same - this one is just about the "title" attribute, as far as I know.

Thanks!

tobscure:
It's likely related to a quirk of Mithril's redrawing algorithm, in which case this may not be so good for new contributors because a solid understanding of said algorithm may be necessary. I'm happy to take a look if you have no luck.

bogdanteodoru:
I will try to solve it out (might be a good way to learn Mithril's algorithm). If I don't find a solution I will follow up.

PR #739, opened 2016-01-12
#667 Locking Thread Then Sticky Thread Shows Locked For Both Badges

bogdanteodoru:
From what I saw, this forces the re-render which seems to solve the problem.

### DIFF
--- a/js/lib/components/Badge.js
+++ b/js/lib/components/Badge.js
@@ -29,6 +29,8 @@ export default class Badge extends Component {
     // of badges.
     attrs.key = attrs.type;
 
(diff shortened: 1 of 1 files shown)


tobscure:
Not quite, but thanks for having a look! It did nudge me to look into the issue so you can consider this a successful PR :)

https://github.com/flarum/core/commit/e5a7013c2caaf25864cb0bd6640696c6ef9cf19f

bogdanteodoru:
@tobscure yeah, more or less. I am a bit disappointed in myself for not fixing this properly. I'll move to another issue then.

PR #748, opened 2016-01-15
Added support for new minifiers

JoshyPHP:
https://discuss.flarum.org/d/1870-new-javascript-minifiers

New minifiers are available in s9e\TextFormatter 0.4.8. Two of them work with a web service that can be faster than the Closure Compiler service hosted by Google. The [web service's repository is here](https://github.com/s9e/WebService-Minifier). The default configuration is tuned for the library, Flarum needs different settings. Below is the code for the actual endpoint configured in Flarum. Going forward you may want to host the web service on your own server. The current instance acts as a caching proxy for the Closure Compiler service and only runs the Closure Compiler application locally if Google's service is unavailable.

```php
<?php

// That's the path to the WebService-Minifier repository
$root = $_ENV['OPENSHIFT_REPO_DIR'] . '/libs/webservice-minifier';
include $root . '/vendor/autoload.php';

$minifier = new s9e\WebServices\Minifier\Minifier;
// Path to the cache dir
$minifier->cacheDir = $_ENV['OPENSHIFT_REPO_DIR'] . '/misc/flarum-minifier-cache/';
// Max ~2MB of code, which is way overkill and should probably set lower
$minifier->maxPayload = 2000000;
$minifier->minifiers = [
	// Uses CCS first because it's probably faster than running the Java application locally
	'ClosureCompilerService' => [
		'compilationLevel' => 'SIMPLE_OPTIMIZATIONS',
		'timeout'          => 60
	],
	// Uses the Java application as a fallback
	'ClosureCompilerApplication' => [
		'closureCompilerBin' => $root . '/bin/compiler.jar'
	]
];
$minifier->handleRequest();
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -39,7 +39,7 @@
         "tobscure/json-api": "^0.2.0",
         "oyejorge/less.php": "~1.5",
         "intervention/image": "^2.3.0",
-        "s9e/text-formatter": "^0.4.4",
+        "s9e/text-formatter": "^0.4.8",
         "psr/http-message": "^1.0",
         "zendframework/zend-diactoros": "^1.1",
         "zendframework/zend-stratigility": "^1.1",
--- a/src/Asset/JsCompiler.php
+++ b/src/Asset/JsCompiler.php
@@ -13,7 +13,8 @@ namespace Flarum\Asset;
 use Exception;
 use Illuminate\Cache\Repository;
 use MatthiasMullie\Minify;
-use s9e\TextFormatter\Configurator\JavaScript\Minifiers\ClosureCompilerService;
+use s9e\TextFormatter\Configurator;
+use s9e\TextFormatter\Configurator\JavaScript\Minifiers\FirstAvailable;
 
 class JsCompiler extends RevisionCompiler
 {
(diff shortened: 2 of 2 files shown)


franzliedke:
While this is certainly cool, I'm not sure we should depend on an external service for this feature.

JoshyPHP:
You already do, though. You depend on Google's service.

With this, you can at least host your own minifier service independent of Google's while keeping Google's as a fallback.

franzliedke:
Oh, interesting, I didn't know that. ;)

JoshyPHP:
Talking about Google's Closure Compiler service, I got reminded today that they enforce some kind of quota, as [mentionned in their documentation](https://developers.google.com/closure/compiler/docs/gettingstarted_ui?hl=en):

> Note: To prevent abuse, the Closure Compiler limits the number of consecutive compiles that you can run. If you see the message Too many compiles performed recently. Try again later, it means you've temporarily exceeded the limit.

I don't know what the exact limit is but it's high enough that I have rarely hit it in the past. For some reason I did exceed that limit today while running tests on Travis. They also enfore a size limit on the unminified code. I don't think it's documented anywhere but I estimate it's around 1 MB.

tobscure:
Thanks @JoshyPHP!

PR #779, opened 2016-02-09
Correct redirect uri in OAuth2 Controller

sijad:


franzliedke:
As detailed in the issue, the `getOriginalRequest()` method is not part of the `ServerRequestInterface` interface.

sijad:
@franzliedke it's just an interface, it implemented [here](https://github.com/zendframework/zend-stratigility/blob/master/src/Http/Request.php#L71)

franzliedke:
No, it's not "just" an interface. We typehint it so that Flarum can be used with other PSR-7 compatible middleware layers.

sijad:
@franzliedke sorry about that, how about now?

franzliedke:
Looking good. I'll probably create a helper for this.

sijad:
thanks for reviewing my code, flarum/flarum-ext-auth-twitter#3 need to be updated too, please let me know what should I do

franzliedke:
Please do the same there.

PR #786, opened 2016-02-10
reordering and removing unused imports

Luceos:
I've noticed quite some unused imports in some files. In order to keep the files clean and tidy I ran a [optimize imports](https://www.jetbrains.com/phpstorm/help/optimizing-imports.html) command.

It does two things:

- remove unused imports
- sorts based on name


franzliedke:
Ah, nice. I didn't know it had a command for doing it for all files. :)

Luceos:
Yes, quite useful :P had to look it up :)

tobscure:
I think StyleCI does this too, but I *still* haven't gotten around to merging in those changes. Soon!

PR #808, opened 2016-02-15
fixed namespace guessing for extension

Luceos:
Fixes namespace guessing in migration files by loading composer autoloading definition. In cases where migration files are used, this will be set. This is a good way of providing extension migration support and is forward an backward compatible. Also works with lowercase namespacing.


Luceos:
I'm considering removing the if/else altogether. There is no reason to support the Flarum\Core namespace this way if it can be intelligently identified through composer.

PR #810, opened 2016-02-16
extensions improvement (mainly related to migrations)

Luceos:
- Fixed some issues with migrations in my last PR for rewriting the whole extension thing.
- Identifying namespace of extensions for migrations has been modified and tested to allow better running of migrations. Including a fallback on composer name in case no autoload has been defined in the composer file.

### DIFF
--- a/src/Extension/Extension.php
+++ b/src/Extension/Extension.php
@@ -223,6 +223,14 @@ class Extension implements Arrayable
     }
 
     /**
+     * @return string
+     */
+    public function getPath()
+    {
+        return $this->path;
+    }
+
+    /**
      * Tests whether the extension has assets.
      *
      * @return bool
@@ -240,6 +248,23 @@ class Extension implements Arrayable
     public function hasMigrations()
     {
         return realpath($this->path . '/migrations/') !== false;
+    }
+
+    /**
+     * Identifies Extension namespace based on autoload from composer.
+     *
+     * @return string|null
+     */
+    public function getNamespace()
+    {
+
+        // Try identifying based on psr-4/2 whatever comes first in autoload.
+        if(is_array($this->autoload) && count($this->autoload)) {
+            return head(array_keys(head($this->autoload)));
+        }
+
+        // Identify based on package name; which is less safe.
+        return Str::studly(str_replace('/', '\\', $this->name));
     }
 
     /**
--- a/src/Api/Controller/UpdateExtensionController.php
+++ b/src/Api/Controller/UpdateExtensionController.php
@@ -42,12 +42,10 @@ class UpdateExtensionController implements ControllerInterface
         $enabled = array_get($request->getParsedBody(), 'enabled');
         $name = array_get($request->getQueryParams(), 'name');
 
-        $extension = $this->extensions->getExtension($name);
-
         if ($enabled === true) {
-            $this->extensions->enable($extension);
+            $this->extensions->enable($name);
         } elseif ($enabled === false) {
-            $this->extensions->disable($extension);
+            $this->extensions->disable($name);
         }
     }
 }
--- a/src/Database/Migrator.php
+++ b/src/Database/Migrator.php
@@ -247,14 +247,17 @@ class Migrator
     {
         $file = implode('_', array_slice(explode('_', $file), 4));
 
-        // flagrow/image-upload
         if ($extension) {
-            $class = str_replace('/', '\\', $extension->name);
+            $class = $extension->getNamespace();
         } else {
-            $class = 'Flarum\\Core';
+            $class = 'Flarum\\Core\\';
         }
 
-        $class .= '\\Migration\\';
+        if (substr($class, -1) != '\\') {
+            $class .= '\\';
+        }
+
+        $class .= 'Migration\\';
 
         $class .= Str::studly($file);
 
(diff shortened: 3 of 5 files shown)


src/Extension/Extension.php
```diff
@@ -243,6 +251,23 @@ public function hasMigrations()
...
+    public function getNamespace()
+    {
+
+        // Try identifying based on psr-4/2 whatever comes first in autoload.
+        if(is_array($this->autoload) && count($this->autoload)) {
+            return head(array_keys(head($this->autoload)));
+        }
+
+        // Identify based on package name; which is less safe.
+        return Str::studly(str_replace('/', '\\', $this->name));
```
Luceos: not even sure we need to maintain such a fallback

src/Extension/Extension.php
```diff
@@ -243,6 +251,23 @@ public function hasMigrations()
...
     /**
+     * Identifies Extension namespace based on autoload from composer.
+     *
+     * @return string|null
+     */
+    public function getNamespace()
+    {
+
+        // Try identifying based on psr-4/2 whatever comes first in autoload.
+        if(is_array($this->autoload) && count($this->autoload)) {
```
franzliedke: Please also remove the empty line above

franzliedke:
We really need to discuss the approach to resolving the migration class names. There's a chance that by adding a new autoloading rule the developer will inadvertently break existing migrations. We don't want that... ;)

Luceos:
I'm aware of that. I do think this is the best solution for now though. Because all extensions - at this time - that use migrations do have an autoload entry.

A future solution would be to scan the files in the migration directory to analyze it for namespacing. But it's dirty too.

tobscure:
*sigh* if only we could use PHP 7's anonymous classes:

```php
<?php

return class {
    public function up()
    {
        // ...
    }

    public function down()
    {
        // ...
    }
}
```

In all seriousness though, I think the best solution would be to search for a `composer.json > extra > flarum-extension > migrations` value, as in:

```json
{
    "name": "flarum/flarum-ext-auth-twitter",
    // etc
    "extra": {
        "flarum-extension": {
            "title": "Twitter Login",
            "icon": {...},
            "migrations": "Flarum\Auth\Twitter\Migration"
        }
    }
}
```

IF that is not found, then default to Vendor\Package\Migration, stripping hyphens from Package:

```
flarum/flarum-ext-tags -> Flarum\Tags\Migration
flarum/flarum-ext-auth-twitter -> Flarum\AuthTwitter\Migration
```

How does that sound?

@Luceos do you reckon you could amend the PR with this approach, and send the appropriate PR to flarum-ext-auth-twitter to add its migration namespace to composer.json? All of the other extensions can default to the Vendor\Package\Migration value.

franzliedke:
@tobscure Yeah, I was thinking about anonymous classes, too. What about returning an array with two closures instead?

tobscure:
@franzliedke classes are nicer, because we can inject dependencies (as in AbstractMigration) and they're better for more complex migrations (private methods, etc.)

franzliedke:
That's true. Although injection would work, too, with Laravel's `Container::call()`.

tobscure:
@franzliedke that's also true. So you're thinking something like:

```php
<?php

use Illuminate\Database\Schema\Builder;

return [
    'up' => function (Builder $schema) {
        $schema->create('whatever', ...);
    },
        $schema->create('whatever', ...);
    }
];
```

I think I like it...

franzliedke:
Yep. Once we go full PHP7, we can use anonymous classes and keep this as a fallback solution (backwards compatibility).

Luceos:
I'll take a look this week. It should be quite easy to include the extra attributes for migration loading. Squashing is not that big an issue anymore ;)

Luceos:
Any advice on squashing when having pushed commits already. Seems I can't find my most recent commit in the git rebase -i list..

tobscure:
@Luceos Scratch my extra-attribute-in-composer.json idea – we'll go with @franzliedke's array-returning method. It will break BC but we're beta so that's OK. We'll need PRs changing the format for all of the migrations in core/extensions (and updating their minimum flarum/core version requirement too).

Luceos:
I must have missed the intent of that discussion al together. I'll revamp the PR, still any assistence on the squashing or would a complete new PR make more sense @flarum/core ?

franzliedke:
I've started doing this on core now. I hope you don't mind, @Luceos

franzliedke:
@tobscure What minimum version should the extensions require, though? We haven't tagged beta5 yet...

Luceos:
@franzliedke no I don't ;) If you would be able to assist with squashing this PR, that would be great too. Want to have this one done too in fact.

Luceos:
Wait what am I asking, the last commit has to be undone and we need to add support for array migrations.

franzliedke:
@Luceos Can you send a new one based on the latest code? Should be easier. The changes to the Migrator class and regarding the namespaces of extension migrations are not needed anymore. The rest looks good, though.

Luceos:
wasn't sure whether that was the easiest way, but sure; let me do so ;)

franzliedke:
@tobscure I completely refactored the migrations in core and all extensions to the new array structure. After that, I also wrote some nice factory methods to ease the writing of some of these more standard ones. Please review.

Also, Toby, can you tell me how the constraints of the affected extensions should be updated exactly (or do it yourself), please?

Issue #698, opened 2015-12-31
Firefox autocomplete with more than 1 account not working


Jump3r99:
I saved passwords for 2 accounts in FF. When I'm trying to select one in logging window FF autocomplete only password - I have to select once more same nick.

Luceos:
This sounds like a browser issue, can you explain a bit more how this is a Flarum issue?

Jump3r99:
I think flarum wrong switch between login data. Password seem to be correct.

dcsjapan:
Confirmed with Firefox 43.0 for OS X. I created a test account on discuss.flarum.org to test this. When I type a "D" in the Username field, Firefox presents a dropdown list allowing me to select "Dominion" as the user name. It then fills in the password for that account, but *does not* fill in the username. I have to type "D" and select "Dominion" from the dropdown list a second time to get the Username field filled.

In other words, it's just as @Jump3r99 described.

I was tempted to think it's a browser issue, but I tested the same thing with two accounts on our esoTalk forums. There, the Firefox autofill function enters *both* the username *and* password* correctly.

It seems possible that Flarum is doing something that's preventing autofilling of the Username field.

franzliedke:
This may be related to #467.

tobscure:
Fixed in #811

PR #811, opened 2016-02-16
Fix login box autocomplete in firefox

### LINKED ISSUE #467: Unable to log in via KeePassX autotype
There is something funky in the Flarum login form that stops it accepting auto-type from KeePassX.

Browser: Mozilla/5.0 (X11; Linux x86_64; rv:31.0) Gecko/20100101 Firefox/31.0 Iceweasel/31.2.0
OS: Debian Jessie
KeePassX: 0.4.3

To reproduce:

Create a KeePassX site entry for Flarum, with a valid Flarum username and password, and the URL http://discuss.flarum.org/

In KeePassX, double-click the URL of the Flarum entry and wait for the browser to open the Flarum main page.

Click Login. Verify that the insertion point is correctly positioned in an empty Username or Email box.

Bring the KeePassX window to the front again, and press Ctrl-V to make KeePassX auto-type username, Tab, password, Enter into the browser window.

Expected result: Login form appears to be filled in and Flarum logs on.

Actual result: White flicker at bottom of browser window, nothing appears in logon form text boxes, logon doesn't happen.

I have credentials for several hundred sites in my KeePassX database. Flarum is the only one that displays this behavior.

No repro on Chromium.

sijad:
use [oninput](http://caniuse.com/#feat=input-event) event fixes ~~#467~~ and #698

### DIFF
--- a/js/forum/dist/app.js
+++ b/js/forum/dist/app.js
@@ -24095,7 +24095,7 @@ System.register('flarum/components/LogInModal', ['flarum/components/Modal', 'fla
                   { className: 'Form-group' },
                   m('input', { className: 'FormControl', name: 'email', type: 'text', placeholder: extractText(app.translator.trans('core.forum.log_in.username_or_email_placeholder')),
                     value: this.email(),
-                    onchange: m.withAttr('value', this.email),
+                    oninput: m.withAttr('value', this.email),
                     disabled: this.loading })
                 ),
                 m(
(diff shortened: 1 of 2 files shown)


sijad:
as I don't have access to debian os I can't be sure if #698 has been fixed...

franzliedke:
Can you explain why this fixes the issue?

bmalex88:
With `autocomplete="off"` for the text inputs this problem is solved, I think. Using `oninput` event is also an improvement, as it only fires when text is changed

sijad:
@franzliedke in FF when you select and click on you email from autocomplete form, FF only fire `change` event for `password` input and not for 'email'
when this happens `this.email` stay empty and as Mithril do redraw the `Modal` it only fill `password` input

franzliedke:
Okay, I guess that's fine then. :+1: from me.

tobscure:
The issue with using `oninput` is that if redraws are slow (perhaps due to lots of things being on the page, behind the modal), then typing becomes slow.

What we should probably do is call `m.redraw.strategy('none')` when setting the field values in the `oninput` handler – this way it will update the component state, but won't trigger a redraw (which is unnecessary anyway).

tobscure:
...However, until this causes visible problems with performance, I think it's nicer to use `bidi` as I have done in 71d2e71908de6051b8ae7f29bae4c4e3e14e88f6 :P

sijad:
@tobscure sorry about that, I didn't know about `bidi`

PR #813, opened 2016-02-17
typehint fix, opening for implementation

Luceos:
In order to use Laravel with Flarum, we need to open up the service providers to get the Laravel application injected. Currently only the `Flarum\Foundation\Application` is allowed, which implements the `Illuminate\Contracts\Foundation\Application`, in order to open up, the last should be imported in the AbstractServiceProvider.

I think the import was simply forgotten initially.

### DIFF
--- a/src/Foundation/AbstractServiceProvider.php
+++ b/src/Foundation/AbstractServiceProvider.php
@@ -11,6 +11,7 @@
 namespace Flarum\Foundation;
 
 use Illuminate\Support\ServiceProvider;
+use Illuminate\Contracts\Foundation\Application;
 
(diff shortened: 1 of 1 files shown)


franzliedke:
Very good, thanks!

tobscure:
The problem is that I believe the Flarum Application provides a few additional methods (`config()`, `isInstalled()`, `isUpToDate()`, `url()`?) which some of our service providers need to make use of. 

Perhaps it'd be better for us to make Flarum\Foundation\Application a decorator class (wrapping around an illuminate application contract) instead?

Luceos:
@tobscure exactly the next issue I ran into today ;)

PR #817, opened 2016-02-18
Revert "typehint fix, opening for implementation"

franzliedke:
Reverts flarum/core#813

/cc @Luceos 
As per Toby's comments in #813, I will revert this. We should then create a wrapper class working on an Illuminate application instance for implementing the extra functionality that our Application class provides.


Luceos:
I agree with that. If I have the time I'll do an initial attempt if you'd like.

PR #830, opened 2016-02-25
refactor of migrations and extensions fix

Luceos:
This PR now supports array closure migrations. Class migrations are completely DEFUNCT when merged! Extensions need to update.

A few additional fixes are added that broke core related to migrations in the new Extension (class) structure.


src/Database/Migrator.php
```diff
@@ -228,15 +240,19 @@ public function getMigrationFiles($path)
...
      * @param  string $path
-     * @return object
+     * @param  string $file
+     * @return array
      */
-    public function resolve($file, $path)
+    public function resolve($path, $file)
     {
-        return $this->files->getRequire("$path/$file.php");
+        $migration = $path . '/' . $file . '.php';
```
franzliedke: String interpolation FTW. :P

Luceos: You're suggesting these variables can hold anything else in my case? That's not true..

franzliedke: Nope. Interpolation, not injection. I was suggesting to change it back to what I had. ;)

~~~php
$migration = "$path/$file.php";
~~~

Luceos: I can't see why interpolation would be useful here, but enfin :grin: I'll add it in.

Luceos:
Please note squashing will be close to impossible due to the merge with master.

franzliedke:
I'll probably squash-merge it again. :P

Is this ready from your perspective?

Luceos:
It is now, squashed the last two.

franzliedke:
Oh, can you go through the long comments in the Migrator class again? I think some of them may have outdated content...

Luceos:
Yes I noticed that too. Some of those are actually clearly copied from Laravel and have no meaning on the methods. Let me clean that up. One sec.

Luceos:
There you are.

franzliedke:
Okay, I [squash-merged it in](https://github.com/flarum/core/commit/e4412178b1ac5b68be4c44a6c5dd0bda4bb27105). Thanks!

We need to talk about our preferred order of `$file` and `$path`, by the way. ;)

Luceos:
@franzliedke that is a discussion worth my time, when are you able to meet in person? :grin: I'd love to debate the preference of all method arguments in flarum/core when we have the chance.

Btw it's hillarious how close our code matched.

Issue #774, opened 2016-02-06
Eliminate `extensions` directory from skeleton


tobscure:
@franzliedke and I discussed the usefulness of the `extensions` directory in the skeleton today. There are a couple of reasons we have this directory in the first place:

* **So we can scan its sub-directories to build a list of available extensions.** However, we realised we can just read the list from `vendor/composer/installed.json`, filtering by package type. We do this anyway to get extension version numbers!

* **To support drag/drop uploading of extensions.** It only works for extensions that don't have any Composer autoloading or dependency requirements though (really only language packs). We decided we want to discourage this because it is very likely to cause confusion by having multiple extension installation methods.

Since both of these reasons are invalid, we think there is no longer any reason to have a separate `extensions` directory. Instead, extensions can just be installed into the `vendor` directory like any other package. It's cleaner and eliminates a potential source of confusion.

Luceos:
This has some issues:

- Composer has to run twice, first to drop the flarum/composer-installer then to move extensions to vendor. It's beta, let's live with that?
- All migrations by extensions had the flarum-(ext-) removed in the namespace. I'm not sure how to handle this. Replacing flarum-ext- in the composer name to resolve to the migration namespace might create issues, because it's still namespace guessing. What's the reason for namespaced migrations that aren't necessarily required to be added to the classmap anyway :stuck_out_tongue:

franzliedke:
Migrations are now not namespaced anymore. *yay*

Luceos:
Jep, that should make this a much easier task.

Luceos:
Any objections to move this to beta 5? I had a wip branch somewhere I need to finish now with the latest changes to the migration namespaces ;)

franzliedke:
Agreed. Since we are already breaking BC with the migrations stuff, we may as well do this change in beta.5 as well and make the upgrade as painful as possible. But only once. ;)

*Wait a sec. @Luceos, you did propose to move this to beta.5, right? I got an email...*

Luceos:
I know I deleted it because I couldn't find my branch. But let's move this to beta 5. The reason is the same as you say, there is already a backward compatibility thing here.

Sorry to confuse you.

Luceos:
So there's this PR, it also adds prefer-stable.

Luceos:
@franzliedke a test run would be great for this one.

tobscure:
[...]

PR #841, opened 2016-03-01
Refactoring to drop extensions dir, see #774

Luceos:
- See #774 & https://github.com/flarum/flarum/pull/35
- No longer loading the composer.json in the extension directory, but fully utilizing the installed.json of composer.

### tested:

- with dev-master of all core extensions and dev-master of flarum/flarum dropping flarum/composer-installer
- loading flarum
- loading admin
- en-/disabling extensions
- re-running migrations

### DIFF
--- a/src/Extension/Extension.php
+++ b/src/Extension/Extension.php
@@ -96,8 +96,9 @@ class Extension implements Arrayable
      */
     protected function assignId()
     {
-        $segments = explode('/', $this->path);
-        $this->id = end($segments);
+        list($vendor, $package) = explode('/', $this->name);
+        $package = str_replace(['flarum-ext-', 'flarum-'], '', $package);
+        $this->id = "$vendor-$package";
     }
 
     /**
--- a/src/Extension/ExtensionManager.php
+++ b/src/Extension/ExtensionManager.php
@@ -18,6 +18,7 @@ use Flarum\Foundation\Application;
 use Flarum\Settings\SettingsRepositoryInterface;
 use Illuminate\Contracts\Events\Dispatcher;
 use Illuminate\Filesystem\Filesystem;
+use Illuminate\Support\Arr;
 use Illuminate\Support\Collection;
 
 class ExtensionManager
@@ -38,6 +39,11 @@ class ExtensionManager
      */
     protected $filesystem;
 
+    /**
+     * @var Collection|null
+     */
+    protected $extensions;
+
     public function __construct(
         SettingsRepositoryInterface $config,
         Application $app,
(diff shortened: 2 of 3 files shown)


Luceos:
Is styleci complaining about a non-existent tab? :laughing:

tobscure:
Great work! Will test as soon as I get time. 

StyleCI is complaining about extra unnecessary whitespace on that line.

src/Extension/Extension.php
```diff
@@ -96,8 +96,9 @@ public function __construct($path, $composerJson)
      */
     protected function assignId()
     {
-        $segments = explode('/', $this->path);
-        $this->id = end($segments);
+        list($vendor, $package) = explode('/', $this->name);
+        $package = str_replace(['flarum-ext-', 'flarum-'], null, $package);
```
franzliedke: Hmm, minor thing: It surprised me that the second parameter to `str_replace` can be `null`. Can you change it to an empty string? That's really what we want to express here, right?

src/Extension/ExtensionManager.php
```diff
@@ -260,13 +282,24 @@ protected function setEnabled(array $enabled)
...
 
+    /**
+     * The extensions path.
+     *
+     * @return string
+     */
     protected function getExtensionsDir()
     {
-        return public_path('extensions');
+        return public_path('vendor');
```
franzliedke: Can we pass that in when defining the manager in the service provider?

src/Extension/ExtensionServiceProvider.php
```diff
@@ -21,11 +21,10 @@ public function register()
     {
         $this->app->bind('flarum.extensions', 'Flarum\Extension\ExtensionManager');
 
-        $config = $this->app->make('flarum.settings')->get('extensions_enabled');
-        $extensions = json_decode($config, true);
+        $extensions = $this->app->make('flarum.extensions')->getEnabledExtensions();
 
         foreach ($extensions as $extension) {
-            if (file_exists($file = public_path().'/extensions/'.$extension.'/bootstrap.php')) {
+            if (file_exists($file = $extension->getPath().'/bootstrap.php')) {
```
franzliedke: Can we shorten this to a method in the extension manager that returns an array of files to be included?

franzliedke:
Looks good, overall! :+1: 
Just some minor corrections...

franzliedke:
Oh, last thing: do we still get any benefit from defining a special type for extension composer packages? Or was that only needed for installing it to a different directory?

tobscure:
@franzliedke we need the flarum-extension type to be able to tell which packages in installed.json are Flarum extensions :)

src/Extension/ExtensionManager.php
Luceos: While loading the installed.json we also use the hardcoded public path to vendor. IMO that's wrong too. So I just kept it like this. We might need to provide a way to get the vendor_path and extension_path seperately from the ExtensionManager, perhaps in the Application class. Same behavior as Laravel in that sense.

If you don't mind, won't be taking this along for this PR.

Luceos:
@flarum/core 

> Just some minor corrections...

All done except for the one to pass the vendor folder into the construct of the Extension Manager, I think there should be an alternative solution for that. Either create a `extension_path` and `vendor_path` helper or implement those paths in the Application.

tobscure:
Fix up the StyleCI error and I'm happy :) @franzliedke we can work out a solution for the hardcoded vendor path after

Luceos:
@tobscure was looking at those.. I really need to set up phpstorm to use the styleci/nitpick params.

Luceos:
@tobscure done.

tobscure:
Thanks so much @Luceos, great work :)

Luceos:
Happy with this too :)

Don't forget about https://github.com/flarum/flarum/pull/35

PR #844, opened 2016-03-03
added integration with codecov to track coverage of tests

Luceos:
Would like to propose tracking code coverage. We can do this automatically by using codecov.io (just use your account to set it up @flarum/core ). The entry below sends the stats over to codecov.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,4 +30,7 @@ notifications:
     on_failure: always
     on_start: false
 
+after_success:
+  - bash <(curl -s https://codecov.io/bash)
+
 sudo: false


franzliedke:
Not sure we want that at this point. Can be slightly depressing... also I'm in the camp of those who doubt the usefulness of code coverage as metric. It can lead to the wrong goals in testing.

Luceos:
@franzliedke we don't need to publish this and let's face the fact at some point we need to know how far the tests are covering code. Maybe you (core devs) don't need it, but those that write the code do ;) It makes their lives much easier.

franzliedke:
That's convincing. :+1:

Luceos:
I meant "those that write the tests do", updated comment ;)

Ps it's also a goal to work towards.

franzliedke:
@tobscure Do you agree?

PR #846, opened 2016-03-04
Remove 'extensions' path for writable check

sijad:


Luceos:
Thanks @sijad totally missed that! :+1:

PR #855, opened 2016-03-11
Add Icon Dropdown Component

dav-is:
The `IconDropdown` component is the same as a `Dropdown`, except the it houses font awesome icons for selection. This adds a cleaner way for users to select a font awesome icon. I noticed in groups
# Usage:

`selection` the icon that is currently selected. Default value of `question`
`icons` the set of icons to display for selection. Possible values: "all", "social", or an array of icon names. Default value of `all`.

```
IconDropdown.component({selection: this.selection, icons: 'social'})
```
# Screenshot:

screenshot

### DIFF
--- a/js/lib/components/IconDropdown.js
+++ b/js/lib/components/IconDropdown.js
@@ -1,0 +1,65 @@
+import Dropdown from 'flarum/components/Dropdown';
+import ItemList from 'flarum/utils/ItemList';
+import icon from 'flarum/helpers/icon';
+
+/**
+ * The `IconDropdown` component is the same as a `Dropdown`, except the it
+ * houses font awesome icons for selection.
+ *
+ * ### Props
+ *
+ * - `selection` - The icon currently selected, required as m.prop
+ * - `icons` - The set of icons to display as an option, set as 'all', 'social', or custom array of fa-icons
+ */
+export default class IconDropdown extends Dropdown {
+  static initProps(props) {
+    super.initProps(props);
+    
+    props.className = 'Dropdown--iconselector';
+    props.buttonClassName = 'Button Button--iconselector';
+    props.menuClassName = 'Dropdown-menu--below';
+  }
+  
+  view() {
+    this.props.children = this.items().toArray();
+
+    return super.view();
+  }
+
+  getButtonContent() {
+    return [
+        icon(this.props.selection()),
+        this.props.caretIcon ? icon(this.props.caretIcon, {className: 'Button-caret'}) : ''
+    ];
+  }
+  
+  items() {
+    //Set Defaults
+    if (typeof this.props.selection == 'undefined') {this.props.selection = m.prop('question');}
+    if (typeof this.props.icons == 'undefined') {this.props.icons = 'all';}
+
+    const items = new ItemList();
+    //Preset icon list
+    const iconlist = {
+        'all':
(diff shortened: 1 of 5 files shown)


Luceos:
Out of curiosity what is the effect on loading times with and without the dropdown? Did you check that?

Nice addition.

jordanjay29:
Love this, let's get it in. I have an extension I'm working on that could soon make use of it, and having it as a Flarum component would be perfect.

Luceos:
@flarum/core wouldn't it make more sense to start classifying the components too (much as in the backend side) so abstracts should move to a separate directory and extended components to another? This might help in the future.

franzliedke:
Don't we have a lib folder for that sort of thing?

Luceos:
Well the lib is where all shared code is placed, so those used by forum and admin. If I'm not mistaken @tobscure ?

tobscure:
Nice! Will get this in with some modification when I have time.

I don't see the need for component sub-folders, I think it would be somewhat arbitrary.

js/lib/components/IconDropdown.js
```diff
@@ -0,0 +1,71 @@
...
+        'social':
+            ["globe", 'amazon', 'angellist', 'apple', 'behance', 'bitbucket', 'codepen', 'connectdevelop', 'dashcube', 'delicious', 'deviantart', 'digg', 'dribbble', 'dropbox', 'drupal', 'facebook', 'flickr', 'foursquare', 'get-pocket', 'git', 'github', 'github-alt', 'gittip', 'google', 'google-plus', 'google-wallet', 'gratipay', 'hacker-news', 'instagram', 'ioxhost', 'joomla', 'jsfiddle', 'lastfm', 'leanpub', 'linkedin', 'meanpath', 'medium', 'odnoklassniki', 'opencart', 'pagelines', 'paypal', 'pied-piper-alt', 'pinterest-p', 'qq', 'reddit', 'renren', 'sellsy', 'share-alt', 'shirtsinbulk', 'simplybuilt', 'skyatlas', 'skype', 'slack', 'slideshare', 'soundcloud', 'spotify', 'stack-exchange', 'stack-overflow', 'steam', 'stumbleupon', 'tencent-weibo', 'trello', 'tripadvisor', 'tumblr', 'twitch', 'twitter', 'viacoin', 'vimeo', 'vine', 'vk', 'wechat', 'weibo', 'weixin', 'whatsapp', 'wordpress', 'xing', 'y-combinator', 'yelp', 'youtube-play' ],
```
franzliedke: Please extract both lists to variables (put them at the beginning of the file, they're only scoped to this module anyway if you don't export them).
I'd prefer having one item per line - even if that gets very long, it's easier to see what's going on and diffs will be much easier to read.

Also, you don't need to wrap the keys in strings. :)

less/lib/Dropdown.less
```diff
@@ -79,6 +79,49 @@
...
+    -webkit-background-clip: padding-box;
+    background-clip: padding-box;
+    border: 1px solid rgba(0,0,0,.15);
+    -webkit-transform: none !important;
+    transform: none !important;
+    -webkit-transition: none !important;
+    -moz-transition: none !important;
+    -o-transition: none !important;
+    transition: none !important;
+    > li > div {
```
franzliedke: Are these very specific selectors necessary?

Also, what's our policy on browser prefixes, @tobscure ?

dav-is: They are necessary. They overwrite the original Dropdown CSS.

dav-is:
I'm vacationing in China currently and don't have time to work on this. It would be great if you all could fix it up.

less/lib/Dropdown.less
franzliedke: That's ... too bad. :(

jordanjay29: Why is it necessary to reuse the same class?

less/lib/Dropdown.less
```diff
@@ -79,6 +79,49 @@
...
 }
+.Dropdown-menu--below {
+    padding-left: 10px !important;
+    padding-top: 10px !important;
+    bottom: inherit !important;
+    position: absolute !important;
+    top: 100% !important;
+    overflow: auto;
+    max-height: 45vh !important;
+    min-width: 189px !important;
```
franzliedke: I must say that all these `!important` are a bad sign, too. It means we have too specific selectors somewhere else.

tobscure: Yep, one of the things I will modify.

less/lib/Dropdown.less
tobscure: @franzliedke we have mixins which automatically apply browser prefixes. See less/lib/mixins/vendor-prefixes.less

dav-is: @jordanjay29 the dropdown component adds the .Dropdown-menu class no matter what

ardacebi:
@dav-is 

- How many font awesome icons does this picker has? 
- Does it have all or just the common ones? 
- How should users input a font awesome icon that it isn’t on the picker if it only has the common icons?
- Does this use FA 4 or FA 5?

Many thanks.

PR #856, opened 2016-03-12
WIP: PostgreSQL support

vjousse:
Early development on PostgreSQL integration, related to flarum/issue-archive#411

As I'm new to Flarum and Laravel, this PR is here to gather early feedback on the psql integration. Feel free to comment/suggest.

### DIFF
--- a/src/Install/InstallServiceProvider.php
+++ b/src/Install/InstallServiceProvider.php
@@ -28,7 +28,6 @@ class InstallServiceProvider extends AbstractServiceProvider
      */
     public function register()
     {
-
         $drivers = new Drivers();
 
         $this->app->bind(
@@ -57,7 +56,6 @@ class InstallServiceProvider extends AbstractServiceProvider
         $this->app->singleton('flarum.install.routes', function () {
             return $this->getRoutes();
         });
-
 
         $this->app->singleton(
             'Flarum\Install\Database\Drivers',
@@ -91,5 +89,4 @@ class InstallServiceProvider extends AbstractServiceProvider
 
         return $routes;
     }
-
 }
--- a/src/Install/InstallServiceProvider.php
+++ b/src/Install/InstallServiceProvider.php
@@ -39,7 +39,7 @@ class InstallServiceProvider extends AbstractServiceProvider
                         'json',
                         'mbstring',
                         'openssl',
-                        'pdo_mysql',
+                        'pdo_mysql|pdo_pgsql',
                     ]),
                     new WritablePaths([
                         public_path(),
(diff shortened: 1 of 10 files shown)


Luceos:
@flarum/core fyi I personally asked @vjousse to do this wip pr to allow for early feedback.

Things to consider:

- Laravel native drivers.
- Raw DB queries in use in migrations and other code need to re-evaluated to support multiple drivers.

vjousse:
@Luceos note sure what you mean by "Laravel native drivers"?

Luceos:
I meant these: https://laravel.com/docs/5.1/database#introduction

src/Install/InstallServiceProvider.php
```diff
@@ -54,6 +57,12 @@ function () {
             return $this->getRoutes();
         });
 
+        $this->app->singleton(
+            'Flarum\Install\Database\Drivers',
+            function () use ($drivers) {
+                return $drivers;
+        });
```
franzliedke: This can be simplified with `$this->app->instance('Flarum\Install\Database\Drivers', $drivers)`, by the way. :)

src/Install/InstallServiceProvider.php
```diff
@@ -39,7 +42,7 @@ function () {
                         'json',
                         'mbstring',
                         'openssl',
-                        'pdo_mysql',
+                        implode('|', $drivers->getSupportedPhpDrivers()),
```
franzliedke: You should probably get the drivers instance from the container instead. Otherwise, there's no point in binding it at all, if it cannot be replaced...

vjousse: Well I agree, but I have to admit that I don't see how to do it (I'm pretty new to laravel).

As far as I understand, InstallServiceProvider is used to define services. The problem is that the 'Flarum\Install\Prerequisite\PrerequisiteInterface' needs another service the 'Flarum\Install\Database\Drivers' one.
If I type hint the Driver class in the 'PhpExtensions' constructor, it's not resolved automagically.

What's the best way to do this without refactoring Composite and/or PhpExtensions too heavily?

Luceos:
Closing this due to inactivity.

skywalker512:
I use this to install Flarum with postgresql everthing is right, but when I started a post the 500 error came to

```

{
  "errors": [
    {
      "code": 500,
      "title": "Internal server error",
      "detail": "PDOException: SQLSTATE[42703]: Undefined column: 7 ERROR:  column \"id\" does not exist\nLINE 1: ...number\", \"read_time\") values ($1, $2, $3, $4) returning \"id\"\n                                                                   ^ in /var/www/html/vendor/illuminate/database/Connection.php:321\nStack trace:\n#0 /var/www/html/vendor/illuminate/database/Connection.php(321): PDOStatement->execute(Array)\n#1 /var/www/html/vendor/illuminate/database/Connection.php(655): Illuminate\\Database\\Connection->Illuminate\\Database\\{closure}(Object(Illuminate\\Database\\PostgresConnection), 'insert into \"us...', Array)\n#2
```

vjousse:
Indeed, I've stopped my work at the installer.

PR #862, opened 2016-03-16
Show alert for unverified User

### LINKED ISSUE #709: Add "you need to confirm your email" informational message with "resend" functionality

sijad:
Things that I wasn't sure about:
* adding js codes to `js/forum/src/initializers/boot.js` isn't a good practice IMO, it should be changed, but I was sure where should I add those codes
* I'm not sure using `resend-confirmation` as API path, also using `GET` method
* in controller I used `actor` as `user`, like `Read all notifications as read` controller
* I've reused some codes of [EmailConfirmationMailer.php](https://github.com/flarum/core/blob/master/src/Core/Listener/EmailConfirmationMailer.php) to resend  the email, I think a separated class for this is appropriate but I wasn't sure where should I add this class

also https://github.com/flarum/flarum-ext-english/pull/68 need to be merged

fixes #709

### DIFF
--- a/js/forum/src/initializers/alertEmailConfirmation.js
+++ b/js/forum/src/initializers/alertEmailConfirmation.js
@@ -1,0 +1,36 @@
+import Alert from 'flarum/components/Alert';
+import Button from 'flarum/components/Button';
+
+/**
+ * The `alertEmailConfirmation` initializer shows an Alert if loggend in
+ * user's email is not confirmed
+ *
+ * @param {ForumApp} app
+ */
+export default function alertEmailConfirmation(app) {
+  const user = app.session.user;
+
+  if (!user || user.isActivated()) return;
+
+  let alert;
+
+  const resendButton = Button.component({
+    className: 'Button Button--link',
+    children: app.translator.trans('core.forum.user_confirmation.resend_button'),
+    onclick: () => {
+      app.request({
+        method: 'POST',
+        url: app.forum.attribute('apiUrl') + '/users/' + user.id() + '/send-confirmation',
+      }).then(() => app.alerts.dismiss(alert));
+    }
+  });
+
+  app.alerts.show(
+    alert = new Alert({
+      type: 'error',
+      dismissible: false,
+      children: app.translator.trans('core.forum.user_confirmation.alert_message'),
+      controls: [resendButton]
+    })
+  );
+}
--- a/src/Api/ApiServiceProvider.php
+++ b/src/Api/ApiServiceProvider.php
@@ -179,6 +179,13 @@ class ApiServiceProvider extends AbstractServiceProvider
             $toController('Flarum\Api\Controller\DeleteAvatarController')
         );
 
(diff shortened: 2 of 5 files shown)


tobscure:
Awesome, will review tomorrow :)

src/Api/ApiServiceProvider.php
```diff
@@ -124,6 +124,13 @@ protected function populateRoutes(RouteCollection $routes)
             $toController('Flarum\Api\Controller\ForgotPasswordController')
         );
 
+        // Resend confirmation email
+        $routes->get(
```
franzliedke: If this causes a mail to be sent, this should not be a `GET` action...

sijad: `POST` then?

franzliedke: Yep.

tobscure:
Sorry for the delay, and thanks very much for the PR! I think we should make the following changes:

* Change the endpoint to POST /api/users/[id]/send-confirmation, make it send the confirmation to the user with that ID only if they are the current actor, otherwise throw a PermissionDeniedException... And only if their account is not already activated, otherwise throw a 403 error. 

* Move the alert code out of boot.js and into a new initializer function, called "alertEmailConfirmation" or something. (Feel free to think of a better name!)

I'm happy to refactor the mailer code once this has been merged :)

sijad:
@tobscure I'm sorry for delay, is it better now?

dcsjapan:
@tobscure Should the message box have a cancel button so it can be canceled without resending?

tobscure:
@dcsjapan All alerts automatically have an 'x' button to dismiss them :)

js/forum/src/initializers/alertEmailConfirmation.js
```diff
@@ -0,0 +1,36 @@
...
+        method: 'POST',
+        url: app.forum.attribute('apiUrl') + '/users/' + user.id() + '/send-confirmation',
+      }).then(() => app.alerts.dismiss(alert));
+    }
+  });
+
+  app.alerts.show(
+    alert = new Alert({
+      type: 'error',
+      dismissible: false,
```
sijad: @tobscure it's not `dismissible` so `x` button is not available, as it appears in every pages I thought it's better that way, let me know if you think revert this

tobscure: Ah OK. I'm about to merge this in so I'll have a play :) Thanks!

dcsjapan:
> All alerts automatically have an 'x' button to dismiss them :)

Okay, I suspected that might be the case. Just checking! :smile:

tobscure:
@sijad I just committed some tweaks – haven't been able to test fully though because emails aren't sending properly on my local server. Would you mind doing me a big favour and quickly checking that my changes are working OK? :)

sijad:
@tobscure what grate changes :+1: and yes they works very well

PR #922, opened 2016-04-09
Update Extension generator

### LINKED ISSUE #915: extension skeleton, generator overhaul
Related to:
https://discuss.flarum.org/d/2435-now-in-beta-5-how-to-build-extension

The code:
https://github.com/flarum/core/blob/master/src/Console/Command/GenerateExtensionCommand.php#L54
- Supresses any errors when mkdir fails, thus failing all over without any indication what went wrong. Mkdir returns a boolean based on the state of running it, also it might be wise to simply use the file adapter anyway?
- The command uses the depreciated extensions directory. Even if you want to create development packages there, it might cause too much confusion.
- Wouldn't it be better if the command used studio instead?

sijad:
It's probably buggy and need to be tested
fixes #915


sijad:
now `'cache:clear'` command also fire after extension generated

franzliedke:
Kewl. I like where this is going. :+1: 

And it's a cool idea to support *both* command line options and asking the user for options.

The biggest remaining problems is the desired behavior, though:
- Obviously, with the `extensions` directory dropped, we have no good place to store generated extensions. Developing in the `vendor` dir is meh.
- You went with a `workbench` directory. Is that what we want?
- Should we use my Studio package instead of dealing with path repositories?
- In combination with that, what about a skeleton extension package on GitHub instead? (I aim to build custom skeletons into Studio anyway.)

sijad:
@franzliedke such nice ideas

I really like downloading skeleton from github

I get `workbench` folder idea of @Luceos's [post](https://discuss.flarum.org/d/1608-extension-development-using-composer-repositories-path), I wasn't sure either
I think your `Studio package` is a way better idea, I'll look into that

src/Console/Command/GenerateExtensionCommand.php
```diff
@@ -10,116 +10,367 @@
...
+     * @param string $src
+     * @param string $dst
+     * @param array $replacements
+     * @return void
+     */
     protected function recursiveCopy($src, $dst, $replacements = [])
     {
         $dir = opendir($src);
-        @mkdir($dst);
+        @mkdir($dst, 0775, true);
```
Luceos: See my comment on #915 ; using `@` suppresses any error messages when creating a directory. The code should however check for its return code (true or false) whether it was succesful.

Luceos:
Amazing PR @sijad love your contribution!

franzliedke:
Hey @sijad, thanks again, I promise I will take care of this soon!

sijad:
no worries, should I close this?

franzliedke:
No, leave it open, it's a good reminder. :)

tobscure:
> what about a skeleton extension package on GitHub instead

Problem with that is there are lots of places in extensions where the extension `vendor/package` or namespace needs to be specified:

* composer.json
* gulpfile.js x2
* main.js x2
* all php files

So it would be a pain to have to do that all manually once you clone the skeleton, which is where a "generator" comes in handy.

sijad:
@tobscure I guess a git repo is easier to deal with, also we can have more options using a [yeoman](http://yeoman.io/) generator.

sijad:
@tobscure this PR need lots of improvements, if you let me know your suggestions how can I make this better, I'd really appreciate it.

src/Console/Command/GenerateExtensionCommand.php
```diff
@@ -10,116 +10,367 @@
...
+        if (! isset($composer['repositories'])) {
+            $composer['repositories'] = [
+                [
+                    'type' => 'path',
+                    'url' => 'workbench\/*\/',
+                ],
+            ];
+        }
+
+        $composer['require'][$packageName] = 'dev-master';
```
dav-is: Shouldn't this be set to `*@dev`?

src/Console/Command/GenerateExtensionCommand.php
```diff
@@ -10,116 +10,367 @@
...
+
+        file_put_contents($composerPath, json_encode($composer, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES));
+
+        passthru('cd '.escapeshellarg(public_path()).'; composer update');
```
dav-is: This should be more specific? Don't want to update all their packages if they don't want to.

`passthru('cd '.escapeshellarg(public_path()).'; composer update '.$packageName);`

src/Console/Command/GenerateExtensionCommand.php
```diff
@@ -10,116 +10,367 @@
...
+            'License',
+            function ($input) {
+                return true;
+            }
+        );
+
+        $package = $vendor.'-'.$name;
 
-        $description = $this->ask('Description:');
+        $dir = public_path().'/workbench/'.$package;
```
dav-is: Shouldn't this be `$dir = base_path('workbench/'.$package);` ?

src/Console/Command/GenerateExtensionCommand.php
```diff
@@ -10,116 +10,367 @@
...
 
-        $authorName = $this->ask('Author name:');
+        if (is_dir($dir)) {
+            throw new Exception($dir.' folder already exists');
+        }
 
-        $authorEmail = $this->ask('Author email:');
+        $composerPath = public_path().'/composer.json';
```
dav-is: `$composerPath = base_path('composer.json');` ?

dav-is:
I think it might have gone unnoticed, but Flarum supports a public directory. That means the script would get a 404 if someone was using the public directory because the composer.json wouldn't be accessible to the public. Also, is there a reason you weren't passing the rest of the directory as a parameter? @sijad

Luceos:
@sijad I'll close this for now. We can always come back to it if the need arises..

Issue #251, opened 2015-08-27
Experiment with custom scrollbars on Windows


tobscure:
Mac OS and iOS hide scrollbars by default, which is good for Flarum's UI. But on Windows, we have things like this:

![](http://i.imgur.com/Fx37aq2.png)

I don't know much about styling custom scrollbars, but I think we should try and make them look a little nicer **if there's an easy CSS-only way.**

tobscure:
See https://github.com/flarum/core/pull/923#issuecomment-209473585

askvortsov1:
@davwheat don't remember off the top which issue we discussed this on, but turns out it's a milestone item, yay!

SychO9:
is this the one #2531 ?

askvortsov1:
Ah yes! Thank you!

PR #923, opened 2016-04-10
Nice CSS scrollbars - Issue #251

mattsterr:
A starting point for issue flarum/issue-archive#423 

I'm not sure about that horizontal scrollbar from a ux point of view, looks as if it should sit just below the list of tags (as opposed to at the top of the discussions). Adding a `height:10px` after line 10-ish will neaten it up, but wasn't sure what people thought should be done about it.  

Here's what I have:
image

### DIFF
--- a/less/lib/scaffolding.less
+++ b/less/lib/scaffolding.less
@@ -8,6 +8,7 @@
 
 ::-webkit-scrollbar {
   width: 7px;
+  height: 10px;
 }
 ::-webkit-scrollbar-track {
   -webkit-border-radius: 10px;
--- a/less/lib/scaffolding.less
+++ b/less/lib/scaffolding.less
@@ -6,6 +6,24 @@
   }
 }
 
+::-webkit-scrollbar {
+  width: 7px;
+}
+::-webkit-scrollbar-track {
+  -webkit-border-radius: 10px;
+  background: fade(#000, 10%);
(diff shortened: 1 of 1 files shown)


Kakifrucht:
I like it, using this in my forum now.

datitisev:
The horizontal scrolling bar might be too thick.
Anyhow, good job!
Does the color come from the forum primary color? If not, it should :/

mattsterr:
@datitisev Agreed, so if we all agree I'll update the code. It does use the primary colour.

Thanks guys! :)

franzliedke:
Thanks for the pull request! :)

We'll have to wait for feedback from @tobscure here. But I _think_ that flarum/issue-archive#423 wasn't about all scrollbars (e.g. the global one), but only about the one below the tags list on tablet widths, so that should be changed.

That said, I'd definitely like it a little thinner, too.

tarunmarkose:
While this does improve the scroll bar, I'm still not able to come to peace
with horizontal scroll on Flarum.

Can't we consider opening up a list or cloud of tags downwards when someone
hits "Tags" in this view? Like a drawer, it can go back up once a tag
selection is made.

tobscure:
Yeah, I think that's a good point @tarunmarkose. Going to update the original issue with that suggestion.

Thanks for the attempt @mattsterr!

Issue #893, opened 2016-03-29
Windows loses image file extensions


jordanjay29:
[As per this topic at discuss.flarum](https://discuss.flarum.org/d/2359-avatar-upload-still-doesnt-work-in-beta-5)

It seems like Flarum is saving the image file to a .tmp extension, but while Linux can recover the true extension and finish the upload, Windows loses its mind and produces a 500 error instead.

dcsjapan:
Discussion of the cause and workarounds begins at [line 16](https://discuss.flarum.org/d/2359-avatar-upload-still-doesnt-work-in-beta-5/16) of that thread.

Averell7:
The real answer begins at post 24 of the above mentioned thread. 
It is proved that creating a temp file with .jpg extension instead of tmp solves the problem in Windows. No problem if the file is really a gif, png or jpg file, it works for all three types.

Averell7:
Suggested solution :
in project/vendor/flarum/core/src/Core/Command/UploadAvatarHandler.php

```
$tmpFile1 = tempnam($this->app->storagePath().'/tmp', 'avatar');
$tmpFile = str_replace(".tmp", ".jpg", $tmpFile1);         
register_shutdown_function(create_function('', "unlink('{$tmpFile1}');"));        
$command->file->moveTo($tmpFile);
```

PR #927, opened 2016-04-12
Fix avatar upload on Windows servers - Issue #893

KazeFlame:
As said in Intervention Image docs on save api: The image type will be defined by the file extension. For example if you pass foo.jpg the image will be saved as a JPG file.

Windows servers will generate temporary name and include .tmp extension.

So instead of using the save api. It will first encode the image to JPG format then save it using the native file_put_contents.

### DIFF
--- a/src/Core/Command/UploadAvatarHandler.php
+++ b/src/Core/Command/UploadAvatarHandler.php
@@ -99,7 +99,7 @@ class UploadAvatarHandler
 
         $encodedImage = $manager->make($tmpFile)->fit(100, 100)->encode('jpg', 100);
 
-        @file_put_contents($tmpFile, $encodedImage);
+        file_put_contents($tmpFile, $encodedImage);
 
         $this->events->fire(
             new AvatarWillBeSaved($user, $actor, $tmpFile)
--- a/src/Core/Command/UploadAvatarHandler.php
+++ b/src/Core/Command/UploadAvatarHandler.php
@@ -96,7 +96,10 @@ class UploadAvatarHandler
         $this->validator->assertValid(['avatar' => $file]);
 
         $manager = new ImageManager;
-        $manager->make($tmpFile)->fit(100, 100)->save();
+
+        $encodedImage = $manager->make($tmpFile)->fit(100, 100)->encode('jpg', 100);
+
+        @file_put_contents($tmpFile, $encodedImage);
 
         $this->events->fire(
             new AvatarWillBeSaved($user, $actor, $tmpFile)


KazeFlame:
fixes #893

src/Core/Command/UploadAvatarHandler.php
```diff
@@ -96,7 +96,10 @@ public function handle(UploadAvatar $command)
         $this->validator->assertValid(['avatar' => $file]);
 
         $manager = new ImageManager;
-        $manager->make($tmpFile)->fit(100, 100)->save();
+
+        $encodedImage = $manager->make($tmpFile)->fit(100, 100)->encode('jpg', 100);
+
+        @file_put_contents($tmpFile, $encodedImage);
```
Luceos: Thanks for this PR. Just two questions;

- using `@` will silence any errors, thus you'll never know whether saving the temporary file succeeded.
- why did you fall back on `file_put_contents`, the manager used `save()` previously, can't we keep using that?

KazeFlame:
I've looked up the code of `Intervention\Image\Image::save` and it uses `@file_put_contents`. The `save` function will automatically detect the format based on the file extension. So if we pass `.tmp` extension, the format will be `tmp` and that causes the error. The `tempnam` function will add `tmp` extension on Windows servers. If the `save` function has a parameter to manually choose the encoding format, we could use the `save` function.
https://github.com/Intervention/image/blob/master/src/Intervention/Image/Image.php#L122-L152

src/Core/Command/UploadAvatarHandler.php
KazeFlame: I've looked up the code of `Intervention\Image\Image::save` and it uses `@file_put_contents`. The `save` function will automatically detect the format based on the file extension. So if we pass `.tmp` extension, the format will be `tmp` and that causes the error. The `tempnam` function will add `tmp` extension on Windows servers. If the `save` function has a parameter to manually choose the encoding format, we could use the `save` function.
https://github.com/Intervention/image/blob/master/src/Intervention/Image/Image.php#L122-L152

franzliedke: And you cannot use `save` after calling `encode` either?

franzliedke:
Thanks for the PR! I'll have to think about this a bit, but will merge it soon.

src/Core/Command/UploadAvatarHandler.php
Luceos: I'm also curious whether that works. Tmp is not one of the identifiable extensions, so the encoding should force during save, see:

http://image.intervention.io/api/encode

KazeFlame: It doesn't work. That's why we have to use the native `file_put_contents` to save the encoded image.

franzliedke:
Thank you very much, @KazeFlame !

PR #934, opened 2016-04-16
#249 AvatarEditor should show an upload icon instead of a user's default avatar

### LINKED ISSUE #249: AvatarEditor should show an upload icon instead of a user's default avatar
The AvatarEditor component (shown on a user's profile if the avatar is editable) should display a big "upload" icon in place of the avatar if the user hasn't uploaded one. This will make discovery of the upload function more obvious.

What needs to be done:

* Add a check in AvatarEditor to see if the user's `avatarUrl` is empty – if it is, show an upload icon instead of their avatar.

datitisev:
This kind of fixes #249
When a user doesn't have an avatar, over their "avatar" it will show an upload button with opacity of 70%.
It isn't very clear, but the users hopefully sees that something is up with the avatar :wink:
When you click it it directly calls the `upload` function

Screenshot:

Flarum User Profile: No avatar and shows upload button

### DIFF
--- a/js/forum/src/components/AvatarEditor.js
+++ b/js/forum/src/components/AvatarEditor.js
@@ -37,10 +37,10 @@ export default class AvatarEditor extends Component {
     return (
       <div className={'AvatarEditor Dropdown ' + this.props.className + (this.loading ? ' loading' : '')}>
         {avatar(user)}
-        <a className="Dropdown-toggle"
-          data-toggle="dropdown"
-          onclick={this.quickUpload.bind(this)}>
-          {this.loading ? LoadingIndicator.component() : icon('pencil')}
+        <a className={ user.avatarUrl() ? "Dropdown-toggle" : "Dropdown-toggle AvatarEditor--noAvatar" }
+           data-toggle="dropdown"
+           onclick={this.quickUpload.bind(this)}>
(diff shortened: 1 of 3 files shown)


js/forum/src/components/AvatarEditor.js
```diff
@@ -37,10 +37,10 @@ export default class AvatarEditor extends Component {
     return (
       <div className={'AvatarEditor Dropdown ' + this.props.className + (this.loading ? ' loading' : '')}>
         {avatar(user)}
-        <a className="Dropdown-toggle"
-          data-toggle="dropdown"
-          onclick={this.quickUpload.bind(this)}>
-          {this.loading ? LoadingIndicator.component() : icon('pencil')}
+        <a className={ user.avatarUrl() ? "Dropdown-toggle" : "Dropdown-toggle AvatarEditor--noAvatar" }
+           data-toggle="dropdown"
+           onclick={this.quickUpload.bind(this)}>
```
franzliedke: Why these extra spaces? :P

datitisev: I don't know, I probably just wanted it so I could read it better just in case it didn't work :stuck_out_tongue_winking_eye:

franzliedke:
Awesome job, looks perfect!

PR #936, opened 2016-04-17
#814 Better code style

### LINKED ISSUE #814: Better post <code> style
https://discuss.flarum.org/d/1589-code-style

datitisev:
This just makes the inline code look like Scoth.IO's.
It fixes #814 

I'll make any necessary changes.

CSS
```css
.Post-body code {
    font-family: source-code-pro, Monaco, Consolas, "Courier New", monospace;
    padding: 2px 6px;
    background: #F2F7F9;
    color: #256FC7;
    font-size: 90%;
    border-radius: 4px
  }
```

---
Screenshot:

Flarum New Code Style

### DIFF
--- a/less/forum/Post.less
+++ b/less/forum/Post.less
@@ -116,6 +116,14 @@
     border-bottom: 2px dotted @body-bg;
     margin: 1em 0;
   }
+  code {
+    font-family: source-code-pro, Monaco, Consolas, "Courier New", monospace;
+    padding: 2px 6px;
+    background: #F2F7F9;
+    color: #256FC7;
+    font-size: 90%;
+    border-radius: 4px
+  }
   pre {
     border: 0;
     padding: 15px;
@@ -127,6 +135,13 @@
     .hljs {
       padding: 0;
       background: none;
+    }
+    code {
+      padding: 0;
+      background: transparent;
+      color: inherit;
+      font-size: 100%;
+      border-radius: 0;
     }
   }
   h1 {
--- a/less/forum/Post.less
+++ b/less/forum/Post.less
@@ -119,7 +119,7 @@
   code {
     font-family: source-code-pro, Monaco, Consolas, "Courier New", monospace;
     padding: 2px 6px;
-    background: #F2F7F9;
+    background: #60646D;
     color: #256FC7;
     font-size: 90%;
     border-radius: 4px
--- a/less/forum/Post.less
+++ b/less/forum/Post.less
@@ -119,8 +119,8 @@
   code {
     font-family: source-code-pro, Monaco, Consolas, "Courier New", monospace;
     padding: 2px 6px;
-    background: #60646D;
-    color: #256FC7;
+    background: @code-bg;
+    color: @code-color;
     font-size: 90%;
     border-radius: 4px
   }
--- a/less/lib/variables.less
+++ b/less/lib/variables.less
@@ -37,6 +37,9 @@
   @control-danger-color:           #d66;
 
   @overlay-bg:                     fade(@secondary-color, 90%);
+
+  @code-bg:                        #F2F7F9;
+  @code-color:                     #60646D;
 }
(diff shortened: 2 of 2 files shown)


franzliedke:
I like it, almost! ;)

I'd prefer the text color for the inline code block to be less blue and more gray (similar to what was suggested [in this discussion](https://discuss.flarum.org/d/1589-code-style)). Currently, it looks too similar to links IMO.

What did you change for the code blocks?

datitisev:
@franzliedke Yeah, I also thought that about the code being blue...

What do you mean, "code blocks"?
Do you mean the extra CSS for code blocks?
That was so it doesn't use the inline code style :smiley:

franzliedke:
> That was so it doesn't use the inline code style

Ah, I see... it resets the other CSS. Makes sense.

Can you change the text color? :)

datitisev:
@franzliedke I would do it, but it is too late for me. :(
All I would have to do is find an appropriate color that also goes with Flarum theme and/or fits nicely :smile:

Thanks for your feedback!

datitisev:
Flarum: New Code Style (Using #60646D

Does this look better?
I'm using `#60646D` for the color.
I don't want to commit "unwelcomed" changes :wink:


---

_I just figured out you can drag & drop images to the GitHub editor... didn't even see this:_
> _Attach files by dragging & dropping or [selecting them](#)._

_😄_

viruxe:
It's subtle. It like it.

less/forum/Post.less
```diff
@@ -116,6 +116,14 @@
     border-bottom: 2px dotted @body-bg;
     margin: 1em 0;
   }
+  code {
+    font-family: source-code-pro, Monaco, Consolas, "Courier New", monospace;
+    padding: 2px 6px;
+    background: #60646D;
+    color: #256FC7;
```
franzliedke: Ah, something I forgot till now. You should use one of the existing color variables probably instead of copying the value here... that way, theming from the admin panel will be easier later on.

datitisev: @franzliedke What variable should I use, then?
I just had a look at them, but I'm not good with color codes (RGB or HEX) at all :wink:.

Reflic: @datitisev Maybe have a look on one of these variables, IMO these should look good.
https://github.com/datitisev/core/blob/814-better-post-code-style/less/lib/variables.less#L30-L31

datitisev: @Reflic Thanks, I'll look into them

datitisev:
I'm sorry about the time that has passed with no progress to this pull.
I have been busy.
I hope to be able to continue tomorrow.

Thanks for all your feedback.

franzliedke:
Yes, your variable suggestion makes sense. :)

franzliedke:
Depending on their values, you might be able to assign other existing variables to `$code-bg` and `$code-color` when creating them...

datitisev:
@franzliedke All I did was replace the values of `color` and `background` with the new variables, whenever it compiles, the result should be exactly the same (or similar).
About the contrast issue, I could probably try to find opposite colors of the ones I'm currently using for the dark mode, and just change the variables there automatically.

I don't understand your suggestion of 
> you might be able to assign other existing variables to $code-bg and $code-color when creating them...

`$code-bg` : `#60646D`
`$code-color` : `#256FC7`

datitisev:
I've just (kind of) fixed the problem with contrast (dark mode)!
Same CSS for both, different result:

captura de pantalla 2016-05-08 a las 4 55 30 pm
captura de pantalla 2016-05-08 a las 4 55 45 pm

How does that look?

franzliedke:
Thank you so much, David!

I finally merged this (manually) and made some [last tweaks](https://github.com/flarum/core/commit/aca497e7be618ff10dbea22ac1b99d15a544676b), thanks for getting it started. :fireworks:

PR #938, opened 2016-04-20
Public and base directory can be separated

bmalex88:
Existing users don't have to make any change. In their case, the base and public path will be the same. In order to use separated base and public path, **before installing Flarum**, you need to specify them as arguments of the `Server` object constructor, in _index.php, api.php_ and _admin.php_ respectively:

`$server = new Flarum\Forum\Server(BASE_PATH, PUBLIC_PATH);`

If `PUBLIC_PATH` is not specified, it will be the taken from `BASE_PATH`. Extensions assets will be copied to the public path, with all the assets (CSS, JS, avatars)

### DIFF
--- a/src/Extension/ExtensionManager.php
+++ b/src/Extension/ExtensionManager.php
@@ -63,11 +63,11 @@ class ExtensionManager
      */
     public function getExtensions()
     {
-        if (is_null($this->extensions) && $this->filesystem->exists(base_path('vendor/composer/installed.json'))) {
+        if (is_null($this->extensions) && $this->filesystem->exists($this->app->basePath().'/vendor/composer/installed.json')) {
             $extensions = new Collection();
 
             // Load all packages installed by composer.
-            $installed = json_decode($this->filesystem->get(base_path('vendor/composer/installed.json')), true);
+            $installed = json_decode($this->filesystem->get($this->app->basePath().'/vendor/composer/installed.json'), true);
 
             foreach ($installed as $package) {
                 if (Arr::get($package, 'type') != 'flarum-extension' || empty(Arr::get($package, 'name'))) {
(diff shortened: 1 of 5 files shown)


src/Foundation/AbstractServer.php
```diff
@@ -25,7 +25,12 @@
     /**
      * @var string
      */
-    protected $path;
+    protected $base_path;
+
+    /**
+     * @var string
+     */
+    protected $public_path;
```
franzliedke: `$basePath` and `$publicPath`, please! :)

bmalex88: Ooops! Habits. Will change in a moment

src/Extension/ExtensionManager.php
```diff
@@ -318,6 +318,6 @@ public function isEnabled($extension)
      */
     protected function getExtensionsDir()
     {
-        return public_path('vendor');
+        return base_path('vendor');
```
franzliedke: Can you please use the methods on `$this->app` in all occurrences in this file?

franzliedke:
Great job! Just some minor comments.

One last thing I'm thinking about: shouldn't we have two methods (one setter for each path) on the `Application` class instead of the `setPaths()` method?

bmalex88:
I guess I should have left it like I did it first, with `setBasePath()` and `setPublicPath()`. I will change it

src/Extension/ExtensionManager.php
bmalex88: I guess you mean replacing `base_path('vendor')`by `$this->app->basePath().'vendor'`?

franzliedke: Yep.

tobscure:
Perfect 👌

tobscure:
Actually I'll do them myself, because I want to start using this!

Issue #937, opened 2016-04-18
Password reset has no validation rules


JohanSombekke:
When a user signs up validation rules are applied to the password (example: at least 8 characters). With a password reset no validation rules are applied.
(Flarum beta5)

franzliedke:
Nice catch, thank you!

poush:
I want to try this. I am familiar with native laravel app structure as well as dingo/api but not with your app setup. So can you help me out where to start with this bug ? or link of some quick introduction to app structure of flarum/core

franzliedke:
@poush That's great, thank you!

The routes for the forum frontend can be found in the [`ForumServiceProvider` class](https://github.com/flarum/core/blob/7192c4391bee006ccc2de3db6caa89803d72d130/src/Forum/ForumServiceProvider.php#L58). There, you can see that the [`SavePasswordController`](https://github.com/flarum/core/blob/7192c4391bee006ccc2de3db6caa89803d72d130/src/Forum/Controller/SavePasswordController.php) is responsible for handling the password reset requests.

That's where you need to add the validation, possibly by using the [`UserValidator` class](https://github.com/flarum/core/blob/7192c4391bee006ccc2de3db6caa89803d72d130/src/Core/Validator/UserValidator.php).

I hope that helps you in getting started. :)

poush:
From little digging into core. I saw there RegisterUserHandler where validation is done in registration process and similarly in RequestPasswordResetHandler when generating password but there is no Handler for saving password.

It goes like savePassword route to SavePasswordController then ends there. So is there need of some ```Flarum\Core\Command\SavePasswordHandler``` ? or I should do validation part in SavePasswordController ?

Luceos:
@poush there is this:

https://github.com/flarum/core/blob/master/src/Core/Command/RequestPasswordResetHandler.php

poush:
@Luceos but this handles the request for token generation and mailing user the token generated.

Luceos:
My apologies, my bad :) Yes this should be modified in
https://github.com/flarum/core/blob/master/src/Forum/Controller/SavePasswordController.php

I wasn't sure but there has been talk about dropping the handler, because it generated too much confusion. So simply hack it in there and you should be good for a merge already. We can refactor it if needed in a follow up issue.

poush:
please check #943

franzliedke:
Thanks a lot, @poush !

poush:
Thanks you guys for guiding me to make my first contribution on flarum

[...]

PR #943, opened 2016-04-28
Validation on password change

poush:


Luceos:
Link to issue: https://github.com/flarum/core/issues/937

franzliedke:
:+1: Looking good, thank you!

@tobscure Apparently, the `AbstractValidator` only checks the rules for the keys that are passed in. That is nice for cases like this one, where we only want to check a specific attribute. But what about other cases, where the `required` rule is supposed to make sure that a key exists?

src/Forum/Controller/SavePasswordController.php
```diff
@@ -52,6 +59,8 @@ public function handle(Request $request)
         $password = array_get($input, 'password');
         $confirmation = array_get($input, 'password_confirmation');
 
+        $this->validator->assertValid(compact('password'));
+
         if (! $password || $password !== $confirmation) {
```
poush: @franzliedke I think we can make use of validator for checking password confirmation so it can throw localised error as in Laravel

franzliedke: That's true, good idea!

tobscure:
@franzliedke The `AbstractValidator` validates attributes rather than entities as a whole. It's the caller's responsibility to make sure all of the attributes that need to be validated are passed in. See for example in RegisterUserHandler: we validate `$user->getAttributes()`, where `$user` has been hydrated in `User::register()` and the `UserWillBeSaved` event. The `required` rule will check for a non-empty value for an attribute rather than its presence.

PR #946, opened 2016-04-28
Fixes #945 - Incorrect Scrubber count value

### LINKED ISSUE #945: Clicking and dragging Scrubber beyond final post causes count to exceed max value
flarum-post-count

augiwan:
Clicking and dragging the Scrubber beyond the final post causes the counter to exceed the total post count. This commit fixes that issue. 

Please review code.

### DIFF
--- a/js/forum/src/components/PostStreamScrubber.js
+++ b/js/forum/src/components/PostStreamScrubber.js
@@ -61,7 +61,7 @@ export default class PostStreamScrubber extends Component {
     const unreadPercent = count ? Math.min(count - this.index, unreadCount) / count : 0;
 
     const viewing = app.translator.transChoice('core.forum.post_scrubber.viewing_text', count, {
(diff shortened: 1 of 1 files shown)


franzliedke:
You can just use `Math.max()`. :)

franzliedke:
See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/max.

augiwan:
@franzliedke We don't want the max value, but the post value limited by the total post count.

When post = 49 and count = 50, then we want 49.
When post = 51 and count = 50, then we want 50.

We could use something like `Math.min(count, Math.max(0, post))` but I thought the conditional operator was clearer.

franzliedke:
Whoops, I meant Math.min:

    Math.min(Math.ceil(this.index + this.visible), count)

franzliedke:
@augiwan Can you push that quick fix or do you want me to make it?

augiwan:
@franzliedke Done! :+1:

franzliedke:
Thank you very much!

PR #947, opened 2016-04-28
Updated s9e\TextFormatter to 0.5.0

JoshyPHP:
The `timeout` property has been removed from the JavaScript minifiers. You need to set it on the HTTP client instead.


tobscure:
Thanks @JoshyPHP! 🍻

PR #949, opened 2016-05-01
Domain based configuration file

bmalex88:
A simple way of allowing multiple domains to use the same source but different configuration files (hence databases, enabled extensions, languages). Existing users will continue to use their default configuration file until next installation.
Generated configuration file will be named as _config.domain.tld.php_

### DIFF
--- a/src/Foundation/AbstractServer.php
+++ b/src/Foundation/AbstractServer.php
@@ -66,7 +66,7 @@ abstract class AbstractServer
                 $host = @parse_url($_SERVER['SERVER_NAME'], PHP_URL_HOST);
 
                 if ($host) {
-                    if (substr($host, 0, 4) == "www.") {
+                    if (substr($host, 0, 4) == 'www.') {
                         $host = substr($host, 4);
                     }
 
--- a/src/Foundation/AbstractServer.php
+++ b/src/Foundation/AbstractServer.php
@@ -61,6 +61,17 @@ abstract class AbstractServer
 
         if (file_exists($file = $this->basePath.'/config.php')) {
             $this->config = include $file;
+        } else {
+            if (isset($_SERVER['SERVER_NAME'])) {
+                $host = @parse_url($_SERVER['SERVER_NAME'], PHP_URL_HOST);
+                if (!$host) $host = $_SERVER['SERVER_NAME'];
+
+                if (substr($host, 0, 4) == "www.") $host = substr($host, 4);
+
+                if (file_exists($file = $this->basePath.'/config.'.$host.'.php')) {
+                    $this->config = include $file;
+                }
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


bmalex88:
I will have to work on separating the elements in _storage_, too. Will PR again later

franzliedke:
Hmm, let's wait for input from @tobscure. How do you do it for the hosting service?

tobscure:
Currently we provision an nginx virtual host for each domain which points to a unique index.php/api.php/admin.php:

```
require '/path/to/flarum/vendor/autoload.php';

$server = new Flarum\Forum\Server('/path/to/flarum');

$server->setConfig(include __DIR__.'/../config.php');
$server->setPublicPath(__DIR__);
$server->setStoragePath(__DIR__.'/../storage');

$server->listen();
```

However this is just a stopgap, we are planning to move towards a more dynamic PHP/database-based multisite setup.

I think we should leave an opinionated multisite setup like this out of core (at least for now), especially given that we already have the APIs to make it happen (see above). Need more information/experience with how people use multisite before we can make a call on a "best practice" way to do it.

bmalex88:
Well, this works, too, but with my approach the configuration file ad storage paths would be automatically generated when installing for a domain pointing to the same source code. You wouldn't have to manually create a configuration file for each domain and you wouldn't need the same index, api and admin file duplicated for each domain

franzliedke:
@bmalex88 Huh? Why does your approach mean files can be generated automatically, and with Toby's, they have to be created manually?

In any case, your approach would be bound to separate domains, which wouldn't have to be the only possible multi-site setup. I guess that's what Toby meant with too opiniated.

ahrasis:
I am not familiar with flarum yet but I think multi forum will be a lot of works. I like multi forum feature but we have done similar to this in ElkArte as an sub forum addon. SMF, VB and XF also have it as an extension / plugin / addon / mod.

It will also not enough to simply create it during installation, admin need to manage that from time to time whether to add or remove domain, rename its name, change its theme, fix its board / tag / categories, determine its language etc. We may even need a separate admin for each forum.

To conclude, it definitely should be an extension of flarum and not its core.

franzliedke:
@ahrasis The functionality does not need to be built into core, I agree. However, we should be careful that building such a thing remains easy with an extension.

ahrasis:
Yes. That is also true.

From what I understand, the intended approach is to allow flarum to have multi forum based on domain with multi database and different storage files. As seems mainly use for testing, I guess the intention is to allow the user to eventually move on and run their own forum independently later on.

This is different from our current approach which I think is simpler i.e. by having multi sub forums  based on domain but with single database and same storage files. The intention is for admin / owner to create more forums without having to reinstall its software while sub forum can be managed from its own control panel.

From my early flarum test, a simple `'url' => 'http://' . $_SERVER['SERVER_NAME'],` in config.php can already allow multiple domain to use the same installation. I will test further to see how certain tags can be limited to domain, how registration can be defaulted to domain based group members and how theme or language can fixed to it as well.

As I said, I am new to flarum, and for the time being I am still not sure how an extension is supposedly created.

Issue #127, opened 2015-06-20
SQL query audit + add database indexes


tobscure:
Lots of low-hanging fruit to improve performance and efficiency. Basically we just need to set up a query log, do every action possible on the forum, and then inspect each query individually and optimize where it's not too hard.

Also add SQL indexes (we don't have any currently apart from primary keys).

There are some more interesting cases too, like:
- Flarum\Forum\Actions\IndexAction line 58: we send an API request to get the current user, but we already have the user data; we just want it formatted in an API response. We could technically call a serializer directly, but then we don't quite get all of the response features (like sideloaded data), and also Forum would depend on Api. I don't think we can optimize this case.
- Flarum\Core\Repositories\EloquentPostRepository (see big comment at top of file)

tobscure:
No SQL query optimization has been done on Flarum yet, so there is probably quite a bit of room for performance gains here.

What needs to be done:

* Create a "Debug" extension which attaches the page load time and a list of queries that were run to each API response (under the "meta" key).
* Generate a large dataset (hundreds of thousands of discussions/posts) to test with.
* Analyse the queries that are run on requests to each API endpoint and work out ways to reduce/optimize them – especially making sure there are no n+1 situations.
* Add database indexes where appropriate.

michaelmior:
I'm going to suggest [https://github.com/maximebf/php-debugbar](php-debugbar) for this purpose as it's easy to use and can quickly collect a lot of helpful data.

tobscure:
Thanks, will check that out.

younes0:
if it can help: Laravel's package of maximebf/debugbar: https://github.com/barryvdh/laravel-debugbar

believer-ufa:
Hello! You can use [Mysql profiling builted functions](https://dev.mysql.com/doc/refman/5.5/en/show-profile.html), for example.

BartVB:
I would like to suggest another course of action. Analysing queries is fun stuff, but this mostly results in theoretical optimisation strategies. If you really want to improve performance it makes more sense to create a script which fills up the database to thousands of users and millions of posts. This will make it much (much) easier to spot bottlenecks and you'll spot actual problems instead of indices which are missing only in theory.

Creating this database can be done with a fixtures script or by converting an existing large forum. Advantage of fixtures is that this is trivial to share with other developers and it's easier to adapt to the specs of your development environment. The main advantage of converting an existing forum is that this results in a very realistic database.

franzliedke:
[...]

PR #952, opened 2016-05-07
SQL Performance tuning

Ximich:
I found bad query:

select * from `flarum_discussions` where `id` in (select `discussion_id` from `flarum_posts` where `id` in (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)) and not exists (select 1 from `flarum_discussions_tags` where 0 = 1 and `flarum_discussions`.`id` = discussion_id)

One complex nested query take more time than 2 simple query. MySQL engine doesn't like nested queries.
So I reduce execution time on my VPS from ~3.1 seconds to 360 milliseconds just switching from one complex nested query to 2 small queries.

MySQL EXPLAIN for one big query:

MySQL> explain select * from `flarum_discussions` where `id` in (select `discussion_id` from `flarum_posts` where `id` in (1 ,2, 4, 5)) and not exists (select 1 from `flarum_discussions_tags` where 0 = 1 and `flarum_discussions`.`id` = discussion_id);
+----+--------------------+--------------------+-------+-------------------------------------------+---------+---------+------+-------+------------------+
| id | select_type        | table              | type  | possible_keys                             | key     | key_len | ref  | rows  | Extra            |
+----+--------------------+--------------------+-------+-------------------------------------------+---------+---------+------+-------+------------------+
|  1 | PRIMARY            | flarum_discussions | ALL   | NULL                                      | NULL    | NULL    | NULL | 39022 | Using where      |
|  3 | DEPENDENT SUBQUERY | NULL               | NULL  | NULL                                      | NULL    | NULL    | NULL |  NULL | Impossible WHERE |
|  2 | DEPENDENT SUBQUERY | flarum_posts       | range | PRIMARY,posts_discussion_id_number_unique | PRIMARY | 4       | NULL |     4 | Using where      |
+----+--------------------+--------------------+-------+-------------------------------------------+---------+---------+------+-------+------------------+

MySQL EXPLAIN for 2 small queries:

explain select `discussion_id` from `flarum_posts` where `id` in (1 ,2, 4, 5);
+----+-------------+--------------+-------+---------------+---------+---------+------+------+-------------+
| id | select_type | table        | type  | possible_keys | key     | key_len | ref  | rows | Extra       |
+----+-------------+--------------+-------+---------------+---------+---------+------+------+-------------+
|  1 | SIMPLE      | flarum_posts | range | PRIMARY       | PRIMARY | 4       | NULL |    4 | Using where |
+----+-------------+--------------+-------+---------------+---------+---------+------+------+-------------+

MySQL > explain select * from `flarum_discussions` where `id` in (2, 3, 4) and not exists (select 1 from `flarum_discussions_tags` where 0 = 1 and `flarum_discussions`.`id` = discussion_id);
+----+--------------------+--------------------+-------+---------------+---------+---------+------+------+------------------+
| id | select_type        | table              | type  | possible_keys | key     | key_len | ref  | rows | Extra            |
+----+--------------------+--------------------+-------+---------------+---------+---------+------+------+------------------+
|  1 | PRIMARY            | flarum_discussions | range | PRIMARY       | PRIMARY | 4       | NULL |    3 | Using where      |
|  2 | DEPENDENT SUBQUERY | NULL               | NULL  | NULL          | NULL    | NULL    | NULL | NULL | Impossible WHERE |
+----+--------------------+--------------------+-------+---------------+---------+---------+------+------+------------------+

### DIFF
--- a/src/Core/Repository/PostRepository.php
+++ b/src/Core/Repository/PostRepository.php
@@ -146,7 +146,7 @@ class PostRepository
 
     protected function getDiscussionsForPosts($postIds, User $actor)
     {
-	    return Discussion::query()
+        return Discussion::query()
             ->select('discussions.*')
             ->join('posts', 'posts.discussion_id', '=', 'discussions.id')
             ->whereIn('posts.id', $postIds)
--- a/src/Core/Repository/PostRepository.php
+++ b/src/Core/Repository/PostRepository.php
@@ -146,9 +146,9 @@ class PostRepository
 
     protected function getDiscussionsForPosts($postIds, User $actor)
     {
-        return Discussion::whereIn('id', function ($query) use ($postIds) {
-            $query->select('discussion_id')->from('posts')->whereIn('id', $postIds);
-        })
+        $posts = Post::whereIn('id', $postIds)->lists('discussion_id');
+
+        return Discussion::whereIn('id', $posts->toArray())
             ->whereVisibleTo($actor)
             ->get();
     }
--- a/src/Core/Repository/PostRepository.php
+++ b/src/Core/Repository/PostRepository.php
@@ -146,9 +146,11 @@ class PostRepository
 
     protected function getDiscussionsForPosts($postIds, User $actor)
     {
-        return Discussion::query()
-            ->whereIn('posts.id', $postIds)
+	    return Discussion::query()
+            ->select('discussions.*')
             ->join('posts', 'posts.discussion_id', '=', 'discussions.id')
+            ->whereIn('posts.id', $postIds)
+            ->groupBy('discussions.id')
             ->whereVisibleTo($actor)
             ->get();
     }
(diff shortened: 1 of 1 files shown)


Ximich:
#127

franzliedke:
Thanks for researching this!

Since we still need to add some indexes in a few places: have you researched how those would affect this particular query? After all, letting MySQL do it in one go *should* not be any slower than doing two queries...

Ximich:
This query does not require adding additional indexes, because table `posts` already have index on discussion_id column.
There are 2 ways to make this query execute efficiently (using indixes):
1. Divide it into two simple queries
2. Use the inner join instead of subquery

MySQL execution plan for subquery variant (original):


MySQL execution plan for inner join variant


As you see, on subquery version there are fullscan on table discussions (all 39022 records).
Version with inner join uses the indexes, so it is more efficient (**0.01** sec vs. **0.72** sec on my VPS):

Ximich:
Also I saw yet another few places with bad queries. So you want I can help you with them too.
Optimization of this particular query is important because this query is used on almost every page (gives most pain to the server).

franzliedke:
Thank you so much! We're happy about any further help with other problematic queries you come across. :)

PR #954, opened 2016-05-09
#446 Replace Post Edit Icon into Text

### LINKED ISSUE #446: Change edited icon into text

datitisev:
---
##### COMMITED: This pull replaces the pencil icon with the tooltip with just the text that was previously on the tooltip.
captura de pantalla 2016-05-08 a las 8 11 38 pm

---
##### SUGGESTION: We could probably also still use the pencil button + the text, like this:
captura de pantalla 2016-05-08 a las 8 13 44 pm

---

What do you guys think?

_PS: GitHub, this fixes #446 :wink:_ 
_I also figured out how to sign commits... had the wrong email :cry: :smiley:_

### DIFF
--- a/js/forum/dist/app.js
+++ b/js/forum/dist/app.js
@@ -24381,8 +24381,8 @@ System.register('flarum/components/PostEdited', ['flarum/Component', 'flarum/hel
 
             return m(
               'span',
-              { className: 'PostEdited', title: title },
-              icon('pencil')
+              { className: 'PostEdited' },
+              title
             );
           }
         }, {
--- a/js/forum/src/components/PostEdited.js
+++ b/js/forum/src/components/PostEdited.js
@@ -18,7 +18,7 @@ export default class PostEdited extends Component {
     const title = extractText(app.translator.trans('core.forum.post.edited_tooltip', {user: editUser, ago: humanTime(post.editTime())}));
 
     return (
-      <span className="PostEdited" title={title}>{icon('pencil')}</span>
+      <span className="PostEdited">{title}</span>
     );
   }
 


franzliedke:
:+1: Looks great.

datitisev:
I don't think the suggestion is that great, because users could still confuse it with the edit button.
Instead, we could use something like
> datitisev :pencil2: a minute ago

dav-is:
Maybe this information should go on the bottom? I just completed a PR for showing the IP address and I put it in the same spot. I think both belong somewhere else because it seems a bit too cluttered if both PRs went through.

datitisev:
@dav-is The bottom is mostly for the post actions, so we may not want to put much in there.
Probably the IP left-aligned would be fine, but not much else...

dav-is:
I'm not sure I really like this idea. It is from September of 2015. And I don't think it's been brought up since. I feel like there will be too much clutter.

datitisev:
Well, we would at least just make it for mobile, there is enough space.

Reflic:
I think we should use the way @datisev already committed. As he said, the icon makes it confusing for the user as it is not clear if it is an action or only a information.

franzliedke:
I'm starting to think we should definitely include the text, no matter how we decide to display this in the end.

And, more importantly, the actual styling (optional icon, show the text or hide it) should be done with CSS. This would mean we could hide the text in an accessibility-friendly way, so that users with screen readers would be told that the post has been edited, but we wouldn't necessarily need to clutter the interface with the text, or at least leave it up to the theme...

franzliedke:
Which means I can merge this. :)

Thanks @datitisev, you're on a roll.

PR #960, opened 2016-05-14
added an event to mutate extensions behavior

Luceos:
Idan Hazan requested a way of mutating the order of extensions. See https://discuss.flarum.org/d/2868-extensions-priority

While discussing this on gitter I thought about writing the code and sending in. Here it is ;)

### DIFF
--- a/src/Event/ExtensionsWereLoaded.php
+++ b/src/Event/ExtensionsWereLoaded.php
@@ -1,0 +1,29 @@
+<?php
+/*
+ * This file is part of Flarum.
+ *
+ * (c) Toby Zerner <toby.zerner@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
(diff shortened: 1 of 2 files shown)


hazanpro:
Thanks :)
Very useful for me!

tobscure:
Thanks for the PR, I think this solution is kind of overreaching though... Supposing that the initial problem is that there needs to be a way to prioritize one extension's event listener over another, yes this solution will change the order in which the event listeners are registered (and thus their priorities), but it will also change the order of the extensions in the extension list, or any other time a list of extensions is retrieved. I don't think that's behaviour we want to allow.

JoshyPHP's solution in that discussion is the recommended approach:

> Look up this API: https://laravel.com/api/5.1/Illuminate/Events/Dispatcher.html#method_listen

Luceos:
Thank you @tobscure just wanted this discussed ;)

PR #965, opened 2016-05-22
Fix Permission Name for Viewing Ips

dav-is:
Apparently permissions must end in 'Posts' if they are about a post. This is something I wasn't aware of when creating my previous PR to add the ability to view post Ips. This fixes the bug preventing mods from seeing IPs.


franzliedke:
Will we need a migration for this, @tobscure?

Probably not, eh, since none of the related changes were part of an official beta release...

PR #989, opened 2016-06-20
Fixes #825

### LINKED ISSUE #825: Support setting MySQL port in installation
https://discuss.flarum.org/d/2115-how-to-configure-different-mysql-port

We should parse the host field for a `:` and explode it if found

nielstholenaar:
See #825 for more information.

### DIFF
--- a/src/Install/Console/InstallCommand.php
+++ b/src/Install/Console/InstallCommand.php
@@ -133,7 +133,8 @@ class InstallCommand extends AbstractCommand
                     'host' => 'required',
                     'database' => 'required|string',
                     'username' => 'required|string',
-                    'prefix' => 'alpha_dash|max:10'
+                    'prefix' => 'alpha_dash|max:10',
+                    'port'   => 'required|integer|min:1|max:65535',
                 ]
             );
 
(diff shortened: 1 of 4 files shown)


dav-is:
Good work, but I thought the plan was to parse the address and if it had a `:*` then it would use that port instead of the default port.

franzliedke:
I'd suggest we keep that extra field and do the parsing only if no custom port was provided.
That way, non-technical users who don't really know what a port number means, don't have to extract it in case they get a hostname with port from their hosting provider.

@nielstholenaar Can you implement that?

src/Install/Console/InstallCommand.php
```diff
@@ -133,7 +133,8 @@ protected function install()
                     'host' => 'required',
                     'database' => 'required|string',
                     'username' => 'required|string',
-                    'prefix' => 'alpha_dash|max:10'
+                    'prefix' => 'alpha_dash|max:10',
+                    'port'   => 'required|integer|min:1|max:65535',
```
franzliedke: The port shouldn't be required. Just use the default (after trying to extract one from the hostname) if none was provided.

JoshyPHP: I don't know about Laravel's database driver specifically but in some other implementations the port needs to be omitted if a socket is used. That's another reason to make the port number optional.

franzliedke:
@nielstholenaar Will you have time to get back to this and incorporate the suggested changes? If not, we'll probably merge it and do it ourselves...

franzliedke:
Thank you! I have taken care of the rest.

PR #995, opened 2016-07-08
Fix UI for iOS while replying to a post

fernsheldon:
In reference to :
https://discuss.flarum.org/d/3347-no-cursor-while-replying-to-a-post-in-mobile-view-on-ios/9

 - Removed the triggerClick
 - Changed the DiscussionControls.replayAction.call method

### DIFF
--- a/js/forum/src/components/ReplyPlaceholder.js
+++ b/js/forum/src/components/ReplyPlaceholder.js
@@ -31,17 +31,15 @@ export default class ReplyPlaceholder extends Component {
       );
     }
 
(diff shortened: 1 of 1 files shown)


franzliedke:
Since this is for iOS, can you have a look, @tobscure ?

tobscure:
Thanks @fernsheldon! I ended up refactoring your solution slightly and keeping the replyAction method call unchanged, but the final result is the same :)

fernsheldon:
@tobscure Thanks :)

PR #997, opened 2016-07-10
Correction in search sensitive

terabin:
The query is executed, but no shown, this will correct this bug.

### DIFF
--- a/js/forum/dist/app.js
+++ b/js/forum/dist/app.js
@@ -2766,7 +2766,7 @@ exports.System = System;
 	}
 
 	if( m.attrs ) m.attrs.bidi = bidi
-	
+
 	m.bidi = bidi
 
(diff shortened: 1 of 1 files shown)


fernsheldon:
Hi @terabin , I was just looking the changes and like @franzliedke said, the changes are in the dist/app.js file.  

What you probably want to do is to find the component that the change is related to, and make the change there. Once done then compile it, so that the dist/app.js gets built with your new changes.   

In this case I think you'd need to make the changes in core/js/forum/src/components/Search.js

franzliedke:
@terabin It looks like you made the changes in the compiled file. I could fix that for you, however you modified a vendor script (one that we pull in via NPM and have not written ourselves). Can you explain (or link to an issue) which problem this PR fixes?

PR #1001, opened 2016-07-17
Fix #999

### LINKED ISSUE #999: Avatar Uploader does not remove files after validation failure
Reference thread
https://discuss.flarum.org/d/3381-i-have-found-avatar-upload-bug-for-upload-shell-backdoor

When a user uploads a file that fails image validation, the file is kept on the server in storage/tmp.
This may trigger antivirus warnings if malicious files are uploaded

The file should be deleted immediately if validation fails for uploads.

Currently occurring in dev-master and possibly beta.5

dav-is:
Simply `unlink()` unvalidated files from the `{{flarumroot}}/storage/tmp` dir

Should fix #999

### DIFF
--- a/src/Core/Validator/AvatarValidator.php
+++ b/src/Core/Validator/AvatarValidator.php
@@ -1,4 +1,5 @@
 <?php
+
 /*
  * This file is part of Flarum.
  *
--- a/src/Core/Validator/AvatarValidator.php
+++ b/src/Core/Validator/AvatarValidator.php
@@ -10,8 +10,6 @@
 
 namespace Flarum\Core\Validator;
 
-use League\Flysystem\FilesystemInterface;
-
 class AvatarValidator extends AbstractValidator
 {
     protected $rules = [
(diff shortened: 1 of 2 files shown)


src/Core/Validator/AvatarValidator.php
```diff
@@ -19,4 +20,14 @@ class AvatarValidator extends AbstractValidator
             'max:2048'
         ]
     ];
+
+    public function assertValid(array $attributes)
+    {
+        $validator = $this->makeValidator($attributes);
+
+        if ($validator->fails()) {
+            unlink($attributes['avatar']->avatarPath);
```
franzliedke: You should be able to use `$attributes['avatar']->getRealPath()` here: http://php.net/manual/en/splfileinfo.getrealpath.php

franzliedke:
Thanks for the PR! Just a small change, then I will merge. :)

tobscure:
:+1: including @franzliedke's suggestion :)

dav-is:
Closed along with #999 and https://github.com/flarum/core/commit/d5797dae794df86ad2e98560e3447c0cca6bf5ea

Issue #623, opened 2015-11-04
Check whether langpack is the default language before disabling


dcsjapan:
At present it is possible to disable a language pack that the admin has selected as the forum's default language. This can result in the translator attempting to fall back to resources that are no longer valid.

Flarum needs to do the following whenever an admin disables an extension:
1. Check whether the extension is a language pack.
2. If yes: check whether that language pack has been selected as the forum's default language.
3. If yes: fail to disable the extension and warn the user to change the default language first.

Or it could present the user with a list of installed language packs _other than the one the user is trying to disable_, allowing the user to selected a new default language right there ... upon which the extension can be disabled.

renyuneyun:
Hi,

I'm new here (also new to php) and have some interest in solving this issue.

Are there any common fields shared by all language packs so that it can be used to identify an extension's type?

I found there a common one in composer.json ('locale' in keywords), but not sure if it could be used or should be used.

Kakifrucht:
The default locale is identified in the settings table of flarum under the key default_locale

renyuneyun:
@kirkbushell 
Thanks for the reply.

So I can get the default one from the database and check if it's the same as the extension's name?
How to get the value of the key default_locale? Hopefully it's not done by writing a sql statement myself...

p.s. Wouldn't this be a kind of waste since only language packs/extentions should be checked?

Kakifrucht:
I guess you would have to check the composer.json, I'm not sure though.

dcsjapan:
@renyuneyun It should be possible to get a short list of just locale extensions, since the admin UI will be doing something similar to [group extensions by type](https://github.com/flarum/core/issues/746). I'm guessing @tobscure plans to add some sort of "extension type" data to composer.json to facilitate that grouping; if so, that would be the thing to use.

franzliedke:
Hi @renyuneyun, thanks for your interest in solving this issue! And sorry it took a while to get back to you with more information... you know, life. ;)

Anyhow, yes, you are right, the "locale" keyword in composer.json is currently the only way to recognize an extension as language pack. Please note that the `ExtensionManager` already does some work parsing all the extension information from the JSON file, so that is not necessary anymore.

Do you need any other pointers to get started?

renyuneyun:
Hi @franzliedke, thanks for your kindness!

I've found out how to get the needed properties and (I think I've) finished the codes to determine whether the extension is the target or not.

[...]

PR #1004, opened 2016-07-20
WIP: solve #623

renyuneyun:
Prevent from disabling default locale

Trying to solve #623

### DIFF
--- a/src/Extension/ExtensionManager.php
+++ b/src/Extension/ExtensionManager.php
@@ -1,4 +1,5 @@
 <?php
+
 /*
  * This file is part of Flarum.
  *
@@ -138,9 +139,11 @@ class ExtensionManager
         $enabled = $this->getEnabled();
 
         if (($k = array_search($name, $enabled)) !== false) {
-            unset($enabled[$k]);
-
             $extension = $this->getExtension($name);
+
+            $this->dispatcher->fire(new ExtensionWillBeDisabled($extension));
+
+            unset($enabled[$k]);
 
             $this->setEnabled($enabled);
 
--- a/src/Core/CoreServiceProvider.php
+++ b/src/Core/CoreServiceProvider.php
@@ -1,4 +1,5 @@
 <?php
+
 /*
  * This file is part of Flarum.
  *
@@ -107,6 +108,7 @@ class CoreServiceProvider extends AbstractServiceProvider
         $events->subscribe('Flarum\Core\Listener\UserMetadataUpdater');
         $events->subscribe('Flarum\Core\Listener\EmailConfirmationMailer');
         $events->subscribe('Flarum\Core\Listener\DiscussionRenamedNotifier');
+        $events->subscribe('Flarum\Core\Listener\ExtensionChangeValidator');
 
         $events->subscribe('Flarum\Core\Access\DiscussionPolicy');
         $events->subscribe('Flarum\Core\Access\GroupPolicy');
--- a/src/Core/Listener/ExtensionChangeValidator.php
+++ b/src/Core/Listener/ExtensionChangeValidator.php
@@ -1,0 +1,56 @@
+<?php
+
+/*
+ * This file is part of Flarum.
+ *
+ * (c) Toby Zerner <toby.zerner@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Flarum\Core\Listener;
+
+use Flarum\Event\ExtensionWillBeDisabled;
+use Flarum\Extension\Extension;
+use Illuminate\Contracts\Events\Dispatcher;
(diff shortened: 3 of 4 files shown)


datitisev:
Please make StyleCI happy and add that extra line of code at the beginning of the file.
Thanks for the pull request, at least the community is working on Flarum ;)

renyuneyun:
Hi @datitisev,

It's strange that the styleci check didn't pass (yes, I was aware of this just after creating the pull request), since I didn't touch anything at the beginning of the file (see the diff) :|

If this should be 'solved', the same thing should be done on many other files since they all have the same format/style...

Seems to be curious :)

dav-is:
I can confirm style cl did this to my PR too

datitisev:
Hi, @renyuneyun . StyleCI just makes your coding style better, cleans up unused imports, etc..
It says you are missing a space on some file between the start (`<?php`) and the header copyright comment. Don't worry much about it.

dav-is:
@datitisev Look through the current flarum files. None of them have that space. This has to be a new style cl rule or something.

datitisev:
@dav-is Huh, weird.

renyuneyun:
Hey everyone,

There is one more thing that I can't make for sure. That is: how to broadcast the event/situation of what is happening. As for this one, it's telling other components "failed to disable the language pack" (and maybe also the reason).

I found that there is an _Event_ mechanism, and I also found some codes using exceptions.

Anyone can give me some guides? What are the usages of these two mechanisms in the software? And, which one should be chosen here?

Luceos:
Hi @renyuneyun 

Our apologies for the late reply. We should have answered you sooner than this.

You are correct in thinking about events. A better implementation of this would be using an event. There is an existing event you can use [ExtensionWasDisabled](https://github.com/flarum/core/blob/master/src/Event/ExtensionWasDisabled.php). If you are able to create a Listener for that event and have it do a validation whether this is not the last locale in the extensions manager, it would probably suffice.

franzliedke:
Hi @renyuneyun, can you give us a quick update on this? Did @Luceos' comment help you? If not, what do you need help with?

src/Extension/ExtensionManager.php
```diff
@@ -138,10 +138,18 @@ public function disable($name)
         $enabled = $this->getEnabled();
 
         if (($k = array_search($name, $enabled)) !== false) {
-            unset($enabled[$k]);
-
             $extension = $this->getExtension($name);
 
+            if (in_array('locale', $extension->keywords)) {
```
dav-is: I'd say it's a better idea to use: `if (in_array('flarum-locale', $extension->extra)) {`

src/Extension/ExtensionManager.php
```diff
@@ -138,10 +138,18 @@ public function disable($name)
...
+                $default_locale = $this->app->make('flarum.settings')->get('default_locale');
+                $locale_code = array_get($extension->extra, 'flarum-locale.code');
+                if (strcmp($default_locale, $locale_code) == 0) {
```
dav-is: `if ($default_locale === $locale_code) {`

Would this be a cleaner option?

renyuneyun:
Sorry... Really sorry that I didn't update anything here...
I've just been to a new university abroad and was trying to adapt myself to that, so I didn't check my github notifications for three weeks... And unfortunately, I mis-configured my email client (thunderbird) and it failed to receive emails from my gmail account for a long time... Until today had I realized that and fixed it...

Before I took the flight (say, went abroad), I finished the ExtensionWillBeDisabled event, the validator (ExtensionChangeValidator).
But I'm not sure how to wait for the event... I mean, is the dispatcher->fire() action asynchronous or synchronous? (And I left it there, thought I could find it out later... That was until now...)

src/Extension/ExtensionManager.php
renyuneyun: But that won't be a good extension (language pack), will it?
Maybe a better way is to check 'locale' in keywods first, and then look for 'flarum-locale' if 'locale' was not found. Since this action will happed not that often, it won't take up resources.

src/Extension/ExtensionManager.php
renyuneyun: Definitely!

I'm a newbie to php, and thought == would compare the pointer/reference so I looked for the strcmp() function.
Thank you for pointing out things like '==='! That's news to me

renyuneyun:
Aha. I just looked at the 'pull requests' list, and found that dav-is has already done that.
Maybe accepting his (or her?) pull request will be a simpler solution?

Apologize again here... I should have left some words which say that I will not be able to work on this those days...


But I'm still not clear why dispatcher->fire() is synchronous... It doesn't fit its name

Luceos:
Going to be resolved by https://github.com/flarum/core/pull/1032

Issue #990, opened 2016-06-20
API: Get all posts by tag?


coderstephen:
Is it possible to get all posts by what their parent discussion is tagged with through the API? If not, would you consider adding this capability or some other efficient means of getting this data? Currently I have to get all discussions with a tag, then request all posts for each discussion, which is really inefficient.

My intent is to get all posts with a given discussion tag, ordered by date.

franzliedke:
I don't think that's currently possible.

Can you explain a bit more what your use-case is? I can't come up with a useful example of when I would want all posts for a tag, regardless of which discussion they belong to...

coderstephen:
@franzliedke I'm currently working on an NNTP proxy server for Flarum. (Raw NNTP part is [here](https://github.com/coderstephen/libnntp), WIP. This part of an effort to make Flarum work for the PHP internals team to replace the mailing list). It is coming along nicely, but the issue with NNTP is that it has no concept of "threaded conversation" other than replies, very much like email. Ideally, I could present all posts in a tag to the client and calculate the "replies" myself by having the first post in a discussion start a thread of messages.

The issue is that this process is currently very slow since I make a request for every discussion in the desired tag + requests for the actual posts under each. Most clients (notably Mozilla Thunderbird) has a 30-second timeout, and generating the post list takes longer than that. I'm sure a single SQL query with joins could produce the results I'm looking for, but I'd rather not require direct DB access and use the API instead.

sijad:
@coderstephen maybe 2018e424ec92e974c8fa9ee4d997c86e614e7126 can help, you need make a new extension though.

franzliedke:
@coderstephen I think we can do it. Would you be interested in sending a PR? Or would you rather get help in getting started with an extension?

coderstephen:
@franzliedke I think I can open a PR soon to add the filter option. I haven't looked at it in a couple of weeks, but this is might be the ideal solution I think for my use case.

PR #1005, opened 2016-07-21
Add posts API filter for tags

coderstephen:
Adds another filter, `filter[tag]`, for the `/posts` API endpoint which filters posts based on the tag of the posts' parent discussion.

Addresses #990.


tobscure:
Looks good, but as a general rule Flarum's core shouldn't contain any references to extensions. So this logic should be moved to be part of flarum-ext-tags (via the ConfigurePostsQuery event). Let us know if you need a hand!

PR #1007, opened 2016-07-24
#1002 Handle word wrapping properly in posts

### LINKED ISSUE #1002: Break words in posts with excessive width
Words longer than the post width can complicate scrolling in certain environments, e.g. when using notebook touchpads or mobile devices. It might be a good idea to adjust the `word-wrap` property for the post body. See [this discussion](https://discuss.flarum.org/d/3380-posts-with-horizontal-scroll) for details.

datitisev:
This PR fixes #1002.

- Added `overflow-wrap: break-word` to .Post-body
- Added `overflow-wrap: normal` to pre in .Post-body for code blocks so there is no wrap (easier to identify different lines, unless we want to add that enter icon to indicate it's the same line)

Screenshot (if needed):

captura de pantalla 2016-07-24 a las 12 18 40 pm

### DIFF
--- a/less/forum/Post.less
+++ b/less/forum/Post.less
@@ -92,6 +92,7 @@
   line-height: 1.7;
   position: relative;
   overflow: auto;
+  overflow-wrap: break-word;
 
   p, ul, ol, blockquote {
     margin-bottom: 1em;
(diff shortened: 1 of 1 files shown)


tobscure:
Thanks @datitisev!

PR #1009, opened 2016-07-26
#200 Refresh doesn't update notifications

### LINKED ISSUE #200: Refresh doesn't update notifcations
**_1 Upvote**_ Although the refresh implementation is a temporary solution until the Pusher extension is developed, I think it would make sense for the refresh functionality to also update any possible notifications you may have received. Currently, it only updates the discussion listing.

datitisev:
Refreshing discussion list (without Pusher) refreshes notifications as well. This fixes #200.
- When clicking "refresh" button for discussion list (on homepage) refresh notifications
- When clicking forum title (on homepage) refresh notifications

**TODO:** Have some way of updating the notifications when this events don't occur. Ideas:
1. The user is AFK with a page open, comes back, and the notification count is refreshed because it refreshes every _n_ minutes
2. Refresh notifications when clicking a discussion, a tag, clicking reply, like... (probably too much, though, maybe have a random chance of updating?)

_(I know this was marked for v1.0, but I think it should be ready ASAP)_

### DIFF
--- a/js/forum/dist/app.js
+++ b/js/forum/dist/app.js
@@ -18237,8 +18237,6 @@ $('#el').spin('flower', 'red');
 'use strict';
 
 System.register('flarum/app', ['flarum/ForumApp', 'flarum/initializers/store', 'flarum/initializers/preload', 'flarum/initializers/routes', 'flarum/initializers/components', 'flarum/initializers/humanTime', 'flarum/initializers/boot', 'flarum/initializers/alertEmailConfirmation'], function (_export, _context) {
-  "use strict";
-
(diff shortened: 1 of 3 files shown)


terabin:
`app.store.find('users', app.session.user.id());`
This will give a error if the user is not logged, you need to add a check.

datitisev:
@terabin Thanks for the comment. I commited the fix.

Luceos:
Someone willing to check this out and test it.. Approve it for merging? Hence the verification tag.

franzliedke:
Thanks a lot - I merged this manually now.

PR #1011, opened 2016-07-29
Fixed fontawesome relative path

zcodes:
If install flarum in the site's root directory, it seems nothing
is wrong because the server software will trim the path, but if
install flarum in a sub directory, the font file of fontawesome
will not load correctly.

### DIFF
--- a/less/lib/lib.less
+++ b/less/lib/lib.less
@@ -1,5 +1,5 @@
 @import "font-awesome.less";
-@fa-font-path: "../assets/fonts";
+@fa-font-path: "./fonts";
 
 @import "normalize.less";
 @import "print.less";
--- a/less/lib/lib.less
+++ b/less/lib/lib.less
@@ -1,5 +1,5 @@
 @import "font-awesome.less";
-@fa-font-path: "../../assets/fonts";
+@fa-font-path: "../assets/fonts";
 
 @import "normalize.less";
 @import "print.less";


franzliedke:
But wouldn't that mean it won't work anymore when installed in the root directory?

zcodes:
It works when installed in the root directory as it is the correct path, I mean when installed in the root directory, it is not easy to find that "../../assets/" is  incorrect, because "../assets", "../../assets/", even "../../../assets" is same as it will be treated as “/assets/" by the server software.

franzliedke:
Ah okay, that makes sense. Thanks for the PR, will merge it soon when I fully understand what is going on. I need some rest. ;)

tobscure:
Not sure I understand... the assets directory is located two levels up from less/lib/lib.less. `../assets/fonts` would refer to `less/assets/fonts`, which doesn't exist. Am I missing something?

zcodes:
this path is for the dist files, not for the source. I find that the less files are compiled into css by `src/Asset/LessCompiler.php`, and from `flarum/flarum`, the asset files are store in `assets` directory with this hierarchy:

```
├── admin-b3681435.js
├── admin-en-df77f08d.js
├── avatars
│   └── ulnctd46jk3qi9ni.jpg
├── fonts
│   ├── FontAwesome.otf
│   ├── fontawesome-webfont.eot
│   ├── fontawesome-webfont.svg
│   ├── fontawesome-webfont.ttf
│   ├── fontawesome-webfont.woff
│   └── fontawesome-webfont.woff2
├── forum-28507fef.js
├── forum-9163da26.css
├── forum-en-df77f08d.js
└── rev-manifest.json
``` 
and in the compiled css file, it will be like 

```
@font-face {
    font-family: 'FontAwesome';
    src: url('../assets/fonts/fontawesome-webfont.eot?v=4.6.3');
    src: url('../assets/fonts/fontawesome-webfont.eot?#iefix&v=4.6.3') format('embedded-opentype'), url('../assets/fonts/fontawesome-webfont.woff2?v=4.6.3') format('woff2'), url('../assets/fonts/fontawesome-webfont.woff?v=4.6.3') format('woff'), url('../assets/fonts/fontawesome-webfont.ttf?v=4.6.3') format('truetype'), url('../assets/fonts/fontawesome-webfont.svg?v=4.6.3#fontawesomeregular') format('svg');
    font-weight: normal;
    font-style: normal
}
```

so `../assets/fonts/` or `./fonts/` should the correct path.

tobscure:
Ah, right you are! In that case, shouldn't the path just be "fonts"?

zcodes:
Ah, "fonts" should be OK. And i changed it in this PR.

tobscure:
Thanks!

Issue #1025, opened 2016-09-03
post.discussion() is not a function


terabin:
I don't know how this error appears, just a strange thing happened.

A member anwsered my post, and notification start to give error:

post.discussion() is not a function
in:

```
/**
 * Generate a URL to a post.
 *
 * @param {Post} post
 * @return {String}
 */
app.route.post = function (post) {
  return app.route.discussion(post.discussion(), post.number());
};
```

Value of two metions notification in this time is false, so, maybe is not correctly loaded?

When I refresh the page, the first time when I open notifications, works, but when I change my page(for example, for another topic), the error appears again.

When another member anwsered the same post, the error is gone.

felagund18:
I found this error too, mine was related to notifications
![screen shot 2016-09-04 at 2 21 30 pm](https://cloud.githubusercontent.com/assets/1384327/18229211/f80acf82-72aa-11e6-919a-c83a20499833.png)
this error was related to ```subject_id```,
after notification is saved in ```notifications```, user deleted it's post. In client side forum.js, while showing notifications, did not check if related model is exists or not... this was causing the error

terabin:
Well, I found the cause too, but for me, makes sense delete notification about a post when the post is deleted, so, I propose this:

https://github.com/flarum/flarum-ext-likes/pull/7

Luceos:
Link to new PR https://github.com/flarum/core/pull/1027

franzliedke:
From [Toby's comment on one of the first PRs]():

> The notification subject type checking should be done using the Blueprint for the notification type... it might be a little complex and requires some knowledge of how the (undocumented) notification system works, so might be easier to just leave for now...

franzliedke:
The frontend issue has been fixed. Toby will create an issue for the remaining backend cleanup.

PR #1027, opened 2016-09-04
Delete Notifications When a post is deleted

terabin:


tobscure:
Thanks for the amended PR. Still not quite right as notifications can have subjects that are not posts, so deleting all by ID like this won't work. But getting there. The solution is slightly more complex, I will have a think about the best way to do it.

terabin:
So, maybe the best way to do this is in each extension?

tobscure:
Nope the best way is still in core, but will probably involve looping through each registered notification type and detecting whether or not the subject is a post. I'll work on it when I have some time.

terabin:
It's just PostLiked, PostMentioned and userMentioned?

terabin:
But each notification type have a different text, so, it's not possible get info about post type in each extension, right? Just curious.

terabin:
I improved the post code, and worked for now, can you confirm? Just thinking about discussion.

renyuneyun:
I guess coding style (StyleCI check) is the reason why this isn't merged (?)

franzliedke:
No, Toby will need to review the logic. I can make some comments on the PHP code really quick...

src/Core/Post.php
```diff
@@ -87,6 +88,21 @@ public static function boot()
...
+            //Delete notifications about this post
+            $notifs = Notification::where('subject_id', $post->id)->orWhere('subject_id', $post->discussion_id)->get();
+
+            foreach ($notifs as $notif)
+            {
+                if (!$notif->subject) {
+                    $notif->delete();
+                }
+                else {
+                    if ($notif->type == $post->type) {
```
franzliedke: You can "collapse" this into one `else if` statement.

src/Core/Post.php
```diff
@@ -87,6 +88,21 @@ public static function boot()
...
+
+            //Delete notifications about this post
+            $notifs = Notification::where('subject_id', $post->id)->orWhere('subject_id', $post->discussion_id)->get();
+
+            foreach ($notifs as $notif)
+            {
+                if (!$notif->subject) {
+                    $notif->delete();
+                }
+                else {
```
franzliedke: ...and this can be on one line: `} else if (...) {`.

Luceos: @franzliedke meant `} else {`

tobscure:
The notification subject type checking should be done using the Blueprint for the notification type... it might be a little complex and requires some knowledge of how the (undocumented) notification system works, so might be easier to just leave for now... do we have an issue for this?

franzliedke:
We have issue #1025, I will reference your comment there. :)

tobscure:
In any case, thanks for your efforts on this one @terabin. Sorry we didn't adopt your solution — as I said above, this problem is more complex than it seems — but we do appreciate the PR!

PR #1033, opened 2016-09-13
Prevent Overwriting of User's Attributes on Register (#897)

dav-is:
I read through the entire SSO process and this is the only security issue I could even remotely find regarding to flarum/issue-archive#332. This will prevent any of the user's attributes to be overwritten. For example, if a SSO provider neglected to verify emails.

### DIFF
--- a/src/Core/Command/RegisterUserHandler.php
+++ b/src/Core/Command/RegisterUserHandler.php
@@ -1,4 +1,5 @@
 <?php
+
 /*
  * This file is part of Flarum.
  *
--- a/src/Core/Command/RegisterUserHandler.php
+++ b/src/Core/Command/RegisterUserHandler.php
@@ -116,7 +116,7 @@ class RegisterUserHandler
         // from the get-go.
         if (isset($token)) {
             foreach ($token->payload as $k => $v) {
-                if ($user->$k == "" || ! isset($user->$k)) {
+                if ($user->$k == '' || ! isset($user->$k)) {
                     $user->$k = $v;
                 }
             }
--- a/src/Core/Command/RegisterUserHandler.php
+++ b/src/Core/Command/RegisterUserHandler.php
@@ -116,7 +116,7 @@ class RegisterUserHandler
         // from the get-go.
         if (isset($token)) {
             foreach ($token->payload as $k => $v) {
-                if ($user->$k == "" || !isset($user->$k)) {
+                if ($user->$k == "" || ! isset($user->$k)) {
                     $user->$k = $v;
                 }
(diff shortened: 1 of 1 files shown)


src/Core/Command/RegisterUserHandler.php
```diff
@@ -115,7 +116,9 @@ public function handle(RegisterUser $command)
         // from the get-go.
         if (isset($token)) {
             foreach ($token->payload as $k => $v) {
-                $user->$k = $v;
+                if ($user->$k == '' || ! isset($user->$k)) {
```
franzliedke: Isn't this the same as `if (empty($user->$k)`?

dav-is: Well empty would return true for 0 or false

franzliedke: So does `== ''`, see https://3v4l.org/Z6cl0.

Please use `===` then. :)

tobscure:
Can you please give a more concrete example of exactly how this could be exploited to screw something up? I agree that there's probably an issue here, but I'm having trouble imagining what that might be, and the complexity of the `if ($user->$k === '' || ! isset($user->$k))` condition is setting off alarm bells for me.

dav-is:
A user's attribute for email_when _mentioned (just making this up) is set to `0`. This setting is set and has a value of false. Therefore, it should not be tampered with. `if (empty($user->$k))` would return true because the value is false, but we still don't want to allow our SSO to modify the set variable.

luceos (changes requested):
the conditional can be shortened using an empty() function

dav-is:
@luceos Did you read my rational behind not using empty()?

luceos:
I skipped your last comment, my apologies. Even more impressed 😄

luceos (approved):

Luceos:
@tobscure or @franzliedke either merge this or assign to a milestone please. After having done that please unassign both of you.

franzliedke:
Thanks for the PR! And your patience. ;)

franzliedke:
I [slightly tweaked the condition](https://github.com/flarum/core/commit/7c1b0bfcf2fccc361cc984375dcc6a6430216b3a) to a statement that signals the intent slightly better - at least in my opinion.

And damn, should have squashed this ;)

franzliedke:
@dav-is Oh, last thing: Do you think we need to do anything about unknown attributes, i.e. a provider trying to set attributes that we don't even support? (Although I think that should just be a no-op, i.e. not be written to the database anyway).

Issue #1006, opened 2016-07-21
User Page Discussion Count differs from discussions user's able to see


datitisev:
> I didn't really know how to name this, sorry.

## PROBLEM

Whenever someone deletes the first post of a discussion, another user can still see that discussion as part of the discussion count on the user's page, but whenever they go to see the actual list of discussions, it says there're no discussions there.
![flarum/core#1006](https://i.imgsafe.org/140b07a051.png)

## STEPS

1. Create a discussion on a user's account with no discussions
2. Delete the first post in whatever account
3. Log out our login into any other account which doesn't have permissions to see deleted posts and/or discussions
4. You will see it says 1 discussion but shows none

tobscure:
Great bug report, thanks :)

muyiwaolu:
What's the status on this, thinking of taking this on as my first PR.

dav-is:
It's important the user's discussion and post count reflect the number of posts the user has made. If a post is hidden from the user viewing the post, the user should see the same number and should see a template post listing all the discussions they don't have permission to see. The API should be able to tell users the post was restricted instead of not found.

franzliedke:
@muyiwaolu Still interested? Feel free to take this on!

dav-is:
It think it's important we preserve the user statistics. I think the number should stay but I think the API should output empty placeholder posts or discussions when they don't have permission.

jordanjay29:
@dav-is has a fair point for user statistics, but I'm not sure the board-wide statistics actually need to be considered here. The number of discussions listed on a user profile could simply be a total count grabbed from the list to be displayed depending on permissions, not necessarily the total number contained in the DB. This maintains the obscurity of secure forums that you may not have permission to access (and read posts within), while not confusing members by having problems (seemingly) with simple arithmetic. 

The display on the user profile does not necessarily need to be the user's post count. It can instead be considered a total number of posts listed on the profile (which can be permissions-curated).

Any widgets or site statistics could pull from the entire post count as per normal.

dav-is:
Except loading all the posts from a certain user and testing to see if they have permissions takes a long time. That's why the post count is controlled by an event listener. @jordanjay

franzliedke:
@tobscure What are your general thoughts on what the UI should look like here?

tobscure:
I'm leaning towards the idea that the post count should only include posts that are visible to the public (or whatever the baseline visibility is for the forum).

[...]

PR #1035, opened 2016-09-14
Inform users that some posts/discussions are withheld from the profile (#1006)

dav-is:
This is the best fix I could think of for flarum/issue-archive#321. Here are some screenshots:

discussions

discussions2

posts

Here is the PR for the translations: https://github.com/flarum/flarum-ext-english/pull/84

### DIFF
--- a/js/forum/src/components/DiscussionsUserPage.js
+++ b/js/forum/src/components/DiscussionsUserPage.js
@@ -1,5 +1,5 @@
 import UserPage from 'flarum/components/UserPage';
-import DiscussionList from 'flarum/components/DiscussionList';
+import UserDiscussionList from 'flarum/components/UserDiscussionList';
 
 /**
  * The `DiscussionsUserPage` component shows a discussion list inside of a user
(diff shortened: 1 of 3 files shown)


franzliedke:
How do you like this UI, @tobscure?

franzliedke:
Thanks for taking the time, @dav-is !

dav-is:
It's my pleasure, @franzliedke

tobyzerner:
Hmm I'm not convinced this is the most elegant solution to this problem. I know it's trying to explain to the user why the number may be wrong, but in doing so it kind of draws too much attention. You shouldn't know anything about posts that you can't see - even the fact that they exist, or who has been making them. But this kind of throws that in your face. It's a security issue.

Going to close this, let's go back to the drawing board in flarum/issue-archive#321. Thanks for the effort nonetheless @dav-is. As I mentioned elsewhere, reopening a dialogue in the issue to talk about the approach before sending a PR might be a good idea :)

dav-is:
@tobscure I don't think it's a security issue. It tells you basically nothing about the missing posts, the number is generated simply by subtracting the number of posts displayed vs the count. Those posts could be deleted, hidden, or it simply be an error. I still believe the number should represent the actual number of posts and not the number the user can see. So I can't imagine a more elequent solution...

PR #1067, opened 2016-11-15
Add AuthorIdGambit to support search discussion by user id

liji:
Add a new  AuthorIdGambit to support search discussion by user id

eg:

/api/discussions?filter[q]=authorid:1
/api/discussions?filter[q]=authorid:1,2,3


tobscure:
We can probably just re-purpose the existing `author:` gambit for this. If a username is an integer, treat it as an ID.

dav-is:
Can a user have a username that is an integer?

tobscure:
I think at the moment it's possible but really it shouldn't be allowed as our JSON API allows use of a username or an ID too (`/api/users/123`, `/api/users/tobscure`)

liji:
Yes, re-purpose the existing Author gambit is fine, like the user API dose.  But I am wondering that why we use the same API for different search criteria, especially the username can be an integer currently, the API may not works as expected. e.g.  an user registered with username 100 ,` /api/users/100` will only retrieve the user with ID 100, this does not look correct If someone's intention is to get the user with username 100.

franzliedke:
@liji Can you change this PR so that the existing gambit reacts differently when it receives a numeric ID?

After that, we will need to implement (or at least make a ticket to do so) a new validation rule for usernames so that they cannot be numbers-only.

liji:
@franzliedke Sure, I will change this PR accordingly.

liji:
@franzliedke this PR has been changed.

src/Core/Search/Discussion/Gambit/AuthorGambit.php
```diff
@@ -50,7 +51,7 @@ protected function conditions(AbstractSearch $search, array $matches, $negate)
 
         $ids = [];
         foreach ($usernames as $username) {
-            $ids[] = $this->users->getIdForUsername($username);
+            $ids[] = is_numeric($username) ? $username : $this->users->getIdForUsername($username);
```
franzliedke: This should at least be `intval($username)`.

liji: Make sense. how about filter_var($username, FILTER_VALIDATE_INT)?

Luceos: Both solutions are very flaky. Mostly because we tolerate digit usernames as well, doing a intval or filter_var will only assume a username to be a user ID. Wouldn't it make more sense to stick to usernames only for now?

datitisev: I don't think I'd look for a user's posts by their ID. If I wanted to know their username, I'd visit their profile page, `/u/ID`. But some people would prefer to have to do just one query, and not visit another page.
Talking about the profile page, as it allows user IDs, wouldn't it make sense to maybe use that method as well for the gambit? I haven't seen anyone bring this up, and I haven't been able to find where in Flarum's code this happens.

dav-is: I'd say change the user validatior in this PR as well that confirms the username is a string. Maybe add a migration that searches for numeric usernames and adds a "." to the end?

dav-is: >I haven't seen anyone bring this up, and I haven't been able to find where in Flarum's code this happens.

Check the discussion in this PR.

Luceos: Taking a look at this PR again it's about the authorId not a user id, let's keep it like that and stick to using an ID integer, not a username.

liji: This PR is applicable only if we decide that username can not be digits only. As I mentioned above, like the current user API, it won't work properly in certain cases.

Luceos: I see, yes then the user validation has to be improved too. This will be very problematic in relation to backward compatibility. We might need to create a migration that renames digit only usernames too.

liji: OK, I agree. Then we can leave this PR for now until we decide to implement the new validation and figure out the migration strategy.

franzliedke:
Closing as stale.

PR #1072, opened 2016-11-21
Bithound Integration Support

dav-is:
Added files for support with https://www.bithound.io

ESLint configuration still needs discussion


matteocontrini:
BitHound doesn't exist anymore, this could be closed

PR #1073, opened 2016-11-23
Add group gambit to support search user by group name

liji:
Add group gambit to support search user by group name

- `/api/users?filter[q]=group:admin`

- `/api/users?filter[q]=group:admin,mod`

refer to flarum/issue-archive#418

### DIFF
--- a/src/Core/Repository/GroupRepository.php
+++ b/src/Core/Repository/GroupRepository.php
@@ -1,4 +1,5 @@
 <?php
+
 /*
  * This file is part of Flarum.
  *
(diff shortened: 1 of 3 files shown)


tobyzerner (approved):
Looks good!

I think there are possibly a few tweaks we could make to improve performance (a subquery perhaps?) but that is also the case for a lot of other gambits + the fulltext search driver, so maybe we'll do all that optimization all together at a later date.

dav-is:
Laughed a bit at the typical styleci struggle

liji:
Yes, I agree, there are things we can do to improve the performance.

Luceos:
@tobscure you might consider putting styleci to auto patch any merge. That resolves a lot of issues contributors might have. The time they now spend on fixing styleci compliance can be used to send in another PR :smile:

franzliedke:
@liji Thank you very much for the implementation and your patience! 👍

Issue #1057, opened 2016-10-20
Meta description not show


wiwatsrt:
Meta description not show
## Bug report
- Version of Flarum v0.1.0-beta.6
## Additional comments

I set description in admin but not show in page
## Code

```
<meta name="description" content="">
```

Jurgyte:
Same here. It seems that in Flarum "flarum/core/src/Http/WebApp/WebAppView.php" is small section:

        $view->title = $this->buildTitle(array_get($forum, 'data.attributes.title'));
        $view->description = $this->description;

Description is not set by anything. Could someone check?

daggeto:
👍  and how to customize that description will be same as first post in page?

Luceos:
There is no setter method for description. This is confirmed, fix should take place somewhere in the mentioned `WebAppView` class:

https://github.com/flarum/core/blob/master/src/Http/WebApp/WebAppView.php#L286

franzliedke:
I can't test this right now, but I believe changing that line to the following should fix the issue:

~~~php
$view->description = array_get($forum, 'data.attributes.description');
~~~

(The `$description` attribute can then also be removed.)

franzliedke:
Pretty sure this was fixed in #1105.

PR #1088, opened 2016-12-15
Set discussion page meta description same as first post

daggeto:
As description is not set in discussion page, in our project, we decide to put first post as page description for better SEO. What do you think about it ?

### DIFF
--- a/src/Forum/Controller/DiscussionController.php
+++ b/src/Forum/Controller/DiscussionController.php
@@ -85,6 +85,12 @@ class DiscussionController extends WebAppController
             }
         }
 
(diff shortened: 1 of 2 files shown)


franzliedke:
@tobscure What do you think about this feature?

In any case, I would like to see #1057 fixed before reviewing / merging this PR.

franzliedke:
#1057 is already fixed as per #1105.

As for the SEO feature, this is still buggy: when you visit a second page, the description of the first post of the second page. Is that intended?

Closing until the behavior we want here is more clearly defined...

PR #1092, opened 2016-12-21
Add cookie helper

### LINKED ISSUE #1084: Cookies must be set with the Secure flag in HTTPS mode

sijad:
Fixes #1084

### DIFF
--- a/src/Http/CookieFactory.php
+++ b/src/Http/CookieFactory.php
@@ -1,0 +1,63 @@
+<?php
+
+/*
+ * This file is part of Flarum.
+ *
+ * (c) Toby Zerner <toby.zerner@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Flarum\Http;
+
+use Dflydev\FigCookies\SetCookie;
+use Flarum\Foundation\Application;
+
+class CookieFactory
+{
+    /**
+     * @var Application
+     */
+    protected $app;
+
+    /**
+     * @param Application $app
+     */
+    public function __construct(Application $app)
+    {
+        $this->app = $app;
+    }
+
+    /**
+     * make a new cookie instance.
+     *
+     * @param  string  $name
+     * @param  string  $value
+     * @param  int     $maxAge
+     * @param  string  $path
+     * @param  bool    $secure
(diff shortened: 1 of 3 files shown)


tobscure:
Nice! Hmm I'd probably rather have this as a CookieFactory class though since we will need to inject SettingsRepository etc. in the future.

sijad:
how about now?

tobscure:
Perfect! 😄

src/Http/CookieFactory.php
```diff
@@ -0,0 +1,63 @@
...
+     * @param  string  $name
+     * @param  string  $value
+     * @param  int     $maxAge
+     * @param  string  $path
+     * @param  bool    $secure
+     * @param  bool    $httpOnly
+     * @param  string  $domain
+     * @return \Dflydev\FigCookies\SetCookie
+     */
+    public function make($name, $value = null, $maxAge = null, $path = null, $secure = null, $httpOnly = true, $domain = null)
```
franzliedke: @sijad Are all these parameters necessary? They don't seem to be used anywhere...

franzliedke: Never mind. I've amended this a bit in https://github.com/flarum/core/commit/974f45e4e8529cc56757229d460af65c290dad1a and https://github.com/flarum/core/commit/8540932638388b758ed475248c4161eb3cd27c5c.

sijad: I just used Laravel cookie, it might be used by other developers see:
https://github.com/laravel/framework/blob/5.3/src/Illuminate/Foundation/helpers.php#L294

PR #1093, opened 2016-12-24
Allow to manually activate users

### LINKED ISSUE #414: Ability to manually confirm a user's email address
Someone mentioned not recieving emails for validation, and manual approval does not exist. Also needs ability to see members online, and possibly a public directory. Basically just needs full user management

renyuneyun:
Add a button to EditUserModal to allow manually activation of users in the web interface.

fix #414

### DIFF
--- a/js/forum/src/components/EditUserModal.js
+++ b/js/forum/src/components/EditUserModal.js
@@ -15,6 +15,7 @@ export default class EditUserModal extends Modal {
 
     this.username = m.prop(user.username() || '');
     this.email = m.prop(user.email() || '');
+    this.isActivated = m.prop(user.isActivated() || false);
     this.setPassword = m.prop(false);
     this.password = m.prop(user.password() || '');
     this.groups = {};
@@ -49,6 +50,16 @@ export default class EditUserModal extends Modal {
                 <input className="FormControl" placeholder={extractText(app.translator.trans('core.forum.edit_user.email_label'))}
                   bidi={this.email} />
               </div>
+              {!this.isActivated() ? (
+                <div>
+                  {Button.component({
+                    className: 'Button Button--block',
+                    children: app.translator.trans('core.forum.edit_user.activate_button'),
+                    loading: this.loading,
+                    onclick: this.activate.bind(this)
+                  })}
+                </div>
+              ) : ''}
             </div>,
 
             <div className="Form-group">
@@ -98,6 +109,24 @@ export default class EditUserModal extends Modal {
         </div>
       </div>
     );
+  }
+
+  activate() {
+    this.loading = true;
+    const data = {
+      username: this.username(),
+      isActivated: true,
+    };
+    this.props.user.save(data, {errorHandler: this.onerror.bind(this)})
+      .then(() => {
+        this.isActivated(true);
+        this.loading = false;
+        m.redraw();
+      })
+      .catch(() => {
+        this.loading = false;
+        m.redraw();
+      });
   }
 
   data() {
--- a/js/forum/src/components/EditUserModal.js
+++ b/js/forum/src/components/EditUserModal.js
@@ -38,14 +38,14 @@ export default class EditUserModal extends Modal {
       <div className="Modal-body">
         <div className="Form">
           <div className="Form-group">
-            <label>Username</label>
+            <label>{app.translator.trans('core.forum.edit_user.username_heading')}</label>
             <input className="FormControl" placeholder={extractText(app.translator.trans('core.forum.edit_user.username_label'))}
               bidi={this.username} />
           </div>
 
           {app.session.user !== this.props.user ? [
             <div className="Form-group">
-              <label>Email</label>
+              <label>{app.translator.trans('core.forum.edit_user.email_heading')}</label>
               <div>
                 <input className="FormControl" placeholder={extractText(app.translator.trans('core.forum.edit_user.email_label'))}
                   bidi={this.email} />
@@ -63,7 +63,7 @@ export default class EditUserModal extends Modal {
             </div>,
 
             <div className="Form-group">
-              <label>Password</label>
+              <label>{app.translator.trans('core.forum.edit_user.password_heading')}</label>
               <div>
                 <label className="checkbox">
                   <input type="checkbox" checked={this.setPassword()} onchange={e => {
@@ -72,7 +72,7 @@ export default class EditUserModal extends Modal {
                     if (e.target.checked) this.$('[name=password]').select();
                     m.redraw.strategy('none');
                   }}/>
-                  Set new password
+                  {app.translator.trans('core.forum.edit_user.set_password_label')}
                 </label>
                 {this.setPassword() ? (
                   <input className="FormControl" type="password" name="password" placeholder={extractText(app.translator.trans('core.forum.edit_user.password_label'))}
@@ -83,7 +83,7 @@ export default class EditUserModal extends Modal {
           ] : ''}
 
           <div className="Form-group EditUserModal-groups">
-            <label>Groups</label>
+            <label>{app.translator.trans('core.forum.edit_user.groups_heading')}</label>
             <div>
               {Object.keys(this.groups)
                 .map(id => app.store.getById('groups', id))


js/forum/src/components/EditUserModal.js
```diff
@@ -33,26 +34,55 @@ export default class EditUserModal extends Modal {
   }
 
   content() {
+    const topObj = this;
+    const activateButton = Button.component({
+      className: 'Button Button--block',
+      children: app.translator.trans('core.forum.edit_user.activate'),
+      loading: topObj.loading,
+      onclick: function() {
```
dav-is: Extract to method

js/forum/src/components/EditUserModal.js
```diff
@@ -33,26 +34,55 @@ export default class EditUserModal extends Modal {
   }
 
   content() {
+    const topObj = this;
```
dav-is: Use this instead of topObj

js/forum/src/components/EditUserModal.js
```diff
@@ -33,26 +34,55 @@ export default class EditUserModal extends Modal {
...
+    const activateButton = Button.component({
```
dav-is: Should place directly in return

js/forum/src/components/EditUserModal.js
```diff
@@ -33,26 +34,55 @@ export default class EditUserModal extends Modal {
...
+          username: topObj.username(),
+          isActivated: true,
+        };
+        topObj.props.user.save(data, {errorHandler: topObj.onerror.bind(topObj)})
+          .then(() => {
+            topObj.isActivated(true);
+            topObj.loading = false;
+            m.redraw();
+          })
+          .catch(() => {
```
dav-is: Don't think you need catch() as you use the errorHandler

js/forum/src/components/EditUserModal.js
```diff
@@ -33,26 +34,55 @@ export default class EditUserModal extends Modal {
...
 
           {app.session.user !== this.props.user ? [
             <div className="Form-group">
-              <label>Email</label>
+              <label>{app.translator.trans('core.forum.edit_user.email_label')}</label>
               <div>
                 <input className="FormControl" placeholder={extractText(app.translator.trans('core.forum.edit_user.email_label'))}
                   bidi={this.email} />
               </div>
+              {this.isActivated() == false ? (
```
dav-is: `{this.isActivated() ? (`

dav-is: Sorry typo
`{!this.isActivated() ? (`

js/forum/src/components/EditUserModal.js
renyuneyun: I am just copy-pasting from `onSubmit()`. So I guess that part (`onSubmit()`) should also be changed?

js/forum/src/components/EditUserModal.js
renyuneyun: At least for what I've tried, using this in the Button object would refer to the Button itself rather than the outer class.
(I'm not familiar with JS, maybe there are other ways that I don't know.)

dav-is: Well since you aren't inside of the actual class of the button, you can use `this`.

renyuneyun: But where I actually use `topObj` is only in the instance of Button...

dav-is: You're supplying parameters to the button

dav-is: Here's an example: https://github.com/flarum/core/blob/34f3d93ce56ad18ada6aad65bfc77244fea09716/js/forum/src/components/ChangePasswordModal.js#L26

renyuneyun: Thanks. Now I know the reason: I need `bind(this)`...

js/forum/src/components/EditUserModal.js
```diff
@@ -100,6 +111,24 @@ export default class EditUserModal extends Modal {
     );
   }
 
+  activate() {
+      this.loading = true;
```
dav-is: Too many spaces here

js/forum/src/components/EditUserModal.js
```diff
@@ -15,6 +15,7 @@ export default class EditUserModal extends Modal {
 
     this.username = m.prop(user.username() || '');
     this.email = m.prop(user.email() || '');
+    this.isActivated = m.prop(user.isActivated() || '');
```
dav-is: `this.isActivated = m.prop(user.isActivated() || false);`

`user.isActivated` is a boolean, so let's keep it that way.

Luceos:
Assigned to Merge Master @franzliedke (hope that still applies, if not let me know).

js/forum/src/components/EditUserModal.js
```diff
@@ -37,22 +38,32 @@ export default class EditUserModal extends Modal {
...
+              {!this.isActivated() ? (
+                <div>
+                  {Button.component({
+                    className: 'Button Button--block',
+                    children: app.translator.trans('core.forum.edit_user.activate_button'),
+                    loading: this.loading,
+                    onclick: this.activate.bind(this)
+                  })}
+                </div>
+              ) : ''}
```
franzliedke: Please avoid the double negation - i.e. change it to `this.isActivated() ? '' : (<div>...</div>)`.

js/forum/src/components/EditUserModal.js
```diff
@@ -100,6 +111,24 @@ export default class EditUserModal extends Modal {
...
+    const data = {
+      username: this.username(),
+      isActivated: true,
+    };
+    this.props.user.save(data, {errorHandler: this.onerror.bind(this)})
+      .then(() => {
+        this.isActivated(true);
+        this.loading = false;
+        m.redraw();
+      })
```
franzliedke: I believe this should set the user property to the returned object, instead of overwriting the `isActivated` property, correct, @tobscure ?

js/forum/src/components/EditUserModal.js
dav-is: Interesting... is there reasoning behind this? It seems a bit confusing for "if is actived show nothing else show this button" instead of "if is not activated show button"

js/forum/src/components/EditUserModal.js
dav-is: Well init is only run when the model is opened. So this.isActivated is set to the activation status when it's opened. When the user is saved, it updates this.props.user but not this.isActivated.

js/forum/src/components/EditUserModal.js
renyuneyun: I agree with dav-is, it will make the logic less straightforward.

Anyway, if this doesn't follow the style guide, I'll change it.
(wow, this sentence is also an example of how common logic expresses)

js/forum/src/components/EditUserModal.js
renyuneyun: Indeed as dav-is's saying, it is a button, not a checkbox.

I initially designed it as a checkbox, but then realised it's almost nonsense to 'deactivate' the user.
As a button, it should take effects immediately.

js/forum/src/components/EditUserModal.js
dav-is: The code used actually isn't a double negation. A double negation would be `if (!!this.isActivated)` which doesn't make much sense.

franzliedke: What I meant with "double negation" is that there is still an `else` clause at the end.

And when I see something like this:
~~~js
if ( ! something) {
  // Do A
} else {
  // Do B
}
~~~

I automatically think: Why not turn it around?

~~~js
if (something) {
  // Do B
} else {
  // Do A
}
~~~

That way, there will be less negation in my head when trying to understand the if and else blocks.

However, we're only using the ternary operator here because there is no inline if statement - therefore I agree with you that it's better to have the one condition that we are checking for here come first - the `else` here is basically a no-op.

franzliedke:
Very good job, @renyuneyun ! Thank you!

Issue #978, opened 2016-06-09
Use queues for sending emails


tobyzerner:
Sometimes submitting a post can be slow because you have to wait for Flarum to send notification emails on the backend before it sends back the 200 OK response.

We should look at using Laravel's Queue component (or similar) to send emails asynchronously. Of course it will be optional as many shared hosts won't have the ability to run a queue worker in the background.

franzliedke:
Yes, please. Using the `sync` driver by default. We can improve this even more, by using one of my two experimental approaches from a long time ago: [tardiqueue](https://github.com/franzliedke/tardiqueue) and [threadqueue](https://github.com/franzliedke/threadqueue).

Both (but probably mostly the first one) are aimed at providing a more performant "poor man's queue" system for use with e.g. shared hosts.

Luceos:
@franzliedke something you might need to consider is the limitation by many shared hosts of the number of processes allowed per hosting plan/website. A common OS that implements this is CloudLinux.

franzliedke:
@Luceos Laravel's `sync` driver is basically just a dummy - it executes a job immediately instead of pushing it onto a queue. It's basically the poor man's queue adapter. So is my tardiqueue extension, it's just a little more advanced.

Luceos:
Yes I understand, I was referring to the threadqueue solution @franzliedke ;)

franzliedke:
Gotcha, yeah, that should only be used where possible.

franzliedke:
That's highly experimental anyhow.

Luceos:
Would like to take a shot at this if you're ok with it @flarum/core

franzliedke:
@Luceos Please do. :+1:

TODO list looks great.

Please create a stripped-down version of the service provider - the one from Laravel registers all the drivers in order to be as configurable as possible (and some console commands which we don't need for now). Ours should only set up the sync driver for now - at one point we will have a config flag that will be used to switch to another driver instead. The [`DatabaseServiceProvider`](https://github.com/flarum/core/blob/9ffdeff608895efbf769259aa6732a3cac0bf582/src/Database/DatabaseServiceProvider.php) might serve as inspiration.

Luceos:
First implementation already sends over sync. I've extended the illuminate service provider and the QueueManager. I'll see if I can make a non-extended provider.

Luceos:
Some thoughts I had:

[...]

PR #1096, opened 2016-12-29
wip queue implementation

luceos:
> __this is a work in progress__

Relates to #978 , please keep all discussion (not code review related) to the issue.

### DIFF
--- a/src/Core/Listener/EmailConfirmationMailer.php
+++ b/src/Core/Listener/EmailConfirmationMailer.php
@@ -19,10 +19,10 @@ use Flarum\Event\UserEmailChangeWasRequested;
 use Flarum\Event\UserWasRegistered;
 use Flarum\Forum\UrlGenerator;
 use Flarum\Settings\SettingsRepositoryInterface;
+use Illuminate\Contracts\Bus\Dispatcher as JobDispatcher;
 use Illuminate\Contracts\Events\Dispatcher;
 use Illuminate\Contracts\Mail\Mailer;
 use Symfony\Component\Translation\TranslatorInterface;
-use Illuminate\Contracts\Bus\Dispatcher as JobDispatcher;
 
 class EmailConfirmationMailer
 {
@@ -52,6 +52,7 @@ class EmailConfirmationMailer
      * @param Mailer $mailer
      * @param UrlGenerator $url
      * @param TranslatorInterface $translator
+     * @param JobDispatcher $queue
      */
     public function __construct(SettingsRepositoryInterface $settings, Mailer $mailer, UrlGenerator $url, TranslatorInterface $translator, JobDispatcher $queue)
     {
--- a/src/Core/Queue/QueueServiceProvider.php
+++ b/src/Core/Queue/QueueServiceProvider.php
@@ -69,8 +69,8 @@ class QueueServiceProvider extends Provider
 
     protected function registerDispatcher()
     {
-        $this->app->singleton(DispatcherContract::class, function($app) {
-            return new Dispatcher($app, function() use ($app) {
+        $this->app->singleton(DispatcherContract::class, function ($app) {
+            return new Dispatcher($app, function () use ($app) {
                 return $app['queue']->connection();
             });
         });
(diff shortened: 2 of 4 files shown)


src/Core/Listener/EmailConfirmationMailer.php
```diff
@@ -11,20 +11,22 @@
...
 use Flarum\Settings\SettingsRepositoryInterface;
+use Illuminate\Contracts\Bus\Dispatcher as JobDispatcher;
 use Illuminate\Contracts\Events\Dispatcher;
 use Illuminate\Contracts\Mail\Mailer;
-use Illuminate\Mail\Message;
 use Symfony\Component\Translation\TranslatorInterface;
 
 class EmailConfirmationMailer
 {
+    use DispatchJobsTrait;
```
franzliedke: Honestly, I don't think we need this trait, if all it does is shortening `$this->queue->dispatch()` to `$this->dispatch()`, at the cost of one more step to find out where this method is coming from, and why we need to inject the seemingly-unused `$queue` attribute. :)

src/Core/Queue/QueueServiceProvider.php
```diff
@@ -0,0 +1,78 @@
...
+            // creating the classes that accept queue configs and instantiate queues.
+            $manager = new QueueManager($app);
+
+            // We've disabled all other queue connectors for now. Once we're able to
+            // configure any of the other connectors, we can easily reset the old
+            // behavior of the parent provider by registering all connectors.
+            $this->registerSyncConnector($manager);
+
+            return $manager;
+        });
```
franzliedke: I don't think we need the `QueueManager` at all,. Just configure the correct `queue.connection` binding here, and we are good. Once we make this configurable (to switch between the default "sync" and other - real - drivers), this would just be a different instance, but it would be configured here,

franzliedke:
Good job on this! 👏 

Just some minor things...

src/Core/Listener/EmailConfirmationMailer.php
Luceos: You might want to create an issue then to remove the need of a DispatchesEventTrait as well, it does exactly the same useless shortcut :) It might make more sense if the trait would load the bound Dispatcher too from my perspective.

franzliedke: To be fair, that one at least encapsulates [some logic](https://github.com/flarum/core/blob/01eba18164d6eedd7cd3b58169fa5779ae6306ad/src/Core/Support/DispatchEventsTrait.php#L32-L36) ;)

Luceos: Must have missed that 👍

src/Core/Queue/QueueServiceProvider.php
Luceos: @franzliedke so you want me to only set up the default connection and then call the QueueServiceProvider of Illuminate?

franzliedke: No. Only register the connection instance. We don't need a manager at all, I think, because we don't need the dynamic switching between different connections.

Luceos: For simplicity purposes we can go that direction. Although I would know of situations where an extension might be happy to use separate queue drivers. Ok, thanks!

tobscure:
Why can't we use Laravel Mailer's built-in queue functionality? https://laravel.com/docs/5.3/mail#queueing-mail

franzliedke:
Because we're not on 5.3 yet. ;)

tobscure:
It exists in 5.1, with a different API: https://laravel.com/docs/5.1/mail#queueing-mail

tobscure:
My concern is that this proposed MailJob API is very restricted (you can't customise the Message object), and once we start trying to open it up, we'll essentially be duplicating Laravel's API unnecessarily?

Luceos:
@tobscure with all respect, I think we need to implement a queue initially. Moving towards mailable makes a lot of sense, but - in my opinion - would be step 2 and can be easily done once @sijad is done with his compatibility towards 5.3+. I picked the mail sending initially, because that one is blocking api request.

Luceos:
@flarum/core Should we hold this off until @sijad has finished the upgrade to L 5.3 or 5.4 or 5.5?

PR #1094, opened 2016-12-27
Make Add Extension modal's title translatable

milescellar:
Associated with: https://github.com/flarum/flarum-ext-english/pull/93

PR #1098, opened 2016-12-31
Make Add Extension modal's title translatable (#1094 for next-front)

milescellar:
This is just the `next-front` branch version of the #1094 PR.

### DIFF
--- a/js/admin/src/components/AddExtensionModal.js
+++ b/js/admin/src/components/AddExtensionModal.js
@@ -15,7 +15,7 @@ export default class AddExtensionModal extends Modal {
   }
 
   title() {
-    return 'Add Extension';
+    return app.translator.trans('core.admin.add_extension.title');
   }
 
   content() {


milescellar:
(Sorry I don't know how to work on different branches on a same PR so I opened a new one)

franzliedke:
@dcsjapan This is the same change as in [the MR for master](https://github.com/flarum/core/pull/1094/files). Is this still correct?

franzliedke:
If this is correct, we can close this PR, because the change in master will merge just fine.

milescellar:
Yeah, it's correct, there are no changes at all, this structure is unchanged for this one, so it's indeed unnecessary. Just in case, I let @dcsjapan confirm before closing it.

milescellar:
Oh, I see, good to know! I'll wait @tobscure 's feedback before sending a new commit to this PR to change that to be really sure. Thanks! 😄

PR #1100, opened 2017-01-02
Update to Zend Stratigility 1.3

### LINKED ISSUE #1069: Update to Stratigility 1.3
Updating will require some changes in our middleware stack to stay forward-compatible with the upcoming version 2.0 of Stratigility, see https://docs.zendframework.com/zend-stratigility/migration/to-v2/.

Important: MR #1066 should also be reverted as part of this.

franzliedke:
* Fix dependency version constraint. (Reverts #1066.)
* Allow exceptions to be raised when dispatching middleware.
* Fix our error handler middleware (do not implement Stratigility's
  error handler interface, catch exceptions instead).

See https://docs.zendframework.com/zend-stratigility/migration/to-v2/.

Closes #1069.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: php
 
 php:
-  - 5.5
   - 5.6
   - 7.0
   - hhvm
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ language: php
 php:
   - 5.6
   - 7.0
+  - 7.1
   - hhvm
 
 matrix:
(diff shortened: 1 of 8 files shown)


Luceos:
@flarum/core  tests fail because Stratigility's minimum php version for 1.3+ is 5.6. We might need to up our version now too.

franzliedke:
Given that 5.6 now only receives security fixes, that probably makes sense anyway.

@tobscure Do you agree?

Luceos:
@franzliedke can you update this PR to update `.travis.yml` and `composer.json` once @tobscure approves?

franzliedke:
That also means, we're now requiring PHP 5.6, but that simply makes sense. :)

tobscure:
Beautiful :)

Luceos:
@franzliedke can you take care of branch deletion if you feel it's now ready as is?

franzliedke:
Done, thanks.

Issue #820, opened 2016-02-19
More attributes in app.blade.php


sijad:
`dir` and `lang` attributes needs to be added to `<html>` tag, it should be like this:

`<html dir="{{ $dir }}" lang="{{ $lang }}">`

it'll allow developers to change document direction in browser easily.
also it'll be a good idea if `<body>` tag had a configurable `class` attribute

update: 
related links:
- [Why use the language attribute?](https://www.w3.org/International/questions/qa-lang-why.en)
- [Why you should use "dir" attribute](https://alicoding.com/why-you-should-use-dir-attribute-more-for-right-to-left-localization/)

franzliedke:
@sijad How exactly do you want to make the body's `class` attribute to be configurable? A custom class name is not necessary, you can just use `body` as a selector in your CSS. I suppose you mean something different... Are you talking about having different classes attached to the body depending on which page is open?

sijad:
had something like [modernizr](https://modernizr.com/) in mind
one of the use cases, body classes can help developer do things only using css (like conditions in css) and no javascript

franzliedke:
Yes, but which things? ;)

sijad:
as said it's just an idea, there can be lots of use cases:
* user is logged in
* user group
* user is admin
* ...
just saying it not that bad have this option, adding these classes in **back-end** means they can also be used in case user javascript is disabled

franzliedke:
Hmm, technically this should be possible, because we refresh the page when logging in / out.

franzliedke:
Thanks to David, direction and language will now be exposed.

tobscure:
So now we just need to expose body class? Is there anything else?

franzliedke:
Yes, we probably need to come up with a list of which things we are talking about (and then create a new ticket with some instructions to make it contributor-friendly).

luceos:
@tobscure this relates to the frontend refactor. If we add a body class this can be closed?

tobscure:
We'll do that in a separate PR.

franzliedke:
Let's close this. When the use-cases for concrete classes come up, these can be tracked / added separately.

PR #1105, opened 2017-01-14
Add more attributes in app.blade.php and add a setDescription method.

datitisev:
Related to #820, but doesn't close it.

- Added `$language` and `$direction` protected properties to WebAppView
- Added `setLanguage` and `setDirection` public methods to WebAppView
- Added missing `setDescription` public method to WebAppView 🙂 

Please feel free to correct me, I feel like `$this->language ? $this->language : $this->locales->getLocale()` can be done in a better, cleaner way 😛 

final result

### DIFF
--- a/src/Http/WebApp/WebAppView.php
+++ b/src/Http/WebApp/WebAppView.php
@@ -329,7 +329,7 @@ class WebAppView
         $view->title = $this->buildTitle(array_get($forum, 'data.attributes.title'));
         $view->description = $this->description ? $this->description : array_get($forum, 'data.attributes.description');
         $view->language = $this->language ? $this->language : $this->locales->getLocale();
-        $view->direction = $this->direction ? $this->direction : "ltr";
+        $view->direction = $this->direction ? $this->direction : 'ltr';
 
         $view->modules = $this->modules;
         $view->payload = $this->buildPayload($request, $forum);
--- a/src/Http/WebApp/WebAppView.php
+++ b/src/Http/WebApp/WebAppView.php
@@ -42,6 +42,20 @@ class WebAppView
     protected $description;
 
     /**
+     * The language of the document, displayed as the value of the attribute `dir` in the <html> tag.
+     *
+     * @var null|string
+     */
+    protected $language = null;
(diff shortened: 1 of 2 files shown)


src/Http/WebApp/WebAppView.php
```diff
@@ -283,7 +327,9 @@ public function render(Request $request)
         $view = $this->view->file(__DIR__.'/../../../views/app.blade.php');
 
         $view->title = $this->buildTitle(array_get($forum, 'data.attributes.title'));
-        $view->description = $this->description;
+        $view->description = $this->description ? $this->description : array_get($forum, 'data.attributes.description');
+        $view->language = $this->language ? $this->language : $this->locales->getLocale();
+        $view->direction = $this->direction ? $this->direction : 'ltr';
```
franzliedke: There is indeed a shorter version of this - see [here](http://stackoverflow.com/questions/2153180/what-is-in-php-5-3).

~~~php
$view->direction = $this->direction ?: 'ltr';
~~~

datitisev: Thanks! That's a neat................ _thing_ 😛  and it makes sense.

franzliedke:
Thanks for all your help! ❤️

Luceos:
@franzliedke @datitisev 

I don't understand why these properties have been made protected with setters only. How should this help extendibility? There's no way for an extension to see now what is currently set, so modifying is a blind guess. In my opinion there should be a very strong argument to make anything protected/private like this.

franzliedke:
Hmm yeah, good point. @tobscure What do you think about this?

tobscure:
We should have getters and setters.

franzliedke:
[Done.](https://github.com/flarum/core/commit/bbcc33b5b579b867f39236330d83f194d3801955) I used public properties now, to avoid some superfluous typing - and the setters did not do any type / sanity checking anyways.

tobscure:
I suppose that makes sense – as long as we can't foresee any instance in the future where we might want to do some type/sanity checking?

PR #1108, opened 2017-01-17
Fix missing meta description

recallfx:
This is improved version of #1088 with possible fix for #1057;


franzliedke:
I appreciate the effort, but I think there is no need for modifying the meta tag on the client side - those are only relevant on initial page load, right?

That said, the bug has been fixed in #1105, and I will see whether we can extract the new feature (using a topic's first post for the description meta tag) from #1088.

Thank you for your contribution! 🍻

PR #1113, opened 2017-01-25
Correct image orientation according to Exif data

oanhnn:
When using mobile, take a photo and upload it as avatar, it's orientation is incorrect.
This commit will fix this problem.


franzliedke:
We require the `mbstring` extension anyway, so this should be good, thank you @oanhnn !

PR #1114, opened 2017-01-30
Add HTTP2 Push Headers

tankerkiller125:
This code adds the HTTP2 Push headers that allow Nginx and some CDN's with HTTP2 enabled to push the JS and CSS assets decreasing overall load time. See https://discuss.flarum.org/d/4768-http2-server-push/ for details.

### DIFF
--- a/src/Http/WebApp/WebAppView.php
+++ b/src/Http/WebApp/WebAppView.php
@@ -300,7 +300,7 @@ class WebAppView
 
         return $view->render();
     }
-    
+
     protected function addLinkHeaders()
     {
         $css = [
--- a/src/Http/WebApp/WebAppView.php
+++ b/src/Http/WebApp/WebAppView.php
@@ -313,13 +313,13 @@ class WebAppView
         ];
         foreach ($css as $cssUrl) {
             if ($cssUrl != null) {
-                header('Link: <' . $cssUrl . '>; rel=preload, as=style', false);
+                header('Link: <'.$cssUrl.'>; rel=preload, as=style', false);
             }
         }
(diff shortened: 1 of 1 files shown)


src/Http/WebApp/WebAppView.php
```diff
@@ -293,13 +293,37 @@ public function render(Request $request)
...
+        ];
+        foreach ($css as $cssUrl) {
+            if ($cssUrl != null) {
+                header('Link: <'.$cssUrl.'>; rel=preload, as=style', false);
+            }
+        }
+
+        foreach ($js as $jsUrl) {
+            if ($jsUrl != null) {
+                header('Link: <'.$jsUrl.'>; rel=preload, as=script', false);
```
franzliedke: Hmm, we do not use the `header()` function, as they touch global state. These headers should instead be set on the [PSR-7 response object](http://www.php-fig.org/psr/psr-7/) that encapsulates the HTTP response. Trouble is, right now the response object is not available in this class - any quick idea, @tobscure?

tankerkiller125: I was originally going to make a middleware for this but I couldn't seem to figure out how that works. Would actually be better as a middleware I think because we could scan the entire page for images and all assets to push. Instead of just the ones defined here.

franzliedke:
Closing this for now, as I am unsure about the performance impact in a real-world scenario (see [this post](https://discuss.flarum.org/d/4768-http2-server-push/25) and the discussion that precedes it).

Thanks for the initiative, @tankerkiller125 !

PR #1124, opened 2017-02-11
fixed issues with $extension visibility and typehinting

Luceos:
Fixes three things:

- Visibility of the `extension` property, which cannot be used from the outside now.
- Updated php doc, which I forgot during the extension rewrite.
- Added one more typehint for Extension

The first one is very important to take along in the next version.


franzliedke:
Uh, thanks. Apparently I [didn't quite do it properly](https://github.com/flarum/core/commit/9349ed13fc74f041ca93e2d057b2c3ac3629a1d8), whoopsie.

Cheers. :beers:

PR #1131, opened 2017-02-16
Affix sidebar when window is resized

### LINKED ISSUE #866: affixSidebar should check window size on resize event
[see here](https://github.com/flarum/core/blob/master/js/forum/src/utils/affixSidebar.js#L17)
not sure if it's a real problem or not, but on tabltes, window resizes when user rotate device to landscape or portrait after document is loaded and it can make a little problem and user may have to refresh the page or rotate it again

franzliedke:
Fixes #866.

Phew... that was harder than I thought.

#### Implementation
- Bind an event handler for window resize events.
- Execute this handler exactly once to get the initial setup working.
- Tweak the handler a little bit (in contrast to the current implementation that is only executed once): Always remove the affix plugin first (see below for why) and then apply it again, if appropriate given the current dimensions.

The challenging part was realizing that the height/offset calculation did not work once the affix was applied. This was due to the surrounding `$sidebar` element essentially having zero height, because its content - the affixed `$affixElement` - had received fixed positioning. Thus, the calculation broke down. To fix this, I had to remove the affix plugin calculation.

#### Downsides
- Because there is no easy way (that I found) to remove the affix from an element, I had to remove it globally, following [this StackOverflow answer](https://stackoverflow.com/questions/15366519/how-do-i-remove-bootstrap-affix-from-an-element-under-certain-conditions#15372789). This means that having multiple affixed regions on one page would probably break down.

#### Questions
To @tobscure mostly...

1. Can we live with this downside of only supporting one affixed region per page?
2. Is my implementation okay? It's a little unnecessary overhead, but in the real world, resizing should not happen as often as it did during my testing (where the impact seemed fine).

### DIFF
--- a/js/forum/dist/app.js
+++ b/js/forum/dist/app.js
@@ -31146,29 +31146,42 @@ System.register('flarum/utils/abbreviateNumber', [], function (_export, _context
 System.register('flarum/utils/affixSidebar', [], function (_export, _context) {
   "use strict";
 
-  function affixSidebar(element, isInitialized) {
+  function affixSidebar(element, isInitialized, context) {
     var _this = this;
 
     if (isInitialized) return;
 
-    var $sidebar = $(element);
-    var $header = $('#header');
-    var $footer = $('#footer');
+    var onresize = function onresize() {
+      var $sidebar = $(element);
+      var $header = $('#header');
+      var $footer = $('#footer');
+      var $affixElement = $sidebar.find('> ul');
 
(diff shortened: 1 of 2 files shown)


tobscure:
Nicely done!

Maybe in the future we can either have more sophisticated affix behaviour (scroll the sidebar for as long as it has more content, and *then* affix it - same thing when scrolling back up), or we can give it its own scrollbar as sijad suggested... or we can just drop the idea of a fixed sidebar altogether (like on the discussions index), not sure how valuable it is.

PR #1135, opened 2017-02-26
Prevent crawlers to index nojs pages

### LINKED ISSUE #912: Prevent crawling of nojs pages
Reported by @dav-is via Gitter chat:

It looks like Google crawled the community site while it was experiencing JS issues and ended up recording some `nojs` URLs.

nojs

Some sort of countermeasures are needed to keep search engines away from these URLs.

It might also be a good idea to use Google web tools to get the site recrawled, so people won't follow the links to these URLs.

sijad:
this resolves #912 

as it's mentioned in [this document](https://developers.google.com/webmasters/control-crawl-index/docs/robots_meta_tag#using-the-robots-meta-tag) by using `robots` meta tag it's possible to prevent crawlers to index specific pages.

maybe content value should be [none](https://developers.google.com/webmasters/control-crawl-index/docs/robots_meta_tag#valid-indexing--serving-directives)? instead of `noindex`

### DIFF
--- a/views/app.blade.php
+++ b/views/app.blade.php
@@ -6,6 +6,9 @@
     <meta name="description" content="{{ $description }}">
     <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1, minimum-scale=1">
     <meta name="theme-color" content="{{ array_get($forum, 'attributes.themePrimaryColor') }}">
+    @if (! $allowJs)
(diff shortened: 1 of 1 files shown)


franzliedke:
I think `noindex` is perfectly fine. I don't see any harm in following the links on the page.

PR #1136, opened 2017-02-28
Fix modal layout when width between 601px and 768px

SDLyu:
This PR to resolve modal layout, there's a blank area when window width between 601px and 768px.

flarum_community

### DIFF
--- a/less/lib/Modal.less
+++ b/less/lib/Modal.less
@@ -155,6 +155,7 @@
     }
   }
   .Modal {
+    max-width: 100%;
     margin: 0;
     -webkit-transform: none !important;
             transform: none !important;


dav-is:
Finally! I've been dealing with this bug on chrome dev tools for a while but was never motivated enough to really look into it.

franzliedke:
Hmm, I still see a small strip of blank area, even with this fix applied.
blank area next to sign up screen

Which browsers were you folks testing on?

SDLyu:
Weird, I have tested Chrome, FireFox and Safari and not see blank area 😕 
## Chrome
flarum_community

## FireFox
flarum_community

## Safari
flarum_community

@franzliedke can you show more information about window width?

dav-is:
@franzliedke Could you provide more information on what you experienced here?

datitisev:
I see it too, on macOS Sierra, Google Chrome Version 59.0.3071.36 (Official Build) beta (64-bit)

img

I tested on DevFlarum and just added the style
img

The `.ModalManager` seems to have right padding of 15px... included in the element's inline style 😕 
img
img

datitisev:
Whoops, I was trying to see if I could reproduce the 15px issue, and somehow ended up pushing the exact same fix that you PR'd here (9c0d921f49e049d4374ad666f065089521ef76a5).

Thank you for your PR! Apologies that it wasn't merged before.

The padding right still exists, but it's less empty space than without this PR's fix applied. It doesn't seem to be coming from our code. A simple `padding-right: 0 !important` fixes it, so that might be what we go with.

PR #1147, opened 2017-03-20
Improve slug + Design minor changes

yagobski:
Fix issue #194:  Slug transliteration 
Minor design change


faouzi.txt
```diff
@@ -0,0 +1 @@
+
```
dav-is: What is this?

less/lib/App.less
```diff
@@ -169,7 +169,7 @@
   list-style: none;
 }
 .Header-logo {
-  max-height: 30px;
+  max-height: 20px;
```
dav-is: Whys the CSS being changed?

src/Util/Str.php
```diff
@@ -24,10 +24,660 @@ class Str
...
+        $str = self::remove_accents($str);
         $str = preg_replace('/[^a-z0-9]/i', '-', $str);
         $str = preg_replace('/-+/', '-', $str);
         $str = preg_replace('/-$|^-/', '', $str);
 
         return $str ?: '-';
     }
+
+    /**
+     * Converts all accent characters to ASCII characters.
```
dav-is: I don't think this extremely long comment is necessary

Issue #442, opened 2015-09-07
Arabic titles doesn't work for URL


samymassoud:
When writing Arabic title it appears like dashes in URL

For example try this title
`موضوع جديد` which is 'New topic` By the way :D

You will see URl like this
`d/3--`

franzliedke:
We already have a ticket for this: #194

Issue #433, opened 2015-09-04
SEO URLs Faulty


jenolan:
There is a discussion on [Flarum](http://discuss.flarum.org) currently ```http://discuss.flarum.org/d/855-Is%20this%20what%20it's%20like%20to%20develop%20a%20software?``` problem is if the 'nojs' is added you end up with ```http://discuss.flarum.org/d/855-Is%20this%20what%20it's%20like%20to%20develop%20a%20software??nojs=1?nojs=1``` replicating eternally.
The SEO slug should be .. 
  1. A-Za-z0-9 only
  2. Space converted to '-'
  3. All other characters eliminated

Fastidious:
I do not see this, @jenolan. I see the slug as:

`http://discuss.flarum.org/d/855-is-this-what-it-s-like-to-develop-a-software`

and

`http://discuss.flarum.org/d/855-is-this-what-it-s-like-to-develop-a-software?nojs=1`

jenolan:
Do http://discuss.flarum.org/?nojs=1 the link for the topic is then http://discuss.flarum.org/d/855-Is%20this%20what%20it's%20like%20to%20develop%20a%20software?

Fastidious:
@jenolan ah, that way, in that order, yes, I see it.

modInfo:
I have a similar problem

So make it flarum:
https://spytajsie.com/d/34-portal-wygl-da-jak-stare-zapytaj-my-lisz-e-da-si-na-niego-zgromadzi-tyle-os-b-jak-kiedy-by-o-na-zapytaj/

Properly should be:
https://spytajsie.com/d/34-portal-wyglada-jak-stare-zapytaj-myslisz-ze-da-sie-na-niego-zgromadzi-tyle-osob-jak-kiedys-byl-o-na-zapytaj/

tobscure:
This is covered by #189 and #194.

Issue #189, opened 2015-07-28
Search engine optimization


tobyzerner:
Flarum renders very basic versions of the discussion list and discussions (within `<noscript>` tags), which are going to be specifically optimised for search engines. This all takes place in Flarum\Forum\IndexAction and DiscussionAction, which extend Support\ClientAction and use Flarum\Support\ClientView. The actions get data from the Flarum's API and pass it onto content views (discussion.blade.php and index.blade.php). I'm happy to explain more if need be because it can get pretty confusing.

Put simply, the code is a mess. It needs to be rethought and refactored. A few thoughts:
- We probably want to use or build some kind of library that makes it easier to work with a JSON-API document payload.
- Similarly, we probably want to use or build some kind of library that will help generate URLs/query strings/etc.

Only after doing this will we be able to effectively optimise the SEO content and do the following:
- Add <link rel="canonical"> tags
- Add <meta> description tags
- Think about other stuff like schema.org
- https://moz.com/blog/seo-cheat-sheet

axisdos:
Noindex for subpages (Inspired by Yoast SEO for WordPress) to avoid duplicate content. and URL Friendly.

franzliedke:
@axisdos Noindex isn't needed when we have the `rel=canonical` tag. :)

DCSJAPAN:
Questions of how to handle foreign language/unicode URLs -- such as https://github.com/flarum/core/issues/194 -- will also have SEO implications.

In that regard, unfortunately, I think different regions may prefer different solutions.

neandrake:
Would this tie into Accessibility at all? I'm not extremely familiar with it, but when structuring the noscript HTML along with the rest of the page, it would be nice if parts of it were identifiable for users with disabilities.

http://www.w3.org/WAI/intro/aria

franzliedke:
Could [this approach](http://www.sitepoint.com/cutting-the-mustard-with-css-media-queries/) be an alternative to the <noscript> tags?

tobscure:
Quite possibly, feels a *little* bit hacky though? I'm not too fussed

dcsjapan:
https://github.com/flarum/core/issues/906 is an SEO-related issue specific to the Tags extension.

01010000101001100:
I investigated links on my Flarum and posted results in this topic on official Flarum forum, have a look:
https://discuss.flarum.org/d/3103-is-the-flarum-s-seo-seriously-messed-up-or-am-i-doing-something-wrong

[...]

PR #1148, opened 2017-03-20
Slug transliteration #194

yagobski:
Currently slugs are generated using only alphanumeric characters, replacing anything else with a hyphen. I improve Slug function to  support some degree of transliteration for non-Latin languages still get slugs.

### DIFF
--- a/src/Util/Str.php
+++ b/src/Util/Str.php
@@ -24,10 +24,280 @@ class Str
     public static function slug($str)
     {
         $str = strtolower($str);
+        $str = self::remove_accents($str);
         $str = preg_replace('/[^a-z0-9]/i', '-', $str);
         $str = preg_replace('/-+/', '-', $str);
         $str = preg_replace('/-$|^-/', '', $str);
 
         return $str ?: '-';
     }
+
+    /**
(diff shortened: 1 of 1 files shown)


yagobski:
Please check this pull request. It will fix the issues: 
Slug transliteration #194
Search engine optimization #189
SEO URLs Faulty #433
Arabic titles doesn't work for URL #442

franzliedke:
Did you take this code from WordPress?

franzliedke:
It looks like you did. Pretty sure that we can't use WordPress code (especially not without attribution) - as WordPress is licensed under GPL.

tobscure:
Plus we still haven't actually decided on a strategy/approach in #194?

yagobski:
Yes I used same function like Wordpress but a little bit modified.
This Slug strategy cover all approach described i #194
Flarum can't be used in other languages without this Slug function.

franzliedke:
I will close this then. 

We are more than happy to accept contributions here, but let's agree on a solution in #194 first. And we will have to use a library with a compatible license. 

Thanks for the initiative!

PR #1155, opened 2017-04-09
Change content column from TEXT to MEDIUMTEXT

### LINKED ISSUE #1044: Use MEDIUMTEXT for post content
Hi,

I build a Fluxbb to Flarum importer. On Fluxbb the equivalent of the column 'content' the type is mediumtext. So potentially a few posts exceed 65535 characters.

importer : https://github.com/mondediefr/fluxbb_to_flarum

ssfinney:
Fixes #1044 by addressing the following:

- A new migration to change the type of the content column in the posts table from TEXT to MEDIUMTEXT.
- Confirmed that `doctrine/dbal` already exists in the core composer.json, so that we can modify existing columns.

### DIFF
--- a/migrations/2017_04_09_152230_change_posts_content_column_to_mediumtext.php
+++ b/migrations/2017_04_09_152230_change_posts_content_column_to_mediumtext.php
@@ -1,0 +1,27 @@
+<?php
+
+/*
+ * This file is part of Flarum.
+ *
+ * (c) Toby Zerner <toby.zerner@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Illuminate\Database\Schema\Blueprint;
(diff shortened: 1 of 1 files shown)


Luceos:
As @tobscure pointed out to me, please use this (new) migration style:

https://github.com/flarum/core/blob/master/migrations/2015_10_24_194000_add_ip_address_to_posts.php

tobscure:
@luceos actually, that helper is only for adding columns. We don't have a helper yet for changing columns. So the current code is fine I think.

Luceos:
@tobscure wouldn't it make sense to cover that too then? My apologies for hijacking your pr @ssfinney :)

tobscure:
@Luceos Sure, but it's not necessarily in the scope of this PR :P

ssfinney:
Thanks guys :)

franzliedke:
Good job @ssfinney :ok_hand:

PR #1158, opened 2017-04-12
Use JSON for Install Command Instead of Yaml

dav-is:
Yaml is a pain to escape things. Currently, we can only safely pass alpha numeric characters into the yaml file safely. When passing the discussion title, something like "Davis' Forum" won't work because of the apostrophe. JSON can be encoded using `json_encode()` and will automatically escape special characters. It's much safer for automated use (which is probably all that would use the artisan commands).

### DIFF
--- a/src/Install/Console/FileDataProvider.php
+++ b/src/Install/Console/FileDataProvider.php
@@ -13,7 +13,6 @@ namespace Flarum\Install\Console;
 
 use Exception;
 use Symfony\Component\Console\Input\InputInterface;
-use Symfony\Component\Yaml\Yaml;
 
 class FileDataProvider implements DataProviderInterface
 {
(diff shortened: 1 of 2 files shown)


Luceos:
Why not support both, instead of introducing a backward breaking change?

dav-is:
I don't see much use case for YAML at all. We're still in beta, breaking changes are going to happen.

Luceos:
@dav-is yes, but you are removing functionality. I know people on the forums have asked about this yaml file and thus have implemented it. Allowing others to freely choose between the two makes more sense than simply removing one because you personally don't like it 😄

issyrocks12:
No reason to make the change if it isn't needed

franzliedke:
Indeed, this is being used. If anything, we should support both.

@dav-is Are you interested in adding JSON **in addtion to** YAML? If not, we will have to close this.

franzliedke:
Closing. Feel free to open a new PR that adds support for JSON in addition to YAML (probably by inspecting the filename).

luceos:
Just found out that JSON is already supported, see:

https://github.com/flarum/core/pull/1193

😕

dav-is:
I wonder how that happened lol

luceos:
@dav-is 🦄 and 🌈

PR #1162, opened 2017-04-19
Bug uploading PNG images as avatars with transparent backgrounds

1337god:
"PNG images with canvas size 100x100 pixels (avatars used on Flarum) usually are smaller than JPG images (even into PNG with transparency). So my advice is let users use JPG or PNG format, because the size reduction using JPG encoding with quality 100 don't exists, it simply increase final size of avatars. If i have to choose one format, i simply use PNG always and every simple problem disappear."

["Bug uploading PNG images as avatars with transparent backgrounds" On Flarum](https://discuss.flarum.org/d/5522-bug-uploading-png-images-as-avatars-with-transparent-backgrounds/23)

### DIFF
--- a/src/Core/Command/UploadAvatarHandler.php
+++ b/src/Core/Command/UploadAvatarHandler.php
@@ -101,7 +101,7 @@ class UploadAvatarHandler
             $manager = new ImageManager;
 
             // Explicitly tell Intervention to encode the image as JSON (instead of having to guess from the extension)
-            $encodedImage = $manager->make($tmpFile)->orientate()->fit(100, 100)->encode('jpg', 100);
+            $encodedImage = $manager->make($tmpFile)->orientate()->fit(100, 100)->encode('png', 100);
             file_put_contents($tmpFile, $encodedImage);
 
             $this->events->fire(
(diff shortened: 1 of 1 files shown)


issyrocks12:
What about [`RegisterUserHandler`](https://github.com/flarum/core/blob/master/src/Core/Command/RegisterUserHandler.php#L176)?

1337god:
@issyrocks12 I had forgot about it sorry! Updating right now!

Luceos:
see #1167

Luceos:
@1337god can you please combine both commits into one PR. I'm going to close this so your PR will be merged in a timely manner, because then only one PR has to be reviewed. I think the change is good enough, you just need to do this in one commit or one PR. Is that possible?

1337god:
Ok done sorry, It was now committed in the same branch as this. (1337god:patch-1)

Zeokat:
I did yesterday a new PR that resolves PNG issues and another little problema with EXIF data was solved. Both issues can be resolved at once since they are related to the same code line.

New PR: https://github.com/flarum/core/pull/1168

PR #1167, opened 2017-04-20
Bug uploading PNG images as avatars with transparent backgrounds

1337god:
"PNG images with canvas size 100x100 pixels (avatars used on Flarum) usually are smaller than JPG images (even into PNG with transparency). So my advice is let users use JPG or PNG format, because the size reduction using JPG encoding with quality 100 don't exists, it simply increase final size of avatars. If i have to choose one format, i simply use PNG always and every simple problem disappear."

Part 2 of https://github.com/flarum/core/pull/1162

### DIFF
--- a/src/Core/Command/RegisterUserHandler.php
+++ b/src/Core/Command/RegisterUserHandler.php
@@ -173,7 +173,7 @@ class RegisterUserHandler
             'target' => $this->uploadDir,
         ]);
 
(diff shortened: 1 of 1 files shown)


issyrocks12:
FYI you don't have to open a new pull request, you can just commit on your old pull request

1337god:
Still getting used to github sorry.

Luceos:
See https://github.com/flarum/core/pull/1162#issuecomment-296595928 please

Issue #1163, opened 2017-04-19
Problems uploading PNG images as avatars with transparent backgrounds


Zeokat:
## Bug report
- Version of Flarum: v0.1.0-beta.6
- Website URL where the bug is visible: https://discuss.flarum.org/u/Zeokat
- The webserver you are running: apache, nginx or something else
- PHP version: PHP 5 or PHP 7, error always appear.
- Hosted environment: local server and VPS
- Hosting provider: DigitalOcean

## Flarum info

```
zeokat@ubuntu:/var/www/html/flarum$ php flarum info
Flarum core 0.1.0-beta.6
PHP 7.0.15-0ubuntu0.16.04.4
Loaded extensions: Core, date, libxml, openssl, pcre, zlib, filter, hash, pcntl, Reflection, SPL, session, standard, mysqlnd, PDO, xml, calendar, ctype, curl, dom, mbstring, fileinfo, ftp, gd, gettext, iconv, intl, json, exif, mcrypt, mysqli, pdo_mysql, Phar, posix, readline, shmop, SimpleXML, sockets, sysvmsg, sysvsem, sysvshm, tokenizer, wddx, xmlreader, xmlwriter, xsl, Zend OPcache
EXT flarum-approval v0.1.0-beta.6
EXT flarum-bbcode v0.1.0-beta.5
EXT flarum-emoji v0.1.0-beta.6
EXT flarum-english v0.1.0-beta.6
EXT flarum-flags v0.1.0-beta.6
EXT flarum-likes v0.1.0-beta.6
EXT flarum-lock v0.1.0-beta.6
EXT flarum-markdown v0.1.0-beta.5
EXT flarum-mentions v0.1.0-beta.6
EXT flarum-sticky v0.1.0-beta.6
EXT flarum-subscriptions v0.1.0-beta.6
EXT flarum-suspend v0.1.0-beta.6
EXT flarum-tags v0.1.0-beta.7
Base URL: http://192.168.247.128/flarum
Installation path: /var/www/html/flarum
```

## Additional comments
The full explanation about what happens and how solve it was disccussed at https://discuss.flarum.org/d/5522-bug-uploading-png-images-as-avatars-with-transparent-backgrounds/23

## Log files

```
N/A
```

franzliedke:
Closed in https://github.com/flarum/core/commit/7837fff10762f25886cfa83a8b1bb9c0626f84c5.

Issue #1161, opened 2017-04-19
Upload profile image error - php exif extension


kulga:
## Bug report
- Version of Flarum: dev-master
- Website URL where the bug is visible: https://devflarum.xyzz.work
- The webserver you are running: nginx
- PHP version: 7.0.17
- Hosted environment: vps

## Flarum info

```
Flarum core 0.1.0-beta.6
PHP 7.0.17
Loaded extensions: Core, date, libxml, openssl, pcre, sqlite3, zlib, bz2, ctype, curl, dom, fileinfo, filter, gd, gettext, hash, iconv, json, mbstring, mcrypt, SPL, pcntl, mysqli, PDO, pdo_mysql, pdo_sqlite, session, posix, recode, Reflection, standard, SimpleXML, sockets, Phar, tokenizer, xml, xmlreader, xmlwriter, zip, Zend OPcache
```

Uploading a image (I presume with exif data - [example](https://www.iconfinder.com/icons/1531895/broken_scare_skull_icon#size=128)) as a profile image results in below error.
Compiling php with exif support (`--enable-exif`) fixes this.

Ideal fixes:
* Documentation should report the need for exif to be enabled
* Check for exif support and only use it if enabled

## Log files

```
POST https://devflarum.xyzz.work/api/users/3/avatar

{
  "errors": [
    {
      "code": 500,
      "title": "Internal server error",
[...]

PR #1168, opened 2017-04-25
Support PNG avatars with transparent backgrounds and fix EXIF rotation

Zeokat:
With this PR i try to solve the issues:

+ [1161](https://github.com/flarum/core/issues/1161): Upload profile image error - php exif extension
+ [1163](https://github.com/flarum/core/issues/1163): Problems uploading PNG images as avatars with transparent backgrounds

I decided to solve both issues with one PR, because both are related.

### DIFF
--- a/src/Core/Command/UploadAvatarHandler.php
+++ b/src/Core/Command/UploadAvatarHandler.php
@@ -101,7 +101,13 @@ class UploadAvatarHandler
             $manager = new ImageManager;
 
             // Explicitly tell Intervention to encode the image as PNG (instead of having to guess from the extension)
-            $encodedImage = $manager->make($tmpFile)->fit(100, 100)->encode('png', 100);
+            // Read exif data to orientate avatar only if EXIF extension is enabled
+            if (extension_loaded('exif')) {
+                $encodedImage = $manager->make($tmpFile)->orientate()->fit(100, 100)->encode('png', 100);
+            }
+            else {
+                $encodedImage = $manager->make($tmpFile)->fit(100, 100)->encode('png', 100);
+            }
             file_put_contents($tmpFile, $encodedImage);
 
             $this->events->fire(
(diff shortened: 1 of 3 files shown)


Luceos:
Thanks for the cross-reference. 👍

src/Core/Command/UploadAvatarHandler.php
```diff
@@ -101,7 +101,7 @@ public function handle(UploadAvatar $command)
             $manager = new ImageManager;
 
             // Explicitly tell Intervention to encode the image as JSON (instead of having to guess from the extension)
-            $encodedImage = $manager->make($tmpFile)->orientate()->fit(100, 100)->encode('jpg', 100);
+            $encodedImage = $manager->make($tmpFile)->fit(100, 100)->encode('png', 100);
```
franzliedke: Can you fix the comment in the line above? That says JSON and probably meant JPEG. Plus now it's PNG ;)

franzliedke:
Also, please check for EXIF being available and call `orientate()` if so, as @kulga suggested in https://github.com/flarum/core/issues/1161#issue-222779188.

Zeokat:
I can't understand clearly your last message @franzliedke . I already checked usage of `orientate()` into a server with a PHP-EXIF module enabled and no errors throwed on screen and `/storage/logs/` remains empty.

datitisev:
@Zeokat Franz is saying that you deleted the `orientate()` method call, and that you should add it back for it to execute only if the server has EXIF available.

screen shot 2017-04-29 at 10 00 19 am

Zeokat:
Oh i see @datitisev , but honestly... code that stuff isn't so trivial becasue we not only need to check if `EXIF` is enanled. On windows servers for example, it also requires `mbstring`. So it's not only check if EXIF is enabled, is check Operating System and then check modules required. Two checks for something useless like the usage of `orientate`... has no sense.

As i said into the issue opened by Kuga https://github.com/flarum/core/issues/1161 , has no sense use `orientate`, we need avatars, we don't need read EXIF to rotate the image.

franzliedke:
@Zeokat mbstring is already part of our [System Requirements](http://flarum.org/docs/installation/#system-requirements).

It's not useless, that we can offer (because we already did) when the server supports it. Would be nice if you could add it in.

Zeokat:
Ok @franzliedke i will try to implement that.

franzliedke:
Thanks for your patience!

Issue #983, opened 2016-06-14
Highlight Post When Linked


dav-is:
It's often confusing when someone links a post which post they were referring to. Maybe it should highlight the linked post in green or something on the first page loaded (then fade away?). Links often don't line up with the post that was link too (I'm sure there's an issue for that somewhere).

datitisev:
@Nayan-Das Feel free to make a PR whenever, just be aware that it may be closed if the solution is not accepted or may require changes. No need to ask, just don't create duplicate PRs 🙂.

PR #1170, opened 2017-04-27
Highlight post when linked

bkolobara:
Hi. I'm getting familiar with the flarum codebase and was thinking this is a good first issue (#983) I can pick up. When a post is linked, it is highlighted in orange for 4s. I think this color works well for both, bright and dark theme. I also removed some "dead code" while I was there.

Would love to get some feedback on it.

### DIFF
--- a/less/forum/PostStream.less
+++ b/less/forum/PostStream.less
@@ -63,18 +63,6 @@
   }
 }
 
-// .item.highlight .post {
-//   &:before {
-//     content: "";
-//     position: absolute;
-//     left: -30px;
-//     top: -5px;
-//     bottom: -5px;
-//     width: 5px;
-//     border-radius: @border-radius;
-//     background: @fl-primary-color;
-//   }
-// }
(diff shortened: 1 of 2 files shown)


dav-is:
Maybe you could add a gif that shows this change's effect?

datitisev:
In my opinion, that doesn't fit Flarum very well, as orange is usually used for important things. I also think it should have some padding on the sides and some border radius.
I think that orange may be too dark, maybe the previous light blue color could be used for the background, but with less opacity.

dav-is:
Why not reuse the animation when hovering over replies?

franzliedke:
Agreed about the reuse suggestion.

For the concrete design (for both), we should probably wait for input from @tobscure.

franzliedke:
That said, thanks for taking this on, @bkolobara! And :clap: for removing the commented code there.

tobscure:
This will be taken care of as part of the updated design that is scheduled for an upcoming release. Thanks for your input though :)

PR #1171, opened 2017-04-29
Allow override layout template.

liverbool:
| Q               | A
| --------------- | ---
| Bug fix?        | no|
| New feature?    | yes |
| BC breaks?      | no |
| License         | MIT |

Please accept this PR ;)

### DIFF
--- a/src/Http/WebApp/AbstractWebApp.php
+++ b/src/Http/WebApp/AbstractWebApp.php
@@ -86,7 +86,7 @@ abstract class AbstractWebApp
      */
     protected function getLayout()
     {
-        return __DIR__.'/../../../views/'.$this->getName().'.blade.php';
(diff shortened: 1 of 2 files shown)


franzliedke:
@liverbool Please provide a little more explanation. :)

franzliedke:
@luceos Since you upvoted... can you explain the nuances here? :see_no_evil:

luceos:
@franzliedke instead of using a path for loading the blade, @liverbool correctly uses the namespace defined (eg for [forum](https://github.com/flarum/core/blob/master/src/Forum/ForumServiceProvider.php#L81)). This allows an extension to override the namespace. A path reference cannot be overridden, hence my upvote.

franzliedke:
Turns out this was already taken care of by Toby. :see_no_evil:

PR #1178, opened 2017-05-08
Update HandleErrors.php

Luceos:
@franzliedke forgot to make variables available to the method, just triggered this but got a warning that all three variables are undefined.


Luceos:
You can reproduce such an error by creating a malformed yaml locale file ;)

PR #1181, opened 2017-05-12
Better 404.html

1337god:
I've made a better 404 page for you guys.

### DIFF
--- a/error/404.html
+++ b/error/404.html
@@ -2,12 +2,17 @@
 <html>
 <head lang="en">
   <meta charset="UTF-8">
-  <title></title>
+  <meta http-equiv="X-UA-Compatible" content="IE=edge, chrome=1">
+  <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1">
+  <title>404 Not Found</title>
 </head>
 <body>
 
-<h1>404 Not Found</h1>
-<p>Looks like this page could not be found.</p>
+<h1>Error 404</h1>
+<p>We are looking for your requested page: <code id="uri"></code> but it seems that it doesn’t exist or it has been moved. Sorry!</p>
 
+<script>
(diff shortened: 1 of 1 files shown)


error/404.html
```diff
@@ -2,12 +2,17 @@
 <html>
 <head lang="en">
   <meta charset="UTF-8">
-  <title></title>
+  <meta http-equiv="X-UA-Compatible" content="IE=edge, chrome=1">
```
franzliedke: Do we really need this here?

error/404.html
```diff
@@ -2,12 +2,17 @@
...
 </head>
 <body>
 
-<h1>404 Not Found</h1>
-<p>Looks like this page could not be found.</p>
+<h1>Error 404</h1>
+<p>We are looking for your requested page: <code id="uri"></code> but it seems that it doesn’t exist or it has been moved. Sorry!</p>
 
+<script>
+	document.getElementById("uri").innerHTML = "/" + (location.pathname+location.search).substr(1);
```
franzliedke: I would prefer for this URL to be generated on the server side. Our [`HandleErrors` middleware](https://github.com/flarum/core/blob/7837fff10762f25886cfa83a8b1bb9c0626f84c5/src/Http/Middleware/HandleErrors.php) could replace some simple strings such as `{{url}}` in the HTML that it reads from the templates.

franzliedke:
Thanks for this suggestion! I like the idea. It would be nice if you could replace the JavaScript with some server-side code, though. :)

error/404.html
1337god: because I will still make a better template, that's why I put it there.
Example: https://thevidme.club/fgdgdg

franzliedke: But for which feature is this needed?

tobscure:
Our error pages have since been redone in b7c1cc5ceffb36bc990919bd749050f49f39554b. Thanks anyway for your PR!

PR #1182, opened 2017-05-14
Add support for slug transliteration and also add support for unicode characters on slugs

Zeokat:
This code can solve issue https://github.com/flarum/core/issues/194

New issue added as it is commented into the source:

>      * TODO: German - GERMANY (de_DE), Danish - DENMARK (da_DK), Serbian - SERBIA (sr_RS)
>      * This languages needs special replacements but will be pending until we have a function
>      * to get forum language (locale used with country code maybe).

### DIFF
--- a/src/Util/Str.php
+++ b/src/Util/Str.php
@@ -1,9 +1,10 @@
 <?php
 
-/*
+/**
  * This file is part of Flarum.
+ * Usage: $slug = Str::slug( 'Cómò ünç 中文百强网' , md5(uniqid(rand(), true)) );
  *
- * (c) Toby Zerner <toby.zerner@gmail.com>
+ * @author: Zeokat 
  *
  * For the full copyright and license information, please view the LICENSE
  * file that was distributed with this source code.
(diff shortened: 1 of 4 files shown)


src/Util/Str.php
```diff
@@ -1,9 +1,10 @@
 <?php
 
-/*
+/**
  * This file is part of Flarum.
+ * Usage: $slug = Str::slug( 'Cómò ünç 中文百强网' , md5(uniqid(rand(), true)) );
  *
- * (c) Toby Zerner <toby.zerner@gmail.com>
+ * @author: Zeokat
```
datitisev: This should remain as Toby. You didn't create the file, you contributed to it. Every single file that's part of Flarum is Toby's property. You just helped make it better. 😉 
Also, is not a documentation comment. It's the header that is in every single PHP file on Flarum/core, so it shouldn't have the usage either.
I'm pretty sure the usage would go in the method as
```php
 * @example
 * <EXAMPLE>
```

Zeokat:
Yeah @Luceos i'm new with GitHub PR. Also the file needs some edits. Is possible delete this PR? (i will try to do a new one with proper edits)

datitisev:
@Zeokat If you push more commits into the branch you are PRing, they will be added to the PR. That's why it's usually a good idea to create new ones from master.

Zeokat:
Yes @Luceos and @datitisev i'm trying to learn how to do it, i'm not a PHP developer. This stuff is new to me 😕

franzliedke:
Thank you, @Zeokat. If you have any questions, we can help you out in our Gitter channels.

PR #1188, opened 2017-05-17
#1184 Fix /api/posts returning 500

### LINKED ISSUE #1184: GET /api/posts returns 500
## Bug report
- Version of Flarum: `dev-master`, `v0.1.0-beta.6`, `v0.1.0-beta.5`, maybe earlier
- Website URL where the bug is visible: https://discuss.flarum.org/api/posts

## Flarum info

_(currently errors for me)_

## Additional comments

Flarum is in latest `dev-master`, even though it's in `workbench/`.
I don't think it has been reported, as I haven't been able to find it. I thought it had been, tho.
Sorry about opening an issue so late. Originally reported at https://discuss.flarum.org/d/2930

## Log files

```php
Uncaught TypeError: Argument 2 passed to Flarum\Api\Controller\ListPostsController::applyFilters() must be of the type array, null given, called in Flarum/workbench/flarum-core/src/Api/Controller/ListPostsController.php on line 117 and defined in Flarum/workbench/flarum-core/src/Api/Controller/ListPostsController.php:132
Stack trace:
#0 Flarum/workbench/flarum-core/src/Api/Controller/ListPostsController.php(117): Flarum\Api\Controller\ListPostsController->applyFilters(Object(Illuminate\Database\Eloquent\Builder), NULL)
#1 Flarum/workbench/flarum-core/src/Api/Controller/ListPostsController.php(69): Flarum\Api\Controller\ListPostsController->getPostIds(Object(Zend\Stratigility\Http\Request))
#2 Flarum/workbench/flarum-core/src/Api/Controller/AbstractSerializeController.p in Flarum/workbench/flarum-core/src/Api/Controller/ListPostsController.php on line 132
```

datitisev:
This makes sure `$filter` is an array even if there is no `filter` property in the url query.
Not sure if we want to use a different solution.
Fixes #1184 

### Changelog
- [Fixed] ListPostsController::applyFilters not receiving array if  `$filter` argument is null

---

_PS: Franz, I somehow remembered `?:` is a thing [after all this time](https://github.com/flarum/core/pull/1105#discussion_r96127460) 😛_

### DIFF
--- a/src/Api/Controller/ListPostsController.php
+++ b/src/Api/Controller/ListPostsController.php
@@ -107,7 +107,7 @@ class ListPostsController extends AbstractCollectionController
      */
     private function getPostIds(ServerRequestInterface $request)
     {
-        $filter = $this->extractFilter($request);
+        $filter = $this->extractFilter($request) ?: array();
         $sort = $this->extractSort($request);
         $limit = $this->extractLimit($request);
         $offset = $this->extractOffset($request);
--- a/src/Api/Controller/ListPostsController.php
+++ b/src/Api/Controller/ListPostsController.php
@@ -107,7 +107,7 @@ class ListPostsController extends AbstractCollectionController
      */
     private function getPostIds(ServerRequestInterface $request)
     {
-        $filter = $this->extractFilter($request) ?: array();
+        $filter = $this->extractFilter($request) ?: [];
         $sort = $this->extractSort($request);
         $limit = $this->extractLimit($request);
         $offset = $this->extractOffset($request);
--- a/src/Api/Controller/ListPostsController.php
+++ b/src/Api/Controller/ListPostsController.php
@@ -107,7 +107,7 @@ class ListPostsController extends AbstractCollectionController
      */
     private function getPostIds(ServerRequestInterface $request)
     {
-        $filter = $this->extractFilter($request) ?: [];
+        $filter = $this->extractFilter($request);
         $sort = $this->extractSort($request);
         $limit = $this->extractLimit($request);
         $offset = $this->extractOffset($request);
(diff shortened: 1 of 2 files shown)


datitisev:
Oh, whoops! Lemme fix Style CI.... Huh, I thought it errored me when I used `array()`. If it doesn't error now, I'll change it.

src/Api/Controller/ListPostsController.php
```diff
@@ -107,7 +107,7 @@ protected function extractOffset(ServerRequestInterface $request)
      */
     private function getPostIds(ServerRequestInterface $request)
     {
-        $filter = $this->extractFilter($request);
+        $filter = $this->extractFilter($request) ?: [];
```
Luceos: This is the wrong place to fix this. If you look at the `extractFilter` method the phpdoc even mentions returning an array. You should do this check over [there](https://github.com/datitisev/core/blob/4ed4067c3f01e646dddcbfe71010c476a21da10a/src/Api/Controller/AbstractSerializeController.php#L189).

Makes more sense, if you ask me 👍

franzliedke:
Hmm, what about the other calls to `extractFilter()` in this file? Also, the return type documentation of `extractFilter()` is currently incorrect... maybe we should change this to always return an array in the method itself? Can you check how this would affect other places where this method is used?

datitisev:
Should be fixed now.
`extractFilter` is used in [two other files](https://github.com/flarum/core/search?utf8=✓&q=extractFilter&type=), but both of them use `array_get` with the filter (or null) as first argument, so I'm guessing PHP just changes null to an array there.

src/Api/Controller/ListPostsController.php
datitisev: Done! :)

Luceos:
I hope your undo doesn't show up on the history @datitisev .

@franzliedke my review already pointed that out, David fixed it at the root cause now.

Approving as is.

franzliedke:
Great, thank you very much!

datitisev:
@Luceos It won't, if you look at "Files changed" all it shows is the end result of the changes. :D 
@franzliedke No problem :)

franzliedke:
Well, it also showed 4 commits. I squash-merged them, though, so it will be just one commit in our master branch. :)

PR #1189, opened 2017-05-18
Update .gitignore to exclude files for distribution

### LINKED ISSUE #762: update .gitattributes exluding files for distribution
This makes the distribution version smaller.

When using `composer install --prefer-dist` the distribution file is used - a zip generated. This archive file will automatically not have the files excluded by using export-ignore in the .gitattributes file.

datitisev:
This adds some files that are only needed for contributing and the `tests` folder to `.gitattributes` to be ignored on Composer's ZIP distribution.
I added these because I don't think there's any reason to install Flarum via ZIP with composer if you are going to contribute. I wasn't sure if I should include `stubs`.
Happy to discuss, tho :)
Closes #762.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,5 +2,10 @@
 .gitignore export-ignore
 .gitmodules export-ignore
 .travis.yml export-ignore
+.editorconfig export-ignore
+.styleci.yml export-ignore
+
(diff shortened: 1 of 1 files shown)


franzliedke:
Nice, thank you!

PR #1193, opened 2017-05-26
Allow JSON to be used for Install Command

dav-is:
This will check if valid JSON is provided and if so, use JSON, otherwise it will try to parse as a YAML file.


tobscure:
Thanks!

PR #1195, opened 2017-06-12
Prevent yourself from locking yourself out of admin group

### LINKED ISSUE #736: API: Prevent from locking yourself out of admin group
Hey,

just a little improvment idea for the API.
You cannot remove the Admin-Group from yourself using the website, but you can by using the API. 
So you can lock yourself out of your Flarum-Instance.

datitisev:
This throws a `PermissionDeniedException` when updating a user via `/api/users/ID` if
- the actor is admin
- they are updating their own user
- the admin group is not in the relationships

It prevents the admin from locking themselves out accidentally.
I'm sure there's a better way to check for the group in the body other than using a filter, but I don't know what else can be used.

Closes #736

---

_Success:_
img

_Error changing group_
img

_Error removing all groups_
img

### DIFF
--- a/src/Core/Listener/AdminGroupLockingPrevention.php
+++ b/src/Core/Listener/AdminGroupLockingPrevention.php
@@ -36,15 +36,15 @@ class AdminGroupLockingPrevention
         $groups = array_get($event->data, 'relationships.groups.data');
 
         // Prevent an admin from removing their admin permission via the API
-        if ($actor->id === $user->id && $actor->isAdmin() && ! $user->isAdmin()) {
-            $userGroups = $user->groups()->get(['group_id'])->all();
+        if ($groups && $actor->id === $user->id && $actor->isAdmin()) {
+            $adminInGroups = array_filter($groups, function ($group) {
+                return $group['id'] == 1;
+            });
+            $keepsAdmin = count($adminInGroups) > 0;
 
-            $newGroups = array_map(function ($group) {
-                return $group->group_id;
-            }, $userGroups);
-            $newGroups[] = Group::ADMINISTRATOR_ID;
-
-            $user->groups()->sync($newGroups);
+            if (! $keepsAdmin) {
+                throw new PermissionDeniedException;
+            }
         }
     }
 }
--- a/src/Core/Listener/AdminGroupLockingPrevention.php
+++ b/src/Core/Listener/AdminGroupLockingPrevention.php
@@ -36,15 +36,15 @@ class AdminGroupLockingPrevention
         $groups = array_get($event->data, 'relationships.groups.data');
 
         // Prevent an admin from removing their admin permission via the API
-        if ($groups && $actor->id === $user->id && $actor->isAdmin()) {
-            $adminInGroups = array_filter($groups, function ($group) {
-                return $group['id'] == 1;
-            });
-            $keepsAdmin = count($adminInGroups) > 0;
+        if ($actor->id === $user->id && $actor->isAdmin() && ! $user->isAdmin()) {
+            $userGroups = $user->groups()->get(['group_id'])->all();
 
-            if (! $keepsAdmin) {
-                throw new PermissionDeniedException;
-            }
+            $newGroups = array_map(function ($group) {
+                return $group->group_id;
+            }, $userGroups);
+            $newGroups[] = Group::ADMINISTRATOR_ID;
+
+            $user->groups()->sync($newGroups);
         }
     }
 }
--- a/src/Core/Listener/AdminGroupLockingPrevention.php
+++ b/src/Core/Listener/AdminGroupLockingPrevention.php
@@ -1,0 +1,52 @@
+<?php
+
+/*
+ * This file is part of Flarum.
+ *
+ * (c) Toby Zerner <toby.zerner@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Flarum\Core\Listener;
+
+use Flarum\Core\Exception\PermissionDeniedException;
+use Flarum\Core\Group;
+use Flarum\Event\UserGroupsWereChanged;
+use Illuminate\Contracts\Events\Dispatcher;
+
+class AdminGroupLockingPrevention
+{
+    /**
+     * @param Dispatcher $events
+     */
+    public function subscribe(Dispatcher $events)
+    {
+        $events->listen(UserGroupsWereChanged::class, [$this, 'whenUserGroupsWereChanged']);
+    }
+
+    /**
+     * @param UserGroupsWereChanged $event
+     * @throws PermissionDeniedException
+     */
+    public function whenUserGroupsWereChanged(UserGroupsWereChanged $event)
+    {
+        if (!$event->actor) return;
+
+        $actor = $event->actor;
+        $user = $event->user;
+
+        // Prevent an admin from removing their admin permission via the API
+        if ($actor->id === $user->id && $actor->isAdmin() && !$user->isAdmin()) {
+            $userGroups = $user->groups()->get(['group_id'])->all();
+
+            $newGroups = array_map(function ($group) {
+                return $group->group_id;
+            }, $userGroups);
+            $newGroups[] = Group::ADMINISTRATOR_ID;
+
+            $user->groups()->sync($newGroups);
+        }
+    }
+}
--- a/src/Core/Listener/AdminGroupLockingPrevention.php
+++ b/src/Core/Listener/AdminGroupLockingPrevention.php
@@ -12,6 +12,7 @@
 namespace Flarum\Core\Listener;
 
 use Flarum\Core\Exception\PermissionDeniedException;
+use Flarum\Core\Group;
 use Flarum\Event\UserWillBeSaved;
 use Illuminate\Contracts\Events\Dispatcher;
 
@@ -38,7 +39,7 @@ class AdminGroupLockingPrevention
         // Prevent an admin from removing their admin permission via the API
         if ($groups && $actor->id === $user->id && $actor->isAdmin()) {
             $adminInGroups = array_filter($groups, function ($group) {
-                return $group['id'] == 1;
+                return $group['id'] == Group::ADMINISTRATOR_ID;
             });
             $keepsAdmin = count($adminInGroups) > 0;
 
--- a/src/Core/Listener/AdminGroupLockingPrevention.php
+++ b/src/Core/Listener/AdminGroupLockingPrevention.php
@@ -12,8 +12,7 @@
 namespace Flarum\Core\Listener;
 
 use Flarum\Core\Exception\PermissionDeniedException;
-use Flarum\Core\Group;
-use Flarum\Event\UserGroupsWereChanged;
+use Flarum\Event\UserWillBeSaved;
 use Illuminate\Contracts\Events\Dispatcher;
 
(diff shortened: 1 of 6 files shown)


Luceos:
Wouldn't it be cleaner to listen to [UserGroupsWereChanged](https://github.com/flarum/core/blob/master/src/Event/UserGroupsWereChanged.php) and reset the Group based on your criteria? This way the check is done no matter the entry point.

datitisev:
Hm.. probably. Where would I listen to the event, though? Not sure in what folder and/or file it would go.

dav-is:
https://github.com/flarum/core/tree/master/src/Core/Listener

franzliedke:
Yes, indeed. This is business logic, and as such does not belong in the controller. @Luceos' suggestion is great. :+1:

datitisev:
@Luceos @franzliedke How does it look now?

I wasn't sure what I should name the file. I'm also not sure if I did the group logic right, as I'm guessing there's an easier way to just get all the group ids in an array.

dav-is:
I did some reading and it appears that the event is thrown before the user is actually saved so maybe that event should be renamed to `UserGroupsWillBeChanged`. If I'm correct, you should be able to simply throw an exception instead of overwriting the changes.

datitisev:
I tried to do so. I tried to throw the Permissions exception, whatever it's called, but the user was still updated in the database, whichever led me to believe i cannot change the event, and rather I have to manually add the group.

Luceos:
> I tried to do so. I tried to throw the Permissions exception, whatever it's called, but the user was still updated in the database, whichever led me to believe i cannot change the event, and rather I have to manually add the group.

Can you explain what you did and how it did not work @datitisev , this might point to a core weakness which should be resolved. I like the suggestion by @dav-is by throwing an exception while that event is triggered.

Ps I think you're picking up an important vulnerability in Flarum, so hence multiple people jump on to this to give direction to the best solution. Great work!

datitisev:
@luceos I basically was just throwing an exception instead of having [all the logic to add the Admin group back](https://github.com/flarum/core/pull/1195/files#diff-7cd3f87003c2cbf37b126369c285e94fR44) to the user.
I may have been doing something wrong, because I was still figuring out stuff, but I'll try it again in a bit.

tobscure:
@dav-is No, the `UserGroupsWereChanged` event is instantiated and stored in the model before the model is saved (line 130), but it is only **dispatched** after the model is successfully saved (line 147).

@datitisev You should listen for the `UserWillBeSaved` event and then validate the contents of `array_get($event->data, 'relationships.groups.data')`.

datitisev:
@tobscure Ok, done! :)

src/Core/Listener/AdminGroupLockingPrevention.php
```diff
@@ -0,0 +1,51 @@
...
+     */
+    public function whenUserWillBeSaved(UserWillBeSaved $event)
+    {
+        $actor = $event->actor;
+        $user = $event->user;
+        $groups = array_get($event->data, 'relationships.groups.data');
+
+        // Prevent an admin from removing their admin permission via the API
+        if ($groups && $actor->id === $user->id && $actor->isAdmin()) {
+            $adminInGroups = array_filter($groups, function ($group) {
```
dav-is: Try using `empty(array_filter(...))` and maybe using more clear variable names?

src/Core/Listener/AdminGroupLockingPrevention.php
```diff
@@ -0,0 +1,51 @@
...
+                return $group['id'] == Group::ADMINISTRATOR_ID;
+            });
+            $keepsAdmin = count($adminInGroups) > 0;
+
+            if (! $keepsAdmin) {
+                throw new PermissionDeniedException;
```
dav-is: Maybe add a message?

src/Core/Listener/AdminGroupLockingPrevention.php
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+namespace Flarum\Core\Listener;
+
+use Flarum\Core\Exception\PermissionDeniedException;
+use Flarum\Core\Group;
+use Flarum\Event\UserWillBeSaved;
+use Illuminate\Contracts\Events\Dispatcher;
+
+class AdminGroupLockingPrevention
```
dav-is: Considering "Prevention" is kind of an abstract noun maybe a better classname would be well suited.

src/Core/Listener/AdminGroupLockingPrevention.php
```diff
@@ -0,0 +1,51 @@
...
+     * @throws PermissionDeniedException
+     */
+    public function whenUserWillBeSaved(UserWillBeSaved $event)
+    {
+        $actor = $event->actor;
+        $user = $event->user;
+        $groups = array_get($event->data, 'relationships.groups.data');
+
+        // Prevent an admin from removing their admin permission via the API
+        if ($groups && $actor->id === $user->id && $actor->isAdmin()) {
```
datitisev: I think for a user to remove himself from all groups `relationships.groups.data` needs to be set to `{}`, as if it's not set to anything it won't change anything..
Correct me if I'm wrong

src/Core/Listener/AdminGroupLockingPrevention.php
datitisev: I tried to name it something short that made _some_ sense, I didn't have any inspiration :P

src/Core/Listener/AdminGroupLockingPrevention.php
dav-is:
```
$groups = [];
if ($groups) {
    echo "true";
} else {
    echo "false";
}
```
Will echo `false`. You should probably experement here.

src/Core/Listener/AdminGroupLockingPrevention.php
dav-is: Do you like the idea of `SelfDemotionGuard`?

datitisev: Yep! I'll change it when I have time.

src/Core/Listener/AdminGroupLockingPrevention.php
datitisev: Anyway, if the user removes himself from all groups, and he's an admin, he should still get denied, shouldn't he?

datitisev: Oh, never mind, I'm just dumb. I see what you mean now :)

src/Core/Listener/AdminGroupLockingPrevention.php
datitisev: I can't, unless I add a new exception & exception handler, as `PermissionDeniedExceptionHandler` only uses the status from the exception itself.

src/Core/Listener/AdminGroupLockingPrevention.php
datitisev: Done!

src/Core/Listener/AdminGroupLockingPrevention.php
datitisev: Done!

src/Core/Listener/AdminGroupLockingPrevention.php
datitisev: Done!

datitisev:
I forgot to push the commit... 🤦‍♂️

src/Core/Listener/SelfDemotionGuard.php
```diff
@@ -0,0 +1,50 @@
...
+     */
+    public function whenUserWillBeSaved(UserWillBeSaved $event)
+    {
+        $actor = $event->actor;
+        $user = $event->user;
+        $groups = array_get($event->data, 'relationships.groups.data');
+
+        // Prevent an admin from removing their admin permission via the API
+        if (isset($groups) && $actor->id === $user->id && $actor->isAdmin()) {
+            $hasGroupWithAdminID = count($groups) > 0 ? ! empty(array_filter($groups, function ($group) {
```
dav-is: I think the `count($groups) > 0` is redundant. If the groups are zero, then the `empty()` will return true. I don't think `array_filter()` is really that costly of an operation especially with an empty array.

src/Core/Listener/SelfDemotionGuard.php
```diff
@@ -0,0 +1,50 @@
...
+     * @param UserWillBeSaved $event
+     * @throws PermissionDeniedException
+     */
+    public function whenUserWillBeSaved(UserWillBeSaved $event)
+    {
+        $actor = $event->actor;
+        $user = $event->user;
+        $groups = array_get($event->data, 'relationships.groups.data');
+
+        // Prevent an admin from removing their admin permission via the API
```
dav-is: You should move this comment into the doc block.

src/Core/Listener/SelfDemotionGuard.php
```diff
@@ -0,0 +1,50 @@
...
+        if (isset($groups) && $actor->id === $user->id && $actor->isAdmin()) {
+            $hasGroupWithAdminID = count($groups) > 0 ? ! empty(array_filter($groups, function ($group) {
+                return $group['id'] == Group::ADMINISTRATOR_ID;
+            })) : false;
+
+            if (! $hasGroupWithAdminID) {
```
dav-is: Instead of using `!` so much just use the inverse. when you define this variable call it `$adminGroupRemoved` and don't use the `!` with `empty()` and here.

datitisev:
@dav-is Done!

luceos:
@datitisev this is how I read the guard:

> It will prevent when __you__ as an __admin__ remove your admin group.

So what happens if someone else removes the admin group on the last admin? There are plenty of reasons this could happen, having extensions can play a major role within Flarum. Maybe this question is outside the scope of this PR 😁

tobscure:
> So what happens if someone else removes the admin group on the last admin?

How could that happen though? The user removing the other user's admin status must be an admin themselves. So as long as you can't un-admin yourself, there will still be at least one admin.

luceos:
@tobscure 

> The user removing the other user's admin status must be an admin themselves.

Are you absolutely sure? I just glimpsed at that code and the [EditUserHandler](https://github.com/flarum/core/blob/master/src/Core/Command/EditUserHandler.php) does run a permission check whether the actor can 'edit' a user, but if "an extension" would incorrectly modify that check the admin would be dropped. Maybe I'm trying to idiot-proof flarum too much 😬

tobscure:
If you have permission to edit a user's groups (via any means), then for all intents and purposes you *are* an admin, because you can promote yourself/others to the admin group.

luceos:
Okay, then disregard my comment. Good thing you're willing to discuss the status quo @tobscure 😸

datitisev:
As the matter of error messages in exceptions was brought up, I could work on that if necessary. 😁 
Anything else this PR needs before it's **squashed** and merged? 😜

franzliedke:
We can squash when merging, which is what I did.

Thanks for the great work. :clap:

franzliedke:
@datitisev I've played around with this a bit and [refactored the listener a bit](https://github.com/flarum/core/commit/8ea13dc8268d7df94970a2b6a7c94e334ae48f0c). This doesn't mean your implementation is bad - I just like these methods flat. :wink:

datitisev:
No problem! 😁

dav-is:
Interesting way of writing that @franzliedke. I think it's a bit more clear. :+1:

PR #1198, opened 2017-06-19
Update Client.php

### LINKED ISSUE #1120: Error handling is bad for controllers parsing internal api calls
## Bug report
- Version of Flarum: current dev-master

---

Pages that parse data by calling on any API endpoint and pushing them into a view for displaying are not handling errors properly with debug on. For instance if any extension produces a 500 internal server error during development, these errors will only produce a white page. The behavior we should expect is that the error is directed towards the view too.

An example of this has happened while assisting @jordanjay29 I've recalled this happened with me before too. A good way to reproduce is create a bad query in any of the listeners that modify the Discussion index (ConfigureDiscussionSearch for instance). The only way to identify the underlying issue is by dumping the `$document` variable in the `getView` method of the Forum `IndexController`.

Feel free to ask for more information if required.

luceos:
Now forwarding exceptions from client to page in case debug mode is on. Fixes #1120.

### DIFF
--- a/src/Api/Client.php
+++ b/src/Api/Client.php
@@ -15,7 +15,6 @@ use Exception;
 use Flarum\Core\User;
 use Flarum\Foundation\Application;
 use Flarum\Http\Controller\ControllerInterface;
-use Illuminate\Contracts\Container\Container;
 use InvalidArgumentException;
 use Zend\Diactoros\ServerRequestFactory;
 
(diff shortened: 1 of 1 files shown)


franzliedke:
Hmm, where is this instantiated? Can't we pass in a parameter so that the client does not have to decide when to raise exceptions on its own?

Luceos:
It's resolved through ioc, because it simply proxies requests to the API controllers in question. Nothing in the construct requires any configuration.

tobscure:
I'm happy with this for now :)

PR #1206, opened 2017-07-06
Search title & content in discussions

datitisev:
This allows searching for query in titles as well as content.
I don't think the implementation is really that well made, but it's a start.

Changelog:
- Renamed `Fulltext` in `Flarum\Core\Search\Discussion` to `Driver`
- Added title search logic to `MySqlFulltextDriver` and renamed to `MySqlDiscussionDriver`

The reason I included the title search with the fulltext is because of the following. Originally, I tried to create another driver & gambit, which worked great, except it would show no discussions because the first gambit (title) had failed, and I didn't want to modify the search logic.

---

image

image


franzliedke:
Nice work, @datitisev.

Can I ask you a favor? Please send a new MR branching off of the `next-back` branch, which will be merged into master soon after the beta-7 release. In that branch, we started a last big namespace restructuring. For example, we renamed the search driver class to `Flarum\Discussion\Search\Fulltext\MySqlFulltextDriver`.

src/Core/Search/Discussion/Driver/MySqlDiscussionDriver.php
```diff
@@ -9,25 +9,35 @@
...
      */
     public function match($string)
     {
-        $discussionIds = Post::where('type', 'comment')
+        $contentDiscussionIds = Post::where('type', 'comment')
             ->whereRaw('MATCH (`content`) AGAINST (? IN BOOLEAN MODE)', [$string])
             ->orderByRaw('MATCH (`content`) AGAINST (?) DESC', [$string])
             ->lists('discussion_id', 'id');
+        $titleDiscussionIds = Discussion::where('is_approved', 1)
+            ->where('title', 'like', "%$string%")
```
franzliedke: Why not use another fulltext search here? This would require another index on the discussion table's `title` column.

tobscure: I believe `LIKE "%$string%"` is going to be *very* slow with anything more than a few dozen thousand discussions.

datitisev: Not sure I'm getting what you mean.  Should I use the whereRaw here again, then?

franzliedke: Yep, that together with the `MATCH AGAINST` clause (which needs a fulltext index on the column, if I am not mistaken).

datitisev:
Continue discussion in https://github.com/flarum/core/pull/1220

PR #1213, opened 2017-07-28
Discussion page design tweaks

tobscure:
- [x] Remove discussion hero, move title to top of post stream
- [ ] Revise post border style
- [ ] Split up post controls and reactions (Like/Reply)
- [ ] Update tags extension (flarum/flarum-ext-tags#43)


pflstr:
Why do you want to remove the discussion hero? One of the great strengths of flarum is its cleanness, its straight design language without a lot of destraction. This design needs consistency, which you are going to give up.

I hope, I will be able to compensate for this change by css modifications and own extensions to preserve the look  as it is now, but why should this be necessary?

Right now, I am working to use the *Flagrow* extension *User Directory.* Basically it does what I need, unfortunately it lacks a Hero und thus breaks the design of the webpage. I first have to fork the extension to add a Hero, which is a pitty, but I see no other way to handle that problem. I'm therefore really concerned, that you are going to follow the same path with the core of flarum.

I just don't understand the reasoning behind that step.

dav-is:
> preserve the look as it is now, but why should this be necessary?

Flarum isn’t stable yet; consequentially, it is vulnerable to drastic (and sometimes disagreeable) design changes. If there’s a certain look you desire, you can simply create a custom extension (no need to fork). Just list the Flagrow extension as a dependency and add the hero in.

tobscure:
just rebased this on master.

datitisev:
I'd say rebasing is better because of commit history, but you gotta be careful. I think it probs removes commit comments as well, as they are different commits and different commit histories.

franzliedke:
Rebasing is absolutely worth it. :+1:

tobscure:
Closed as this has evolved into more substantial skin changes, which I will PR after the release of beta 8.

PR #1214, opened 2017-07-28
Remove user bio feature

franzliedke:
The feature is very limited in scope, and we hope for community
extensions to take over this feature and make it much better.

**After merge:**
- [ ] Remove translations in flarum-ext-english

@dav-is If I am informed correctly, you may be interested in this. :)


dav-is:
Yep 👍🏻 I’ll add it into my social buttons extension sometime before beta 8

moliata:
Hello, by bio you mean about user? If so, I don't recommend doing that. Every big forums platform have that as a core component. Indeed, I love to write more about myself on my community without having to install a custom extension. You should leave this, regards.

Tyler0130:
I would prefer to have this packaged with Flarum. If that means it's removed from the core and put into a bundled extension, that's cool, but I don't want to have to use a third party extension for something as simple as the user bio feature. While I understand it's limited in scope, I like it's simplicity and would prefer to keep that way and extend it with extensions if I felt the need to add more functionality to it or what have you in the future.

tobscure:
@Tyler0130 I think I agree with you that it should be a simple bundled extension for the reason you've stated 👍  we'll look at doing that soon but for now this PR is just about getting it out of core.

datitisev:
Maybe you could make the extension packaged with Flarum before merging this PR, and having the migration to create the table in the extension only do it if it doesn't exist already, and if so, not do anything.

Could that maybe be an option, or something similar?

franzliedke:
We can leave the column in place for now, and remove it with the first stable release (enough time for anybody who wants to switch to the extension) - just need a ticket as reminder.

franzliedke:
Okay, I removed the destructive migration for now.

We will remove the database column forever for the stable release - I've created [a ticket](https://github.com/flarum/core/issues/1280) as reminder.

Everybody who wants to keep their user's bios, should switch to the extension.
@dav-is I would suggest that you give the field a slightly different name. That way, the updater from beta to stable won't kill your extension's data. In your extension's installation migration, you can check for the existence of the column and, if present, import the data from there.

Tyler0130:
@franzliedke What extension are you referring to? It was my understanding that this would be moved to an official bundled extension, based on @tobscure's comment.

tobscure:
@Tyler0130 I want similar functionality as an official bundled extension in the future, but we have not yet discussed/decided what form this will take and when we will implement it.

matteocontrini:
Why can't the bio column not be deleted with the upgrade? So that you can upgrade and then think about installing an extension for the bio feature,  without losing data?

luceos:
@matteocontrini if I'm not mistaken, that is exactly the case. The bio field is still available in the database to be used (or rather migrated to a new field) by a replacement extension.

matteocontrini:
@luceos It's still available but from what I've understood the field will be removed with the stable release. I think it's ok to not include the field from the schema when creating a new Flarum installation, but for those upgrading from previous versions I would expect it to be left there to avoid data loss.

Tyler0130:
@tobscure I see. I must have misunderstood what you had said then because I was expecting to have a bundled extension before it was removed from the core and I feel that would have been the right time to do so. Personally, I'm not pleased with a feature being removed like this and then have to risk it being removed from the database during an update at some point, all before a bundled extension is even discussed. Not happy about that at all, to be truthful.

dav-is:
@tyler0130 Stable release is a ways away and I doubt the bios will be the only thing succeptable to data loss. Although, there will definetly be a warning before upgrading, so simply back up your database and there’s no data lost 🙌🏻.

Tyler0130:
@dav-is I'm not worried about losing the data, because I know I can back it up. I'm worried about the data being usable without having to develop my own extension just to use a feature that was already working quite well for me.

I'm okay with the feature being removed from the core, as long as there is some kind of bundled extension that's available that can make use of the data before this update is released in a stable release.

If there isn't plans to have a bundled extension before this update is released, that's a problem for me.

tobscure:
Bundled extension will come before stable release.

PR #1218, opened 2017-08-02
Add isEmpty function to ItemList

zinsserzh:
Allow extension developers to check if an ItemList is empty and if so remove it from its parent element.

### DIFF
--- a/js/lib/utils/ItemList.js
+++ b/js/lib/utils/ItemList.js
@@ -27,9 +27,10 @@ export default class ItemList {
    * @public
    */
   isEmpty() {
-    for(var prop in this.items) {
-      if(this.items.hasOwnProperty(prop))
+    for (const i in this.items) {
+      if(this.items.hasOwnProperty(i)) {
         return false;
+      }
     }
 
     return true;
--- a/js/lib/utils/ItemList.js
+++ b/js/lib/utils/ItemList.js
@@ -21,6 +21,21 @@ export default class ItemList {
   }
 
   /**
+   * Check whether the list is empty.
+   *
+   * @returns {boolean}
+   * @public
+   */
+  isEmpty() {
+    for(var prop in this.items) {
+      if(this.items.hasOwnProperty(prop))
+        return false;
+    }
+
+    return true;
+  }
+
+  /**
    * Check whether an item is present in the list.
    *
    * @param key
--- a/js/admin/dist/app.js
+++ b/js/admin/dist/app.js
@@ -23425,14 +23425,25 @@ System.register("flarum/utils/ItemList", [], function (_export, _context) {
         }
 
         /**
-         * Check whether an item is present in the list.
+         * Check whether the list is empty.
          *
-         * @param key
          * @returns {boolean}
+         * @public
          */
 
 
         babelHelpers.createClass(ItemList, [{
+          key: "isEmpty",
+          value: function isEmpty() {
+            for (var i in this.items) {
+              if (this.items.hasOwnProperty(i)) {
+                return false;
+              }
+            }
+
+            return true;
+          }
+        }, {
           key: "has",
           value: function has(key) {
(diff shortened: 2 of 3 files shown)


franzliedke:
Thanks for the pull request!

Please compile the JavaScript in the js/forum and js/admin folders and commit the changes to the compiled code as well.

js/lib/utils/ItemList.js
```diff
@@ -21,6 +21,21 @@ export default class ItemList {
...
 
   /**
+   * Check whether the list is empty.
+   *
+   * @returns {boolean}
+   * @public
+   */
+  isEmpty() {
+    for(var prop in this.items) {
+      if(this.items.hasOwnProperty(prop))
```
dav-is: You’re going to want to use `{}` with if blocks to remain consinstant. Also add a space after if.

js/lib/utils/ItemList.js
```diff
@@ -21,6 +21,21 @@ export default class ItemList {
   }
 
   /**
+   * Check whether the list is empty.
+   *
+   * @returns {boolean}
+   * @public
+   */
+  isEmpty() {
+    for(var prop in this.items) {
```
dav-is: Missing space after for

dav-is: You’ll also want to define `prop` as a `const` instead of a `var`

dav-is: Probaly should use the variable name `i` [for consistency](https://github.com/zinsserzh/core/blob/f6810675437ff4605d991e098d631907b4d2a94d/js/lib/utils/ItemList.js#L109)

zinsserzh:
Just to I understand it correctly. The pull request guideline said

> Do NOT checkin the JavaScript dist files in commits.

and you asked me to compile the JavaScript. Were you asking me to compile the code and include dist/ in the commits?

datitisev:
I think those guidelines are in need of updating, as they were last updated on 2015. So yes, you need to compile the code and commit `dist/`

zinsserzh:
Is there any building guide? I'm not sure if I'm doing it correctly:
run `bower install` under `js/`
run `npm install` under `js/forum/` and `js/admin/`
run `gulp` under `js/forum/` and `js/admin/`

luceos:
Wouldn't it make more sense to adopt something like collect.js in general for the ItemList?

tobscure:
What is the utility of this? Equivalent to `items.toArray().length`? (Not saying we shouldn't have it, just want to clarify)

@luceos Collect doesn't do the same thing, ItemList has quite a specialised function/API.

dav-is:
@tobscure I’d say transforming an item list to an array to check if it’s empty is pretty expensive. Imagine a itemlist with like 300 different font awesome icons. You would need to turn that into an array, which might take a second considering it’s 300 items that probaly have decent sized content. Only to recieve the length of 300 != 0. This implementation would stop processing once it realizes there’s at least one item in the list. 

I thought of a possible alternative to this PR, but I’m not as proficient in js, but is it possible to reassign the index of an item in an object? Like you could have `items` with only `items[0]` which would be the usual but is it possible to have `items[55]` and that be the only item in a list? I think if my understanding is correct that would prohibit `return typeof this.items[0] == undefined` from being an alternative.

To the author of the PR, let me know if I misunderstood anything.

zinsserzh:
Use case: https://github.com/wuethrich44/flarum-ext-sso/pull/2
After I remove elements from an ItemList, I also want to remove the ItemList itself if it empty. I don't need to know the length. I only need to know if it is empty. `items.isEmpty()` looks much better than `items.toArray().length == 0` and isEmpty doesn't need to scan the entire list.

zinsserzh:
hmmm... Merge please?

tobscure:
Merged, thanks!

PR #1226, opened 2017-08-16
Fix oauth controller wrong session method call

epoxa:
This rare error emerges when wrong oauth2state passed to AbstractOAuth2Controller:
Call to undefined method Symfony\Component\HttpFoundation\Session\Session::forget()
Obviously method "remove" meant. It works well after the correction inside my custom extension.


tobscure:
Thanks @epoxa!

PR #1241, opened 2017-09-09
Get Mime type without fileinfo extension.

jabedbd:
Some time it is not able to install or enable the fileinfo extension in shared hosting. This change able to run flarum without fileinfo extension.

### DIFF
--- a/src/Extension/Extension.php
+++ b/src/Extension/Extension.php
@@ -183,7 +183,7 @@ class Extension implements Arrayable
                 if (file_exists($file)) {
                     $mimetype = pathinfo($file, PATHINFO_EXTENSION) === 'svg'
                         ? 'image/svg+xml'
-                        : finfo_file(finfo_open(FILEINFO_MIME_TYPE), $file);
+                        : image_type_to_mime_type(exif_imagetype($file));
                     $data = file_get_contents($file);
 
                     $icon['backgroundImage'] = 'url(\'data:'.$mimetype.';base64,'.base64_encode($data).'\')';


franzliedke:
Hmm, wouldn't that just mean we now require the exif extension instead?

jabedbd:
exif extension mostly installed by default with shared hosting server.

franzliedke:
Are there any stats about enabled extensions on various hosting providers?

jabedbd:
I didn't found any stats about installation of fileinfo and exif. Everything i said from my little experiences as a poor shared hosting user :p 
My second source of research is stackoverflow. I found a lot of issues there with the fileinfo and shared hosting where issues about exif is less (almost zero) :D

AndreiTelteu:
Would be better to check if those functions exists ?
```php
$mimetype = pathinfo($file, PATHINFO_EXTENSION) === 'svg'
    ? 'image/svg+xml'
    : (function_exists('finfo_file')
        ? finfo_file(finfo_open(FILEINFO_MIME_TYPE), $file)
        : image_type_to_mime_type(exif_imagetype($file)));
```
Not this ugly, maybe do a separate method for this.

luceos:
I tried researching the availability of both extensions, but found nothing conclusive. Thus I'd say let's not change anything and possibly introduce unforeseen issues.

I do like the idea of @AndreiTelteu , but then we're checking against several extensions without any fixed dependency, thus also not having the certainty of a fallback.

tldr, I'm saying no to this.

tobscure:
What about a simple switch/case for the file extension? (eg. svg → image/svg+xml, png → image/png, etc)

There's a chance we will refactor this code in the future to actually refer to the image file rather than a `data:` URI, but I'm still unsure about that so happy to discuss a solution to reduce dependency on certain PHP extensions.

luceos:
Why are we even writing a custom method to identify the mimetype. Looking at Symfony:

```
The mime type is guessed using a MimeTypeGuesser instance, which uses finfo(), mimecontenttype() and the system binary "file" (in this order), depending on which of those are available.
```

luceos:
@franzliedke can add your opinion, it feels like we spent too much time on this already at the cost of @jabedbd

franzliedke:
I agree: if we manage to get rid of requiring fileinfo (without replacement with another extension), then I'd be happy to accept such a PR. Toby's suggestion seems simple enough.

@jabedbd Would you like to implement this feature?

jabedbd:
@franzliedke  yes,i like to do it. let me try.

luceos:
Why add a hack switch statement when we can use symfony which is already part of flarum anyway @franzliedke ?

franzliedke:
@luceos If I understood your earlier post correctly, Symfony only falls back to the `file` binary, which isn't exactly what we are doing here (but which should be just as fine, right)?

franzliedke:
@jabedbd Can you push that change (to a simple `switch` statement) or do you want us to do it?

jabedbd:
@franzliedke i'm busy with a project. please do it.

franzliedke:
Done in https://github.com/flarum/core/commit/ba96f311a960314104becc716cd3dc7b5291cc6d.

Wasn't there an issue about our reliance on certain PHP extensions over others. I can't seem to find it, not sure why...

Issue #557, opened 2015-09-29
Add screen name functionality


tobyzerner:
This is a central issue for discussion about allowing other characters in usernames, and the related complications.

Currently we only allow alphanumeric, underscores, and hyphens in usernames (Laravel's alphadash rule). I think it's pretty clear that we need to relax this constraint though since Flarum is already quite popular in non-English environments.

However, one does not simply allow other characters. We need to consider:
- **User profile URLs.** Currently we use the username as the sole identifier in the URL, as in forum.com/u/username. Unicode and spaces _can_ work in URLs, but things get ugly with URL encoding. Is this worth dealing with? Do we want to use the numeric ID in the URL instead of the username? What about making a latin slug from the username, and putting that in the URL in addition to the ID?
- **Mentions.** In particular, whitespace. @JoshyPHP has suggested a syntax like `@"foo bar"`. `@[foo bar]` would also work. What do we think? Any other issues with this?

Related discussion:
http://discuss.flarum.org/d/915-spaces-non-ascii-characters-in-usernames
#465

FirestarterUA:
Very userful library for thats task: https://github.com/jbroadway/urlify

dcsjapan:
Taking your points one at a time:

- Handling of **user profile URLs** should really be discussed in the wider context of slug transliteration and encoding (see #194) as this is something we're going to have to deal with in any event. I'll add a suggestion to that issue when I'm done here.

 > What about making a latin slug from the username, and putting that in the URL ...?

 While this is possible for many languages, it would be very impractical for some, and nigh impossible for others. I'll explain in more detail in the issue linked above; in any case, this should be dealt with as a URL issue rather than as a username issue.

- **Mentions** could certainly be handled as @JoshyPHP suggested. The only problem I see there is that users will tend to expect Flarum's mentions to work just like Twitter.

 Would it be possible to have the delimiters (quotation marks or brackets) inserted automatically by the autocomplete function? That would make the process more transparent to users. (Of course you would have to adjust the autocomplete function so it can recognize unicode usernames and present them as candidates.)

Dakoom:
I think that usernames without spaces is too much "computer style". I would try to find a way to permit the spaces. It could be that when there is a space in a username, in the profile URL you get automaticaly something like forum.com/u/user_name.

kirkbushell:
Better yet, a proper URL, like user-name. I agree that this should be supported.

oldfetch404:
A while back I used a simple URL format:
[...]

PR #1246, opened 2017-09-19
User display names

tobscure:
(I'm going into `master` here because I need to use this imminently for a client project.)

It is not uncommon for forums to be intergrated with sites where users don't have a unique "handle" - they might just have their first name, or a full name, which is not guaranteed to be unique.

This PR introduces the concept of "display names" for users. By default display names are just an alias for usernames, and they cannot be changed in core (we are not introducing a new column into the DB).

But extensions may override this and set them to something different. eg. the client integration I'm working on uses a numeric ID as the `username`, and adds a `display_name` column which is set to a non-unique first name.

The important thing is that all code should use `display_name` whenever intending to output a human-readable name - `username` is reserved for cases where you want to output a unique identifier (which may or may not be human-friendly) - eg. @mentions, user profile URL slugs.

The new "GetDisplayName" API is probably sub-optimal, but I didn't worry too much because we can come up with something better when we bring this into `next-back`.

Just wanting some feedback on the concept and implementation before I merge.

ref #557

### DIFF
--- a/js/forum/dist/app.js
+++ b/js/forum/dist/app.js
@@ -28942,7 +28942,9 @@ System.register('flarum/components/UsersSearchSource', ['flarum/helpers/highligh
             query = query.toLowerCase();
 
             var results = app.store.all('users').filter(function (user) {
-              return user.username().toLowerCase().substr(0, query.length) === query;
+              return [user.username(), user.displayName()].some(function (value) {
+                return value.toLowerCase().substr(0, query.length) === query;
+              });
             });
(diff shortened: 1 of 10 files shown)


tobscure:
Note that this is not the same thing as "screen names" as suggested in https://github.com/flarum/core/issues/557#issuecomment-148813969. A screen name extension which just wants to add an optional field, and leave the username as the primary human-readable identifier, would not need to change the display name.

PR #1252, opened 2017-10-05
Illuminate 5.5 for next-back

luceos:
So here's the work progress for the upgrade to the (Laravel) Illuminate components 5.5 Long Term Support.

---

Trace of work:

- Collection `lists` modified to `pluck`
- Validation exception testing fixed
- Increased minimum constraint of PHP to 7.0, matching Illuminate framework.
- Removed PHP 5.6 and added PHP 7.2 in travis config.
- Implemented Illuminate Translator contract.


tobscure:
Didn't realise 5.5 requires PHP 7... @franzliedke are we OK with this?

Eldenroot:
@tobscure - please check this: http://php.net/supported-versions.php

Flarum should support PHP 7.0 and above. Almost 90% of webhosting offer php 7.0.

franzliedke:
I'm fine with PHP 7.

src/Foundation/Site.php
```diff
@@ -178,8 +178,6 @@ protected function getApp()
             $config->set('mail.username', $settings->get('mail_username'));
             $config->set('mail.password', $settings->get('mail_password'));
 
-            $app->register(\Flarum\BusServiceProvider::class);
-
```
franzliedke: No replacement?

franzliedke:
@luceos Can you take care of [my comment here](https://github.com/flarum/core/commit/ea4d889b76ba03ddf55a77d6333c0a8352b9b854#commitcomment-24825880)? This can only be done when using Laravel 5.5. :)

src/Foundation/Site.php
luceos: This is a WIP, I was busy stripping down flarum/flarum to get it to work with flarum/core 👍

franzliedke: Not sure what you mean, sorry. Does that mean you will revert this before finishing the PR?

luceos:
@franzliedke [this](https://github.com/flarum/core/pull/1252#issuecomment-334926692) `@json` added.

Moving forward with this work.

src/Foundation/Site.php
luceos: Fixed. Re-added the ability to auto forward Commands to their respective Handler classes.

luceos:
This deprecates `Symfony\Component\Translation\TranslatorInterface` in our app. It is no longer bound and needs to be replaced with `Illuminate\Contracts\Translation\Translator`, you can also use `translator`.

src/Foundation/Site.php
luceos: @franzliedke see https://github.com/flarum/core/pull/1252/commits/72c232d5a388479eaf712939ffeeb19bec848a19

luceos:
Used this in my flarum/flarum after cloning that repo's `next-back` branch:

```json
    "require": {
        "flarum/core": "dev-next-back-5.5 as 0.1.0-beta.99",
        "flarum/flarum-ext-english": "dev-next-back"
    },
```

luceos:
@flarum/core This is ready for initial review. It has been tested with:

- flarum/flarum at next-back
- flarum/flarum-ext-english at next-back

Nothing else. I went to some of the steps:

- create user
- create post
- access admin
- enable extension

In general I've noticed nothing strange.

src/Locale/LocaleManager.php
```diff
@@ -11,12 +11,12 @@
 
 namespace Flarum\Locale;
 
-use Symfony\Component\Translation\Translator as SymfonyTranslator;
+use Illuminate\Contracts\Translation\Translator;
 
 class LocaleManager
```
tobscure: LocaleManager::getTranslator docblock is still returning SymfonyTranslator.

AbstractFrontend::addTranslations depends on the SymfonyTranslator::getCatalogue, which is not present in the Illuminate contract.

I think LocaleManager needs to require a Flarum\Locale\Translator instance directly, rather than the Illuminate contract - that way we know it fulfils both the Illuminate and Symfony contracts.

src/Locale/LocaleServiceProvider.php
```diff
@@ -41,19 +40,17 @@ public function register()
...
 
             return $translator;
         });
         $this->app->alias('translator', Translator::class);
-        $this->app->alias('translator', TranslatorInterface::class);
+        $this->app->alias('translator', TranslatorContract::class);
     }
 
-    private function getDefaultLocale()
+    private function getDefaultLocale(): string
```
tobscure: A return type! That tickles 😊

src/Locale/Translator.php
```diff
@@ -28,7 +29,6 @@ public function getCatalogue($locale = null)
         } else {
             $this->assertValidLocale($locale);
         }
-
```
tobscure: ?

src/Post/RegisteredTypesScope.php
```diff
@@ -13,9 +13,9 @@
 
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Database\Eloquent\Model;
-use Illuminate\Database\Eloquent\ScopeInterface;
+use Illuminate\Database\Eloquent\Scope;
 
-class RegisteredTypesScope implements ScopeInterface
+class RegisteredTypesScope implements Scope
```
tobscure: Scopes have changed a bit in 5.2. See https://laravel.com/docs/5.3/upgrade#upgrade-5.2.0 > Eloquent > Global Scopes.

I think we can get rid of the `remove` method, and all of the weird whereIndex/bindingIndex/bindingCount stuff that makes it possible.

Please test post type functionality after doing this to make sure it works.

src/Locale/LocaleServiceProvider.php
luceos: Wasn't really sure, but since we're adopting php 7.0+ and this method really only returns a string it seemed appropriate. Also because it seems to become an industry standard to get of unnecessary phpdoc's completely when type hinting do the same. What's your opinion or preference?

src/Locale/Translator.php
luceos: Most likely happened during debugging as the Translator instance wasn't being resolved correctly.

src/Locale/LocaleServiceProvider.php
tobscure: Yes let's go for it! I only meant that it felt funny, but in a good kind of way 😆

src/Post/RegisteredTypesScope.php
luceos: Done. Tested.

src/Locale/LocaleServiceProvider.php
luceos: Done, patched the whole class.

src/Foundation/Site.php
```diff
@@ -166,6 +167,8 @@ protected function getApp()
         $app->register(ViewServiceProvider::class);
         $app->register(ValidationServiceProvider::class);
 
+        $app->register(BusProvider::class);
```
tobscure: No need for the `BusProvider` alias?

src/Api/Controller/SendConfirmationEmailController.php
```diff
@@ -43,17 +43,17 @@ class SendConfirmationEmailController implements ControllerInterface
...
 
     /**
      * @param \Flarum\Settings\SettingsRepositoryInterface $settings
      * @param Mailer $mailer
      * @param UrlGenerator $url
-     * @param TranslatorInterface $translator
+     * @param Translator $translator
      */
-    public function __construct(SettingsRepositoryInterface $settings, Mailer $mailer, UrlGenerator $url, TranslatorInterface $translator)
+    public function __construct(SettingsRepositoryInterface $settings, Mailer $mailer, UrlGenerator $url, Translator $translator)
```
tobscure: Can we please undo these changes and stick with injecting the Symfony TranslatorInterface. I think we want that to be our primary, implementation of the Illuminate contract is secondary just so the validator still works.

tobscure:
@luceos Symfony 4.0 is out!

src/Foundation/Site.php
luceos: Nope, we load both Illuminate BusServiceProvider and Flarum BusServiceProvider. Or we call the Illuminate from inside our own.

Issue #201, opened 2015-07-29
Show text next to discussion badges


franzliedke:
On the discussion hero, badges like "Locked" or "Pinned" should include their textual meaning. We have enough space...

As discussed on Gitter, because [Nielsen](http://www.nngroup.com/articles/icon-usability/).

tobscure:
What needs to be done:

* Alter the Badge component to display the label prop, wrapped in a `<span className="Badge-label">`.
* Alter Badge.less to hide .Badge-label, except in .DiscussionHero and .UserHero.

MacPrawn:
There's already a tooltip on the badge, however. I suspect that should be removed/hidden also, when the text is displayed? (i.e. in .DiscussionHero and .UserHero)

franzliedke:
We had an initial implementation in PR #1263, but it was never finished and needed the following improvements:
- wrapping `<span>` instead of `<div>`
- styling for the badge text to look like part of the badge
- I'd prefer for the label to come from a prop - that way, no CSS rules are needed for showing them only in certain contexts. If there is a label, let's show it. If we don't want one, we shouldn't pass it in.

klu17:
Is this still open? @franzliedke

franzliedke:
@klu17 As far as I can tell, there was no updated version of #1263. Do you want to give it a shot?

klu17:
@franzliedke Yes :)

SychO9:
Feel free to do so @klu17 !

Checkout https://docs.flarum.org/contributing.html for setting things up, and if you need any help, there is the [community forums](http://discuss.flarum.org/) and [discord chat](http://flarum.org/discord).

PR #1263, opened 2017-10-23
201 - Adding label next to discussion badges

smithandre:
Adding labels next to the discussion badges, addressing https://github.com/flarum/core/issues/201

### DIFF
--- a/js/lib/components/Badge.js
+++ b/js/lib/components/Badge.js
@@ -29,7 +29,7 @@ export default class Badge extends Component {
           <span {...attrs}>
             {iconName ? icon(iconName, {className: 'Badge-icon'}) : m.trust('&nbsp;')}
           </span>
-          <span className="Badge-label">{' ' + this.props.label + ' '}</span>
+          <span className="Badge-label">{this.props.label}</span>
         </div>
     );
   }
--- a/js/lib/components/Badge.js
+++ b/js/lib/components/Badge.js
@@ -25,9 +25,12 @@ export default class Badge extends Component {
     attrs.title = extract(attrs, 'label') || '';
 
     return (
-      <span {...attrs}>
-        {iconName ? icon(iconName, {className: 'Badge-icon'}) : m.trust('&nbsp;')}
-      </span>
+        <div>
+          <span {...attrs}>
+            {iconName ? icon(iconName, {className: 'Badge-icon'}) : m.trust('&nbsp;')}
+          </span>
+          <span className="Badge-label">{' ' + this.props.label + ' '}</span>
+        </div>
     );
   }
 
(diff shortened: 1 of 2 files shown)


js/lib/components/Badge.js
```diff
@@ -25,9 +25,12 @@ export default class Badge extends Component {
...
     return (
-      <span {...attrs}>
-        {iconName ? icon(iconName, {className: 'Badge-icon'}) : m.trust('&nbsp;')}
-      </span>
+        <div>
+          <span {...attrs}>
+            {iconName ? icon(iconName, {className: 'Badge-icon'}) : m.trust('&nbsp;')}
+          </span>
+          <span className="Badge-label">{this.props.label}</span>
+        </div>
```
franzliedke: Should this wrapping div receive some styling to prevent wrapping between the two spans?

franzliedke: Also, it looks like the indentation changed? Was that on purpose? If not, please revert.

franzliedke:
When you change any of our JS code, please make sure you also recompile and commit the code in the district directory.

franzliedke:
*dist directory

js/lib/components/Badge.js
tobscure: Tthe wrapping tag should be a `span` as well because badges are inline elements.

josephnle:
@franzliedke Just curious and wanted to clarify since I didn't do this for my PR. The [Contribution Guide](https://github.com/flarum/flarum/blob/master/CONTRIBUTING.md#development-setup) tells us not to commit compiled assets. Should we start committing the compiled JS?

franzliedke:
Uhm, looks like we need to update that guide, right, @tobscure. At least that has been our policy for the last couple of months.

Good catch, @josephnle.

Pleaes take care of the requested wrapping `<span>`. If you don't commit the compiled JS, we can do it after the merge.

franzliedke:
Indeed, I was wrong. We take care of the compilation ourselves, in order to minimize conflicts between different pull requests.

@josephnle Still interested in finishing this? Otherwise, we will have to take care of the rest. :smile:

josephnle:
@franzliedke Would be happy to help out and finish this, but this isn't my PR. 😛 Perhaps you meant to tag @smithandre?

franzliedke:
Oh, haha. Yes, that question was for @smithandre. :grinning:

js/lib/components/Badge.js
franzliedke: Do we need to return exactly one object from the `view` method or could we return multiple as well, by the way?

franzliedke:
@tobscure I'll let you finish this. :grin:

franzliedke:
Closing this, because reviews weren't followed up.

PR #1273, opened 2017-10-26
Prevent saving invalid custom less

### LINKED ISSUE #1227: Invalid custom CSS causes forum to 500
## Bug report
- Version of Flarum:  0.1.0-beta.7
- Website URL where the bug is visible: Removed
- The webserver you are running: apache
- PHP version:
```
PHP 7.0.22-0ubuntu0.16.04.1 (cli) ( NTS )
Copyright (c) 1997-2017 The PHP Group
Zend Engine v3.0.0, Copyright (c) 1998-2017 Zend Technologies
    with Zend OPcache v7.0.22-0ubuntu0.16.04.1, Copyright (c) 1999-2017, by Zend Technologies
```
- Hosted environment: vps
- Hosting provider: KVM VPS provided by ourselves

## Flarum info

```
Flarum core 0.1.0-beta.7
PHP 7.0.22-0ubuntu0.16.04.1
Loaded extensions: Core, date, libxml, openssl, pcre, zlib, filter, hash, pcntl, Reflection, SPL, session, standard, mysqlnd, PDO, xml, calendar, ctype, curl, dom, mbstring, fileinfo, ftp, gd, gettext, iconv, json, exif, mysqli, pdo_mysql, Phar, posix, readline, shmop, SimpleXML, sockets, sysvmsg, sysvsem, sysvshm, tokenizer, wddx, xmlreader, xmlwriter, xsl, Zend OPcache
EXT flarum-approval v0.1.0-beta.7
EXT flarum-bbcode v0.1.0-beta.5
EXT flarum-emoji v0.1.0-beta.6
EXT flarum-english v0.1.0-beta.7
EXT flarum-flags v0.1.0-beta.7
EXT flarum-likes v0.1.0-beta.6
EXT flarum-lock v0.1.0-beta.7
EXT flarum-markdown v0.1.0-beta.5
EXT flarum-mentions v0.1.0-beta.7
EXT flarum-sticky v0.1.0-beta.7
EXT flarum-subscriptions v0.1.0-beta.6
EXT flarum-suspend v0.1.0-beta.7
EXT flarum-tags v0.1.0-beta.8
Base URL: <removed>
Installation path: /srv/forum
```

## Additional comments
Forum displays 500 when CSS with @media is present
Removing the offending CSS in the database makes the forum load correctly.

```
update frm_settings set value="" where `key`="custom_less";
```

offending CSS:
```
/*p {
color:#a8897c;
}
 
.Button {
color:#a8897c;
}
 
.IndexPage-newDiscussion {
color:white;
background-color:#a8897c;
}*/
 
body {
/*background-image:url("<removed>");*/
}
 
#app {
background-color:rgba(0,0,0,0);
background-image:url("<removed>");
background-repeat: no-repeat;
background-position: 50% -50px;
background-size: 1600px 100%;
}
 
/*.Button-label, .DiscussionListItem-title, .Dropdown-menu>li>a {
color:black;
}*/
 
@media {
color:black;
}
 
.IndexPage-newDiscussion .Button-label {
color:white;
}
```

## Expected behaviour
* Forum software should not crash
* Setting "debug mode" should display an error (only way to find the error was the apache logs)

### Apache Error log
```
[Wed Aug 16 23:01:17.283172 2017] [:error] [pid 8658] [client <REMOVED>] PHP Fatal error:  Uncaught Error: Call to a member function accept() on null in /srv/forum/vendor/oyejorge/less.php/lib/Less/Visitor.php:37\nStack trace:\n
#0 /srv/forum/vendor/oyejorge/less.php/lib/Less/Tree/Media.php(32): Less_Visitor->visitObj(NULL)\n#1 /srv/forum/vendor/oyejorge/less.php/lib/Less/Visitor.php(28): Less_Tree_Media->accept(Object(Less_Visitor_joinSelector))\n
#2 [internal function]: Less_Visitor->visitObj(Object(Less_Tree_Media))\n
#3 /srv/forum/vendor/oyejorge/less.php/lib/Less/Visitor.php(45): array_map(Array, Array)\n
#4 /srv/forum/vendor/oyejorge/less.php/lib/Less/Tree/Ruleset.php(64): Less_Visitor->visitArray(Array)\n
#5 /srv/forum/vendor/oyejorge/less.php/lib/Less/Visitor.php(28): Less_Tree_Ruleset->accept(Object(Less_Visitor_joinSelector))\n
#6 /srv/forum/vendor/oyejorge/less.php/lib/Less/Visitor/joinSelector.php(17): Less_Visitor->visitObj(Object(Less_Tree_Ruleset))\n
#7 /srv/forum/vendor/oyejorge/less.php/lib/Less/Parser.php(420): Less_Visitor_joinSelector->run(Object(Less_Tre in /srv/forum/vendor/oyejorge/less.php/lib/Less/Visitor.php on line 37
```

clarkwinkelmann:
Fixes #1227

This could be moved to a separate event listener, but I don't know where to put it.

### DIFF
--- a/src/Core/Listener/CheckCustomLessFormat.php
+++ b/src/Core/Listener/CheckCustomLessFormat.php
@@ -21,10 +21,10 @@ class CheckCustomLessFormat
 {
     public function subscribe(Dispatcher $events)
     {
-        $events->listen(PrepareSerializedSetting::class, [$this, 'checkCustomLessFormatBeforeSaving']);
+        $events->listen(PrepareSerializedSetting::class, [$this, 'check']);
     }
 
-    public function checkCustomLessFormatBeforeSaving(PrepareSerializedSetting $event)
+    public function check(PrepareSerializedSetting $event)
     {
         if ($event->key === 'custom_less') {
             $parser = new Less_Parser();
--- a/src/Core/Listener/CheckCustomLessFormat.php
+++ b/src/Core/Listener/CheckCustomLessFormat.php
@@ -1,0 +1,43 @@
+<?php
+
+/*
+ * This file is part of Flarum.
+ *
+ * (c) Toby Zerner <toby.zerner@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Flarum\Core\Listener;
+
+use Flarum\Core\Exception\ValidationException;
+use Flarum\Event\PrepareSerializedSetting;
+use Illuminate\Contracts\Events\Dispatcher;
+use Less_Exception_Parser;
+use Less_Parser;
+
+class CheckCustomLessFormat
+{
+    public function subscribe(Dispatcher $events)
+    {
+        $events->listen(PrepareSerializedSetting::class, [$this, 'checkCustomLessFormatBeforeSaving']);
+    }
+
+    public function checkCustomLessFormatBeforeSaving(PrepareSerializedSetting $event)
+    {
+        if ($event->key === 'custom_less') {
+            $parser = new Less_Parser();
+
+            try {
+                // Check the custom less format before saving
+                // Variables names are not checked, we would have to set them and call getCss() to check them
+                $parser->parse($event->value);
+            } catch (Less_Exception_Parser $e) {
+                throw new ValidationException([
+                    'custom_less' => $e->getMessage(),
+                ]);
+            }
+        }
+    }
+}
--- a/src/Forum/ForumServiceProvider.php
+++ b/src/Forum/ForumServiceProvider.php
@@ -46,7 +46,7 @@ class ForumServiceProvider extends AbstractServiceProvider
     {
         $this->populateRoutes($this->app->make('flarum.forum.routes'));
 
-        $this->loadViewsFrom(__DIR__.'/../../views', 'flarum.forum');
+        $this->loadViewsFrom(__DIR__ . '/../../views', 'flarum.forum');
 
         $this->checkCustomLessFormatBeforeSaving();
 
@@ -157,7 +157,7 @@ class ForumServiceProvider extends AbstractServiceProvider
                     // Check the custom less format before saving
                     // variables names are not checked, we would have to set them and call getCss() to check them
                     $parser->parse($event->value);
-                } catch(Less_Exception_Parser $e) {
+                } catch (Less_Exception_Parser $e) {
                     throw new ValidationException([
                         'custom_less' => $e->getMessage(),
                     ]);
--- a/src/Forum/ForumServiceProvider.php
+++ b/src/Forum/ForumServiceProvider.php
@@ -11,13 +11,17 @@
 
 namespace Flarum\Forum;
 
+use Flarum\Core\Exception\ValidationException;
 use Flarum\Event\ConfigureForumRoutes;
 use Flarum\Event\ExtensionWasDisabled;
 use Flarum\Event\ExtensionWasEnabled;
+use Flarum\Event\PrepareSerializedSetting;
 use Flarum\Event\SettingWasSet;
 use Flarum\Foundation\AbstractServiceProvider;
 use Flarum\Http\Handler\RouteHandlerFactory;
 use Flarum\Http\RouteCollection;
+use Less_Exception_Parser;
+use Less_Parser;
 
 class ForumServiceProvider extends AbstractServiceProvider
 {
@@ -43,6 +47,8 @@ class ForumServiceProvider extends AbstractServiceProvider
         $this->populateRoutes($this->app->make('flarum.forum.routes'));
 
         $this->loadViewsFrom(__DIR__.'/../../views', 'flarum.forum');
+
+        $this->checkCustomLessFormatBeforeSaving();
 
         $this->flushWebAppAssetsWhenThemeChanged();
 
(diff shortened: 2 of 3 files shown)


clarkwinkelmann:
StyleCI fails on something I didn't change in this PR :thinking:

src/Forum/ForumServiceProvider.php
```diff
@@ -141,6 +147,25 @@ protected function populateRoutes(RouteCollection $routes)
...
 
+    protected function checkCustomLessFormatBeforeSaving()
+    {
+        $this->app->make('events')->listen(PrepareSerializedSetting::class, function (PrepareSerializedSetting $event) {
+            if ($event->key === 'custom_less') {
+                $parser = new Less_Parser();
+
+                try {
+                    // Check the custom less format before saving
+                    // variables names are not checked, we would have to set them and call getCss() to check them
```
tobscure: So what happens if you do have core variables in your custom LESS? Does it fail validation?

clarkwinkelmann: The `parse` method does not handle variables at all apparently, it only checks and parse format. So you can input any variable name without an exception being thrown.

If we were to call $parser->getCss() afterwards, that part would try to fill variables and fail if their name is invalid.

Would it be worth it to also load and check variables ? I'm not sure if the part injecting variables has been designed to be reusable

tobscure: I see, that makes sense.

Hmm I think ideally when you save custom LESS it should attempt to recompile the forum CSS straight away, rather than doing it on the next forum page load. That way we can be sure that the custom LESS is valid in the context of all core/extension LESS (and variables).

That will obviously take some bigger refactoring to do though...

clarkwinkelmann: Indeed :grin: 

But it's tricky. If the check is made before saving the custom less, it's not easy to push the new value into the WebApp just for testing. Maybe by pushing another class that implements the `SettingsRepositoryInterface` interface when creating the WebApp ? The value is fetched directly from the settings https://github.com/flarum/core/blob/40ebc13292b64271418aa3d5683d0927e6db778e/src/Forum/WebApp.php#L55

And if the check is done after... I have no idea how to rollback in case of exception :sweat_smile:

src/Forum/ForumServiceProvider.php
```diff
@@ -141,6 +147,25 @@ protected function populateRoutes(RouteCollection $routes)
         );
     }
 
+    protected function checkCustomLessFormatBeforeSaving()
+    {
+        $this->app->make('events')->listen(PrepareSerializedSetting::class, function (PrepareSerializedSetting $event) {
```
franzliedke: Can you move this logic to a dedicated event handler, please? Similar to what we do [here](https://github.com/flarum/core/blob/4f688fc9a25da9bfefe86b6cb418e2ca2a754519/src/User/UserServiceProvider.php#L89).

I feel this service provider is too generic a place to handle this stuff.

clarkwinkelmann:
@franzliedke is it better like this ? It's still not very clean, because unlike the `UserServiceProvider` you showed as example there's not a single place to plug everything to `$this->app->make('events')` in `ForumServiceProvider`.

src/Core/Listener/CheckCustomLessFormat.php
```diff
@@ -0,0 +1,43 @@
...
+use Less_Parser;
+
+class CheckCustomLessFormat
+{
+    public function subscribe(Dispatcher $events)
+    {
+        $events->listen(PrepareSerializedSetting::class, [$this, 'checkCustomLessFormatBeforeSaving']);
+    }
+
+    public function checkCustomLessFormatBeforeSaving(PrepareSerializedSetting $event)
```
luceos: I think just calling this method `check` or `safeguard` or something would suffice. No need to use such a long method name which repeats most of the class name.

tobscure:
I honestly don't think there's much point in implementing this in its current form, we should instead think about how to refactor [as per my comment](https://github.com/flarum/core/pull/1273#discussion_r147020282).

clarkwinkelmann:
@luceos indeed. I put it there because tha'ts where resources were recompiled. It sure doesn't make sense for saving settings :innocent: 

@tobscure as you wish :smile: But I won't have enough time to submit a PR for that other option in a near future...

src/Core/Listener/CheckCustomLessFormat.php
clarkwinkelmann: Sure. I moved the lot from the Provider and didn't rename it :innocent:

clarkwinkelmann:
How do you wish to continue with this ? I'm willing to take time to properly finish this PR, but I don't want to spend time on it if you've planned something else @tobscure :smiley:

tobscure:
@clarkwinkelmann yep go for it :)

clarkwinkelmann:
I made the changes suggested by @luceos Should I leave the Listener inside `Core\Listener` ? There does not seem to be a folder for admin listeners in particular ?

I noticed there's a lot of calls to `$events = $this->app->make('events');` in both the ForumServiceProvider and AdminServiceProvider. Wouldn't it make more sense to call all `$events->listen` in a single method and only resolve the `events` object from the container once ? :thinking:

franzliedke:
Looks great, @clarkwinkelmann. Thank you!

As for the namespace and the resolution of the event dispatcher: both will be obsolete with the upcoming merge of our namespace restructuring, so don't worry.

PR #1276, opened 2017-10-27
Fix blurry chrome image rendering for Avatar / Logo - closes #1259

### LINKED ISSUE #1259: Blurry downscaled images (avatars, emoticons, etc) in Chrome
## Explanation

Downscaled images looks blurry on Chrome and they looks bad. In FireFox, downscaled images are OK.
You can see that this issue is part of Chrome and was already discussed and solved with an extra CSS line at https://stackoverflow.com/questions/37906602/blurry-downscaled-images-in-chrome

## Technical details

```
Not relevant.
```

## Flarum info

```
Not relevant.
```

## Log files

```
Not relevant.
```

markv:
Adds css changes to images to fix blurry rendering in chrome - #1259

### DIFF
--- a/less/lib/App.less
+++ b/less/lib/App.less
@@ -171,6 +171,7 @@
 .Header-logo {
   max-height: 30px;
   vertical-align: middle;
+  // Prevent blurriness in Chrome
   image-rendering: -webkit-optimize-contrast;
 }
 
--- a/less/lib/App.less
+++ b/less/lib/App.less
@@ -171,6 +171,7 @@
 .Header-logo {
   max-height: 30px;
   vertical-align: middle;
+  image-rendering: -webkit-optimize-contrast;
 }
 
 // On phones, the header is displayed inside of the drawer. We lay its
--- a/less/lib/Avatar.less
+++ b/less/lib/Avatar.less
@@ -14,6 +14,7 @@
     height: 100%;
     border-radius: 100%;
     vertical-align: top;
+    image-rendering: -webkit-optimize-contrast;
   }
 }
 
--- a/less/lib/Avatar.less
+++ b/less/lib/Avatar.less
@@ -14,6 +14,7 @@
     height: 100%;
     border-radius: 100%;
     vertical-align: top;
+    // Prevent blurriness in Chrome
     image-rendering: -webkit-optimize-contrast;
   }
 }


franzliedke:
Awesome, thanks for your contribution!

What about doing this for all `img` tags? Or is that potentially bad for post content, such as photographs?

Also, can I ask you to add a small comment to these lines? Something like "Prevent blurriness in Chrome"...

Zeokat:
This fix works for avatars and logo, but don't fix emoticons and all downscaled images. It doesn't completely fix the issues described in https://github.com/flarum/core/issues/1259

markv:
@franzliedke I was trying to contain the impact of the change on the site. I'm not sure what side effects adding that css prop to every img tag in the project would have.

@Zeokat I've made a separate PR for the emoticons: https://github.com/flarum/flarum-ext-emoji/pull/12

markv:
Ideally it would be better to resize the image on the server instead of down sampling in the client

Zeokat:
You are totally right @markv , the best way to solve this issue will be resize the images properly and store them on the server. It will help optimizing the page load speed and also will solve this issue.

Chrome render downscaled images blurry as part of its _"optimizations"_ to save some CPU cicles. Anyway, have a blurry logo, avatars or emoticons is visually very ugly.

As i know that @franzliedke usually requests a solid source to include this kind of changes, in the next Google Image Optimization guide, you can read why deliver scaled images is a must: https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/image-optimization#delivering_scaled_image_assets

tobscure:
The reason we serve assets with more pixels than might be needed is to support retina displays. From those Google guidelines:

"you should ensure that the number of unnecessary pixels is minimal, and that your **large assets** in particular are delivered as close as possible to their display size"

These assets (header logo, avatars, and emoji) are not large, and I think the complexity that would be added to generate multiple versions of the assets and serve them based on display density would not be worth the trouble. 

Additionally, since all of these assets may be displayed at different sizes according to the theme (particularly avatars), we should serve a reasonable upper limit. 

I do not think we should add this image-rendering property to all img tags because for large images (in posts) I presume it would use much more CPU with relatively little benefit.

Zeokat:
@tobscure have logo/avatars/emoticons blurry causes that the look of the forum turns to very ugly. At least fix those parts please.

tobscure:
@Zeokat Yep I'm happy for these PRs to be merged, I was just saying I don't think we should make it a blanket rule for all img tags.

markv:
@tobscure usually the way to deal with this issue is any non-photo assets (logo, emojis, icons) are either provided as font icons or svg so they are scalable to any resolution.

tobscure:
@markv Good point. We already use an icon font in Flarum (Font Awesome). [Looks like providing emojis as SVGs should be possible](https://github.com/s9e/TextFormatter/blob/master/docs/Plugins/Emoji/Configuration.md#use-svg-images) - if someone could PR that would be fantastic. Perhaps we could allow uploading forum logo as SVG too, open to a PR for that!

franzliedke:
Thanks a lot, @markv! Good job!

Just a hint for your next pull request: it is always safer to create a new branch in your forked repository before creating a pull request. This way, your master can not diverge from our master branch (as it did now because I squash-merged the branch).

Zeokat:
@tobscure PR for SVG emoticons done at https://github.com/flarum/flarum-ext-emoji/pull/13

PR #1298, opened 2017-11-30
remove autolink

sijad:


luceos:
I've just did a check on my current vendor folder, the autoLink method of the String object is not in use anywhere else.

Issue #881, opened 2016-03-20
Improve (or avoid) the nojs experience


dcsjapan:
Inspired by this thread, following an update of discuss.flarum.org:

https://discuss.flarum.org/d/2303-locked-out-via-nojs

The message displayed on the basic HTML site could suggest that users try clearing their browser cache and reloading the page to see if that fixes the problem.

tobscure:
:+1: we should also remove the redirection to nojs=1 so that they can hit refresh and it will attempt to boot the JS app again, rather than just displaying the nojs version

JoshyPHP:
Or maybe add a link to the JS-enabled page in the alert banner.

franzliedke:
Getting this nojs stuff right is tricky indeed. I think I suggested automatically redirecting back to the JS version with JavaScript. This takes care of those people that have JS disabled. But it can also cause infinite back-and-forth redirects if there is some JS error on the site.

How would this work without a redirect, @tobscure ?

I like the solution by @JoshyPHP best.

tobscure:
Maybe if we catch a JS error, we show an error message (using JavaScript) which says "please try reloading the page, or click here to view the HTML-only version"

datitisev:
Couldn't we use something with the history APi to replace state to '/' but while they are in the noJs page?

tobscure:
@datitisev That's a very clever idea :)

franzliedke:
But it doesn't solve the problem for those people with JS disabled, once they enable JS and refresh the page, they will still see that error message. (Only then the history API will kick in.)

tobscure:
People with JS disabled won't be redirected to nojs=1 in the first place – they'll just see the `<noscript>` content, and when they enable JS and refresh, it will work

franzliedke:
Ah, I totally missed that part. ;)

datitisev:
Glad I was able to help :)

Too bad I get my ideas in my sleep though.
I had to wait for morning and hope I would remember. ;)

webeindustry:
It seems there is a fix for this, I found here:

[https://laracasts.com/discuss/channels/laravel/how-to-force-browser-to-update-js-and-css-files-cache-after-deploy?page=1](url)

It requires creating a versioning system for javascript. So when an update is applied, the client will be forced to clear the cache from the older version.

dcsjapan:
:+1: for ways to avoid the nojs experience altogether.

Troubleshooting suggestions and redirects would still be good, if all else fails.

[...]

Issue #1024, opened 2016-08-20
Uglify JS in flarum-gulp, commit minified files and source maps to the repo


tobscure:
See flarum/flarum-gulp#2

Advantages:
- No minification required on PHP end -> better performance, especially when enabling/disabling extensions
- Don't have to rely on a third-party minifier (nullifies #887)
- Full control over minification process - less likely for installation-specific errors to occur
- Source maps! Small asset filesize even in debug mode

Disadvantages:
- Lose some of the extra compression of closure compiler
  - Could potentially still use closure compiler in flarum-gulp instead of uglifyJS?
- Have to commit minified files/sourcemaps to the repo (could result in larger diffs/repo size)
  - But we already commit unminified dist files to the repo, so it's not a huge change

sijad:
> Could potentially still use closure compiler in flarum-gulp instead of uglifyJS?

https://github.com/google/closure-compiler-js supports gulp

tobscure:
@franzliedke do you agree with this idea?

franzliedke:
If it wouldn't mean that we get very ugly diffs every time, I would immediately say yes... I am slightly against it for that reason right now, but let me sleep over it.

tobscure:
GitHub's interface will hide diffs for minified/ugly/large files, I believe.

sijad:
only developers need source maps, right? can they build them or download them from other repo somehow?

franzliedke:
Oh, Google's Closure Compiler is now available as JS version, too: https://developers.googleblog.com/2016/08/closure-compiler-in-javascript.html?m=1

franzliedke:
The problem with ugly diffs is not seeing them or not seeing them, it's the size of the repo that will quickly balloon, because every small change to a JS file will mean that Git will essentially store both the old and the new version completely every time.

UrGuardian4ngel:
Those diffs can also be suppressed using [`.gitattributes`](https://www.kernel.org/pub/software/scm/git/docs/gitattributes.html), [marking them as binary files](https://www.kernel.org/pub/software/scm/git/docs/gitattributes.html#_marking_files_as_binary). E.g.:

```
# .gitattributes
js/*/dist/*.js -diff
```

But I'm not sure in which ways this will affect the size or 'ballooning' of the repository (if at all).

franzliedke:
If we decide to commit the minified files, then yes, we should do that. Good idea, @UrGuardian4ngel. :)

tobscure:
[...]

PR #1309, opened 2017-12-15
Fix two asset cache invalidation bugs

gwillem:
Proof of concept fix for #881 (the infamous nojs bug).

Currently, cache validation never happens because the revision is never updated. This PR replaces the faulty hash method with a hash of the actual content. Cache invalidation is now triggered by the flush method, by saving an empty revision to `rev-manifest.json`.

### DIFF
--- a/src/Asset/RevisionCompiler.php
+++ b/src/Asset/RevisionCompiler.php
@@ -69,34 +69,27 @@ class RevisionCompiler implements CompilerInterface
      */
     public function getFile()
     {
-        $old = $current = $this->getRevision();
+        $oldrev = $this->getRevision();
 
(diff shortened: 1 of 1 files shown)


gwillem:
My proposed solution probably should be refactored, however, may I humbly suggest to add this to the `v0.1.0-beta.8` milestone? This bug affects nearly every new user.

tobscure:
I will take a look at this (and the rest of the asset compilation logic) as part of #1024.

For my own reference, I noted down a few todos a while ago:

* Rename FrontendAssets functions to `make`
* Add `clear` function to RevisionCompiler
* Use Filesystem dependency in RevisionCompiler

PR #1330, opened 2018-01-01
Prevent editing fields in sign up modal according to identification data

clarkwinkelmann:
Currently the difference between "forced value" fields and "suggested value" fields coming from OAuth providers is completely removed from the front-end. This means that a custom provider forcing a value for "username" does not automatically locks this field in the sign up modal.

This PR adds that feature. I tested it with a custom flagrow/passport configuration to either force username, email or username-and-email based accounts. I also checked Twitter still works (it's the only provider I know that doesn't force either username or password).

I'm not sure about the naming. I named the list of fields `identification_field` because they are part of the `identification` data in the backend, but maybe something like `locked_fields`, `immutable_fields` or `provider_fields` would make more sense ?


tobscure:
I only vaguely remember when I wrote the OAuth system, but I believe this may have been somewhat intentional. Consider this scenario:

1. Click "Log in with Facebook"
2. Flarum opens the sign-up form with pre-filled (suggested) values
3. I see that my Facebook email is not the one I actually want to use, and wish to change it

Not sure if this is actually the best thing to do or not, I suppose in that case if you change the pre-filled email, the account you create will no longer be associated with Facebook... however we do have [another issue](https://github.com/flarum/core/issues/897) regarding saving the Facebook ID in the database and using that as the identification rather than the email, which would solve that.

clarkwinkelmann:
@tobscure it's already happening server side. `getIdentification` data can't be changed, while `getSuggestions` data is merely a pre-fill value. This PR only makes the front-end aware that you can't edit any `getIdentification` value, while still letting you edit any suggested value.

js/forum/src/components/SignUpModal.js
```diff
@@ -60,6 +60,10 @@ export default class SignUpModal extends Modal {
     ];
   }
 
+  fieldCantBeEdited(field) {
```
tobscure: Could we please change this to `isProvided(field)`?

src/Forum/AuthenticationResponseFactory.php
```diff
@@ -104,7 +104,14 @@ private function getPayload(array $identification, array $suggestions, User $use
...
             $token->save();
 
-            $payload = array_merge($identification, $suggestions, ['token' => $token->id]);
+            $payload = array_merge(
+                $identification,
+                $suggestions,
+                ['token' => $token->id],
+                // List of the fields that can't be edited during sign up
+                // Only includes attributes on the sign up form, otherwise this could leak private attribute names
+                ['identification_fields' => array_keys(array_only($identification, ['username', 'email']))]
```
tobscure: We're already sending `$identification` values through in the payload above (line 108), I don't think we need to worry about leaking private attribute names.

Please change the key to `identificationFields`.

tobyzerner (changes requested):
I understand this better now and happy for it to be merged, pending a couple of changes :)

js/forum/src/components/SignUpModal.js
clarkwinkelmann: I agree to change this, but isn't `isProvided` a bit vague about its purpose as well ?

src/Forum/AuthenticationResponseFactory.php
clarkwinkelmann: Didn't even notice it. It makes no sense to filter it then indeed :sweat_smile: 

My reasoning was to prevent a token like `twitter_id` (either its name or value) to be sent to the browser... But it's usually a public value anyway. I'll remove that

js/forum/src/components/SignUpModal.js
tobscure: `isProvided` reads nicely for me. "disable this field if its value is provided"

`isIdentifier` could also work, "disable this field if it's an identification field"

clarkwinkelmann:
Made the changes. Had to merge master into the PR to get my dev setup working again, I hope this cause no issue.

I'm still unsure if you want me to run `scripts/compile.sh` and commit the changes ? When I run it it changes a lot more than what I just did. Including in admin.js where I didn't change anything in this PR.

Just tell me and I'll include them :smile:

tobscure:
Thanks Clark!

PR #1332, opened 2018-01-02
Always update Composer height on show

clarkwinkelmann:
This prevents visual glitches when the Composer component was just replaced

The main issue that gets fixed with this happens when you click on "new discussion" while the composer is already open (many people must already have seen that one I guess):

image


clarkwinkelmann:
This might actually have been fixed in https://github.com/flarum/core/commit/1a102766a9c9570b92621a9d7d3c9e011bc6bfbe

I just realized my core wasn't up to date. And now that it is I can't get it running :hankey: I'll try to give an update soon

Issue #851, opened 2016-03-07
New PHP Extension API


tobscure:
In every extension, we have an AddClientAssets listener which is basically the same:

``` php
class AddClientAssets
{
    public function subscribe(Dispatcher $events)
    {
        $events->listen(ConfigureClientView::class, [$this, 'addAssets']);
    }

    public function addAssets(ConfigureClientView $event)
    {
        if ($event->isForum()) {
            $event->addAssets([
                __DIR__.'/../../js/forum/dist/extension.js',
                __DIR__.'/../../less/forum/extension.less'
            ]);
            $event->addBootstrapper('flarum/auth/facebook/main');
        }

        if ($event->isAdmin()) {
            $event->addAssets([
                __DIR__.'/../../js/admin/dist/extension.js',
                __DIR__.'/../../less/forum/extension.less'
            ]);
            $event->addBootstrapper('flarum/auth/facebook/main');
        }
    }
}
```

Given that this asset file-structure is a best practice, we could reduce this duplication by providing an instantiable `AddDefaultClientAssets` listener. In bootstrap.php, instead of:

``` php
return function (Dispatcher $events) {
    $events->subscribe(Listener\AddClientAssets::class);
};
```

you would use:

``` php
return function (Dispatcher $events, Extension $extension) {
    $events->subscribe(new AddDefaultClientAssets($extension));
};
```

This would add the default asset file paths if they exist, along with JS bootstrappers using the Extension's ID as a prefix.

We could also extend this idea of shortcut listeners to other things, e.g.:

``` php
return function (Dispatcher $events) {
    $events->subscribe(new AddForumRoute($extension, 'get', '/auth/facebook', 'auth', FacebookAuthController::class));

    $events->subscribe(new AddPostType(DiscussionStickiedPost::class));
};
```

This needs discussion because to me it's a little unclear how far we would want to go with providing these helpers. Where do we draw the line?

Luceos:
I only see new "shortcuts" popup, with easier adoption and faster development as a reason. But isn't this pointing us at a bigger problem? Endlessly adding helpers will only increase the size of the platform, which is not something we should do lightly.

How can we optimize the code to make extension development easier, without losing any flexibility at the same time?

franzliedke:
I agree with Toby's suggestion.

But yes, we have to be careful with all of these. But here, just like with my migration helpers, I'd say it's warranted. Since we're doing this only by adding new helper classes, the changes are very un-intrusive, so I don't see any big problems with that approach.

[...]

PR #1335, opened 2018-01-03
Start of PHP extenders API

franzliedke:
Refs #851.

This adds the first three extender classes (`Locale`, `FormatterConfiguration` and `Assets`) for the new backend extension API. More are in the pipeline.

I have written this in such a way that it won't break existing extensions - if an extension bootstrapper script returns a closure, it is wrapped in a `Compat` extender. This allows the backend to deal with extenders only, whereas extensions still have to migrate.

Whether we will leave this in core for the beta.8 release is up for discussion - we need it at least until all bundled extensions are converted (and we need more extender implementations to do so).

I have also pushed `next-back-extenders` branch to some extensions (bbcode, emoji, english, and markdown).

### DIFF
--- a/src/Event/AbstractConfigureRoutes.php
+++ b/src/Event/AbstractConfigureRoutes.php
@@ -14,6 +14,9 @@ namespace Flarum\Event;
 use Flarum\Http\RouteCollection;
 use Flarum\Http\RouteHandlerFactory;
 
+/**
+ * @deprecated
+ */
 abstract class AbstractConfigureRoutes
 {
     /**
--- a/src/Event/ConfigureApiRoutes.php
+++ b/src/Event/ConfigureApiRoutes.php
@@ -12,9 +12,7 @@
 namespace Flarum\Event;
 
 /**
- * Configure API routes.
- *
- * This event is fired when API routes are being registered.
+ * @deprecated
  */
 class ConfigureApiRoutes extends AbstractConfigureRoutes
 {
--- a/src/Event/ConfigureForumRoutes.php
+++ b/src/Event/ConfigureForumRoutes.php
@@ -14,9 +14,7 @@ namespace Flarum\Event;
 use Flarum\Forum\Controller\FrontendController;
 
 /**
- * Configure forum routes.
- *
- * This event is fired when routes for the forum client are being registered.
+ * @deprecated
  */
 class ConfigureForumRoutes extends AbstractConfigureRoutes
 {
--- a/src/Event/ConfigureLocales.php
+++ b/src/Event/ConfigureLocales.php
@@ -15,6 +15,9 @@ use DirectoryIterator;
 use Flarum\Locale\LocaleManager;
 use RuntimeException;
 
+/**
+ * @deprecated
+ */
 class ConfigureLocales
 {
     /**
--- a/src/Event/ConfigureMiddleware.php
+++ b/src/Event/ConfigureMiddleware.php
@@ -13,6 +13,9 @@ namespace Flarum\Event;
 
 use Zend\Stratigility\MiddlewarePipe;
 
+/**
+ * @deprecated
+ */
 class ConfigureMiddleware
 {
     /**
--- a/src/Extend/Assets.php
+++ b/src/Extend/Assets.php
@@ -1,0 +1,81 @@
+<?php
+
+/*
+ * This file is part of Flarum.
+ *
+ * (c) Toby Zerner <toby.zerner@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Flarum\Extend;
+
+use Flarum\Frontend\Event\Rendering;
+use Illuminate\Contracts\Container\Container;
+use Illuminate\Events\Dispatcher;
+
+class Assets implements Extender
+{
+    protected $appName;
+
+    protected $assets = [];
+    protected $bootstrapper;
+
+    public function __construct($appName)
+    {
+        $this->appName = $appName;
+    }
+
+    public function defaultAssets($baseDir)
+    {
+        $this->asset("$baseDir/js/{$this->appName}/dist/extension.js");
+        $this->asset("$baseDir/less/{$this->appName}/extension.less");
+
+        return $this;
+    }
+
+    public function asset($path)
+    {
+        $this->assets[] = $path;
+
+        return $this;
+    }
+
+    public function bootstrapper($name)
+    {
+        $this->bootstrapper = $name;
+
+        return $this;
+    }
+
+    public function apply(Container $container)
+    {
+        $container->make(Dispatcher::class)->listen(
+            Rendering::class,
+            function (Rendering $event) {
+                if (! $this->matches($event)) {
+                    return;
+                }
+
+                $event->addAssets($this->assets);
+
+                if ($this->bootstrapper) {
+                    $event->addBootstrapper($this->bootstrapper);
+                }
+            }
+        );
+    }
+
+    private function matches(Rendering $event)
+    {
+        switch ($this->appName) {
+            case 'admin':
+                return $event->isAdmin();
+            case 'forum':
+                return $event->isForum();
+            default:
+                return false;
+        }
+    }
+}
--- a/src/Extend/Compat.php
+++ b/src/Extend/Compat.php
@@ -1,0 +1,38 @@
+<?php
+
+/*
+ * This file is part of Flarum.
+ *
+ * (c) Toby Zerner <toby.zerner@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Flarum\Extend;
+
+use Illuminate\Contracts\Container\Container;
+
+/**
+ * This class is used to wrap old bootstrap.php closures (as used in versions up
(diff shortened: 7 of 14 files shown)


tobscure:
Beautiful! Seeing those `next-back-extenders` branches makes me very happy 😊 

Rather than wrapping in an Compat extender, can we just support functions implicitly? As per the original proposal I wrote: https://github.com/flarum/core/issues/851#issuecomment-259126656

franzliedke:
You mean for returning an array of extenders *and* functions? For e.g. auth-github, the bootstrap.php would then look like the following:

~~~php
use Flarum\Event\ConfigureForumRoutes;
use Flarum\Extend;
use Illuminate\Contracts\Events\Dispatcher;

return [
    (new Extend\Assets('forum'))
        ->defaultAssets(__DIR__)
        ->bootstrapper('flarum/auth/github/main'),
    (new Extend\Assets('admin'))
        ->asset(__DIR__.'/js/admin/dist/extension.js')
        ->bootstrapper('flarum/auth/github/main'),
    function (Dispatcher $events) {
        $events->listen(
            ConfigureForumRoutes::class,
            function (ConfigureForumRoutes $event) {
                $event->get('/auth/github', 'auth.github', 'Flarum\Auth\GitHub\GitHubAuthController');
            }
        );
    }
];
~~~

That would work as well, of course. I would probably convert them to `Compat` extenders when loading the extenders in our `ExtensionManager`, just to have a nice unified interface when working with the array of extenders later.

I wonder, though, do you want this for BC sake or for advanced use-cases that are not covered by extenders? In the latter case, I would actually prefer having a "compat" extender (probably named differently) explicitly, because it would clearly signal to the extension author that they are leaving the safety net of our BC guarantee (extenders are under SemVer, other code isn't necessarily, right?). And I very much like having that signaled through a name. Speaking of which, what would be a good name here? :thinking: 

To make this clearer, this is what the auth-github bootstrapper currently looks like on my machine (because I haven't yet written the extender for creating routes):

~~~php
use Flarum\Event\ConfigureForumRoutes;
use Flarum\Extend;
use Illuminate\Contracts\Events\Dispatcher;

return [
    (new Extend\Assets('forum'))
        ->defaultAssets(__DIR__)
        ->bootstrapper('flarum/auth/github/main'),
    (new Extend\Assets('admin'))
        ->asset(__DIR__.'/js/admin/dist/extension.js')
        ->bootstrapper('flarum/auth/github/main'),
    new Extend\Compat(function (Dispatcher $events) {
        $events->listen(
            ConfigureForumRoutes::class,
            function (ConfigureForumRoutes $event) {
                $event->get('/auth/github', 'auth.github', 'Flarum\Auth\GitHub\GitHubAuthController');
            }
        );
    })
];
~~~

tobscure:
You're absolutely right. I'll have a think about how we should name it.

Eldenroot:
OK, any list of planned extenders?

franzliedke:
Voila. I added the `Route` extender (whose implementation can be improved, I think to `extend` the route collection bindings instead of retrieving them out of the container - but that can safely be done later).

That allowed me to finish our three auth extensions as well.

@tobscure If you're fine with this, would you merge this into master (along with the extension branches)? Since the introduction is backwards-compatible so far, I'd like to continue (and then split up) the work on the master branch.

@Eldenroot I'll list them in a call for contributions on the forum in the coming days; we're still in the experimentation phase.

tobscure:
@franzliedke Just one thought before I do – instead of `new Extend\Assets('forum')` and `new Extend\Route('admin', ...)`, would it be worth finding some way to constant-ize the `'forum'`/`'admin'` part – options I can think of are:

* class constants (eg. `Extend::FORUM` or `Extend\Assets::FORUM`)
* subclassing (eg. `new Extend\ForumAssets()`, `new Extend\AdminRoute(...)`
* static methods (eg. `Extend\Assets::forum()`, `Extend\Route::admin(...)`)

This may also be important for BC-planning in case we introduce some kind of fundamental difference between forum/admin assets/routes in the future. In that case, the subclassing would probably be the way to go – we have two truly separate APIs, and the common parent class is just an implementation detail.

And using strings just feels a bit flimsy :P Then again, we've always used strings for things like HTTP methods (get, post, etc) so maybe it's not that big of a deal...

luceos:
I'm in favor of not using strings too, they are too easy to mistype. Of the three propositions I dislike subclassing most. I think the static methods look cleanest, but then again it's no longer a class instantiation, so that breaks our proposed logic..

If you want to hold onto subclassing, can we namespace them? `Extend\Assets\Forum` and `Extend\Routes\Admin`?

---

Speaking of which, can we please stick to either plural or singular? I see Route and Assets being used. Either use Asset and Route or Assets and Routes. Otherwise we all need to keep looking at class inspections all the time 😁

tobscure:
@luceos The plurality here reflects how the extenders actually function.

With `Assets` you instantiate it and then either add `defaultAssets` or an `asset` or a `bootstrapper` or any combination. With `Route` you are merely defining a single route.

But I agree we should consider make it consistent if possible. That would either mean splitting up the Assets extender:

```php
return [
    new Extend\DefaultAssets('forum'), // these classes could be namespaced under Extend\Assets
    new Extend\Bootstrapper('forum', 'whatever'),
    new Extend\Asset('forum', 'whatever')
];
```

or allowing the definition of multiple routes in a single extender instance:

```php
return [
    (new Extend\Routes('forum'))
        ->get(...)
        ->post(...)
];
```

Consider the scenario where you want to *remove* an existing route:

```php
new Extend\RemoveRoute('forum', 'whatever')
// vs
(new Extend\Routes('forum'))->remove('whatever')
```

tobscure:
> In the context of assets, they identify instances of web apps, which - again in this context - are the same.

I think I have a curse of trying to think ahead for every different possible scenario (resulting in analysis paralysis!) My thought here is that the admin and forum web apps only happen to be similar by their implementation, but conceptually they are distinct. For example, imagine a hypothetical where we rewrite the admin (but not the forum) in a different JS framework, and it no longer has the concept of "bootstrappers". Or perhaps just the forum defaultAssets diverge from the admin defaultAssets. Not the best examples but perhaps they demonstrate the possibility of a discrepancy in the future between the admin/forum APIs. In any case, I am happy to concede that I am probably overthinking... but at least you know my thought process now!

> Let's wait with consistency changes until all extenders have been created. (I am thinking we will probably want sub-namespaces at that point.)

OK, sounds good.

franzliedke:
> I think I have a curse of trying to think ahead for every different possible scenario

I can understand the sentiment; finding the right balance of thinking ahead and dealing with the present is the tricky, unsolved part. ;)

As far as I can tell, the examples you mentioned can all be taken care of by introducing new extenders when the time comes / any need arises.

franzliedke:
Can we merge this then? :grin: 

Of course, none of this needs to be considered final. Once all extenders are done, we should go through them again and try to find obvious issues / sort out inconsistencies.

franzliedke:
:drooling_face: I'll merge the branches in the various extensions.

franzliedke:
@tobscure @luceos Just so you know: I switched to `Extend\Routes`: https://github.com/flarum/flarum-ext-auth-github/commit/ccfe95740778aae0300da234c829634642632f66 - more consistent, easier route registration, allows multiple registrations in one extender. :)

tobscure:
Cool. That also makes it possible for us to have a `remove` method in the future :)

I still can't shake the elegance of the "everything is a function" approach, if we make Extenders merely glorified functions (invokable classes)... and they could even be "composites" that return their own set of extenders/functions.

```php
// bootstrap.php
return [
    function (Some $dependency) {
        // do something
    },
    (new Routes('forum'))
        ->get('/auth/github', 'auth.github', GitHubAuthController::class)
];

// Extend\Routes.php
class Routes {
    // ...
    public function __invoke() {
        return array_map(function ($route) {
            return new Route($this->appName, $route['method'], $route['path'], $route['name'], $route['handler']);
        }, $this->routes);
    }
}

// Extend\Route.php
class Route {
    // ...
    public function __invoke(Container $container)
    {
        // add the route
    }
}
```

Isn't that kinda beautiful? All the extension runner has to do is loop through recursively and go "is this callable?" YES → call it and recurse on the result (if there is one), or NO → recurse. Breaking up into Routes/Route extenders is not the most realistic example, but more just a demo of how simple it is.

Anyway, I'll shut up now :D

(Also, should we consider renaming `bootstrap.php` to `extend.php` or `extension.php` or something? Because it's not really "boostrapping" anymore, is it?)

franzliedke:
@tobscure I am torn.

Yes, you are right, using callables / invokables is that much more elegant (no need for a custom interface with my `apply(Container $container)` method). But, I like the explicit-less and narrow scope (in terms of public API) of the current approach. If we accept any closure, people can do anything (including things likely to break between releases). In that case, we would need to make the differences in BC guarantees more explicit in the docs.

Tradeoffs, tradeoffs...

tobscure:
People can do anything anyway, it's just a matter of documentation. With functions I think it would be just as easy to make it explicit in the docs: we just say if you're *not* using an Extender, you're playing with fire. Extenders guarantee BC, if you need to do something that can't be achieved with an Extender, report an issue.

franzliedke:
Okay, I've turned extenders into invokables, as you suggested. This basically renders the `Compat` extender obsolete, right?

Let's keep all further discussion to #851.

PR #1338, opened 2018-01-06
Remove execute permissions from php/less files

gwillem:


franzliedke:
Thank you sir! :bowing_man:

PR #1342, opened 2018-01-11
Overhaul model visibility scoping

### LINKED ISSUE #1333: The posts list of a user shows no posts if there are too many posts marked as is_private

tobscure:
This should fix #1333 and is required to properly implement visibility permissions for searching (see #1339).

Note that these changes have not been extensively tested, but I have been over the logic multiple times and it seems to check out. (famous last words)

See associated changes in:
https://github.com/flarum/flarum-ext-approval/commit/64d1ef0ec0d0d7849a91ce5f58592aefef05858f
https://github.com/flarum/flarum-ext-tags/commit/74c4430a290b0566b054ddd3be811ac8e913e2d4

@luceos These changes will break byobu. A lot. Once this is merged, if you'd like I can try and send a PR for byobu soon, since we will need it upgraded in order to deploy to discuss.flarum.org anyway.


luceos:
I'm a bit packed with work right now. I might be able to take a look at the end of the week. If you want to push forward feel free to send in a PR before then.

Issue #1319, opened 2017-12-25
Don't raise a million PostWasDeleted events when deleting a discussion


tobscure:
Require extensions to listen for DiscussionWasDeleted instead and act appropriately.

datitisev:
What if someone is making an extension to notify on post deletion AND discussion deletion, like webhooks?
I'd make it so there's a way to check if the main event is from deleting a discussion, and if so ignore the event / dont fire the event.

tobscure:
Changed my mind about this, deleting a discussion doesn't happen that often so doesn't matter if the queries are a little bit inefficient - because it makes life a heck of a lot easier for extension developers.

franzliedke:
@tobscure Isn't the problem that we query and hydrate *all* post instances for the thread?

What about an event "Posts Deleted" instead that contains an array of IDs?

(In any case, this should go via queue, once we have a queue system in place.)

tobscure:
It's much easier to program extensions to perform whatever action in response to one event (post deleted) with all the post information available, rather than that and an additional event (posts deleted) where they may need to fetch the post information themselves anyway.

We could certainly chunk the post query though so memory doesn't become an issue.

franzliedke:
Let's try using `cursor` for now. :)

tobscure:
Didn't know that existed! Sounds good, will reopen this.

tobscure:
Oops, just saw you already did it. The GitHub app on my phone doesn't show commit refs among comments :(

franzliedke:
For the record: with the huge database changes of #1236, we're now back to Toby's original proposal, which is to not throw all the events. The database takes care of deletion now via foreign keys with cascades.

Issue #1280, opened 2017-11-02
Remove user bio column from database


franzliedke:
The database column should be removed in time for the stable release. This gives people a grace period to switch to an extension as replacement.

See #1214.

franzliedke:
Quoting @tobscure from https://github.com/flarum/core/pull/1214#issuecomment-343702662:

> Bundled extension will come before stable release.

So that will use the existing data if available, otherwise the migration script will remove it.

askvortsov1:
A few things:

1. https://github.com/FriendsOfFlarum/user-bio isn't bundled, was it supposed to be?
2. How can we remove this without breaking https://github.com/FriendsOfFlarum/user-bio? We could have fof change the column name on their end and us remove the bio on our end, but i don't think thats a very elegant solution.

franzliedke:
After some discussion in our last team meeting, I don't think we're willing to bundle (and support) more extensions officially until stable.

As for migration: since we're in beta, I think it would be fine if user-bio just offered a one-time script or snippet to execute once before we remove the field. We can even make that an optional step in our upgrade instructions for users of that extension.

askvortsov1:
Could we add a migration to remove the column IF there's no data in it? That way we could not have it on fresh installations without hacky workarounds

franzliedke:
No. The state of all databases should be the same after running a certain migration. I could live with migrations that have checks for different states coming *into* the migration, but the results should be deterministic.

askvortsov1:
Fair. So we just have user bio rename the field? I'll open an issue on their end, but us actually removing the column should be delayed until the stable release to give users time to migrate user bio.

franzliedke:
Either that (a migration that creates a new field and prefills it from the old field *if that exists*), or a one-time migration scripts for users with old data. I'll leave that choice to them.

> us actually removing the column should be delayed until the stable release to give users time to migrate user bio.

Yup, see issue description. :wink:

KyrneDev:
Why don't we just remove the migration from where it is created instead of just dropping it? Old forums will still have the column and new forums won't unless the install the extension. I don't see any benefit to dropped it outright.

askvortsov1:
That is a much, much better idea than this hackery I've come up with

PR #1344, opened 2018-01-11
Database changes

### LINKED ISSUE #1236: Database naming changes
As per https://gist.github.com/tobscure/b189580dd17b9def574f92692faa0e56

---

Edit by @luceos

- [x] update phpdoc
- [x] update existing properties in flarum/core where a renameColumn was executed #1344 
- [x] update properties in core extensions where a renameColumn was executed
- [x] modify the accessor in User for the avatar_url https://github.com/flarum/core/commit/c31c1ea062d7be4bc0214330e4a1e5d130815b73
- [x] migrations for core; #1344 
- [x] migrations for suspend ext
- [x] migrations for tags ext
- [ ] add indices see [comment by Toby](https://github.com/flarum/core/issues/127#issuecomment-370190948)
- [ ] solve preferences drop from users, table or column
  - [ ] user notifications table
  - [ ] all other preferences to columns on users

luceos:
**Closes #1236**

**Related PRs**
https://github.com/flarum/flarum-ext-akismet/pull/10
https://github.com/flarum/flarum-ext-approval/pull/15
https://github.com/flarum/flarum-ext-flags/pull/15
https://github.com/flarum/flarum-ext-likes/pull/16
https://github.com/flarum/flarum-ext-lock/pull/16
https://github.com/flarum/flarum-ext-mentions/pull/34
https://github.com/flarum/flarum-ext-sticky/pull/11
https://github.com/flarum/flarum-ext-subscriptions/pull/18
https://github.com/flarum/flarum-ext-suspend/pull/19
https://github.com/flarum/flarum-ext-tags/pull/53


franzliedke:
Please request reviews when this is out of WIP. :)

luceos:
Absolutely. Feel free to discuss anything that comes up in the meantime, might save some time refactoring it all over again 👍

migrations/2018_01_30_220100_create_posts_users_table.php
```diff
@@ -18,7 +18,6 @@
             $table->integer('post_id')->unsigned();
             $table->integer('user_id')->unsigned();
 
-            $table->foreign('post_id')->references('id')->on('posts')->onDelete('cascade');
```
tobscure: #1339 will change the posts table to InnoDB (https://github.com/flarum/core/pull/1339/commits/d4850eccb7b6c5bb95bb2090f032250f24e88b25), so this can be re-added with a migration that has a more recent date/time.

luceos: I think this migration is already newer than yours:

```
migrations/2018_01_11_120604_change_posts_table_to_innodb.php
```

franzliedke:
Whew, quite big. If we get #1346 merged first, this one will be easier to review, right? (Still, this might need to be split up, to be honest.)

luceos:
@flarum/core just to prevent confusion. I force pushed a clean version of the branch here, I'm basing it off of master. WIP.

luceos:
Seems like most of the changes in core are done. Part of the referenced issue is done. Now need to take care of the extensions.

luceos:
Waiting on #1418

luceos:
@flarum/core beta 8 will drop the preferences column which holds (primarily) the notification settings from the frontend of the forum. @tobscure proposed to move each of the preferences to a separate column on the user table. As a result the users table would receive a column for each of the checks shown here:

image

To me that feels like too many unnecessary columns for something that's a collection. Can't we move them to their own table instead?

tobscure:
Yes, notification preferences should probably have their own table (`notification_preferences`?) with the following columns:

* `user_id`
* `type` (eg. postMentioned)
* `channel` (eg. alert)

Alternatively we could have a new column on the users table called `notifications` which contains a JSON structure, but I think we should normalize first and see if there's any reason to switch later.

Other preferences (`discloseOnline`, `followAfterReply`) can become new columns on the users table.

migrations/2018_01_30_222800_create_notification_preferences_table.php
```diff
@@ -0,0 +1,30 @@
...
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Schema\Builder;
+
+return [
+    'up' => function (Builder $schema) {
+        $schema->create('notification_preferences', function (Blueprint $table) {
+            $table->integer('user_id')->unsigned();
+            $table->string('type');
+            $table->string('channel');
+            $table->boolean('enabled')->default(false);
```
tobscure: Do we need an `enabled` column? Doesn't the mere presence of a row in this table indicate that it is enabled?

If there is a good reason to have it, according to the conventions it should be called `is_enabled`

migrations/2018_01_30_222800_create_users_table_preferences_columns.php
```diff
@@ -0,0 +1,27 @@
...
+ * file that was distributed with this source code.
+ */
+
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Schema\Builder;
+
+return [
+    'up' => function (Builder $schema) {
+        $schema->table('users', function (Blueprint $table) {
+            $table->boolean('disclose_online')->default(false);
```
tobscure: This too should be prefixed with `is_` according to the convention. Perhaps `is_online_public`, or maybe `is_last_seen_public` to associate it with `last_seen_time`...

tobscure:
Would it be possible to move the preferences stuff into a separate PR? This one

migrations/2018_01_30_222800_create_notification_preferences_table.php
luceos: We actually need the enabled column, because there would not be a way to distinguish between a user without notification_preferences set yet (and in need for default) and someone who disabled all of them?

Good point about is_enabled.

migrations/2018_01_30_222800_create_users_table_preferences_columns.php
luceos: I like the `is_last_seen_public`. Let's go for that.

luceos:
@tobscure I guess that might make sense with the size and consequences of these changes.

migrations/2018_01_30_222800_create_notification_preferences_table.php
tobscure: Ah, didn't think of that. Good reason.

luceos:
@flarum/core the users.preferences modification has been moved to #1467 ; please note that is a child PR of this one and needs to be merged to master before a release as (for now) the users.preferences column is dropped inside this PR (bundled in the users-b8 changes migration). The migrations needed to move preferences to the new table or a new users column is part of the child PR and runs before the users-b8 migration.

migrations/2018_01_15_071700_change_user_discussions_discussions_users.php
```diff
@@ -0,0 +1,22 @@
...
+
+return [
+    'up' => function (Builder $schema) {
+        $schema->rename('users_discussions', 'discussions_users');
+    },
+
+    'down' => function (Builder $schema) {
+        $schema->rename('discussions_users', 'users_discussions');
+    }
+];
```
tobscure: This could use the `Migration::renameTable`` helper?
Also for consistency the migration should be called `rename_users_discussions_to_discussions_users.php`

migrations/2018_01_15_072600_change_email_tokens_b8_columns.php
```diff
@@ -0,0 +1,27 @@
...
+            $table->renameColumn('id', 'token');
+        });
+    },
+
+    'down' => function (Builder $schema) {
+        $schema->table('email_tokens', function (Blueprint $table) {
+            $table->renameColumn('token', 'id');
+        });
+    }
+];
```
tobscure: This could use the `Migration::renameColumn` helper? (I'm starting to think these helpers aren't that valuable, since they can only be used in such limited instances)

migrations/2018_01_18_130400_rename_permissions_to_group_permissions.php
```diff
@@ -0,0 +1,14 @@
...
+ *
+ * (c) Toby Zerner <toby.zerner@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Flarum\Database\Migration;
+
+return Migration::renameTable('permissions', 'groups_permissions');
```
tobscure: The migration file should be called rename_permissions_to_group***s***_permissions.php

migrations/2018_01_18_130400_rename_users_groups_to_groups_users.php
```diff
@@ -0,0 +1,14 @@
...
+ *
+ * (c) Toby Zerner <toby.zerner@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Flarum\Database\Migration;
+
+return Migration::renameTable('users_groups', 'group_user');
```
tobscure: The table should be called group**s**_user**s**

migrations/2018_01_18_132900_create_notifications_from_table.php
```diff
@@ -0,0 +1,40 @@
...
+        $schema->getConnection()->table('notifications')->chunkById(100, function ($notifications) use ($schema) {
+            foreach ($notifications as $notification) {
+                $insert = [
+                    'id' => $notification->id,
+                    'from_user_id' => $notification->sender_id
+                ];
+
+                $schema->getConnection()->table('notifications_from')->updateOrInsert($insert, $insert);
+            }
+        });
```
tobscure: > we don't add multiple table statements into one migration because migrations could be half run during an exception

Does this migration break that rule? Should the insertion part be in a separate migration?

migrations/2018_01_18_134400_change_password_tokens_id_to_token.php
```diff
@@ -0,0 +1,14 @@
...
+ *
+ * (c) Toby Zerner <toby.zerner@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Flarum\Database\Migration;
+
+return Migration::renameColumn('password_tokens', 'id', 'token');
```
tobscure: For consistency, this migration file should be named change_password_tokens_**rename**_id_to_token.php

migrations/2018_01_18_135000_change_posts_b8_columns.php
```diff
@@ -0,0 +1,42 @@
...
+            $table->renameColumn('created_at', 'time');
+            $table->renameColumn('edited_at', 'edit_time');
+            $table->renameColumn('hidden_at', 'hide_time');
+
+            $table->renameColumn('edited_user_id', 'edit_user_id');
+            $table->renameColumn('edited_user_id', 'hidden_user_id');
+        });
+
+        $prefix = $schema->getConnection()->getTablePrefix();
+        $schema->getConnection()->statement('ALTER TABLE '.$prefix.'posts MODIFY content FULLTEXT');
```
tobscure: What's going on here?

migrations/2018_01_19_084700_change_settings_value_longblob.php
```diff
@@ -0,0 +1,26 @@
...
+
+return [
+    'up' => function (Builder $schema) {
+        $prefix = $schema->getConnection()->getTablePrefix();
+//        $schema->getConnection()->statement('ALTER TABLE '.$prefix.'settings MODIFY "value" LONGBLOB');
+    },
+
+    'down' => function (Builder $schema) {
+        $schema->table('posts', function (Blueprint $table) {
+//            $table->longText('value')->nullable()->change();
```
tobscure: ?

migrations/2018_01_30_222900_change_users_b8_columns.php
```diff
@@ -0,0 +1,43 @@
...
+        $schema->table('users', function (Blueprint $table) {
+            $table->renameColumn('is_activated', 'is_email_confirmed');
+            $table->renameColumn('join_time', 'joined_at');
+            $table->renameColumn('last_seen_time', 'last_seen_at');
+            $table->renameColumn('discussions_count', 'discussion_count');
+            $table->renameColumn('comments_count', 'comment_count');
+            $table->renameColumn('read_time', 'marked_all_as_read_at');
+            $table->renameColumn('notifications_read_time', 'read_notifications_at');
+            $table->renameColumn('avatar_path', 'avatar_url');
+            $table->dropColumn('bio');
```
tobscure: Let's not drop bio just yet, since we don't have a replacement yet. See https://github.com/flarum/core/issues/1280

src/Api/Serializer/DiscussionSerializer.php
```diff
@@ -37,28 +37,28 @@ protected function getDefaultAttributes($discussion)
...
 
         $attributes = parent::getDefaultAttributes($discussion) + [
-            'commentsCount'     => (int) $discussion->comments_count,
-            'participantsCount' => (int) $discussion->participants_count,
-            'startTime'         => $this->formatDate($discussion->start_time),
-            'lastTime'          => $this->formatDate($discussion->last_time),
+            'commentsCount'     => (int) $discussion->comment_count,
+            'participantsCount' => (int) $discussion->participant_count,
+            'startTime'         => $this->formatDate($discussion->created_at),
+            'lastTime'          => $this->formatDate($discussion->last_posted_at),
```
tobscure: We should rename the API attributes as well, but understandable leaving that to a separate PR.

src/Http/CookieFactory.php
```diff
@@ -79,7 +80,7 @@ public function make($name, $value = null, $maxAge = null)
         if ($maxAge) {
             $cookie = $cookie
                 ->withMaxAge($maxAge)
-                ->withExpires(time() + $maxAge);
+                ->withExpires(Carbon::now()->timestamp + $maxAge);
```
tobscure: Why the use of Carbon instead of `time()`?

migrations/2018_01_18_132900_create_notifications_from_table.php
tobscure: Perhaps the insertion part should be in the same migration as where the `sender_id` column is dropped

migrations/2018_01_18_130400_rename_permissions_to_group_permissions.php
tobscure: Can we please also add a foreign key for group_id?

tobscure:
Can you please go through and add foreign keys to any columns that should have them but don't :)

migrations/2018_01_18_132900_create_notifications_from_table.php
luceos: Yes, this won't work; if the update or insert call throws an exception the migration will not be marked done, but re-running the migration isn't possible either because the first part already finished..

migrations/2018_01_18_130400_rename_users_groups_to_groups_users.php
luceos: Seems like that's not possible:

https://laravel.com/docs/5.6/eloquent-relationships#many-to-many

tobscure: disregard this, as per our discussion on Discord we are changing our convention to match Laravel's :)

migrations/2018_01_18_132900_create_notifications_from_table.php
luceos: Fixed in https://github.com/flarum/core/pull/1344/commits/1e372f3881aebb6aa930b53201067e593684607c.

src/Http/CookieFactory.php
luceos: Reverted in https://github.com/flarum/core/pull/1344/commits/b2e873ba7bc0c31b1d7fc1c0e96d347c93bebc5f

migrations/2018_01_19_084700_change_settings_value_longblob.php
luceos: Fixed in https://github.com/flarum/core/pull/1344/commits/8779e40ec5ddf978b0bca25b51f470c1a04153f5,

we decided to not change the column type (as discussed in the issue).

migrations/2018_01_18_135000_change_posts_b8_columns.php
luceos: Fixed in https://github.com/flarum/core/pull/1344/commits/5139ce647e9360f89eecf7599fb0e098766b08fa

luceos:
So I've added some constraints. We need to review them.

- [ ] discussions first_post_id will block deletion of a discussion because the Discussion deleted callback is fired after the discussion is being deleted.
- [ ] do we need to add more `cascade` or `set null` instead of strict constraints in the commit?

Relates to https://github.com/flarum/core/pull/1344/commits/90792abf103c5d0ed505edd6dc759728fe1e1f0b

tobscure:
@luceos Yes we should definitely add `cascade` and `set null` to the constraints:

discussions.user_id = set null
discussions.last_posted_user_id = set null
discussions.hidden_user_id = set null

discussions_users.user_id = cascade

posts.user_id = set null
posts.edited_user_id = set null
posts.hidden_user_id = set null

Also some tables need to be renamed as per the Laravel convention:

discussions_users -> discussion_user
groups_permissions -> group_permission
etc.

luceos:
@tobscure 

Discussion::booted

```php
        static::deleted(function (Discussion $discussion) {
            $discussion->raise(new Deleted($discussion));

            // Delete all of the posts in the discussion. Before we delete them
            // in a big batch query, we will loop through them and raise a
            // PostWasDeleted event for each post.
            $posts = $discussion->posts()->allTypes();

            foreach ($posts->cursor() as $post) {
                $discussion->raise(new PostDeleted($post));
            }

            $posts->delete();

            // Delete all of the 'state' records for all of the users who have
            // read the discussion.
            $discussion->readers()->detach();
        });
```

When deleting a discussion from the tests you get the following error:

```
Illuminate\Database\QueryException: SQLSTATE[23000]: Integrity constraint violation: 1451 Cannot delete or update a parent row: a foreign key constraint fails (`flarum`.`discussions`, CONSTRAINT `discussions_first_post_id_foreign` FOREIGN KEY (`first_post_id`) REFERENCES `posts` (`id`)) (SQL: delete from `discussions` where `title` = test - too-obscure)
```

https://travis-ci.org/flarum/core/jobs/405707602#L661

---

Creating a posts.discussion_id cascade would make the above deleted ineffective as the models might not exist.

tobscure:
I reckon just get rid of that chunk of code and let the foreign key cascade take care of deleting the discussion's posts from the database. Relevant: https://github.com/flarum/core/issues/1319 (even though I had changed my mind, this is forcing our hand)

We can probably get rid of `$discussion->readers()->detach()` as well since the FK cascade should take care of that too.

tobscure:
@luceos I've taken care of most of the outstanding issues and tested every obvious bit of user-facing functionality in core. This should be good to go after extensions are updated.

tobscure:
I've done PRs for all core extensions that need them (they are linked in the OP).
I've resolved merge conflicts with master.

@luceos @franzliedke When you have a chance can you please confirm that this is all working from your end! Start with master core/extensions, switch over to the database-changes branches, run `php flarum migrate`, and test it out. If no issues, please go ahead and push those merge buttons :D

franzliedke:
I am guessing the timestamps of all migrations should be updated just before merging, to prevent clashes with migrations from other branches that were merged in the meantime?

(E.g. potentially #1514)

tobscure:
One thing I just noticed: Laravel's default timestamp columns (added via `Illuminate\Database\Schema\Blueprint::timestamps`) use the TIMESTAMP datatype, not DATETIME. Do we want to follow that convention and convert all of our DATETIME columns to TIMESTAMP?

luceos:
@tobscure i think so too.

tobscure:
Have decided against using TIMESTAMP datatype. It is MySQL-specific and causes [issues with doctrine/dbal in migrations](https://github.com/laravel/framework/issues/16526). Further reading: https://softwareengineering.stackexchange.com/questions/126208/datetime-vs-timestamp-in-mysql-and-php-in-practice

migrations/2018_01_30_223700_create_user_user_table.php
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Flarum\Database\Migration;
+use Illuminate\Database\Schema\Blueprint;
+
+return Migration::createTable(
+    'user_user',
```
franzliedke: Where does this table come from?

src/Database/Migration.php
```diff
@@ -125,35 +137,39 @@ public static function addSettings(array $defaults)
...
             },
 
-            'down' => function (Builder $schema) use ($keys) {
+            'down' => function (Builder $schema) use ($rows) {
                 $db = $schema->getConnection();
 
-                foreach ($keys as $key) {
-                    $db->table('permissions')->where($key)->delete();
+                foreach ($rows as $row) {
+                    $db->table('group_permission')->where($row)->delete();
```
franzliedke: Hmm, this is interesting. Doesn't changing the implementation of this migration helper break old migrations using this type? After all, at the point when those migrations were created (and where they should be run in sequence) the table was still called `permissions`.

That tells me we should inline this kind of migration instead of providing helpers for it.

src/Database/Migrator.php
```diff
@@ -63,6 +63,9 @@ public function __construct(
         $this->repository = $repository;
 
         $this->schemaBuilder = $connection->getSchemaBuilder();
+
+        // Workaround for https://github.com/laravel/framework/issues/1186
+        $connection->getDoctrineSchemaManager()->getDatabasePlatform()->registerDoctrineTypeMapping('enum', 'string');
```
franzliedke: As per [this comment](https://github.com/laravel/framework/issues/1186#issuecomment-379640952), this should not be needed anymore?

src/User/User.php
```diff
@@ -128,25 +125,9 @@ public static function boot()
...
-            $posts->delete();
-
-            $user->read()->detach();
-            $user->groups()->detach();
-            $user->accessTokens()->delete();
-            $user->notifications()->delete();
         });
 
-        static::$dispatcher->dispatch(
+        static::$dispatcher->fire(
```
franzliedke: Hmm... as far as I can tell, we are using the dispatcher contract here, which only has `dispatch`. So this should be changed back (along with similar changes, if any).

luceos:
> (Could we add a prefix to our build matrix on Travis? 🤔)

I guess so. It's worth a shot 👍

migrations/2018_01_30_223700_create_user_user_table.php
tobscure: It currently isn't used anywhere. It's for user-user relations, like if an extension wanted to allow users to "follow" each other or become friends. But we could potentially remove it from this PR and only add it in when we write the corresponding core code?

src/Database/Migration.php
tobscure: Yes, you're right.

src/Database/Migrator.php
tobscure: It wasn't working when I tried it, but we should probably double-check

src/User/User.php
tobscure: Agreed

tobscure:
> I know there's a separate PR for the preferences thing, but indices still need to be done, right?

We will do indices in a separate PR? Would rather not delay this one any longer than we need to.

franzliedke:
> We will do indices in a separate PR?

Yep, there's even [a dedicated ticket](https://github.com/flarum/core/issues/127).

migrations/2018_01_30_223700_create_user_user_table.php
franzliedke: Yes, please.

src/Database/Migration.php
tobscure: @franzliedke This is kind of a pain in the butt to inline. It's quite commonly used, and there's lots of boilerplate required (only add the permission if the group exists, and if it's not already present in the group_permission table).

Could we potentially keep the helper but add a second parameter to indicate the name of the table to use (default to `group_permission`, and old migrations can be updated to pass `permissions`)?

Come to think of it, there is a flaw in our current migrator in that core migrations are run first, then migrations from each extension sequentially... but really they should all be run at the same time in chronological order to prevent issues with this kind of stuff?

src/Database/Migrator.php
tobscure: Just tested with DBAL 2.8.0; still not working.

tobscure:
I've taken care of the tweaks suggested by @franzliedke (apart from the addPermissions one, awaiting feedback) - @luceos can you please take care of testing table prefixes?

src/Database/Migration.php
franzliedke: Okay, but how would we keep that helper for extensions (makes sense), but prevent it being used in core (where table names in migrations should never need to be changed)?

We probably need to add a facility for extensions to reference tables by model, so that their migrations don't break when core changes table names?

By the way, both `groups` and 'group_permission` are referenced in that file.

franzliedke:
This needs a rebase.

franzliedke:
Or rather, a merge, since we have other merges in the history.

src/Database/Migration.php
tobscure: > Okay, but how would we keep that helper for extensions (makes sense), but prevent it being used in core (where table names in migrations should never need to be changed)?

We should probably just inline the helper for the old core migrations.

> We probably need to add a facility for extensions to reference tables by model, so that their migrations don't break when core changes table names?

We could just put the onus on extensions to update their migrations to remain compatible, but I suppose it's pretty easy to do the following:

```php
return Migration::addColumns(User::getTable(), []);
```

tobscure:
Update: have added model/JSON-API attribute renaming to this PR as it is completed project-wide. Awaiting feedback from @franzliedke on https://github.com/flarum/core/pull/1344#discussion_r212592101 and @luceos testing table prefixes. In the meantime I will work on an indices PR.

src/Database/Migration.php
franzliedke: Hmm, let's avoid referencing model classes in migrations. Extensions should just update the table / column names in their existing migrations - they won't be run again anyway since the timestamps are saved in the database. In the future, things like this won't happen - or if they do, only in major updates, where SemVer should protect the extensions from breaking, as long as they have proper version constraints.

Other than that, :+1: for your poposal.

franzliedke:
I have added [a separate ticket](https://github.com/flarum/core/issues/1564) for the table prefix thing, no need for that to block this PR.

src/Database/Migration.php
franzliedke: Hmm, we could at least make the logic reusable - this is what I came up with:

~~~php
    public static function addRows($table, array $row, ...$foreignKeys)
    {
        $references = [];

        if (! empty($foreignKeys)) {
            if (count($foreignKeys) % 2 == 1) {
                throw new \InvalidArgumentException(
                    'Two arguments needed for every foreign key'
                );
            }

            while (count($foreignKeys) > 0) {
                $table = array_shift($foreignKeys);
                $keyMap = array_shift($foreignKeys);

                $references[$table] = $keyMap;
            }
        }

        return [
            'up' => function (Builder $schema) use ($table, $row, $references) {
                $db = $schema->getConnection();

                if ($db->table($table)->where($row)->exists()) {
                    return;
                }

                foreach ($references as $table => $keyMap) {
                    $column = key($keyMap);
                    $referencedColumn = current($keyMap);

                    if ($db->table($table)->where($column, $row[$referencedColumn])->doesntExist()) {
                        return;
                    }
                }

                $db->table($table)->insert($row);
            },

            'down' => function (Builder $schema) use ($table, $row) {
                $schema->getConnection()->table($table)->where($row)->delete();
            }
        ];
    }
~~~

You would have to use it like this:
~~~php
Migration::addRow(
    'group_permission',
    [
        'group_id' => Group::MODERATOR_ID,
        'permission' => 'discussion.hidePosts',
    ],
    'groups',
    ['id' => 'group_id']
);
~~~
...with the third and fourth parameter being optional. Not sure if that API is too confusing, or whether we can come up with something better?

franzliedke:
Table prefixes now have a separate PR.

This only needs resolving the migration thing now. :raised_hands:

tobscure:
Which I will do by the end of the week :D

src/Database/Migration.php
tobscure: I've just inlined the core migration in https://github.com/flarum/core/pull/1344/commits/622bf265103e53fe0388b6c97e568010636cbcaa. I think it's simple (and specific) enough that it doesn't warrant a new helper.

tobscure:
@franzliedke Please merge this (and the related PRs) if you are happy.

franzliedke:
@tobscure Should we remove the `addPermissions` migration shortcut then?

tobscure:
@franzliedke No, because extensions can still safely use it – remember, their migrations need to be able to run on the most up-to-date Flarum database, so the `addPermissions` helper will always be appropriate.

luceos:
😄 this got merged.. .. ..

.. ..

omg

PR #1346, opened 2018-01-17
WIP: add (ci) tests

luceos:
I've implemented some changes to our tests, so that we can run installation and interact with the database. As a start I've added a few tests, a.o. to verify the workings of the RegisterUserHandler. The benefit is that I can now confirm that my migrations are all in good order as well.

I'd love to hear your opinion about this.

Please ignore the amount of commits I needed 😊 , halfway I replaced my mariadb with mysql so that local tests could run too.

Please note I removed HHVM and the gitter.im webhook.


tests/Api/Controller/AbstractTestController.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
+namespace Flarum\Tests\Api\Controller;
+
+use Flarum\Http\Controller\ControllerInterface;
+use Flarum\Tests\Test\TestCase;
+use Flarum\User\User;
+use Psr\Http\Message\ResponseInterface;
+
+abstract class AbstractTestController extends TestCase
```
tobscure: Should this be `AbstractControllerTest`?

luceos: .. so this way this class is automatically ignored by phpunit.

tobscure: But shouldn't phpunit ignore abstract classes anyway?

tobscure: or do https://stackoverflow.com/questions/5934347/mark-a-phpunit-test-class-as-to-be-ignored

luceos: > But shouldn't phpunit ignore abstract classes anyway?

Yes. I think so as well. I'll refactor.

luceos: Renamed to `ApiControllerTestCase` giving a proper indication what it is for.

tobscure:
This is looking good!

Instead of hitting mysql should we be mocking repos or anything? (Please ignore if stupid question, complete testing noob here)

luceos:
There are a lot of reasons to use mocking. I'm especially a huge fan of them when calling remote endpoints. However when talking about our system requirements I've always had a preference of running tests against all of them. Using a ci matrix it would be pretty easy not only to test flarum against different php versions, but also different database drivers. One reason for us to do this is the previous choice to use raw sql statements inside migrations. I prefer being on the safe side and have our ci pipelines error whenever we deviate from the supported system requirements.

tests/Test/TestCase.php
```diff
@@ -9,17 +9,21 @@
...
 use Mockery;
 use PHPUnit\Framework\TestCase as Test;
 
 abstract class TestCase extends Test
 {
+    use Concerns\CreatesForum, Concerns\MakesApiRequests;
+
     public function setUp()
     {
         Mockery::close();
```
franzliedke: Wait a second, isn't `Mockery::close` supposed to be called on tear-down?

src/Install/Console/DataProviderInterface.php
```diff
@@ -13,6 +13,8 @@
 
 interface DataProviderInterface
 {
+    public function getDebug(): bool;
```
franzliedke: Can we change this to `getDebugMode()` or `isDebugMode()`, please? :smile:

franzliedke:
On my system, the `Flarum\Tests\Install\InstallCommandTest::allows_forum_installation` test fails with

> Illuminate\Database\QueryException: SQLSTATE[42S01]: Base table or view already exists: 1050 Table 'migrations' already exists (SQL: create table `migrations` (`migration` varchar(255) not null, `extension` varchar(255) null) default character set utf8mb4 collate utf8mb4_unicode_ci)

src/Install/Console/DataProviderInterface.php
luceos: I don't think changing that interface is within the scope of this PR. Feel free to update it in another branch or create an issue.

tests/Test/TestCase.php
luceos: Using mockery::close is part of the old tests. I'll check whether this is still needed.

luceos:
@franzliedke that installation test will fail in case a config.php is found inside /tests/tmp. I'll need to do something about cleaning up after all tests are done.

src/Install/Console/DataProviderInterface.php
luceos: Will do.

tests/Install/InstallCommandTest.php
```diff
@@ -0,0 +1,38 @@
...
+            /** @var InstallCommand $command */
+            $command = $this->app->make(InstallCommand::class);
+            $command->setDataSource($this->configuration);
+
+            $body = fopen('php://temp', 'wb+');
+            $input = new StringInput('');
+            $output = new StreamOutput($body);
+
+            $command->run($input, $output);
+        }
```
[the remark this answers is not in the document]
luceos: Cleaning up might become very expensive. In addition I created a test suite to fully test the installation procedure (using the command). The conditional allows working with an installed Flarum, but also one that still requires installation.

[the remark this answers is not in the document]
franzliedke: I appreciate the intent, but disagree with the solution. :wink: 
Tests should run the same code in any environment.

luceos:
@franzliedke 

- before: https://travis-ci.org/flarum/core/jobs/330278185
- after: https://travis-ci.org/flarum/core/jobs/343747139

977ms vs 9+ seconds.

migrations/2015_12_17_194247_change_settings_value_column_to_text.php
```diff
@@ -21,7 +21,7 @@
 
     'down' => function (Builder $schema) {
         $schema->table('settings', function (Blueprint $table) {
-            $table->binary('value')->change();
+            $table->longText('value')->change();
```
luceos: Figure out whether this one is still needed if we don't do a refresh.

franzliedke:
@luceos Okay, I've applied some of the easy-peasy changes directly on master (.travis.yml updates and test namespace changes), please rebase this when you get the chance.

This makes the diff a little easier to look at and lets us focus at the hard problems in this PR.

Thanks for the great work! :heart:

luceos:
Prepared statements used, reduces testing time to 2.09s 🎉

luceos:
@franzliedke assigned you for review. Last point of discussion is how to handle the installation procedure testing. It now assumes a blank database, this is pretty arduous to handle locally. Do we want to skip those tests in case the database already exists?

tobscure:
Do we have a plan in case we need to use transactions inside the app somewhere in the future?

luceos:
@tobscure illuminate can perfectly handle transaction nesting ;)

tobscure:
@luceos but can MySQL?

luceos:
@tobscure Laravel (illuminate) handles nested transactions per default. Whether this is MySQL (innodb or myisam) or pgsql etc. Also it will bundle transactions to execute in one go, here's some snipppets to convince you:


Bundling:
```php


    /**
     * Commit the active database transaction.
     *
     * @return void
     */
    public function commit()
    {
        if ($this->transactions == 1) {
            $this->getPdo()->commit();
        }

        $this->transactions = max(0, $this->transactions - 1);

        $this->fireConnectionEvent('committed');
    }
```

Using save points (innodb):

```php


    /**
     * Create a transaction within the database.
     *
     * @return void
     */
    protected function createTransaction()
    {
        if ($this->transactions == 0) {
            try {
                $this->getPdo()->beginTransaction();
            } catch (Exception $e) {
                $this->handleBeginTransactionException($e);
            }
        } elseif ($this->transactions >= 1 && $this->queryGrammar->supportsSavepoints()) {
            $this->createSavepoint();
        }
    }
```

PR #1349, opened 2018-01-24
Add console configuration event

clarkwinkelmann:
This is based on the work previously done in https://github.com/flagrow/console

I'm not sure if having a reference to `Flarum\Foundation\Application` in the event is necessary. I remember there was no way to access it otherwise in beta7, but maybe this has changed in beta8 (use case: get the version or the path of the application when registering a command).

Adding this event will allow extensions to hook into the commands offered by `php flarum`


luceos:
I like this very much, this would reduce the code necessary for our console package.

tobscure:
Thanks @clarkwinkelmann!

PR #1352, opened 2018-01-30
Hotfix/tag manager

bryanleetc:


views/app.blade.php
```diff
@@ -15,6 +15,16 @@
...
 
+    <script>
+    // Data layer object for Google Tag Manager
+    window.Oberlo = [{}];
+
+    (function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
+      new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
+      j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
+      'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
+    })(window,document,'script','Oberlo','GTM-PBK2KPC');</script>
```
luceos: Aside from the discussion whether Google Tag Manager belongs in core (I'd say it would be best placed inside an extension), not pushing your own GTM code would be wise 😊

PR #1361, opened 2018-02-09
Extendable notification methods

clarkwinkelmann:
It's currently not easy to add new notifications methods in the settings grid (see what I had to do [here](https://github.com/flagrow/telegram/blob/master/js/forum/src/addNotificationMethod.js#L9) to achieve it in beta7). Simply using an ItemList makes it a lot easier.

While at it I also fixed the jsdoc for notification types that didn't exactly match the code anymore.


tobscure:
Thanks!

PR #1366, opened 2018-02-10
Use Illuminate Session component instead of Symfony

tobscure:
Symfony's component relies on PHP's native session functionality, which
is not ideal. It automatically sets its own cookie headers, resulting in
this issue: https://github.com/flarum/core/issues/1084#issuecomment-364569953

The Illuminate component is more powerful and has a simpler API for
extension with other drivers and such, and fits in nicely with other
components we use (the majority of which are from Illuminate).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,6 +35,7 @@
         "illuminate/filesystem": "5.5.*",
         "illuminate/hashing": "5.5.*",
         "illuminate/mail": "5.5.*",
+        "illuminate/session": "5.5.*",
         "illuminate/support": "5.5.*",
         "illuminate/validation": "5.5.*",
         "illuminate/view": "5.5.*",
--- a/src/Forum/Controller/AbstractOAuth2Controller.php
+++ b/src/Forum/Controller/AbstractOAuth2Controller.php
@@ -62,7 +62,7 @@ abstract class AbstractOAuth2Controller implements ControllerInterface
 
         if (! $code) {
             $authUrl = $this->provider->getAuthorizationUrl($this->getAuthorizationUrlOptions());
-            $session->set('oauth2state', $this->provider->getState());
+            $session->put('oauth2state', $this->provider->getState());
 
             return new RedirectResponse($authUrl.'&display=popup');
         } elseif (! $state || $state !== $session->get('oauth2state')) {
--- a/src/Forum/Controller/LogOutController.php
+++ b/src/Forum/Controller/LogOutController.php
@@ -102,7 +102,7 @@ class LogOutController implements ControllerInterface
 
         // If a valid CSRF token hasn't been provided, show a view which will
         // allow the user to press a button to complete the log out process.
-        $csrfToken = $session->get('csrf_token');
+        $csrfToken = $session->token();
 
         if (array_get($request->getQueryParams(), 'token') !== $csrfToken) {
             $return = array_get($request->getQueryParams(), 'return');
--- a/src/Forum/Controller/ResetPasswordController.php
+++ b/src/Forum/Controller/ResetPasswordController.php
@@ -50,6 +50,6 @@ class ResetPasswordController extends AbstractHtmlController
 
         return $this->view->make('flarum.forum::reset-password')
             ->with('passwordToken', $token->id)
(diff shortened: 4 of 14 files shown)


tobscure:
I'm not so sure I agree with your point about the managers. Isn't it good to be able to give advanced users the flexibility of using different session drivers (and filesystem, logging, etc.)? For example, if we were to be using the equivalent of a `file` session driver by default, that wouldn't scale too well, so an advanced user may wish to swap it out for a `redis` driver.

Furthermore, the abstraction you're talking about doesn't seem that expensive. The `SessionManager` caches and returns an instance of the `Session` interface, which is a single layer on top of whatever driver is being used. Surely the performance decrease would be negligible.

I've only had a quick look, but the only PSR-7 specific library I can find is [psr7-sessions/storageless](https://github.com/psr7-sessions/storageless) – seemingly the equivalent of using Laravel's `cookie` session driver. It looks nice, but it does come with some limitations, namely the assumptions listed in its readme: 

> your sessions are fairly small and contain only few identifiers and some CSRF tokens. Small means `< 400` bytes
> data in your session is **freely readable by the client**

I'm not sure that we can guarantee that extensions won't wish to store larger, sensitive data within sessions.

luceos:
Even though I understand the argument of not releasing a framework, one of the things I strongly am in favour of is extensibility. The drawback of forcing one type of implementation is the restriction you put in place for extensions. A good example of this is the (previous) issue with avatars being stored in the local asset folder, replacing that with (for instance) an aws driven storage is pretty much impossible. Let's not make a habit of closing down our app completely.

franzliedke:
Ah, but we don't limit extensibility by not using a manager class. All an extension needs to do is replace the binding in our IoC container.

Of course other implementations should be usable, that's why we type hint interfaces almost everywhere. But since we currently do not support configuring another driver, we should stick with YAGNI and omit the manager.

luceos:
@franzliedke that's something I can agree on 👍

tobscure:
Ah gotcha. Kinda. Honestly still not really convinced about the benefit to not using the Laravel stuff. We could easily allow that config to be changed via a php file, and get that flexibility for free, without having to write an extension or anything.

Happy to concede to your expertise though. Are you suggesting to still use the Laravel interface and just bypass the service provider/manager? Or something else entirely?

franzliedke:
Hi @tobscure, I've rebased your branch on the latest master and added a commit that takes care of the following:
- Implement session instantiation directly in the middleware (this is important for supporting alternative server environments, such as long-running asynchronous pure-PHP servers, e.g. PHP-PM).
- The service provider instantiates a session handler (this can be replaced by extensions), not an actual session store.
- Garbage collection is moved to the existing middleware.
- Unused session config is removed.
- Avoid using Illuminate's session manager - this is now more or less a side effect of the previous items.

Another question to consider: Should we use Laravel's `EncryptedStore`? I don't see any downsides in doing so...

Let me know what you think. :smile:

tobscure:
Looks good!

Not a huge fan of the session binding being in `UserServiceProvider`? But don't really have a better suggestion for where to put it 🤔

Also kinda liked all the session stuff, including the GC, being together in one middleware... but I'm just nitpicking, ultimately happy for this to go in without any changes if you are :)

franzliedke:
I see your point about GC. Let's make it better at some point, e.g. have a GC manager (or something along those lines) that randomly chooses one of the queries that do GC. (Extensions can then register their own ones.) At that point, having a dedicated middleware suddenly makes sense. :smile:

luceos:
Binding the SessionHandlerInterface only in the UserServiceProvider currently breaks master.

franzliedke:
Quick update: I had forgotten to remove Laravel's `SessionServiceProvider` from this PR, so [I did that now](https://github.com/flarum/core/commit/46f80e8d722261b5b984158ec6a2f940090d8b30).

As for the installer stuff (@luceos' comment above and also [this issue](https://github.com/flarum/core/issues/1370)), it all comes down to how we currently set up the `Application` instance when Flarum has not yet been installed. I have to completely rethink which service providers are registered (and when they are booted) when that is the case. I am on it, though.

PR #1307, opened 2017-12-14
Remove hosted s9e minifying service as it is down

gwillem:
See also #887 and

https://discuss.flarum.org/d/1870-new-javascript-minifiers

Issue #879, opened 2016-03-20
Set up ESLint


tobyzerner:
http://eslint.org/

Need to discuss style rules we want to use.

tobscure:
@franzliedke and I discussed and have decided to use the Airbnb preset with as little modification as possible.

sijad:
isn't https://github.com/airbnb/javascript/blob/master/linters/.eslintrc?
if you still have plan to use airbnb. a few dev dependencies should be added to `flarum-gulp` tough.

dav-is:
I found this config works well with Flarum:
```
{
	"extends": "eslint-config-airbnb",
	"rules": {
		"import/no-unresolved": "off",
		"import/no-extraneous-dependencies": "off",
		"react/jsx-filename-extension": ["error", { "extensions": [".js", ".jsx"] }],
		"react/react-in-jsx-scope": "off",
		"react/prop-types": "off",
		"import/extensions": ["error", "never"]
	},
	"env": {
        	"jquery": true
    	},
	"globals": {
        	"app": false,
        	"m": false
    	}
}
```

sijad:
here's mine: https://github.com/sijad/flarum-ext-links/blob/master/.eslintrc

sijad:
[jsx-filename-extension.md](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-filename-extension.md) consider #1030

franzliedke:
We have decided we want to adopt prettier instead of eslint and will do so as part of (?) the frontend rewrite.

PR #1367, opened 2018-02-10
Webpack

tobscure:
Doing this transition now is required to futureproof the extension API
and will put us in very good stead for future migration to TypeScript.

I've set it up so that minimal changes to the core JS files have been
required so far. The only thing is that import paths with the 'flarum/'
prefix won't work, so I've removed all of those.

This branch/PR will likely end up containing the whole new JS extension
API as well, because using webpack means we have to explicitly expose a
public API (whereas before, when we just exposed our whole JS codebase
as public).

As such, I've started refactoring the very top layer of the codebase
(ie. the Application class, the boot/initialization process). Lots more
coming, and this will be a very good exercise in "one thing at a time"
for me!

TODO:

- [x] core
- [x] akismet
- [x] approval
- [x] auth-facebook
- [x] auth-github
- [x] auth-twitter
- [x] bbcode
- [x] embed
- [x] emoji
- [x] flags
- [x] likes
- [x] lock
- [x] markdown
- [x] mentions
- [x] pusher
- [x] statistics
- [x] sticky
- [x] subscriptions
- [x] suspend
- [x] tags (in progress)

closes #1307


datitisev:
Wouldn't this be easier as a webpack plugin, that way it's all set up like `flarum-gulp`?

tobscure:
Maybe, maybe not, I'll look into it later when I convert extensions

js/forum/src/index.js
```diff
@@ -0,0 +1,26 @@
+import 'expose-loader?$!expose-loader?jQuery!jquery';
```
franzliedke: What do these things do? 😕

tobscure: This [exposes](https://github.com/webpack-contrib/expose-loader) the jquery module as a global variable, so all existing core code would work without changes. Todo includes adding `import $ from 'jquery'` to the top of any file that uses it so we can potentially get rid of this.

franzliedke:
> This branch/PR will likely end up containing the whole new JS extension
API as well, because using webpack means we have to explicitly expose a
public API (whereas before, when we just exposed our whole JS codebase
as public).

Is that unavoidable? The PR is large enough as is, and it would be easier to review these other changes separately?

tobscure:
@franzliedke The problem is that with webpack, extensions can no longer just do:

```js
import Whatever from 'flarum/Whatever';
```

They have to do:

```
import { Whatever } from 'flarum';
```

So to merge this PR, at the very least we must expose a public API and then change all extension `import` statements.

But you're right, the PR will be too big to include the whole new API. I'll see if I can come up with some kind of magic to make extensions continue to work without any changes.

luceos:
Perhaps we can solve this PR nightmare by creating a base branch called "new-frontend" then PR every change against that branch, eg this Webpack change. If we're sure the new-frontend works we can merge that branch to master at specific intervals while still keeping that "project" separate from all other work.

franzliedke:
I'd love it if we could either provide the new API (`import { Whatever } from 'flarum';`) to the old code, before merging Webpack, or provide the old API (`import Whatever from 'flarum/Whatever';`) with the new code, so that we can merge Webpack quickly, but without breaking extensions.

datitisev:
I personally prefer the old API, I think it looks better and is simpler.

franzliedke:
@datitisev I think part of the idea behind the new setup would be that we can limit what is considered "public" API. I.e. extension developers can not consider their work to be on the safe side when they use internal classes. 

This is similar to what we're doing on the backend with extender classes - these are dedicated, stable public API that you as an extension author can rely on. At the same time, we still have the freedom to change internals and refactor to our heart's content, without having to release new major versions and break backwards compatibility all the time.

tobscure:
Good news: I've managed to work out a way to support the old import method within the new Webpack setup, so extensions will need minimal changes to work initially. This will be deprecated and removed by 0.1 stable though.

luceos:
That's great news @tobscure , that will allow extensions to gradually start using webpack 🤗

tobscure:
@luceos Well they still have to upgrade to Webpack to be compatible, but at least they won't have to completely rewrite all their code straight away!

luceos:
Based on your work, would it also be possible to add more code coverage (as a follow up task) like linting for example?

tobscure:
@luceos yes https://github.com/flarum/core/issues/879

sijad:
personally I prefer new API:

`
import {
  Whatever,
  Whatever2,
  Whatever3,
  Whatever4,
} from 'flarum';
` 

than:

`
import Whatever from 'flarum/Whatever';
import Whatever2 from 'flarum/Whatever2';
import Whatever3 from 'flarum/Whatever3';
import Whatever4 from 'flarum/Whatever4';
`

it looks better or maybe I've used to it (because of react(-native) API).

tobscure:
@sijad Yes that's where we're headed! Just painting the old-style imports (deprecated) for backwards compatibility, they will be removed by the stable release.

sijad:
doesn't beta 8 break php backwards compatibility?

tobscure:
Yes, we're not going for complete backwards compatibility, just trying to reduce the developer burden when upgrading.

src/Frontend/Asset/LocaleJsCompiler.php
```diff
@@ -22,23 +22,12 @@ public function setTranslations(array $translations)
 
     public function compile()
     {
-        $output = "
- System.register('locale', [], function(_export) {
-    return {
-        execute: function() {
-            _export('default', function(app) {
-                app.translator.translations = ".json_encode($this->translations).";\n";
+        $output = 'flarum.core.app.translator.translations='.json_encode($this->translations).";\n";
```
franzliedke: @tobscure Please review whether we need some JSON encoding options here, for security. (Compare with Laravel's `CompilesJson.php`, for example.)

tobscure: Good thought. Since this is being outputted in a JavaScript file (not within HTML), no encoding options are required. https://stackoverflow.com/a/6817147/887933

franzliedke: Ah, cheers for double checking.

PR #1372, opened 2018-02-16
FontAwesome v5.0.6

### LINKED ISSUE #1326: Update to Font Awesome v5
A suggestion by [Eldenroot](https://discuss.flarum.org/u/Eldenroot) from [this forum topic](https://discuss.flarum.org/d/8018-update-to-fontawesome-icon-v5)

I must say also those icons look very neat! I like that u can have 3 types of the same icon (solid, regular, light)

[Font awesome v5](https://fontawesome.com/)

afrastgeek:
In this PR, I'm updating flarum to use the latest FontAwesome icon font (v5.0.6).

**After this PR, there's probably a need to update icon name in flarum's extension.**

## Issues that will be fixed by this PR:
- #1326 (Update to Font Awesome v5)

## Changes

> Version 5 has been re-written and re-designed completely from scratch.

Most notable change from version 4, in brief :
- New Prefixes (5 instead of 1)
- Icon Name Changes

Please see: [FontAwesome 5 Upgrade Guide](https://fontawesome.com/how-to-use/upgrading-from-4) for more detailed information

### Icon replacement
In FontAwesome 5, icon replacement also occured.
Some of the old icon now got appended `-alt` in it's icon name, and the old-name got a new icon.
(e.g.: the old `expand` become `expand-alt`)

## BC Breaks

- Icon name must be specified in it's long form class syntax.
  Previously, icon can be used just by specifying it's name.
  E.g: `wrench` now became `fa fa-wrench`.

- Groups Icon
  After updating flarum to a version containing this PR, you must manually update icon name in `groups` database table.

## To-Do

Below is list of task I used for this implementation.

- [x] Use latest font-awesome shim package (12e8a2f)
- [x] Adapt file names and directory structure changes (12e8a2f)
- [x] Adapt icon name changes in less (bd61a62, d13bae3)
- JS Changes:
  - [x] ~~Add new parameter for `iconPrefix` (e.g.: `fa`, `far`, `fab`) to the `icon` helper (0121c1e)~~
  - [x] ~~Update js component to use `iconPrefix` parameter (**Partially implemented** on c8fa44a)~~
  - [x] ~~Adapt icon name changes in js (8f96908, 6af9019)~~
- [x] ~~Display non-default (`far`, `fab`) extension icon in Extension Administration Page. (1b348d2)~~
- [x] Refactor Icon Helper API, see https://github.com/flarum/core/pull/1372#issuecomment-366470195 (75ad8ca)
- [x] Re-checking leftover icon classes (1fa8d6c)

## Size Comparison

In #1326 ,  @tobscure requested for a file-size comparisons.

Here's a table comparing compiled assets files size (in bytes), before and after.
Using `ls -l` and `du -b`.

| Type | File Name | Size Before | Size After | Size After Refactor |
| ------------- | ------------- | -------------:| -------------:| -------------:|
| file      | public/assets/admin-*.css |    74 382 |    78 143 |    78 143 |
| file      | public/assets/admin-*.js  |   507 943 |   507 957 |   505 334 |
| directory | public/assets/fonts/*     | 1 085 661 | 1 688 588 | 1 688 588 |
| file      | public/assets/forum-*.css |   113 425 |   117 661 |   118 082 |
| file      | public/assets/forum-*.js  |   865 708 |   865 906 |   867 466 |

### DIFF
--- a/js/lib/helpers/icon.js
+++ b/js/lib/helpers/icon.js
@@ -1,13 +1,12 @@
 /**
- * The `icon` helper displays a FontAwesome icon. The fa-fw class is applied.
+ * The `icon` helper displays an icon.
  *
- * @param {String} name The name of the icon class, without the `fa-` prefix.
+ * @param {String} fontClass The full icon class, prefix and the icon’s name.
  * @param {Object} attrs Any other attributes to apply.
- * @param {String} prefix The icon class prefix.
  * @return {Object}
  */
-export default function icon(name, attrs = {}) {
-  attrs.className = 'icon ' + name + ' ' + (attrs.className || '');
+export default function icon(fontClass, attrs = {}) {
+  attrs.className = 'icon ' + fontClass + ' ' + (attrs.className || '');
 
   return <i {...attrs}/>;
 }
--- a/js/lib/helpers/icon.js
+++ b/js/lib/helpers/icon.js
@@ -3,10 +3,11 @@
  *
  * @param {String} name The name of the icon class, without the `fa-` prefix.
  * @param {Object} attrs Any other attributes to apply.
+ * @param {String} prefix The icon class prefix.
  * @return {Object}
  */
-export default function icon(name, attrs = {}) {
-  attrs.className = 'icon fa fa-' + name + ' ' + (attrs.className || '');
+export default function icon(name, attrs = {}, prefix = 'fa') {
+  attrs.className = 'icon ' + prefix + ' fa-' + name + ' ' + (attrs.className || '');
 
   return <i {...attrs}/>;
 }
--- a/js/lib/helpers/icon.js
+++ b/js/lib/helpers/icon.js
@@ -6,8 +6,8 @@
  * @param {String} prefix The icon class prefix.
  * @return {Object}
  */
-export default function icon(name, attrs = {}, prefix = 'fa') {
-  attrs.className = 'icon ' + prefix + ' fa-' + name + ' ' + (attrs.className || '');
+export default function icon(name, attrs = {}) {
+  attrs.className = 'icon ' + name + ' ' + (attrs.className || '');
 
   return <i {...attrs}/>;
 }
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
         "php": ">=7.0",
         "dflydev/fig-cookies": "^1.0.2",
         "doctrine/dbal": "^2.5",
-        "components/font-awesome": "^4.6",
+        "components/font-awesome": "^5.0.6",
         "franzl/whoops-middleware": "^0.4.0",
         "illuminate/bus": "5.5.*",
         "illuminate/cache": "5.5.*",
--- a/js/admin/src/components/AdminNav.js
+++ b/js/admin/src/components/AdminNav.js
@@ -41,7 +41,7 @@ export default class AdminNav extends Component {
 
     items.add('basics', AdminLinkButton.component({
       href: app.route('basics'),
-      icon: 'pencil',
+      icon: 'pencil-alt',
       children: app.translator.trans('core.admin.nav.basics_button'),
       description: app.translator.trans('core.admin.nav.basics_text')
     }));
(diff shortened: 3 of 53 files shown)


datitisev:
What if, because the fonts directory is larger now, there could be an option to select what extra icon sets you want to include in the website from the admin interface.
That way, only the general icons are loaded initially, unless extra are added from the admin interface or an extension requires it.
Not necessarily for this PR, but I think it’d be a good idea, if possible.

tobscure:
What an awesomely comprehensive PR! Thanks for your work @afrastgeek 🤩

My only thought is that maybe we should simplify the `icon` helper API and simply pass the full class name as the first parameter. i.e. instead of:

```js
icon('star', undefined, 'far')
```

You'd just do:

```js
icon('far fa-star')
```

It's more transparent, less confusing, and since the parameter values are tightly coupled to FontAwesome anyway, we don't lose any flexibility – we just make that relationship more clear.

Thoughts? @afrastgeek @franzliedke @luceos 

@datitisev I think that's in more the realm of Extensions 😄

franzliedke:
Great stuff, thank you so much @afrastgeek!
Can you quickly explain what the different prefixes do?

I agree with @tobscure's suggestion. Would you mind implementing that in this PR?

tobscure:
@franzliedke Check out the [upgrade guide](https://fontawesome.com/how-to-use/upgrading-from-4) for a good explanation (scroll down to *Icon Name Changes*)

datitisev:
@tobscure Well, I was thinking about it so there would be less resources to load, and extensions can request more as they are needed, instead of always loading all icons and not using over 95% of them.
I guess it could also be done with an extension, though.

afrastgeek:
@datitisev I like the idea. Maybe we need to check what are the font type needed, and load the font file accordingly. But I don't know how it should be done and how much it would improve the performance, so for now i won't implement that in this PR.

---

@tobscure @franzliedke It's my pleasure to contribute to this awesome forum app.

I agree to refactor the icon helper. Previously I was thinking about not disturbing the icon helper API (lesser changes), so i created a new parameter, but when implemented, it gets ugly.

@tobscure suggestion are pretty clean, and yes, it much simpler, and less confusing.

I will update this PR soon, thank you for your input.

---

@franzliedke new prefixes in FontAwesome 5 are represent the icon style.

- `fa` or `fas` for solid icon
- `far` for regular icon (Partially Pro)
- `fal` for light icon (Pro Only)
- `fab` for brand icon

luceos:
I'm with @tobscure on this related to using icon classes. The current implementation has no flexibility to use any other icon pack, using the full class names (including the far, fal and fab prefixes) will allow more flexibility in the long run. 

I'm not sure I like using the shim to support the legacy font awesome set, because it will unnecessarily increase the compiled js file.

afrastgeek:
@luceos yes, I will update the PR to use @tobscure suggestion.
And I'm agree to not use the shim. I believe it only give a short-term advantage.

tobscure:
Sorry can someone explain what the shim is?

afrastgeek:
FontAwesome provide a [Version 4 Shim](https://fontawesome.com/how-to-use/upgrading-from-4#v4-shim). It facilitate user to use old v4 syntax when using the new v5 files.

> Our v4 shim allows you to keep your old version 4 icon names, aliases, and syntax, and we’ll even upgrade you from FA4’s web fonts to FA5’s new SVG with JavaScript implementation.

tobscure:
I see, thanks. I agree we do not need to include the shim.

js/lib/helpers/icon.js
```diff
@@ -3,10 +3,11 @@
  *
  * @param {String} name The name of the icon class, without the `fa-` prefix.
  * @param {Object} attrs Any other attributes to apply.
+ * @param {String} prefix The icon class prefix.
```
franzliedke: Please fix the docblock to match the implementation. ;)

franzliedke: Also, the method description is not completely true anymore.

afrastgeek: I'm sorry, missed this one. Thanks for the catch.

franzliedke:
Thank you so much @afrastgeek. I can only second Toby: what an awesome contribution. :heart:

afrastgeek:
You're welcome! I'm happy doing this.
Let me know if the PR gave some kind of a bug or anything.

Issue #1369, opened 2018-02-14
If I use phone to view the  * others like this, the title will move.


skywalker512:
## Explanation

If I use phone to view the  ```** others like this.``` the title will move.

![_ _20180214095650](https://user-images.githubusercontent.com/12960906/36184305-dfec2c08-116d-11e8-9112-a3c346c31471.gif)

## Technical details

- Website URL where the bug is visible: https://discuss.flarum.org

TurboProgramming:
Isn't discuss pulling the latest commits from dev-master? Shouldn't it be the same?

luceos:
@TurboProgramming discuss is occasionally being updated with dev-master, but this only happens when the calculated risk for problems will be close to zero. The current dev-master is still very volatile, the imminent merge of the database changes making it even more so. Discuss will be updated before beta 8 is tagged though.

PR #1378, opened 2018-03-02
Fix modal title background in media phone

DogSports:
Fix issue #1369

### DIFF
--- a/less/lib/Modal.less
+++ b/less/lib/Modal.less
@@ -152,6 +152,7 @@
     &:before {
       content: " ";
       .header-background();
+      position: absolute;
     }
   }
   .Modal {


franzliedke:
Thank you, @DogSports !

Looks correct, as far as I can tell. I'll let Toby approve and merge this, though.

tobscure:
This is one solution! But it's interesting actually, the way we were intending it to be was for the modal heading to be `position: fixed`, and the bug is that the `.App-primaryControl, .App-titleControl, .App-backControl` components are *not* `position: fixed` as well.

You can see this in https://github.com/flarum/core/blob/master/less/lib/App.less#L43 - these elements are supposed to be `position: fixed` if the app has been scrolled (ie. `.App.affix`), but I obviously didn't realise at the time that the modal element is displayed as a sibling to the `.App` element rather than as a child.

The ultimate solution probably will involve a rethink of whether modal should be a child to `.App` instead of a sibling, but for now I'm happy to merge this as a temporary solution.

PR #1385, opened 2018-03-11
Adding <meta> descriptions and Laravel's slug generator function with transliteration

johannsa:
Without additional dependencies we can fix the issue of some non-ASCII characters disappearing from the slugs (replaced by separator) by using the more complete slug() function on Illuminate\Support\Str. The entire namespace Flarum\Util, containing only Flarum\Util\Str, would be redundant and is marked as deprecate.

**Update: I have repurposed this PR to include my work on adding <meta> descriptions for the core and, on separate PR, for the flarum-ext-tags extension.**

What do you think about this?


johannsa:
@luceos That's fine. I made dropped the class and with this the whole Flarum\Util namespace is also gone.

I confirm that before this PR I did a search and I cannot find any remaining usages of Flarum\Util\Str within the core or the bundled extensions. There were many usages of the Illuminate/Support/Str though.

johannsa:
Note: This **may** close #194
But I think eventually the separator and style of URL (`https://discuss.flarum.org/d/187-word-association-game` vs `https://discuss.flarum.org/d/187`) should be made configurable.

src/Discussion/Discussion.php
```diff
@@ -461,6 +463,8 @@ public static function setStateUser(User $user)
     protected function setTitleAttribute($title)
     {
         $this->attributes['title'] = $title;
-        $this->slug = Str::slug($title);
+        $app = Application::getInstance();
+        $locale = $app->make(LocaleManager::class)->getLocale() ?? 'en';
```
franzliedke: This locale depends on the currently active user. Shouldn't we *always* use either `en` or at least the board's default language here, so that we get consistent results?

luceos: Good point, I think the default locale would make the most sense.

johannsa:
@franzliedke, @luceos The changes requested were made. Also, I am including the changes I was making to add <meta> description.

src/Discussion/Discussion.php
```diff
@@ -461,6 +464,24 @@ public static function setStateUser(User $user)
...
+     * Get the description.
+     *
+     * Returns the stored description (if any) or generates one based on startPost´s content.
+     *
+     * @param string $description
+     * @return string
+     */
+    protected function getDescriptionAttribute($description)
+    {
+        $description = ($description == null) ? $this->startPost->content : $description;
```
tobscure: This is gonna be inefficient when listing discussions D:

johannsa: @tobscure Are you referring to `$this->startPost->content`, right? The other option is to always have one (from discussion creation) on the database, not only when you (i.e.: via 3rd party extension) want to have a custom non-generated description there.

tobscure: Yeah I think we need to do something like this - any better ideas @franzliedke?

johannsa:
**Tasks**
- [ ] Uncommit the dist files.
- [ ] Make Flarum to store a generated (first 300 characters plus '...' if needed) description upon discussion creation and/or update.
- [ ] Make the migration to generate (and store) description for existing discussions.

franzliedke:
@johannsa Are you still interested in finishing this? We'd highly appreciate it. :smiley:

src/Discussion/Discussion.php
luceos: The line we're talking about could also have been improved with the null coalescence operator.

I, however, agree to store a generated description, but only if none is set beforehand.

tobscure:
@johannsa If you're interested in finishing this, could you please extract the **slug generation** changes into a separate PR which I consider ready to be merged. The **description** stuff needs more work.

franzliedke:
I extracted the switch to Laravel's slugger (that has basic transliteration support) into a new PR - #1975.

PR #1392, opened 2018-03-19
Always return a promise in PostStream#update

### LINKED ISSUE #1390: "Cannot read property 'then' of undefined" when posting reply
I posted a new post in https://discuss.flarum.org/d/6542-reactions-by-reflar, and the composer got stuck on loading while the post was successfully posted.

image

Error in console:
```js
Uncaught TypeError: Cannot read property 'then' of undefined
    at forum-def494fc.js:2133
    at forum-def494fc.js:553
    at thennable (forum-def494fc.js:544)
    at fire (forum-def494fc.js:551)
    at Deferred.self.resolve (forum-def494fc.js:526)
    at forum-def494fc.js:536
    at Array.map (<anonymous>)
    at finish (forum-def494fc.js:536)
    at forum-def494fc.js:558
    at thennable (forum-def494fc.js:544)
```

I looked at the error source and found it here:
```js
app.store
  .createRecord("posts")
  .save(data)
  .then(post => {
    if (app.viewingDiscussion(discussion)) {
      app.current.stream.update().then(function() { // <---- exception is thrown here
        return app.current.stream.goToNumber(post.number());
      });
    } else {
      var alert = void 0;
      var viewButton = Button.component({
        className: "Button Button--link",
        children: app.translator.trans("core.forum.composer_reply.view_button"),
        onclick: function onclick() {
          m.route(app.route.post(post));
          app.alerts.dismiss(alert);
        }
      });
    }
  });

```

datitisev:
_Most likely_ fixes #1390

Replaces `return` with `return Promise.resolve()` so a Promise is returned, but not an error (Promise.reject)

### DIFF
--- a/js/forum/src/components/PostStream.js
+++ b/js/forum/src/components/PostStream.js
@@ -122,7 +122,7 @@ class PostStream extends Component {
    * @public
    */
   update() {
-    if (!this.viewingEnd) return;
+    if (!this.viewingEnd) return Promise.resolve();
 
     this.visibleEnd = this.count();
 
--- a/js/forum/src/components/PostStream.js
+++ b/js/forum/src/components/PostStream.js
@@ -124,7 +124,6 @@ class PostStream extends Component {
   update() {
     if (!this.viewingEnd) return m.deferred().resolve().promise;
 
-
     this.visibleEnd = this.count();
 
     return this.loadRange(this.visibleStart, this.visibleEnd).then(() => m.redraw());
--- a/js/forum/src/components/PostStream.js
+++ b/js/forum/src/components/PostStream.js
@@ -122,7 +122,8 @@ class PostStream extends Component {
    * @public
    */
   update() {
-    if (!this.viewingEnd) return Promise.resolve();
+    if (!this.viewingEnd) return m.deferred().resolve().promise;
+
 
     this.visibleEnd = this.count();
 


franzliedke:
Can we rely on `Promise` being available globally? Does that need to be polyfilled? Is it already? Can we import it instead?

datitisev:
Promises are available in ES5, and that’s what we are using. They are probably Pollyfilled when compiling the code into the dist file anyway.

franzliedke:
I am not going to rely on a "probably". :wink: Let's wait for Toby's expertise.

@datitisev Thanks for jumping on this, in any case! :+1:

tobscure:
Mithril 0.2 doesn't use native Promises (nor do we include a polyfill). It uses its own implementation called [`m.deferred`](https://mithril.js.org/archive/v0.2.5/mithril.deferred.html).

Basically, I think it should be this:

```js
const deferred = m.deferred();
deferred.resolve();
return deferred.promise;
```

datitisev:
I think babel pollyfills them, as there are Promises in forum/dist.js file:

image

Seems like SystemJS uses Promises, so it should be fine if we do as well.
```sh
js/forum/dist/app.js-657-var System = {
js/forum/dist/app.js-658-    set: set,
js/forum/dist/app.js-659-    get: get,
js/forum/dist/app.js-660-    has: has,
js/forum/dist/app.js-661-    import: function(name) {
js/forum/dist/app.js:662:        return new Promise(function(resolve, reject) {
js/forum/dist/app.js-663-            var normalizedName = normalizeName(name, []);
js/forum/dist/app.js-664-            var mod = get(normalizedName);
js/forum/dist/app.js-665-            return mod ? resolve(mod) : load(name).then(function () {
js/forum/dist/app.js-666-                return get(normalizedName);
js/forum/dist/app.js-667-            });
```

datitisev:
Done!

tobscure:
Great thanks! According to the [`m.deferred`](https://mithril.js.org/archive/v0.2.5/mithril.deferred.html#signature) documentation, the `resolve` method returns void, but looks like you're correct and that's not :)

datitisev:
@tobscure Yeah, for a second I was going to do it like you said (and like the documentation said), but then looking at other code in PostStream made it look like one line was fine.

PR #1396, opened 2018-03-21
public path should not be writable, but base path should

luceos:
Franz please review.


luceos:
@franzliedke adding an empty config.php could also solve that.

tobscure:
@franzliedke could do, but is there any disadvantage to the base path being writeable - as long as all the files and subfolders are not writeable?

franzliedke:
Right, that makes sense. To be honest, file and folder permissions are not my strong suit. :wink:

PR #1397, opened 2018-03-21
Installer: add check for file existence & fix path resolving

datitisev:
- Adds check for if file exists
- Uses path output from `realpath($path) ?: stream_resolve_include_path($path) ?: $path)`
    - `realpath`, returns FALSE if 
        - path doesn't exist
        - running script doesn't have executable permissions on all directories in the hierarchy
    - `stream_resolve_include_path`, returns FALSE if
         - path doesn't exist (as far as I can tell)
    - `$path`

image


franzliedke:
Why would these directories not exist, though? They are in the flarum/flarum repository.

I don't really want to support the case of flarum/flarum not being in sync with flarum/core - that's mostly a beta problem anyway...

luceos:
@franzliedke that might be what you want, but let's consider the real life situation where people will do an actual upgrade from a previous version. Having this scenario only reduces support load in the long run.

Please understand I agree with not supporting all scenarios, but both @datitisev and I ran into this because we're both rather stupid and simply tried upgrading an existing installation.

src/Install/Prerequisite/WritablePaths.php
```diff
@@ -23,9 +23,15 @@ public function __construct(array $paths)
     public function check()
     {
         foreach ($this->paths as $path) {
-            if (! is_writable($path)) {
+            $filepath = (realpath($path) ?: stream_resolve_include_path($path) ?: $path);
```
franzliedke: Why the `stream_resolve_include_path`?

datitisev: It was the only other function I found similar to realpath that would work for permission errors. Both realpath and stream_resolve_include_path fail, however, if the dir doesn’t exist, so maybe just calculate that in is_writable?
And also maybe realpath is unnecessary, as we are using it to show the path of a nonwritable directory, when it doesn’t even work with nonwritable directories

datitisev:
@franzliedke Also, if someone accidentally deleted a directory, they would only get the permissions error, making them confused.
Furthermore, this also fixes the permissions error not showing a path because real path returned false when called with a nonwritable path, the point of the error.

datitisev:
Anything needed for this to be merged, or just waiting on related master development?

src/Install/Prerequisite/WritablePaths.php
tobscure: You're right, `realpath` will only work if the directory exists so not ideal here. `stream_resolve_include_path` is really doing something else completely, not right for this situation.

I would suggest adding a private function to this class called `getAbsolutePath` inspired by the function in this user comment: http://php.net/manual/en/function.realpath.php#84012

franzliedke:
@datitisev Thanks bunches!

datitisev:
@franzliedke No problem! 🙂

PR #1402, opened 2018-04-10
fixes log in without remember on master

luceos:
Currently dev-master has an issue with log in. This PR fixes that. The cause is a cookie generated `flarum_flarum_session` the double prefixing is generated inside the CookieFactory.

### DIFF
--- a/src/Http/Middleware/StartSession.php
+++ b/src/Http/Middleware/StartSession.php
@@ -40,7 +40,9 @@ class StartSession implements MiddlewareInterface
     protected $config;
 
     /**
-     * @param CookieFactory $cookie
+     * @param SessionHandlerInterface $handler
+     * @param CookieFactory           $cookie
+     * @param ConfigRepository        $config
      */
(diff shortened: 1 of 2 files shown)


src/Http/Middleware/StartSession.php
```diff
@@ -70,7 +72,7 @@ private function makeSession(Request $request)
         $cookieName = $this->cookie->getName($this->config['cookie']);
 
         return new Store(
-            $cookieName,
```
franzliedke: Can we let `$this->config['cookie']` be the value of `$cookieName` now? Alternatively, just inline it. ;)

tobscure: `$cookieName` is still used on line 77 though?

franzliedke: Precisely. Either make it be used twice, or inline the temp var, please. :wink:

tobscure: Ah, gotcha :)

franzliedke:
Feel free to merge, with or without my proposed change.

PR #1403, opened 2018-04-10
deprecations and type hinting

luceos:
removes some deprecated code and adds some type hinting compliant with php 7.0


tobscure:
I have just fixed this issue in an alternative way by upgrading everything to PSR-15 / stratigility 3.0, will PR later today. Thanks though :)

franzliedke:
See [my comment on that PR](https://github.com/flarum/core/pull/1404#issuecomment-380350809), I am not convinced we should use Stratigility 3.x.

luceos:
I guess we can move forward by merging this one as well then?

franzliedke:
This will be made obsolete when I finish #1416.

PR #1404, opened 2018-04-11
Use PSR-15 for middleware, update to zend-stratigility 3.x

tobscure:


franzliedke:
I am sorry, but I am against this.

I have been there, too - had it on my local machine and everything. But requiring PHP 7.1 turned me off. So I looked around for alternatives, found [northwoods/broker](https://github.com/northwoods/broker), where I submitted a [PR for path prefixing](https://github.com/northwoods/broker/pull/7) so that we can use that instead...

luceos:
I don't think php 7.1 should be that much of an issue. We are already upping beta 8 to 7.0; 7.1 is much further progressed and has even an increased performance gain; what are you arguments against it @franzliedke ?

tobscure:
One thing to consider is that by the time we release "stable" PHP 7.0 will likely be very close to EOL. http://php.net/supported-versions.php

That said, I don't mind what implementation we use as long as we can use PSR-15.

tobscure:
Also my apologise @franzliedke, I was not aware you were already updating to PSR-15. That's probably my bad, I might've missed you mentioning it? But nonetheless `master` was in a broken state and we can't do a whole lot while that's the case.

franzliedke:
The middleware stuff worked on my machine (hehe). What exactly was broken?

PR #1405, opened 2018-04-13
Improved foundational backend unit tests

luceos:
This deprecates #1346 

This PR only adds the bare necessary changes for tests to begin with.

Connected PRs adding more tests:

- Api: #1406


src/Install/Console/InstallCommand.php
```diff
@@ -138,6 +138,7 @@ public function setDataSource(DataProviderInterface $dataSource)
     protected function install()
     {
         try {
+            $this->debug = $this->dataSource->isDebugMode();
```
franzliedke: Can we pass this to `$this->storeConfiguration()` instead of storing it as an instance variable? You can do this on master, if you agree.

tests/Test/Concerns/RetrievesAuthorizedUsers.php
```diff
@@ -0,0 +1,37 @@
...
+    ];
+
+    public function getAdminUser()
+    {
+        return User::find(1);
+    }
+
+    public function getNormalUser()
+    {
+        User::unguard();
```
franzliedke: Shouldn't the following code be run inside a `User::unguarded()` closure? Currently, the `User` model stays in unguarded model throughout the tests, right?

src/Install/Console/InstallCommand.php
luceos: Let's refactor that command as a whole after merging 👍

tests/Install/DefaultInstallationCommandTest.php
```diff
@@ -0,0 +1,51 @@
...
+use Flarum\Install\Console\InstallCommand;
+use Flarum\Install\InstallServiceProvider;
+use Flarum\Tests\Test\TestCase;
+use Flarum\User\User;
+use Symfony\Component\Console\Input\StringInput;
+use Symfony\Component\Console\Output\StreamOutput;
+
+class DefaultInstallationCommandTest extends TestCase
+{
+    protected $isInstalled = false;
```
franzliedke: Still needed?

franzliedke: Never mind, found it.

PR #1408, opened 2018-04-14
Fix profile page layout jumping when scrolling

clarkwinkelmann:
The first post height was affected by the floating block when affix wasn't active.

This issue is best observed when there are many menu items on the left (link to settings, link to Masquerade, ...) and the first post listed for the user is very short.

For some reason the floating property of the .sideNav caused the first post to always be as tall as the menu. But then when you scroll and the affix effect is applied to the nav, this side effect cease and this cause a jump in the list of posts when the first post gets back to intended height.

Here's how the bug looked like:

without scrolling past the affix trigger point:
image
after scrolling a little and triggering the affix on the nav:
image

I tried a few things to fix it and I don't think my solution is very pretty. But setting the height of the container to 0 fixes the issue. Tested in Firefox and Chrome.

### DIFF
--- a/less/common/sideNav.less
+++ b/less/common/sideNav.less
@@ -98,6 +98,7 @@
 @media @desktop-up {
   .sideNav {
     float: left;
+    height: 0;
 
     &, > ul {
       width: 190px;
--- a/less/common/sideNav.less
+++ b/less/common/sideNav.less
@@ -98,6 +98,9 @@
 @media @desktop-up {
   .sideNav {
     float: left;
+
+    // Prevents the sidenav from changing the height of the content beside it because of floating wizardry
+    // Without this, the content on the right (like a post) would have a height of at least the height of the sidenav
     height: 0;
 
     &, > ul {


pflstr:
The reason for the space at the bottom is the CSS rule

```
.Post::after {
    clear: both;
}
```

Once the `ul` inside `<nav class="sideNav UserPage-nav">` gets the `.affix` and thus becomes `position: fixed`, the abobe `clear: both` looses it's undesired effect.

franzliedke:
Thanks for debugging this!

@pflstr Why does a `height: 0` fix this problem then?
@clarkwinkelmann Can I ask you to rebase this on the latest master? :see_no_evil:

tobscure:
`height: 0` seems a bit hacky, so if we're going to use it we should at least add a comment explaining its purpose.

clarkwinkelmann:
I put my rebase talents into action and hope I didn't break anything :innocent: 

I've also added a comment, I hope it's useful. The problem is that I don't really know myself why this fixes the issue (floating wizardry), I can only explain what it solves :see_no_evil:

pflstr:
> Why does a `height: 0` fix this problem then?

I have no idea.

For an elegant fix: Just remove:

```
.Post::after {
    clear: both;
}
```

Overriding it with `.Post::after { clear: none }` solved it for me.

What's the purpose of the `clear: both`?

tobscure:
@pflstr `clear: both` is necessary after each post because posts contain floated elements in their footer which need to be cleared. `overflow: hidden` is not an option because there are elements such as dropdown menus that would be clipped.

The issue was happening because the sideNav was floated left, and so the `clear: both` after the first post was clearing the height of the floated sideNav. By setting that sideNav's height to 0, the problem was resolved, but this is too hacky for my liking.

Fixed by switching to flexbox 🤘 6a6b9ac

PR #1411, opened 2018-04-25
allowing configurable flood gate

luceos:
This will allow any extension to return a boolean to prevent or enforce opening or closing the gate. Stacked listeners might be .. problematic but let's get one solution in for now.


tobscure:
👌

PR #1416, opened 2018-05-01
[WIP] Sites foundational rewrite

luceos:
This is a WIP PR partially finished by @franzliedke. It's to see what's in it, perhaps Franz is able to share what still needs to be done?

### DIFF
--- a/src/Foundation/InstalledApp.php
+++ b/src/Foundation/InstalledApp.php
@@ -1,0 +1,132 @@
+<?php
+
+/*
+ * This file is part of Flarum.
+ *
+ * (c) Toby Zerner <toby.zerner@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Flarum\Foundation;
+
+use Flarum\Database\Console\GenerateMigrationCommand;
+use Flarum\Database\Console\MigrateCommand;
+use Flarum\Database\Console\ResetCommand;
+use Flarum\Foundation\Console\CacheClearCommand;
+use Flarum\Foundation\Console\InfoCommand;
+use Flarum\Http\Middleware\DispatchRoute;
+use Flarum\Settings\SettingsRepositoryInterface;
+use Zend\Diactoros\Response\HtmlResponse;
+use Zend\Stratigility\MiddlewarePipe;
+use function Zend\Stratigility\middleware;
+use function Zend\Stratigility\path;
+
+class InstalledApp implements AppInterface
+{
+    /**
+     * @var Application
+     */
+    protected $laravel;
+
+    /**
+     * @var array
+     */
+    protected $config;
+
+    public function __construct(Application $laravel, array $config)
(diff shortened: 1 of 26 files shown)


luceos:
@franzliedke I've looked at this and must say it's a pretty tough rewrite. Well done. I haven't been able to see where you wanted to go in terms of Site instantiation yet as I guess you have some changes in flarum/flarum that go along with all this?

src/Foundation/InstalledApp.php
```diff
@@ -0,0 +1,125 @@
...
+        if ($this->inMaintenanceMode()) {
+            return $this->getMaintenanceMiddleware();
+        } elseif ($this->needsUpdate()) {
+            return $this->getUpdaterMiddleware();
+        }
+
+        $pipe = new MiddlewarePipe;
+
+        $api = parse_url($this->laravel->url('api'), PHP_URL_PATH);
+        $pipe->pipe(path($api, $this->laravel->make('flarum.api.middleware')));
```
sijad: this can be simplified to:

```
$pipe->pipe(path('/api', $this->laravel->make('flarum.api.middleware')));
```

getting site url and parse it to get only its path doesn't make sense.

src/Foundation/InstalledApp.php
```diff
@@ -0,0 +1,125 @@
...
+
+        $admin = parse_url($this->laravel->url('admin'), PHP_URL_PATH);
+        $pipe->pipe(path($admin, $this->laravel->make('flarum.admin.middleware')));
```
sijad: this also can be simplified like above,
also I think this shouldn't be hardcoded. changing admin URL is very common for security purposes.

src/Foundation/InstalledApp.php
```diff
@@ -0,0 +1,125 @@
...
+
+        $forum = parse_url($this->laravel->url(''), PHP_URL_PATH) ?: '/';
+        $pipe->pipe(path($forum, $this->laravel->make('flarum.forum.middleware')));
```
sijad: should be `$pipe->pipe(path('/', $this->laravel->make('flarum.forum.middleware')));`

src/Foundation/InstalledApp.php
```diff
@@ -0,0 +1,125 @@
...
+    protected $laravel;
+
+    /**
+     * @var array
+     */
+    protected $config;
+
+    public function __construct(Application $laravel, array $config)
+    {
+        $this->laravel = $laravel;
```
luceos: `$laravel` isn't really the right variable name to use here. I understand wanting to use it, but it would highly confuse people. Also, everywhere else in Flarum we have assigned Application to the variable name `$app` why move away from that here? Just because the class is called `InstalledApp`?

tobscure:
@franzliedke What is the status of this?

franzliedke:
I need to rebase this and ask for input on some things. It's the next thing on my list.

franzliedke:
Something's cooking... :man_cook:

PR #1417, opened 2018-05-02
fixes session loading when forum not yet installed

luceos:
Currently the session isn't configured whenever Flarum is not yet installed. This PR offers an intermediate fix until Franz' full rewrite passes, it might fix it, it might not.

The cause of the SessionHandlerInterface not being bound is because the UserServiceProvider is only registered when the Forum is installed. Simply registering that UserServiceProvider in all situations doesn't work because it subscribes the `EmailConfirmationMailer` which requires the `SettingsRepositoryInterface` and that one requires a database which hasn't been set up yet. One of the drawbacks of resolving from IoC in listener constructs I guess 🤷‍♂️ 

Let me know what you think!

### DIFF
--- a/src/Http/Server.php
+++ b/src/Http/Server.php
@@ -113,6 +113,7 @@ class Server
     protected function getInstallerMiddleware(MiddlewarePipe $pipe)
     {
         $this->app->register(InstallServiceProvider::class);
+        $this->app->register(SessionServiceProvider::class);
 
         // FIXME: Re-enable HandleErrors middleware, if possible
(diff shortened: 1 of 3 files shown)


franzliedke:
My temporary fix for this was removing the `StartSession` middleware from the installer's middleware stack. This breaks the auto-login feature of the web installer, but I guess that could be considered less invasive...

luceos:
@franzliedke thanks for clarifying, would it be okay for me to push that change so I can continue with #1236 (no way to install currently) 😊 

And thanks for popping in, much appreciated!

PR #1420, opened 2018-05-04
Use ItemList for signup and login modals

clarkwinkelmann:
This will make extending the signup and register a lot easier.

Currently we had to either:

Hack into Mithril template:

https://github.com/flagrow/terms/blob/2c76e691c757b6e9a88293a40e1e61300143505c/js/forum/src/addFieldsToRegister.js#L49

Inject with jQuery:

https://github.com/sijad/flarum-ext-recaptcha/blob/b3935410a3bbf5ec3c966e7a8bf843a779558bf6/js/forum/src/main.js#L22

With the new solution we can now just:

    extend(SignUpModal.prototype, 'fields', items => {
        items.add('somefield', m('.Form-group', [
            m('input.FormControl', {
                type: 'text',
            }),
        ]));
    });

    extend(LogInModal.prototype, 'fields', items => {
        items.add('somefield', m('.Form-group', [
            m('input.FormControl', {
                type: 'text',
            }),
        ]), 25);
    });

I used a few hard-coded priorities in the ItemList so it's easier to inject fields at the place you want between core fields. Maybe you'll prefer to use other values.


tobscure:
Perfect, thanks @clarkwinkelmann!

PR #1430, opened 2018-05-14
adds a few additional api controller tests

luceos:
Some additional tests.

Ran into an issue with the ShowDiscussionControllerTest that a Guest is unable to view a discussion. @flarum/core any suggestions what might be the cause?

Please note the author is able to view the discussion. I can't identify whether discussions are only showing up when they have at least one comment post..


luceos:
```php
// Hide private discussions by default.
        $query->where(function ($query) use ($actor) {
            $query->where('discussions.is_private', false)
                ->orWhere(function ($query) use ($actor) {
                    $this->events->dispatch(
                        new ScopeModelVisibility($query, $actor, 'viewPrivate')
                    );
                });
        });

```

this part in the policy is removing the result, found this by using:

```php
if ($query->find(7)) dd($actor);
```

still investigating

PR #1432, opened 2018-05-14
Configure external links before dispatching textformatter configuring event

clarkwinkelmann:
This way extensions can override the link attributes.

I'm not sure why this was done in the opposite order until now. That part of the code seems to be 3 years old. It seems a lot more logical for me to configure the links before dispatching the event, so that extension have an opportunity to override it.

Use case include extensions that might want to remove set `target` or `rel` to another value. It's impossible right now as the template is always overridden after the event.


tobscure:
🤦‍♂️

luceos:
:blush:

PR #1436, opened 2018-05-21
Properly set comments_count and discussions_count

### LINKED ISSUE #1104: comments_count total incorrect for users
Hi,

Sorry for my english.

I have finding that when i remove a comment, in "users" table the value to "comments_count" line is set to n-1 and when i remove definitly the comment, the value to "comments_count" line is again seted to n-1.

So the total comments is incorrect because for one comment deleted, -2 is subtracted from the "comments_count" current total

I think that is not normal, what thinking you ?

Thank you :)

datitisev:
I've just added a quick check to make sure the post wasn't hidden before being deleted (would happen via extensions, I assume), and if it wasn't, subtract from the `comments_count`.

Ultimately we would most likely want to make the `comments_count` accurate and take into account posts with `is_private` and `is_hidden` and decide whether we want to see those, but for now this is a quick hack to fix it.

Closes #1104.


tobscure:
Maybe we should make `UserMetadataUpdater` more like `DiscussionMetadataUpdater` and add `refreshCommentsCount` and `refreshDiscussionsCount` methods to the `User` model?

luceos:
I'm not sure @tobscure ; these models are already pretty large. It can be a design choice to move it there, I personally like the cleanliness of the Meta Updater and would rather move more logic there 🤷‍♂️

src/User/UserMetadataUpdater.php
```diff
@@ -49,7 +49,9 @@ public function whenPostWasPosted(Posted $event)
      */
     public function whenPostWasDeleted(PostDeleted $event)
     {
-        $this->updateCommentsCount($event->post, -1);
+        if (! isset($event->post->hide_time)) {
```
luceos: There's no need to do an `isset` on any Eloquent Model property. These are all processed by `getAttribute` and if the property does not exist will return `null`. So here the value is either:

- `null`
- a `\Carbon\Carbon` or most likely `\Illuminate\Support\Carbon` object

tobscure:
@luceos not that big of a deal, but I would argue that the logic to "refresh the user's comments count" belongs in the model (it could be useful elsewhere), whereas the purpose of the metadata updater listener is simply to hook up the events with that logic.

In any case my main point is that we should refresh the counts in response to events, rather than increment/decrement them

luceos:
@tobscure agreed, that makes more sense.

datitisev:
I think that’d be a good idea. Should I try to incorporate that into this PR? It fixes the same problem but other imminent ones too, so...

src/User/UserMetadataUpdater.php
franzliedke: But these will go through [`__isset()`](https://github.com/laravel/framework/blob/39165ddea22cf837a34a8b01e4974cebb5c1af6d/src/Illuminate/Database/Eloquent/Model.php#L1480), which returns `false` if `getAttribute` returns `null`.

tobscure:
Yes please!

datitisev:
I implemented the changes into this PR.

I'm not sure if I should still be checking for the user's existence, however. Should I keep these methods in UserMetadataUpdater and just call the user's methods after checking for its existence?

image

luceos:
@datitisev the previous logic would use the `$post->user`, your new logic uses the actor. Please note that these two can be entirely different entities.

datitisev:
@luceos Oops! My bad 🤦‍♂️. Fixed!

tobscure:
And we'll still need a check for if those users exist. Because if they don't, calling `refreshXCount` on them is not gonna work :)

datitisev:
@tobscure Is this good? I'm not sure if a check for deleted is needed, as DiscussionMetadataUpdater doesn't have checks for delete events... and I also didn't want to repeat the same code so I put the code into those private methods.

tobscure:
Private methods are good... we don't need to check if the discussion/post exists though, but rather the user that we want to update!

datitisev:
I knew that... lol.... 😶 Third time's the charm! 😂

luceos:
Instead of doing a check whether the user exists we can also use the optional Helper instead, see https://laravel.com/docs/5.6/helpers#method-optional

datitisev:
Guessing yes, as Toby upvoted your comment, so I changed it to use that instead.

matteocontrini:
Is this scheduled for beta 8? :)

datitisev:
@matteocontrini I think so, yes.
@luceos Your review is outdated. Care to review again?

franzliedke:
@matteocontrini The underlying bug report (#1104) was not explicitly scheduled for the next beta. Before the release, we will go through all the open PRs again and get many of them merged, though - so the chances are good. :wink:

src/User/User.php
```diff
@@ -747,4 +748,30 @@ public static function getNotificationPreferenceKey($type, $method)
...
+    }
+
+    /**
+     * Refresh the user's comments count.
+     *
+     * @return $this
+     */
+    public function refreshDiscussionsCount()
+    {
+        $this->discussions_count = Discussion::where('start_user_id', $this->id)->count();
```
luceos: I'm unsure why we are using the relation `posts()` in `refreshCommentsCount` and are using a direct query here. I would personally prefer a new relation `discussionsStarted`, what do you think @flarum/core ?

If we do, it needs:

- [ ] the relation
- [ ] the phpdoc change

tobscure: Yep. Not sure why we don't already have it. It can just be called `discussions`.

datitisev:
I... think I created the relation correctly.
Why? Because it works 😆.

datitisev:
No problem! 😄

PR #1443, opened 2018-05-29
Replace ControllerInterface with PSR-15 interface

franzliedke:
The custom interface already had the same signature as the
one from the standard (except for the return type hint), so
why not use that one now? :)


tobscure:
Love it :D

PR #1445, opened 2018-06-03
Migrations: Simplify interface

franzliedke:
Extracted from #1416. To prevent people from depending on the old behavior (and reduce the amount of changes that need to be reviewed in that PR), let's get this part merged first.

1. This unifies the implementations of all pre-defined migration helpers.
2. This improves their implementation: By not pulling dependencies such as a settings repository from the IoC container (one that is decorated with caching), migrations more clearly and explicitly define their desired implementation: We want to write this setting to the database, period!
3. Most importantly, IoC configuration can be simplified in different contexts, e.g. when installing.
4. I really disliked having that `app()->call()` in the migrator class. It contradicts the point of dependency injection, by hiding what dependencies are required.

(All extensions have previously been converted to only depend on schema builder in closure migrations, see https://github.com/flarum/flarum-ext-tags/commit/1d1b1b5176a17e8a9a455c3a267a699d1bf2df2a for an example.)

### DIFF
--- a/src/Database/DatabaseMigrationRepository.php
+++ b/src/Database/DatabaseMigrationRepository.php
@@ -11,16 +11,16 @@
 
 namespace Flarum\Database;
 
-use Illuminate\Database\ConnectionResolverInterface as Resolver;
+use Illuminate\Database\ConnectionInterface;
 
 class DatabaseMigrationRepository implements MigrationRepositoryInterface
 {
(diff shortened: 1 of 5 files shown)


tobscure:
Looks good!

tobscure:
Although, see discussion in #1446

PR #1450, opened 2018-06-05
[WIP] Frontend/Assets refactor

### LINKED ISSUE #837: Cache clean command
Let's re-add some cache cleaning through the artisan command.

This will prevent possible "oops" issues with users deleting the wrong folders. Folders/files to clear:
- assets/*.js
- assets/*.css
- storage/cache/\* (recursive)

### LINKED ISSUE #865: Add more control over the assets
Add some event to control assets:
- Add event to check file before add in [addFile](https://github.com/flarum/core/blob/master/src/Asset/AssetManager.php#L49) function
- Add event for [getCssFile](https://github.com/flarum/core/blob/master/src/Asset/AssetManager.php#L95) and [getJsFile](https://github.com/flarum/core/blob/master/src/Asset/AssetManager.php#L103)
- ability to change/alter `AssetManager` compilers (`JsCompiler` and `LessCompiler`), this option could help for CDN

this changes can help for CDN, Theming, ... (in my situation add RTL support)

P.S. all this changes could be done by using [monkey patching](https://github.com/flarum/core/issues/246#issuecomment-181993118) @franzliedke

### LINKED ISSUE #881: Improve (or avoid) the nojs experience

### LINKED ISSUE #951: We need removeHeadString and removeFootString in ClientView
I was trying to create an extension for selecting a custom font for the entire forum from the administrator and replacing the default one, which is hard-coded (???) in the `ClientView`'s constructor.
But I realized that there's no function for replacing or removing a string from the `headStrings` array.

Is there a workaround for this or should I add these functions to the class (`removeHeadString`, `removeFootString`, `replaceHeadString` and `replaceFootString`?

### LINKED ISSUE #1024: Uglify JS in flarum-gulp, commit minified files and source maps to the repo

tobscure:
Fixes #881 
Fixes #1024 
Fixes #837 
Fixes #865 
Fixes #951

The main changes are:

- Use Laravel's Filesystem component for asset IO, meaning theoretically
  assets should be storable on S3 etc.

- More reliable checking for asset recompilation when debug mode is on,
  so you don't have to constantly delete the compiled assets to force
  a recompile. Should also fix issues with locale JS files being
  recompiled with the same name and cached.

- Remove JavaScript minification, because it will be done by Webpack
  (exception is for the TextFormatter JS).

- Add support for sourcemaps (both JS and LESS)!

- Refactor FrontendView class:
  - Remove all getters/setters - just use public properties instead.
  - Refine the public API.
  - Pass the Request as a dependency on construction so that "default
    content" can be added to the instance. Then extensions can alter it
    as they see fit.

The bits I would like feedback on are:

* The FrontendView class. Is there any way we can simplify it / reduce its number of dependencies? I've been scratching my head over this all day and I can't think of anything. Maybe it is ok as it is 🤔 

* The other Frontend classes: FrontendAssets, AbstractFrontend + Forum/AdminFrontend, AbstractFrontendController + Forum/AdminFrontendController. Does the way I've structured this make sense, is it intuitive enough? Any ideas to improve?


tobscure:
@franzliedke It's difficult because a lot of the changes are intertwined :/ In any case, the diff is not particularly useful as I'm not really interested in feedback on what I've *changed* (because I already know I've made it better than it was before) - but rather how this section of the code is looking as a whole:

https://github.com/flarum/core/tree/frontend-refactor/src/Frontend
https://github.com/flarum/core/tree/frontend-refactor/src/Forum

tobscure:
For my own benefit, here's how I would summarise these classes and how they work together:

* `FrontendView` is a `Renderable` that will show the HTML needed to boot a given frontend. It does this by:
  * Exposing various public properties for extensions to modify
  * Initializing asset compilers (via `FrontendAssets`) which are also exposed so extensions can add assets
  * Building a payload to be passed into the JS app, including API responses from the forum and current user, session information, locale information

* `FrontendAssets` is responsible for building asset compilers for a given frontend.
  * It is used by the `FrontendView`
  * It is also used by `Forum`/`AdminServiceProvider` to flush all assets when extensions are enabled/disabled

* `FrontendAssetsFactory` and `FrontendViewFactory` are present to assist in the creation of the above classes, and group their dependencies together in `AbstractFrontend`

* `AbstractFrontend` is a factory for building a frontend's view and assets
  * It is extended as `Forum`/`AdminFrontend` to add frontend-specific information (i.e. the name of the frontend [and thus where to locate its default assets/translations], add TextFormatter JS for the forum, add settings to the payload for the admin)
  * These implementations are used by frontend controllers to build the frontend view, and by `Forum`/`AdminServiceProvider` to build the assets for flushing

* `AbstractFrontendController` uses an `AbstractFrontend` to build a frontend view and render that.
  * It is extended as `Forum`/`Admin\FrontendController` to depend upon a specific `AbstractFrontend` implementation
  * The `getView` method can then be extended by child controllers to alter the view's properties (e.g. SEO content, title, description)

---

I guess the above architecture works but it's a bit unclear in my mind, there is a lot of "factory" behaviour for one...

I suppose one thing minor we could look at doing is extracting the `FrontendView`'s payload-building code into a `FrontendPayloadBuilder`?

tobscure:
e.g. something like this:

```php

return [
    (new Extend\Routes('forum'))
        ->get(
            '/embed/{id:\d+(?:-[^/]*)?}[/{near:[^/]*}]',
            'flarum.embed.discussion',
            function (Extend\Routes\RouteBuilder $route) {
                $route->discussion()->useFrontend('embedded-discussion');
            }
        ),

    (new Extend\Frontend('embedded-discussion'))
        ->layout(__DIR__.'/views/layout.blade.php')
        ->js(__DIR__.'/js/forum.js')
        ->css(__DIR__.'/less/forum.less')
];
```

franzliedke:
I wouldn't focus too much on the constructor signature at this point, especially since most objects are instantiated by the IoC container. Even regarding the public API surface, I can only suggest to see what is needed - which we can only figure out during the first few releases. With relatively small, fine-grained objects, further development / extension points can more likely be added on top instead of having to modify existing classes. Or so I hope.

---

Concerning the concrete issues:
An important aspect here seems to be the question **inheritance** or **composition**. OO afficionados usually prefer the latter, because it allows more flexible combination/reuse of classes, whereas with inheritance, you are tied to the things that a concrete parent class gives you. So, I would love to see us move in a direction that allows for more composition.

I haven't thought this through to the very end due to a lack of time for experimentation, so I am sure there are loose ends. Anyway, here it goes...

You were previously suggesting to overwrite the abstract controller, but only changing the controller signature in order to inject the correct dependencies for each controller. Having only that in a subclass smells, and it made me think we are not really dealing with different behavior here - we want to inject different subtypes. This means, **maybe those different controllers aren't different classes, they're different *instances***.

So, what happens if we opt for a generic, non-abstract controller class, that depends on several interfaces, replacing your previous `AbstractFrontendController`:

~~~php
class FrontendController implements Controller
{
  public function __construct(Frontend $frontend, Content $content, Dispatcher $events)
  {
    $this->frontend = $frontend;
    $this->content = $content;
    $this->events = $events;
  }

  public function render(Request $request)
  {
    $view = $this->frontend->makeView($request);

    $this->content->renderInto($view, $request);

    $this->events->dispatch(new Rendering($this, $view, $request));

    return $view;
  }
}
~~~

This is everything but abstract, it is a complete implementation of an interface. Variation in behavior can happen by injecting various implementations of the interfaces we depend on.

Then, *in a service provider*, we would instruct the IoC container how to construct *concrete instances* of controllers.

~~~php
$this->app->bind(
  'controller.discussion',
  function ($app) {
    return new FrontendController(
      new ForumFrontend, new DiscussionContent, $app->make(Dispatcher::class)
    );
  }
);
~~~

Of course, the differing behavior still needs some subclasses (or better, multiple implementations of interfaces such as `Frontend`). This is where `ForumFrontend` or `DiscussionContent` come into play. But with your generic implementation, the behavior of the controller does not need to vary, right?

That leaves us with extensibility: Besides dispatching events in the controller, frontend, view etc., we can now use Laravel's container to `extend` various classes. (This might require using `$app->make()` for instantiating more of the dependencies in my service provider snippet, but that's not a problem.) The container has two very useful, but rarely-used methods: `resolving` and `extend`. The former lets you specify a callback for when an object is pulled out of a container:

~~~php
$app->resolving(
  ForumFrontend::class,
  function ($frontend) {
    // Set some variables on the frontend or whatever...
  }
);
~~~

This stuff could be wrapped in an extender, of course.
`Container::extend` can be used to wrap objects in decorators, before they are resolved:

~~~php
$app->extend(
  'your.favorite.controller',
  function ($controller) {
    return new MagicFancyControllerDecorator($controller);
  }
)
~~~

(Decorators are mostly useful in combination with narrow interfaces.)

Okay, that's a bit of a braindump. I hope it's useful. Again, I might have missed some important parts, but it should certainly do as a conversation starter. :beers:

franzliedke:
Off-topic: Are you intentionally merging into `webpack`? This will get us a very huge diff. :scream:

tobscure:
> Off-topic: Are you intentionally merging into webpack? This will get us a very huge diff. 😱

You're right! I will avoid that, I'll merge the webpack branch into master before I merge any of this. For now it's just for the sake of WIP.

Thanks for your post @franzliedke, very helpful lesson. But I'm still having a lot of trouble actually applying the concept to this situation and being happy with how our public API looks. I've been at this for days and I feel like my brain is dead, so I thought I would do a fun exercise:

I converted the whole thing into procedural code. 😬

Below is everything that needs to happen in order to render the admin view. The challenge is to refactor it into small parts with manageable dependencies. I have marked the admin-specific behaviour with comments.

Importantly, another instance (for the sake of example, an "embedded admin" route), should have the ability to "inherit" all of this behaviour (including whatever logic is applied to the "admin" instance in the `Rendering` event) while changing certain properties (`$assetsName`) and add its own behaviour (i.e. customise `$attributes`).

I've done this because I have a feeling I've just been going about architecting this the wrong way. So going back to the drawing board.

Anyone wanna take a crack?

```php
class AdminFrontendController 
{
    public function __construct(
        FilesystemAdapter $assetsDir,
        Application $app,
        ApiClient $api,
        ViewFactory $view,
        LocaleManager $locales,
        SettingsRepositoryInterface $settings
        PermissionsRepository $permissions,
        ExtensionManager $extensions,
        UserSerializer $userSerializer
    ) {
    }

    public function handle(Request $request)
    {
        $locale = $request->getAttribute('locale');

        // <<< ADMIN
        $assetsName = 'admin';
        // >>>

        $js = new JsCompiler($this->assetsDir, $assetsName.'.js', $this->app->inDebugMode());
        $css = new LessCompiler($this->assetsDir, $assetsName.'.css', $this->app->inDebugMode());
        $localeJs = new LocaleJsCompiler($this->assetsDir, $assetsName.'-'.$locale.'.js', $this->app->inDebugMode());
        $localeCss = new LessCompiler($this->assetsDir, $assetsName.'-'.$locale.'.css', $this->app->inDebugMode());     

        // <<< ADMIN
        $js->addFile('js/admin.js');
        $css->addFile('less/admin.js');
        // >>>

        foreach ($this->locales->getJsFiles($locale) as $file) {
            $view->localeJs->addFile($file);
        }

        foreach ($this->locales->getCssFiles($locale) as $file) {
            $view->localeCss->addFile($file);
        }

        $lessVariables = [
            'config-primary-color'   => $this->settings->get('theme_primary_color') ?: '#000',
            'config-secondary-color' => $this->settings->get('theme_secondary_color') ?: '#000',
            'config-dark-mode'       => $this->settings->get('theme_dark_mode') ? 'true' : 'false',
            'config-colored-header'  => $this->settings->get('theme_colored_header') ? 'true' : 'false'
        ];

        $lessVariablesString = function () use ($lessVariables) {
            foreach ($lessVariables as $name => $value) {
                $less[] = "@$name: $value;";
            }

            return implode('', $less);
        };

        $css->addString($lessVariablesString);
        $localeCss->addString($lessVariablesString);

        // <<< ADMIN
        $filter = '/^.+(?:\.|::)(?:admin|lib)\./';
        // >>>

        $translations = array_get($this->locales->getTranslator()->getCatalogue()->all(), 'messages', []);
        $filtered = array_filter(array_keys($translations), function ($id) use ($filter) {
            return preg_match($filter, $id);
        });
        $localeJs->setTranslations(array_only($translations, $filtered));

        $response = $this->api->send(ShowForumController::class, $actor);
        $forum = json_decode($response->getBody(), true);

        $attributes = new StdObject;
        $attributes->title = '';
        $attributes->language = $locale;
        $attributes->direction = 'ltr';
        // <<< ADMIN
        $attributes->layout = 'flarum::frontend.admin';
        // >>>
        $attributes->content = null;
        $attributes->document = null;
        $attributes->variables = [];
        $attributes->meta = [
            'viewport' => 'width=device-width, initial-scale=1, maximum-scale=1, minimum-scale=1',
            'description' => array_get($forum, 'data.attributes.forumDescription'),
            'theme-color' => array_get($forum, 'data.attributes.themePrimaryColor')
        ];
        $attributes->canonicalUrl = null;
        $attributes->head = [
            'font' => '<link rel="stylesheet" href="//fonts.googleapis.com/css?family=Open+Sans:400italic,700italic,400,700,600">'
        ];
        $attributes->foot = [];
        $attributes->js = $js;
        $attributes->css = $css;
        $attributes->localeJs = $localeJs;
        $attributes->localeCss = $localeCss;

        if (array_get($request->getQueryParams(), 'nojs')) {
            $attributes->meta['robots'] = 'noindex';
        }

        // <<< ADMIN
        $attributes->variables['settings'] = $this->settings->all();
        $attributes->variables['permissions'] = $this->permissions->map();
        $attributes->variables['extensions'] = $this->extensions->getExtensions()->toArray();
        // >>>

        // <<< ADMIN
        $name = 'admin';
        // >>>

        event(new Rendering($name, $attributes, $request));

        $this->view->share('allowJs', ! array_get($request->getQueryParams(), 'nojs'));
        $this->view->share('forum', array_get($forum, 'data'));

        $title = ($attributes->title ? $attributes->title.' - ' : '').array_get($forum, 'data.attributes.title');

        $data = array_merge([$forum['data']], $forum['included']);

        if ($actor->exists) {
            $this->userSerializer->setActor($actor);
            $resource = new Resource($actor, $this->userSerializer);
            $document = new Document($resource->with('groups'));
            $data[] = $document->toArray();
        }

        $payload = [
            'resources' => $data,
            'session' => [
                'userId' => $actor->id,
                'csrfToken' => $request->getAttribute('session')->token()
            ],
            'document' => $attributes->document,
            'locales' => $this->locales->getLocales(),
            'locale' => $locale
        ];

        $payload = array_merge($payload, $attributes->variables);

        $layout = $this->view->make($attributes->layout)
            ->with('content', $this->view->make('flarum::frontend.content')
                ->with('content', $attributes->content));

        $cssUrls = array_filter([$css->getUrl(), $localeCss->getUrl()]);

        $head = array_map(function ($url) {
            return '<link rel="stylesheet" href="'.e($url).'">';
        }, $cssUrls);

        if ($faviconUrl = array_get($forum, 'data.attributes.faviconUrl')) {
            $head[] = '<link rel="shortcut icon" href="'.e($faviconUrl).'">';
        }

        if ($attributes->canonicalUrl) {
            $head[] = '<link rel="canonical" href="'.e($attributes->canonicalUrl).'">';
        }

        $head = array_merge($head, array_map(function ($content, $name) {
            return '<meta name="'.e($name).'" content="'.e($content).'">';
        }, $attributes->meta, array_keys($attributes->meta)));

        $head = implode("\n", array_merge($head, $attributes->head));

        $jsUrls = array_filter([$js->getUrl(), $localeJs->getUrl()]);

        $js = implode("\n", array_map(function ($url) {
            return '<script src="'.e($url).'"></script>';
        }, $jsUrls));

        $foot = implode("\n", $attributes->foot);

        $appView = $this->view->make('flarum::frontend.app')->with([
            'title' => $title,
            'payload' => $payload,
            'layout' => $layout,
            'language' => $attributes->language,
            'direction' => $attributes->direction,
            'js' => $js,
            'head' => $head,
            'foot' => $foot
        ]);

        return new HtmlResponse($appView->render());
    }
}
```

tobscure:
Woo! Just pushed the new frontend architecture I've been working on for the past... oh gosh, it's been 5 days. 😐

Once again, the diff and scale of the changes is huge, so I would suggest reading the commit messages and then looking at the relevant parts of the codebase in full:

https://github.com/flarum/core/tree/frontend-refactor/src/Frontend
https://github.com/flarum/core/tree/frontend-refactor/src/Forum

Probably my no. 1 takeaway from this is that, as @franzliedke has rightly pointed out, Laravel's container is very powerful (especially `resolving` and `extend`), and we should be making more use of it for extensibility! I have a few ideas about where we can use it to remove some of those lingering `Flarum\Event` classes.

With this new architecture, the complete embed extension will be able to look something like this – no PHP classes necessary:

```php
use Flarum\Forum\Content;

return [
    (new Extend\Routes('forum'))
        ->get(
            '/embed/{id:\d+(?:-[^/]*)?}[/{near:[^/]*}]',
            'flarum.embed',
            Extend\Routes::toFrontend('flarum.embed.frontend', Content\Discussion::class)
        ),

    (new Extend\Frontend('flarum.embed.frontend'))
        ->extends('forum')
        ->assets('flarum.embed.assets')
        ->layout('flarum.embed::layout'),

    (new Extend\Assets('flarum.embed.assets'))
        ->extends('forum')
        ->js(__DIR__.'/js/embed.js')
        ->css(__DIR__.'/less/embed.less')
];
```

src/Forum/RecompileFrontendAssets.php
```diff
@@ -0,0 +1,84 @@
...
+            // is actually written yet.
+
+            $settings = $this->container->make(SettingsRepositoryInterface::class);
+
+            $this->container->extend(
+                SettingsRepositoryInterface::class,
+                function ($settings) use ($event) {
+                    return new OverrideSettingsRepository($settings, $event->settings);
+                }
+            );
```
franzliedke: Where is this decorated instance being resolved? Interesting use case for sure. :wink:

src/Frontend/RecompileFrontendAssets.php
```diff
@@ -0,0 +1,98 @@
...
+    /**
+     * @param FrontendCompilerFactory $assets
+     * @param LocaleManager $locales
+     * @param Container $container
+     */
+    public function __construct(FrontendCompilerFactory $assets, LocaleManager $locales, Container $container)
+    {
+        $this->assets = $assets;
+        $this->locales = $locales;
+        $this->container = $container;
```
franzliedke: This container is not being used in this class. If it's just used in the subclasses, just inject it there, please.

src/Settings/OverrideSettingsRepository.php
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * (c) Toby Zerner <toby.zerner@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Flarum\Settings;
+
+class OverrideSettingsRepository implements SettingsRepositoryInterface
```
franzliedke: This class might benefit from a quick docblock explaining its purpose. :smile:

src/Frontend/FrontendServiceProvider.php
```diff
@@ -20,23 +21,55 @@
...
-     * {@inheritdoc}
-     */
     public function register()
     {
-        $this->registerAssetsFilesystem();
-    }
+        // Yo dawg, I heard you like factories, so I made you a factory to
+        // create your factory. We expose a couple of factory functions that
+        // will create frontend factories and configure them with some default
+        // settings common to both the forum and admin frontends.
```
franzliedke: Very nice. :sweat_smile:

franzliedke:
Looking nice, great job! :clap:

franzliedke:
Quick side note: The classes in `src/Frontend` all have `Frontend` in their names as well... is that intentional? They are in that namespace, after all. :wink:

src/Forum/RecompileFrontendAssets.php
tobscure: `Flarum\Forum\Asset\CustomCss`, which is instantiated in a closure within `flarum.forum.assets` defined in `ForumServiceProvider`.

It seems like a *mildly* hacky solution... now that I think of it, perhaps a contextual binding would be better?

tobscure:
@franzliedke We don't really have a consistent convention for this? Like take the `Discussion` namespace for example, you've got `DiscussionRepository` and `DiscussionPolicy` as opposed to just `Repository` and `Policy`. Maybe that's different though because `Repository` and `Policy` are so common. Hmm 🤔 what about this:

```
CompilerFactory
FrontendController
FrontendServiceProvider
View
ViewFactory
```

src/Forum/RecompileFrontendAssets.php
tobscure: Decided against the contextual binding; it would imply specific knowledge of the asset compilation system internals, whereas a normal binding is more of a blanket test.

tobscure:
This will fix https://github.com/flarum/core/issues/881 and https://github.com/flarum/core/issues/1024 and https://github.com/flarum/core/issues/837

src/Extend/Assets.php
```diff
@@ -12,29 +12,37 @@
...
+    public function css($path)
     {
-        $this->assets[] = $path;
+        $this->css[] = $path;
 
         return $this;
     }
 
+    /**
+     * @deprecated
```
luceos: I'd prefer seeing an indication of why it was deprecated or what to do instead. Referencing an issue or another class method should mostly suffice.

tobscure:
Re: `AssetInterface` - this is one of the things I was less happy with but I couldn't think of a better way to do it. The reason there need to be four entry points (rather than one `compile` method) is because the compilers are built independently of each other – see in `FrontendCompilerFactory`. This is so that when clearing the cache we can rebuild them selectively (i.e. build just the CSS ones, or build locale ones for every locale). I played around with having four separate interfaces (i.e. a `JsAssetInterface`, a `CssAssetInterface`, a `LocaleJsAssetInterface`, and a `LocaleCssAssetInterface`) which the asset classes could then implement selectively, but that required code to apply them selectively too in the CompilerFactory which felt just as clunky. I'm not sure, any thoughts would be appreciated.

src/Extend/Assets.php
franzliedke: Let's just remove this instead? The extenders have not even been released anywhere - we "only" need to update dev-master of our extensions...

franzliedke:
IMO, this closes #865 as well.

tobscure:
@franzliedke do you have any thoughts on (1) [naming](https://github.com/flarum/core/pull/1450#issuecomment-396095394) and (2) [AssetInterface](https://github.com/flarum/core/pull/1450#issuecomment-398019666)?

franzliedke:
### Naming

Your reasoning is legit. Nothing to add, except for one thing: Can we find a better name for `FrontendView`? Maybe we could go with `Document`? It seems to be representing a full HTML document... Only problem: That would conflict with the preloadable JSON-API document.

### AssetInterface

To be honest, I am having difficulties following along and understanding the exact purpose, but that may just be a problem of the current time. :wink: 
Do I understand its purpose correctly when I say that the `AssetInterface` represents a bunch of assets that belong together (e.g. because they are from the same source, such as an extension)? If so, I think we have an interesting abstraction here, but lack a good name.

Not sure if that helps, but maybe thinking once more in the direction of multiple instances (being registered in the container) can once more put us on a different track, maybe getting us the simple interface @luceos proposed?

In any case, I am happy to have this merged now (still needs a rebase and re-target on master). Maybe we could schedule a pair programming session to talk through `AssetInterface` and play around with some alternatives.

tobscure:
> Do I understand its purpose correctly when I say that the AssetInterface represents a bunch of assets that belong together (e.g. because they are from the same source, such as an extension)?

That's correct.

> Maybe we could schedule a pair programming session to talk through AssetInterface and play around with some alternatives.

Sounds good! Will discuss times with you on Discord.

tobscure:
Rebased on master in #1471

PR #1453, opened 2018-06-07
fix empty base url

sijad:
not sure why but in my local install `originalUri` attribute is empty, therefore `baseUrl` is empty, `url` attribute in `config.php` was empty and `mail_from` has been set `noreply@` in `settings` table.

not sure why this happening, maybe it's because of my nginx config?

### DIFF
--- a/src/Install/Controller/InstallController.php
+++ b/src/Install/Controller/InstallController.php
@@ -77,7 +77,7 @@ class InstallController implements RequestHandlerInterface
             'email'                 => array_get($input, 'adminEmail'),
         ]);
 
-        $baseUrl = rtrim((string) $request->getAttribute('originalUri'), '/');
+        $baseUrl = rtrim((string) $request->getUri(), '/');
         $data->setBaseUrl($baseUrl);
 
         $data->setSetting('forum_title', array_get($input, 'forumTitle'));


src/Install/Controller/InstallController.php
```diff
@@ -77,7 +77,7 @@ public function handle(Request $request): ResponseInterface
             'email'                 => array_get($input, 'adminEmail'),
         ]);
 
-        $baseUrl = rtrim((string) $request->getAttribute('originalUri'), '/');
+        $baseUrl = rtrim((string) $request->getAttribute('originalUri', $request->getUri()), '/');
```
tobscure: Can we simply use `$request->getUri()`? Not sure why we are getting the "original" URI, whatever that is.

sijad: I guess we can, I was not sure what is `originalUri` either.

tobscure: @sijad would you be able to please try it out?

sijad: I've tested `$baseUrl = rtrim((string) $request->getUri(), '/');` with clean install, everything seems fine, should I change this PR to remove `originalUri`?

franzliedke: Yep. In fact, `originalUri` is the real problem here. It [was available in zend-stratigility v1](https://docs.zendframework.com/zend-stratigility/v2/migration/#original-request-response-and-uri), but apparently we forgot to adopt the v2 way when upgrading. Simply `$request->getUri()` will do. :+1:

franzliedke: Correction: The `rtrim` thing should stay in place, so your last suggestion is just right.

sijad:
please review again

PR #1456, opened 2018-06-12
check class string before concat

### LINKED ISSUE #1454: [b8] CommentPost article has undefined class
image

https://discuss.flarum.org/d/12942-shared-hosting/11

sijad:
fixes #1454

I couldn't find anything else. there might be more of this.
there is few `+=` in `Dropdown` components, but they seem alright.

### DIFF
--- a/js/src/forum/components/CommentPost.js
+++ b/js/src/forum/components/CommentPost.js
@@ -80,7 +80,7 @@ export default class CommentPost extends Post {
     const post = this.props.post;
     const attrs = super.attrs();
 
-    attrs.className += ' '+classList({
+    attrs.className = (attrs.className || '') + ' ' + classList({
       'CommentPost': true,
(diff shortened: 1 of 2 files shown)


franzliedke:
@sijad Please rebase this.

What about adding the capability to start from an existing class string to the `classList()` helper?

tobscure:
@franzliedke Ultimately we want to replace our `classList` util with the [classnames package
](https://www.npmjs.com/package/classnames). Then we can do something like:

```js
attrs.className = classNames(attrs.className, { ... });
```

franzliedke:
Ah yes, I had seen that before; that's probably what prompted me to propose something similar.

franzliedke:
Here or in a separate PR? (We should probably get this bugfix in and then refactor towards that separately.)

tobscure:
Separate PR because the changes will be more far-reaching.

PR #1457, opened 2018-06-14
Upgrade text formatter

### LINKED ISSUE #1377: Update to newer TextFormatter
In the course of my Flarum migration service I noticed many posts coming form other platforms can't be parsed by Flarum. And by can't be parsed I mean it throws errors when passing the content to TextFormatter.

There are both critical PHP and javascript parsing issues. I tried with a newer version of TextFormatter and these seem fixed.

I expected master to run on a newer TextFormatter but it seems like it isn't. Is there any plan to update ? I can contribute to that if no plan was made.

Here's a simple example, create a new post with content `    ` (4 spaces). You won't be able to edit that post anymore :stuck_out_tongue:

sijad:
fixes #1377

### DIFF
--- a/src/Formatter/Formatter.php
+++ b/src/Formatter/Formatter.php
@@ -112,6 +112,9 @@ class Formatter
         $configurator->rendering->engine = 'PHP';
         $configurator->rendering->engine->cacheDir = $this->cacheDir;
 
+        $configurator->enableJavaScript();
+        $configurator->javascript->exportMethods = ['preview'];
+
         $configurator->Escaper;
         $configurator->Autoemail;
         $configurator->Autolink;
@@ -193,11 +196,6 @@ class Formatter
      */
     public function getJs()
     {
-        $configurator = $this->getConfigurator();
-        $configurator->enableJavaScript();
-        $configurator->javascript->exportMethods = ['preview'];
-        $configurator->javascript->setMinifier('MatthiasMullieMinify');
-
-        return $configurator->finalize()['js'];
+        return $this->getComponent('js');
     }
 }
--- a/src/Formatter/Formatter.php
+++ b/src/Formatter/Formatter.php
@@ -115,7 +115,6 @@ class Formatter
         $configurator->javascript->exportMethods = ['preview'];
 
         $minifier = $configurator->javascript->setMinifier('FirstAvailable');
-        $minifier->add('ClosureCompilerService');
         $minifier->add('MatthiasMullieMinify');
         $minifier->add('Noop');
 
--- a/src/Formatter/Formatter.php
+++ b/src/Formatter/Formatter.php
@@ -198,9 +198,6 @@ class Formatter
         $configurator->javascript->exportMethods = ['preview'];
         $configurator->javascript->setMinifier('MatthiasMullieMinify');
 
-        return $configurator->finalize([
-            'returnParser' => false,
-            'returnRenderer' => false
-        ])['js'];
+        return $configurator->finalize()['js'];
     }
 }
--- a/src/Formatter/Formatter.php
+++ b/src/Formatter/Formatter.php
@@ -96,8 +96,7 @@ class Formatter
      */
     public function flush()
     {
-        $this->cache->forget('flarum.formatter.parser');
-        $this->cache->forget('flarum.formatter.renderer');
+        $this->cache->forget('flarum.formatter');
     }
 
     /**
@@ -145,16 +144,16 @@ class Formatter
     /**
      * Get a TextFormatter component.
      *
-     * @param string $name "renderer" or "parser"
+     * @param string $name "renderer" or "parser" or "js"
      * @return mixed
      */
     protected function getComponent($name)
     {
-        $cacheKey = 'flarum.formatter.'.$name;
-
-        return $this->cache->rememberForever($cacheKey, function () use ($name) {
-            return $this->getConfigurator()->finalize()[$name];
+        $formatter = $this->cache->rememberForever('flarum.formatter', function () {
+            return $this->getConfigurator()->finalize();
         });
+
+        return $formatter[$name];
     }
 
     /**
--- a/src/Formatter/Formatter.php
+++ b/src/Formatter/Formatter.php
@@ -114,6 +114,11 @@ class Formatter
         $configurator->enableJavaScript();
         $configurator->javascript->exportMethods = ['preview'];
 
+        $minifier = $configurator->javascript->setMinifier('FirstAvailable');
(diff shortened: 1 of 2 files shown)


src/Formatter/Formatter.php
```diff
@@ -197,9 +197,6 @@ public function getJs()
         $configurator->enableJavaScript();
         $configurator->javascript->exportMethods = ['preview'];
 
-        return $configurator->finalize([
-            'returnParser' => false,
-            'returnRenderer' => false
-        ])['js'];
+        return $configurator->finalize()['js'];
```
sijad: I think this can be optimized.
`enableJavaScript()` can moved to `getConfigurator()` and `getComponent()` can be used here, e.g:

```
return $this->getComponent('js');
```

but as `getComponent()` results will cache I'm not sure if it's good idea or not

tobscure: I sounds like a reasonable thing to do.

sijad: should `getConfigurator()` result get cached in memory, e.g:

```php
protected function getConfigurator()
{
    if (! $this->configurator) {
        $configurator = new Configurator;
        ...
        ...
        ...
        $this->configurator = $configurator;
    }

    return $this->configurator;
}
```

tobscure: We should probably just be caching `$this->getConfigurator()->finalize()` [here](https://github.com/sijad/core/blob/90dd1bb8b1f9d7a539941b9045cb5151e0dab6e7/src/Formatter/Formatter.php#L153) and refactoring to get the parser/renderer/JS from that.

sijad: like this?

```php
return $this->cache->rememberForever('flarum.formatter', function () {
    return $this->getConfigurator()->finalize();
});
```

tobscure: Exactly.

tobscure:
This is good to merge, #1460 can be done independently

sijad:
I haven't checked `flarum-ext-mentions` yet.

franzliedke:
@sijad Please remove the "WIP" if this is ready from your side.

src/Frontend/Asset/JsCompiler.php
```diff
@@ -89,30 +81,13 @@ class_exists(Configurator::class);
...
-        $hostedMinifer = $minifier->add('HostedMinifier');
-        $hostedMinifer->url = 'http://s9e-textformatter.rhcloud.com/flarum-minifier/';
-        $hostedMinifer->httpClient->timeout = 30;
-
         $ccs = $minifier->add('ClosureCompilerService');
         $ccs->compilationLevel = 'SIMPLE_OPTIMIZATIONS';
         $ccs->httpClient->timeout = 30;
 
         $minifier->add('MatthiasMullieMinify');
+        $minifier->add('Noop');
```
franzliedke: We are still using `MatthiasMullieMinify` here, but the dependency was removed from composer.json. In text-formatter itself, it is only a dev dependency. Mistake or what am I missing?

sijad: Oh god, thanks for noticing it.

sijad:
I've not checked `flarum-ext-mentions` to see if there's any conflict, but as `flarum-ext-emoji` will break too, I guess it's good too go.

src/Frontend/Asset/JsCompiler.php
franzliedke: I am confused. It's gone now, because of Toby's Webpack changes, right?

tobscure: It's gone from the JsCompiler, but we need to add minification into the Formatter class because nothing else is gonna minify the formatter JS.

sijad: Doesn't it re-minify everything?

tobscure: No, JsCompiler no longer minifies anything, we've moved that responsibility to webpack so the `js/dist/*` files come pre-minified.

sijad: I mean doesn't it re-minifiy webpack-minified version once again?
https://github.com/flarum/core/blob/3f683dd6ee739f850765e9827f8b410f9c9187f9/src/Frontend/Asset/RevisionCompiler.php#L152

sijad: should we only minify [added strings](https://github.com/flarum/core/blob/3f683dd6ee739f850765e9827f8b410f9c9187f9/src/Frontend/Asset/RevisionCompiler.php#L64)?

tobscure: No, it doesn't: https://github.com/flarum/core/blob/3f683dd6ee739f850765e9827f8b410f9c9187f9/src/Frontend/Asset/JsCompiler.php#L39-L42

It's no longer the responsibility of the compiler to minify, it's the responsibility of the caller. Please add minification to the Formatter JS output as per http://s9etextformatter.readthedocs.io/JavaScript/Minifiers/#matthiasmullie92minify

tobscure:
@sijad awesome! We need PRs/assurance that mentions and emoji will work before we can merge this into master.

tobscure:
I think you're right, it would be better to use just the MatthiasMullieMinify one to make it always happen locally.

franzliedke:
@sijad Please take care of Toby's latest suggestion, and then rebase on latest master. :)

src/Formatter/Formatter.php
```diff
@@ -112,6 +111,13 @@ protected function getConfigurator()
         $configurator->rendering->engine = 'PHP';
         $configurator->rendering->engine->cacheDir = $this->cacheDir;
 
+        $configurator->enableJavaScript();
+        $configurator->javascript->exportMethods = ['preview'];
+
+        $minifier = $configurator->javascript->setMinifier('FirstAvailable');
+        $minifier->add('MatthiasMullieMinify');
+        $minifier->add('Noop');
```
franzliedke: @JoshyPHP Do we need the `FirstAvailable` strategy if we only use `MatthiasMullieMinify`? In other words, can that one fail in a way that is caught by `FirstAvailable`?

JoshyPHP: I wouldn't know off the top of my head but after reading the API doc I think you can either use `FirstAvailable` in conjunction with `Noop` or you can a single minifier with `$keepGoing = true`, which also mutes exceptions.

```php
$configurator->javascript->setMinifier('MatthiasMullieMinify')->keepGoing = true;
```

franzliedke: @sijad Please do that then. :see_no_evil:

PR #1462, opened 2018-06-16
Use container events as an alternative to the ConfigureMiddleware event

tobscure:
By moving the DispatchRoute middleware into an `afterResolving`
callback, this will allow a new Middleware extender to add a `resolving`
callback to the appropriate container binding, removing the need for the
ConfigureMiddleware event.

The ConfigureMiddleware event has been deprecated and should be removed
in beta 9.

### DIFF
--- a/src/Admin/AdminServiceProvider.php
+++ b/src/Admin/AdminServiceProvider.php
@@ -11,7 +11,6 @@
 
 namespace Flarum\Admin;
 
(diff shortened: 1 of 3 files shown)


franzliedke:
Please rebase.

tobscure:
> So - if we already had the middleware extender - this would simply change the implementation of that extender, right?

Yep!

tobscure:
Rebased.

PR #1464, opened 2018-06-17
Recursive Extenders: Allow Extenders to return more Extenders

tobscure:
This will allow us to make composite Extenders without duplicating logic.

Also allows a single extender to be returned from bootstrap.php rather than an array (see flarum/flarum-ext-english#128)

### DIFF
--- a/src/Extension/Extension.php
+++ b/src/Extension/Extension.php
@@ -119,7 +119,16 @@ class Extension implements Arrayable
             return;
         }
 
-        $extenders = array_flatten((array) require $bootstrapper);
+        $this->runExtenders($app, require $bootstrapper);
+    }
+
+    private function runExtenders(Container $app, $extenders)
+    {
+        if (! is_array($extenders)) {
+            $extenders = [$extenders];
+        }
+
+        $extenders = array_flatten($extenders);
 
         foreach ($extenders as $extender) {
             // If an extension has not yet switched to the new bootstrap.php
@@ -129,7 +138,11 @@ class Extension implements Arrayable
                 $extender = new Compat($extender);
             }
 
-            $extender($app, $this);
+            // Extenders may return additional Extenders when they are invoked.
+            // Recursively run them if there were any returned.
+            if ($return = $extender($app, $this)) {
+                $this->runExtenders($app, $return);
+            }
         }
     }
 
--- a/src/Extend/ExtenderInterface.php
+++ b/src/Extend/ExtenderInterface.php
@@ -16,5 +16,10 @@ use Illuminate\Contracts\Container\Container;
 
 interface ExtenderInterface
 {
+    /**
+     * @param Container $container
+     * @param Extension|null $extension
+     * @return void|ExtenderInterface|ExtenderInterface[]
+     */
     public function __invoke(Container $container, Extension $extension = null);
 }


src/Extension/Extension.php
```diff
@@ -119,7 +119,16 @@ public function extend(Container $app)
             return;
         }
 
-        $extenders = array_flatten((array) require $bootstrapper);
+        $this->runExtenders($app, require $bootstrapper);
```
luceos: I'm confused. The previous code already cast the required file into an array, so both a single item and array would become an array. There should be no need to create this unnecessary logic with the `is_array` further down the class, or am I missing something here?

Have you tried using a single Extender with the old code?

tobscure: Yes, `(array)` casts an object into an array rather than wrapping an array around the object, so the `is_array` check is necessary.

franzliedke:
I fail to see the use-case, to be honest. Can you share (or link to) an example?

tobscure:
**Use-case 1: Consolidate certain tasks/dependencies in a single extender**
eg. Instead of grabbing a dispatcher from the container each time we want to register listeners, just delegate to Extend\Listener.
https://github.com/flarum/core/blob/a8a4f5cba81f98f5e7bf8ccba92fb059477a0ea2/src/Extend/Formatter.php#L33-L54
Becomes:
```php
public function __invoke(Container $container, Extension $extension = null)
{
    return [
        new Listener(
            Configuring::class,
            function (Configuring $event) {
                call_user_func($this->callback, $event->configurator);
            }
        ),

        new Listener(
            [Enabled::class, Disabled::class],
            function ($event) use ($container, $extension) {
                if ($event->extension === $extension) {
                    $container->make(ActualFormatter::class)->flush();
                }
            }
        )
    ];
}
```

This is just one example, I can imagine having this flexibility could open up other possibilities for code reuse as well.

**Use-case 2: Allow extensions to define extenders dynamically**

```php
// flarum-ext-debug/extend.php
return [
    function (Application $app) {
        if ($app->config('debug')) {
            return [
                (new Extend\Assets('forum'))->css('debug.css');
                // not a good example, as assets shouldn't be declared dynamically
                // but again this flexibility could be useful
            ];
        }
    }
];
```

franzliedke:
I'd prefer to avoid the `Listener` extender, by the way - hoping to keep events as implementation details in most cases.

So... can we wait with introducing this until we actually, really, need it? Not sure I like the additional interface complexity, to be honest.

franzliedke:
Regarding **Use-case 2**: Would that work without returning a closure (which is a deprecated feature)?

tobscure:
You're right about use-case 2. However...

> I'd prefer to avoid the Listener extender, by the way - hoping to keep events as implementation details in most cases.

I've been thinking about this in the context of the JavaScript API and I'd like to argue that **Events should be a part of our public API**, instead of abstracting them behind extenders. Here's why.

```php
(new Extend\Discussion())
    ->whenRenamed(function (Discussion $discussion, User $actor) {})
```

1. It's hard to do inline documentation for the closure arguments, especially documentation that IDEs will understand. As far as I can tell there's [no standard](https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md) way to do this.

    ```php
    /**
     * @param callable $listener ???
     * @return $this
     */
    public function whenRenamed(callable $listener) {}
    ```

2. It's prone to API breakage/mess. Imagine we want to add another parameter to the Renamed event. We want to pass along what the old discussion title was.

    ```php
    (new Extend\Discussion())
        ->whenRenamed(function (Discussion $discussion, User $actor, string $oldTitle) {})
    ```

    Oh, and we also want to pass along the Request that initiated the renaming action – bit of a contrived example, but say if you want to log the IP address of the user who performed the action.

    ```php
    (new Extend\Discussion())
        ->whenRenamed(function (Discussion $discussion, User $actor, string $oldTitle, Request $request) {})
    ```

    That function signature is getting a bit convoluted, which brings me to...

3. Typehinting becomes a nightmare. The responsibility for typehinting these callbacks falls on the consumer. You have to make sure you're importing the right ones, and in a `extend.php` file with a lot of listeners the `use` block is gonna get big.

This is all *especially* problematic in the JavaScript API, where because there are so many places you can hook into when rendering some components (eg. when rendering IndexPage, you can modify items in the sidebar, the nav, the toolbar, you can change parameters passed to the discussion list, etc.) you would just end up with a ridiculous number of arguments.

So how do we solve these problems? We group all of the arguments together in an object!

```php
    (new Extend\Discussion())
        ->whenRenamed(function (Renamed $event) {})
```

Look familiar?

Now we can add as many parameters to the Renamed event as we want and they will all be nicely documented and won't cause any trouble. We could even add a parameter intended to *replace* another parameter which we then deprecate.

So now the event is part of our public API, why are we bothering to wrap it in an Extender? We're just needlessly increasing our API surface area. Why don't we just do this instead:

```php
    new Extend\Listen(DiscussionRenamed::class, function (DiscussionRenamed $event) {})
```

franzliedke:
I am absolutely with you as far as typehinting domain events ("discussion was started" etc.) is concerned. Therefore, these ones should be public API, yes.

I'd still like to avoid the "listener" extender, though - because, as I see it, this one actually increases our API surface: it allows you to reach into all events that we raise internally, which I don't want to make too easy (to prevent it from becoming a common thing). Talking about events such as `ConfiguringMiddleware` (okay, that one is deprecated) - events that are implementation details, not from the domain.

Adding more small, single-purpose classes does increase the API surface as well, of course, but it does so in a limited way which allows us to determine what we want to be extensible, and what not (because we need to guarantee these continue to work).
Somewhat related: https://twitter.com/whyevernotso/status/1008473953906483200

(In the long term, we probably need to allow a more generic construct that does the same thing the current `Compat` extender does, in order to allow unforeseen extensions or experimental things. But these need to look complicated, in order to discourage using them, as they won't be guaranteed compatibility.)

franzliedke:
All of that is just about the backend extension API. Events in our JS code can be a different matter.

luceos:
A possible solution to allow the Listener extender while marking it as a pro-only solution is to mark it as deprecated permanently..

🤔 out loud

PR #1465, opened 2018-06-17
Rename and improve FormatterConfiguration extender

tobscure:
In the future we may have multiple Formatter instances, so by moving the config
callback to its own instance method we can leave the constructor
available to specify which formatter (like Assets and Routes). This
format also allows for other methods to be added.

Additionally, this adds logic to automatically flush the Formatter cache
whenever the extension is enabled or disabled.


src/Extend/Formatter.php
```diff
@@ -0,0 +1,55 @@
...
+
+        // Also set up an event listener to flush the formatter cache whenever
+        // this extension is enabled or disabled.
+        $flush = function ($event) use ($container, $extension) {
+            if ($event->extension === $extension) {
+                $container->make(ActualFormatter::class)->flush();
+            }
+        };
+
+        $events->listen(Enabled::class, $flush);
```
luceos: The `listen` method accepts an array too, so you can provide both events with one callback (not assigned to a variable).

tobscure: Learn something new every day!

src/Extend/Formatter.php
```diff
@@ -0,0 +1,55 @@
...
+    protected $callback;
+
+    public function configure(callable $callback)
+    {
+        $this->callback = $callback;
+
+        return $this;
+    }
+
+    public function __invoke(Container $container, Extension $extension = null)
```
luceos: I've seen this in Franz' code too. He was referring to `$laravel` in some situations. In the end they are all `$app` and it doesn't make sense to differ the naming all over our code. Any reason for resolving `Container` here instead of `Application`?

franzliedke: This is the extender interface. We're dealing with a container (contract) instance here, not an "app".

franzliedke:
What will be the use case for multiple formatters, by the way?

src/Extend/Formatter.php
tobscure: On that note, if an extender needed to access the Application instance, would it be expected to retrieve it from the Container? Is there an argument that we should be passing Application into extenders?

tobscure:
@franzliedke eg. if I want to set up a basic formatter for user bios which only allows bold/italics/links.

src/Extend/Formatter.php
franzliedke: I am not a big fan of extending `Application` from `Container` anyway, I might change that. So yes, you'd have to resolve it.

src/Extend/Formatter.php
```diff
@@ -0,0 +1,55 @@
...
+        // Also set up an event listener to flush the formatter cache whenever
+        // this extension is enabled or disabled.
+        $events->listen(
+            [Enabled::class, Disabled::class],
+            function ($event) use ($container, $extension) {
+                if ($event->extension === $extension) {
+                    $container->make(ActualFormatter::class)->flush();
+                }
+            }
+        );
```
franzliedke: Why did you decide to do this here, instead of generically listen to these events - for *any* extension - to flush the cache (similar to how we do it for other cache flushes, IIRC)?

franzliedke:
I rebased this, it's green now.

src/Extend/Formatter.php
tobscure: Because if an extension doesn't alter the formatter, then we don't need to clear the formatter cache when it is enabled/disabled, so why would we?

src/Extend/Formatter.php
tobscure: Separating `Application` from `Container` sounds good to me. :+1:

Issue #1387, opened 2018-03-14
Split permission for editing and deleting posts


tjrgg:
Flarum has a simple permission system. Accordingly, editing and deleting posts have been grouped into the same permission. I disagree with this choice. For our community in particular, we don't allow our moderators to edit other people's posts, only delete them and we'd like to continue this. We have an exception for announcements and updates, but those go under a separate tag, so we'd just give moderators edit permission for those tags. In any event, none of this is possible with how the permissions are currently setup, which makes this problematic.

I'm assuming I can't be the only one that would prefer to be able to control these permissions separately. I'd like to see the "Edit and delete posts" split into two separate permissions for editing posts and deleting them. This could be accomplished by replacing the current permission with two new ones, modifying the existing one and adding a new one, or keeping the existing one and adding two new permissions and requiring the existing permission along with the new ones, whatever works best.

@luceos felt that this is something that should be in the core, so hence the issue here. I was initially suggesting an extension, but I'm all for seeing this in the core for a future update (perhaps beta 8).

tobscure:
That's a fair use-case which I hadn't thought of. Happy for this to be in core, by adding a new "deletePosts" permission alongside the existing "editPosts" one.

luceos:
Relates to https://discuss.flarum.org/d/7195-split-permission-for-editing-and-deleting-posts

franzliedke:
Fixed by #1466.

PR #1466, opened 2018-06-17
Create new hidePosts permission

afrastgeek:
In this PR, I'm adding a new `hidePosts` permission.

I thought it's better to make `editPosts` literally a permission to just edit posts. And let the `hidePosts` handle permission to delete posts. (which means: Groups with `editPosts` permission will be unable to delete posts. You must have `hidePosts` permission to do so.)

I also prefer to use `hidePosts` since it's represent soft delete and `deletePosts` may represent hard deletion. Just like permission in discussion.

## Issues that will be fixed by this PR:
- #1387 (Split permission for editing and deleting posts)

### DIFF
--- a/js/src/admin/AdminApplication.js
+++ b/js/src/admin/AdminApplication.js
@@ -55,7 +55,7 @@ export default class AdminApplication extends Application {
       required.push('discussion.hide');
     }
     if (permission === 'discussion.deletePosts') {
-      required.push('discussion.editPosts');
+      required.push('discussion.hidePosts');
     }
 
     return required;
--- a/js/src/admin/components/PermissionGrid.js
+++ b/js/src/admin/components/PermissionGrid.js
@@ -206,9 +206,15 @@ export default class PermissionGrid extends Component {
 
     items.add('editPosts', {
       icon: 'fas fa-pencil-alt',
-      label: app.translator.trans('core.admin.permissions.edit_and_delete_posts_label'),
+      label: app.translator.trans('core.admin.permissions.edit_posts_label'),
       permission: 'discussion.editPosts'
     }, 70);
+
+    items.add('hidePosts', {
+      icon: 'far fa-trash-alt',
+      label: app.translator.trans('core.admin.permissions.delete_posts_label'),
+      permission: 'discussion.hidePosts'
+    }, 60);
 
     items.add('deletePosts', {
       icon: 'fas fa-times',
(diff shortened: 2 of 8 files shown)


franzliedke:
@afrastgeek Due to the changes related to Webpack, unfortunately this branch now has a merge conflict. Can you take care of this by rebasing and fixing the conflicts (mostly file renames, I suppose) or do you want us to do it?

afrastgeek:
@franzliedke Sure thing! Let me try to fix this first.
I hope i can do it later today.

TurboProgramming:
@franzliedke Looks like there are no any conflicts at the moment, should this be merged?

franzliedke:
Thank you, @afrastgeek! This will be in beta 8. :smile: 

@luceos @tobscure Do we need a migration here? Or will this be an explicit step in the upgrade guide so that people can customize to their likings? (If so, we need to add this to flarum/docs#14.)

afrastgeek:
Awesome! Sure, it's my pleasure.

tobscure:
A migration to set a default value for this new permission is probably a good idea. We'll need to add a default value in the installation code too, I think.

afrastgeek:
@tobscure Pardon me, but I believe [this line](https://github.com/flarum/core/pull/1466/files#diff-2e6d4ed85cd06d3e11f7f8428746214eR322) add default value at installation for this new permission. Isn't it?

tobscure:
@afrastgeek sorry, you're right, missed that!

franzliedke:
Okay, I [added the migration](https://github.com/flarum/core/commit/665f2413482541a4ff85ae70654088d1e8e32a13) and updated the [changelog](https://github.com/flarum/docs/issues/14).

franzliedke:
Damn, the migration [breaks the tests](https://travis-ci.org/flarum/core/jobs/404242197) (Hat-tip @luceos) - because we are running both the initial set-up of permissions in the installer, and all migrations in the migrations folder.

I guess we should just move the default permissions from the installer to a new migration with an old timestamp? /cc @tobscure

tobscure:
Can we make the `addPermissions` migration helper check for the existence of the permission first, and only add it if it doesn't exist? It might be safer to do that anyway in case someone has already manually set such a permission in their database (for whatever reason).

franzliedke:
@tobscure Apparently, [it already does that](https://github.com/flarum/core/blob/dee54a008f5e465472cd7188821cc7444a6bdc2a/src/Database/Migration.php#L144-L148). I'll refactor all permissions into migrations.

PR #1467, opened 2018-06-19
users.preferences to user notification preferences and dedicated columns

luceos:
This includes the user preferences changes needed for beta 8. As requested this has been split up from (and depends on) #1344 

- [x] add drop and reinstate statement migration, see https://github.com/flarum/core/pull/1344/commits/d9b357c18e6d1564abcde070c737a6e683e0e9b0

---

> The migrations contained in this PR have to run in-between the migrations inside the PR #1344 ; the users.preferences column is dropped inside a migration of that parent PR.


franzliedke:
Re-targeted this to master.

franzliedke:
@luceos Can you rebase this on the latest master, please?

luceos:
@flarum/core we need to decide how to move forward with this; we decided back then to store notification preferences inside their own column and (I assume) other preferences should be stored by the extension on the user table as a separate column. I'm curious whether it might be wiser to have a user_preferences table instead that combines all preferences (again but as a table instead of a json column).

franzliedke:
Toby strongly advocated using columns on the users table in [his original proposal](https://gist.github.com/tobyzerner/b189580dd17b9def574f92692faa0e56#users):

> NO Entity-Attribute-Value tables. See [here](http://www.antradar.com/blog-wordpress-and-the-curse-of-eav) and [here](http://karwin.blogspot.com.au/2009/05/eav-fail.html).

---

I tend to agree, especially as the use-case seems to be limited.

My proposal:
- New table for `notification_preferences`, that's a real domain entity, not a super-generic Entity-Attribute-Value table
- Move core preferences (`locale` and `discloseOnline` - are there others?) to dedicated columns
- Leave the column for BC - we can consider marking it deprecated and recommend using separate columns.

By the way: [JSON datatype may be useful](https://mariadb.com/kb/en/library/documentation/replication/optimization-and-tuning/optimizing-tables/entity-attribute-value-implementation/) if we decide to keep the column forever.

luceos:
followAfterReply is part of subscriptions. And this PR still needs to take care of the changes on the user (notification) preferences.

luceos:
@franzliedke I'm struggling a lot with the backward compatibility. The migration to drop the users.preferences column is in the code, but could be removed. I'd personally be rigorous and drop the column in the coming release, because keeping all the methods relating to it in the User class are a huge fat bloat.. Care to weigh in 🙏 🤗

luceos:
Keep deprecated methods noop, don't do a thing.

luceos:
Going to pull this into a new branch off of master.

Issue #1317, opened 2017-12-25
New JS Extension API


tobyzerner:
As per https://gist.github.com/tobyzerner/7613421bf93a953a33958e323d7ab67e

This will include some other specific changes:

* Refactor entity controls:
  * flarum/controls/DiscussionControls
  * controls = new DiscussionControls(discussion)
  * controls.items() (don't bother with different “sections”? Need to review the way approval/flags/akismet adding post controls currently works)
  * Better yet: discussionControls(discussion)

* `ItemList` improvements
  * Add a `toVnodes` method, remove the `listItems` helper
  * Allow method chaining
  * Consider these methods:
    * items.after('delete').add('something', hello)
    * items.before('delete').add('something', hello)
    * items.after('delete', 10).add() // specify a fallback priority
    * items.after(['foo', 'bar'], 10).add() // specify multiple items to add after

franzliedke:
When this is done, we might want to consider flarum/framework#246 as closed?

luceos:
Related to ItemList could you possible move the login related entries of the LoginModal into a ItemList while you're at this? With your fluent setters this would be awesome.

See https://github.com/flarum/core/blob/master/js/forum/src/components/LogInModal.js

This would allow an extension to disable default authentication (relates to flarum/issue-archive#380) but already gives us more flexibility.

CDK2020:
I would love if the edit user modal could also be an item list.

https://github.com/flarum/core/blob/master/js/src/forum/components/EditUserModal.js

franzliedke:
Do we want to move this to beta-9 now that the ItemList stuff has ben merged?

I guess this one should still be kept open, because "only" basic infrastructure (Webpack, heh) was done so far. In contrast to flarum/framework#851, which I will close in favor of some smaller issues, one per missing extender type.

askvortsov1:
Closing in favor of https://github.com/flarum/core/issues/2414

PR #1468, opened 2018-06-21
[WIP] JS Extender API foundation

tobscure:
Kicking off #1317

### DIFF
--- a/js/src/common/extend/Model.js
+++ b/js/src/common/extend/Model.js
@@ -1,0 +1,41 @@
+export default class Routes {
+  type;
+  attributes = [];
+  hasOnes = [];
+  hasManys = [];
+
+  constructor(type, model = null) {
+    this.type = type;
+    this.model = model;
+  }
+
+  attribute(name) {
+    this.attributes.push(name);
+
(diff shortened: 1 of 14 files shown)


franzliedke:
This puts the infrastructure in place for running an extension's JS extender, right? :+1: So, can we merge this, to keep future PRs small? :)

tobscure:
Yes, just a couple more minor changes I need to push in order to ensure that these extenders will actually work.

datitisev:
Question about importing lodash-es: Webpack will only import the export flattenDeep, and not the whole npm module, into the dist file, right?
Just wanna make sure we aren’t adding a lot of unused JS code 🙂.
I forgot so, one can just get the code from https://30secondsofcode.org

tobscure:
Correct! Webpack does tree-shaking automatically.

js/src/common/extend/Model.js
```diff
@@ -0,0 +1,41 @@
+export default class Routes {
```
franzliedke: That should be `Model`, I suppose. ;)

PR #1469, opened 2018-06-22
Allow a single extender to be returned

tobscure:
Casting an object to an array does not have the intended effect of
wrapping the object in an array. Instead we need to explicitly check
if the returned value is an array or not.

Required to merge flarum/flarum-ext-english#128


tobscure:
@luceos https://stackoverflow.com/a/2476954/887933 - casting an object to an array is like calling `get_object_vars`. We want to wrap it in an array.

PR #1511, opened 2018-07-17
Horizontal divider between main buttons and tags

TurboProgramming:
Addition of horizontal separator between main buttons and tags.

[sideNav.less](https://github.com/Flarum/core/blob/1f257d22a441e9101d4d47bb900be872d5a38943/less/common/sideNav.less) --> line 68 - 116

**Supports:**
- Dark Mode 🎨

**Screenshot**
screenshot_2018-07-17 appybuilder community

### DIFF
--- a/less/common/sideNav.less
+++ b/less/common/sideNav.less
@@ -65,8 +65,6 @@
       margin-top: 15px;
       border-bottom: 1px solid @control-bg;
     }
-    
-    
     .sideNav .Dropdown--select .Dropdown-menu>.Dropdown-separator {
     background:none
     }
--- a/less/common/sideNav.less
+++ b/less/common/sideNav.less
@@ -65,6 +65,58 @@
       margin-top: 15px;
       border-bottom: 1px solid @control-bg;
     }
+    
+    
+    .sideNav .Dropdown--select .Dropdown-menu>.Dropdown-separator {
+    background:none
+    }
+    
+    .sideNav--horizontal .Dropdown-separator {
+    display:none
+    }
+    
+    .sideNav .Dropdown-separator {
+    display:none
+    }
+    
+    .Dropdown-separator {
+    margin:8px 0;
+    background-color:#e4e8f6;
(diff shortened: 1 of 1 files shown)


datitisev:
Why are you modifying the styles for the composer separator if this is only supposed to affect the navbar ?
Also, does this work when the tags extension is _not_ enabled?

less/common/sideNav.less
```diff
@@ -65,6 +65,56 @@
...
+    @media (min-width: 768px) {
+    .Dropdown--split .Dropdown-menu li:first-child,.Dropdown--split .Dropdown-menu li:first-child+li.Dropdown-separator {
+    display:none
+    }
+ }
+    .Dropdown-separator {
+    margin:0
+    }
+    
+    .TagsPage .sideNav .Dropdown-separator {
```
franzliedke: Styling (or code) related to a core extension should never be part of flarum/core.

franzliedke:
That said, I believe Toby styled things intentionally. Unless you can provide a very good reason why we should change this, I am pretty confident this won't get merged.

Closing for now.

TurboProgramming:
A space could be a great way to separate stuff but this divider really looked great.

With inspector, it doesn't work, but I believe it's an issue with less and vanilla CSS again.

TurboProgramming:
@franzliedke Can you reopen this PR as I will be modifying some stuff?

Maybe we'll decide to accept this change, I can't modify it for the PR.

TurboProgramming:
> Why are you modifying the styles for the composer separator if this is only supposed to affect the navbar?

I removed the lines, I added them by mistake. Sorry 😞 

> Also, does this work when the tags extension is not enabled?

Probably no, if this PR gets opened again, I can make modify some stuff to make it work only if the tags extension is enabled.

(@datitisev)

franzliedke:
We can reopen it in case Toby likes the idea.

For now, let's save some time for all of us.

PR #1512, opened 2018-07-19
Grant "own" abilities via permissions so that they can be revoked

### LINKED ISSUE #1419: [Suspend] Suspended users can edit/delete own posts
There's not much needed to recreate this, suspend a test user with a post or two and then attempt to edit or delete the past posts as that user.

This appears to be unintentional, but presents some real problems for security and moderation. It's possible for a user to erase evidence of why they were suspended, to openly dispute their suspension, or use it to retaliate by editing/removing critical posts.

tobscure:
**Fixes #1419**

Whereas previously the ability to rename or hide one's own discussion,
or to edit one's own post, was granted in policy logic, this change
grants it by default and then revokes it using inverse policy logic.

This means that the abilities are granted only to Members, so that when
extensions like Suspend erase a user's groups, the user will not be
incorrectly granted permission to edit their own stuff.

**Related PRs**
https://github.com/flarum/flarum-ext-tags/pull/56

### DIFF
--- a/src/Discussion/DiscussionPolicy.php
+++ b/src/Discussion/DiscussionPolicy.php
@@ -56,11 +56,16 @@ class DiscussionPolicy extends AbstractPolicy
     /**
      * @param User $actor
      * @param string $ability
+     * @param Discussion $discussion
      * @return bool|null
      */
-    public function can(User $actor, $ability)
+    public function can(User $actor, $ability, Discussion $discussion)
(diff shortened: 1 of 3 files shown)


tobscure:
Realised this approach won't work because then the permissions will be revoked when they shouldn't be, like for moderators. Will have to think of another solution for #1419.

PR #1513, opened 2018-07-20
Make "own" permissions depend on the user's ability to reply

### LINKED ISSUE #1419: [Suspend] Suspended users can edit/delete own posts

tobscure:
Permission to rename/hide/edit one's own discussion/post is only granted
if the user has permission to reply to the discussion. This makes sense
if you think of these actions as forms of "replying" to a discussion.

**Fixes #1419** because suspended users do not have permission to reply to
discussions, therefore they will not be granted these "own" permissions.

**Related PRs**
https://github.com/flarum/flarum-ext-tags/pull/57
https://github.com/flarum/flarum-ext-lock/pull/15

### DIFF
--- a/src/Discussion/DiscussionPolicy.php
+++ b/src/Discussion/DiscussionPolicy.php
@@ -123,7 +123,7 @@ class DiscussionPolicy extends AbstractPolicy
      */
     public function rename(User $actor, Discussion $discussion)
     {
-        if ($discussion->user_id == $actor->id) {
+        if ($discussion->user_id == $actor->id && $actor->can('reply', $discussion)) {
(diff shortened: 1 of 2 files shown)


luceos:
It would be awesome to be able to confirm this inner working by adding a test (at some point), just to make sure it does not break along the way.

franzliedke:
@tobscure Please rebase this and the two extension PRs now that DB changes have been merged.

PR #1514, opened 2018-07-20
Auth token and avatarUrl security improvements

### LINKED ISSUE #1357: Avatar upload can be used to leak server IP
## Explanation

I'm a big fan of services like Cloudflare but I've noticed the way Flarum handles avatars allows anybody to leak the server IP.

When registering or updating your user settings, you can pass an `avatarUrl` attribute that will be downloaded and set as avatar.

This looks deliberate, however this allows any user to ask the server to perform a request to any given url.

Maybe this should be limited to admins or controlled via a permission ?

## Technical details

- Version of Flarum: master

Both `Flarum\User\Command\RegisterUserHandler` and `Flarum\User\Command\EditUserHandler` allow the use of `avatarUrl` without restrictions.

### LINKED ISSUE #1359: Shortcomings of abstract oauth 2 controller
The `AbstractOAuth2Controller` is a bliss, no doubt about it. However,

- No events are triggered during the oauth procedure.
- Although the AuthToken class stores the `$identification` array (which is used within the select), one cannot store additional information to pass along to a triggered event.

What I'd like to propose is to add Events related to the OAuth2 flow. And in addition store the $suggestions as well as the source/service authorizing the authentication.

A usecase for this would be the patronage extension, automatically adding users authenticated from patreon who pledge the campaign to be assigned a specific user group.

###

Affected:

- [ ] https://github.com/flarum/core/tree/1359-oauth2-improvements
- [ ] https://github.com/flarum/flarum-ext-auth-github/tree/1359-oauth2-improvements
- [ ] https://github.com/flarum/flarum-ext-auth-facebook/tree/1359-oauth2-improvements

tobyzerner:
**Supersedes flarum/framework#1388**
**Fixes flarum/framework#1357**
**Fixes flarum/framework#1359**

This PR updates the AuthenticationResponseFactory class so that you can add additional data to an auth token's payload *without* it being used to identify the user. This allows the avatarUrl to be safely stored in the auth token, rather than passed through the frontend as a "suggestion".

The ability to set an avatarUrl as an API attribute is removed completely, fixing flarum/framework#1357.

This PR also removes AbstractOAuth2Controller. There is no reason to provide an implementation for a specific oAuth2 library in core; it's not generic enough (eg. auth-twitter can't use it).

**Confirmed**

**Required changes:**

* Related core extension PRs: 
https://github.com/flarum/flarum-ext-auth-facebook/pull/15
https://github.com/flarum/flarum-ext-auth-github/pull/12
https://github.com/flarum/flarum-ext-auth-twitter/pull/11


luceos:
Scanned over the PR real quick, it seems to me that passing over the full payload will be sufficient for any use case.

Not sure if the column size is sufficient for all use cases, especially if multiple scopes have been enabled providing json about campaigns, pledges and what not.. I think that was also the reason I had not pushed the full payload along; this could be circumvented in the implementation of a Patreon controller easily by removing unnecessary payload.

tobscure:
@luceos My concern is that when registration takes place the items in the payload are automatically set as attributes on the user. I guess this won't be flexible enough?

luceos:
> payload are automatically set as attributes on the user. 

I am not sure i understand but if you suggest that any property in the payload will be assigned to the user than you will have an issue. Eg we would need to know what tier the user is on at patreon.com; we allow passing that over so that after creation we can assign the role. The tier or tier_id column does not exist, so the database will complain that a user is being saved with attributes that it does no columns for.

So we need a way to send additional payload from the response of the oauth2 server to after user creation/updating without setting them on the user.

franzliedke (dismissed):
Code looks great.

src/Forum/AuthenticationResponseFactory.php
```diff
@@ -12,107 +12,79 @@
...
-
-            $payload = array_merge(
-                $identification,
-                $suggestions,
-                ['token' => $token->id],
-                // List of the fields that can't be edited during sign up
-                ['identificationFields' => array_keys($identification)]
-            );
-        }
+        $payload = array_merge($identification, $suggestions);
```
franzliedke: Not sure if this is critical at all, but I could imagine that preferring the keys in `$identification` over those in `$suggestions` (by swapping the parameter order here) might be somewhat safer?

tobyzerner:
I'm going to amend this to address the flexibility issues discussed above, and to fix flarum/issue-archive#332.

tobscure:
Have just pushed changes. See example usage in https://github.com/flarum/flarum-ext-auth-facebook/pull/15. /cc @luceos @franzliedke

migrations/2018_07_22_000000_create_login_providers_table.php
```diff
@@ -0,0 +1,28 @@
...
+
+use Flarum\Database\Migration;
+use Illuminate\Database\Schema\Blueprint;
+
+return Migration::createTable(
+    'login_providers',
+    function (Blueprint $table) {
+        $table->increments('id');
+        $table->unsignedInteger('user_id');
+        $table->string('provider', 100);
```
luceos: Is provider supposed to be a FQCN? If so I'm not sure that 100 will suffice, some providers might have very long namespaces..

tobscure: Nah it's just a string to uniquely identify the provider. e.g. in 
https://github.com/flarum/flarum-ext-auth-facebook/pull/15 it's `facebook`.

luceos (approved):

franzliedke (approved):
I rebased, fixed the conflict and fixed the tests (related to migrations being out-of-date).

auth-twitter and auth-github still need an update. Once these are done, feel free to merge. Great work. :+1:

franzliedke:
So this fixes flarum/framework#1357, flarum/framework#1359 and flarum/issue-archive#332? If that's correct, please make sure all of them get closed (and assigned to mileston beta8) as well.

EDIT: There are some other issues referenced in some of these, you might wanna double-check those, too.

PR #1528, opened 2018-08-05
Change token error message

### LINKED ISSUE #1337: Clicking activation link multiple times results in error message
Flarum 0.1.0-beta.7

When an activation email is sent and the link is clicked multiple times (usually by accident), the first request will activate the user and second request will result in the error message "Invalid confirmation token".

This confuses users as it indicates something went wrong, when everything is in fact in order.

There should be a more friendly message when an user clicks an activation link while having an already activated account. Or maybe send them to the forum main page directly.

Thanks!
Flarum is great, keep it up!

datitisev:
**Fixes #1337**

- Error message mentions possibility of the confirmation token having already been used

- Language
- _Maybe_ creating an exception for custom error messages (similar to 404 but more customizable)

**Screenshot**
image

**Confirmed**

### DIFF
--- a/src/Forum/Controller/ConfirmEmailController.php
+++ b/src/Forum/Controller/ConfirmEmailController.php
@@ -20,7 +20,6 @@ use Illuminate\Contracts\Bus\Dispatcher;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface as Request;
 use Psr\Http\Server\RequestHandlerInterface;
-use Zend\Diactoros\Response\HtmlResponse;
 use Zend\Diactoros\Response\RedirectResponse;
 
 class ConfirmEmailController implements RequestHandlerInterface
--- a/src/Forum/Controller/ConfirmEmailController.php
+++ b/src/Forum/Controller/ConfirmEmailController.php
@@ -64,7 +64,7 @@ class ConfirmEmailController implements RequestHandlerInterface
                 new ConfirmEmail($token)
             );
         } catch (InvalidConfirmationTokenException $e) {
-            return new HtmlResponse('Invalid confirmation token');
+            return new HtmlResponse('The confirmation token is either invalid or has already been used');
         }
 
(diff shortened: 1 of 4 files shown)


tobscure:
Thanks. Hmm you're right we do need some way to easily render the nice error view (as in Http\Middleware\HandleErrors) with a custom message/translation. Any thoughts @franzliedke?

datitisev:
I created a new exception, `CustomException` (may want to rename to `CustomErrorException` tho that sounds stupid), which uses the translator for the message provided and accepts any valid HTTP status (`status >= 400 && status < 600`).

I wasn't sure if to use the view property or check by class, but I thought that by checking for the view property, other extensions could use their own error views by just settings the `view` property on the exception.

luceos:
I don't think calling the exception literally a CustomException is what we need. How about something like `InvalidTokenException` or `TokenAuthorizationException`?

datitisev:
Hm, I was going for what @tobscure commented above.
Though I now see that he was going more for a change in `HandleErrors`, but that'd still need an exception to throw, and simply throwing `Exception` doesn't seem like a good idea.
I'll wait until we decide what we want to do.

tobscure:
@luceos is right. We don't even need to make a new exception, we just need to add a handler that can translate the existing InvalidConfirmationTokenException into an error view. Pretty much exactly what we're already doing in Flarum\Api\ExceptionHandler - perhaps we could even move that whole namespace out into Flarum\Http, add English translations to the responses, and then simply convert the API error format into a view.

luceos:
Feel free to give this a shot @datitisev 👍

franzliedke:
@datitisev Do you need help to get started with Toby's suggestion?

datitisev:
I forgot about this, to be completely honest. I'll try to get started on it soon. Haven't actually begun yet.

franzliedke:
#1843 should make this much easier to build now.

datitisev:
:+1:

PR #1530, opened 2018-08-09
ModalManager: Fix issue when app.current is not defined

datitisev:
- Prevent TypeError when `app.current` is not defined when setting `app.current.retain` to `true`

- If `app.current` should *ever* be undefined

**Screenshot**
image

**Confirmed**


datitisev:
This error occurs when `super.init()` is not called from within the `init` method in a class that extends `Page`.

tobscure:
@datitisev OK. I think it's still fair to add this check here in case there's ever an instance where a modal is displayed but there isn't a current page.

PR #1539, opened 2018-08-12
Stop '@' and '*' from crashing the forum

### LINKED ISSUE #1498: "@" and "*" crashing forum if searched
## Bug Report

**Current Behavior**
I know, this is not something that prevents usability, but it crashes the forum: if I search for @ or * using the search bar, the forum will crash until the ?q=%40 or ?q=* gets deleted from the URL bar.

**Steps to Reproduce**
1. Go to homepage
2. Search for @ or *
3. See error

**Expected Behavior**
The forum should really search for articles including those characters.

**Screenshots**

After searching for "@":

screenshot_2018-07-05 flarum community 1

After searching for "*":

screenshot_2018-07-05 flarum community 2

After refreshing the website without deleting the search query from URL:

screenshot_2018-07-05 flarum community 3

**Environment**
Flarum Community
https://discuss.flarum.org

**Additional Context**
I only found these for now, but we may have other characters doing the same (or worse).

datitisev:
**Fixes #1498**

- FulltextGambit replaces `@` and `*` with nothing
- DiscussionListItem now retrieves the current search from `app.current.searching` (same as `Search`) instead from the unformatted query
- `IndexPage#searching` removes `@` and `*`
- `Search#view` formats the input on change before setting to `this.value`, therefore making uses unable of typing `@` or `*` (nothing happens)

- __Same issue__ occurs with characters such as `?`. We may want to restrict search to alphanumeric characters
image

**Screenshot**

Before:
image

After:
image

**Confirmed**

- [X] Backend changes: tests are green (run `php vendor/bin/phpunit`).

### DIFF
--- a/js/src/forum/components/DiscussionListItem.js
+++ b/js/src/forum/components/DiscussionListItem.js
@@ -72,7 +72,7 @@ export default class DiscussionListItem extends Component {
         jumpTo = post.number();
       }
 
-      const phrase = this.props.params.q;
+      const phrase = app.current.searching();
       this.highlightRegExp = new RegExp(phrase+'|'+phrase.trim().replace(/\s+/g, '|'), 'gi');
     } else {
       jumpTo = Math.min(discussion.lastPostNumber(), (discussion.readNumber() || 0) + 1);
--- a/js/src/forum/components/IndexPage.js
+++ b/js/src/forum/components/IndexPage.js
@@ -280,7 +280,7 @@ export default class IndexPage extends Page {
    * @return {String}
    */
   searching() {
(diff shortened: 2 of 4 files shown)


tobscure:
Thanks for the attempt. I'm closing this as I think this is taking the wrong approach. As per https://github.com/flarum/core/issues/1498#issuecomment-404962789, all we need to do is remove leading `*` characters from the search term. Also, the search query needs to be [escaped](https://stackoverflow.com/questions/3446170/escape-string-for-use-in-javascript-regex) in DiscussionListItem before being inserted into the regular expression.

PR #1540, opened 2018-08-12
Allow admins to see last online text

### LINKED ISSUE #1256: User last online text should be visible to admins
The ability to hide online status from other users shouldn't apply to admins, they should have a default override. There are instances where an admin would need to know if a user has been online even if they haven't posted anything. 

Ideally this would be a permission that could be handed out to mods as well. If ever there appears an Edit User permission, this would go hand in hand, IMHO. For now, just admins seems suitable, since Flarum only allows admins to edit or delete users.

datitisev:
**Fixes #1256**

- Allow admins to see last online text

- We may want a permission for this, however I don't think it is needed because it's such a small thing

**Screenshots:**
image
image

**Confirmed**

**Required changes**
- [x] Related core extension PRs: flarum/flarum-ext-english#132


tobscure:
Hmm I wonder about the utility of a new permission. Could be useful to allow mods to see last online activity?

tobscure:
`user.viewLastSeenAt`

datitisev:
@tobscure Done!

luceos:
@datitisev just to point out, commits are also pushing the PR into the notifications dashboard/bell of GitHub.

PS I love this PR! Things like these make a huge difference to mods.

tobscure:
This is not quite right. I think `viewLastSeenAt` should be an ability for a user. i.e.

* the permission should be called `user.viewLastSeenAt`
* no need to add an attribute to `ForumSerializer` as it's not a global permission
* the permission check in `UserSerializer` should read `$this->actor->can('viewLastSeenAt', $user)`

See flarum-ext-suspend for an example

datitisev:
Should the icon be the solid or regular version?
I see most icons are solid, but some like the trash for deleting discussions are regular.

image

ardacebi:
Probably regular. It should look better with the others.

franzliedke:
Can we get a PR to ext-english as well?

@luceos Maybe worth another item in our checklist? :thinking:

datitisev:
@franzliedke Ok, I'll do that when I have time. 

FYI, It is an item in the checklist, but it wasn't required for the previous changes that didn't add any translations and only checked for admin.

franzliedke:
Now that the database changes have been merged, we have a small conflict. Can you take care of that, please, @datitisev ?

migrations/2018_07_21_000100_seed_default_group_permissions.php
```diff
@@ -27,6 +27,7 @@
     ['permission' => 'discussion.hidePosts', 'group_id' => Group::MODERATOR_ID],
     ['permission' => 'discussion.rename', 'group_id' => Group::MODERATOR_ID],
     ['permission' => 'discussion.viewIpsPosts', 'group_id' => Group::MODERATOR_ID],
+    ['permission' => 'user.viewLastSeenAt', 'group_id' => Group::MODERATOR_ID],
```
franzliedke: Adding this to the existing migration means we need to add this as a manual action to the upgrade guide, right? Or am I missing something?

datitisev: @franzliedke Not sure... I guess that'd be something to test along with everything else for a beta 7 -> 8 conversion because of merging db changes...

datitisev: @franzliedke As it's a new file, I assume there'd be no problems... Flarum will load the whole thing on new or updated installation.

tobscure: Yeah, this migration is dated after beta 7 release.

PR #1543, opened 2018-08-12
Change tab title after renaming discussion

### LINKED ISSUE #1521: After changing title, discussion title changes but page (tab) title stays the same
## Bug Report

**Current Behavior**
After a title change, the discussion title that is located under the tags does change, but the page (tab) title won't change until next refresh.

**Steps to Reproduce**
1. Go to a discussion
2. Click on edit title
3. Edit title

**Expected Behavior**
Page (tab) title should change with the discussion title.

**Screenshots**
htfdhfdhgdfhgfd

**Environment**
Flarum Community
https://discuss.flarum.org

datitisev:
**Fixes #1521**

- Set title after renaming discussion

**Confirmed**


franzliedke:
What's with the lockfile change?

js/src/forum/components/RenameDiscussionModal.js
```diff
@@ -58,6 +58,7 @@ export default class RenameDiscussionModal extends Modal {
           app.current.stream.update();
         }
         m.redraw();
+        app.setTitle(title());
```
franzliedke: Hmm, this feels wrong in terms of the virtual DOM / always re-render model. Shouldn't this title be set whenever the discussion page is re-rendered? (This is just one of the places where we trigger a redraw of that page - that redraw in turn should take care of the title change.)

CDK2020:
@franzliedke It's from installing the npm modules, we can remove it if you need.

datitisev:
@franzliedke Not sure. All it seems to do is modify a property for every package. All I did was run `npm install`, didn't update any package.

image

franzliedke:
Strange. Probably a version conflict with NPM? Which version do you have?

CDK2020:
@franzliedke it's version 6.2.0

js/src/forum/components/RenameDiscussionModal.js
CDK2020: That's true, however if it did update this PR wouldn't exist

datitisev:
@franzliedke I'll reset the lockfile changes if it's causing such a commotion 😂

js/src/forum/components/RenameDiscussionModal.js
datitisev: Looks like the title is currently only set on https://github.com/flarum/core/blob/master/js/src/forum/components/DiscussionPage.js#L163 (DiscussionPage#show), which only runs when DiscussionPage#refresh runs, aka on first load of Discussion, I assume.

tobscure:
Almost. Correct place to do this would be in `config` (theoretically `view` can be called without the view actually being rendered, whereas `config` is only called if the view has been rendered).

datitisev:
@tobscure Done! Had to add a check for if `this.discussion` was set or not, seems like it was running before `DiscussionPage#show` was

PR #1547, opened 2018-08-15
Fix for SourceMaps not adding extension source maps

### LINKED ISSUE #1538: JS errors don't show the correct location when clicked in the console
## Bug Report

**Current Behavior**
A clear and concise description of the behavior.

**Steps to Reproduce**
1. Add error to js dist file
2. Look for error on js console
3. Click wrong location

**Expected Behavior**
It should show where the error in JS is

**Screenshots**
image

image

**Environment**
- Flarum version: dev-database-changes
- Browser: Chrome 68

**Possible Solution**
Could be related to Webpack maps.

CDK2020:
Extension's sourcemaps weren't being added to the main assets map file causing wrong line numbers to be shown on a break point. 

* **Fixes** #1538

**Screenshots**

Before: 
image

After: 
image

Before: 
image

After: 
image

### DIFF
--- a/src/Frontend/Asset/ExtensionAssets.php
+++ b/src/Frontend/Asset/ExtensionAssets.php
@@ -46,6 +46,9 @@ class ExtensionAssets implements AssetInterface
     public function js(SourceCollector $sources)
     {
         if ($this->js) {
+			$sources->addString(function() {
+				return $this->js.'.map';
+			});
             $sources->addString(function () {
                 $name = $this->extension->getId();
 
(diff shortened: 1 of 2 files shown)


CDK2020:
So I have source maps working almost properly, the only thing I am having trouble with is that comments & whitespace in the js code is causing the source map to be off be a few lines

datitisev:
@CDK2020 That could maybe be because the extension's source map itself ignores comments & whitespace... that may be something we may want to change, though I think it won't be such big of an issue as one may think.

CDK2020:
David just tried it out and it worked fine, removing the WIP label as this is ready to go now

datitisev:
I cannot seem to reproduce that bug. For me, it works fine.

image
image

However, I think it'd be useful if _somehow_ the error indicated the correct extension without having to click on it.
The path of the file is `Module../src/admin/index.js`, it'd be nice if it said something along the lines of `extension-name/src/admin/index.js`. Not sure if that's possible, though.
img

tobscure:
Thanks for the effort. Fixed in a simpler way in 5f83285.

PR #1560, opened 2018-08-24
Update ExtensionPage Style

CDK2020:
# EDIT: Please see [this comment](https://github.com/flarum/core/pull/1560#issuecomment-416083728) for the most updated info

#### Changes proposed in this pull request:
Make the ExtensionsPage 2 column and hide the description on mobile

#### Screenshots:

**Desktop:**

Before:

image

After:
image

**Mobile:**

Before:

image

After:

image


ardacebi:
Thank you so much for this @CDK2020, especially for the 2 column feature. :)

gurjyot:
Actually, having a description for extensions in mobile wasn't bad. I would rather suggest providing a drop-down description in mobile view if you wanna hide it in mobile. Or just don't hide it at all. 
What you did for desktop view, that's perfect. :)

ardacebi:
@gurjyot I'm partial with you on the mobile description thing. I never tested the extension list on a mobile device but what kind of issues can it produce?

CDK2020:
@gurjyot I think hiding it is the best option, It gets very squished with the 2 column look. Plus, there weren't descriptions before.

CDK2020:
@gurjyot for your reference. here are the two changes proposed when squashed.

With descriptions removed:

image


Descriptions included: 
image

gurjyot:
Even if we go with the first way, not having complete names upfront is just a mess. 
In this situation, I would like a single column implementation much better.

datitisev:
For a smaller layout, i.e. <300px (I think that 992px is probably more for tablets), most likely one column would be preferable.

CDK2020:
I've updated it to have 3 columns, but also reduce the number of columns if the screen can't fit the description, as well as completely hiding the description when it can't be fully shown (a screen less than 410px wide)

image

image

image

image

less/admin/ExtensionsPage.less
```diff
@@ -96,3 +99,13 @@
   margin-left: -50px;
   position: absolute;
 }
+
+@media (max-width: 410px) {
```
franzliedke: Shouldn't this use on of our named breakpoints?

less/admin/ExtensionsPage.less
```diff
@@ -96,3 +99,13 @@
...
+   .ExtensionListItem-description {
+        display: none;
+    }
+    .ExtensionListItem-version {
+        display: block;
+        margin-top: 8px;
```
franzliedke: Also, indentation seems wrong here. :stuck_out_tongue_winking_eye:

tobscure:
I don't think it's a huge deal. In this case the 410px is preferable because that's the min-width of the columns?

CDK2020:
@tobscure that's correct, this is the min-width of the columns. We can't use the breakpoints because those will hide the description before we need to.

franzliedke:
In that case, you should create a new named breakpoint, to prevent magic numbers in the stylesheet.

CDK2020:
@franzliedke after some research, there is no display on a phone currently in use with a width less than 410, in that case, I now favor removing the media constraint all together since it will foreseeably never be used

datitisev:
@CDK2020 There may be some browsers that decrease the viewport size to make everything bigger, in which case you may want to hide the description.
It should probably remain included for those rare cases. It barely adds to the compiled CSS size.

CDK2020:
@datitisev can you give an example of a browser doing that? How do you suggest we satisfy both your edge case as well as Franz's request?

CDK2020:
@datitisev I feel like it is counter intuitive to make a breakpoint for just this one case, @franzliedke what part about `max-width` do you not like?

franzliedke:
@CDK2020 The 410 can be considered a "magic number". When reading this stylesheet, you're left to wonder: Why this value and not any other? By using a breakpoint variable, we a) give it a name and b) place its definition with the other breakpoints, so that it cannot be forgotten when those change.

tobscure:
I would've thought that the variable should be named for this specific use-case, eg. `@extension-list-column-width`. Probably don't need a variable for the media query, that can just go straight in ExtensionsPage.less. Is this what you had in mind @franzliedke?

CDK2020:
@tobscure I've updated it per your request. Sorry for the trouble, I didn't quite understand what Franz was asking for.

franzliedke:
Hmm, now you've added your very specific variable under the "Breakpoints" heading in variables.scss. That's not the right place.

If this number is specific for this page (as it seems to be), it's fine to have it in ExtensionsPage.less - but at least we give the concept a name, and cannot forget to change the breakpoint when we change the width of the column.

CDK2020:
Hopefully that is the last commit needed!

tobscure:
Thanks @CDK2020!

PR #1562, opened 2018-09-03
InfoCommand improvements

datitisev:
- Ignore error output of `git rev-parse-head`
  - Would output the following if a git repository existed but had no commits:
    ```bash
    fatal: ambiguous argument 'HEAD': unknown revision or path not in the working tree.
    Use '--' to separate paths from revisions, like this:
    'git <command> [<revision>...] -- [<file>...]'
    ```
- Pad extension id & version to make look cleaner

**Screenshot**

Before:
image

After:
image

**Confirmed**

- [X] Backend changes: tests are green (run `php vendor/bin/phpunit`).

### DIFF
--- a/src/Foundation/Console/InfoCommand.php
+++ b/src/Foundation/Console/InfoCommand.php
@@ -64,8 +64,9 @@ class InfoCommand extends AbstractCommand
 
         foreach ($this->extensions->getEnabledExtensions() as $extension) {
             /* @var \Flarum\Extension\Extension $extension */
-            $name = $extension->getId();
-            $version = $this->findPackageVersion($extension->getPath(), $extension->getVersion());
+            $name = str_pad($extension->getId(), 25);
+            $fallback = str_pad($extension->getVersion(), 15);
+            $version = $this->findPackageVersion($extension->getPath(), $fallback);
 
             $this->info("EXT $name $version");
         }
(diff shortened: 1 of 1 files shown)


datitisev:
@franzliedke Looks like this now:

image

---

I added a new heading & formatted all the text to use `<info>` color. I tried to make the first row heading to use all 3 cells, but with the formatting it extends a bit too far.

<details> 
  <summary>Failed attempt </summary>
  
</details>

src/Foundation/Console/InfoCommand.php
```diff
@@ -91,20 +116,20 @@ protected function fire()
...
+    private function findPackageVersion($path, $fallback = null)
     {
         if (file_exists("$path/.git")) {
             $cwd = getcwd();
             chdir($path);
 
             $output = [];
             $status = null;
-            exec('git rev-parse HEAD', $output, $status);
+            exec('git rev-parse HEAD 2> /dev/null', $output, $status);
```
franzliedke: Hmm, how does this behave on Windows?

franzliedke: @datitisev See commit 96045ca, that should fix it, I think.

datitisev: @franzliedke Looks like it still works on macOS. :+1:

Issue #1504, opened 2018-07-11
Start a Discussion Tag Popup closes in IE 11


D-mata:
When creating a new discussion using "Start a Discussion", the modal that appears if a tag is not chosen closes immediately after it appears and leaves the default overlay.

**Steps to Reproduce**
1. Go to 'Start a Discussion'
2. Fill out Title and Post text
3. Then click "Post Discussion"
4. See the popup appear then fade away

franzliedke:
Can anybody reproduce this?

@D-mata Do you see any error messages in the browser's console?

TurboProgramming:
I tested on IE 9, and it works fine. Not sure about IE 11.

![ezgif-4-a352cfbd77](https://user-images.githubusercontent.com/17576065/42734980-511e3246-8855-11e8-8b85-87e29cad711d.gif)

**Environment**
http://104.251.211.123 (Beta 7)

TurboProgramming:
I just tested this on IE 11 too. I can't reproduce this and with IE 11, I'm able to create, delete and edit a post fine.

D-mata:
![flarum-ie11-issue](https://user-images.githubusercontent.com/11344451/43019502-129d186e-8c12-11e8-86b2-30fac11129a9.gif)

There are no errors in the console window, and none in network calls.

TurboProgramming:
https://www.browserling.com/browse/win/7/ie/11/http%3A%2F%2Fdiscuss.flarum.org

I tried here and I see the same output @D-mata is getting. Try it yourself.

TurboProgramming:
> I tried here and I see the exact same output @D-mata is getting. Try it yourself. (Confirming)

Just to make things right, I confirmed this and experiencing the same @D-mata sent on his GIF. Can we remove the needs confirmation label?

D-mata:
So it appears to be this line that doesnt work in IE11:
```
 .one('bsTransitionEnd', function ()
           that.$element.trigger('focus').trigger(e)
```
Removing the command doesn't seem to affect anything. I may try to submit the code change later today.

tankerkiller125:
It appears that as of beta11.1 discuss no longer functions at all in IE11, what is interesting is that my forum that has polyfill.io enabled (added myself) everything works completely as expected including the modules.

therealsujitk:
##Temporary Solution

[...]

PR #1570, opened 2018-09-18
Update AvatarEditor.js onchange to oninput

### LINKED ISSUE #1125: MS EDGE, Avatar upload, doesn't work.
When user browsing flarum site using MS Edge tries to upload avatar image, no image is uploaded. No errors appear in browser console.

D-mata:
onchange does not work in IE11 and other IE browsers. This change works with all modern browsers as well.

**Fixes #1125**

**Screenshot**

**Confirmed**

**Required changes:**

### DIFF
--- a/js/src/forum/components/AvatarEditor.js
+++ b/js/src/forum/components/AvatarEditor.js
@@ -150,7 +150,7 @@ export default class AvatarEditor extends Component {
     const user = this.props.user;
     const $input = $('<input type="file">');
 
-    $input.appendTo('body').hide().click().on('change', e => {
+    $input.appendTo('body').hide().click().on('input', e => {
       this.upload($(e.target)[0].files[0]);
     });
   }


D-mata:
https://github.com/flarum/core/issues/1504

Referenced here

luceos:
Have you tested this? It's a required step for Frontend changes in the PR template...

D-mata:
flarum-ie11-issue

Yes, here is an example.

D-mata:
https://github.com/flarum/core/issues/1572
I posted it here.

franzliedke:
@D-mata I think @luceos is asking whether the proposed fix solves the problem for you. Your GIF (and the linked issue) only show the issue, as far as I can tell.

franzliedke:
I am a bit scared that this will cause regressions elsewhere (i.e. other browsers). Can you link to a source that points to this as a common solution for such a problem?

MichaelBelgium:
Would there be a chance if the issue is about arrow functions? [Arrow functions are never supported in IE](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#Browser_compatibility)

What if the fix is this:

```
$input.appendTo('body').hide().click().on('change', function(e)  {
      this.upload($(e.target)[0].files[0]);
    });
```

franzliedke:
@MichaelBelgium No, the code is compiled down to IE-compatible syntax, see [this file](https://github.com/flarum/core/blob/565131e2a75d80970229970185b45435e18f11ae/js/dist/forum.js) (you can search there for `on("change")`).

MichaelBelgium:
> @MichaelBelgium No, the code is compiled down to IE-compatible syntax, see [this file](https://github.com/flarum/core/blob/565131e2a75d80970229970185b45435e18f11ae/js/dist/forum.js) (you can search there for `on("change")`).

Ooh, ok got it

franzliedke:
If anybody has time to research / test whether and how this works in other browsers, that would be a great help!

askvortsov1:
> If anybody has time to research / test whether and how this works in other browsers, that would be a great help!

Not sure if this is what you're looking for, but just according to browser compatability maps, oninput is supported more than onchange:

onInput:
https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/oninput#Browser_compatibility

onChange:
https://developer.mozilla.org/en-US/docs/Web/API/PresentationAvailability/onchange

tankerkiller125:
@askvortsov1 The documentation you linked to are for two completely separate APIs, the correct docs are located as https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/oninput and https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onchange

askvortsov1:
Ah that's awkward my bad, sorry for the confusion

PR #1576, opened 2018-09-22
Make user posts page similar to discussion page, add empty text

datitisev:
- Make `PostsUserPage` look more like `DiscussionsUserPage` (which uses `DiscussionList`)

- Locale text

**Screenshot**

**Confirmed**

**Required changes:**

- [x] Related core extension PRs: https://github.com/flarum/flarum-ext-english/pull/133

### DIFF
--- a/js/src/forum/components/PostsUserPage.js
+++ b/js/src/forum/components/PostsUserPage.js
@@ -1,6 +1,7 @@
 import UserPage from './UserPage';
 import LoadingIndicator from '../../common/components/LoadingIndicator';
 import Button from '../../common/components/Button';
+import Placeholder from '../../common/components/Placeholder';
 import CommentPost from './CommentPost';
 
 /**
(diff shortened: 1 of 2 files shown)


js/src/forum/components/PostsUserPage.js
```diff
@@ -59,6 +60,15 @@ export default class PostsUserPage extends UserPage {
...
     }
 
+    if (this.posts.length === 0 && !this.loading) {
+      const text = app.translator.trans('core.forum.user.posts_empty_text');
+      return (
+        <div className="PostsUserPage">
+          {Placeholder.component({text})}
+        </div>
+      );
+    }
```
franzliedke: This entire if statement could probably go to the very top of the method - we don't need to set the `footer` variable if we're not using it anyway...

datitisev: @franzliedke Hm, true. It would only shave off a millisecond or two in performance, though, if at all. The function is not running all the time... unless a madman with a loop & `m.redraw` comes along 😂

tobscure: 1736fe3f585b551c7b858b1c0f49a2b5b83664a8

franzliedke: It's not so much about performance as about clarity. Thanks, Toby!

PR #1577, opened 2018-09-22
Database indices

### LINKED ISSUE #127: SQL query audit + add database indexes

tobscure:
**Fixes #127**

**Confirmed**

**Related changes**

- https://github.com/flarum/flarum-ext-flags/pull/16
- https://github.com/flarum/flarum-ext-lock/pull/17
- https://github.com/flarum/flarum-ext-sticky/pull/12

### DIFF
--- a/migrations/2018_09_15_041340_add_users_indicies.php
+++ b/migrations/2018_09_15_041340_add_users_indicies.php
@@ -1,0 +1,33 @@
+<?php
+
+/*
+ * This file is part of Flarum.
+ *
+ * (c) Toby Zerner <toby.zerner@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Schema\Builder;
+
+return [
+    'up' => function (Builder $schema) {
+        $schema->table('users', function (Blueprint $table) {
+            $table->index('joined_at');
+            $table->index('last_seen_at');
+            $table->index('discussion_count');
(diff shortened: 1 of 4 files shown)


franzliedke:
Code looks good. Just two questions:

1. How did you evaluate which indices we need? Did you try out our queries with EXPLAIN and evaluated possible indices? Did you go through our queries and added indices accordingly? Or are we just throwing indices at the wall to see which of them stick? :wink: Asking because they obviously have a downside as well, both in terms of memory and write performance.
If the Laracon videos were online already, I'd link to a great talk on the subject here. ;)
2. The ticket mentions indices in some of the extensions as well. Can you open those PRs, too? (I see there are some branches lying around.)

tobscure:
It's been a while since I worked them out, but I'm 99% sure I did base it on our queries and verified their effectiveness with EXPLAIN. The reason for each index is contained in https://github.com/flarum/core/issues/127#issuecomment-370190948

Sorry forgot about the extension PRs! Will do now.

Issue #1463, opened 2018-06-16
Add API for extension "installed", "updated", "before uninstall" events


tobscure:
Outcome of #1446. Extensions need to be able to do execute logic after they are installed, after they have been updated, and before they are uninstalled.

Likely should be implemented as events wrapped by a new Extender as the public-facing API. Something like:

```php
(new Extend\Extension)
    ->installed($callback)
    ->updated($callback)
    ->willUninstall($callback)
```

luceos:
This would be terrific, eg to unset a settings key after uninstall. One suggestion would be to have a `updating` as well; eg for Byobu (although no longer the case) to soft delete discussions in order to prevent complications during upgrading to open up all private discussions. Then after upgrading to re-enable those discussions.

franzliedke:
Not sure this can work as an extender. When installing an extension, the extenders of that extension are not yet loaded. We might have to hook into composer.json instead, or add more optional files (e.g. `hooks/installed.php` which returns a closure or something callable).

(Also, we don't need to register all those extenders on every request, when they only do something upon installation / deinstallation.)

tobscure:
It would be nice for API consistency if we could have these as an extender though? After all, extenders are declarative in the same way that `composer.json` is. Maybe we could make it so that the Extension extender *actually does nothing* when invoked. And, when after an extension has just been enabled, we could do something like:

```php
$extenders = require $this->getExtendPath();

foreach ($extenders as $extender) {
    if ($extender instanceof Extension) {
        $extender->runInstalledCallbacks();
    }
}
```

franzliedke:
They are a very different thing. Extenders extend Flarum at runtime, these events we are discussing are only relevant for the installation lifecycle of an extensions. I recommend to keep them apart.

There might be a good reason for the proposed extender to exist, actually, but that would allow extensions to listen for *other* extensions to be installed/enabled/etc.

tobscure:
Hmm. Upon further thought I think there's a bit more to talk about here, funnily enough related to #1464 and #1465.

Let's start with this: Extenders should be able to hook into an extension's lifecycle events. The new Formatter extender needs to be able to say "whenever this extension is enabled/disabled, clear the formatter cache". Right?

And take the embed extension as another example. It will use a Frontend extender to [define a new frontend](https://github.com/flarum/core/pull/1450#issuecomment-396021388). That Frontend extender should have the ability to say "when this extension is uninstalled, delete the frontend assets associated with it".

[...]

PR #1581, opened 2018-09-26
Extension enable/disable hooks

franzliedke:
**My shot at #1463**

- Introduce a new `LifecycleInterface` that extenders *can* implement. If they do, the interface's methods (`onEnable()`, `onDisable()`) will be called when the extension is enabled or disabled. They make it easy for extenders to bring along custom logic for these lifecycle events, instead of having to write messy event listeners.
- Change the `Formatter` extender accordingly. Nice demonstration of how much cleaner this becomes. We seem to be discovering a nice abstraction. :wink: 
- Remove the additional flushing of the formatter cache being done in the `FormatterServiceProvider`. Extensions using the `Formatter` hook were already setting up these listeners (and now they're doing it via the lifecycle hooks). In fact, this means the cache will only be flushed when actually necessary, because only extensions with a `Formatter` extender will actually cause a flush.

Additional, smaller changes:
- Small refactorings I came across while working in and around the `ExtensionManager`

- Should these lifecycle callbacks be in an additional "marker interface" or part of the existing `ExtenderInterface`? The current solution means less code for extenders that don't need them, at the price of an `instanceof`.
- Do we need other callbacks at this point? [The ticket](https://github.com/flarum/core/issues/1463) referenced other types ("before uninstall", "updating"), but so far the extenders don't seem to have a use-case for them...

**Confirmed**

### DIFF
--- a/src/Database/Console/MigrateCommand.php
+++ b/src/Database/Console/MigrateCommand.php
@@ -67,11 +67,7 @@ class MigrateCommand extends AbstractCommand
         $extensions = $this->container->make('Flarum\Extension\ExtensionManager');
         $extensions->getMigrator()->setOutput($this->output);
 
-        foreach ($extensions->getExtensions() as $name => $extension) {
-            if (! $extension->isEnabled()) {
-                continue;
-            }
-
+        foreach ($extensions->getEnabledExtensions() as $name => $extension) {
             $this->info('Migrating extension: '.$name);
 
             $extensions->migrate($extension);
--- a/src/Extend/Compat.php
+++ b/src/Extend/Compat.php
@@ -32,7 +32,7 @@ class Compat implements ExtenderInterface
         $this->callback = $callback;
     }
 
-    public function __invoke(Container $container, Extension $extension = null)
+    public function extend(Container $container, Extension $extension = null)
     {
         $container->call($this->callback);
     }
--- a/src/Extend/ExtenderInterface.php
+++ b/src/Extend/ExtenderInterface.php
@@ -16,5 +16,5 @@ use Illuminate\Contracts\Container\Container;
 
 interface ExtenderInterface
 {
-    public function __invoke(Container $container, Extension $extension = null);
+    public function extend(Container $container, Extension $extension = null);
 }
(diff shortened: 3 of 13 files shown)


franzliedke:
Oh, and there is also [`Flarum\Frontend\RecompileFrontendAssets`](https://github.com/flarum/core/blob/1f0e9b62801ccae8339ed3d66919af8ccbbfe953/src/Frontend/RecompileFrontendAssets.php), which **also** flushes some caches when extensions are enabled / disabled. I guess in the long term these should also be part of an extender - the `Frontend` one, I suppose (conditionally based on whether assets have been registered)?

tobscure:
This looks awesome! Happy with the marker interface, I think it's nicer than having to have these empty methods on the extenders that don't need them (which is the majority).

There is still no way for extensions to define their own custom logic on enable/disable/?other lifecycle events. Did you have a plan for that? Just brainstorming some use-cases:

* When **installed** (enabled for the first time), an extension needs to copy assets into the public folder
* When **updated** (this concept doesn't really exist for extensions atm), an extension needs to copy updated assets into the public folder
* When **enabled**, an extension needs to download a file from somewhere and copy it into the public folder
* When **disabled**, an extension needs to flush some cache not covered by existing extenders?
* When **uninstalled**, an extension needs to delete assets from the public folder

Not the best examples, but I think it highlights the need for us to flesh out how the other lifecycle events actually occur. I think what we probably need to do is to store the extension's version number in the database (settings table?) when it is enabled for the first time ("installed" lifecycle hook), same as how we store Flarum's version number. And then if the version of the extension code is different to that stored in the database, we need to prompt running the extension's migrations and call an "update" lifecycle hook.

> Oh, and there is also [`Flarum\Frontend\RecompileFrontendAssets`](https://github.com/flarum/core/blob/1f0e9b62801ccae8339ed3d66919af8ccbbfe953/src/Frontend/RecompileFrontendAssets.php), which **also** flushes some caches when extensions are enabled / disabled. I guess in the long term these should also be part of an extender - the `Frontend` one, I suppose (conditionally based on whether assets have been registered)?

Yep, that would make sense I think.

franzliedke:
Regarding custom logic: At this point, it would be very easy to add a `CustomLifecycle` extender (obviously with a better name) for exactly this use-case.

CDK2020:
Man, from an extension developer's standpoint, this is amazing. Thanks for your hard work @franzliedke. 

1+ for the extra events from Toby though.

PR #1582, opened 2018-09-28
Up the version

luceos:
@CDK2020 suggested to create a PR in advance so we don't forget it (this time) 😂 

Good point.


franzliedke:
I removed this from the milestone, because that contains the actual work. Also, it's smaller now. :grin: 

Still tracking this in the beta-8 "project".

franzliedke:
There we go. :hatching_chick:

PR #1586, opened 2018-10-01
Add permission for throttling

### LINKED ISSUE #1255: Throttling is bothersome to some permission groups
Although related to #929;

I wish there was an option to enable throttling only for specific permission groups. As a moderator on discuss it sometimes block my processes to handle tasks efficiently when having to wait for the grace period.

datitisev:
**Fixes #1255**

- Add permission for floodgate throttling

- Not sure if the logic should be in the controllers, or somewhere in the handler... mainly because extensions will have to implement the permission themselves as well
- Should this be a bypass permission instead?
- Should the name be `floodgate.throttling`, simple `throttling`, or something else?

**Screenshot**
image

**Confirmed**

**Required changes:**

- [ ] Related core extension PRs: flarum/english (not done yet, waiting for confirmation)


CDK2020:
I think it should be bypass. there will be a lot less people who will be allowed to bypass throttling then those who should be throttled. Not that this matters but the grammar of the function makes more sense. `$actor->can('throttle');` vs. `$actor->can('bypass');`

franzliedke:
Thanks for the PR! We'll tackle this once beta.8 is released.

migrations/2018_07_21_000100_seed_default_group_permissions.php
```diff
@@ -16,10 +16,11 @@
     // Guests can view the forum
     ['permission' => 'viewDiscussions', 'group_id' => Group::GUEST_ID],
 
-    // Members can create and reply to discussions, and view the user list
+    // Members can create and reply to discussions, view the user list, and be throttled
     ['permission' => 'startDiscussion', 'group_id' => Group::MEMBER_ID],
     ['permission' => 'discussion.reply', 'group_id' => Group::MEMBER_ID],
     ['permission' => 'viewUserList', 'group_id' => Group::MEMBER_ID],
+    ['permission' => 'floodgate.throttling', 'group_id' => Group::MEMBER_ID],
```
franzliedke: Editing old, already-released migrations, is a no-go.

src/Api/Controller/CreateDiscussionController.php
```diff
@@ -65,7 +65,7 @@ protected function data(ServerRequestInterface $request, Document $document)
         $actor = $request->getAttribute('actor');
         $ipAddress = array_get($request->getServerParams(), 'REMOTE_ADDR', '127.0.0.1');
 
-        if (! $request->getAttribute('bypassFloodgate')) {
+        if ($actor->can('floodgate.throttling') && ! $request->getAttribute('bypassFloodgate')) {
```
franzliedke: I would turn this around as well - throttling should be on for everyone (the current default) unless explicitly disabled.

That also takes us to naming... what about something like `postWithoutThrottle`? That's the thing you are *permitted* to do.

datitisev:
See #1938.

PR #1597, opened 2018-10-08
Add migration to add 'fa fa-' to group icons

datitisev:
- Add migration that adds `fa fa-` to group icons that aren't empty and don't contain `fa-`

- Do we want to update each one individually, or use the following raw SQL query?
  ```php
  $connection = $schema->getConnection();
  $prefix = $connection->getTablePrefix();
  $connection->statement("UPDATE ${prefix}groups SET icon = CONCAT('fa fa-', icon) WHERE 
  (`icon` <> '') AND (`icon` NOT LIKE '%fa-%')");
  ```
  Both work exactly the same, I just feel like the above would be a tad faster as it doesn't perform a 
query per group, only one to update them all (don't quote me on that, I have no idea on SQL 
benchmarks).
  However, it would most likely not impact anybody as I doubt most Flarum instances have over 10 groups anyway.
- Do we want a down migration to remove the `fa fa-` prefix?

**Screenshot**

After migration ran, only group ID 4 (Mod) was modified, adding the `fa fa-` to the beginning.

image

**Confirmed**

### DIFF
--- a/migrations/2018_10_08_144700_add_shim_prefix_to_group_icons.php
+++ b/migrations/2018_10_08_144700_add_shim_prefix_to_group_icons.php
@@ -1,5 +1,14 @@
 <?php
 
+/*
+ * This file is part of Flarum.
+ *
+ * (c) Toby Zerner <toby.zerner@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
 use Illuminate\Database\Schema\Builder;
 
 return [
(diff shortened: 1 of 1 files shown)


tobscure:
Looks great, thanks! I think the current method of updating each individually is fine, and has the benefit of using the query builder rather than a raw statement.

Wouldn't worry about a down migration.

Doesn't the prefix need to be `fas fa-`?

datitisev:
@tobscure Oh, I thought that the v4 shims were included (https://fontawesome.com/how-to-use/on-the-web/setup/upgrading-from-version-4#shims). I used `fa` and forgot that it's the same as `fas` for some backwards compatibility with v4.

I'll change it to `fas`. Both work, but we probably want to change it to `fas` so users don't think it's still FA 4.
image

PR #1605, opened 2018-10-19
Fixes logging in

luceos:

Seems the created_at column has no default value. This was always the case, at least that's what I can tell from a clean install and no migrations changing that default value.

```
$table->timestamp('created_at');
```

This or create a new migration?

**Confirmed**

### DIFF
--- a/src/Http/AccessToken.php
+++ b/src/Http/AccessToken.php
@@ -48,6 +48,7 @@ class AccessToken extends AbstractModel
 
         $token->token = str_random(40);
         $token->user_id = $userId;
+        $token->created_at = Carbon::now();
         $token->last_activity_at = Carbon::now();
         $token->lifetime_seconds = $lifetime;
 


franzliedke:
1. How is this related to logging in?
2. Do we use this column at all?
3. What's the difference between `last_activity_at` and Eloquent's normal `updated_at`? (As far as I can tell, we even reimplemented `touch`.)

tobscure:
2. I don't think so, but it's still good to keep a record.
3. They mean different things. `updated_at` would refer to the last time the row was updated; `last_activity_at` refers to the last time an access token was used.

luceos:
@franzliedke 

Looking at the test suite, there's a [test](https://github.com/flarum/core/blob/master/tests/Api/Controller/CreateTokenControllerTest.php) for the [CreateTokenController](https://github.com/flarum/core/blob/master/src/Api/Controller/CreateTokenController.php).

It should have created an access token otherwise the test would have failed.. I haven't looked into this, but it might be worth doing so anyway just to confirm we won't run into this again..

luceos:
@franzliedke 

Looking at the test suite, there's a [test](https://github.com/flarum/core/blob/master/tests/Api/Controller/CreateTokenControllerTest.php) for the [CreateTokenController](https://github.com/flarum/core/blob/master/src/Api/Controller/CreateTokenController.php).

It should have created an access token otherwise the test would have failed.. I haven't looked into this, but it might be worth doing so anyway just to confirm we won't run into this again..

luceos:
@franzliedke 

Looking at the test suite, there's a [test](https://github.com/flarum/core/blob/master/tests/Api/Controller/CreateTokenControllerTest.php) for the [CreateTokenController](https://github.com/flarum/core/blob/master/src/Api/Controller/CreateTokenController.php).

It should have created an access token otherwise the test would have failed.. I haven't looked into this, but it might be worth doing so anyway just to confirm we won't run into this again..

luceos:
@franzliedke 

Looking at the test suite, there's a [test](https://github.com/flarum/core/blob/master/tests/Api/Controller/CreateTokenControllerTest.php) for the [CreateTokenController](https://github.com/flarum/core/blob/master/src/Api/Controller/CreateTokenController.php).

It should have created an access token otherwise the test would have failed.. I haven't looked into this, but it might be worth doing so anyway just to confirm we won't run into this again..

luceos:
@franzliedke 

Looking at the test suite, there's a [test](https://github.com/flarum/core/blob/master/tests/Api/Controller/CreateTokenControllerTest.php) for the [CreateTokenController](https://github.com/flarum/core/blob/master/src/Api/Controller/CreateTokenController.php).

It should have created an access token otherwise the test would have failed.. I haven't looked into this, but it might be worth doing so anyway just to confirm we won't run into this again..

luceos:
@franzliedke 

Looking at the test suite, there's a [test](https://github.com/flarum/core/blob/master/tests/Api/Controller/CreateTokenControllerTest.php) for the [CreateTokenController](https://github.com/flarum/core/blob/master/src/Api/Controller/CreateTokenController.php).

It should have created an access token otherwise the test would have failed.. I haven't looked into this, but it might be worth doing so anyway just to confirm we won't run into this again..

franzliedke:
GitHub made me hear that loud and clear. :grinning: 

Another candidate for a regression test, I guess. Will take a look, thanks!

luceos:
@franzliedke 

Looking at the test suite, there's a [test](https://github.com/flarum/core/blob/master/tests/Api/Controller/CreateTokenControllerTest.php) for the [CreateTokenController](https://github.com/flarum/core/blob/master/src/Api/Controller/CreateTokenController.php).

It should have created an access token otherwise the test would have failed.. I haven't looked into this, but it might be worth doing so anyway just to confirm we won't run into this again..

luceos:
OMG, woops, I wasn't aware about GitHub's issue, so tried pushing that comment several times and then I looked up their status page 😊

PR #1608, opened 2018-10-20
New extender for adding variables to HtmlDocument payload

### LINKED ISSUE #1602: PHP Extender: Rendering
We need another extender that allows setting variables when rendering a HTML frontend (e.g. the forum). This will replace code like this in the statistics extension:

~~~php
public function addStatisticsData(Rendering $event)
{
    if ($event->isAdmin()) {
        $event->view->setVariable('statistics', array_merge(
            $this->getStatistics(),
            ['timezoneOffset' => $this->getUserTimezone()->getOffset(new DateTime)]
        ));
    }
}
~~~

Follow-up to #851.

franzliedke:
**Fixes #1602**

Add a `document()` method to the `Frontend` extender as proposed in #1602.

The method accepts a closure as well as a class name, to allow for more complex logic to be encapsulated in a class.

**Confirmed**

**Required changes:**

- [x] Fix database columns in flarum/statistics
- [ ] Adopt extender in flarum/statistics

### DIFF
--- a/src/Extend/Frontend.php
+++ b/src/Extend/Frontend.php
@@ -14,6 +14,7 @@ namespace Flarum\Extend;
 use Flarum\Extension\Extension;
 use Flarum\Frontend\Asset\ExtensionAssets;
 use Flarum\Frontend\CompilerFactory;
+use Flarum\Frontend\HtmlDocumentFactory;
 use Flarum\Http\RouteHandlerFactory;
 use Illuminate\Contracts\Container\Container;
 
@@ -24,6 +25,7 @@ class Frontend implements ExtenderInterface
     protected $css = [];
     protected $js;
     protected $routes = [];
+    protected $content = [];
 
     public function __construct($frontend)
     {
@@ -51,10 +53,22 @@ class Frontend implements ExtenderInterface
         return $this;
     }
 
+    /**
+     * @param callable|string $callback
+     * @return $this
+     */
+    public function content($callback)
+    {
+        $this->content[] = $callback;
+
+        return $this;
+    }
+
     public function extend(Container $container, Extension $extension = null)
     {
         $this->registerAssets($container, $this->getModuleName($extension));
         $this->registerRoutes($container);
+        $this->registerContent($container);
     }
 
     private function registerAssets(Container $container, string $moduleName)
@@ -90,6 +104,26 @@ class Frontend implements ExtenderInterface
                 $factory->toFrontend($this->frontend, $route['content'])
             );
         }
+    }
+
+    private function registerContent(Container $container)
+    {
+        if (empty($this->content)) {
+            return;
+        }
+
+        $container->resolving(
+            "flarum.$this->frontend.frontend",
+            function (HtmlDocumentFactory $view, Container $container) {
+                foreach ($this->content as $content) {
+                    if (is_string($content)) {
+                        $content = $container->make($content);
+                    }
+
+                    $view->add($content);
+                }
+            }
+        );
     }
 
     private function getModuleName(?Extension $extension): string
--- a/src/Admin/Content/AdminPayload.php
+++ b/src/Admin/Content/AdminPayload.php
@@ -52,7 +52,7 @@ class AdminPayload implements ContentInterface
         $this->events = $events;
     }
 
-    public function populate(HtmlDocument $document, Request $request)
+    public function __invoke(HtmlDocument $document, Request $request)
     {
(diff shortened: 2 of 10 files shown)


datitisev:
@franzliedke I think that @tobscure had in mind (and mentioned a few times) is that Flarum flattens the extend.php array, so you can use the extenders in multiple files and require them (however that's done) in extend.php

image

src/Extend/Frontend.php
```diff
@@ -51,10 +54,22 @@ public function route($path, $name, $content = null)
         return $this;
     }
 
+    /**
+     * @param callable|string $callback
+     * @return $this
+     */
+    public function document($callback)
+    {
+        $this->documentCallback = $callback;
```
tobscure: I think we should allow multiple document callbacks to be registered on a single extender. There is loads of stuff that can be done to HtmlDocument so people might want to organise it into separate callbacks/classes:

```php
    ->document(AddFooter::class)
    ->document(AddSeoStuff:class)
```

franzliedke:
@datitisev You mean having a separate file with a rather long extender callback instead of my proposed solution with an extra class?

Not sure whether that works here, as we're just calling one more method on the `Frontend` extender, which is very short otherwise.

src/Extend/Frontend.php
tobscure: I think we should allow multiple document callbacks to be registered on a single extender. There is loads of stuff that can be done to HtmlDocument so people might want to organise it into separate callbacks/classes:

```php
    ->document(AddFooter::class)
    ->document(AddSeoStuff:class)
```

franzliedke:
> How would it work with the Listener, anyway? I don't see a specific method being called or anything... though that may be because of my nonexistent Laravel knowledge.

The class would have an `__invoke` method, which means it can be treated like a closure or any other callable.

src/Extend/Frontend.php
```diff
@@ -92,6 +107,26 @@ private function registerRoutes(Container $container)
...
+        $container->resolving(
+            "flarum.$this->frontend.frontend",
+            function (HtmlDocumentFactory $view, Container $container) {
+                if ($this->documentCallback instanceof Closure) {
+                    $callback = $this->documentCallback;
+                } else {
+                    $callback = $container->make($this->documentCallback);
+                }
+
+                $view->addCallback($callback);
```
tobscure: Is it necessary to add a new collection of "callbacks" to the HtmlDocumentFactory? That's what "content" is for. We could do this instead:

```php
$content = new class implements ContentInterface {
    public $callback;
    public function populate(HtmlDocument $view, Request $request)
    {
        $this->callback($view, $request);
    }
});

$content->callback = $callback;

$view->add($content);
```

src/Extend/Frontend.php
franzliedke: In one extension?

tobscure: Yeah

src/Extend/Frontend.php
franzliedke: Hmm, good point. What about getting rid of `ContentInterface` and making those classes invokable, too?

franzliedke: Or we can keep the interface, and the existing listener class would implement that interface instead of being invokable.

src/Extend/Frontend.php
tobscure: Hmm I wonder if `content` might be a better name for the method for consistency with the internal API. And you could either pass a callback, or a class name which implements `ContentInterface`

src/Extend/Frontend.php
tobscure: I like your second suggestion, as per my comment in the other thread :)

src/Extend/Frontend.php
franzliedke: I took care of the multiple calls, will do the rest tomorrow. :zzz:

src/Extend/Frontend.php
franzliedke: I've done this now. I changed `ContentInterface` to use `__invoke` instead of `populate`, so that we can just drop in closures instead (we weren't type-hinting anyway).

franzliedke:
All done from my side.

franzliedke:
flarum/statistics can now get by without implementing the `ContentInterface` (it just defines `__invoke`, without any `implements ContentInterface`). This would mean we can consider `ContentInterface` private API for now, and I kind of like that.

PR #1613, opened 2018-10-24
removed form title from email subject

ozzzzam:
From this update, users will have full freedom to customize the email subject as they want.

**Fixes #0000**

**Screenshot**

**Confirmed**

**Required changes:**

### DIFF
--- a/src/Api/Controller/SendConfirmationEmailController.php
+++ b/src/Api/Controller/SendConfirmationEmailController.php
@@ -89,7 +89,7 @@ class SendConfirmationEmailController implements RequestHandlerInterface
 
         $this->mailer->raw($body, function (Message $message) use ($actor, $data) {
             $message->to($actor->email);
-            $message->subject('['.$data['{forum}'].'] '.$this->translator->trans('core.email.activate_account.subject'));
+            $message->subject($this->translator->trans('core.email.activate_account.subject'));
         });
 
         return new EmptyResponse;


ozzzzam:
Better if possible to get the discussion sooner :(

franzliedke:
@ozzzzam Sorry, I don't understand. What do you mean?

tobscure:
The `[Forum Title]` part of the subject was intended to be a **prefix**, i.e. not part of each individual subject translation. From that point of view I would only be in favour of adding a new translation for an email subject "wrapper", something like this:

```
core.email.subject_template: "[{title}] {subject}"
core.email.activate_account.subject: "Activate Your Account"
```

But I suppose there is also the use-case of someone wanting to customise the whole individual subject? Like if instead of "[Forum Title] Activate Your Account" you wanted "Activate Your Forum Title Account". Should we support this use-case?

luceos:
> But I suppose there is also the use-case of someone wanting to customise the whole individual subject? Like if instead of "[Forum Title] Activate Your Account" you wanted "Activate Your Forum Title Account". Should we support this use-case?

I say yes. If you consider flagrow/linguist allows you to easily update one single translation key to your liking it enables you to update email subjects and even body quite easily. Let's not enforce what we think makes sense in the context.

I personally don't like these prefixes anyway, because you already have a sender. In addition the body often has a signature. The only thing the prefix achieves is reduce readability on devices that limit how much you can see from the subject..

ozzzzam:
Any update on this

franzliedke:
Patience, please. We are on the verge of releasing beta.8, that is more important than anything else. After the release, we will go through all outstanding PRs and take care of them.

tobscure:
@franzliedke to clarify, do we want:

```yaml
core.email.subject_template: "[{title}] {subject}"
core.email.activate_account.subject: "Activate Your Account"
```

Or keep it simpler and have the prefix in each subject:

```yaml
core.email.activate_account.subject: "[{title}] Activate Your Account"
```

franzliedke:
You seemed to prefer the first option, because it was more in line with the previous implementation (a "global" prefix).

Personally, I'd go with the second, as it is fully flexible.

luceos:
I'm with @tobscure on this one. I'd rather have one global mail prefix than one that keeps popping up on every locale string..

Alternatively we might even consider making the prefix a setting configurable in admin, so including the brackets.

franzliedke:
My absolute favorite would be @luceos' original suggestion of getting rid of the prefix and relying on the sender. We can make that more easily configurable, I guess?

tobscure:
Is there any chance we'll want the sender to be different in certain instances though? e.g. in an email notification for a new PM, should the sender be the name of the user who sent the PM?

luceos:
Sending email on behalf of someone else has its downsides, because there are two options:

- a mismatching sender name vs sending email address; I always find this shady practice
- sender name and email address are from the actual user/sender; this won't work because most mail apps will mark it as spam

The best way for a PM notification email is to play the handling third party, eg "Hey tobscure, luceos sent you a PM". This allows the same sender to be used.

franzliedke:
For the record: I think @luceos' suggestion (of sticking to the forum name as sender - we could even add an additional setting) is both sensible and simple enough to implement.

franzliedke:
I guess that means once we have an issue that details the desired solution, we can close this PR? :confused:

franzliedke:
As discussed in chat, we're already using the forum name as sender name everywhere:
https://github.com/flarum/core/blob/8dd3bd420b0a919b691407cf61cc7f9d52f4b7c9/src/Mail/MailServiceProvider.php#L59

So accepting this just removes duplication.

If anybody is actually interested in making the sender name configurable with something else than the forum name, feel free to open a ticket. :grin:

PR #1617, opened 2018-10-24
Split up the installer logic

### LINKED ISSUE #1587: Can't install Flarum, if Markdown or BBCode is enabled
## Bug Report

**Current Behavior**
I want to install Flarum, but when it's time for the Markdown (or BBCode) extension to migrate it throws an error.

**Steps to Reproduce**
1. Clone the latest Flarum skleton

2. Edit the following files

`composer.json`
```json
{
    "name": "flarum/flarum",
    "description": "Delightfully simple forum software.",
    "type": "project",
    "keywords": ["forum", "discussion"],
    "homepage": "https://flarum.org/",
    "license": "MIT",
    "authors": [
        {
            "name": "Toby Zerner",
            "email": "toby.zerner@gmail.com"
        },
        {
            "name": "Franz Liedke",
            "email": "franz@develophp.org"
        }
    ],
    "support": {
        "issues": "https://github.com/flarum/core/issues",
        "source": "https://github.com/flarum/flarum",
        "docs": "https://flarum.org/docs/"
    },
    "require": {
        "flarum/core": "dev-master",
        "flarum/flarum-ext-english": "dev-master",
        "flarum/flarum-ext-likes": "dev-master",
        "flarum/flarum-ext-lock": "dev-master",
        "flarum/flarum-ext-markdown": "dev-master",
        "flarum/flarum-ext-mentions": "dev-master",
        "flarum/flarum-ext-sticky": "dev-master",
        "flarum/flarum-ext-subscriptions": "dev-master",
        "flarum/flarum-ext-suspend": "dev-master",
        "flarum/flarum-ext-tags": "dev-master"
    },
    "config": {
        "preferred-install": "dist",
        "sort-packages": true
    },
    "minimum-stability": "dev",
    "prefer-stable": true
}
```

`install.json`
```json
{
    "debug": "true",
    "baseUrl": "http://flarumsecond.localhost",
    "databaseConfiguration": {
        "driver": "mysql",
        "host": "localhost",
        "port": 3306,
        "database": "flarum_second",
        "username": "flarum",
        "password": "password",
        "prefix": ""
    },
    "adminUser": {
        "username": "admin",
        "password": "password",
        "password_confirmation": "password",
        "email": "admim@example.com"
    },
    "settings": {
        
    }
}
```

3. Run the install command
`./flarum install -f install.json`

4. After the Flarum migrations, there will be an error when migrating the Markdown extension
```
Enabling extension: flarum-markdown

In Container.php line 752:

  Class cache.store does not exist
```
See the [full log](https://gist.github.com/LukBukkit/4886345eca13249ce64e3aaac3b87d61).

A similar problem also happens when installing on the web, but there isn't a nice error message shown, instead of the message mentioned above there's another error:
```
Error booting Flarum: Too few arguments to function Illuminate\Support\Manager::createDriver(), 0 passed in D:\lukas\Coding\PhpstormProjects\flarum\flarum\vendor\illuminate\support\Manager.php on line 88 and exactly 1 expected
```
Read more about this on [#1169](https://github.com/flarum/core/issues/1169#issuecomment-425908605).

**Expected Behavior**
It should install without any error.

**Environment**
- Flarum version: 655aa9cea91dab571b67d917f37cc95ce3615285
- Webserver: apache and cli
- Hosting environment: Windows 10
- PHP version: 7.1.11
- Browser: Firefox Dev. Edition 63.0b10 (64-Bit)

```
$ ./flarum info
Flarum core 0.1.0-beta.7 (737d33826e9d4caf1d07fa8ee5509d1ff06066d6)
PHP version: 7.1.11
Loaded extensions: Core, bcmath, calendar, ctype, date, filter, hash, iconv, json, mcrypt, SPL, pcre, readline, Reflection, session, standard, mysqlnd, tokenizer, zip, zlib, libxml, dom, PDO, bz2, SimpleXML, xml, wddx, xmlreader, xmlwriter, openssl, curl, fileinfo, gd, gettext, mbstring, exif, mysqli, pdo_mysql, pdo_sqlite, soap, sockets, sqlite3, tidy, Phar, ftp
+----------------------+------------+--------+
| Flarum Extensions    |            |        |
+----------------------+------------+--------+
| ID                   | Version    | Commit |
+----------------------+------------+--------+
| flarum-english       | dev-master |        |
| flarum-likes         | dev-master |        |
| flarum-lock          | dev-master |        |
| flarum-mentions      | dev-master |        |
| flarum-sticky        | dev-master |        |
| flarum-subscriptions | dev-master |        |
| flarum-suspend       | dev-master |        |
| flarum-tags          | dev-master |        |
+----------------------+------------+--------+
Base URL: http://flarumsecond.localhost
Installation path: P:\Programme\XAMPP\htdocs\flarum-test2
Debug mode: ON

```

**Possible Solution**
I'll look into this tomorrow.

**Additional Context**
The installation works fine with the latest beta build (0.1-beta7) and when using the latest repository (development) version without the `Markdown` and `BBCode` extensions. See this [installation log](https://gist.github.com/LukBukkit/1ba3cab0ac819c83d12b4d685dd47a9e), where I excluded those extensions from the `composer.json`.

franzliedke:
**Fixes #1587**

This is probably the most complicated way I could find to fix #1587.

Jokes aside, this was done with a few goals in mind:
- Reduce coupling between the installer and the rest of Flarum's 
  "Application", which we are building during installation.
- Move the installer logic to several smaller classes, which can then
  be used by the web frontend and the console task, instead of the
  former hacking its way into the latter to be "DRY".
- Separate installer infrastructure (the "pipeline", with the ability
  to revert steps upon failure) from the actual steps being taken.

The problem was conceptual, and would certainly re-occur in a similar
fashion if we wouldn't tackle it at its roots.

It is fixed now, because we no longer use the ExtensionManager for
enabling extensions, but instead duplicate some of its logic. That is 
fine because we don't want to do everything it does, e.g. omit 
extenders' lifecycle hooks (which depend on the Application instance
being complete).

> for each desired change, make the change easy (warning: this may be 
> hard), then make the easy change

- Kent Beck, https://twitter.com/kentbeck/status/250733358307500032

Bugs and conceptual problems

**Confirmed**


tobscure:
Looking good, will review over the weekend!

luceos:
Anything specific you want us to look at @franzliedke ?

franzliedke:
High-level concepts mostly.

src/Install/Console/InstallCommand.php
```diff
@@ -121,232 +100,82 @@ protected function fire()
...
-            'flarum-pusher',
-        ];
-
-        foreach ($extensions->getExtensions() as $name => $extension) {
-            if (in_array($name, $disabled)) {
-                continue;
-            }
-
-            $this->info('Enabling extension: '.$name);
+        if (! filter_var($admin['email'], FILTER_VALIDATE_EMAIL)) {
```
luceos: You are resolving the Validation Factory in the construct, is there any reason not to use the (most likely better) native `email` validation?

src/Install/Console/InstallCommand.php
```diff
@@ -121,232 +100,82 @@ protected function fire()
...
+            throw new Exception('You must enter a valid email.');
+        }
 
-            $extensions->enable($name);
+        if (! $admin['username'] || preg_match('/[^a-z0-9_-]/i',
```
luceos: As we're using the Validation Factory, wouldn't it be better to have a Custom validation rule for usernames that adheres to our (regular expression) requirements?

src/Install/Console/InstallCommand.php
franzliedke: Fun fact: this is what Laravel's validation layer does for emails. :wink:

src/Install/Console/InstallCommand.php
```diff
@@ -101,7 +80,7 @@ protected function fire()
     {
         $this->init();
 
-        $prerequisites = $this->getPrerequisites();
+        $prerequisites = $this->installation->prerequisites();
         $prerequisites->check();
```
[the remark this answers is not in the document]
franzliedke: Good idea, I've taken care of this.

src/Install/Controller/InstallController.php
```diff
@@ -51,54 +50,65 @@ public function handle(Request $request): ResponseInterface
...
+                'welcome_title' => 'Welcome to '.array_get($input, 'forumTitle'),
+            ])
+            ->build();
 
         try {
-            $this->command->run($input, $output);
+            $pipeline->run();
         } catch (Exception $e) {
-            return new HtmlResponse($e->getMessage(), 500);
+            return new Response\HtmlResponse($e->getMessage(), 500);
```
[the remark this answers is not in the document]
franzliedke: True, fixed.

tobscure:
Ref flarum/docs#43

src/Install/Console/InstallCommand.php
```diff
@@ -121,232 +100,82 @@ protected function fire()
...
-        if ($this->dataSource === null) {
-            if ($this->input->getOption('defaults')) {
-                $this->dataSource = new DefaultsDataProvider();
-            } elseif ($this->input->getOption('file')) {
-                $this->dataSource = new FileDataProvider($this->input);
-            } else {
-                $this->dataSource = new UserDataProvider($this->input, $this->output, $this->getHelperSet()->get('question'));
-            }
+        if ($this->input->getOption('defaults')) {
+            $this->dataSource = new DefaultsDataProvider();
```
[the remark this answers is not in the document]
franzliedke: Yep, done.

src/Install/Steps/EnableBundledExtensions.php
```diff
@@ -0,0 +1,112 @@
...
+                $extension->setVersion(Arr::get($package, 'version'));
+
+                return $extension;
+            })->filter(function (Extension $extension) {
+                return ! in_array($extension->getId(), self::DISABLED_EXTENSIONS);
+            })->sortBy(function (Extension $extension) {
+                return $extension->composerJsonAttribute('extra.flarum-extension.title');
+            })->mapWithKeys(function (Extension $extension) {
+                return [$extension->getId() => $extension];
+            });
```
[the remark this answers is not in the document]
franzliedke: I think I've taken care of this concern for now by switching to a whitelist of extensions which are automatically enabled, as [suggested here](https://github.com/flarum/core/pull/1631#issuecomment-437744083).

For the mid-term future, I'd like to move to enabling extensions not in the installer, but in a first step after it's done. That will mean the installer does not have to deal with extensions at all, until it has a fully installed "app" that it can ask to do those things. This is why [I shot this PR down](https://github.com/flarum/core/pull/1724).

franzliedke:
Okay, I do now consider this complete (enough).

Additional changes since the last reviews:
- Instead of a blacklist, we now have a whitelist that controls which extensions are enabled during installation instead (see [this discussion](https://github.com/flarum/core/pull/1631#issuecomment-437744083)).
- I removed `php flarum install --defaults` - this mostly existed for easy setup in our old Vagrant-based development VM. That VM is gone, dev setup can easily be done with the `--file` option as well.
- Reduced duplication
  - Default settings are managed in one place.
  - Domain invariants (e.g. password length) are checked only once in a few new classes.
  - Only UI-specific validations (such as confirming a password by entering it twice) is implemented twice, depending on whether the UI requires it.
- Some (not all yet) of the validation messages appear immediately when installing on the config. Previously, you would first enter all the values before being warned that there are problems, which was annoying.
- Building and storing config file was combined into one step.
- We now check for the correct MariaDB version when not using MySQL (see flarum/docs#43).
- When installation fails, published assets (font files) are now unpublished again.

This should take care of all of @tobscure's comments, and then some. :grinning: 

Last but not least, I fixed two down migrations and the `Migrator`'s `reset()` method - however, we still cannot fully migrate down due to issues with foreign keys and indices, I will tackle that separately.

Next steps (in separate PRs for later versions):
- Test suite improvements (already done, but this PR needs to be merged first)
- Better and more validation
- Don't enable extensions during installation, but in a separate step afterwards (to properly resolve #1587 and #1631) - in combination with a quick and pretty way to enable extensions via the API as a second step in the GUI installer
- Separate installer from updater (no need to run all migrations to setup a new Flarum instance, should be quicker and more stable) - *to be discussed*

franzliedke:
> TODO from #1674: Ensure that all migrations are fully rolled back when there is a failure.

I wasn't able to take care of this - our core migrations currently cannot be rolled back. I would prefer to fix this as part of tackling the "Separate installer from updater" suggestion above.

src/Extension/Extension.php
```diff
@@ -307,6 +313,25 @@ public function hasAssets()
...
     }
 
+    public function copyAssetsTo(FilesystemInterface $target)
+    {
+        if (! $this->hasAssets()) {
+            return;
+        }
+
+        $mount = new MountManager([
+            'source' => $source = new Filesystem(new Local($this->getPath().'/assets')),
```
luceos: Wouldn't it make more sense to have a global Filesystem that gets the extension subdirectory in the code below. That would reduce the need of object instantiation. 50 extensions migrating assets otherwise generate a pretty large memory footprint due to these instantiations..

src/Install/AdminUser.php
```diff
@@ -0,0 +1,58 @@
...
+    {
+        return $this->username;
+    }
+
+    public function getAttributes(): array
+    {
+        return [
+            'username' => $this->username,
+            'email' => $this->email,
+            'password' => (new BcryptHasher)->make($this->password),
```
luceos: Is it wise to hardcode the password hasher like this instead of binding it in the container? Some use cases:

- incompatible OS for bcrypt
- newer crypt lib available in Laravel but not supported here
- inability to exchange the hasher from ioc during installation, not sure whether this has any value though 🤷‍♂️

src/Install/DatabaseConfig.php
```diff
@@ -0,0 +1,98 @@
...
+        $this->port = $port;
+        $this->database = $database;
+        $this->username = $username;
+        $this->password = $password;
+        $this->prefix = $prefix;
+
+        $this->validate();
+    }
+
+    public function getConfig(): array
```
luceos: Just curious whether making this an `Arrayable` or `__invoke` kind of class would be cleaner?

Arrayable makes `toArray` obligatory: https://laravel.com/api/5.7/Illuminate/Contracts/Support/Arrayable.html

src/Install/Installation.php
```diff
@@ -0,0 +1,164 @@
...
+ *
+ * (c) Toby Zerner <toby.zerner@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Flarum\Install;
+
+class Installation
```
luceos: Props on this class ;)

src/Install/Prerequisite/Composite.php
```diff
@@ -25,21 +27,14 @@ public function __construct(PrerequisiteInterface $first)
...
     {
         return array_reduce(
             $this->prerequisites,
-            function ($previous, PrerequisiteInterface $prerequisite) {
-                return $prerequisite->check() && $previous;
+            function (Collection $errors, PrerequisiteInterface $condition) {
+                return $errors->concat($condition->problems());
             },
-            true
+            collect()
```
luceos: What is the benefit here of using the function `collect()` over direct class instantiation? The class is imported in this code already anyway?

src/Install/Prerequisite/PhpVersion.php
```diff
@@ -20,13 +22,15 @@ public function __construct($minVersion)
...
     }
 
-    public function check()
+    public function problems(): Collection
     {
         if (version_compare(PHP_VERSION, $this->minVersion, '<')) {
-            $this->errors[] = [
+            return collect()->push([
                 'message' => "PHP $this->minVersion is required.",
                 'detail' => 'You are running version '.PHP_VERSION.'. Talk to your hosting provider about upgrading to the latest PHP version.',
```
luceos: Most commonly the suggestion is to talk to the "system administrator", referring to a hosting provider might be too opinionated?

tests/Test/Concerns/CreatesForum.php
```diff
@@ -108,23 +105,10 @@ protected function teardownApplication()
...
-                'charset'   => 'utf8mb4',
-                'collation' => 'utf8mb4_unicode_ci',
-                'prefix'    => $dbConfig['prefix'],
-                'port'      => $dbConfig['port'],
-                'strict'    => false
-            ],
-            'url'      => $this->configuration->getBaseUrl(),
+            'debug'    => true,
+            'database' => $this->getDatabaseConfiguration()->getConfig(),
+            'url'      => 'http://flarum.local',
```
luceos: This value `url` can't be hardcoded. We should be able to override it from env at least, so that we properly test in CI. This is especially true when running browser tests (in the future). The same might be true for debug, eg by checking what information is returned from the API when running debug on/off.

src/Install/DatabaseConfig.php
franzliedke: If this were a PHP core interface like [`JsonSerializable`](http://php.net/manual/en/class.jsonserializable.php), maybe this would even be called implicitly in some cases. It isn't, though. :confused: 

Still, I'll follow your suggestion - at least this way I don't have to come up with my own semi-good method name. :laughing:

src/Install/Prerequisite/Composite.php
franzliedke: None, just laziness. Do you have a preference?

src/Install/Prerequisite/PhpVersion.php
franzliedke: Thanks, good suggestion.

tests/Test/Concerns/CreatesForum.php
franzliedke: True, but soon irrelevant, when I follow up with my testing PR. So I'll ignore this here if you don't mind. ;)

migrations/2018_09_22_004100_change_registration_tokens_columns.php
```diff
@@ -24,7 +24,7 @@
     },
 
     'down' => function (Builder $schema) {
-        $schema->table('auth_tokens', function (Blueprint $table) {
+        $schema->table('registration_tokens', function (Blueprint $table) {
```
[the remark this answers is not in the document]
franzliedke: It was a mistake. I fixed this when I tried to make the TODO from #1674 work in this branch.

Since that turned out to be harder than expected, I postponed the TODO, but it still makes sense to get the fix in. Do you want me to extract this to `master`?

src/Extension/Extension.php
franzliedke: Do we support enabling multiple extensions at once? Because if we don't, the only place that would be affected by this is the installer, which a) isn't particularly performance-critical and b) will soon lose this functionality (see my list of next steps). So I'd take the lazy route here...

src/Install/AdminUser.php
franzliedke: I want to keep the installer as decoupled from runtime Flarum as possible. This makes things much easier. When the core changes its default hashing mechanism, the installer will have to change, too.

IMO, we should keep extensions out of the installer. That just adds too much complexity to their lifecycle. They can still change the hashing mechanism at runtime, if they must. (They would have to be able to deal with "legacy" passwords generated by a Flarum without this extension anyway.)

franzliedke:
I adapted your suggestions, or replied to your comments.

migrations/2018_09_22_004100_change_registration_tokens_columns.php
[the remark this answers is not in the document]
luceos: Nope, I can imagine we merge this soon anyway. Just curious about why it got in in the first place.

src/Install/Prerequisite/Composite.php
luceos: My preference is to use class instantiation over function. But then again, maybe we need to have a stricter guideline for this.

luceos:
Looked over it, feel free to resolve where needed. And dismiss the review/merge once done with the last few modifications.

tobscure:
Bravo @franzliedke! 👏

PR #1627, opened 2018-11-06
Fix dropdown menu hidden by PermissionsPage header when going up

datitisev:
- Remove `overflow: auto` from `.PermissionsPage-permissions`
  - Scroll isn't even managed by this container, the whole page content scrolls, so not sure why it was here in the first place

**Screenshot**
Before:
image

After:
image

**Confirmed**

### DIFF
--- a/less/admin/PermissionsPage.less
+++ b/less/admin/PermissionsPage.less
@@ -32,7 +32,6 @@
 
 .PermissionsPage-permissions {
   padding: 20px 0 200px;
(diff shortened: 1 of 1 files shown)


datitisev:
@tobscure Oh, no, I hadn't considered that. I can confirm that's a problem. `overflow-x: auto;` fixes that.

franzliedke:
I fixed that in master, thanks.

PR #1630, opened 2018-11-09
add security address in issue template as well

luceos:
as per suggestion by @CDK2020, let's also add an important statement in the issue template


franzliedke:
Thank you! IMO, we should even add a new category "Security Issue" on [this page](https://github.com/flarum/core/issues/new/choose).

PR #1631, opened 2018-11-11
Boot new application instance before enabling extensions on installation

### LINKED ISSUE #1587: Can't install Flarum, if Markdown or BBCode is enabled

tobscure:
**Fixes #1587**

#1617 is a big rewrite of the installer and will take a while longer to perfect before it is release-ready. Thus I think we should implement a stopgap solution in the current installer to fix #1587 so that we can get beta 8 out the door sooner, and delay the rewrite until beta 9.

As mentioned in https://github.com/flarum/core/pull/1617#discussion_r232439491, I'm a little bit concerned about not enabling extensions via the ExtensionManager and therefore ignoring lifecycle events. What if one of the default extensions actually needs to make use of a lifecycle event in the future? (eg. managing some files, building a cache, something like that.)

Thus, this PR simply boots a new (installed) application instance before enabling bundled extensions, so that all application services are available to extension lifecycle hooks.

Is this an acceptable solution for the purposes of beta 8. Further discussion about the installer overhaul can continue in https://github.com/flarum/core/pull/1617.

**Confirmed**

### DIFF
--- a/src/Install/Console/InstallCommand.php
+++ b/src/Install/Console/InstallCommand.php
@@ -16,6 +16,8 @@ use Flarum\Console\AbstractCommand;
 use Flarum\Database\DatabaseMigrationRepository;
 use Flarum\Database\Migrator;
 use Flarum\Extension\ExtensionManager;
+use Flarum\Foundation\Application as FlarumApplication;
+use Flarum\Foundation\Site;
 use Flarum\Group\Group;
 use Flarum\Install\Prerequisite\PrerequisiteInterface;
 use Flarum\Settings\DatabaseSettingsRepository;
@@ -186,9 +188,20 @@ class InstallCommand extends AbstractCommand
 
             $this->createAdminUser();
 
-            $this->enableBundledExtensions();
-
             $this->publishAssets();
+
+            // Now that the installation of core is complete, boot up a new
+            // application instance before enabling extensions so that all of
+            // the application services are available.
+            Site::fromPaths([
+                'base' => $this->application->basePath(),
+                'public' => $this->application->publicPath(),
+                'storage' => $this->application->storagePath(),
+            ])->bootApp();
+
+            $this->application = FlarumApplication::getInstance();
+
+            $this->enableBundledExtensions();
         } catch (Exception $e) {
             @unlink($this->getConfigFile());
 


tobscure:
I guess we can break the problem down a bit more.

For the first-party extensions we bundle with Flarum, I agree we're unlikely to ever need to do anything in the lifecycle hooks that would be meaningful on first installation.

However, I wouldn't be confident saying this for third-party extensions. As a concrete example, what about an advanced statistics extension which builds a cache of forum data analytics whenever it is enabled.

Will third-party extensions ever be enabled on installation? Well, the current logic will actually enable all extensions *except for* a select few that are blacklisted. So if you install any extensions before installing Flarum, they would get enabled without running their lifecycle hooks. Perhaps we should change it to be a whitelist instead.

Also, a possibility in the future (when we do composer web interface stuff) is to have a step in the GUI installer where you choose which extensions you would like to install and enable. In this case I think we would likely want to do what you've suggested anyway – you would do the core installation first, and *then* set up your extensions in a subsequent request.

I agree that's the cleanest way forward conceptually, maybe something we can think about in #1617.

PR #1633, opened 2018-11-11
Improve logging

### LINKED ISSUE #991: Flarum API doesn't log errors
rights are okay though:

``` php
chmod 775 /www/icra/icrafl
chmod -R 775 /www/flarum/assets /www/flarum/storage
chgrp www-data /www/icra/icrafl
chgrp -R www-data /www/flarum/assets /www/flarum/storage
```

I also get 500 internal error pages
debug is set to true

### LINKED ISSUE #1619: Ignore RouteNotFound exceptions
One of the reasons for log rotation (#1598) was the size of the flarum.log file on an inactive forum. Now that log rotation is in place, devops can easily create archives and/or delete older files to keep the disk in check.

The underlying problem, however, hasn't been resolved. `RouteNotFound` exceptions are still piling up in the log file. 

- This makes providing support pretty rough. You can ask for the log file and get a multi MB file back with (maybe) one non-RouteNotFound exception tucked away somewhere deep inside lines and lines of traces.
- A decently configured (pretty much all default set ups) also have website access logs. These provide better information when hitting a 404 anyhow, because they log the actual path.
- Crawlers (also those with bad intent) will trigger the exception and cause a flood of unnecessary log lines.

I hope we can tackle this soon, the native Laravel Handler has support for ignoring exceptions when handling and rendering. See 

https://github.com/laravel/framework/blob/c30a784f56e5f0c4c54ffbb601280bc278227a9b/src/Illuminate/Foundation/Exceptions/Handler.php#L60

tobscure:
**Fixes #1619**
**Fixes #991**

* Catch Throwables so that we handle and log internal PHP errors too
* Stop logging errors that use a custom view (eg. 404 - RouteNotFound)
* Log errors that occur in the API stack

See commit messages for more details.

This is not a definitive implementation for optimal error handling and logging. Ideally we should have a more advanced error handling solution like the [one referenced](https://github.com/laravel/framework/blob/c30a784f56e5f0c4c54ffbb601280bc278227a9b/src/Illuminate/Foundation/Exceptions/Handler.php#L60) in #1619. But I thought it would be good to implement a least-resistance stopgap in the meantime.

Is this acceptable for the purposes of beta 8. 

**Confirmed**

### DIFF
--- a/src/Api/ErrorHandler.php
+++ b/src/Api/ErrorHandler.php
@@ -12,6 +12,7 @@
 namespace Flarum\Api;
 
 use Exception;
+use Throwable;
 use Tobscure\JsonApi\Document;
 use Tobscure\JsonApi\ErrorHandler as JsonApiErrorHandler;
 
@@ -34,8 +35,12 @@ class ErrorHandler
      * @param Exception $e
      * @return JsonApiResponse
      */
-    public function handle(Exception $e)
+    public function handle(Throwable $e)
     {
+        if (! $e instanceof Exception) {
+            $e = new Exception($e->getMessage(), $e->getCode(), $e);
+        }
+
         $response = $this->errorHandler->handle($e);
 
         $document = new Document;
--- a/src/Api/ExceptionHandler/FallbackExceptionHandler.php
+++ b/src/Api/ExceptionHandler/FallbackExceptionHandler.php
@@ -1,0 +1,79 @@
+<?php
+
+/*
+ * This file is part of Flarum.
+ *
+ * (c) Toby Zerner <toby.zerner@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Flarum\Api\ExceptionHandler;
+
+use Exception;
+use Psr\Log\LoggerInterface;
+use Tobscure\JsonApi\Exception\Handler\ExceptionHandlerInterface;
+use Tobscure\JsonApi\Exception\Handler\ResponseBag;
+
+class FallbackExceptionHandler implements ExceptionHandlerInterface
+{
+    /**
+     * @var bool
+     */
+    protected $debug;
+
+    /**
+     * @var LoggerInterface
+     */
+    protected $logger;
+
+    /**
+     * @param bool $debug
+     * @param LoggerInterface $logger
+     */
+    public function __construct($debug, LoggerInterface $logger)
+    {
+        $this->debug = $debug;
+        $this->logger = $logger;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function manages(Exception $e)
+    {
+        return true;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function handle(Exception $e)
+    {
+        $status = 500;
+        $error = $this->constructError($e, $status);
+
+        if (! $this->debug) {
+            $this->logger->error($e);
+        }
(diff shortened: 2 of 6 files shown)


src/Api/ErrorHandler.php
```diff
@@ -34,8 +35,12 @@ public function __construct(JsonApiErrorHandler $errorHandler)
      * @param Exception $e
      * @return JsonApiResponse
      */
-    public function handle(Exception $e)
+    public function handle(Throwable $e)
     {
+        if (! $e instanceof Exception) {
+            $e = new Exception($e->getMessage(), $e->getCode(), $e);
+        }
```
franzliedke: This is only needed until tobscure/json-api supports `Throwable`, right?

src/Api/ExceptionHandler/FallbackExceptionHandler.php
```diff
@@ -0,0 +1,79 @@
...
+     * {@inheritdoc}
+     */
+    public function handle(Exception $e)
+    {
+        $status = 500;
+        $error = $this->constructError($e, $status);
+
+        if (! $this->debug) {
+            $this->logger->error($e);
+        }
```
franzliedke: Why only log in debug mode?

src/Api/ErrorHandler.php
tobscure: Yep.

src/Api/ExceptionHandler/FallbackExceptionHandler.php
tobscure: My thought was that in debug mode you can see the error displayed when you encounter it, and therefore no need to log it. I suppose it'd be simpler to just log everything regardless though.

tobscure: On second thought, errors that occur in the forum stack don't get logged when debug mode is on, because they go through the HandleErrorsWithWhoops middleware. Whatever we do, it should be consistent between the forum/API stacks. Should we add logging to the Whoops middleware or keep things as-is?

src/Api/ExceptionHandler/FallbackExceptionHandler.php
```diff
@@ -0,0 +1,79 @@
...
+    {
+        return true;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function handle(Exception $e)
+    {
+        $status = 500;
```
luceos: Some Illuminate Exceptions allow setting the HTTP status code as well. These can be found as implementations of the HttpExceptions from Symfony. Forcing uncaught exceptions to be shown as 500 might be conflicting with the reason of the author, eg Unauthorized Exceptions should become 401, whereas Forbidden Exceptions should become 403. We should at least try to see if the Exception has a method `getStatusCode()` and use that.

tobscure: This is what the `Api\ExceptionHandler`s are for. If an exception is thrown that we haven't accounted for in an ExceptionHandler, then it should count as a 500 Internal Server Error regardless, because we don't necessarily want to expose what kind of exception/status code it is. Plus, just using the status code would be awkward because the message would still be "internal server error".

src/Api/ExceptionHandler/FallbackExceptionHandler.php
franzliedke: Yes, I agree. Always logging means a) no surprises and b) persistence.

Issue #1641, opened 2018-11-13
Refactor exception handling, rendering, and reporting


tobyzerner:
From https://github.com/flarum/core/pull/1633#pullrequestreview-173722732:

> What I would like to see in beta.9:
> * One middleware (always enabled) that converts all known exception types with custom views to pretty error pages, always.
> * Another middleware (either Whoops or pretty pages) that a) displays the exceptions and b) logs them. (The logging could even be a separate middleware that re-throws the exceptions, unless that messes with the stack traces.)
> 
> IMO, there is no need for NotFound and similar exceptions to ever be handled with the fallback error handling (Whoops or pretty pages) - we are just using the exception mechanism to handle known error cases.

May be worth taking some inspiration from https://laravel.com/docs/5.7/errors

I would also like to see the ability to configure logging - we may as well make use of the power of [Laravel's logging component](https://laravel.com/docs/5.7/logging).

---

Step-by-step TODO:

- [x] Return proper status codes for Whoops / debug mode
- [ ] Token error message - notes from PR (#1528)
- [ ] Pretty error pages should differ based on exception type, not status code (e.g. CSRF token mismatch)
- [x] Exceptions should expose whether they are loggable / reportable (see https://github.com/FriendsOfFlarum/sentry/commit/8fc69285943d1a9a55d844dbc87d4d63b2748d87#r34287868)

luceos:
One thing I think is a must-have is the ability to easily enable Sentry. This would especially be helpful for taking immediate action when encountering unexpected errors on discuss. I will gladly provide an extension if need be.

franzliedke:
I'd be most interested in your opinion on this:

> IMO, there is no need for NotFound and similar exceptions to ever be handled with the fallback error handling (Whoops or pretty pages) - we are just using the exception mechanism to handle known error cases.

luceos:
Catching NotFound exceptions allows for customising the returned response, eg with a decent error page or through extensions to list related discussions. Unless of course an extension could be configured to handle an uncaught exception (add to the handler stack, I think this is currently the case).

I would be very interested to know whether we can create decent error pages for any HttpException (with 500 probably being the 'exception').

franzliedke:
@luceos Yes, I want to keep that part. My question was about the Whoops error page, which is currently shown (in debug mode) for NotFound exceptions and the like. IMO, these exceptions should always be handled by the error handler stack.

Debug mode (a.k.a. Whoops error pages) would then only have an effect for generic 500 errors.

[...]

PR #1648, opened 2018-11-20
Set Whoops middleware HTTP status to error code

datitisev:
**Refs #1641 ([comment](https://github.com/flarum/core/issues/1641#issuecomment-439217306))**

- When in debug mode, sets the HTTP status code to the error code
  - [WhoopsRunner sets the status code to 500](https://github.com/franzliedke/whoops-middleware/blob/master/src/WhoopsRunner.php#L28)

**Screenshot**

Before:
image

After:
image

Both:
image

**Confirmed**

### DIFF
--- a/src/Http/Middleware/HandleErrorsWithWhoops.php
+++ b/src/Http/Middleware/HandleErrorsWithWhoops.php
@@ -46,7 +46,8 @@ class HandleErrorsWithWhoops implements Middleware
                 $this->logger->error($e);
             }
 
-            return WhoopsRunner::handle($e, $request);
+            return WhoopsRunner::handle($e, $request)
+                ->withStatus($e->getCode() ?? 500);
         }
     }
 }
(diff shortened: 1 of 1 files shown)


src/Http/Middleware/HandleErrorsWithWhoops.php
```diff
@@ -46,7 +46,8 @@ public function process(Request $request, Handler $handler): Response
                 $this->logger->error($e);
             }
 
-            return WhoopsRunner::handle($e, $request);
+            return WhoopsRunner::handle($e, $request)
+                ->withStatus($e->getCode() ?? 500);
```
tobscure: Can we make sure the code is in a valid range before we use it as a HTTP status? i.e. >= 400 && < 600

franzliedke:
I'd like to leave this open for beta.9, when we tackle #1641. I want some more time to think about this and test it out.

Thanks a lot!

PR #1649, opened 2018-11-21
Throw FileNotFoundException when FileSource path does not exist

datitisev:
- Throw `League\Flysystem\FileNotFoundException` when asset file is not found

**Screenshot**
image

**Confirmed**

### DIFF
--- a/src/Frontend/Compiler/Source/FileSource.php
+++ b/src/Frontend/Compiler/Source/FileSource.php
@@ -11,6 +11,8 @@
 
 namespace Flarum\Frontend\Compiler\Source;
 
+use League\Flysystem\FileNotFoundException;
+
 class FileSource implements SourceInterface
 {
     /**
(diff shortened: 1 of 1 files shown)


tobscure:
@Franz I was discussing with @datitisev whether we should be depending on `League\Flysystem` for the exception here, rather than just using a `\InvalidArgumentException` or `\RuntimeException`. Thoughts?

datitisev:
@franzliedke What should the message of the exception be?

And why shouldn't we use an existing FileNotFoundException, which provides exactly what the issue is to the user?
Symfony and Illuminate also have FileNotFoundExceptions, if it's just because it's in League then that's not a problem.
I don't really understand why we wouldn't want to use it, sorry.

franzliedke:
I guess it just feels off to use a class from a framework in a file that is otherwise completely free of any framework coupling.

franzliedke:
You can use the exception text from Flysystem's exception. ;)

franzliedke:
Please don't change the dist file.

datitisev:
Oops! Sorry, looks like I accidentally added it when I was testing the exception 🤦‍♂️

luceos:
that branch name 😂 

let's make it a convention 🤣

Issue #1601, opened 2018-10-17
PHP Extender: Frontend


franzliedke:
We need a `Frontend` extender that allows *extending* (read: inheriting) a "frontend" (group of routes, middleware and assets such as the forum). This is needed for reviving the embed extension.

More thoughts and discussion starting at https://github.com/flarum/core/issues/851#issuecomment-407207717

tobscure:
I don't think this should block beta 8, can't the embed extension can still function using the Compat extender?

franzliedke:
Possibly. That would need to be done as well, though - it might be easier to finish the extender instead. Still torn on the best way forward, I have some WIP code lying around

tobscure:
Closed in edaca3160ed856ef2b42d043c8276c33dc8ea6cd

datitisev:
I think this broke something. `Flarum\Extend\Frontend@content` doesn't seem to be working anymore, I have the following code trying to make flarum/statistics work again, and `app.data.hello` is undefined
![image](https://user-images.githubusercontent.com/6401250/48905907-c83bc000-ee30-11e8-91af-80e024ce0183.png)

I also have a bunch of logs in the callbacks (not included in the screenshot) and none of them show up in the corresponding log file.
Throwing an exception does nothing too.

franzliedke:
In the first `content` call, the listener class is probably implementing the `ContentInterface`, which has been removed.

datitisev:
@franzliedke Actually, it's not.

I think I see the reason why.

- In `Extend\Frontend@registerContent`, when the container is resolving `flarum.$this->frontend.frontend`, one of the argument is `Flarum\Extend\Frontend` instead of `Flarum\Frontend\Frontend`.

- It also seems like the container never resolves, I put an extension inside the callable and it does not throw... only exceptions before resolving the container throw.

  https://github.com/flarum/core/blob/master/src/Extend/Frontend.php#L143

  ![image](https://user-images.githubusercontent.com/6401250/48906168-a989f900-ee31-11e8-830a-2d19b92c4e04.png)
  _first `throw` actually errors with `$content` not being defined, but still shows that the code is 
  being executed_

PR #1652, opened 2018-11-22
Fix frontend extender using old container & wrong class

datitisev:
**Fixes https://github.com/flarum/core/issues/1601#issuecomment-441031735**
**Caused by https://github.com/flarum/core/commit/edaca3160ed856ef2b42d043c8276c33dc8ea6cd**

- Fixes `Extend\Frontend@registerContent` using wrong container name when resolving, and wrong `Frontend` class in argument
  - Argument was using `Frontend` extender instead of `Frontend\Frontend`

**Confirmed**

### DIFF
--- a/src/Extend/Frontend.php
+++ b/src/Extend/Frontend.php
@@ -14,6 +14,7 @@ namespace Flarum\Extend;
 use Flarum\Extension\Extension;
 use Flarum\Frontend\Assets;
 use Flarum\Frontend\Compiler\Source\SourceCollector;
+use Flarum\Frontend\Frontend as FlarumFrontend;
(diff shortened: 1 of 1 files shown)


franzliedke:
Thanks for the fix!

I cherry-picked this to master and tweaked the alias slightly to be consistent with another extender.

franzliedke:
Still an interesting question for @tobscure: Did you intentionally change the name of these bindings? Feels somewhat inconsistent now with the other bindings for routes, middleware etc.

tobscure:
I did, because it felt more inconsistent for someone to be able to define a new frontend resulting in `flarum.myRandomName.frontend` and `flarum.myRandomName.assets` bindings.

PR #1658, opened 2018-11-29
Support extend by callback?

lubobill1990:
**Fixes #0000**
In the document, it seems extend.php can support callback.

**Screenshot**

**Confirmed**

### DIFF
--- a/src/Foundation/InstalledSite.php
+++ b/src/Foundation/InstalledSite.php
@@ -149,7 +149,11 @@ class InstalledSite implements SiteInterface
         $laravel->boot();
 
         foreach ($this->extenders as $extension) {
-            $extension->extend($laravel);
+            if (is_callable($extension)) {
+                $laravel->call($extension);
+            } else {
+                $extension->extend($laravel);
+            }
         }
 
         return $laravel;


tobscure:
Callbacks are already supported (but deprecated) via https://github.com/flarum/core/blob/master/src/Extension/Extension.php#L108

lubobill1990:
Perhaps the code you list here is for extensions installed with composer?

When I add callback to extend in root/extend.php, exception will be thrown and complains that callback::extend() is not exist

tobscure:
Ah I see. The purpose of supporting callbacks is for backwards-compatibility purposes, so we haven't added support for them in the root `extend.php` which is new in beta 8. You can instantiate the `Extend\Compat` extender if you'd like to use a callback in your root `extend.php`.

lubobill1990:
I know, thanks toby.

lubobill1990:
@tobscure I have a question: What's the preferred way to add a event listener to Flarum\Post\Event\Posted in beta 8?

tobscure:
Using a callback ;) We haven't finished implementing extenders for a lot of tasks (including listening for events), there will be issues to track progress soon!

lubobill1990:
Looking forward to it and thanks for your effort!

PR #1674, opened 2018-12-02
Only create migration table if it doesn't exist

CDK2020:
Currently, if you try to install and it fails, you can't recover without dropping the whole install because Flarum will try to create the migration table again. This fix allows you to attempt to install again even after a fail (such as not using Innodb)

* Can we simplify this logic?

**Confirmed**

### DIFF
--- a/src/Database/DatabaseMigrationRepository.php
+++ b/src/Database/DatabaseMigrationRepository.php
@@ -99,10 +99,12 @@ class DatabaseMigrationRepository implements MigrationRepositoryInterface
     {
         $schema = $this->connection->getSchemaBuilder();
 
-        $schema->create($this->table, function ($table) {
-            $table->string('migration');
-            $table->string('extension')->nullable();
-        });
+        if (!$schema->hasTable('migrations')) {
+            $schema->create($this->table, function ($table) {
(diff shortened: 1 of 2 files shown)


tobscure:
An alternative would be using `dropIfExists` before running create. Not sure what would be best here.

CDK2020:
@tobscure that would defeat the whole purpose. If the migration table it dropped it will attempt to re-run all of the previously ran migrations and this would fail again. If instead it just glosses over creating the migration table, it will run smoothly and only run migrations it hasn't done yet.

CDK2020:
@tobscure Whoops, didn't even see that function, but there we go!

franzliedke:
I don't know, I don't know. I guess it can be argued that the installation should *always* be run fully.

Not all migrations are necessarily fully atomic, so trying to resume installation might actually result in a somewhat inconsistent state, which would make debugging a pain. The only way to ensure consistency is having installation run all parts, always.

(To be more constructive: we could look into rolling back everything on failure more completely, e.g. as part of #1617.)

tobscure:
@franzliedke good point. #1617 is a better approach to this.

PR #1697, opened 2018-12-13
Do not resolve services in extenders

franzliedke:
**First shot at flarum/issue-archive#269** (however there's lots of work left)

We want to avoid resolving services in extenders - instead we add closures that are executed when the relevant services are resolved by someone else.

My change of `RecompileFrontendAssets` away from the "event subscriber" pattern - in order to only resolve its dependencies when the corresponding events are fired (very few requests), rather than at the time the event listeners are bound (every single request).

**Confirmed**


tobyzerner (approved):
Beautiful 🧜‍♀️

PR #1700, opened 2018-12-14
Don't resolve services when binding listeners

franzliedke:
**Next part of solving flarum/issue-archive#269**

Replace event subscribers with event listeners in core, where it makes sense.

Exceptions:
- **Subscribers without constructor dependencies** - These have no cost of dependencies, and are useful for grouping a bunch of listeners that simply operate on models.
  - `Flarum\User\UserMetadataUpdater`
  - `Flarum\Discussion\DiscussionMetadataUpdater`
- **Policy subclasses** - These are a bit more complex, as a base class and lots of logic is involved - haven't thought about the best way to get rid of the subscriber model here.

**Confirmed**


franzliedke:
I pushed these (and then some) directly on master. Have gathered enough experience doing all of that for all the extensions. :wink:

PR #1709, opened 2018-12-19
allow get avatar url with event listener

lubobill1990:
**Fixes #0000**

Allow extension to return customized avatar url just like display name

Not much.
The one thing is if the namespace of GetAvatarUrl event is proper, because GetDisplayName event exists at Flarum\Event, however I think it's better to be at Flarum\User\Event

**Screenshot**

**Confirmed**

**Required changes:**


luceos:
I have looked at this PR twice now and I'm confused what the benefit is of using an event over directly mutating the model property at whatever opportunity. Could you clarify that?

franzliedke:
I agree. @lubobill1990 Can you clarify what you want to *achieve* in your extension?

At this point (at least once all of their use cases are covered by extenders), events are basically implementation details - so if you want to introduce hooks for extensions, extenders should be built / adapted.

lubobill1990:
@luceos @franzliedke 
One reason of adding this event is because of the exist of GetDisplayName event, I think they are similar.
Another reason is extensions have better only change the columns created by their own migrations.

The scenario of my app using this is:
After signup with oauth, such as facebook, extensions normally will save the user info in another table, such as username, email, avatar url. And if an user didn't uploaded an avatar, then the avatar of facebook will be used.

Of course, it can be implemented as changing the avatar_url column of that user, but in my opinion, extensions should be encouraged to only change the tables or columns they create by themselves.

lubobill1990:
@luceos @franzliedke @tobscure Hello guys, any news on this pull request?

franzliedke:
No. Please be patient.

composer.json
```diff
@@ -1,5 +1,5 @@
 {
-    "name": "flarum/core",
+    "name": "zhishiq/flarum-core",
```
franzliedke: Don't change this please.

lubobill1990: It's weird, I didn't include this change to the pull request...

lubobill1990:
@franzliedke Hi Franz, I've removed the unexpected code from the commit~

luceos:
> One reason of adding this event is because of the exist of GetDisplayName event, I think they are similar.

That one event exists, doesn't prove the need for another.

> Another reason is extensions have better only change the columns created by their own migrations.

That depends completely on what the extension does and the intent of the column. In this case having a single source of truth, the avatar_url column seems sufficient enough?

>  And if an user didn't uploaded an avatar, then the avatar of facebook will be used.

If a user signed up using a third party and the avatar is adopted, the user should be able to override it and even delete it. Forcing an avatar back to the one used on the provider would reduce the ability of a user to decide whether he wants one or not.

Also there's a RegisteringFromProvider event specifically for situations where users sign up through a third party auth service.

> And also, this event can be used as returning default avatar url when it's empty.

Okay makes sense, but you could also set the avatar_url to the default one using the Registered event in case no avatar_url was set.

> Of course, it can be implemented as changing the avatar_url column of that user, but in my opinion, extensions should be encouraged to only change the tables or columns they create by themselves.

The avatar_url however is a perfect example of a column that could easily be modified by extensions. The reason for wanting to use your own column is out of fear that different extensions interact conflictingly with the avatar, but that is also the case when using an event. 

I still see no benefit to adding this event.

lubobill1990:
😂

lubobill1990:
I think GetAvatarUrl event adds flexibility to flarum when brining little complexity to flarum. And it also encourages a self control convention for extension community.

Anyway, if the core team thinks it's not necessary, changing my own extension doesn't take much efforts.

luceos:
@lubobill1990 this is one opinion. You should wait for the decision by merge master @franzliedke

lubobill1990:
@luceos Thanks. This event is important for me. When I submit this pull request, I'm writing an extension providing API for wechat mini app. User register through API, and API save avatar url in its own table.

And now, I'm writing another extension to provide default avatar based on user's ID and username with gravatar. It's not appropriate to overwrite current avatar url column.

tobscure:
Rather than adding an event which should be considered private, I'd rather we come up with a solution (API) for this problem using Extenders. @franzliedke

franzliedke:
Closing in favor of a (future) solution using an extender.

PR #1723, opened 2019-01-07
Open up Site for local implementation

luceos:

The visibility of the config file related methods of Site are private. This disallows implementing a different way of using one. Eg virtualised or in the cloud values are set in the environment, which allows for more flexibility. 

I'm proposing opening up the methods to protected, so that a local implementation can take a different course in getting the config file.

Do we want this?

**Confirmed**

### DIFF
--- a/src/Foundation/Site.php
+++ b/src/Foundation/Site.php
@@ -39,12 +39,12 @@ class Site
         }
     }
 
-    private static function hasConfigFile($basePath)
+    protected static function hasConfigFile($basePath)
     {
         return file_exists("$basePath/config.php");
     }
 
-    private static function loadConfig($basePath): array
+    protected static function loadConfig($basePath): array
     {
         $config = include "$basePath/config.php";
 


franzliedke:
Not a big fan, to be honest. I even kinda like [this approach](https://matthiasnoback.nl/2018/09/final-classes-by-default-why/). You don't even need to override this class, just create a `InstalledSite` object directly with the properties that you derive from environment variables.

Hope that's okay?

luceos:
The problem here is that I also still would need an Installer in case the forum needs to be installed on a development machine. So I'd still need both Installed and UninstalledSite..

franzliedke:
If you're dealing with environment variables, methods like `hasConfigFile()` don't make much sense anyway...

I think implementing a different way of instantiating any implementation of `SiteInterface` should be simple enough. `Site` mostly exists to keep the default code in flarum/flarum as short as possible (and as an abstraction layer to change the skeleton without forcing people to do manual updates there).

As you are basically creating your own skeleton anyway, I'll trust you'll be able to manage a few more lines of code. :grin:

PR #1724, opened 2019-01-07
Allow extending UninstalledSite / Installer

luceos:

The goal of Flarum is to allow easy extensibility. In certain specific situations you might want to hook into the Installer or UninstalledSite as well. Currently there is absolutely no way of hooking any extender into it, even though the skeleton could be changed to allow that.

This PR adds a trait that moves the `extendWith` call and it's related array of ExtenderInterfaces outside of InstalledSite and adds it to UninstalledSite.

- [ ] Do we want to allow this?
- [ ] Do we want to use a trait/concern?

**Confirmed**

### DIFF
--- a/src/Foundation/Concerns/Extending.php
+++ b/src/Foundation/Concerns/Extending.php
@@ -1,5 +1,14 @@
 <?php
 
+/*
+ * This file is part of Flarum.
+ *
+ * (c) Toby Zerner <toby.zerner@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
 namespace Flarum\Foundation\Concerns;
 
 use Flarum\Foundation\SiteInterface;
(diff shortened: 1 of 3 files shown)


franzliedke:
My original plan (we're 80% there) was to decouple the installer as much as possible from Flarum's core, including from things like a Laravel container. That would mean running extenders in that "app" wouldn't work anymore.

Personally, I thought the way forward would be 1) decoupled installer, 2) enabling extensions **after** installing core, 3) running things like a converter. So 2) and 3) would happen when Flarum is already "installed", i.e. with an instance of `InstalledApp`.

We need to discuss this, probably best in a developer meeting.

frinkfromthedrink:
Sorry for jumping in here, but just thought I'd share my opinion. I personally feel that if people feel the need to hook into the extender, they can fork Flarum and customize it. Installers shouldn't need to be changed, and if people feel changes need made, the best way is a PR. Allowing people to modify the installer can open up security vulnerabilities that might not be noticed.

franzliedke:
Indeed. I think I'd rather have the installer as decoupled as possible from the rest of the application, and focus on setting up core and core only. This is so much simpler - and building alternative installers is still easily possible, especially with the changes in #1617.

@luceos I hope you can live with this. :smiley:

PR #1743, opened 2019-02-01
Improve test suite structure

franzliedke:
This is the promised revamp of the test suite.

Aims:
- **Separate unit from integration tests** so that the former can run quickly (you can do so with the `composer test:unit` shortcut.
- **Make integration tests independent.** The new infrastructure allows for setting up all necessary data ("fixtures") before each test run. Each test class defines a set of rows which are expected to be found in the database. We then truncate those tables, insert the relevant data, and voila. IMO, a good trade-off between performance (don't truncate everything!) and developer overhead (you need to remember to set up all required state in the `setUp()` method).
- **Do not run the installer on every test run.** Instead, developers (and Travis) should run a setup script (that makes use of the new `Installation` class) to setup a fully migrated database against which all integration tests (mostly testing the API right now) can run.

This means, the `DefaultInstallationTest` is now removed. IMO, this is an improvement, because previously the whole test suite depended on this test passing. It also wasn't doing much but taking a whole lot of time. We still get the basic "did the installer work?" to some degree by running the setup script first.

**Confirmed**

**Future improvements**

- Test the API, not controllers. This further decouples the tests from our codebase, which makes safe refactorings possible (a major goal of testing).
- Improve the setup script with more customizability (when needed) and more helpful output. I already worked a bit on making it quick-start-friendly, but maybe we can be even more friendly.


tests/integration/TestCase.php
```diff
@@ -0,0 +1,81 @@
...
+            );
+        }
+
+        return $this->database;
+    }
+
+    protected function prepareDatabase(array $tableData)
+    {
+        // We temporarily disable foreign key checks to simplify this process.
+        $this->database()->statement('SET FOREIGN_KEY_CHECKS = 0');
```
luceos:
```suggestion
        $this->database()->getSchemaBuilder()->disableForeignKeyConstraints();
```

That might be better cross-database-driver compatible ;)

tests/integration/TestCase.php
```diff
@@ -0,0 +1,81 @@
...
+            $this->database()->table($table)->truncate();
+        }
+
+        // Then, insert all rows required for this test case.
+        foreach ($tableData as $table => $rows) {
+            $this->database()->table($table)->insert($rows);
+        }
+
+        // And finally, turn on foreign key checks again.
+        $this->database()->statement('SET FOREIGN_KEY_CHECKS = 1');
```
luceos:
```suggestion
        $this->database()->getSchemaBuilder()->enableForeignKeyConstraints();
```

luceos:
> Basically, it was a preference-driven choice to differentiate between "real" namespaces and those I introduced to differentiate between types of tests / test suites.

I'm not sure I'm in favour of that preference. I'd rather not introduce something (new) developers/contributors have to be specifically taking into consideration.

I'll try to pull this in tomorrow and run it locally.

tobscure:
Nice @franzliedke, this is fantastic!

> Future improvements: Test the API, not controllers. This further decouples the tests from our codebase, which makes safe refactorings possible (a major goal of testing).

This excites me

PR #1744, opened 2019-02-04
Capsule manager

### LINKED ISSUE #1740: No database reconnector configured
## Bug Report

**Current Behavior**

Whenever you have long running processes that are cut off by the server (mysql settings), a `LogicException` is thrown because there's no reconnector available:

```
Lost connection and no reconnector available.
```

The origin of this is that we are instantiating the MySQL connection manually from within the DatabaseServiceProvider using the ConnectionFactory. This Factory compared to DatabaseManager (which is also available inside our vendor files) does not set up the reconnector.

You won't easily run into this, unless you have long running background tasks in cli. Eg I'm importing users, permission groups, discussions and posts using a testing database. At some point the connection is lost and Flarum cannot continue.

**Steps to Reproduce**

Write a command that inserts/duplicates something, eg users. Keep it running for at least 300k entries.

**Expected Behavior**
The connection is reconnected.

**Environment**
- Flarum version: dev-master/beta 9
- Webserver: nginx
- Hosting environment: local
- PHP version: 7.1

**Possible Solution**
I've considered adding the reconnector from the DatabaseServiceProvider. But just doing that might not be enough. Looking at the DatabaseManager, Laravel does far more than just set the connection. In addition I tested this locally and the issue persisted.

**Additional Context**
☕️

luceos:
**Fixes #1740**

This PR moves away from configuring our ConnectionInterface completely independently. Because this causes an issue when the database connection is lost. 

- [x] Is it OK to use the Capsule Manager instead of the DatabaseManager.

**Confirmed**


tobscure:
🎉

PR #1749, opened 2019-02-12
Allowing access to admin area when Flarum is offline (maintenance)

### LINKED ISSUE #255: Offline mode doesn't allow access to administrators in beta 8
## Bug Report

**Current Behavior**
Enabling `offline` mode in `config.php` disallows access to already logged-in administrators. In beta 7, the administrators were allowed to access the `/admin` area.

**Steps to Reproduce**
1. Open a forum that is on beta 8 and where you have access to an administrator account
2. Make sure you're logged in as an adminsitrator
3. Set the key `offline` to the value of `true` in the `config.php` file
4. Open the forum again in the browser and see the maintenance message on both `/` and `/admin`

**Expected Behavior**
The administrator should be able to see _at least_ the administration area, as it was in beta 7, according to flarum/framework#459.

**Screenshots**
immagine

**Environment**
- Flarum version: 0.1.0-beta.8.1.
- PHP version: 7.2.12

luceos:
**Fixes flarum/issue-archive#255**

This allows access to admin when maintenance mode is enabled. We still have no way of allowing admins to log in though.

- [ ] Does a middleware suit the needs?
- [ ] Any recommendations for alternative implementations?

**Confirmed**


src/Api/ApiServiceProvider.php
```diff
@@ -49,6 +49,8 @@ public function register()
         $this->app->singleton('flarum.api.middleware', function (Application $app) {
             $pipe = new MiddlewarePipe;
 
+            $pipe->pipe($app->makeWith(HttpMiddleware\MaintenanceMode::class, ['maintenance' => $app->isDownForMaintenance()]));
```
franzliedke: I didn't know `makeWith()` [existed](https://github.com/laravel/framework/blob/5.7/src/Illuminate/Container/Container.php#L588-L598), interesting. :grinning: 

Since it's not part of the contract, could you just use `make()` instead?

src/Http/Middleware/MaintenanceMode.php
```diff
@@ -0,0 +1,78 @@
...
+            $status = 503;
+
+            event(new RenderMaintenancePage($template, $status));
+
+            $view = $this->view->make($template);
+
+            return new HtmlResponse($view->render(), $status);
+        }
+
+        return $handler->handle($request);
```
franzliedke: Can we go with something like this, to de-duplicate the conditions and have a common level of abstraction in this method?

~~~php
if (! $this->maintenance) {
    return $handler->handle($request);
}

if ($this->isApiRequest()) {
    return $this->apiResponse();
} else {
    return $this->htmlResponse();
}
~~~

src/Foundation/InstalledApp.php
```diff
@@ -48,7 +48,7 @@ public function __construct(Container $container, array $config)
     public function getRequestHandler()
     {
         if ($this->inMaintenanceMode()) {
-            return new MaintenanceModeHandler();
+//            return new MaintenanceModeHandler();
```
luceos: Not sure why I commented this out, but it should probably go, including the handler.

src/Api/ApiServiceProvider.php
luceos: Odd 🤔 I must have looked it up in the docs which mentions makeWith instead of make:

image

tobyzerner (changes requested):
I like this except for two things:

1. Not a fan of the RenderMaintenancePage event. I would just leave it out altogether until (if) we need it, and we can consider how it should work then.

2. Instead of a custom `maintenance.blade.php` template, can we just use `error.default` and pass through the message.

src/Api/ApiServiceProvider.php
tobscure: I think maybe there was some point at which Taylor wanted to remove the ability to pass params through the `make` API, and added a `makeWith` method to keep people who rely on that functionality happy. But there was backlash so it got reverted?

franzliedke (changes requested):
- This needs a rebase because of the latest middleware changes.
- Disabling the API completely is tricky - won't that break quite a few admin features as well?

franzliedke:
We could also take this back to the drawing board and discuss possible solutions / improvements in flarum/issue-archive#255.

luceos:
Let's do that. This pr is ancient.

PR #1768, opened 2019-03-29
moved GetDisplayName event to User namespace

### LINKED ISSUE #1742: Move GetDisplayName to User namespace
The GetDisplayName event is stored under the src/Event, this is incorrect. The event is purely in use for the Flarum/User namespace and as such should be moved to Flarum/User/Event.

All related code has to change and a changelog item should be created.

luceos:
**Fixes #1742**

This is pretty straightforward.


luceos:
@franzliedke might as well merge this for beta 9 as well.

franzliedke:
Once we have an extender here, the event itself may be refactored away at any time. :smile: 

(It's not really a domain *event*, we just use the event mechanism to hook into the process of determining the right display name.)

PR #1773, opened 2019-04-15
Queue support

luceos:
**Relates to #978**

**TO DO**

- [ ] Rewrite mail handling, possibly using Mailables
- [ ] Document the driver configuration
- [ ] Create a test to check the sync driver works.

This PR adds queue ability, it sets the driver to sync by default. It will also enable the commands if the driver is something else than the sync or null driver.

Does this implementation comply to our conventions and do you foresee issues in extensibility.

**Confirmed**

**Required changes:**


luceos:
@flarum/core I have been considering adding a Configuring event here as well, to immediately override the default queue. Instead of needing to work with a `afterResolving` callback. What do you think?

tobscure:
Looks good, I think `afterResolving` is fine since its use should only be internal anyway (the public API should be an extender).

franzliedke:
I agree with @tobscure, the event shouldn't be necessary.

src/Queue/QueueServiceProvider.php
```diff
@@ -0,0 +1,77 @@
...
+    ];
+
+    public function register()
+    {
+        $this->app->singleton(ConnectorInterface::class, function ($app) {
+            return $app['queue']->connection();
+        });
+
+        $this->app->singleton(Factory::class, function ($app) {
+            $manager = new QueueManager($app);
```
franzliedke: Do we really benefit from the manager here? I don't think we need to support multiple queue backends, and using the manager just means that we'll have to translate extension configuration (e.g. settings in a GUI) into Laravel config and let the manager turn that into a connector instance. Why not turn our settings into a driver directly?

luceos: Makes sense, didn't even consider that. We don't even need the configuration, any listener that binds into the resolving of this binding can change what queue driver is being used.

luceos: Okay so simplifying the binding doesn't work, because the Mailer instance expects a Factory, not an implementation..

luceos: The Validator too. I've reverted my latest commits. Please review again @flarum/core .

luceos:
@franzliedke is this something we can take along for beta 9? :ninja:

luceos:
This is starting to become very ugly. The queue worker needs a cache and does so by resolving `cache` from ioc and then calling a method `driver()`. We don't use the cache manager.

```php
    protected function runWorker($connection, $queue)
    {
        $this->worker->setCache($this->laravel['cache']->driver());

        return $this->worker->{$this->option('once') ? 'runNextJob' : 'daemon'}(
            $connection, $queue, $this->gatherWorkerOptions()
        );
    }
```

I'll try to get an implementation done and then review how to improve all this.

luceos:
@franzliedke subtle request to give this PR a review 🤞

franzliedke:
Oh wow, this part of Laravel is indeed a bit too highly coupled. Thankfully, [change is on the horizon](https://github.com/laravel/framework/pull/29224). As we can't wait for that, I will experiment with alternatives in the meantime.

franzliedke:
I pushed a few commits that let us avoid the managers. Let me know what you think.

PR #1775, opened 2019-04-18
Remove post id from serializer attributes

clarkwinkelmann:
Remove the unneeded, duplicate post ID from the post serialized attributes. It's already present as the resource ID and I have not found any use of that duplicate attribute that's been there since 4 years. This attribute is not present in the javascript model for Post so I don't think it's used at all.

**Confirmed**

I just noticed this and didn't want to open an issue just for that, but can't test dev-master right now. I can verify this PR on a dev-master stup this weekend if needed.


tobscure:
nice find!

PR #1776, opened 2019-04-19
[BUGFIX] The avatar_path should be changed to avatar_url

lubobill1990:
In a previous commit, after avatar_path is renamed, this place is not changed.

The consequence is that, when avatar is uploaded, the previous image file is not deleted.

**Fixes #0000**

**Screenshot**

**Confirmed**

- [ x ] Frontend changes: tested on a local Flarum installation.
- [ x ] Backend changes: tests are green (run `php vendor/bin/phpunit`).

### DIFF
--- a/src/User/AvatarUploader.php
+++ b/src/User/AvatarUploader.php
@@ -11,6 +11,7 @@
 
 namespace Flarum\User;
 
+use Illuminate\Support\Arr;
 use Illuminate\Support\Str;
 use Intervention\Image\Image;
 use League\Flysystem\FilesystemInterface;
@@ -46,7 +47,7 @@ class AvatarUploader
 
     public function remove(User $user)
     {
-        $avatarPath = array_get($user->getAttributes(), 'avatar_url');
+        $avatarPath = Arr::get($user->getAttributes(), 'avatar_url');
 
         $user->afterSave(function () use ($avatarPath) {
             if ($this->uploadDir->has($avatarPath)) {
--- a/src/User/AvatarUploader.php
+++ b/src/User/AvatarUploader.php
@@ -46,7 +46,7 @@ class AvatarUploader
 
     public function remove(User $user)
     {
-        $avatarPath = $user->avatar_path;
+        $avatarPath = array_get($user->getAttributes(), 'avatar_url',);
 
         $user->afterSave(function () use ($avatarPath) {
             if ($this->uploadDir->has($avatarPath)) {
--- a/src/User/AvatarUploader.php
+++ b/src/User/AvatarUploader.php
@@ -46,7 +46,7 @@ class AvatarUploader
 
     public function remove(User $user)
     {
-        $avatarPath = array_get($user->getAttributes(), 'avatar_url',);
+        $avatarPath = array_get($user->getAttributes(), 'avatar_url');
 
         $user->afterSave(function () use ($avatarPath) {
             if ($this->uploadDir->has($avatarPath)) {


lubobill1990:
@luceos @franzliedke This is a bugfix, hope it be included to beta9

src/User/AvatarUploader.php
```diff
@@ -46,7 +46,7 @@ public function upload(User $user, Image $image)
 
     public function remove(User $user)
     {
-        $avatarPath = $user->avatar_path;
+        $avatarPath = array_get($user->getAttributes(), 'avatar_url');
```
luceos: Please import the Laravel Arr class and use `Arr::get()`.

Also this method might need a check whether the url has a protocol or not. There is a difference between `https://somestorage.com/avatar.png` vs `/assets/avatars/avatar.png`.

tobscure: @luceos do you think the `if ($this->uploadDir->has($avatarPath))` check will be sufficient?

luceos: My thought exactly @tobscure, I am not entirely sure though. What do you think @franzliedke?

lubobill1990: @luceos Already changed to Arr::get. So is it not preferred in flarum project to use functions without namespaces?

lubobill1990:
@luceos Any problems with this PR, I regard the fix is obvious.

src/User/AvatarUploader.php
luceos: Most Laravel helpers are deprecated in the framework itself. It won't make sense to keep them around as we'd have to refactor them in the future anyway.

src/User/AvatarUploader.php
```diff
@@ -46,7 +47,7 @@ public function upload(User $user, Image $image)
 
     public function remove(User $user)
     {
-        $avatarPath = $user->avatar_path;
+        $avatarPath = Arr::get($user->getAttributes(), 'avatar_url');
```
franzliedke: This breaks encapsulation of our model object. :confused: 

Instead, I think I would prefer to see a change to the `AvatarChanged` event to hold the following four values:
- the user whose avatar was changed
- the avatar
- the old `avatar_url` value
- the new `avatar_url` value

...and then implement avatar deletion in an event listener.

Thoughts?

lubobill1990: Because AvatarUploader.php is at the same encapsulation level, which means AvatarUploader can be seen as at the same package level with User.php, so I regard it's acceptable to call `Arr::get($user->getAttributes(), 'avatar_url');` in AvatarUploader.php.

Although, it also makes sense to put both old avatar_url and new avatar_url into AvatarChanged, and remove old avatar with listening to AvatarChanged event.

I suggest to accept this pull request first, because it's a bug rather than an enhancement. And make a new issue to make the enhancement. I'll do that issue then.

luceos:
I like the event. Makes it more extensible. Other than that I see no valid argument to make the logic more complex just for the sake of encapsulation.

luceos:
Go away Stalebot.

askvortsov1:
Considering 308f2c9efd4da9e78db0f14c594bc19c98e18519, should this be closed, or would y'all still like to see a PR to expand the event? @luceos

franzliedke:
You're right! Nice catch.

franzliedke:
@askvortsov1 That said, yeah, I wouldn't mind about a PR improving this with an event-based approach. As I said, it would improve encapsulation. Right now, we're reaching into the depths of the Eloquent model from the `AvatarUploader`.

PR #1779, opened 2019-05-01
Fix dropdown icon not showing in UserCard when on UserPage

clarkwinkelmann:
The rule hiding the icon in the UserHero was too broad and applied to UserCard in the list of posts as well
The float rule was redundant

**Screenshot**

Previous (when seen on user profile page):

image

**Confirmed**

### DIFF
--- a/less/forum/UserPage.less
+++ b/less/forum/UserPage.less
@@ -1,10 +1,5 @@
 .UserPage {
-  .UserCard-controls {
-    float: right;
-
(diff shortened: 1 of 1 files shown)


clarkwinkelmann:
I realize now it goes back to beta 7 or even earlier, so I'm surprised nobody noticed this before. I have not found any existing issue about it on the tracker. Just wanted to mention this for future reference if needed.

PR #1787, opened 2019-05-29
remove use of like

luceos:
This relates to https://github.com/flarum/tags/pull/61#discussion_r288466243

The point is that by using `like` in situation where an `_` is used, it allows a one character replacement in MySQL. As such when you have a key named `tes_` and `tes1` by using `$key = 'tes_'` it might return `tes1` instead, which is unwanted.

There are two places this happens. The first one is in the UserRepository, which is being used by the xhr to retrieve a user profile. @meezaan reported he already experienced this issue on his installation where usernames had spaces replaced with underscores. The second one is for the settings repository, this delete method is only used inside the Setting Migration, so it can use an exact match too.

### DIFF
--- a/src/User/UserRepository.php
+++ b/src/User/UserRepository.php
@@ -75,7 +75,7 @@ class UserRepository
      */
     public function getIdForUsername($username, User $actor = null)
     {
-        $query = User::where('username', 'like', $username);
+        $query = User::where('username', $username);
 
         return $this->scopeVisibleTo($query, $actor)->value('id');
     }
--- a/src/Settings/DatabaseSettingsRepository.php
+++ b/src/Settings/DatabaseSettingsRepository.php
@@ -45,8 +45,8 @@ class DatabaseSettingsRepository implements SettingsRepositoryInterface
         $query->$method(compact('key', 'value'));
     }
 
-    public function delete($keyLike)
+    public function delete($key)
     {
-        $this->database->table('settings')->where('key', 'like', $keyLike)->delete();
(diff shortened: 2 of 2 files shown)


src/User/UserRepository.php
```diff
@@ -75,7 +75,7 @@ public function findByEmail($email)
      */
     public function getIdForUsername($username, User $actor = null)
     {
-        $query = User::where('username', 'like', $username);
+        $query = User::where('username', $username);
```
franzliedke: Right now, we do a case-insensitive search here - this would change that. Is that intentional? At least for user search, case-insensitivity would probably be nice, right?

luceos: This method is used on profile pages, these url's are generated by the app and should already match the username precisely. Nevertheless I agree that an exact lowercase query would be better: how about `lower(username) = $username` and use strtolower on that var beforehand?

franzliedke: Yes, it's (probably) fine for those pages, but the method is also used in user searches (`AuthorGambit`). Your solution sounds fine, but it may be more complex when we tackle UTF-8 usernames / display names.

luceos:
As @tobyzerner points the initial select is already case insensitive, can we close this without a change?

franzliedke:
Well, this PR doesn't change case-sensitivity (which is great, and I didn't know that), but it does prevent the special characters `_` and `%` from affecting the query. Apparently, we do not need `LIKE` here.

Issue #1789, opened 2019-06-03
Console not making use of error handling


luceos:
I've encountered a slight problem with the inability to handle errors in console. Both the API and Frontends (admin/forum) are using the middleware pipe which can be used by extensions to hook into. For this use case I was testing the functionality of [fof/sentry](https://github.com/FriendsOfFlarum/sentry) for Bokt, an extension we aim to be using on discuss soon as well.

The console side of Flarum is very slim, it uses the Symfony Console Application and, aside from registering commands, that's pretty much it. What we need to do is implement the same error handler stack of the middleware pipe. I haven't looked into possible solutions yet and would to hear your opinion @franzliedke as you refactored that part of Flarum recently.

The problem with not having an error handler is that any exception thrown in the application will kill the process. If you don't output the console output to a log you'll never be able to identify an issue at all. Being able to log these errors to a third party service like Sentry helps in taking action as soon as possible, especially when actively using the queue.

luceos:
I've looked into this and symfony console offers a way to work with erroring. In order to do so we need to add an EventDispatcher to the Application. Some information:

- https://symfony.com/doc/current/components/console/events.html
- https://symfony.com/blog/new-in-symfony-3-3-better-handling-of-command-exceptions

franzliedke:
Sounds like a good idea, and the changes should be mostly in `Flarum\Console\Server`. Thanks for doing the research, @luceos.

Open for contributions here!

luceos:
This was fixed in the merged PR's. It now uses the general error handling stack.

PR #1810, opened 2019-07-08
Console event subscribing

luceos:
**Relates to #1789**

> Please note this PR does not add error logging for console in core!

This PR adds an EventDispatcher to the Console Application. This event dispatcher is passed on to the Configuring event so that extensions can hook into it.

This is one path for implementation. It's the cleanest I could think of. I think we don't want or need a bloated console side of our software. This implementation adds a dependency to the symfony/event-dispatcher package, this package is required to match the Interface requirement inside Symfony Application.

An alternative implementation would be to start using Illuminate Console instead. The implementation would possibly be easier, no research done yet.

An example implementation of this proposal can be found on the fof/sentry package, PR: https://github.com/FriendsOfFlarum/sentry/pull/1


luceos:
Should this PR be based off of your new exception handling logic @franzliedke ?

franzliedke:
Yeah, I was hoping this would be possible now.

luceos:
In order to add your new Error Registry and handler I need to merge this to move forward, can you review this approach @franzliedke ?

luceos:
I am merging this as I see no alternative right now. Let me know if you disagree.

composer.json
```diff
@@ -56,6 +56,7 @@
         "s9e/text-formatter": "^1.2.0",
         "symfony/config": "^3.3",
         "symfony/console": "^4.2",
+        "symfony/event-dispatcher": "^4.3.2",
```
franzliedke: Why the specific constraint?

src/Console/Server.php
```diff
@@ -46,6 +47,8 @@ private function extend(ConsoleApplication $console)
         $app = Application::getInstance();
 
         $events = $app->make(Dispatcher::class);
-        $events->fire(new Configuring($app, $console));
+        $events->fire(new Configuring($app, $console, $dispatcher = new EventDispatcher()));
+
+        $console->setDispatcher($dispatcher);
```
franzliedke: I was assuming we would handle events ourselves and pass the exceptions to our array of [reporters registered with the application](https://github.com/flarum/core/blob/12dc4fff573b7cc053e6eaf95756297469cfb7b0/src/Foundation/ErrorServiceProvider.php#L73). That way, the Sentry exception would not need any further changes.

If you would furthermore pipe the error through the registry, [you could use that](https://github.com/flarum/core/blob/12dc4fff573b7cc053e6eaf95756297469cfb7b0/src/Http/Middleware/HandleErrors.php#L61-L67) to determine whether the error is even worth reporting.

Issue #1134, opened 2017-02-25
Prevent duplicate content for alternative home page URL


dav-is:
www.google.com/search?q=site%3Adiscuss.flarum.org/all

This is an issue on a few levels. Why can users access https://discuss.flarum.org/all ? I'm told that there's an infinite number of pages too. Like `https://discuss.flarum.org/all?page={x}`.

franzliedke:
There is no NoJS content in these search results (anymore?), as far as I can tell.

As for the `/all` page - that should only be indexed when Flarum's "Home page" setting is set to "Tags", right? When not, it should have another rel=canonical tag that points to the base URL. And vice versa: When the home page is set to "Tags", the URL `/tags` should not be indexed...

TurboProgramming:
![46745674567](https://user-images.githbusercontent.com/17576065/42312178-5f821aa6-8048-11e8-8c56-c761e412f0dd.png)

@franzliedke

luceos:
@TurboProgramming what is the screenshot about? I see no duplicates, nor an `/all` or `/tags` entry..

TurboProgramming:
@luceos Just a screenshot of a Google search for all replies appearing on the results, this may be something complicated for the user.

But I don't know whether this is related to the current issue.

franzliedke:
The screenshot by @TurboProgramming are about a different problem... see issue #1140, which I just closed. :grin:

*(Moderator note: I am hiding the post with the screenshot - and some replies - so that we can focus on topic here.)*

PR #1814, opened 2019-07-10
Add canonical URL to discussion list

datitisev:
**Refs #1134**

Add a canonical URL that defaults to the base URL if the default route is set to the All Discussions route.

Not sure if this is how we want to check the default route or use the logic for `/all` without query.

**Confirmed**

**Required changes:**

- [ ] Related core extension PRs:
    - flarum/tags will need a PR too to complete requirements of #1134.

### DIFF
--- a/src/Forum/Content/Index.php
+++ b/src/Forum/Content/Index.php
@@ -58,9 +58,11 @@ class Index
         ];
 
         $apiDocument = $this->getApiDocument($request->getAttribute('actor'), $params);
+        $defaultRoute = app()->make('flarum.settings')->get('default_route');
 
         $document->content = $this->view->make('flarum.forum::frontend.content.index', compact('apiDocument', 'page'));
         $document->payload['apiDocument'] = $apiDocument;
+        $document->canonicalUrl = $defaultRoute === '/all' ? app()->url() : $request->getUri()->withQuery('');
 
         return $document;
     }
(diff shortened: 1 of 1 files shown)


src/Forum/Content/Index.php
```diff
@@ -58,9 +58,11 @@ public function __invoke(Document $document, Request $request)
         ];
 
         $apiDocument = $this->getApiDocument($request->getAttribute('actor'), $params);
+        $defaultRoute = app()->make('flarum.settings')->get('default_route');
 
         $document->content = $this->view->make('flarum.forum::frontend.content.index', compact('apiDocument', 'page'));
         $document->payload['apiDocument'] = $apiDocument;
+        $document->canonicalUrl = $defaultRoute === '/all' ? app()->url() : $request->getUri()->withQuery('');
```
franzliedke: Please inject the URL generator and the settings repository instead of using `app()`.

datitisev: @franzliedke What about `app()->url()`? That has some more logic than just getting the forum url (for example, if `url` is an array). Duplicating that logic would be pointless, right ?

franzliedke: No, all of that should be taken care of if you call `$this->url->to('forum')->base()`.

franzliedke: Using `app()` means we're hiding the dependencies of our class deep within its methods' implementation, whereas injecting them means listing them very clearly right in the constructor signature.

datitisev: Done @franzliedke.

PR #1817, opened 2019-07-10
Allow extensions to add routes with custom controller before resolving default route

### LINKED ISSUE #1819: Extensions unable to change default route controller
This is an unintended side effect introduced in https://github.com/flarum/core/commit/e4514d84131a82785157dd15550ba2278a5cc154.

By not populating the routes _after_ resolving, this means the default route is resolved before any extensions are able to add routes to the route collection.

Neither `api` nor `admin` need the same fix as they do not have customizable/backend default routes.

I found the bug after trying to find out why fof/tags suddently stopped passing the correct API document, and discovered that it was not using my custom `Content` class and instead passing an array of discussions to the page.

datitisev:
Fixes #1819.

**Confirmed**

### DIFF
--- a/src/Forum/ForumServiceProvider.php
+++ b/src/Forum/ForumServiceProvider.php
@@ -47,7 +47,10 @@ class ForumServiceProvider extends AbstractServiceProvider
         });
 
         $this->app->singleton('flarum.forum.routes', function () {
-            $routes = new RouteCollection;
+            return new RouteCollection;
+        });
+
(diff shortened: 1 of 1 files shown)


franzliedke:
I went with [a slightly different approach](https://github.com/flarum/core/commit/bd040233597f5c21ce272639bf85cd9b8580f46f) so that the registration of forum routes is more consistent with the other ones.

Thanks!

PR #1829, opened 2019-07-28
Add updating 'page' parameter in discussion list

dsevillamartin:
**Refs flarum/framework#1820** (pts 2 & 4)

- Add/update `page` query param when loading more discussions
    - Keeps track of current page & offset
    - Works with tags extension

The logic that adds the parameter looks a bit ugly, but it is to make sure existing query parameters remain (and not only those added through JS).

**Confirmed**

### DIFF
--- a/js/src/forum/components/DiscussionList.js
+++ b/js/src/forum/components/DiscussionList.js
@@ -29,6 +29,18 @@ export default class DiscussionList extends Component {
     this.moreResults = false;
 
     /**
+     * Current page in discussion list
+     */
+    this.page = Number(m.route.param('page')) || 1;
+
+    /**
+     * Number of discussions to offset for pagination
+     *
+     * @type {number}
+     */
+    this.offset = (this.page - 1) * 20;
+
+    /**
      * The discussions in the discussion list.
      *
      * @type {Discussion[]}
@@ -140,10 +152,10 @@ export default class DiscussionList extends Component {
   /**
    * Load a new page of discussion results.
    *
-   * @param {Integer} offset The index to start the page at.
+   * @param {Number} offset The index to start the page at.
    * @return {Promise}
    */
-  loadResults(offset) {
+  loadResults(offset = this.offset) {
     const preloadedDiscussions = app.preloadedApiDocument();
 
     if (preloadedDiscussions) {
@@ -165,7 +177,9 @@ export default class DiscussionList extends Component {
   loadMore() {
     this.loading = true;
 
-    this.loadResults(this.discussions.length).then(this.parseResults.bind(this));
+    this.page++;
+
+    this.loadResults((this.offset += 20)).then(this.parseResults.bind(this));
   }
 
   /**
@@ -180,7 +194,21 @@ export default class DiscussionList extends Component {
     this.loading = false;
     this.moreResults = !!results.payload.links.next;
 
+    // Construct a URL to this discussion with the updated page, then
+    // replace it into the window's history and our own history stack.
     m.lazyRedraw();
+
+    const query = m.route.parseQueryString(document.location.search);
+
+    if (this.page !== query.page) {
+      if (this.page !== 1) query.page = this.page;
+
+      const url = new URL(document.location.href);
+
+      url.search = m.route.buildQueryString(query);
+
+      window.history.replaceState(null, document.title, url);
+    }
 
     return results;
   }
--- a/js/src/forum/components/DiscussionList.js
+++ b/js/src/forum/components/DiscussionList.js
@@ -22,6 +22,20 @@ export default class DiscussionList extends Component {
     this.loading = true;
 
     /**
+     * Whether or not discussion results are loading for the previous page.
+     *
+     * @type {Boolean}
+     */
+    this.loadingPrev = false;
+
+    /**
+     * Whether or not there are previous results that can be loaded
+     *
+     * @type {boolean}
+     */
+    this.previousResults = false;
+
+    /**
      * Whether or not there are more results that can be loaded.
      *
      * @type {Boolean}
@@ -30,15 +44,38 @@ export default class DiscussionList extends Component {
 
     /**
      * Current page in discussion list
+     *
+     * @type {number}
      */
     this.page = Number(m.route.param('page')) || 1;
+
+    /**
+     * First page loaded in discussion list
+     *
+     * @type {number}
+     */
+    this.firstLoadedPage = this.page;
+
+    /**
+     * Last page loaded in discussion list
+     *
+     * @type {number}
+     */
+    this.lastLoadedPage = this.page;
+
+    /**
+     * Discussions per page
+     *
+     * @type {number}
+     */
+    this.offsetBy = 20;
 
     /**
      * Number of discussions to offset for pagination
      *
      * @type {number}
      */
-    this.offset = (this.page - 1) * 20;
+    this.offset = (this.page - 1) * this.offsetBy;
 
     /**
      * The discussions in the discussion list.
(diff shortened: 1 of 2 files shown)


js/src/forum/components/DiscussionList.js
```diff
@@ -187,8 +201,22 @@ export default class DiscussionList extends Component {
...
     this.moreResults = !!results.payload.links.next;
 
+    // Construct a URL to this discussion with the updated page, then
+    // replace it into the window's history and our own history stack.
     m.lazyRedraw();
 
+    const query = m.route.parseQueryString(document.location.search);
+
+    if (this.page !== query.page) {
+      if (this.page !== 1) query.page = this.page;
```
luceos: Couldn't we set this at the `this.page` definition already?

datitisev: @luceos What do you mean ? `this.page` keeps track of current page, and this line adds it to the query string if it isn't `1`. Do you mean have `this.page` set to undefined (i.e. unset) ?

luceos: I meant, can we use the query page earlier?

```js
this.page = Number(m.route.param('page')) || 1;
```
or is that too soon?

franzliedke:
I vote to delay this one for the next beta release.

Too much SEO voodoo that I'd like to investigate and read up upon first, after seeing flarum/framework#1820 and https://webmasters.googleblog.com/2014/02/infinite-scroll-search-friendly.html.

luceos:
I'd prefer to merge this now and revisit this before stable if we must. But I'm giving you the final say on that @franzliedke.

js/src/forum/components/DiscussionList.js
clarkwinkelmann: I'm not able to test this right now, but wouldn't scrolling up from page 2 back to page 1 keep page 2 in the querystring ?

datitisev: @clarkwinkelmann That is correct. There are no scroll handlers in the DiscussionList right now...

@luceos Hm, perhaps, haven't tested it.

datitisev:
TODO: scroll listener to update the page parameter when scrolling & previous page button in discussion (if I has time).

datitisev:
I thought adding a previous page button was more of a priority than scroll listener - as it allows the users to actually go to previous pages without changing the URL.

I had to add a bunch of properties to the discussion list, mainly
- `loadingPrev` - to show loading spinner if its loading the previous page at the top of the discussion list
- `firstPageLoaded` and `lastPageLoaded` - to load pages before & after this, not repeating already-loaded pages
- made `offsetBy` a property in case extensions want to change that easily
- `addResultsToBeginning` for when loading previous pages, the results are appended to the top

The page parameter changing on page loading does not work on IE, as it doesn't have the `URL` constructor, so I added a check for it.

I also added a check for numeric value in the PHP side to fix notices if the `page` parameter is something like `2<ab`

js/src/forum/components/DiscussionList.js
```diff
@@ -215,4 +300,17 @@ export default class DiscussionList extends Component {
   addDiscussion(discussion) {
     this.discussions.unshift(discussion);
   }
+
+  /**
+   * Get the "Load More" or "Load Previous" page buttons
+   *
+   * @param isNext
+   */
+  getLoadButton(isNext = true) {
```
franzliedke: Let's turn this into two methods to avoid the boolean parameter.

src/Forum/Content/Index.php
```diff
@@ -65,6 +65,11 @@ public function __invoke(Document $document, Request $request)
         $q = Arr::pull($queryParams, 'q');
         $page = Arr::pull($queryParams, 'page', 1);
 
+        // Fix PHP notices if $page is not numeric when calculating offset
+        if (! is_numeric($page) || $page < 1) {
+            $page = 1;
+        }
```
franzliedke: This was already done (slightly differently) in master a while ago.

less/forum/DiscussionList.less
```diff
@@ -7,7 +7,7 @@
   list-style-type: none;
   position: relative;
 }
-.DiscussionList-loadMore {
+.DiscussionList-loadMore, .DiscussionList-loadPrev {
```
franzliedke: We can reuse this generic class name for both buttons.

js/src/forum/components/DiscussionList.js
```diff
@@ -166,12 +217,20 @@ export default class DiscussionList extends Component {
   /**
    * Load the next page of discussion results.
    *
+   * @param isNext the page to load is the next page, false for previous page
    * @public
    */
-  loadMore() {
-    this.loading = true;
+  load(isNext = true) {
```
franzliedke: Same here, let's avoid the boolean parameter.

js/src/forum/components/DiscussionList.js
```diff
@@ -166,12 +217,20 @@ export default class DiscussionList extends Component {
...
+    if (isNext) {
+      this.loading = true;
+      this.page = ++this.lastLoadedPage;
+    } else if (this.firstLoadedPage !== 1) {
+      this.loadingPrev = true;
+      this.page = --this.firstLoadedPage;
+      this.addResultsToBeginning = true;
```
franzliedke: Doesn't this mean that weird things will happen when you first

I guess I am arguing not to have this value stored as state on the component, but rather as values closed over by the event handlers below.

js/src/forum/components/DiscussionList.js
```diff
@@ -73,7 +124,7 @@ export default class DiscussionList extends Component {
           })}
         </ul>
         <div className="DiscussionList-loadMore">
-          {loading}
+          {this.loading ? LoadingIndicator.component() : this.moreResults && this.getLoadButton()}
```
franzliedke: Should be `loadingNext` for consistency.

franzliedke (changes requested):
Just went through the code and added a few comments. As agreed a few weeks ago, I will apply these changes myself, to not distract you from the frontend refactoring.

franzliedke:
Okay, I took care of most of these and rebased.

What's left for solving (and discussion):
- State management is still not right, especially in regards to triggering multiple actions in parallel. See [this comment](https://github.com/flarum/core/pull/1829#discussion_r395544112).
- Now that I've seen it, I am unsure this approach makes sense without a scroll handler. When you click multiple times on "Load previous" and "Load more", which page are you expected to be "on" (reflected by the URL parameter)? I am kind of thinking it would always be the earliest one that's loaded - because the top of the page would remain the same once you refresh the page. This would mean only "Load previous" would change the URL state.
- I think we can use the links from API responses instead of building URLs ourselves - that should simplify the state on the client.

js/src/forum/components/DiscussionList.js
datitisev: I don't remember what that code does, but I guess it could cause some issues.

franzliedke:
What we agreed on in today's dev meeting:

- For consistency with scrolling inside a discussion, the page parameter in the URL should update whenever moving up *or* down.
- Refreshing shows (roughly) the same section (i.e. page) that you were looking at before the refresh.
- I will finish the outstanding changes and take care of the remaining comments / suggestions.

franzliedke:

franzliedke:
Actually, I can take care of this one.

luceos:
@franzliedke I moved this to beta 14 as it wasn't merged yet. If this is a mistake, please amend where needed.

datitisev:
This PR will need to be re-done to resolve the conflicts & use the DiscussionListState, so I'll mark it as WIP for now.

franzliedke:
To be honest, beta.14 might be unrealistic for this one. There was quite some complex stuff left to-do and that was before our refactoring around the discussion list (state).

askvortsov1:
Agreed: we can return to it within this release cycle if we somehow end up with extra time at the end, but otherwise I'm fine with pushing to beta 15.

dsevillamartin:
Closing in favor of flarum/issue-archive#162

PR #1830, opened 2019-07-29
Hidden posts visibility

### LINKED ISSUE #1827: Deleted posts are visible to all visitors
Deleted posts are visible to all visitors. 

The issue can be reproduced by disabling the Tags Extension. Writing a post, delete it and logout. The deleted post is collapsed and Grayed out, but can still be opened. 

- Flarum version: Flarum core 0.1.0-beta.9

Requires fix in the PostPolicy
https://github.com/flarum/core/blob/master/src/Post/PostPolicy.php

luceos:
**Fixes #1827**

Added a default `1=0` statement that can be bypassed with a `orWhere` from the dispatched event.

- [ ] Are tests okay.
- [ ] Is the implementation using whereRaw the right one.

**Confirmed**

**Required changes:**

-  Related core extension PRs: (Remove if irrelevant)
    - [ ] flarum/tags might need a pr

### DIFF
--- a/src/Post/PostPolicy.php
+++ b/src/Post/PostPolicy.php
@@ -99,6 +99,7 @@ class PostPolicy extends AbstractPolicy
                             ->from('discussions')
                             ->whereColumn('discussions.id', 'posts.discussion_id')
                             ->where(function ($query) use ($actor) {
+                                $query->whereRaw('1=0');
                                 $this->events->dispatch(
                                     new ScopeModelVisibility(Discussion::query()->setQuery($query), $actor, 'hidePosts')
                                 );
--- a/src/Post/PostPolicy.php
+++ b/src/Post/PostPolicy.php
@@ -99,12 +99,13 @@ class PostPolicy extends AbstractPolicy
                             ->from('discussions')
                             ->whereColumn('discussions.id', 'posts.discussion_id')
                             ->where(function ($query) use ($actor) {
-                                $query->whereRaw('1=0');
-                                $query->orWhere(function ($query) use ($actor) {
-                                    $this->events->dispatch(
-                                        new ScopeModelVisibility(Discussion::query()->setQuery($query), $actor, 'hidePosts')
-                                    );
-                                });
+                                $query
+                                    ->whereRaw('1=0')
+                                    ->orWhere(function ($query) use ($actor) {
+                                        $this->events->dispatch(
+                                            new ScopeModelVisibility(Discussion::query()->setQuery($query), $actor, 'hidePosts')
+                                        );
+                                    });
                             });
                     });
             });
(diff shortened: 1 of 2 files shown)


clarkwinkelmann:
Just by reading the code, it looks good.

I do think changes are needed in the tags extension, as wheres now become an orwhere ?

Is it in https://github.com/flarum/tags/blob/master/src/Access/DiscussionPolicy.php maybe ?

Do you want me to test locally ?

src/Post/PostPolicy.php
```diff
@@ -99,6 +99,7 @@ public function find(User $actor, $query)
                             ->from('discussions')
                             ->whereColumn('discussions.id', 'posts.discussion_id')
                             ->where(function ($query) use ($actor) {
+                                $query->whereRaw('1=0');
```
franzliedke: Could you try whether this would work as well?

~~~php
->where(function ($query) use ($actor) {
  $query->whereRaw('1=0')
    ->orWhere(function ($query) use ($actor) {
      $this->events->dispatch(...);
    });
});
~~~

Because if I understand [Laravel's implementation](https://github.com/laravel/framework/blob/3ec56b8810bc40016252977c69ba27057dcecf7d/src/Illuminate/Database/Query/Builder.php#L1361-L1372) correctly, the `orWhere()` would not be applied if none of the event listeners adds a  WHERE clause to the query.

franzliedke: The one test that registers a listener for this event could then also use `where()` instead of `orWhere()`.

luceos: Assuming you meant to replace the whereRaw, this doesn't work. If you wanted to use orWhere after the whereRaw, it doesn't add anything, it only makes the code less readable.

luceos: However setting the test to use `whereRaw` instead of `orWhereRaw` does make a difference. Committing.

tests/integration/api/Controller/ShowDiscussionControllerTest.php
```diff
@@ -78,6 +83,51 @@ public function guest_cannot_see_empty_discussion()
...
+     * @test
+     */
+    public function when_allowed_guests_can_see_hidden_posts()
+    {
+        /** @var Dispatcher $events */
+        $events = app(Dispatcher::class);
+
+        $events->listen(ScopeModelVisibility::class, function (ScopeModelVisibility $event) {
+            $event->query->whereRaw('1=1');
+        });
```
franzliedke: Shouldn't this listener check for the event's properties so that it only affects the query we want to affect?

src/Post/PostPolicy.php
```diff
@@ -99,9 +99,12 @@ public function find(User $actor, $query)
                             ->from('discussions')
                             ->whereColumn('discussions.id', 'posts.discussion_id')
                             ->where(function ($query) use ($actor) {
-                                $this->events->dispatch(
-                                    new ScopeModelVisibility(Discussion::query()->setQuery($query), $actor, 'hidePosts')
-                                );
+                                $query->whereRaw('1=0');
+                                $query->orWhere(function ($query) use ($actor) {
```
franzliedke: Could you chain this with the above call? :pray:

franzliedke:
Never mind my comment about the `DiscussionPolicy` - that one does not have the `SELECT 1` subquery.

Issue #1832, opened 2019-07-30
Applying the "reply to discussions" permission to a tag may break the "lock" mode


matteocontrini:
## Bug Report

**Current Behavior**
If you have a tag where you explicitly set that users can reply to discussions, users may be able to reply to discussions that are locked.

**Steps to Reproduce**
1. Log in as an admin
2. Create a new "restriction by tag" for any tag (X) in the Permissions admin page
3. Set the permission that allows users to reply to discussions with tag X
4. Create a new discussion with tag X and lock it
5. **Disable and re-enable the lock extension** (this is required to reproduce the issue)
6. Log in as a regular user
7. Try to reply to the created discussion in tag X and succeed

**Expected Behavior**
Users should not be able to reply to discussions that are locked (unless they have the "canLock" permission).

**Screenshots**

Admin view:

![immagine](https://user-images.githubusercontent.com/2164763/62144570-5f2b1700-b2f2-11e9-9f79-bfc9fd0d4b9a.png)

User view:

![immagine](https://user-images.githubusercontent.com/2164763/62144589-67835200-b2f2-11e9-97bb-8f33a5466791.png)

**Environment**
Reproduced on beta.flarum.site and nightly.flarum.site.

askvortsov1:
This will be fixed with the policy rewrite I'm working on right now.

matteocontrini:
Hello, is there a chance to have this fixed for the next beta? It's the only issue marked as "critical" but it's been almost a year since it's been reported... Thanks 😃

datitisev:
I think this issue is more complicated than it seems. We've discussed multiple ways on fixing it, and the policy refactor PR may fix it.

askvortsov1:
Fixed in #2461

PR #1838, opened 2019-08-07
Allow extensions to specify a load priority

CDK2020:
This would allow a simple fix to #1832 as well as other load priority problems. I have confirmed that after using these proposed changes + adding a priority to the lock extension the issue went away.

Allow extensions to specify an **optional** load "priority" in their `composer.json` under Flarum extension that dictates the order they are loaded in (higher means first)

Is this the best way to do this?

**Confirmed**

**Required changes:**

- Extensions such as lock need a priority set

### DIFF
--- a/src/Extension/ExtensionManager.php
+++ b/src/Extension/ExtensionManager.php
@@ -276,7 +276,7 @@ class ExtensionManager
 
         return $enabled->sortByDesc(function ($extension, $name) {
             return $extension->composerJsonAttribute('extra.flarum-extension.priority');
-        });;
+        });
     }
 
     /**
--- a/src/Extension/ExtensionManager.php
+++ b/src/Extension/ExtensionManager.php
@@ -21,6 +21,8 @@ use Flarum\Foundation\Application;
 use Flarum\Settings\SettingsRepositoryInterface;
 use Illuminate\Contracts\Container\Container;
 use Illuminate\Contracts\Events\Dispatcher;
+use Illuminate\Database\ConnectionInterface;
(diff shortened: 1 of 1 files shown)


src/Extension/ExtensionManager.php
```diff
@@ -283,7 +283,13 @@ public function getEnabledExtensions()
      */
     public function extend(Container $app)
     {
-        foreach ($this->getEnabledExtensions() as $extension) {
+        $extensions = new Collection($this->getEnabledExtensions());
```
luceos: I would suggest that the sorted collection comes from `getEnabledExtensions()` instead of being used here.

CDK2020: `getEnabledExtensions` is currently an array, it needs to be converted into a collection before we can sort it (unless you want an array `usort` instead)

franzliedke:
I'd like to think about this some more. If I have no better idea, we can merge it before the next beta release.

franzliedke:
I cannot say I like this very much, because it does not tackle the root cause (the tags extension overwrites the fallback permission check, lock adds an additional performance "middleware", so to speak - maybe we should differentiate these concepts?).

I can revisit this when creating an extender for the permission system.

Leaving for @luceos to review and merge.

luceos:
I dislike this approach too, it bloats the composer.json file unnecessarily. I rather go with @franzliedke proposal right now and do this properly from the get go.

Sorry, but this feels like one of those PR's that might haunt us in the long run.

PR #1848, opened 2019-08-17
Allow formatting post content without the HTTP request

### LINKED ISSUE #1840: Allow calling formatContent() on a CommentPost without passing a request
When the `Formatter` was made dependent on the current HTTP request in 0ab9facc4bd59a260575e6fc650793c663e5866a, the `$request` parameter on the `render` method on the formatter was made optional, but [it wasn't](https://github.com/flarum/core/commit/0ab9facc4bd59a260575e6fc650793c663e5866a#diff-3b6ca93f8d90cf7a0b08c9cdc12eba0eR172) on the `formatContent($request)` method in the `CommentPost` class.

As discussed [here](https://discuss.flarum.org/d/20986-contenthtml-attribute-on-post), there might be the need to generate the HTML content in contexts where the request is not available.

The proposal is to change, in the `CommentPost` class, this signature from:

```php
public function formatContent(ServerRequestInterface $request)
```

to:

```php
public function formatContent(ServerRequestInterface $request = null)
```

matteocontrini:
**Fixes #1840**

This change allows the `formatContent()` method on the `CommentPost` to be called without an HTTP request.

**Confirmed**


luceos:
As the code shows, you can also use the Formatter directly to render. That's why I'm in dubio..

clarkwinkelmann:
Indeed, it's not an absolutely needed change because it can be done without using this method.

But I think the change is so simple, it would be counter-productive to not allow re-using that method in every contexts. It makes extensions easier to develop by not having to resolve the formatter yourself. It's easier to explain how to use it, and it's also maybe easier to read code that uses the method.

PR #1850, opened 2019-08-19
Add Mailgun region setting

### LINKED ISSUE #1834: Add Mailgun region setting
## Feature Request

Mailgun introduced the Europe region some time ago. Following that Laravel added support to choose between US and EU in the mail driver via the `endpoint` setting https://laravel.com/docs/5.7/mail#mailgun-driver

We should add this setting to Flarum's Admin. It's already supported by the version of the library we use.

Needed for all Mailgun customers on the Europe region.

Users can use the Europe SMTP server in the meantime.

vingrad:
**Fixes #1834 **

Support for the new option "endpoint setting" for mailgun.

**Screenshot**

image

**Confirmed**

**Required changes:**


js/src/admin/components/MailPage.js
```diff
@@ -108,6 +105,28 @@ export default class MailPage extends Page {
...
   }
 
+    getMailSettingsElementForm(field) {
+
+    if (field === 'mail_mailgun_endpoint') {
+
+      let endpoints = {
+        'api.eu.mailgun.net': 'EU Endpoint',
+        'api.mailgun.net': 'US Endpoint',
+      };
```
franzliedke: I would prefer not to have this field-specific code here in the frontend, as the rest is so nicely generic.

Instead, we will need the API to differentiate between text fields and select boxes. (That means, the drivers will need this option in their `availableSettings()` method as well.)

js/src/admin/components/MailPage.js
```diff
@@ -108,6 +105,28 @@ export default class MailPage extends Page {
     );
   }
 
+    getMailSettingsElementForm(field) {
+
+    if (field === 'mail_mailgun_endpoint') {
+
+      let endpoints = {
+        'api.eu.mailgun.net': 'EU Endpoint',
+        'api.mailgun.net': 'US Endpoint',
```
franzliedke: These values should be translatable.

vingrad:
I did some changes. Now it's more generic.
But I did not found a good way to passthrough Translation class  to the MailgunDriver. So you will see now only hostnames for endpoints, but I think, that it's also Ok.

js/src/admin/components/MailPage.js
vingrad: Please check the current implementation.

js/src/admin/components/MailPage.js
```diff
@@ -18,7 +18,9 @@ export default class MailPage extends Page {
     this.values = {};
 
     const settings = app.data.settings;
-    this.fields.forEach(key => this.values[key] = m.prop(settings[key]));
+      console.log('settings', settings);
```
franzliedke: Please remove this line.

js/src/admin/components/MailPage.js
```diff
@@ -29,12 +31,15 @@ export default class MailPage extends Page {
...
-          this.values[key] = m.prop(settings[key]);
+
+
+    for (let driverName in this.driverFields) {
+        for (let driverOptionKey in this.driverFields[driverName]) {
+            this.fields.push(driverOptionKey);
+            this.values[driverOptionKey] = m.prop(settings[driverOptionKey]);
         }
-      );
+    }
```
franzliedke: Does this change any functionality or did you just change the loop?

src/Mail/LogDriver.php
```diff
@@ -28,9 +28,9 @@ public function __construct(LoggerInterface $logger)
         $this->logger = $logger;
     }
 
-    public function availableSettings(): array
+    public function availableSettings(): object
     {
-        return [];
+        return (object) [];
```
franzliedke: What's the benefit of using an object over an array here?

vingrad: We need an object for additional parameters like the type of the form element and also for possible values.

image

js/src/admin/components/MailPage.js
vingrad: It's just other loop, because we need to iterate over an object.

js/src/admin/components/MailPage.js
vingrad: Sorry, done.

src/Mail/LogDriver.php
franzliedke: @vingrad Can you expand on that and explain the why? That structure is perfectly valid as an array as well, and it should serialize to JSON in the same way?

franzliedke:
I've taken care of the remaining comments, simplified everything and squash-merged it into master in 4c89e2eb772779b7a1d499caa789a59515a23701.

Thanks for your efforts, @vingrad !

PR #1852, opened 2019-08-20
Fix error thrown if textarea doesn't exist in TextEditor

### LINKED ISSUE #1851: t[0] is sometimes undefined when posting reply
## Bug Report

**Current Behavior**
An error can occur when posting a reply, preventing the composer from closing. Not sure exactly why, but it can be easily fixed.

```js
TypeError: t[0] is undefined TextEditor.js:137:12
    getSelectionRange TextEditor.js:137
    O reply.js:14
    k reply.js:32
    l mithril.js:2033
    c mithril.js:2004
    l mithril.js:2024
    resolve mithril.js:1941
    then mithril.js:1962
    then mithril.js:1905
    k reply.js:32
    onclick addPostReplyAction.js:18
    L/t[e]< mithril.js:1286
```

**Environment**
- Website URL: https://discuss.flarum.org
- Browser: Firefox 68

datitisev:
**Fixes #1851**

Simply checks for `$textarea[0]` being thruthy, else it returns. Avoids errors if textarea isn't found in TextEditor.


franzliedke:
From the issue:

> Not sure exactly why

Can we please find out why this would occur first? I am not a fan of scatter if statements like this without understanding the root cause.

From what I see, the `getSelectionRange()` method is only called once, right after opening the post composer. The `TextEditor` component *always* renders a `textarea`, so why shouldn't it have one?

datitisev:
It occurred to me when posting a post. The loading spinner stayed indefinitely and the error was in the console, so I can only assume it was thrown when clicking post.

I don't really have any more information. It occurred on discuss, only once as far as I know though.

js/src/forum/components/TextEditor.js
```diff
@@ -122,6 +122,8 @@ export default class TextEditor extends Component {
   setSelectionRange(start, end) {
     const $textarea = this.$('textarea');
 
+    if (!$textarea[0]) return;
```
luceos: @datitisev wouldn't a length check make more sense?

Aside from that, feel free to merge this.

PR #1853, opened 2019-08-20
Use GitHub Actions

datitisev:
Uses https://github.com/flarum/action-build for JS building.

Extensions will need a workflow for building before merging this PR. The `build.yml` workflow from this PR should work just fine.

image


franzliedke:
Merging this, as extensions have been taken care of. Thanks a lot!

PR #1854, opened 2019-08-20
Error handling: Fix status codes

franzliedke:
**Refs #1641**

This builds on top of #1843 to make error status codes consistent.

Before my refactoring, two exceptions were handled differently by the API
layer (which let the exception handlers determine the status code) vs. the
middleware used by the frontends (which used the "code" that was set
when throwing the exceptions).

When extracting the logic, I mostly used the exception codes, which
now changed the behavior of the API, e.g. when raising a "permission
denied" error when incorrect login credentials were entered.

For the case of login, this actually meant that you would now get a rather
confusing generic error message when trying to login with an incorrect
password. This is now fixed.

- Does this make sense?
- Logic correct?
- Any cases I missed?

**Confirmed**

### DIFF
--- a/src/Api/Controller/ListNotificationsController.php
+++ b/src/Api/Controller/ListNotificationsController.php
@@ -15,12 +15,14 @@ use Flarum\Api\Serializer\NotificationSerializer;
 use Flarum\Discussion\Discussion;
 use Flarum\Http\UrlGenerator;
 use Flarum\Notification\NotificationRepository;
-use Flarum\User\Exception\PermissionDeniedException;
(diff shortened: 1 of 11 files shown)


franzliedke:
Good catch, on it!

> EDIT: Do we just want the API to use 401 error codes ? Or should a page like /admin also return 401 if not signed in ?

Right now, it returns a 403 and "An error occurred while trying to load this page." Yeah, we might want to fix that.

franzliedke:
Re. admin page: status code is fixed with this PR, the error message has been fixed by adding a line to the English language pack.

PR #1855, opened 2019-08-22
Add DB prefix to PHP tests

datitisev:
Only for PHP 7.3 & MySQL 5.7.

Also simplified job names to show it has a custom prefix.

Only way this worked was by excluding every other job :/

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -33,9 +33,6 @@ jobs:
           - php: 7.2
             service: mariadb
             prefix: flarum_
-          - php: 7.3
-            service: mariadb
-            prefix: flarum_
 
     services:
       mysql:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,7 +8,7 @@ jobs:
 
     strategy:
       matrix:
-        php: [7.1, 7.2, 7.3]
+        php: [7.1, 7.2, 7.3, 7.4]
         service: ['mysql:5.7', mariadb]
         prefix: ['', flarum_]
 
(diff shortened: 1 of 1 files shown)


luceos:
Might want to run prefix mode on mariadb too, that would remove one entry in the exclude, right?

datitisev:
@luceos Yep. Didn't do so as Travis only had the prefix with MySQL. If wanted, I can try to add PHP 7.4 snapshot tests to this PR.

luceos:
Yeah but 7.4 is allowed to fail; not sure you can set that?

datitisev:
@luceos Not necessary allowed to fail, but we can make it exit with a success status code even if it does fail.

datitisev:
Added PHP 7.4, it will always show as success in the status check, but you can see the log when clicking details.

matteocontrini:
Hi everyone, I wanted to mention that GitHub Actions can be run inside Docker containers as well, so using an image with PHP 7.4 preinstalled would be a way to avoid installing it every time.

datitisev:
@matteocontrini Thanks for the info, though I don't think it really matters that much as our tests fail on PHP 7.4 anyway, so it's just there to remind us \:).

luceos:
A test that succeeds on 7.4 even when they fail without direct feedback doesn't make a lot of sense. If looking at the logs are the only way to check actual compliance with the new version, I'd rather not have this version right now. Feel free to push this back to the github actions team, allowed to fail with clear visual feedback is a must-have.

franzliedke:
Agreed, let's avoid 7.3 for now then.

@datitisev Shouldn't it be possible to model this only with "include" and no "exclude"? :thinking:

datitisev:
@franzliedke I tried it, but `include` only seems to be for including extra information??? Not sure, I just got weird errors that didn't offer any insight into what the problem was. Something about `G15` or `G16` (?).

Ok, I'll remove PHP 7.4.

PR #1860, opened 2019-08-26
Add back defaults for language and direction attributes

### LINKED ISSUE #1772: No longer any default direction or locale
## Bug Report

**Current Behavior**
Any beta 8 ~~or dev-master~~ forum has `<html dir="" lang="">` at the top of the document.

**Steps to Reproduce**
1. Look at source HTML of any forum

**Expected Behavior**
The values should not be empty

**Environment**
- Flarum version: beta8, ~~dev-master~~

**Possible Solution**
The default values were removed in #1471

The default values were previously defined here in `FrontendView` https://github.com/flarum/core/pull/1471/files#diff-17e7b99805890a6754d57096b8dfbc9bL295 and before that it was `WebAppView` when it was added #1105

There is no longer any default value since https://github.com/flarum/core/blob/9115b9e28f9c5d1b53708090f183d71e095678ce/src/Frontend/Document.php#L164

We either need to force language packs (including english) to use those methods, or define default values again for the current language and direction as ltr.

**Additional Context**
Investigated after https://discuss.flarum.org/d/19701-w3c-xhtml-validation-service-flarum-error

datitisev:
**Fixes #1772**

- Adds back `language` and `direction` attribute defaults

- Do we want a method in the `LanguagePack` extender to set the direction ? It would simplify having to use the Frontend extender and checking for the current locale

**Confirmed**

### DIFF
--- a/src/Frontend/Content/Meta.php
+++ b/src/Frontend/Content/Meta.php
@@ -12,13 +12,30 @@
 namespace Flarum\Frontend\Content;
 
 use Flarum\Frontend\Document;
+use Flarum\Locale\LocaleManager;
 use Illuminate\Support\Arr;
 use Psr\Http\Message\ServerRequestInterface as Request;
 
 class Meta
 {
+    /**
+     * @var LocaleManager
+     */
+    private $locales;
+
+    /**
+     * @param LocaleManager $locales
+     */
+    public function __construct(LocaleManager $locales)
+    {
+        $this->locales = $locales;
+    }
+
     public function __invoke(Document $document, Request $request)
     {
+        $document->language = $this->locales->getLocale();
+        $document->direction = 'ltr';
+
         $document->meta = array_merge($document->meta, $this->buildMeta($document));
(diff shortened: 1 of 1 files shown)


luceos:
> Do we want a method in the LanguagePack extender to set the direction ? It would simplify having to use the Frontend extender and checking for the current locale

Yes but as a different PR.

src/Frontend/Document.php
```diff
@@ -162,8 +170,8 @@ protected function makeView(): View
             'title' => $this->makeTitle(),
             'payload' => $this->payload,
             'layout' => $this->makeLayout(),
-            'language' => $this->language,
-            'direction' => $this->direction,
+            'language' => $this->language ?: $this->locales->getLocale(),
+            'direction' => $this->direction ?: 'ltr',
```
Jeckerson: Why not constant for default direction?

franzliedke:
Did that for you. :wink:

PR #1868, opened 2019-09-04
Fix new post injected above unread sticky

SychO9:
**Fixes #1751**

**Changes proposed in this pull request:**
instead of prepending the new discussion to the list, this refreshes the list

**Confirmed**

- [x] Frontend changes: tested on a local Flarum installation.
- ~Backend changes: tests are green (run `php vendor/bin/phpunit`).~

PR #1881, opened 2019-09-09
Fix JS errors that only require a check for truthy value

datitisev:
This should fix a bunch of errors that have been thrown recently on our beta demo.

Fixes:
```js
TypeError: Cannot read property 'top' of undefined
  at apply(webpack://@flarum/core/./src/common/components/Dropdown.js:69:15)
```
```js
TypeError: Cannot read property 'content' of null
  at apply(webpack://@flarum/core/./src/forum/components/ReplyPlaceholder.js:55:46)

TypeError: null is not an object (evaluating 'app.composer.component.content')
  at apply(webpack://@flarum/core/./src/forum/components/ReplyPlaceholder.js:70:30)
```
```js
TypeError: undefined is not an object (evaluating '.offset().top')
  at anchorScroll(webpack://@flarum/core/./src/common/utils/anchorScroll.js:16:37)
  at onSuccess(webpack://@flarum/core/./src/forum/components/PostStream.js:115:17)
  ...
  at document(webpack://@flarum/core/./src/forum/components/PostStream.js:66:26)
```
```js
TypeError: Cannot read property 'addDiscussion' of undefined
  at onSuccess(webpack://@flarum/core/./src/forum/components/DiscussionComposer.js:95:34)
```

### DIFF
--- a/js/src/common/utils/anchorScroll.js
+++ b/js/src/common/utils/anchorScroll.js
@@ -12,10 +12,18 @@
  * @param {Function} callback The callback to run that will change page content.
  */
 export default function anchorScroll(element, callback) {
+  const $element = $(element);
+
+  // return if element doesn't exist
+  if (!$element.length) {
+    callback();
+    return;
+  }
+
   const $window = $(window);
-  const relativeScroll = $(element).offset().top - $window.scrollTop();
+  const relativeScroll = $element.offset().top - $window.scrollTop();
 
   callback();
 
-  $window.scrollTop($(element).offset().top - relativeScroll);
+  $window.scrollTop($element.offset().top - relativeScroll);
 }
--- a/js/src/common/utils/anchorScroll.js
+++ b/js/src/common/utils/anchorScroll.js
@@ -14,7 +14,8 @@
 export default function anchorScroll(element, callback) {
   const $element = $(element);
 
-  // return if element doesn't exist
+  // $element.offset() is null if element doesn't exist,
+  // so we call the callback now and return instead of scrolling.
   if (!$element.length) {
     callback();
     return;
--- a/js/src/forum/components/DiscussionComposer.js
+++ b/js/src/forum/components/DiscussionComposer.js
@@ -98,7 +98,9 @@ export default class DiscussionComposer extends ComposerBody {
       .save(data)
       .then((discussion) => {
         app.composer.hide();
-        app.cache.discussionList.refresh();
+
+        if (app.cache.discussionList) app.cache.discussionList.refresh(discussion);
+
         m.route(app.route.discussion(discussion));
       }, this.loaded.bind(this));
   }
--- a/js/src/common/components/Dropdown.js
+++ b/js/src/common/components/Dropdown.js
@@ -60,6 +60,10 @@ export default class Dropdown extends Component {
       m.redraw();
 
       const $menu = this.$('.Dropdown-menu');
+
+      // return if menu is not found
+      if (!$menu.length) return;
+
       const isRight = $menu.hasClass('Dropdown-menu--right');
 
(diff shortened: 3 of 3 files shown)


js/src/forum/components/DiscussionComposer.js
```diff
@@ -92,7 +92,9 @@ export default class DiscussionComposer extends ComposerBody {
     app.store.createRecord('discussions').save(data).then(
       discussion => {
         app.composer.hide();
-        app.cache.discussionList.addDiscussion(discussion);
+
+        if (app.cache.discussionList) app.cache.discussionList.addDiscussion(discussion);
```
franzliedke: This is likely made obsolete by #1868.

js/src/common/utils/anchorScroll.js
```diff
@@ -12,10 +12,18 @@
  * @param {Function} callback The callback to run that will change page content.
  */
 export default function anchorScroll(element, callback) {
+  const $element = $(element);
+
+  // return if element doesn't exist
```
franzliedke: These comments are stating what's pretty obvious in the code - in this case, it's not even 100% correct, because the callback is still executed.

Please explain *why* this is necessary, not *what* it does.

js/src/forum/components/DiscussionComposer.js
datitisev: That PR would still need a check anyway... I'll remove the extra lines I added so maybe git just sees it as adding the `if` before, and not changing the whole line.

js/src/common/utils/anchorScroll.js
matteocontrini: This reminds me of Clean Code, chapter 4 😄 🚀

franzliedke: @datitisev It seems like you also misunderstood me. I would love to see comments explaining *why* this is needed (because of something that happens before this code is called), not what error it prevents in that same method. That's mostly obvious from looking at the code.

What I don't see when looking at this code:
- Why would the `discussionList` ever not exist when the post composer is open?
- When would somebody pass a not-existing element into `anchorScroll`?

franzliedke: But, again, let's focus on the other PRs first.

datitisev: > Why would the discussionList ever not exist when the post composer is open?

I can only assume during redraws or something

> When would somebody pass a not-existing element into anchorScroll?

Not intentionally. If you look at the stack trace, it was because of the PostStream. From there, I assume it tried to scroll to a post (through its index/number) that hadn't been loaded yet.

There's really not much I can gather from the stack traces, and this is a pretty simple fix. This would be a temporary solution, until (or if) we find and & solve the underlying issues.

franzliedke:
I fixed one of them (the `ReplyPlaceholder` one) in master as we've discovered the root cause here is a race condition (`setInterval` yadda-yadda).

Still unsure about the other ones. I prefer fixing root causes. :wink:

franzliedke:
Please rebase. :pray:

franzliedke:

js/src/forum/components/DiscussionComposer.js
```diff
@@ -98,7 +98,9 @@ export default class DiscussionComposer extends ComposerBody {
       .save(data)
       .then((discussion) => {
         app.composer.hide();
-        app.cache.discussionList.refresh();
+
+        if (app.cache.discussionList) app.cache.discussionList.refresh(discussion);
```
askvortsov1, line 102: This line should no longer be necessary after #2150

franzliedke:
I hope this will no longer be necessary with Typescript. Or if so, it will be hopelessly conflicted.

PR #1888, opened 2019-09-20
Speed up Discussion Search FullTextGambit as it is too slow

### LINKED ISSUE #240: Slow loading searches and tags on large data sets

### LINKED ISSUE #1839: Slow loading searches and tags on large datassets
## Feature Request
Allow search and tags to work with large databases and result sets.

We have a Flarum DB with 8 million users, 2 million discussions and over 15 million threads. Tags can take up 10 seconds for us to load and search always times out. If I try running the queries Flarum produces directly, the average search takes over 100 seconds to return 21 results in the end. 

Please note that this is just a suggestion and a quick fix that I've come up with given the time constraints I have to go into production.

I believe the general slowness is tied to using search gambits and eloquent events and just in general the way eloquent makes things so flexible, but it all comes at the cost of ugly and inefficient queries when they are assembled in the end. As such, on a big dataset, the tags pages and searches become un-useable so we need a clean way to resolve this. I will paste a git patch here for what I have written to deal with the problem in src/Discussion/Search/DiscussionSearcher.php - but I am not a Laravel or Eloquent developer, so there is almost certainly a better way to fix in this in SQL. Eventually, it would be great to have an Elastic Search add-on, but bad queries probably need to be fixed anyway.

Because search is a core feature of Flarum.

None, but we have thought of other ideas like flattening the data and putting it into a view or using Elastic Search, but we think these should be step 2, once have optimised the way Flarum gets data from the database.

The patch below means that on our database, even keeping the not-so-pretty SQL produced by the whereVisibleTo() method on a model, we can bring back searches and tags in under 1 second. Tags took about 5-10 seconds before and searches over a minute. This code now also works with sticky tag searches and follows, but as you can see from the code, it is quite limiting and essentially a hack.

**Patch:**
```
diff --git a/src/Discussion/Search/DiscussionSearcher.php b/src/Discussion/Search/DiscussionSearcher.php
index d342acfc..e3e41c8b 100644
--- a/src/Discussion/Search/DiscussionSearcher.php
+++ b/src/Discussion/Search/DiscussionSearcher.php
@@ -61,19 +61,54 @@ class DiscussionSearcher
     {
         $actor = $criteria->actor;
 
+        // Check what we've got
+        $tagPosition = strpos($criteria->query, ' tag:');
+        if ($tagPosition > 0) {
+            $kw = explode(' tag:', $criteria->query);
+            $keyword = $kw[0];
+            $tag = $kw[1];
+        } else {
+            $tag = str_replace(' tag:', '', $criteria->query);
+        }
+        $isTagSearch = $tagPosition !== false;
+        $isFollowing = strpos($criteria->query, 'is:following');
+        $isByobu = strpos($criteria->query, 'byobu:') !== false;
+        $isAuthor = strpos($criteria->query, 'author:') !== false;
+        $isNormalSearch = !$isByobu && !$isTagSearch && !$isAuthor && !$isFollowing && $criteria->query !== null;
+
         $query = $this->discussions->query()->select('discussions.*')->whereVisibleTo($actor);
 
-        // Construct an object which represents this search for discussions.
-        // Apply gambits to it, sort, and paging criteria. Also give extensions
-        // an opportunity to modify it.
+        if ($isTagSearch) {
+            $query->leftJoin('discussion_tag', 'discussions.id', '=', 'discussion_tag.discussion_id')
+                  ->leftJoin('tags', 'discussion_tag.tag_id', '=', 'tags.id')
+                  ->where('tags.slug', $tag)
+                  ->orderBy('is_sticky', 'DESC');
+            if ($tagPosition > 0) {
+                // Search inside a tag
+                $query->where('discussions.title', 'like', '%' . $keyword . '%');
+            }
+
+        } elseif ($isNormalSearch) {
+            $query->where('id', 'in', function($query) {
+                $query->select('posts.discussion_id')->from('posts')->where('posts.content','like', '%' . $criteria->query . '%')->orderBy('id');
+            })
+                  ->orWhere('discussions.title', 'like', '%' . $criteria->query . '%')
+                  ->orderBy('discussions.last_posted_at', 'desc');
+        }
+
         $search = new DiscussionSearch($query->getQuery(), $actor);
 
-        $this->gambits->apply($search, $criteria->query);
-        $this->applySort($search, $criteria->sort);
+        if (!$isNormalSearch) {
+            $this->applySort($search, $criteria->sort);
+        }
+
         $this->applyOffset($search, $offset);
         $this->applyLimit($search, $limit + 1);
 
-        $this->events->dispatch(new Searching($search, $criteria));
+        if (!$isTagSearch && !$isNormalSearch) {
+            $this->gambits->apply($search, $criteria->query);
+            $this->events->dispatch(new Searching($search, $criteria));
+        }
 
         // Execute the search query and retrieve the results. We get one more
         // results than the user asked for, so that we can say if there are more

```
**New Search method in DiscussionSearcher.php:**
```
 public function search(SearchCriteria $criteria, $limit = null, $offset = 0)
    {
        $actor = $criteria->actor;

        // Check what we've got
        $tagPosition = strpos($criteria->query, ' tag:');
        if ($tagPosition > 0) {
            $kw = explode(' tag:', $criteria->query);
            $keyword = $kw[0];
            $tag = $kw[1];
        } else {
            $tag = str_replace(' tag:', '', $criteria->query);
        }
        $isTagSearch = $tagPosition !== false;
        $isFollowing = strpos($criteria->query, 'is:following');
        $isByobu = strpos($criteria->query, 'byobu:') !== false;
        $isAuthor = strpos($criteria->query, 'author:') !== false;
        $isNormalSearch = !$isByobu && !$isTagSearch && !$isAuthor && !$isFollowing && $criteria->query !== null;

        $query = $this->discussions->query()->select('discussions.*')->whereVisibleTo($actor);

        if ($isTagSearch) {
            $query->leftJoin('discussion_tag', 'discussions.id', '=', 'discussion_tag.discussion_id')
                  ->leftJoin('tags', 'discussion_tag.tag_id', '=', 'tags.id')
                  ->where('tags.slug', $tag)
                  ->orderBy('is_sticky', 'DESC');
            if ($tagPosition > 0) {
                // Search inside a tag
                $query->where('discussions.title', 'like', '%' . $keyword . '%');
            }

        } elseif ($isNormalSearch) {
            $query->where('id', 'in', function($query) {
                $query->select('posts.discussion_id')->from('posts')->where('posts.content','like', '%' . $criteria->query . '%')->orderBy('id');
            })
                  ->orWhere('discussions.title', 'like', '%' . $criteria->query . '%')
                  ->orderBy('discussions.last_posted_at', 'desc');
        }

        $search = new DiscussionSearch($query->getQuery(), $actor);

        if (!$isNormalSearch) {
            $this->applySort($search, $criteria->sort);
        }

        $this->applyOffset($search, $offset);
        $this->applyLimit($search, $limit + 1);

        if (!$isTagSearch && !$isNormalSearch) {
            $this->gambits->apply($search, $criteria->query);
            $this->events->dispatch(new Searching($search, $criteria));
        }

        // Execute the search query and retrieve the results. We get one more
        // results than the user asked for, so that we can say if there are more
        // results. If there are, we will get rid of that extra result.
        $discussions = $query->get();

        $areMoreResults = $limit > 0 && $discussions->count() > $limit;

        if ($areMoreResults) {
            $discussions->pop();
        }

        return new SearchResults($discussions, $areMoreResults);
    }
```

meezaan:
**Fixes flarum/issue-archive#240 (potentially)**

The relevance based search in the original gambit has a huge performance impact (a negative one), so it actually makes Flarum unusable on large data sets. As an example, on our database, the original query took 105s to return results. The one in this PR takes 2-3s.

**Screenshot**

**Confirmed**

- [ x] Frontend changes: tested on a local Flarum installation and in production with 20m posts.
- [ x] Backend changes: tests are green (run `composer test`).

### DIFF
--- a/src/Discussion/Search/Gambit/FulltextGambit.php
+++ b/src/Discussion/Search/Gambit/FulltextGambit.php
@@ -15,7 +15,6 @@ use Flarum\Discussion\Search\DiscussionSearch;
 use Flarum\Post\Post;
 use Flarum\Search\AbstractSearch;
 use Flarum\Search\GambitInterface;
-use Illuminate\Database\Query\Expression;
 use LogicException;
 
 class FulltextGambit implements GambitInterface
@@ -42,7 +41,7 @@ class FulltextGambit implements GambitInterface
         $subquery = Post::whereVisibleTo($search->getActor())
             ->select('posts.discussion_id')
             ->from('posts')
-            ->where('posts.content','like', '%' . $bit . '%')
+            ->where('posts.content', 'like', '%' . $bit . '%')
             ->where('posts.type', '=', 'comment')
             ->where('posts.is_private', '=', 0)
             ->orderBy('id');
@@ -50,7 +49,7 @@ class FulltextGambit implements GambitInterface
             ->where(function ($query) use ($subquery, $bit) {
                 $query
                     ->where('id', 'in', $subquery)
-                    ->orWhere('discussions.title', 'like', '%' . $bit . '%');
+                    ->orWhere('discussions.title', 'like', '%'.$bit.'%');
             })
             ->where('discussions.is_private', '=', 0)
             ->orderBy('discussions.last_posted_at', 'desc');
--- a/src/Discussion/Search/Gambit/FulltextGambit.php
+++ b/src/Discussion/Search/Gambit/FulltextGambit.php
@@ -43,29 +43,18 @@ class FulltextGambit implements GambitInterface
         // the ID of the most relevant post.
         $subquery = Post::whereVisibleTo($search->getActor())
             ->select('posts.discussion_id')
-            ->selectRaw('SUM(MATCH('.$grammar->wrap('posts.content').') AGAINST (?)) as score', [$bit])
-            ->selectRaw('SUBSTRING_INDEX(GROUP_CONCAT('.$grammar->wrap('posts.id').' ORDER BY MATCH('.$grammar->wrap('posts.content').') AGAINST (?) DESC, '.$grammar->wrap('posts.number').'), \',\', 1) as most_relevant_post_id', [$bit])
-            ->where('posts.type', 'comment')
-            ->whereRaw('MATCH('.$grammar->wrap('posts.content').') AGAINST (? IN BOOLEAN MODE)', [$bit])
-            ->groupBy('posts.discussion_id');
-
-        // Join the subquery into the main search query and scope results to
-        // discussions that have a relevant title or that contain relevant posts.
+            ->from('posts')
+            ->where('posts.content','like', '%' . $bit . '%')
+            ->where('posts.type', '=', 'comment')
+            ->where('posts.is_private', '=', 0)
+            ->orderBy('id');
         $query
-            ->addSelect('posts_ft.most_relevant_post_id')
-            ->leftJoin(
-                new Expression('('.$subquery->toSql().') '.$grammar->wrapTable('posts_ft')),
-                'posts_ft.discussion_id', '=', 'discussions.id'
-            )
-            ->addBinding($subquery->getBindings(), 'join')
-            ->where(function ($query) use ($grammar, $bit) {
-                $query->whereRaw('MATCH('.$grammar->wrap('discussions.title').') AGAINST (? IN BOOLEAN MODE)', [$bit])
-                      ->orWhereNotNull('posts_ft.score');
-            });
-
-        $search->setDefaultSort(function ($query) use ($grammar, $bit) {
-            $query->orderByRaw('MATCH('.$grammar->wrap('discussions.title').') AGAINST (?) desc', [$bit]);
-            $query->orderBy('posts_ft.score', 'desc');
-        });
+            ->where(function ($query) use ($subquery, $bit) {
+                $query
+                    ->where('id', 'in', $subquery)
+                    ->orWhere('discussions.title', 'like', '%' . $bit . '%');
+            })
+            ->where('discussions.is_private', '=', 0)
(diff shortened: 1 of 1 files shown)


meezaan:
So the unit tests will obviously fail because the query is different.

franzliedke:
Have you experimented with removing other, smaller parts of the query, e.g. the `IN BOOLEAN MODE`? I would expect a fulltext index to still perform better than a `LIKE` query.

franzliedke:
@meezaan Also, can you experiment with reverting flarum/framework#1741 and then amending it so that [this test](https://github.com/flarum/core/commit/26c3bcdb74f78d6ad6b46f19ec410e0219059aee) still passes?

meezaan:
@franzliedke Thanks for the feedback. I will try and find the time to do this, but just FYI I we are working on an Elastic Search extension and that is my priority (because I will be moving on shortly). As it stands, the results we are getting with this SQL are good enough - but other parts slow down with the large data - the Following page, loading large tags, etc. and and there's no text search there - it's the late order by clause that kills us, and what we really want to do is have enough WHERE clauses to reduce the number of results to order. Adding @tariqwalji here who will be taking over from me.

meezaan:
I will close this - this problem has likely been addressed using Elastic Search? @tariqwalji?

luceos:
We settled on reverting the change that included the discussion subject and will investigate alternative solutions like a delegated search driver (perhaps something similar to Laravel Scout) before (or soon after) stable.

meezaan:
Thanks @luceos.

PR #1890, opened 2019-09-21
Avatar Canvas CORS problem

### LINKED ISSUE #445: Avatar Canvas CORS problem
The [calculateAvatarColor](https://github.com/flarum/core/blob/master/js/lib/models/User.js#L85) adds a nice identity part for users but it leaves a few things to be desired. Right now there is a CORS issue when hot-linking images from external sources. There are two ways of tackling this from what I've found:
1. Don't use HTML Canvas together with external URIs.
2. Allow tainted canvases - not widely adopted by browsers

Since hot-linking isn't a native function in the core I can see why this wouldn't be high priority but the default [User.color](https://github.com/flarum/core/blob/master/js/lib/models/User.js#L34) uses the memoization function `computed` making it a bit ugly to override (and still fall back to calling the original function); and thus I purpose a change in the default method implementation to make monkey patching easier for extensions. It is possible to work around but it breaks the monkey patching solution to some extent.

More info: https://developer.mozilla.org/en-US/docs/Web/HTML/CORS_enabled_image

JCKodel:
**Fixes #445**

I have SSO implemented in my forum, so all avatar urls are remote (from my Azure storage).

When an user card was shown, there was an error in JS by Color Thief (same as issue #445), the canvas would stuck inside body and the user card would be grey.

Following the instructions on https://lokeshdhakar.com/projects/color-thief/, I added a `image.crossOrigin = 'Anonymous';` that fixes my issue.

There is just one add to include the crossOrigin attribute.

**Screenshot**

image

For example, one of those avatars are in https://codeartbr.blob.core.windows.net/user-avatars/d3wNdIniabWR8mLdMQO6yzYriLw2.jpg 

**Confirmed**

### DIFF
--- a/js/src/common/models/User.js
+++ b/js/src/common/models/User.js
@@ -91,7 +91,7 @@ Object.assign(User.prototype, {
       user.freshness = new Date();
       m.redraw();
     };
-    image.crossOrigin = 'Anonymous';
+    image.crossOrigin = 'anonymous';
     image.src = this.avatarUrl();
   },
 


franzliedke:
Thanks for your PR for this long-standing issue!

Have you tested this locally with a CDN setup? I am confused this works because the [documentation](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/crossOrigin) only lists the value as lower-case `anonymous`. But apparently that does not make a difference? Can we switch to the value from the documentation in case not every browser is as generous?

luceos:
@franzliedke I was considering an alternative solution where 'on-upload' or change we will read the avatar and do what color thief does to save that into a dedicated users.bg_color column using php.

I merged this using the proposed recommended string value.

datitisev:
Looks like the color thief documentation had `Anonymous` capitalized, which is porobably why this PR had it like that. Most likely doesn't matter, but yeah lets keep it lowercase.

image

---

If we do decide to move this to the backend at some point (which might be a good idea, not sure how much of a performance impact there is with color thief), that would be something good for a queue. :+1:

PR #1894, opened 2019-09-27
Update window size

### LINKED ISSUE #1727: [Facebook] Login popup window should be 580px wide instead of 600px
## Bug Report

**Current Behavior**
The login window shows a warning to Facebook app developers about the use of the popup mode.

**Steps to Reproduce**
1. Go to a forum with the flarum/auth-facebook extension enabled
2. Try to login/signup with a Facebook account that is a developer for the Facebook application
4. See the warning at the bottom

**Screenshots**
immagine

**Possible Solution**
According to [this](https://stackoverflow.com/questions/16542490/facebook-auth-dialog-developer-warning-concerning-the-use-of-display-type-po) StackOverflow question, the issue is related to the popup window size. Although I can't find the reference in the Facebook docs, they say that the popup window should be no more than 580x400 pixels.

Currently, the `LoginButton` component creates a window of 600x400 pixels.

https://github.com/flarum/core/blob/167059027e5a066d618599c90164ef1b5a509148/js/src/forum/components/LogInButton.js#L15-L26

I think that changing the window size to 580x400 would fix this problem, and it should not have any other unintended consequence with other providers.

niladam:
This small change attempts to fix #1727, as i just got my eye on it and figured i could simply fix it while seeing it.

**Fixes #1727**

Modifies the file `js/src/forum/components/LogInButton.js` by changing the window width from `600` to `580`

Not a really big change that requires any focus, however since i don't have a flarum installation at hand, i cannot test it. I think testing is in order.

**Screenshot**
Not tested.

**Confirmed**

### DIFF
--- a/js/src/forum/components/LogInButton.js
+++ b/js/src/forum/components/LogInButton.js
@@ -13,7 +13,7 @@ export default class LogInButton extends Button {
     props.className = (props.className || '') + ' LogInButton';
 
     props.onclick = function() {
-      const width = 600;
+      const width = 580;
       const height = 400;
       const $window = $(window);
 


franzliedke:
Thanks for the PR!

I cannot reproduce the underlying problem anymore, see [the issue](https://github.com/flarum/core/issues/1727#issuecomment-537241791). Can you?

Issue #1734, opened 2019-01-25
Increase username flexibility/privacy with mentions and user profile


luceos:
## Feature Request

**Is your feature request related to a problem? Please describe.**

When you have a forum that tries to protect usernames (as they are used for logging in too) by adding the display_name (eg I've migrated that column onto the users table), the profile url and mentioning logic should be able to use that column instead of the username column.

**Describe the solution you'd like**
I'm pretty blocked right now and would like your input for the best solution:

- Dispatch events on those specific controllers to allow "finding" the user.
- Dispatch an event or allow configuration of the field to use for users (this is most likely prone to error).

**Justify why this feature belongs in Flarum's core, rather than in a third-party extension**

Ease of configuration and protecting user privacy. This might also relate to the PR Toby did #1721 (see point two in https://github.com/flarum/core/pull/1721#issuecomment-451306374)

**Describe alternatives you've considered**

🤷‍♀️

luceos:
@tobscure you might have an opinion on this as you might have already ran into this with that client, maybe I missed something that core allows here.

tobscure:
My feeling is that usernames should always be considered public, they are like a friendly "user ID". We could add an option to only allow logging in via email address for enhanced security?

luceos:
Doesn't that complicate the idea of display- vs username? Isn't displayname supposed to be public, not the other way around? Using email as sole authorization might not be the best bet either, communities might allow multiple users with the same mail address (role playing boards for instance).

tobscure:
Not necessarily, the distinction is not intended to be public vs. private but just machine-friendly (eg. Toby001) vs human-friendly (eg. a full name like Toby Zerner)

> Using email as sole authorization might not be the best bet either, communities might allow multiple users with the same mail address (role playing boards for instance).

I don't know if we should allow this, people can always sign up multiple accounts by using `+` in their email address (eg. `example+account1@example.com`). It's pretty commonplace that emails are unique for accounts across the internet

luceos:
I think we differ in opinion where we talk about "we should allow this", in my opinion I would formulate this "we should allow this by default". 

Display name really feels like something publicly usable and visible, whereas username is something that could/should be private (by default).

tobscure:
[...]

Issue #1356, opened 2018-01-31
Usernames cannot be numeric


luceos:
Currently username can be any matching `regex:/^[a-z0-9_-]+$/i` and the other rules (eg, min: 3). A drawback of this regex is that it allows for numeric usernames. As a consequence consider:

User
- id: 1337 username: Toby
- id: 1338 username: 1337

Hitting
/api/user/1337 returns user 1337 for any other user.

What we have to do is disallow any numeric usernames, the UserValidator has to be modified. A consideration is what to do with existing users.

Zeokat:
In my opinion, numeric user names are allowed on most of software and disallow them is not the way.

Maybe, when someone build an extension to integrate login with service "A" and this service allow numeric usernames, what will happen? The problem here, is not the numeric username is how API works.

tobscure:
Open to discussing alternatives, I think the main question is what should the API call be to retrieve a single user by their username?

I guess it could be `GET /users?filter[username]=Toby`, although that endpoint returns a list (collection) of one users rather than a just a single resource.

Zeokat:
Would not it be enough to obtain user details by its id?

tobscure:
Not always. Prime example is when you browse to a user page like http://discuss.flarum.org/u/Toby - we need to look up the user by their username!

pflstr:
Right now, we are generating duplicate content with http://discuss.flarum.org/u/Toby and http://discuss.flarum.org/u/1. Shouldn't one of them redirect to the other one or at least a canonical URL direct search engines to the version to be indexed?

Zeokat:
Now my question changes to: Wouldn't be enough to obtain user details by its username?
(Thanks for your patience)

tobscure:
Sometimes you need to look up a user by their ID too. eg. when the API returns discussions/posts, they have relationships with their authors which specify a user ID.

clarkwinkelmann:
There could be a separate endpoint to lookup by username ? like `/api/users/by-username/{value}` ? There's no real need for the same endpoint to work with both IDs and usernames ?

Or tweak the existing endpoint to accept the username query with a different syntax. For example if you query `/api/users/username`, it acts like the GET/PUT/DELETE endpoints (or just GET) except you also have to pass a `username` / `filter[username]` GET/POST attribute to tell which username you look for. As long as it doesn't collide with the way you pass new values to PUT/DELETE it would be fine. With this solution you just have to blacklist a single token that will switch the endpoint from id-based to username-based.

[...]

PR #1901, opened 2019-10-01
Add Content for User page, preload user & throw 404 accordingly

### LINKED ISSUE #1846: Invalid/Not existed username handle in API ShowUserController
## Bug Report

**Current Behavior**
spinning icon with error popup when go to url with invalid/not existed usename

**Steps to Reproduce**
1. Go to https://discuss.flarum.org/u/notexisteduser
2. See error

**Expected Behavior**
A user not found handler, not a spinning icon with error popup

**Environment**
- Flarum version: beta 0.9

**Possible Solution**
```
    protected function data(ServerRequestInterface $request, Document $document)
    {
        $id = Arr::get($request->getQueryParams(), 'id');
        if (! is_numeric($id)) {
            $id = $this->users->getIdForUsername($id);
        }
        if ($actor->id == $id) { 
//should be === when $actor->id = 0 and $id = null
            $this->serializer = CurrentUserSerializer::class;
        }
        return $this->users->findOrFail($id, $actor);
 //-> need better handler when actor is guest and $id = null, save some DB queries
    }
```

dsevillamartin:
**Fixes flarum/framework#1846**

- Preload user
  - Tested it works correctly, API call to /api/users/<username> no longer needed as `preloadedApiDocument()` call adds it to the forum payload
- Return 404 if user is not found

Doesn't add a no-JS view. I assume we want one, not sure about what items to include though. Posts/discussions would be hard, and require backend handling of those routes as well.

**Screenshot**
image

**Confirmed**

### DIFF
--- a/js/src/forum/components/UserPage.js
+++ b/js/src/forum/components/UserPage.js
@@ -94,7 +94,7 @@ export default class UserPage extends Page {
     app.preloadedApiDocument();
 
     app.store.all('users').some(user => {
-      if (user.username().toLowerCase() === lowercaseUsername && user.joinTime()) {
+      if ((user.username().toLowerCase() === lowercaseUsername || user.id() === username) && user.joinTime()) {
         this.show(user);
         return true;
       }
--- a/js/src/forum/components/UserPage.js
+++ b/js/src/forum/components/UserPage.js
@@ -88,6 +88,9 @@ export default class UserPage extends Page {
   loadUser(username) {
     const lowercaseUsername = username.toLowerCase();
 
+    // Load the preloaded user object, if any, into the global app store
+    // We don't use the output of the method because it returns raw JSON
+    // instead of the parsed models
     app.preloadedApiDocument();
 
     app.store.all('users').some(user => {
--- a/js/src/forum/components/UserPage.js
+++ b/js/src/forum/components/UserPage.js
@@ -88,6 +88,8 @@ export default class UserPage extends Page {
   loadUser(username) {
     const lowercaseUsername = username.toLowerCase();
 
+    app.preloadedApiDocument();
+
     app.store.all('users').some(user => {
       if (user.username().toLowerCase() === lowercaseUsername && user.joinTime()) {
         this.show(user);
--- a/src/Forum/Content/User.php
+++ b/src/Forum/Content/User.php
@@ -3,10 +3,8 @@
 /*
  * This file is part of Flarum.
  *
- * (c) Toby Zerner <toby.zerner@gmail.com>
- *
- * For the full copyright and license information, please view the LICENSE
- * file that was distributed with this source code.
+ * For detailed copyright and license information, please view the
+ * LICENSE file that was distributed with this source code.
  */
 
 namespace Flarum\Forum\Content;
--- a/src/Forum/Content/User.php
+++ b/src/Forum/Content/User.php
@@ -12,9 +12,9 @@ namespace Flarum\Forum\Content;
 use Flarum\Api\Client;
 use Flarum\Api\Controller\ShowUserController;
 use Flarum\Frontend\Document;
-use Flarum\Http\Exception\RouteNotFoundException;
 use Flarum\Http\UrlGenerator;
 use Flarum\User\User as FlarumUser;
+use Illuminate\Database\Eloquent\ModelNotFoundException;
 use Illuminate\Support\Arr;
 use Psr\Http\Message\ServerRequestInterface as Request;
 
@@ -66,7 +66,7 @@ class User
      * @param FlarumUser $actor
      * @param array $params
      * @return object
-     * @throws RouteNotFoundException
+     * @throws ModelNotFoundException
      */
     protected function getApiDocument(FlarumUser $actor, array $params)
     {
@@ -74,7 +74,7 @@ class User
         $statusCode = $response->getStatusCode();
 
         if ($statusCode === 404) {
-            throw new RouteNotFoundException;
+            throw new ModelNotFoundException;
         }
 
         return json_decode($response->getBody());
--- a/src/Forum/Content/User.php
+++ b/src/Forum/Content/User.php
@@ -1,0 +1,82 @@
+<?php
+
+/*
+ * This file is part of Flarum.
+ *
+ * (c) Toby Zerner <toby.zerner@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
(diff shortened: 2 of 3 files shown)


src/Forum/Content/User.php
```diff
@@ -0,0 +1,82 @@
...
+
+        $document->title = $apiDocument->data->attributes->title;
+        $document->canonicalUrl = $this->url->to('forum')->route('user', ['username' => $userId]);
+        $document->payload['apiDocument'] = $apiDocument;
+
+        return $document;
+    }
+
+    /**
+     * Get the result of an API request to show a discussion.
```
franzliedke: Huh?

src/Forum/Content/User.php
```diff
@@ -0,0 +1,82 @@
...
+     *
+     * @param \Flarum\User\User $actor
+     * @param array $params
+     * @return object
+     * @throws RouteNotFoundException
+     */
+    protected function getApiDocument(\Flarum\User\User $actor, array $params)
```
franzliedke: Import, please.

js/src/forum/components/UserPage.js
```diff
@@ -88,6 +88,8 @@ export default class UserPage extends Page {
   loadUser(username) {
     const lowercaseUsername = username.toLowerCase();
 
+    app.preloadedApiDocument();
```
franzliedke: Why are we not using the result of this method?

Because we only want to support client-side navigation and can not rely on the user getting here directly from a server-rendered page?

src/Forum/Content/User.php
```diff
@@ -0,0 +1,82 @@
...
+
+    public function __invoke(Document $document, Request $request)
+    {
+        $queryParams = $request->getQueryParams();
+        $actor = $request->getAttribute('actor');
+        $userId = Arr::get($queryParams, 'username');
+
+        $params = [
+            'id' => $userId,
+        ];
```
franzliedke: This doesn't seem to be the same lookup logic as in the client-side variant...

franzliedke (changes requested):

src/Forum/Content/User.php
datitisev: Copy paste leftover 🙈

js/src/forum/components/UserPage.js
datitisev: The result of this method simply returns the `apiDocument` passed through by the server.

I'm not using it as calling `preloadedApiDocument` loads the objects and relationships into the global `app.store`, so the `.find` code below can find new users. I tried using the object it returned, but it was just JSON and not model classes, so the rest of the code wouldn't work with it.

> Because we only want to support client-side navigation and can not rely on the user getting here directly from a server-rendered page?
It returns `null` if it has already been called. But that doesn't matter in this case because of my explanation above.

src/Forum/Content/User.php
datitisev: The API uses the `id` query param, while the [oute uses `username`.

https://github.com/flarum/core/blob/2f656146a7a47192ce664c3a889349777d92e467/src/Api/Controller/ShowUserController.php#L49-L55

https://github.com/flarum/core/blob/6c96c932e0b6fd5bdf292013f23f4ff590a679bc/src/Forum/routes.php#L30-L33

franzliedke: Thanks for the explanation, and the second edition on the phone. :laughing:  This makes sense!

src/Forum/Content/User.php
```diff
@@ -0,0 +1,83 @@
...
+
+        $apiDocument = $this->getApiDocument($actor, $params);
+
+        $document->title = $apiDocument->data->attributes->title;
+        $document->canonicalUrl = $this->url->to('forum')->route('user', ['username' => $userId]);
```
franzliedke: Is that really the URL we want to be canonical?. At least in the frontend, we're always building the URL with the username.

That said, this has lead to problems (#1356), but we probably want to fix them by using a unique *slug* instead: https://github.com/flarum/core/issues/1734#issuecomment-570601659. So, using the username would be the closest to that desired reality for now, IMO.

src/Forum/Content/User.php
```diff
@@ -0,0 +1,83 @@
...
+        $actor = $request->getAttribute('actor');
+        $userId = Arr::get($queryParams, 'username');
+
+        $params = [
+            'id' => $userId,
+        ];
+
+        $apiDocument = $this->getApiDocument($actor, $params);
+
+        $document->title = $apiDocument->data->attributes->title;
```
franzliedke: I don't think a user has a `title` attribute. What does this show?

js/src/forum/components/UserPage.js
franzliedke: Ah, the method name is confusing in that case, but I don't yet have a better name yet.

We definitely want a comment here, though, to explain that we are relying on this side effect to happen.

src/Forum/Content/User.php
datitisev: Uh... I have no idea, I'll see what it does.

src/Forum/Content/User.php
datitisev: So instead of using the user-provider ID/username (as either works) we want to stick to the username? Ok, will do.

js/src/forum/components/UserPage.js
datitisev: @franzliedke Will do, no problem.

src/Forum/Content/User.php
datitisev: Done, I've changed it to `displayName` instead. I assume that's what we'd want to show.

src/Forum/Content/User.php
datitisev: Done. Changed it to `username`.

js/src/forum/components/UserPage.js
datitisev: Comment added.

datitisev:
Do we want to throw `ModelNotFoundException` instead of `RouteNotFoundException`? I feel like the current implementation could be confusing, as the users route exists but the user model specified doesn't. And the API endpoint throws the `ModelNotFoundException`, so we should probably be consistent there.

franzliedke:
Agreed. The model is not found, so we should use that exception.

datitisev:
@franzliedke Switched exception thrown :+1:

franzliedke (approved):
Beautiful, thank you!

PR #1906, opened 2019-10-08
feat: re-add debug button/modal

### LINKED ISSUE #1687: Re-add debug button in AJAX errors
Removed in https://github.com/flarum/core/commit/64686ef7a97a17dccd70fd9eb99240ab97d5ef45 for Beta 8. 

However, this is a much easier method of debugging for support requests, and does not force users to learn a new UI (web/dev tools in browser) or parse which errors in log files are relevant immediately. 

I suggest restoring it for future versions.

pReya:
**Fixes #1687**

Reintroduce the debug button in error modals, which was removed in https://github.com/flarum/core/commit/64686ef7a97a17dccd70fd9eb99240ab97d5ef45

**Screenshot**

Screenshot 2019-10-08 20 29 57

**Confirmed**

**Required changes:**
No tests or documentation changes needed (I guess?)

### DIFF
--- a/js/src/common/Application.js
+++ b/js/src/common/Application.js
@@ -330,12 +330,14 @@ export default class Application {
           children = app.translator.trans('core.lib.error.generic_message');
       }
 
+      const isDebug = app.forum.attribute('debug');
+
       error.alert = new Alert({
         type: 'error',
         children,
-        controls: app.forum.attribute('debug') ? [
+        controls: isDebug && [
           <Button className="Button Button--link" onclick={this.showDebug.bind(this, error)}>Debug</Button>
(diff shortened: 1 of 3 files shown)


datitisev:
I think, even though we didn't include it in the issue comments, that we also wanted to log the error (`console.error` the URL, params (?), and response) so it can be viewed in the console after the error as well, and after the error message is dismissed.

I didn't look at your implementation, but is it basically the same as how it was before?

pReya:
Yes, it is exactly the same old implementation – no code changes. I will take a look into the console debug output next week and update this PR accordingly.

PR #1910, opened 2019-10-12
Add discussion_id foreign key to posts table

### LINKED ISSUE #1909: Posts are not permanently deleted when permanently deleting discussions
## Bug Report

**Current Behavior**
When one permanently deletes a discussion, their posts remain in the DB but cannot be seen.

This can & has caused issues with flarum/flags and other extensions that get the post's discussion without checking for it first.

**Steps to Reproduce**
1. Create a discussion with posts
2. Permanently delete discussion
3. Look at DB, posts remain

**Expected Behavior**
The discussion's posts should also be permanently deleted.

**Possible Solution**
Most likely add an `onDelete` clause to the foreign key (or create a foreign key if it doesn't exist)

datitisev:
**Fixes #1909**

Create foreign key for `discussion_id`, cascade on delete.

**Confirmed**

### DIFF
--- a/migrations/2019_10_12_195349_change_posts_add_discussion_foreign_key.php
+++ b/migrations/2019_10_12_195349_change_posts_add_discussion_foreign_key.php
@@ -1,0 +1,28 @@
+<?php
+
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Schema\Builder;
+
+return [
+    'up' => function (Builder $schema) {
+        // Delete rows with non-existent entities so that we will be able to create
+        // foreign keys without any issues.
+        $connection = $schema->getConnection();
+        $connection->table('posts')
+            ->orWhereNotExists(function ($query) {
+                $query->selectRaw(1)->from('discussions')->whereColumn('id', 'discussion_id');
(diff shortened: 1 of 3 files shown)


datitisev:
Looks like some tests create posts without them having a discussion, so they now fail as an existing discussion is required... 🤔

datitisev:
I was able to fix the failing tests. It just required reordering the table inserts, but for it to work I had to remove the `first_post_id` value as the discussion needs to be created first.

This can be fixed by using `prepareDatabase` and not caring about the foreign key constraints while we add the discussions & posts...

Do we want a test in the `DeleteDiscussionControllerTest` that checks that the post is deleted (checking foreign key `on delete` works)?

luceos:
> Do we want a test in the DeleteDiscussionControllerTest that checks that the post is deleted (checking foreign key on delete works)?

I'd say yes.

datitisev:
I have done a test thingy, not sure if this is how I should've done it because there aren't many tests that aren't simple GET requests...

Seems to work, `find` returns `null` because the post no longer exists in `posts`.

PR #1917, opened 2019-10-24
Modal isDismissible() should protect the modal from closing by pressing Esc

Taraflex:
- js/src/common/components/ModalManager.js

By default Bootstrap Modal does not prevent modal from closing when you press Esc, even if backdrop: 'static' is enabled. keyboard: false required.

**Confirmed**

### DIFF
--- a/js/src/common/components/ModalManager.js
+++ b/js/src/common/components/ModalManager.js
@@ -53,7 +53,12 @@ export default class ModalManager extends Component {
 
     m.redraw(true);
 
-    this.$().modal({backdrop: this.component.isDismissible() ? true : 'static'}).modal('show');
+    var dismissible = !!this.component.isDismissible();
+    this.$().modal({
+      backdrop: dismissible || 'static',
+      keyboard: dismissible
+    }).modal('show');
+
     this.onready();
   }
 


js/src/common/components/ModalManager.js
```diff
@@ -53,7 +53,12 @@ export default class ModalManager extends Component {
 
     m.redraw(true);
 
-    this.$().modal({backdrop: this.component.isDismissible() ? true : 'static'}).modal('show');
+    var dismissible = !!this.component.isDismissible();
```
luceos: Using const or let changes the scope of the declared variable over var.

```suggestion
    const dismissible = !!this.component.isDismissible();
```

datitisev: Not sure why this suggestion was resolved and not applied...?

datitisev:
This makes sense to me. If it shouldn't be dismissible there's no reason why it should be with the escape key.

davwheat:
This needs merging.

As @datitisev says, if you want to allow a developer to mark a modal as not dismissible, why can the user still dismiss it?

It's either this or rename the method to "hasCloseButton()" or similar.

franzliedke:
This change makes sense to me as well, but I cannot actually reproduce closing a non-dismissable modal with the keyboard. (Tried this out with the "Please Wait" modal when you press "Clear Cache" on the admin dashboard.)

@Taraflex Could you explain where you encountered this problem, please? :pray:

franzliedke:
Okay, strange. When I make the login modal non-dismissible, this has the desired effect. Must have something to do with focus management.

Moving forward with this, thanks @Taraflex !

franzliedke:
Applied in https://github.com/flarum/core/commit/fd4c0d30d8af429c7b58008aa1ef15873c81942b. Thanks!

PR #1921, opened 2019-10-27
Improve handling of incomplete mail configuration

### LINKED ISSUE #1763: Mail: Handle incomplete email configuration
**Continues #1169**

- [x] Show a warning message when a driver is selected but not fully configured (i.e. missing values for required fields).
- [x] Fall back to a `NullDriver` that does not send any mails when that is the case.

This needs an extension to the `DriverInterface` - something along the lines of `getRequiredFields()` or, more flexibly, `isFullyConfigured($settings)`.

**Open question:**
Do we want to allow selecting the "log" driver explicitly? What's the value in that?

datitisev:
**Fixes #1763**

- Create `NullTransport`
- Add `requiredFields` method to `DriverInterface`
- Warning to JS side if required fields aren't filled
- `*` required indicator next to required fields

- Not sure if the code on the JS is good... the whole file is a mess 😖.
- Do we want to use `requiredFields` for the name? And what about drivers where all fields are required? I simply made a call to the `availableSettings` method because it was pointless copy/pasting it.
- How's the logic for deciding to use NullTransport? Seems to work just fine right now, the code isn't very pretty though.
- Is the translation key ok?

**Screenshot**

**Confirmed**

**Required changes:**

- [ ] Related core extension PRs
  - flarum/english ~~PR required~~ - Franz has a commit waiting on his machine


luceos:
The whole implementation is .. scary. Makes me think of that Backup logic I've shown you @datitisev ; can't we port that? So a `setting => validation` mapping and parsed into the frontend for easier rendering? Very much out of scope, but worth considering before stable?

datitisev:
@luceos Hm yeah, perhaps that'd be better. It'd allow for frontend validation rules as well, for example. But isn't that basically more doable with the changes from #1850 then?

datitisev:
Ok, so I attempted adding validation to the mail system. I don't like how I hooked the PHP side of things together, but it works as of right now. Please do improve it if you see some glaring issues with the implementation 🙂.

I also fixed `flattenMap` not working in IE, so I added a flatten function from https://www.30secondsofcode.org/snippet/deepFlatten/ (they don't require attribution, see https://www.30secondsofcode.org/about).

The weird regular expression is for the hostname, not sure if we want it.

image

src/Forum/ValidateMailConfiguration.php
```diff
@@ -0,0 +1,89 @@
...
+        $this->getValidator($driver, $event->settings)->validate();
+    }
+
+    public function getWorkingDriver()
+    {
+        $settings = $this->settings->all();
+        $driver = $this->getDriver($settings);
+        $validator = $this->getValidator($driver, $settings);
+
+        return$validator->passes()
```
luceos:
```suggestion
        return $validator->passes()
```

How .. is this even working 🙈

Also `getWorkingDriver` does not implicate what this method exactly does. I think it returns the configured driver if validated, otherwise return the fallback. Personally I think this could all be part of `getDriver` instead, but not sure whether that method would explode..

datitisev: Not sure why it worked nor why StyleCI did not check my commits 🤔 

`getDriver` returns an instance of the driver set through the settings if it exists, else it returns the `NullDriver`

`getWorkingDriver` checks for validation of that configured driver and returns it if the validation passes, else it returns the `NullDriver`.

I didn't like the implementation either.

src/Forum/ValidateMailConfiguration.php
```diff
@@ -0,0 +1,91 @@
...
+/*
+ * This file is part of Flarum.
+ *
+ * (c) Toby Zerner <toby.zerner@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Flarum\Forum;
```
franzliedke: I would expect this (and its setup) to be in `Flarum\Mail`?

src/Forum/ValidateMailConfiguration.php
franzliedke: To be honest, the purpose of the whole class is a bit unclear to me. It seems to be both a driver factory and a validator of some sort, that is instantiated when creating drivers (e.g. for sending) and when storing settings.

I guess I could see more clearly if the validation logic would go to each driver implementation. Essentially a method that you can ask whether the current configuration is valid, and which exact problems exist if it isn't.

franzliedke:
Okay... I finished this up. It's not 100% perfect in terms of UX (e.g. server-side validation only), but a good step forward.
@datitisev Thanks for your initial implementation!

This is what it looks like now:
Screenshot of mail settings with errors

@jordanjay29 @luceos @datitisev Please let me know what you think.

---

@gwillem will be happy about this, I believe! :smiley:

gwillem:
Woot! I am HAPPY! :laughing: :tada: :champagne:

franzliedke:
Now, I also added a backwards-compatibility layer, just like [I did](https://github.com/flarum/core/commit/1e7fbf1ed9d8f936356fba900d09b61deb3073f6) after introducing the mail driver field types in PR #1850.

Both will be removed after releasing beta.12.

datitisev:
@franzliedke Looks good to me, I can see that it's cleaner :+1:.

PR #1922, opened 2019-10-28
Pass event to KeyboardNavigatable whenCallback

clarkwinkelmann:
This is something I noticed while working on Kilowhat Mailing. I wanted to set an `onSelect` callback, but still allow TABing out of the field when nothing was typed. Because `preventDefault()` is already called before the `onSelect` callback when we set it, the only place I could prevent the event from being prevented (you follow?) was inside the `whenCallback`. But in the `whenCallback` it wasn't possible to know the key that's been pressed.

So I just added a new parameter to the whenCallback. Existing implementations didn't read any parameter so it won't break anything, and new implementations will have the benefit of being able to access the event.

An example implementation can be seen here: https://github.com/kilowhat/flarum-ext-mailing/blob/69fffe46b2b0ea2b07b89acd43d4687e780e6bb4/js/src/forum/components/EmailUserModal.js#L39 Mailing is using its own "FixedKeyboardNavigatable" that just extends the original, changing the same method as I do in this PR.

I don't see how this could break any existing stuff. I don't even know if any extension (other than Mailing) is using the `KeyboardNavigatable` utility.

**Screenshot**
No visible changes.

**Confirmed**

### DIFF
--- a/js/src/forum/utils/KeyboardNavigatable.js
+++ b/js/src/forum/utils/KeyboardNavigatable.js
@@ -7,10 +7,24 @@
  */
 export default class KeyboardNavigatable {
   constructor() {
+    /**
+     * Callback to be executed for a specified input.
+     *
+     * @callback KeyboardNavigatable~keyCallback
+     * @param {KeyboardEvent} event
+     * @returns {boolean}
+     */
     this.callbacks = {};
 
-    // By default, always handle keyboard navigation.
-    this.whenCallback = () => true;
+    /**
+     * Callback that determines whether keyboard input should be handled.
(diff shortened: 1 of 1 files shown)


luceos:
@clarkwinkelmann feel free to push that when comment patch or have us tackle that. Awaiting your response.

clarkwinkelmann:
Will do soon! The comment was still waiting in my inbox.

clarkwinkelmann:
I have added the documentation. I also documented the other callbacks while I was at it.

I think that's fully JSdoc compliant https://jsdoc.app/tags-callback.html PHPStorm seems to be happy at least.

PR #1924, opened 2019-10-29
Fix avatar files not being deleted

### LINKED ISSUE #1918: Avatar files not deleted
## Bug Report

**Description**
When a user uploads a new avatar or removes an old one, only the variable in the database changes, but the file itself (in `/public/assets/avatars`) is not deleted.

**Why the mistake?**
This variable has already [been changed](https://github.com/flarum/core/blob/master/migrations/2018_01_30_222900_change_users_rename_columns.php#L22) before, but not in [this file](https://github.com/flarum/core/blob/master/src/User/AvatarUploader.php#L49) and it returns `null`. Therefore, it omits `AvatarUploader.php:52` and the file is not deleted.

**Possible solution**
You can change a variable in [`AvatarUploader.php:49`](https://github.com/flarum/core/blob/master/src/User/AvatarUploader.php#L49) to `$user->avatar_url`, but then the entire url will be returned.

clarkwinkelmann:
**Fixes #1918**

**Confirmed**

### DIFF
--- a/tests/unit/User/AvatarUploaderTest.php
+++ b/tests/unit/User/AvatarUploaderTest.php
@@ -1,0 +1,89 @@
+<?php
+
+/*
+ * This file is part of Flarum.
+ *
+ * (c) Toby Zerner <toby.zerner@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Flarum\Tests\unit\User;
+
+use Flarum\User\AvatarUploader;
+use Flarum\User\User;
+use Illuminate\Contracts\Events\Dispatcher;
+use Illuminate\Database\Eloquent\Model;
+use Intervention\Image\ImageManagerStatic;
+use League\Flysystem\FilesystemInterface;
+use Mockery as m;
+use Mockery\Adapter\Phpunit\MockeryPHPUnitIntegration;
+use PHPUnit\Framework\TestCase;
+
+class AvatarUploaderTest extends TestCase
+{
+    use MockeryPHPUnitIntegration;
+
+    private $dispatcher;
+    private $filesystem;
+    private $uploader;
+
+    public function setUp()
+    {
+        $this->dispatcher = m::mock(Dispatcher::class);
+        $this->dispatcher->shouldIgnoreMissing();
+        Model::setEventDispatcher($this->dispatcher);
+
+        $this->filesystem = m::mock(FilesystemInterface::class);
+        $this->uploader = new AvatarUploader($this->filesystem);
+    }
+
+    public function test_removing_avatar_removes_file()
+    {
+        $this->filesystem->shouldReceive('has')->with('ABCDEFGHabcdefgh.png');
+        $this->filesystem->shouldReceive('delete')->with('ABCDEFGHabcdefgh.png')->andReturn(true);
+
+        $user = new User();
+        $user->changeAvatarPath('ABCDEFGHabcdefgh.png');
+        // Necessary because AvatarUpload looks into the original attributes for the raw value,
+        // which isn't usually automatically updated without saving to the database
+        $user->syncOriginal();
+
+        $this->uploader->remove($user);
+        $user->syncOriginal();
+
+        $this->assertEquals(null, $user->getOriginal('avatar_url'));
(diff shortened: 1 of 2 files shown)


franzliedke:
Bleh, this whole feature is a mess (method / mutator names, retrieving the URL generator using `app()`, ...). And now we're pulling internal state out of the model...

But it fixes the problem, so thanks. :+1: 

Can I ask you to add a unit test for this method? Since we have an injected `Filesystem` instance, this should actually be fairly simple to test...

franzliedke:
ping @clarkwinkelmann

clarkwinkelmann:
I have not been able to take a look yet :grimacing:

luceos:
@clarkwinkelmann please consider finishing this so we can pick this up for the release 🙇

clarkwinkelmann:
Sorry didn't manage to do this earlier. I have taken a look, but am not sure how to proceed.

Do we want a unit test just for the `AvatarUploader`, or one integration test via the API to change avatar ?

There's no fake FilesystemInterface built in, so either we need another library for tests or use the actual filesystem ? What do we prefer?

franzliedke:
I think I was hoping for a unit test specifically for this class / method. We depend on Mockery, so you can use that instead of creating a custom class for stubbing.

clarkwinkelmann:
So I have written some tests. They pass, but they don't actually work correctly.

If I remove the `shouldReceive()` assertions, they still pass :thinking: :thinking: 

I noticed the others tests already using Mockery are actually missing `Mockery::close()` which tests for unused assertions for example. It seems the official way is to use the `MockeryPHPUnitIntegration` trait or extend their own test case http://docs.mockery.io/en/latest/reference/phpunit_integration.html

But whatever way I include Mockery, the `filesystem` mocks don't make any difference and the tests always pass :man_shrugging: 

If anyone could take a look at my tests and figure out why Mockery isn't working as expected that would be cool :innocent:

tests/unit/User/AvatarUploaderTest.php
```diff
@@ -0,0 +1,89 @@
...
+use Illuminate\Database\Eloquent\Model;
+use Intervention\Image\ImageManagerStatic;
+use League\Flysystem\FilesystemInterface;
+use Mockery as m;
+use Mockery\Adapter\Phpunit\MockeryPHPUnitIntegration;
+use PHPUnit\Framework\TestCase;
+
+class AvatarUploaderTest extends TestCase
+{
+    use MockeryPHPUnitIntegration;
```
franzliedke: Thanks for researching this. I just [configured checks and included this for all unit tests on the `master` branch](https://github.com/flarum/core/commit/4f1adba38726f76ca15ad0cd43e7a84cc708d8a8) - now you just need to extend from the `Flarum\Tests\unit\TestCase` base class.

tests/unit/User/AvatarUploaderTest.php
```diff
@@ -0,0 +1,89 @@
...
+        Model::setEventDispatcher($this->dispatcher);
+
+        $this->filesystem = m::mock(FilesystemInterface::class);
+        $this->uploader = new AvatarUploader($this->filesystem);
+    }
+
+    public function test_removing_avatar_removes_file()
+    {
+        $this->filesystem->shouldReceive('has')->with('ABCDEFGHabcdefgh.png');
+        $this->filesystem->shouldReceive('delete')->with('ABCDEFGHabcdefgh.png')->andReturn(true);
```
franzliedke: I don't know if this explains the issues you're having, but that return value is probably meant to belong to the `has()` mock above?

clarkwinkelmann: Indded. I made a mistake here. At first I had added `andReturn` for each of them but then realized it was only really useful for `has`. And I removed the wrong one here.

I can try pulling the new master tomorrow afternoon or evening. If someone wants to tackle it before that time, feel free to do.

clarkwinkelmann:
This time everything is working. I was missing the count expectation on the mockery calls. And also the saving callbacks weren't called.

Now everything seems good. All unit tests are green.

I have one failing integration test locally but it doesn't seem related to this and CI doesn't complain so it's probably something with my database being invalid.

Someone can check and merge if it's not too late ?

PR #1925, opened 2019-10-29
Allow creating access tokens without passing a CSRF token

### LINKED ISSUE #1905: Can't create user token via API without CSRF
## Bug Report

**Current Behavior**
POSTing to `/api/token` with user credentials in order to create a user token is not possible without CSRF or an Api Key.

**Steps to Reproduce**
1. Perform the request with correct (or invalid) credentials. For example via `curl -d "anything=anything" -X POST https://beta.flarum.site/api/token`
2. `{"errors":[{"status":"400","code":"csrf_token_mismatch"}]}`

**Expected Behavior**
This API endpoint can be useful for mobile apps or other client-side tools that want to authenticate the user and get a token to then access CSRF-protected API endpoints. To enable that, this first endpoint would need to not require the CSRF token.

**Environment**
- Flarum version: beta9, beta10

**Possible Solution**
We should whitelist this route. And maybe have a way to let extensions whitelist other routes as we have seen it's difficult (https://discuss.flarum.org/d/21421-handling-requests-without-csrf-tokens)

If the issue described at that link can not be made easier together with this, maybe it's worth its own issue.

**Additional Context**
first discussed here https://discuss.flarum.org/d/20867-csrf-token-mismatch-on-beta-9-upgrade/6 but I didn't open an issue at the time.

brought up again here https://discuss.flarum.org/d/21581-admin-page-to-generate-api-key

clarkwinkelmann:
Also allows extension to register their own CSRF ignored paths

**Fixes #1905**

Fixes the linked issue and makes it possible for extension to register their own ignored paths.

Do we want to expose the middleware itself through a singleton or should we create a new separate class to contain the ignored paths ?

**Confirmed**

### DIFF
--- a/src/Api/ApiServiceProvider.php
+++ b/src/Api/ApiServiceProvider.php
@@ -47,6 +47,14 @@ class ApiServiceProvider extends AbstractServiceProvider
             return $routes;
         });
 
+        $this->app->singleton('flarum.api.csrf-middleware', function () {
+            $checkCsrf = $this->app->make(HttpMiddleware\CheckCsrfToken::class);
+
+            $checkCsrf->ignorePath('/token');
+
+            return $checkCsrf;
+        });
+
         $this->app->singleton('flarum.api.middleware', function (Application $app) {
             $pipe = new MiddlewarePipe;
 
@@ -62,7 +70,7 @@ class ApiServiceProvider extends AbstractServiceProvider
             $pipe->pipe($app->make(HttpMiddleware\RememberFromCookie::class));
             $pipe->pipe($app->make(HttpMiddleware\AuthenticateWithSession::class));
             $pipe->pipe($app->make(HttpMiddleware\AuthenticateWithHeader::class));
-            $pipe->pipe($app->make(HttpMiddleware\CheckCsrfToken::class));
+            $pipe->pipe($app->make('flarum.api.csrf-middleware'));
             $pipe->pipe($app->make(HttpMiddleware\SetLocale::class));
 
             event(new ConfigureMiddleware($pipe, 'api'));
--- a/src/Admin/AdminServiceProvider.php
+++ b/src/Admin/AdminServiceProvider.php
@@ -51,6 +51,8 @@ class AdminServiceProvider extends AbstractServiceProvider
             return $routes;
         });
 
+        $this->app->singleton('flarum.admin.csrf-middleware', HttpMiddleware\CheckCsrfToken::class);
+
         $this->app->singleton('flarum.admin.middleware', function (Application $app) {
             $pipe = new MiddlewarePipe;
 
@@ -65,7 +67,7 @@ class AdminServiceProvider extends AbstractServiceProvider
             $pipe->pipe($app->make(HttpMiddleware\StartSession::class));
             $pipe->pipe($app->make(HttpMiddleware\RememberFromCookie::class));
             $pipe->pipe($app->make(HttpMiddleware\AuthenticateWithSession::class));
-            $pipe->pipe($app->make(HttpMiddleware\CheckCsrfToken::class));
+            $pipe->pipe($app->make('flarum.admin.csrf-middleware'));
             $pipe->pipe($app->make(HttpMiddleware\SetLocale::class));
             $pipe->pipe($app->make(Middleware\RequireAdministrateAbility::class));
 
--- a/src/Forum/ForumServiceProvider.php
+++ b/src/Forum/ForumServiceProvider.php
@@ -61,6 +61,8 @@ class ForumServiceProvider extends AbstractServiceProvider
             $this->setDefaultRoute($routes);
         });
 
+        $this->app->singleton('flarum.forum.csrf-middleware', HttpMiddleware\CheckCsrfToken::class);
+
         $this->app->singleton('flarum.forum.middleware', function (Application $app) {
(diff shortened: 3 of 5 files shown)


tobyzerner:
My initial thought: Why is the CSRF token checked at all for any API route? REST APIs are meant to be stateless, CSRF token is a stateful concept.

(Also, hi 👋)

clarkwinkelmann:
Hi Toby :wave: 

Well... we allow cookie auth for the API (that's our primary way of storing session), so we have to check CSRF.

But actually... if no cookie session is active (no cookie was passed), we could skip the CSRF check completely. Thoughts on that ? :thinking:

datitisev:
That makes sense... CSRF check fails without a cookie anyway, which kinda makes the API a bit pointless if you still need an CSRF token.

Not sure if we discussed this before and decided to stick with CSRF tokens for all API calls, though.

tobyzerner:
> But actually... if no cookie session is active (no cookie was passed), we could skip the CSRF check completely.

Exactly!

cmcjacob:
> > But actually... if no cookie session is active (no cookie was passed), we could skip the CSRF check completely.
> 
> Exactly!

I'm just a spectator, so don't mind me.  But I think you're a little out of place here, questioning practices based on the code you wrote.  Maybe it's best you try to fix your fuckups after people have had their round at them.

cmcjacob:
> > But actually... if no cookie session is active (no cookie was passed), we could skip the CSRF check completely.
> 
> Exactly!

Someone needs to say it:  you left a pile of spaghetti code, and expected it to be some corporate takeover? Fuck off.

franzliedke:
I tend to agree with Toby. [So does this StackOverflow question / answer](https://security.stackexchange.com/questions/168230/csrf-protection-for-json-api-with-cookie-auth).

Technically speaking, even cookie-authenticated API access shouldn't be a problem, because normal HTTP forms (which would transfer cookie information) can not submit JSON payloads. AJAX requests could, but they wouldn't have the cookies in a cross-site scenario.

Am I missing something?

clarkwinkelmann:
@franzliedke I think all our API endpoints accept classic POST data as well. But the only place we use that feature is when sending image data.

I actually tested the `/token` endpoint locally by sending POST data via `curl -d`, which worked fine.

So if we wanted to drop CSRF from those endpoints completely, we'd have to block the ability to pass classic POST data and only allow JSON. But we would still be left with the avatar endpoints and similar others that have to use classic POST.

I can start another PR where we only check CSRF when acting on cookie-based session as discussed above.

luceos:
From what I can see, the CreateTokenController is now able to generate a token for any user, even those that haven't confirmed their email account. Opening this up as such would potentially increase the number of spam accounts.

I have no objection against those csrf checks, but perhaps we can allow them on specific endpoints only for validated users if no csrf is available?

clarkwinkelmann:
What would be the problem ? You're still limited by the same permissions. If a non-confirmed user creates a token, they will get permission denied on all actions until they validate their account via email.

I'm in favor of simply dropping CSRF checks for all non-cookie requests as discussed above, so we could close this and I'll create another PR.

luceos:
> Non activated users can already create tokens, they just need to send the CSRF token like other users.

Okay than I have no object against dropping those CSRF check either.

src/Api/ApiServiceProvider.php
```diff
@@ -47,6 +47,14 @@ public function register()
             return $routes;
         });
 
+        $this->app->singleton('flarum.api.csrf-middleware', function () {
+            $checkCsrf = $this->app->make(HttpMiddleware\CheckCsrfToken::class);
+
+            $checkCsrf->ignorePath('/token');
+
+            return $checkCsrf;
+        });
```
luceos: Now that you are binding those middlewares explicitly per frontend, we could make your code shorter by setting the ignorePath method when the binding is resolved like so:

```php
        $this->app->resolving('flarum.api.csrf-middleware', function (HttpMiddleware\CheckCsrfToken $check) {
            $checkCsrf->ignorePath('/token');
        });
```

Untested +)

clarkwinkelmann:
I'm closing this PR, I'll create another PR dropping CSRF for non-cookie requests.

Now... do we still need a way for extensions to register excluded paths with CSRF ? I've seen a single person asking for that on the forum, and I suspect it wasn't the right solution for them. I can't think of any situation where it makes sense and doesn't create a security issue. So it's probably best to not even allow it. Which means completely abandoning this PR.

andreasjacobsen93:
We actually depend on registering excluded paths from CSRF check, on a return URL from a custom sign-on provider. So I'm a little disappointed to see the PR closed without merge ;)

clarkwinkelmann:
@andreasjacobsen93 isn't the return url a GET ?

andreasjacobsen93:
No, it's a POST request and we don't have a CSRF token since we come from the provider. We have other means to verify the user based on metadata from the provider, so it should not be a problem to bypass the CSRF check on that specific route.

We tried registering a middleware that set the CSRF bypass value, but it seems it couldn't be registered before the CSRF check.

So it would be nice to be able to register a specific route to be ignored by CSRF check.

luceos:
@andreasjacobsen93 can you please create a new issue and reference this PR. I have a feeling this might be lost otherwise and that's not something we (both) want.

PR #1927, opened 2019-11-03
Visually distinguish post from self and others

### LINKED ISSUE #169: Visually distinguish posts that you authored
Somehow

tankerkiller125:
**Relates to #169**

This pull request adds an indication as to whether a post is by the user currently browsing the discussion, creating a visual difference between other users posts.

- Code Quality
- Not sure if the LESS changes are in correct location in file
- Not sure if theirs a better way in JS to detect current user vs the post user

**Screenshot**

image

**Confirmed**

### DIFF
--- a/js/src/forum/components/Post.js
+++ b/js/src/forum/components/Post.js
@@ -39,7 +39,7 @@ export default class Post extends Component {
   view() {
     const attrs = this.attrs();
 
-    attrs.className = 'Post ' + (this.loading ? 'Post--loading ' : '') + (attrs.className || '');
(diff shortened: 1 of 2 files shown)


luceos:
I'm not sure what the others think of this, but can we drop the LESS part of this for now and just merge setting the element class? We can then continue talking about the details of implementation.

tankerkiller125:
@luceos I've removed the LESS changes

luceos:
🎉 well done, going to do some minor changes and add another class as well.

luceos:
Here's my commit:

https://github.com/flarum/core/commit/1f2566c32c5a3466ad55eec4026804ced83c6eff

Decided to name the class slightly easier to read. Wanted to add another class that would be enabled for posts from the OP, but we don't have access to `this.props.post.discussion().user()` as the relation isn't included.

PR #1929, opened 2019-11-09
Some improvements to request error handling and modal error formatting

datitisev:
- Only show error(s) detail in debug modal, if found
- Log the request URL, method, status, and response
  - Only logs error(s) detail if found, else it logs the whole error object
  - Logs with a group so the error data can be hidden

The alert appearing was causing the modal to "close" but the background overlay remained... not sure if that's a bug or intended (?) so I closed the currently opened modal - though I don't think want to do that.

**Screenshot**
<details>
<summary>POST /api/flags 404</summary>

image
image
image
</details>
<details>
<summary>POST /flags 404</summary>

image
image
</details>

**Confirmed**

### DIFF
--- a/js/src/common/Application.js
+++ b/js/src/common/Application.js
@@ -324,12 +324,15 @@ export default class Application {
         }
 
         const isDebug = app.forum.attribute('debug');
+        // contains a formatted errors if possible, response must be an JSON API array of errors
+        // the details property is decoded to transform escaped characters such as '\n'
+        const formattedError = error.response && Array.isArray(error.response.errors) && error.response.errors.map((e) => decodeURI(e.detail));
 
         error.alert = new Alert({
           type: 'error',
           children,
           controls: isDebug && [
-            <Button className="Button Button--link" onclick={this.showDebug.bind(this, error)}>
+            <Button className="Button Button--link" onclick={this.showDebug.bind(this, error, formattedError)}>
               Debug
             </Button>,
           ],
@@ -338,6 +341,17 @@ export default class Application {
         try {
           options.errorHandler(error);
         } catch (error) {
+          if (isDebug && error.xhr) {
+            const { method, url } = error.options;
+            const { status = '' } = error.xhr;
+
+            console.group(`${method} ${url} ${status}`);
+
+            console.error(...(formattedError || [error]));
+
+            console.groupEnd();
+          }
+
           this.alerts.show(error.alert);
         }
 
@@ -350,12 +364,13 @@ export default class Application {
 
   /**
    * @param {RequestError} error
+   * @param {string[]} [formattedError]
    * @private
    */
-  showDebug(error) {
+  showDebug(error, formattedError) {
     this.alerts.dismiss(this.requestError.alert);
 
-    this.modal.show(new RequestErrorModal({ error }));
+    this.modal.show(new RequestErrorModal({ error, formattedError }));
   }
 
   /**
--- a/js/src/common/components/RequestErrorModal.js
+++ b/js/src/common/components/RequestErrorModal.js
@@ -6,16 +6,26 @@ export default class RequestErrorModal extends Modal {
   }
 
   title() {
-    return this.props.error.xhr ? this.props.error.xhr.status + ' ' + this.props.error.xhr.statusText : '';
+    return this.props.error.xhr ? `${this.props.error.xhr.status} ${this.props.error.xhr.statusText}` : '';
   }
 
   content() {
+    const { error, formattedError } = this.props;
+
     let responseText;
 
-    try {
-      responseText = JSON.stringify(JSON.parse(this.props.error.responseText), null, 2);
-    } catch (e) {
-      responseText = this.props.error.responseText;
+    // If the error is already formatted, just add line endings;
+    // else try to parse it as JSON and stringify it with indentation
+    if (formattedError) {
+      responseText = formattedError.join('\n\n');
+    } else {
+      try {
+        const json = error.response || JSON.parse(error.responseText);
+
+        responseText = JSON.stringify(json, null, 2);
+      } catch (e) {
+        responseText = error.responseText;
+      }
     }
 
     return (


luceos:
Won't closing the currently active Modal cause unwanted side effects? I can imagine this would happen on log in, register etc; even so.. can't we make the error be pushed on top of the overlay somehow?

Ps, I also assume that changing these outputs will break all extensions returning custom error output?

datitisev:
@luceos 

> Won't closing the currently active Modal cause unwanted side effects?

Perhaps. The problem isn't the error alert being behind the modal, is that showing the alert results in the modal fade out animation to trigger, but the backdrop remains. Might be a separate issue though.

> Ps, I also assume that changing these outputs will break all extensions returning custom error output?

I don't understand what you are saying here. The `formattedError` variable is only used for the error logging and for showing it in the `RequestErrorModal` - what extensions would be using it?

luceos:
Ah your answer clarifies it. Then I have nothing to add ;)

luceos:
Asked for second review from @clarkwinkelmann ; if approved feel free to merge.

datitisev:
@clarkwinkelmann Escape is required to transform `\n` into actual line breaks.

I changed the function call to `decodeURI` and removed the closing of the modal - not sure why, but I can no longer reproduce what I first had happen.

clarkwinkelmann:
Those are great cases for inline comments I think :smile: 

If newlines are the only reason to use it, wouldn't `.replace()` work just as well ?

If we keep it like it's now I suggest adding

> Escape is required to transform \n into actual line breaks

as a comment near `decodeURI` so we know why it was used in the future.

Otherwise approved :+1:

js/src/common/Application.js
```diff
@@ -331,19 +331,31 @@ export default class Application {
...
-          <Button className="Button Button--link" onclick={this.showDebug.bind(this, error)}>Debug</Button>
+          <Button className="Button Button--link" onclick={this.showDebug.bind(this, error, formattedError)}>Debug</Button>
         ]
       });
 
       try {
         options.errorHandler(error);
       } catch (error) {
+        if (isDebug) {
+          console.group(`${error.options.method} ${error.options.url} ${error.xhr ? error.xhr.status : ''}`);
```
franzliedke: Sorry, but we need to verify whether this works in IE11. :stuck_out_tongue: 

[MDN](https://developer.mozilla.org/en-US/docs/Web/API/Console/group#Browser_compatibility) makes me think it does, but this comment [on caniuse](https://caniuse.com/#feat=mdn-api_console) isn't specific enough (Does this affect all functions? If not, which?) so that I am still worried:

> Only functions when developer tools are opened. Otherwise, the 'console' object is undefined.

js/src/common/Application.js
```diff
@@ -331,19 +331,31 @@ export default class Application {
...
+
+          console.error(...formattedError || [error]);
```
franzliedke: Is `error` inside an array because the `...` operator is evaluated last?

In other words, is this `...(formattedError || [error])` or `(...formattedError) || [error]`?

js/src/common/Application.js
```diff
@@ -331,19 +331,31 @@ export default class Application {
...
         children,
         controls: isDebug && [
-          <Button className="Button Button--link" onclick={this.showDebug.bind(this, error)}>Debug</Button>
+          <Button className="Button Button--link" onclick={this.showDebug.bind(this, error, formattedError)}>Debug</Button>
         ]
       });
 
       try {
         options.errorHandler(error);
       } catch (error) {
```
franzliedke: Woah, can we rename this `error` variable in the `catch` block? I find it super confusing to follow along which object we're dealing with here. :grin: 

---

:clock1: Three minutes later...

---

Okay, I see now, the default `errorHandler` implementation just re-throws the same object, which means we have an AJAX error, which makes sense.

Could we change this to check for the existence of `options.errorHandler()` instead? The `try`/`catch` only confuses the reader IMO.

js/src/common/Application.js
datitisev: I don't see any errors in console in IE if I don't have the developer tools open - granted, IE doesn't show anything from before you open them. But it didn't break anything, so 🤷‍♂.

The `console` object is definitely not undefined when loading the page though - even if the developer tools haven't been opened. Logged through Custom Header script:

image

js/src/common/Application.js
datitisev: After some quick testing, this looks to be `...(formattedError || [error])`. Right now, `(...formattedError) || error` doesn't to seem valid (thinks it's an arrow function).

image

js/src/common/Application.js
datitisev: Done. Had to remove the default `errorHandler` implementationa as it's not needed. 

I noticed that when the error handler is set - e.g. modals using `this.onerror` - nothing logs to the console. I assume we want to log the error to the console regardless if the error handler is set?

js/src/common/components/RequestErrorModal.js
```diff
@@ -7,17 +7,25 @@ export default class RequestErrorModal extends Modal {
...
   content() {
+    const { error, formattedError } = this.props;
+
     let responseText;
 
-    try {
-      responseText = JSON.stringify(JSON.parse(this.props.error.responseText), null, 2);
-    } catch (e) {
-      responseText = this.props.error.responseText;
+    if (!formattedError) {
```
franzliedke: When there's an `else`, there's no need for a negated `if` condition - this avoids dealing with double negatives.

js/src/common/Application.js
franzliedke: Hmm, no, I think we should reserve the logging to unexpected errors - exactly the way you did it.

Really, it depends on whether we're dealing with a known or unknown error, but we have no mechanism for that yet (in custom error handlers).

js/src/common/components/RequestErrorModal.js
datitisev: True, makes sense. Will fix soon™.

clarkwinkelmann:
I have still not had time to test this locally, but from reading the code:

Aren't we losing the feature of the `errorHandler` function not handling a function ?

I'm actually using this feature in my Formulaire extension. If the errors are not validation errors for my fields, I re-throw the error, making Flarum render it in the error alert.

This PR appears to remove that feature. Now if `errorHandler` throws an exception, it would fail silently and not show any error alert.

I suppose any custom `errorHandler` could call `app.alerts.show(error.alert);`, but that was a nice feature, and would also show up if the error was unintentionally re-thrown from the `errorHandler`.

clarkwinkelmann:
Would it make sense to log all errors to console, even if `options.errorHandler` is defined ?

We're already going to log most validation errors to the console with the current implementation I believe. We might as well log all request errors.

This would also simplify re-implementing the feature I said was removed in my previous comment as the conditions can be moved up in the code.

datitisev:
You make a good point. I guess there's no reason *not* to log e.g. validation errors - it can make debugging easier even when it's just a handled error.

The catching of `errorHandler` might be a mistake on my part - if we agree, I can add the try/catch back and always log the error.

franzliedke:

datitisev:
I rebased & squashed as I realized I hadn't done it correctly.

I also restored the `errorHandler` default of throwing the error, and moved back to `try/catch` instead of `if/else`.

clarkwinkelmann:
Still looking good to me

askvortsov1:
This doesn't clash with any of the state changes afaik, and we might as well put it in now instead of after TS when more work will be needed to get it updated. As 3 core devs have approved and there seems to be no conflicts, merging.

Issue #1920, opened 2019-10-27
Registration issue


WelaurS:
## Bug Report

**Current Behavior**
If an error occurred while trying to register, the user is created despite the error

**Steps to Reproduce**
1. Install forum, but don't setup mail settings.
2. Try to register user
3. You will got error: https://i.imgur.com/Gi19irD.jpg

**Expected Behavior**
If the error occurred the user account won't created.

**Environment**
- Flarum version: 0.1.0-beta
- Website URL: -
- Webserver: nginx
- Hosting environment: VPS
- PHP version: PHP 7.2.19-0ubuntu0.18.04.2
- Browser: Chrome 78.0.3904.70
--

```
Flarum core 0.1.0-beta.10
PHP version: 7.2.19-0ubuntu0.18.04.2
Loaded extensions: Core, date, libxml, openssl, pcre, zlib, filter, hash, pcntl, Reflection, SPL, sodium, session, standard, mysqlnd, PDO, xml, bz2, calendar, ctype, curl, dom, mbstring, fileinfo, ftp, gd, gettext, iconv, json, exif, mysqli, pdo_mysql, Phar, posix, readline, shmop, SimpleXML, sockets, sysvmsg, sysvsem, sysvshm, tokenizer, wddx, xmlreader, xmlwriter, xsl, zip, Zend OPcache
+----------------------+----------------+--------+
| Flarum Extensions    |                |        |
+----------------------+----------------+--------+
| ID                   | Version        | Commit |
+----------------------+----------------+--------+
| flarum-approval      | v0.1.0-beta.8  |        |
| flarum-bbcode        | v0.1.0-beta.8  |        |
| flarum-emoji         | v0.1.0-beta.10 |        |
| flarum-lang-english  | v0.1.0-beta.10 |        |
| flarum-flags         | v0.1.0-beta.9  |        |
| flarum-likes         | v0.1.0-beta.9  |        |
| flarum-lock          | v0.1.0-beta.9  |        |
| flarum-markdown      | v0.1.0-beta.10 |        |
| flarum-mentions      | v0.1.0-beta.10 |        |
| flarum-statistics    | v0.1.0-beta.9  |        |
| flarum-sticky        | v0.1.0-beta.9  |        |
| flarum-subscriptions | v0.1.0-beta.9  |        |
| flarum-suspend       | v0.1.0-beta.9  |        |
| flarum-tags          | v0.1.0-beta.10 |        |
+----------------------+----------------+--------+
[...]

PR #1931, opened 2019-11-13
Notifications into the queue

### LINKED ISSUE #1869: [Subscriptions] Queue "new post" email
**Relates to #978.**

A nice first use-case for playing around with our new queueing support (#1773).

1. Send `new_post` email via queue (which uses the `sync` driver by default, so this should not change behavior).
2. Document the setup for real queue workers using e.g. Redis, and try it out on discuss.

Once this is implemented, we can start showcasing this to extension developers, and tackle other tasks in the system that could benefit from being offloaded to the queue.

luceos:
Forces notifications into a dedicated SendNotificationsJob and passed to the queue
One static method re-used in the job ::getAttributes, is that okay or use a trait?
Do we want to use this solution and refactor into a better Hub after stable, postpone this implementation or use it in b11?

**Fixes #1869**

Forces all notifications into the queue, not just those from subscriptions. Guaranteeing a more streamlined user experience.

- [ ] Static method call from NotificationSyncer is reused on the SendNotificationsJob, the `::getAttribute`. Is this okay or do we want to use a trait?
- [ ] Do we want to use this solution and plan out a better notification hub for after stable?
- [ ] This implementation assumes all notification channels are only async, this implementation works well with that assumption.

I decided against trying to introduce Laravel notifications channels, because that would bloat our app and increase the weight of this implementation too much.

**Confirmed**

**Required changes:**


datitisev:
>  Do we want to use this solution and plan out a better notification hub for after stable?

Probably, yeah.

> This implementation assumes all notification channels are only async

What about the `sync` queue? We can't assume notification channels are async if there's only ne other queue implementation available for Flarum right now - and it's not even an official one

What problems does this implementation have with the `sync` queue? As far as I can tell while skimming the code, it should work fine with it.

luceos:
This implementation is unrelated to the queue driver directly. What I considered is that if any of the broadcast channel require or expect immediate user interaction, they will fail. I agree there should be no issue running these notifications with sync up until a certain number of recipients is reached and a specific amount of channels is configured. The longer processing the notifications takes the more likely the XHR will time out and the user receives no feedback on whatever they were actually doing..

src/Notification/NotificationSyncer.php
```diff
@@ -215,14 +168,16 @@ protected function setDeleted(array $ids, $isDeleted)
...
+
     /**
      * Construct an array of attributes to be stored in a notification record in
      * the database, given a notification blueprint.
      *
      * @param \Flarum\Notification\Blueprint\BlueprintInterface $blueprint
      * @return array
      */
-    protected function getAttributes(Blueprint\BlueprintInterface $blueprint)
+    public static function getAttributes(Blueprint\BlueprintInterface $blueprint)
```
franzliedke: Since this method only operates on the internals of the `$blueprint` object anyway, I am thinking we could move it to the `BlueprintInterface` and properly encapsulate the data. A quick code search showed me that this is the only place where the three methods `getFromUser()`, `getSubject()` and `getData()` are used, so this tells me they are a goodcandidate to be condensed into one. Would certainly allow for more customized behavior per blueprint in the future.

The question would be how to do this in a backwards-compatible way. My suggestion:
- Add the `getAttributes()` method to each blueprint *implementation* right now. Those will probably need far less conditions, because they already know their attributes.
- Here, check for the method to exist and fall back to the current implementation (backwards compatibility for non-official extensions).
- *Only in the next release*, replace the three small methods with the new `getAttributes()` method on the interface.

That would also answer your question about re-using this static method. :wink:

src/Notification/Job/SendNotificationsJob.php
```diff
@@ -0,0 +1,66 @@
...
+     */
+    private $recipientIds;
+
+    public function __construct(BlueprintInterface $blueprint, array $recipientIds = [])
+    {
+        $this->blueprint = $blueprint;
+        $this->recipientIds = $recipientIds;
+    }
+
+    public function handle(NotificationMailer $mailer)
```
franzliedke: Why did you decide to auto-inject this parameter in the `handle()` method and not in the constructor?

src/Notification/Job/SendNotificationsJob.php
```diff
@@ -0,0 +1,66 @@
...
+        if ($this->blueprint instanceof MailableInterface) {
+            $this->email($mailer, $this->blueprint, $recipients);
+        }
+    }
+
+    protected function email(NotificationMailer $mailer, MailableInterface $blueprint, array $recipients)
+    {
+        foreach ($recipients as $user) {
+            if ($user->shouldEmail($blueprint::getType())) {
+                $mailer->send($blueprint, $user);
```
franzliedke: Do we want to queue these separately? Otherwise, a late failure here would probably cause a lot of re-sends when this job is retried...

src/Notification/Job/SendNotificationsJob.php
```diff
@@ -0,0 +1,66 @@
...
+        $attributes = NotificationSyncer::getAttributes($this->blueprint);
+
+        Notification::insert(
+            array_map(function (User $user) use ($attributes, $now) {
+                return $attributes + [
+                        'user_id' => $user->id,
+                        'created_at' => $now
+                    ];
+            }, $recipients)
+        );
```
franzliedke: Do we want the platform notifications to be generated here in the queue as well, or in sync? I could see arguments for both sides... :thinking:

franzliedke:
Just noticed this will also fix #978. :muscle:

src/Notification/NotificationSyncer.php
luceos: I see where you want to go, the thing is Interfaces cannot have methods with bodies, will go for an abstract then. Is that okay?

luceos: Personally all this backward compatibility consideration is highly unnecessary, I think there is maybe one or two outside core and fof that do this. And we overcomplicate the next release with technical debt..

src/Notification/Job/SendNotificationsJob.php
luceos: Jobs never get dependencies inject with Laravel. Construct arguments are for job related payload, see https://laravel.com/docs/5.7/queues#class-structure

src/Notification/Job/SendNotificationsJob.php
luceos: I considered both as well. But I think it would be smart to stick to one way of processing for all notification channels, unless we really have to make exceptions. Right now, pusher notifications can be sent async as well without issue.

luceos:
@franzliedke I decided in favor of an abstract, can give an opinion on this?

src/Notification/Job/SendNotificationsJob.php
franzliedke: Fair enough. We should make sure the insertion of `Notifications` is idempotent, though - so that they are not re-created when a failure happens during mail delivery. Or is the email already sent via queue as well?

src/Notification/Job/SendNotificationsJob.php
franzliedke: Right, my bad.

src/Notification/NotificationSyncer.php
franzliedke: My whole point was to avoid the unnecessary conditions in `getAttributes()` by inlining the methods it calls, in each class' implementation.

The three instance methods of `BlueprintInterface` will simply be collapsed into one `getAttributes()` method.

You should add this method for the `DiscussionRenamedBlueprint`, as well. If you ignore BC concerns, then you will have to do it for all bundled extensions' blueprints as well.

src/Notification/Job/SendNotificationsJob.php
luceos: I considered the same thing. If we want to harden this right now there are two options:

- Dispatch them separately.
- Catch exceptions and ignore them.

..

franzliedke:
Rebased, diff is looking much better now.

src/Notification/Job/SendNotificationsJob.php
```diff
@@ -0,0 +1,63 @@
...
+                    'user_id' => $user->id,
+                    'created_at' => $now
+                ];
+            }, $recipients)
+        );
+
+        event(new Notifying($this->blueprint, $recipients));
+
+        if ($this->blueprint instanceof MailableInterface) {
+            $this->chain([new SendEmailNotificationJob($this->blueprint, $recipients)]);
```
franzliedke: The `SendEmailNotificationJob` only wants one `Recipient`, as far as I can tell, so the loop should be moved here. Right?

src/Notification/NotificationSyncer.php
```diff
@@ -115,7 +112,16 @@ public function sync(Blueprint\BlueprintInterface $blueprint, array $users)
...
+            $this->queue->push(new SendNotificationsJob($blueprint, $newRecipients));
+        }
+
+        if ($blueprint instanceof MailableInterface) {
+            /** @var User $recipient */
+            foreach ($newRecipients as $recipient) {
+                if ($recipient->shouldEmail($blueprint::getType())) {
+                    $this->queue->push(new SendEmailNotificationJob($blueprint, $recipient));
+                }
+            }
```
franzliedke: This whole block (starting at the check for `MailableInterface`) seems unnecessary now that the `SendNotificationsJob` dispatches the job for sending emails (which is good). Assuming the other comment about the missing loop gets tackled...

src/Notification/Job/SendNotificationsJob.php
luceos: Oh wow, that code should have been removed. I'm not merging this for beta 12 as I want to go over this again and confirm it's inner workings before sending it out to our community.

src/Queue/AbstractJob.php
```diff
@@ -0,0 +1,13 @@
...
+namespace Flarum\Queue;
+
+use Illuminate\Bus\Queueable;
+use Illuminate\Contracts\Queue\ShouldQueue;
+use Illuminate\Queue\InteractsWithQueue;
+use Illuminate\Queue\SerializesModels;
+
+class AbstractJob implements ShouldQueue
+{
+    use InteractsWithQueue, Queueable, SerializesModels;
```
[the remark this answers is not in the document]
franzliedke: We'll keep it, it does seem to offer some nice methods.

src/Notification/NotificationSyncer.php
franzliedke: Since the BC stuff worked well in previous releases, I added a small layer here, thus avoiding the abstract base class. It will be pretty simple to remove again for beta.14.

src/Notification/Job/SendNotificationsJob.php
franzliedke: @luceos already changed this so they are now dispatched separately in the `NotificationSyncer`. Works fine for the moment.

franzliedke:
Okay, we cannot quite consider #978 to be fixed with this yet - this "only" tackles notification emails. There are a few other emails, so those need to be queued as well. See #1920 for more discussion on that topic.

franzliedke:
I removed by weird `BlueprintBC` backwards compatibility layer again, see 1cda9dca4f1ee10b06edb0c9456e30e899dbad2e - having `getAttributes()` in one place is actually quite nice.

I moved all of it to the `Notification` model, though - the `getAttributes()` stuff was only ever called in conjunction with fetching or creating notification models, so this felt like it belongs there.

PR #1938, opened 2019-11-17
Add permission to bypass throttling

### LINKED ISSUE #1255: Throttling is bothersome to some permission groups

datitisev:
**Fixes #1255**.
**Continues #1586**.

- Permission is called `floodgate.postWithoutThrottle`
  - Should it go under "moderation" instead? Seems to be more of a utility for those with power than normal members... 
  - Should there be a different permission for starting discussions?  

**Confirmed**

**Required changes:**

- [ ] Related core extension PRs: flarum/english (not done yet, waiting for confirmation)

### DIFF
--- a/js/src/admin/components/PermissionGrid.js
+++ b/js/src/admin/components/PermissionGrid.js
@@ -210,6 +210,12 @@ export default class PermissionGrid extends Component {
       permission: 'discussion.delete'
     }, 80);
 
+    items.add('postWithoutThrottle', {
+      icon: 'fas fa-swimmer',
+      label: app.translator.trans('core.admin.permissions.post_without_throttle_label'),
+      permission: 'postWithoutThrottle'
+    }, 70);
+
     items.add('editPosts', {
       icon: 'fas fa-pencil-alt',
       label: app.translator.trans('core.admin.permissions.edit_posts_label'),
@@ -233,7 +239,7 @@ export default class PermissionGrid extends Component {
       label: app.translator.trans('core.admin.permissions.edit_users_label'),
       permission: 'user.edit'
     }, 60);
-    
+
     return items;
   }
 
--- a/src/Post/Floodgate.php
+++ b/src/Post/Floodgate.php
@@ -33,6 +33,10 @@ class Floodgate
      */
     public function assertNotFlooding(User $actor)
     {
+        if ($actor->can('postWithoutThrottle')) {
+            return;
+        }
+
         if ($this->isFlooding($actor)) {
             throw new FloodingException;
         }


luceos:
- Yes under moderation
- No not for discussion. If memory serves me right, the floodgate is only active when posting.

datitisev:
@luceos Ok, I'll move it to moderation. 

If the discussion posting doesn't have a floodgate, why does the controller assert that it shouldn't throttle?

luceos:
Feel free to merge once the pr is merged to English.

datitisev:
@luceos :+1: What should the wording be? "Bypass posting floodgate"/"Bypass post throttling"...?

jordanjay29:
> 
> 
> @luceos 👍 What should the wording be? "Bypass posting floodgate"/"Bypass post throttling"...?

How about "Reply multiple times without waiting" in keeping with our other Permission names?

franzliedke:
Hmm, what about the permission name? That `floodgate.` prefix is new, not sure what the convention is there. But the `discussion.replyWithoutApproval` seems to be the closest existing thing, so maybe we should scope this one under `discussion`, too? :thinking: 

@tobyzerner If you happen to read this... can you explain the idea behind the naming scheme? Why are some of the permission names scoped, and others aren't?

js/src/admin/components/PermissionGrid.js
```diff
@@ -210,6 +210,12 @@ export default class PermissionGrid extends Component {
       permission: 'discussion.delete'
     }, 80);
 
+    items.add('postWithoutThrottle', {
+      icon: 'fas fa-swimmer',
```
franzliedke: What's the idea behind [that icon](https://fontawesome.com/icons/swimmer?style=solid)? Swimming with the flood? :laughing:

tobyzerner:
They are scoped so that specific policies can check them. eg. DiscussionPolicy's `can` method.

js/src/admin/components/PermissionGrid.js
datitisev: Floodgate? No idea, it's what I originally chose 🤔

luceos:
Also only those permissions prefixed with discussion can be configured per tag.

Ralkage:
I actually pondered on the idea of changing the default hard-coded time limit for the floodgate to something configurable as a group permission to either set it to 0 (bypass floodgate) or a set value (i.e. 30 seconds, 2 minutes, etc) but that may be more work to do I assume.

I'm referencing https://github.com/flarum/core/blob/d492579638fb52dafbfe65f1f36a95eb6047f7f3/src/Post/Floodgate.php#L45-L52 this method of course.

datitisev:
@Ralkage There's probably an issue for that already, and it's outside the scope of this PR.

franzliedke:
I removed the prefix in the permission name, and encapsulated the permission check in the `Floodgate` class itself.

PR #1942, opened 2019-11-23
history back function fix

w-4:
it shouldn't check for canGoBack again after the array pop()

posted it yesterday to:
https://discuss.flarum.org/d/22045-bug-back-button-on-mobile-always-returns-to-all-discussions-page

**Confirmed**

**Required changes:**

### DIFF
--- a/js/src/forum/utils/History.js
+++ b/js/src/forum/utils/History.js
@@ -92,7 +92,7 @@ export default class History {
 
     this.stack.pop();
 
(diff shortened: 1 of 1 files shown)


matteocontrini:
That check was actually added twice, it seems:

https://github.com/flarum/core/commit/985b87da6c9942c568a1a192e2fdcfde72e030ee

https://github.com/flarum/core/commit/37e0a5579bdc8819b211408aec7f02909b8c86a0

PR #1944, opened 2019-11-27
update postStreamScrubber

w-4:
fixes the scrubber in the case that posts have margins

**Fixes #0000**

Fixes the scrubber in case posts have margins.

The check "top <= viewportTop" is not necessary. it's sufficient to take the index from the first item that passed the previous checks. it's the postStream-item that is first visible in the viewport.

**Screenshot**

**Confirmed**

### DIFF
--- a/js/src/forum/components/PostStreamScrubber.js
+++ b/js/src/forum/components/PostStreamScrubber.js
@@ -198,6 +198,7 @@ export default class PostStreamScrubber extends Component {
     let index = $items.first().data('index') || 0;
     let visible = 0;
     let period = '';
+    let indexUpdated = false;
 
     // Now loop through each of the items in the discussion. An 'item' is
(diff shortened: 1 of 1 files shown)


datitisev:
@w-4 Can you provide an additional explanation on what this PR solves and how/why it solves it? Thanks 🙂.

w-4:
@datitisev To be honest, I don't really remember. I quickly tested it locally with and without the fix.
and for me it fixes the following behavior:
go to https://discuss.flarum.org/d/187-word-association-game/1797
press F5 (refresh the page) -> each time it jumps to a different post.

might be related to:
https://github.com/flarum/core/pull/2088

askvortsov1:
Thank you for the PR! I'm going to close this for now, we can always revisit it from #2088.

PR #1945, opened 2019-11-28
Fix PostStreamScrubber click

w-4:
### Fixes a bug. To replicate:
Go to any discussion on the desktop version, f.e. https://discuss.flarum.org/d/187-word-association-game
Click on the scrubber scrollbar, when it is not on the first position. Instead of the clicked position it will jump back to the first position. 

This is because $scrollbar.offset().top adjusts based on the scrolled position, while clientY of the mouseevent is independent of the scrolled position. Use pageY instead.

Shouldn't affect mobile, since there's no click functionality there.

**Confirmed**


franzliedke:
Thanks a lot @w-4 and sorry for letting you wait so long!

PR #1952, opened 2019-12-08
Middleware extender

### LINKED ISSUE #1919: PHP Extender: Middleware registration
This extender will replace the already-deprecated `ConfigureMiddleware` event.

tankerkiller125:
**Fixes #1919**

Adds the Middleware extender for the extension API

- General Code Quality
- Does this meet the general requirements of the extender? (Zend Middlewares do not have a way to organize or remove middleware from my reading)
- Docblocks? (Other Extenders lack Docblocks)

**Screenshot**

No end user changes, extension developers can now use the following in `extend.php`:
```php
(new Extend\Middleware('forum'))->add(FirstMiddleware::class),
(new Extend\Middleware('forum'))->add(SecondMiddleware::class)
// OR
(new Extend\Middleware('forum'))->add(FirstMiddleware::class)->add(SecondMiddleware::class)
```

**Confirmed**


src/Extend/Middleware.php
```diff
@@ -0,0 +1,41 @@
...
+use Flarum\Extension\Extension;
+use Illuminate\Contracts\Container\Container;
+use Zend\Stratigility\MiddlewarePipe;
+
+class Middleware implements ExtenderInterface
+{
+    private $middlewares = [];
+    private $stack;
+
+    public function __construct(string $stack)
```
luceos:
```suggestion
    public function __construct(string $frontend)
```

Let's make the naming of the Frontend we're impacting uniform. Other implementation like the one to add css/js etc uses frontend for this variable. Please modify here and the property.

tankerkiller125: Will do, I was unsure on the name of that one since the original `ConfigureMiddleware` used `$stackName`

luceos:
Related to your docblock question. My reasoning usually is this:

If the method name, arguments and return type hints are obvious enough, no docblock might be necessary. Sometimes you might need to clarify things, especially when the code could be used by lesser experienced users. Adding middleware, in my opinion, is pretty high level, so the way you wrote it works fine for me.

datitisev:
We might want to make the class properties `protected` - all the other extenders have them as such.

tankerkiller125:
@datitisev I've fixed that now, don't know how I did that considering I was using Frontend as a template, must have been a lapse in thought.

franzliedke:
I would rather see all those properties `private` for all extenders. Most of them are so small; if you want to change something about them, just create a new class. You can't trust us not to change e.g. the variable names anyway.

Anyway, let's merge for now.

franzliedke:
Possible next steps: 
- [ ] More use-cases (removing, adding at different positions, replacing)
- [ ] Extender tests (let's pair on those when it's my turn to coach)

Issue #1478, opened 2018-07-02
switcher for show language dropdown not correctly displayed


manurohr:
## Bug Report

**Current Behavior**
In the admin page, I can choose if I want to show langauge dropdown on frontend or not. I decide not to display the dropdown for the list of languages. After I save the setting, the switcher is set back turned on. However on the frontend it has taken the effect, I do not see any dropdown even the switcher is turned on in the admin page.

**Steps to Reproduce**
1. Go to admin page where basic settings are.
2. Turn switcher on for showing dropdown of languages.
3. Save the setting.
4. Refresh and check switcher again.

**Expected Behavior**
The switcher shall be turned off after having saved the setting and refreshed the page.

**Environment**
Flarum 0.1.0-beta.7PHP 7.0.27-0+deb9u1

datitisev:
Ok, let me see if I understood this:

> The language switcher does not appear on the admin pages when enabled

I think this is an intended design, as the admin page has no user controls (when you click your avatar + username) apart from “Logout”.

manurohr:
No, I mean this setting is still turned on even I do not see any language selector on header. Because if I do something else such as writing forum description and save it, the turned on language selector setting would be saved as well. See attached pictures.

<img width="301" alt="user_controls_header" src="https://user-images.githubusercontent.com/38522534/42163221-6107a822-7e0a-11e8-9501-4676a2254445.png">
<img width="357" alt="user_selector" src="https://user-images.githubusercontent.com/38522534/42163222-612d82b8-7e0a-11e8-94f8-cbdda03e496a.png">

datitisev:
Oh, I see what you mean now.
It works for me just fine. Have you tried clearing the Flarum cache & hard refreshing?

franzliedke:
Can anybody reproduce this with the latest dev-master?

manurohr:
@datitisev I think you commented in the wrong ticket. Are you refering to this issue #1477 ?
If yes, please change your startpage to tagspage, then check it again.

datitisev:
@manurohr Oh shoot, you are right lol. I’ll test #1477 again, post it there and then put my comment in a spoiler so it doesn’t occupy space.

This is what happens when you have too many issues open lol.

1iuhongyu:
I can reproduce this problem, the chain problem is the display problem of the front-end language tab. [Language tab issue #1729](https://github.com/flarum/core/issues/1729)

waql:
[...]

PR #1953, opened 2019-12-09
Fix showLanguageSelector reverting to ON whenever BasicsPage is visited

andreasjacobsen93:
The showLanguageSelector reverts to ON state everytime the BasicsPage is visited and saved.

It seems to be a line in BasicsPage which check whether the show_language_selector value is a number, but since it is assigned as m.prop(settings['show_language_selector']) it will always be a function type.

This PR fixes the check to be directly on the settings object where the number is stored.


andreasjacobsen93:
There still seems to be a problem, because settings['show_language_selector'] is a string type.

This can be fixed in another PR.

matteocontrini:
Related issue is #1478, btw.

PR #1954, opened 2019-12-09
[WIP] dismiss translation files of extensions not installed

### LINKED ISSUE #1837: Only translations for enabled extensions should be loaded from language pack
According to [this discussion](https://discuss.flarum.org/d/19971-extensions-translation-in-their-own-repository-or-in-language-pack), translations of extensions should be in language packs instead of extensions itself. I worked recently to make this approach more convenient for translators, but there is still a performance issue with it. 

Right now Flarum loads all translations from language packs, so if I install a language pack with translations for 1000 extension, Flarum will load all 1k translation files even if I don't need 970 of them. I'm not sure how heavy this is from PHP side, but translations are also included in the generated JS file, so this may result in a huge JS file (full of unnecessary translations) which will slow down page load.

I suggest loading only translation files with names that match the ID of one of the installed extensions (+core files). This change may break some language packs which does not follow this convention, so this probably should be addressed in beta 10.

luceos:
**Fixes #1837**

If I'm not mistaken, this PR is an easy implementation to ignore translation files for extensions not installed. It will however re-read them every time the translation cache files are updated. In addition one can still force files into the translator if one has to.

Is this okay?

**Confirmed**

### DIFF
--- a/src/Extend/LanguagePack.php
+++ b/src/Extend/LanguagePack.php
@@ -11,6 +11,7 @@ namespace Flarum\Extend;
 
 use DirectoryIterator;
 use Flarum\Extension\Extension;
+use Flarum\Extension\ExtensionManager;
 use Flarum\Locale\LocaleManager;
 use Illuminate\Contracts\Container\Container;
 use InvalidArgumentException;
@@ -49,13 +50,13 @@ class LanguagePack implements ExtenderInterface, LifecycleInterface
 
         $container->resolving(
             LocaleManager::class,
-            function (LocaleManager $locales) use ($extension, $locale, $title) {
-                $this->registerLocale($locales, $extension, $locale, $title);
+            function (LocaleManager $locales) use ($extension, $locale, $title, $container) {
+                $this->registerLocale($container, $locales, $extension, $locale, $title);
             }
         );
     }
 
-    private function registerLocale(LocaleManager $locales, Extension $extension, $locale, $title)
+    private function registerLocale(Container $container, LocaleManager $locales, Extension $extension, $locale, $title)
     {
         $locales->addLocale($locale, $title);
 
@@ -75,8 +76,15 @@ class LanguagePack implements ExtenderInterface, LifecycleInterface
             $locales->addCssFile($locale, $file);
         }
 
+        /** @var ExtensionManager $extensions */
+        $extensions = $container->make(ExtensionManager::class);
+
         foreach (new DirectoryIterator($directory) as $file) {
-            if ($file->isFile() && in_array($file->getExtension(), ['yml', 'yaml'])) {
+            $slug = $file->getBasename(".{$file->getExtension()}");
+
+            if ($file->isFile()
+                && in_array($file->getExtension(), ['yml', 'yaml'])
+                && (strstr($slug, '-') || $extensions->isEnabled($slug))) {
                 $locales->addTranslations($locale, $file->getPathname());
             }
         }


src/Extend/LanguagePack.php
```diff
@@ -75,8 +76,15 @@ private function registerLocale(LocaleManager $locales, Extension $extension, $l
...
+        /** @var ExtensionManager $extensions */
+        $extensions = $container->make(ExtensionManager::class);
+
         foreach (new DirectoryIterator($directory) as $file) {
-            if ($file->isFile() && in_array($file->getExtension(), ['yml', 'yaml'])) {
+            $slug = $file->getBasename(".{$file->getExtension()}");
+
+            if ($file->isFile()
+                && in_array($file->getExtension(), ['yml', 'yaml'])
+                && (strstr($slug, '-') || $extensions->isEnabled($slug))) {
```
rob006: What is the purpose of `strstr($slug, '-')` here?

datitisev: To transform the package name (`vendor/name`) into an extension ID (`vendor-name`).

E.g. `flarum/auth-github` into `flarum-auth-github`. This is the same extension ID system that is used when putting extensions in JS `flarum.extensions` (they are all IDs).

This extension ID is then used to see if the locale has been loaded (keys that look like `vendor-name.admin`) - this _is_ a breaking change as extensions that do not follow this would not have their locale loaded, but... it's necessary for performance improvements. 

We might want to make a method for it though, instead of simply repeating the same code without explaining what it does well.

Hopefully that made sense 🙂.

rob006: > Hopefully that made sense slightly_smiling_face.

Sorry, it does not make sense for me at all. :P `strstr()` does not make any replacement, it returns substring after first `-`, which then will be cast to boolean. Essentially it will be interpreted as `true` for every extension (since they must contain at least one `-` for separating vendor and package name), which is both cryptic (`strpos() !== 0` seems to be more common and efficient way for searching if string contains another string) and incorrect (it no longer matters if extension is enabled).

datitisev: Oh, I didn't even look at the context of the code, I apologize. I see what you were asking about now.

Yeah, I guess we should be using `strpos` here instead.

> If you only want to determine if a particular needle occurs within haystack, use the faster and less memory intensive function strpos() instead.

luceos: I think this code misses the negate. Because this conditional attempts to verify the file name is not `core.yml` or `validation.yml`.

rob006: Then why you just don't use `in_array() to check this? Using `strstr()` is harder to understand and less reliable (I can name a file as `myextensions.yml` and it also always be loaded).

luceos: What array would you use in this `in_array` conditional @rob006 ?

rob006: @luceos `core` and `validation`?

luceos: Yeah also possible. The reason I decided against that is because we agreed to demand filenames equal to the extension shortened identifier (removing flarum-ext-). This code would allow additional files for core. Not sure what we prefer here.

rob006: > This code would allow additional files for core.

IMO these files should be explicitly whitelisted (this list could be exposed as public constant). Relying on `-` existence in file name allows language pack to use any name for always loaded translations, which may conflict with new conventions added in future. For example in future Flarum could use `extend.yml` file loaded in specific condition, but such file could be already used by some extensions (because Flarum will always load it, so why not?).

datitisev:
Closed because we may want to approach this issue differently.

rob006:
@datitisev @luceos Any details about this "different approach"? Solution from this PR looks pretty fine, I was hoping that #1837 will be fixed in next beta.

clarkwinkelmann:
I don't remember what was discussed about this. That seemed like an appropriate solution.

Regarding the comments above. What about always loading any file that doesn't contain any `-` for now ? This would solve the question about how to whitelist those files, and would also reduce breaking changes for any language pack that named their files differently without a dash.

Language packs will also be able to easily add files that are not subject to the filter, by simply not making them look like an extension ID.

askvortsov1:
@datitisev is this good to go? If so, can we reopen?

rob006:
> Language packs will also be able to easily add files that are not subject to the filter, by simply not making them look like an extension ID.

This looks more like a drawback than a feature. And potential BC break in the future. If Flarum decides to use some special file name for specific condition, there is always a risk that this name is already in use in existing language packs. Reserving all names without dash is IMO much better and reliable solution.

src/Extend/LanguagePack.php
clarkwinkelmann: strstr($slug, '-')

should likely be replaced with

    strpos($slug, '-') === -1

If we don't find the reason why this was dismissed I would be happy to re-open and test this change.

clarkwinkelmann:
Okay so after internal discussion this appears to have been closed because it was still a work in progress and has not actually been tested working yet.

If someone want to finish and test it, feel free to start a new PR based on this one, or from master.

If someone from core takes on this task, they'll re-open the PR.

PR #1958, opened 2019-12-20
Change Extenders properties to private

tankerkiller125:

Changes the Extenders properties from protected to private as noted by @franzliedke in https://github.com/flarum/core/pull/1952#issuecomment-564900520

- I'm guessing this is something that will need further discussion overall

**Screenshot**

**Confirmed**

### DIFF
--- a/src/Extend/Compat.php
+++ b/src/Extend/Compat.php
@@ -23,7 +23,7 @@ use Illuminate\Contracts\Container\Container;
  */
 class Compat implements ExtenderInterface
 {
-    protected $callback;
+    private $callback;
 
     public function __construct($callback)
     {
--- a/src/Extend/Formatter.php
+++ b/src/Extend/Formatter.php
@@ -17,7 +17,7 @@ use Illuminate\Events\Dispatcher;
 
 class Formatter implements ExtenderInterface, LifecycleInterface
 {
-    protected $callback;
+    private $callback;
 
     public function configure($callback)
     {
--- a/src/Extend/Frontend.php
+++ b/src/Extend/Frontend.php
@@ -25,12 +25,12 @@ use Illuminate\Contracts\Container\Container;
 
 class Frontend implements ExtenderInterface
 {
-    protected $frontend;
+    private $frontend;
 
-    protected $css = [];
-    protected $js;
-    protected $routes = [];
-    protected $content = [];
+    private $css = [];
+    private $js;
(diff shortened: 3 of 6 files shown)


luceos:
I am against this. I think all (if not all) code should either use public or protected, never private. I understand that making things private makes it easier to guarantee forward compatibility, but the flexibility gained from being able to override methods is huge. We cannot foresee all circumstances where Flarum will be implemented, keeping core open as much as possible would be beneficial..

franzliedke:
> We cannot foresee all circumstances where Flarum will be implemented, keeping core open as much as possible would be beneficial..

That's not necessary - we provide an interface and all custom use-cases can be taken care of by implementing this interface. If somebody needs to do duplicate some of the code in one of the core extenders, so be it. At least they're not coupled to them. All of our extenders are reasonable small anyway.

I'd say this is very important precisely *because* Flarum is menat to be extensible (which means there will be hundreds of ideas we did not foresee). Under such circumstances, it is important to be able to stick to compatibility guarantees. For extenders, their internal implementation details are not part of these guarantees - only their public interface and the observable behavior (how they change Flarum at runtime, which we soon will start to define through tests). I don't want to be held back in making changes to those internals just because people started to subclass core extenders.

tankerkiller125:
@franzliedke I've rebased, I think theirs still discussion going on here, I'll try to keep this pull request up to date while those discussions take place.

askvortsov1:
Obviously not a core dev, but wanted to put in my 2 cents. I agree with Franz and Matt on this. Our goal in the coming months before stable is to provide, well, a stable API for extensions using extenders. If we find that common use cases require additions to our API, we can offer that, but in my opinion, flexibility should be offered through a well-maintained, expansive, carefully designed API rather than by allowing extension devs to hack around our extenders.

franzliedke:
@luceos Can we convince you? I don't want to ignore your concerns...

luceos:
I'm fine with it. But do expect additional prs when the property visibility limits the ability to write extensions 🙈

askvortsov1:
@luceos I'm hoping that with the extender architecture we come up with, there won't be a need for extensions to register their own extenders in 99% of cases.

At least in my understanding, the primary criteria for something being in core vs as an extension is the extent to which it serves as a base for other systems to extend upon. From that perspective, anything that would need an extender should be covered in core.

PR #1961, opened 2020-01-03
Get translations from fallback catalogues

### LINKED ISSUE #1774: Translation fallbacks are not loaded in the front-end
I investigated the issues a few people had with translations and found that the fallback translations simply aren't loaded in the front-end. Most recent discussion about this on Discuss https://discuss.flarum.org/d/19684-missing-translations-flarum-shows-translation-id

Situation is as follow: an extension isn't translated in a given language. The visitor switches the language of the forum to that language.

Expected: get texts in english.

Actual: no translations available, we end up with the keys being displayed.

We can point the lack of implementation to https://github.com/flarum/core/blob/9115b9e28f9c5d1b53708090f183d71e095678ce/src/Frontend/AddTranslations.php#L57 where we essentially only give the front-end strings in the active language. Meaning the fallback strings are never retrieved.

I did some tests to see what is needed to implement it. It seems as "simple" (see issue below) as taking the example from the Symfony documentation https://symfony.com/doc/current/components/translation/usage.html#retrieving-the-message-catalogue

```php
$catalogue = $this->locales->getTranslator()->getCatalogue($locale);
$translations = $catalogue->all('messages');
while ($fallbackCatalogue = $catalogue->getFallbackCatalogue()) {
    $translations = array_replace($fallbackCatalogue->all('messages'), $translations);
}
```

Trying to run this code, I kept getting "Maximum execution time exceeded" PHP errors. So I suppose maybe the current implementation was made that way because of performance concerns.

Still, it's very problematic. We need those fallback translations in the front-end.

If I find an efficient way of merging those arrays I'll make a PR but in the meantime I wanted to officially report it so we can refer to this issue.

jbrgr:
**Fixes #1774**

Compute the front-end translations by combining the messages with the entries of the fallback catalogue(s), see [Symfony documentation](https://symfony.com/doc/current/components/translation/usage.html#retrieving-the-message-catalog).

The initialization of the fallback catalogue(s) has not been verified since it apparently works as expected (at least in my environment). Maybe this needs to be double-checked.

**Screenshot**
_(A language other than English needs to be selected to let the fallback catalogue take effect.)_
Current:
image
Fixed:
image

**Confirmed**

- [ ] ~Frontend changes: tested on a local Flarum installation.~

### DIFF
--- a/src/Frontend/AddTranslations.php
+++ b/src/Frontend/AddTranslations.php
@@ -55,6 +55,7 @@ class AddTranslations
     {
         $catalogue = $this->locales->getTranslator()->getCatalogue($locale);
         $translations = $catalogue->all('messages');
+
         while ($catalogue = $catalogue->getFallbackCatalogue()) {
             $translations = array_replace($catalogue->all('messages'), $translations);
         }
--- a/src/Frontend/AddTranslations.php
+++ b/src/Frontend/AddTranslations.php
@@ -53,7 +53,11 @@ class AddTranslations
 
     private function getTranslations(string $locale)
     {
-        $translations = $this->locales->getTranslator()->getCatalogue($locale)->all('messages');
+        $catalogue = $this->locales->getTranslator()->getCatalogue($locale);
+        $translations = $catalogue->all('messages');
+        while ($catalogue = $catalogue->getFallbackCatalogue()) {
+            $translations = array_replace($catalogue->all('messages'), $translations);
+        }        
 
         return Arr::only(
             $translations,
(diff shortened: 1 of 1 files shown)


jbrgr:
Thanks for looking into that.

I didn't find a specific rule about where to add blank lines, PSR-2 just mentions that
> Blank lines MAY be added to improve readability and to indicate related blocks of code.

I don't have a strong opinion on that but if it looks better I can just add it.

clarkwinkelmann:
Great! Let's wait for somebody else to review and it will be good to merge :+1:

PR #1963, opened 2020-01-04
Change Zend namespace to Laminas

tankerkiller125:
**Fixes #0000**

Changes Zend namespace to Laminas to reflect the recent name change of the Zend Framework

- Did I find all the Zend usages?

**Screenshot**

**Confirmed**

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -64,9 +64,9 @@
         "symfony/translation": "^3.3",
         "symfony/yaml": "^3.3",
         "tobscure/json-api": "^0.3.0",
-        "zendframework/zend-diactoros": "^1.8.4",
-        "zendframework/zend-httphandlerrunner": "^1.0",
-        "zendframework/zend-stratigility": "^3.0"
+        "laminas/laminas-diactoros": "^1.8.4",
+        "laminas/laminas-httphandlerrunner": "^1.0",
+        "laminas/laminas-stratigility": "^3.0"
     },
     "require-dev": {
         "mockery/mockery": "^1.0",
--- a/composer.json
+++ b/composer.json
@@ -46,6 +46,10 @@
         "illuminate/validation": "5.7.*",
         "illuminate/view": "5.7.*",
         "intervention/image": "^2.3.0",
+        "laminas/laminas-diactoros": "^1.8.4",
+        "laminas/laminas-httphandlerrunner": "^1.0",
+        "laminas/laminas-stratigility": "^3.0",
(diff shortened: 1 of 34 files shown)


franzliedke:
Thanks for tackling this!!!

We do have a small problem, though, and that is extensions [using some Diactoros classes](https://github.com/flarum/auth-facebook/blob/6a9454954c5d45aa6544d3729b41ff2eb0deb460/src/FacebookAuthController.php#L22). They would have to switch to Laminas to be compatible with the next beta - which is something we are trying to avoid - for one beta cycle. Unfortunately, not even our bundled extensions have explicitly declared the dependency - they just trust that flarum/core will include it.

Is it possible to load both dependencies? We can then use Laminas everywhere in our code, and remove the old Zend dependencies for beta.13. That will give extensions one release cycle time to switch over, with both the new and old code working.

datitisev:
@franzliedke Laminas has a package that should allow the use of ZendFramework classes with Laminas, see [laminas/laminas-zendframework-bridge](https://github.com/laminas/laminas-zendframework-bridge)

tankerkiller125:
@datitisev Based on reading through that code package it appears that we should be able to add it as a composer dependency along with the Laminas changes proposed here and it should protect backwards compatibility. I'll see if I can test tonight or tomorrow to ensure that that statement is correct.

tankerkiller125:
I've added the Laminas bridge as a dependency in core, it works exactly as expected (generating autoload classes that just reference Laminas) this should resolve any backwards compatibility issues for beta.12 and the package can then be removed in beta.13 (breaking the Zend namespace) this is definitely something we will want to make sure to publish in the changelog for developers.

franzliedke:
I see the following follow-ups:
- [ ] a ticket for removing the BC layer, targeted for beta.13
- [ ] update bundled extensions to use the new namespaces
  - [ ] auth-facebook
  - [ ] auth-github
  - [ ] auth-twitter
  - [ ] pusher
  - [ ] tags
- [ ] update the docs: `docs/extend/routes.md` references old namespace

@tankerkiller125 Can you take care of these, too? :pray:

tankerkiller125:
@franzliedke I'll take care of those, labels/milestone for the ticket will have to be added by someone else (no access) https://github.com/flarum/core/issues/1964

luceos:
Milestone modified for #1964

franzliedke:
I like how all of these extension PRs had ever-so-slightly different description texts. :rofl: 

In all seriousness: thanks for taking care of this, and so swiftly!

PR #1967, opened 2020-01-15
(#845) Hide permission groups

### LINKED ISSUE #845: Add a way to hide certain groups
https://discuss.flarum.org/d/2181-special-users-profile-icons

luceos:
~~I have been unable to get the admin checkbox to work, even though it exactly resembles the one from the Tag extension. Help appreciated.~~ Fixed.

For reviewers:

- Is this approach okay? Alternative suggestions?

**Fixes #845**

- added permission
- changed serializer
- created column
- modified policy

**Confirmed**

**Required changes:**

- [ ] Related core extension PRs: https://github.com/flarum/lang-english/pull/153

### DIFF
--- a/migrations/2020_01_14_130500_change_permission_groups_add_is_hidden.php
+++ b/migrations/2020_01_14_130500_change_permission_groups_add_is_hidden.php
@@ -1,0 +1,15 @@
+<?php
+
+/*
+ * This file is part of Flarum.
+ *
+ * For detailed copyright and license information, please view the
+ * LICENSE file that was distributed with this source code.
+ */
+
+
+use Flarum\Database\Migration;
+
+return Migration::addColumns('groups', [
+    'is_hidden' => ['boolean', 'default' => 0]
+]);
--- a/migrations/2020_01_14_130500_change_permission_groups_add_is_hidden.php
+++ b/migrations/2020_01_14_130500_change_permission_groups_add_is_hidden.php
@@ -7,7 +7,6 @@
  * LICENSE file that was distributed with this source code.
  */
 
-
 use Flarum\Database\Migration;
 
 return Migration::addColumns('groups', [
--- a/js/src/admin/components/EditGroupModal.js
+++ b/js/src/admin/components/EditGroupModal.js
@@ -3,7 +3,7 @@ import Button from '../../common/components/Button';
 import Badge from '../../common/components/Badge';
 import Group from '../../common/models/Group';
 import ItemList from '../../common/utils/ItemList';
-import Checkbox from '../../common/components/Checkbox';
+import Switch from '../../common/components/Switch';
 
 /**
  * The `EditGroupModal` component shows a modal dialog which allows the user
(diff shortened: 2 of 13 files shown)


migrations/2020_01_14_130500_change_permission_groups_add_is_hidden.php
```diff
@@ -0,0 +1,14 @@
...
+ * This file is part of Flarum.
+ *
+ * For detailed copyright and license information, please view the
+ * LICENSE file that was distributed with this source code.
+ */
+
+use Flarum\Database\Migration;
+
+return Migration::addColumns('groups', [
+    'is_hidden' => ['boolean', 'default' => 0]
```
franzliedke: Can we switch all of this to `is_visible`, please? I'd like to avoid having to deal with double negations everywhere. :pray:

luceos: This requires a follow up issue for Tags then? I followed that example 🙈

franzliedke: Uh, right. Nah, let's rather be consistent then and leave it like it is.

We could discuss what we want in the frontend, though. A "hide" checkbox or one labeled "show"?

luceos: I think the logic tags used is really clear. I think this is the right approach. Not sure how others perceive it.

PS the admin js still has to be verified as I ran into some issues..

luceos:
@franzliedke done, should be ready for the green button now 👍

migrations/2020_01_14_130500_change_permission_groups_add_is_hidden.php
askvortsov1: Do we want to use a hidden_at datetime instead?

tests/integration/api/Controller/ListGroupsControllerTest.php
```diff
@@ -11,21 +11,67 @@
 
 use Flarum\Api\Controller\ListGroupsController;
 use Flarum\Group\Group;
+use Flarum\User\User;
 
 class ListGroupsControllerTest extends ApiControllerTestCase
```
askvortsov1: This will need to be rebased over the new request-based tests for controllers.

migrations/2020_01_14_130500_change_permission_groups_add_is_hidden.php
franzliedke: Personally, the "event" of hiding the group would not be interesting to me. It's not like a discussion where one could expect this flag to change from time to time / where a record is useful. This is a setting that is probably configured once when creating the group and then forgotten.

clarkwinkelmann: This is similar to the `is_restricted` property on Tags. I also think boolean makes the most sense here.

luceos:
@franzliedke rebased.

franzliedke:

franzliedke:
Closing in favor of #2129.

PR #1970, opened 2020-01-17
New extender for error handling

### LINKED ISSUE #1781: New PHP Extender: Registering custom exception handler
Currently, extensions cannot register a custom `ErrorHandler`. For example, an extension cannot throw a custom exception in a controller and also rely on a custom handler to generate an appropriate JSON response.

With the following sample, if `MyException` is thrown in a controller the registered error handler is not called.

```php
# extend.php
return [
    function (ErrorHandler $handler)
    {
        $handler->registerHandler(new MyErrorHandler);
    },
];
```

```php
class MyErrorHandler implements ExceptionHandlerInterface
{
    public function manages(Exception $e)
    {
        return $e instanceof MyException;
    }

    public function handle(Exception $e)
    {
        $status = $e->getCode();
        $error = [
            'status' => (string) $status,
            'message' => $e->getMessage()
        ];

        return new ResponseBag($status, [$error]);
    }
}
```

The problem is that the ErrorHandler is configured with a FallbackHandler that catches all the exceptions (line 82), so if you register a new one it will be appended after the fallback one.

https://github.com/flarum/core/blob/bbe62f400fd849f1f86ce7a4931c59fab4aec6fc/src/Api/ApiServiceProvider.php#L68-L85

I'll leave the discussion and proposals for an improvement to actual PHP/Flarum developers 😄 

Thanks.

franzliedke:
**Fixes #1781**

This extender implements several methods for extending the new error
handling stack implemented in #1843.

Most use-cases should be covered, but I expect some challenges for more
complex setups. We can tackle those once they come up, though. Basic
use-cases should be covered.

**Confirmed**

- [x] Tested on a local Flarum installation.

### DIFF
--- a/src/Extend/ErrorHandling.php
+++ b/src/Extend/ErrorHandling.php
@@ -1,0 +1,120 @@
+<?php
+
+/*
+ * This file is part of Flarum.
+ *
+ * For detailed copyright and license information, please view the
+ * LICENSE file that was distributed with this source code.
+ */
+
+namespace Flarum\Extend;
+
+use Flarum\Extension\Extension;
+use Flarum\Foundation\ErrorHandling\Reporter;
+use Illuminate\Contracts\Container\Container;
+
+class ErrorHandling implements ExtenderInterface
+{
+    private $statuses = [];
+    private $types = [];
+    private $handlers = [];
+    private $reporters = [];
+
+    /**
+     * Define the corresponding HTTP status code for a well-known error type.
+     *
+     * This can be used to configure the status code (second parameter) to be
+     * used for the error response when encountering an exception with a certain
+     * type (first parameter). This type can be provided by the exception class
+     * itself (if it implements {@see \Flarum\Foundation\KnownError}), or
+     * explicitly defined by using the {@see type} method (useful for exception
+     * classes not under your control).
+     */
+    public function status(string $errorType, int $httpStatus)
+    {
+        $this->statuses[$errorType] = $httpStatus;
+
+        return $this;
+    }
+
+    /**
+     * Define the internal error type for a specific exception class.
+     *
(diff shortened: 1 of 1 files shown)


franzliedke:
Tagging some people:
- @matteocontrini because you reported the original issue
- @datitisev because you maintain fof/sentry if I remember correctly
- @clarkwinkelmann because you've created a bunch of extenders already

matteocontrini:
So taking in consideration the use case mentioned in the issue the usage of this extender would be something like:

```php
return [
    (new Extend\ErrorHandling)
        ->handler(MyException::class, MyExceptionHandler::class);
];
```

Where `MyExceptionHandler` exposes an `handle` method. Is that correct?

Shouldn't there be an interface for exception handlers, perhaps?

franzliedke:
@matteocontrini That depends on the exact use-case. If all you're doing is configuring the status code for an exception, you should use the other methods.

Handlers are really only necessary for adding more details to error output (in the JSON-API layer). #1843 removed most of our layers, and [only](https://github.com/flarum/core/blob/1e7fbf1ed9d8f936356fba900d09b61deb3073f6/src/Foundation/ErrorHandling/ExceptionHandler/ValidationExceptionHandler.php) [two](https://github.com/flarum/core/blob/1e7fbf1ed9d8f936356fba900d09b61deb3073f6/src/Foundation/ErrorHandling/ExceptionHandler/IlluminateValidationExceptionHandler.php) remain - those embed validation errors into the JSON-API response. So, assuming you're doing something simpler than that, please try out the other methods. 

> Shouldn't there be an interface for exception handlers, perhaps?

I thought about it and it does not help much, as that would mean you cannot type hint the argument to the `handle()` class.

franzliedke:
You're right, thanks! Added another docblock.

PR #1971, opened 2020-01-18
Additional Middleware Extender functionality

### LINKED ISSUE #1957: Sorted middleware stack
Based on work done in #1919 we now need to be able to mutate the middleware stack before it is finalised. In order to do so we will bind the middleware stack (array) into IoC and resolve it for the pipeline.

The methods we need to allow for are at least:

- `insertAfter`
- `insertBefore`
- `replace`
- `add`
- `remove`

Then when we use the Middleware extender, we can allow direct mutation of this stack as well. An example would be:

```php
$stack->insertAfter(Flarum\Http\Middleware\AuthenticateWithHeader::class, App\Http\Middleware\AuthenticateWithLdap::class);
```
Or similar using the extender:

```php
(new Flarum\Extend\Middleware('forum'))
    ->insertAfter(Flarum\Http\Middleware\AuthenticateWithHeader::class, App\Http\Middleware\AuthenticateWithLdap::class);
```
One could possibly use the magic methods to proxy the method to the stack.

tankerkiller125:
**Fixes #1957**

- Implements the remove, insertBefore, insertAfter and replace functionality for middlewares
- Changed the original `flarum.{frontend}.middleware` container to array
- Moved middleware piping to `flarum.{frontend}.handler` container

- General code style (hopefully StyleCI catches most of it)
- Variable names (could we name them better?)
- Does anyone see a better way to handle the foreach loops?

**Confirmed**

### DIFF
--- a/src/Forum/ForumServiceProvider.php
+++ b/src/Forum/ForumServiceProvider.php
@@ -59,7 +59,7 @@ class ForumServiceProvider extends AbstractServiceProvider
             $this->setDefaultRoute($routes);
         });
 
-        $this->app->singleton('flarum.forum.middleware', function() {
+        $this->app->singleton('flarum.forum.middleware', function () {
             return [
                 HttpMiddleware\ParseJsonBody::class,
                 HttpMiddleware\CollectGarbage::class,
--- a/src/Forum/ForumServiceProvider.php
+++ b/src/Forum/ForumServiceProvider.php
@@ -59,7 +59,20 @@ class ForumServiceProvider extends AbstractServiceProvider
             $this->setDefaultRoute($routes);
         });
 
-        $this->app->singleton('flarum.forum.middleware', function (Application $app) {
+        $this->app->singleton('flarum.forum.middleware', function() {
+            return [
+                HttpMiddleware\ParseJsonBody::class,
+                HttpMiddleware\CollectGarbage::class,
+                HttpMiddleware\StartSession::class,
+                HttpMiddleware\RememberFromCookie::class,
+                HttpMiddleware\AuthenticateWithSession::class,
+                HttpMiddleware\CheckCsrfToken::class,
+                HttpMiddleware\SetLocale::class,
+                HttpMiddleware\ShareErrorsFromSession::class
+            ];
+        });
+
+        $this->app->singleton('flarum.forum.handler', function (Application $app) {
             $pipe = new MiddlewarePipe;
 
             // All requests should first be piped through our global error handler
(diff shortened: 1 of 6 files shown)


tankerkiller125:
@luceos I agree that a test would be beneficial however I have zero experience writing them (especially outside of the full laravel Framework) so if I do write test I'll need help from someone who know how.

luceos:
I'm fine with doing a test after merging, awaiting the second reviewer.

src/Forum/ForumServiceProvider.php
```diff
@@ -69,21 +82,16 @@ public function register()
...
+                $pipe->pipe($this->app->make($middleware));
+            }
 
             event(new ConfigureMiddleware($pipe, 'forum'));
 
             return $pipe;
         });
 
-        $this->app->afterResolving('flarum.forum.middleware', function (MiddlewarePipe $pipe) {
+        $this->app->afterResolving('flarum.forum.handler', function (MiddlewarePipe $pipe) {
```
franzliedke: IMO, we can now get rid of this `afterResolving` call. This was previously necessary to let the extender add middleware, but that now uses a different approach. So, could you move this `$pipe->pipe()` call to right before the `return $pipe` in line 91?

tests/integration/api/Auth/AuthenticateWithApiKeyTest.php
```diff
@@ -159,7 +159,7 @@ public function process(
             });
         });
 
-        $pipe = app('flarum.api.middleware');
+        $pipe = app('flarum.api.handler');
```
franzliedke: Huh, this stuff actually requires leaving in the `afterResolving()` hack for the API stack. I'd like to avoid that. Any good idea on how to handle this differently here?

franzliedke:
Great job, thank you!

I agree, we can start work on testing extenders in a next step. IIRC, that was actually my suggestion for your next task. I'd happy to do another round of pairing in order to get this started. :smiley:

tests/integration/api/Auth/AuthenticateWithApiKeyTest.php
tankerkiller125: This is in fact why I left the `afterResolving()` in, I couldn't figure out any good ways to get around this issue. I also don't know if any extensions possibly need to use the `afterResolving()` method for anything.

franzliedke: As for extensions relying on it, that won't be the case, as the binding name's was changed. :wink: 
Any (known) use-cases should be covered by the extender.

I took a look at the test and think the hack is not really needed, so I will push a replacement to master.

franzliedke: And done in 76f7d566b2b2fc2e675283a8d621de02a16c10b1. Please rebase on the latest master and get rid of the `afterResolving()` calls. Then we can (squash-)merge this. :raised_hands:

franzliedke:
I took care of the rebase and squash-merged this. Thank you!!!

tankerkiller125:
Sorry about not doing it myself, this week at work was crazy (company meetings, remote employees visiting, etc.)

PR #1975, opened 2020-01-24
Use Laravel's slugger for basic transliteration

franzliedke:
This is better than the current system, as it adds transliteration rules
for special characters, rather than just throwing all of them away.

For languages that cannot be transliterated to ASCII in a reasonable
manner, more possible improvements are outlined in #194.


franzliedke:
IMHO, the JS helper (which is used by the tags extension to suggest tag slugs) does not need the same, complex mechanism, as its result can always be overwritten by the user, e.g. when the transliteration produces nonsense.

That way, it can remain on the client-side. :smiley:

PR #1977, opened 2020-01-25
Make AbstractPolicy compatible with both object and class as $model

clarkwinkelmann:
Extensions were previously unable to use class name -based gate calls on policies that extend `AbstractPolicy`. For example calling `$user->can('create', YourOwnModel::class)` was never calling the `create` method defined in the policy. This PR adds the missing check to support that syntax.

I have also added unit tests. I'm not sure where we want to put the test Policy that's required for the test. I'm using the `User` model as the policy target, but we could use any model, including one created only for the tests.

**Confirmed**

### DIFF
--- a/src/User/AbstractPolicy.php
+++ b/src/User/AbstractPolicy.php
@@ -35,7 +35,7 @@ abstract class AbstractPolicy
      */
     public function getPermission(GetPermission $event)
     {
-        if (! $event->model instanceof $this->model) {
+        if (! $event->model instanceof $this->model && $event->model !== $this->model) {
             return;
         }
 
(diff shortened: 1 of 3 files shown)


clarkwinkelmann:
uuuh I guess this was a bit forgotten.

I even removed the patch in my Formulaire extension being certain we shipped this with beta 12. But it wasn't :hear_no_evil: 

Do we want someone else to review before merge ?

clarkwinkelmann:
TODO: Move class definition to the test file like it was done for extenders

franzliedke:
@clarkwinkelmann I've taken care of the TODO in master.

PR #1978, opened 2020-01-25
Add disabled prop to the Select component

clarkwinkelmann:
Add ability to mark a `Select` component disabled. There's no special style, it just marks the underlying `<select>` disabled, which was previously not possible.

**Screenshot**

image

**Confirmed**

### DIFF
--- a/js/src/common/components/Select.js
+++ b/js/src/common/components/Select.js
@@ -8,14 +8,15 @@ import icon from '../helpers/icon';
  * - `options` A map of option values to labels.
  * - `onchange` A callback to run when the selected value is changed.
  * - `value` The value of the selected option.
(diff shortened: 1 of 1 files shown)


datitisev:
What about perhaps using `extract` to extract the `onchange`, `options`, and `value` props, and add the rest of the attrs to the `select` element as attributes?

clarkwinkelmann:
I thought about that possibility as well.

But looking at the other custom input components, I didn't see many use `extract`. We should probably use it everywhere if we prefer to use it.

datitisev:
I'm in favor of allowing any attribute to be added to the select - this would mean that this PR wouldn't be necessary again for a different attribute.

PR #1985, opened 2020-02-06
Support PHP 7.4

### LINKED ISSUE #1980: Flarum Beta 10, 11 Not Support in PHP 7.4.1
## Follow Official Installation Tutorial
https://flarum.org/docs/install.html

## Bug Report

**Current Behavior**
Flarum beta 10, 11 (or all...?) not support in PHP 7.4.1

**Steps to Reproduce**

> 1. Using PHP 7.4.1.
> 2. `composer create-project flarum/flarum . --stability=beta` 
>  or 
> `composer create-project flarum/flarum:v0.1.0-beta.10 . --stability=beta`
> 3. Set all about site setting (httpd-vhosts.conf, httpd.conf).
> 4. Open the browser, and get the **ERROR**:
> ```
> Notice: Trying to access array offset on value of type null in C:\xampp\htdocs\ft\vendor\illuminate\support\ServiceProvider.php on line 84
> ...
> ``` 
> Full error like this:
> image

**Expected Behavior**
If using PHP 7.3.6, the installer is work.
image

**Environment**
- Flarum version: v0.1.0-beta.10, v0.1.0-beta.11
- Website URL: localhost
- Webserver: Apache
- Hosting environment: localhost
- PHP version: 7.3.6, **7.4.1**
- Browser: Google Chrome 79.0.3945.130

```

C:\xampp\htdocs\ft>php --version
PHP 7.4.1 (cli) (built: Dec 17 2019 19:24:02) ( ZTS Visual C++ 2017 x64 )
Copyright (c) The PHP Group
Zend Engine v3.4.0, Copyright (c) Zend Technologies

C:\xampp\htdocs\ft>php flarum info
PHP Notice:  Trying to access array offset on value of type null in C:\xampp\htdocs\ft\vendor\illuminate\support\ServiceProvider.php on line 84

Notice: Trying to access array offset on value of type null in C:\xampp\htdocs\ft\vendor\illuminate\support\ServiceProvider.php on line 84


  Command "info" is not defined.  

```

franzliedke:
**Fixes #1980**

This fixes two issues in some of our dependencies, that caused problems when installing Flarum on a PHP 7.4 system. We only encountered deprecation warnings - which were either shown or not, depending on the system's configured error reporting level.

- Switch to Wikimedia's fork of less.php
- Work around an issue in Laravel (the "old" version we use is no longer supported)

### DIFF
--- a/src/Foundation/UninstalledSite.php
+++ b/src/Foundation/UninstalledSite.php
@@ -108,7 +108,10 @@ class UninstalledSite implements SiteInterface
                 'lifetime' => 120,
                 'files' => $this->paths['storage'].'/sessions',
                 'cookie' => 'session'
-            ]
+            ],
+            'view' => [
+                'paths' => [],
+            ],
         ]);
     }
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,7 +8,7 @@ jobs:
 
     strategy:
       matrix:
-        php: [7.1, 7.2, 7.3]
+        php: [7.1, 7.2, 7.3, 7.4]
         service: ['mysql:5.7', mariadb]
         prefix: ['', flarum_]
 
@@ -31,6 +31,12 @@ jobs:
             service: 'mysql:5.7'
             prefix: flarum_
           - php: 7.2
+            service: mariadb
+            prefix: flarum_
+          - php: 7.3
+            service: 'mysql:5.7'
+            prefix: flarum_
+          - php: 7.3
             service: mariadb
             prefix: flarum_
 
--- a/composer.json
+++ b/composer.json
@@ -57,7 +57,6 @@
         "middlewares/request-handler": "^1.2",
         "monolog/monolog": "^1.16.0",
         "nikic/fast-route": "^0.6",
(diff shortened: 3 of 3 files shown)


franzliedke:
I now configured our CI to run tests on PHP 7.4 as well. :ok_hand:

src/Foundation/UninstalledSite.php
```diff
@@ -108,7 +108,10 @@ private function getIlluminateConfig()
                 'lifetime' => 120,
                 'files' => $this->paths['storage'].'/sessions',
                 'cookie' => 'session'
-            ]
+            ],
+            'view' => [
+                'paths' => [],
+            ],
```
luceos: Can you clarify the need for these @franzliedke ?

franzliedke: Of course! I should have explained that in the commit message, right?

The problem is [this line in Laravel's `ServiceProvider` class](https://github.com/laravel/framework/blob/5.7/src/Illuminate/Support/ServiceProvider.php#L84). When the key does not exist, this now raises a notice in PHP 7.4. Since several of the service providers we use call `loadViewsFrom()`, this affected the installer.

PR #1998, opened 2020-02-13
Fixed discussion list pane only showing up after internal clicks

### LINKED ISSUE #1993: [Tags]: Topic slideout not working if page opened from link
## Bug Report

**Current Behavior**
The topics slideout that pops out when you move your cursor to the left edge of the screen only works if the current page was reached by clicking on a link from the flarums tag page. If the page was reached via opening a URL, refreshing the page, or opening the page in new tab, the slideout doesn't work.

**Steps to Reproduce**
1. Go to https://discuss.flarum.org/d/859-flarum-community-guidelines (or any page)
2. Move your cursor to the left side of the screen. The expected slide-out doesnt pop out
3. Open the same page, but by directly clicking on it from the tags page.
4. Move your cursor to the left side of the screen. The slideout works.

**Expected Behavior**
The slideout should work regardless of how the discussion page was reached.

**Environment**
- Discuss (flarum.org)
- Browser: Chrome 80

**Possible Solution**
I'll look into making a PR to fix.

askvortsov1:
**Fixes #1993**

Right now, the discussion list pane only shows up on discussion pages if the page is accessed in a tab that has previously visited pages. This is mitigated by 2 fixes.

1. If the there is no cached discussion list when accessing a discussion page, one is created. (in DiscussionPage)
2. The cached discussion list is set to be retained by the browser, so as to not be lost if the page refreshes (in IndexPage).

**Confirmed**

- [ ] Backend changes: Backend Not Affected.

### DIFF
--- a/js/src/forum/components/DiscussionPage.js
+++ b/js/src/forum/components/DiscussionPage.js
@@ -1,6 +1,7 @@
 import Page from './Page';
 import ItemList from '../../common/utils/ItemList';
 import DiscussionHero from './DiscussionHero';
+import DiscussionList from './DiscussionList';
 import PostStream from './PostStream';
 import PostStreamScrubber from './PostStreamScrubber';
 import LoadingIndicator from '../../common/components/LoadingIndicator';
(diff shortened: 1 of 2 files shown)


franzliedke:
I am closing this for now to prevent accidental merges. If the discussion in #1993 turns out in favor of this approach, then we can easily reopen it. :hugs:

PR #2008, opened 2020-02-17
WIP: Improve Shared Hosting Experience (Part 1)

### LINKED ISSUE #262: Improve the skeleton for shared hosting

askvortsov1:
**Fixes flarum/issue-archive#262 **

Started: 
- Added command to switch between shared hosting optimized and non-optimized
- Added compat system to support this if the skeleton isn't updated to the most recent version
- Added 403 forbidden routing to non-public resources (e.g. vendor) in nginx conf

Not Started:
- Add mode switch option to installer (both CLI and web) (Only if "public" in path).
- Add warning message if non-public resources (e.g. vendor/composer/installed.json) are publically accessible (put together a list of these to test), although this should be mitigated by the nginx changes

- Should the compat system be separated out into its own command?
- I really don't like having to maintain copies of the site.php, index.php, .htaccess, and .nginx.conf files in 2 repositories. How can we keep them in 1 while not breaking the experience for people who haven't fully updated their system?
- Sometimes (e.g. if running in a docker container), the public/assets folder may be mounted / set as if on a separate file system. From my understanding, php deals with this by using copy-unlink as opposed to the usual rename operation. However, since public/assets is a folder with contents, copy does not instantly work. As such, I used exec(mv) to move the files that needs to be moved. Is there a way to do this that doesn't necessitate exec? 
- Are there other code style / best practice issues I might be missing?

**Confirmed**
Haven't started on unit tests yet as this is still very WIP. Have been testing on a local installation


askvortsov1:
Here's one potential idea I had to the duplicate file system: we seize on this opportunity to improve the general installation process. We move as much as possible of the skeleton into core, and replace flarum/flarum with a shell script that installs core and places the skeleton. From then, the skeleton structure could be updated / replaced with a console command (expanded compat?). This could make the installation and upgrade process simpler, as there's no longer a need to keep the skeleton updated: flarum would do that itself.

askvortsov1:
Another idea: When transitioning out of shared hosting mode, we could allow people to specify the path of their "public" directory. This could provide an easy mechanism for putting flarum on a subdomain with shared hosting.

askvortsov1:
I'm closing this as this issue needs more discussion before settling on a solution

Issue #1891, opened 2019-09-23
Finish PHP extenders


franzliedke:
Since [introducing the extender API in beta.8](https://github.com/flarum/core/issues/851), a bit of time has passed and we still need to add extenders for the following use-cases.

This list should cover what we need to implement all official, bundled extensions' backends only using existing extenders.

## Remaining Post-Beta 15 Steps:

- [x] Finish up and merge Search extenders
- [x] Review and merge user password checker extenders
- [x] Look over extender naming (perhaps mutate => attributes for ApiSerializer?)
- [x] Add a 4th argument default value to Setting extender
- [x] Add a beforeSending callback method to Notification extender
- [x] Figure out what to do with GetModelIsPrivate
- [x] Add `subscribe` to event extender, but register subscribers only after application is fully booted to avoid early resolution.

## Extenders To Replace Events

At this stage, our goal is to replace the current event-based extension system for non-domain events, not to introduce new extensibility functionality (except where we can break current ambiguous events into more granular, use-driven extenders).

- [x] **Extend\ApiSchema**  (#1624)
  - [x] Api\Event\Serializing
  - [x] Api\Event\WillGetData
  - [x] Api\Event\WillSerializeData
  - [x] Event\GetApiRelationship

- [x] **Extend\Console**
  - [x] Console\Event\Configuring

- [x] **Extend\Formatter** (#2098) @askvortsov1 
  - [x] Formatter\Event\Configuring
  - [x] Formatter\Event\Parsing
  - [x] Formatter\Event\Rendering

- [x] **Extend\Floodgate** (#2170) @askvortsov1 
  - [x] Post\Event\CheckingForFlooding

- [x] **Extend\User**
  - [x] User\Event\GetDisplayName
  - [x] Event\ConfigureUserPreferences (#2005)
  - [x] User\Event\CheckingPassword (#2176) @askvortsov1 
  - [x] User\Event\PrepareUserGroups

- [x] **Extend\Policy**
  - [x] User\Event\GetPermission (#2056) @askvortsov1 
  - [x] User\Event\ScopeModelVisibility @askvortsov1 

- [x] **Extend\Search** (#2104) @askvortsov1 
  - [x] Event\AbstractConfigureGambits
  - [x] Event\ConfigureDiscussionGambits
  - [x] Event\ConfigureUserGambits
  - [x] Discussion\Event\Searching
  - [x] User\Event\Searching
  - [x] Event\ConfigurePostsQuery

- [x] **Extend\Routes**
[...]

PR #2009, opened 2020-02-19
Add CSRF exempt route support to middleware extender

### LINKED ISSUE #1941: Allow extensions to register their own CSRF ignored paths
## Feature Request

As discussed in https://github.com/flarum/core/pull/1925 it is sometimes needed to bypass the CSRF check on a specific route, due to some external limitations.

We are using a sign-on service provider with php saml, which returns to our site with a POST request containing metadata used to verify the user. Therefore we need our extension to be able to register a route to be ignored by CSRF check.

Another option we looked at is trying to add a middleware that sets the bypass CSRF value, but it seems middleware cannot be registered in a specific order from an extension and therefore cannot run before the CSRF check.

askvortsov1:
**Fixes #1941, #1905 **

Part of #1891 

- Add an addCsrfExemptRoute method to the Middleware extender
- Improve Middleware processing system to support this

As per https://github.com/flarum/core/issues/1941#issuecomment-586217862, CSRF exemption is likely to be one of the most common use cases of the middleware extender. As such, considering that there isn't a lot of complexity, it might be easier to implement on our end instead of forcing everyone to override the CheckCsrfToken middleware.

TODO:
- Add integration tests for this (will be done after @flarum/core decides which extender this should be a part of).

1. Should this be a separate extender? I thought that since it affects middleware, and we already have lots of extenders planned, it would be best to have it as a part of the middleware extender
2. Is the mechanism for providing the exempt paths to the CheckCsrfToken middleware the best approach possible? I modeled this largely off of the Routes extender
3. Code style?
4. Is the mechanism for matching paths to exempt routes what we want? (currently done via wildcards).

**Confirmed**
- [x] Backend manual testing

### DIFF
--- a/src/Admin/AdminServiceProvider.php
+++ b/src/Admin/AdminServiceProvider.php
@@ -53,7 +53,7 @@ class AdminServiceProvider extends AbstractServiceProvider
             return [];
         });
 
-        $this->$app->singleton('flarum.admin.middleware.csrf', function () {
+        $this->app->singleton('flarum.admin.middleware.csrf', function () {
             return new HttpMiddleware\CheckCsrfToken($this->app->make('flarum.admin.csrfExemptRoutes'));
         });
 
--- a/src/Admin/AdminServiceProvider.php
+++ b/src/Admin/AdminServiceProvider.php
@@ -53,7 +53,7 @@ class AdminServiceProvider extends AbstractServiceProvider
             return [];
         });
 
-        $app->bind('flarum.admin.middleware.csrf', function ($app) {
+        $this->$app->singleton('flarum.admin.middleware.csrf', function ($app) {
             return new HttpMiddleware\CheckCsrfToken($this->app->make('flarum.admin.csrfExemptRoutes'));
         });
 
--- a/src/Admin/AdminServiceProvider.php
+++ b/src/Admin/AdminServiceProvider.php
@@ -53,7 +53,7 @@ class AdminServiceProvider extends AbstractServiceProvider
             return [];
         });
 
-        $this->$app->singleton('flarum.admin.middleware.csrf', function ($app) {
+        $this->$app->singleton('flarum.admin.middleware.csrf', function () {
             return new HttpMiddleware\CheckCsrfToken($this->app->make('flarum.admin.csrfExemptRoutes'));
         });
 
--- a/src/Admin/AdminServiceProvider.php
+++ b/src/Admin/AdminServiceProvider.php
@@ -48,6 +48,11 @@ class AdminServiceProvider extends AbstractServiceProvider
             return $routes;
         });
 
+        $this->app->singleton('flarum.admin.csrfExemptRoutes', function () {
+            // No exempt routes by default
+            return [];
+        });
+
         $this->app->singleton('flarum.admin.middleware', function () {
             return [
                 HttpMiddleware\ParseJsonBody::class,
@@ -71,7 +76,11 @@ class AdminServiceProvider extends AbstractServiceProvider
             ));
 
             foreach ($this->app->make('flarum.admin.middleware') as $middleware) {
-                $pipe->pipe($this->app->make($middleware));
+                if ($middleware == HttpMiddleware\CheckCsrfToken::class) {
+                    $pipe->pipe(new HttpMiddleware\CheckCsrfToken($this->app->make('flarum.admin.csrfExemptRoutes')));
+                } else {
+                    $pipe->pipe($this->app->make($middleware));
+                }
             }
 
             $pipe->pipe(new HttpMiddleware\DispatchRoute($this->app->make('flarum.admin.routes')));
--- a/src/Admin/AdminServiceProvider.php
+++ b/src/Admin/AdminServiceProvider.php
@@ -53,13 +53,17 @@ class AdminServiceProvider extends AbstractServiceProvider
             return [];
         });
 
+        $app->bind('flarum.admin.middleware.csrf', function ($app) {
+            return new HttpMiddleware\CheckCsrfToken($this->app->make('flarum.admin.csrfExemptRoutes'));
+        });
+
         $this->app->singleton('flarum.admin.middleware', function () {
             return [
                 HttpMiddleware\ParseJsonBody::class,
                 HttpMiddleware\StartSession::class,
                 HttpMiddleware\RememberFromCookie::class,
                 HttpMiddleware\AuthenticateWithSession::class,
-                HttpMiddleware\CheckCsrfToken::class,
+                'flarum.admin.middleware.csrf',
                 HttpMiddleware\SetLocale::class,
                 Middleware\RequireAdministrateAbility::class,
             ];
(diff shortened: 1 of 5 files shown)


src/Extend/Middleware.php
```diff
@@ -97,5 +105,9 @@ public function extend(Container $container, Extension $extension = null)
 
             return $existingMiddleware;
         });
+
+        $container->extend("flarum.{$this->frontend}.csrfExemptRoutes", function ($existingExemptRoutes) {
+            return $existingExemptRoutes + $this->csrfExemptRoutes;
```
tankerkiller125: The other additions to arrays happen via a `[]` method, I feel like we should keep that consistent across the extender/code unless there is a specific reason that you've chosen to use a `+` here.

```suggestion
            return $existingExemptRoutes[] = $this->csrfExemptRoutes;
```

tankerkiller125: After further discussion on discord I realized that there was no foreach loop here, And based on that and some testing I believe that using `array_merge` would be the ideal way of doing this (best performance)

datitisev:
I don't know if we want to have middleware-specific code in the service providers - that seems like a bad idea (not to mention repetition of code and complicating the handler code).

I feel like it'd be better to retrieve the exempt routes when actually running it (or earlier) in the CsrfMiddleware class itself - not sure how to do this as I don't think you can access the `frontend` but that's another thing.

Hopefully, this makes sense and I'm not spewing out nonsense - feel free to expand on this or contradict me.

src/Extend/Middleware.php
askvortsov1: Will do, thanks! Only thing is your change has ExemptRules, whereas should be exemptRoutes. I'll address this in a separate commit though

askvortsov1:
> I don't know if we want to have middleware-specific code in the service providers - that seems like a bad idea (not to mention repetition of code and complicating the handler code).
> 
> I feel like it'd be better to retrieve the exempt routes when actually running it (or earlier) in the CsrfMiddleware class itself - not sure how to do this as I don't think you can access the `frontend` but that's another thing.
> 
> Hopefully, this makes sense and I'm not spewing out nonsense - feel free to expand on this and/or contradict me.

To be honest, I agree that having the csrf exempt routes in the service provider isn't 100% ideal, repetition wise: I modeled this off of route registration. The only question is, where else to store it so that middleware can access it, as middleware (as well as the middleware extender) works within the scope of a service provider. Another requirement is that multiple middleware modifying this should not break the system. This is the best solution I've found so far, but if you have suggestions for better ways, I'm completely open.

src/Extend/Middleware.php
askvortsov1: Fixed in 15c08f6

datitisev:
@askvortsov1 I don't have a better working suggestion, just wanted to voice my thoughts on it in case someone did. Otherwise we can keep it this way.

tankerkiller125:
@datitisev This was the best method we were able to come up with in a discussion on discord. I'm sure someone somewhere will come up with a better method at some point but I think right now this one makes the most sense.

src/Http/Middleware/CheckCsrfToken.php
```diff
@@ -17,8 +17,22 @@
...
+    public function __construct(array $exemptRoutes)
+    {
+        $this->exemptRoutes = $exemptRoutes;
+    }
+
     public function process(Request $request, Handler $handler): Response
     {
+        foreach ($this->exemptRoutes as $exemptRoute) {
+            $path = $request->getUri()->getPath() ?: '/';
+            if (fnmatch($exemptRoute, $path) or fnmatch($exemptRoute, (string) $request->getUri())) {
```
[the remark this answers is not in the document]
askvortsov1: yay!

src/Admin/AdminServiceProvider.php
```diff
@@ -72,7 +77,11 @@ public function register()
             ));
 
             foreach ($this->app->make('flarum.admin.middleware') as $middleware) {
-                $pipe->pipe($this->app->make($middleware));
+                if ($middleware == HttpMiddleware\CheckCsrfToken::class) {
+                    $pipe->pipe(new HttpMiddleware\CheckCsrfToken($this->app->make('flarum.admin.csrfExemptRoutes')));
+                } else {
+                    $pipe->pipe($this->app->make($middleware));
+                }
```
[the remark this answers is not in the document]
askvortsov1: Done, that's a lot better! Thanks!

askvortsov1:
> I don't see how this fixes #1905, to be honest. First, because that route is not whitelisted, and second, because of what @clarkwinkelmann outlined in [#1925 (comment)](https://github.com/flarum/core/pull/1925#issuecomment-555646206).

I modified it to be whitelisted. And as to whether we need this, I believe that we absolutely do. 

> Yes, please. From my perspective, the relation to middleware is an implementation detail and not necessarily obvious to the person interested in whitelisting their routes.

I don't think there's a consensus on this. Personally (and I believe @luceos and @tankerkiller125 agree with me), I think that Middleware is a good place for several reasons:

1. We already have A LOT of extenders planned. Combining and narrowing them down will be more intuitive for users imo.
2. The entire process of processing and dispatching requests is done through the middleware pipe. 3. Additionally, CSRF itself is done through the middleware pipe.

I think this is the best place for it, although we could argue for having it in the Routes extender as well. Regardless, I think that putting it in its own extender adds clutter for not a lot of functionality.

> > Is the mechanism for matching paths to exempt routes what we want? (currently done via wildcards).

To be honest, I don't really see any negatives. CSRF can already be disabled via the middleware extender, so it's not like we're adding a security exemption that users wouldn't have anyway.

askvortsov1:
Another thing I just thought of: the changes @franzliedke recommended, made in 7411d46479accdc7a00ed68f5c6ed25952aaa40f, break the ability to replace / addbefore / addafter / remove / etc the csrf middleware. Not sure how we want to handle that...

franzliedke:
> Another thing I just thought of: the changes @franzliedke recommended, made in 7411d46, break the ability to replace / addbefore / addafter / remove / etc the csrf middleware. Not sure how we want to handle that...

No, it's just another name for the binding. We usually use the class name (when it can be instantiated automatically), but a different string identifier - like in that example - is fine, too.

franzliedke:
> I don't think there's a consensus on this. Personally (and I believe @luceos and @tankerkiller125 agree with me), I think that Middleware is a good place for several reasons:

VETO! :wink: 

> 1. We already have A LOT of extenders planned. Combining and narrowing them down will be more intuitive for users imo.

I don't see a practical difference between calling multiple methods on one extender or calling some methods on multiple extenders. We will have many extenders in any case, and that's totally fine.

> 2. The entire process of processing and dispatching requests is done through the middleware pipe.
> 3. Additionally, CSRF itself is done through the middleware pipe.

That was my point: it's an implementation detail. The entire concept of middleware is and should be completely irrelevant to somebody who just has the use case "I don't want this route to be CSRF-protected". This is inline with the idea behind extenders: hide as many implementation details as possible so that a) users don't have to know them and b) we can change them easily without breaking compatibility.

> I think that having this in the Routes extender would be good as well. Regardless, I think that putting it in its own extender adds clutter for not a lot of functionality.

Having a `Csrf` extender may serve as inspiration for things we could add to that extender in the future. Classes tend to attract behavior, and as long as they're small, I like that. :wink:

askvortsov1:
Alright, I'll refactor this, thanks for the review! I'll probably end up making a new HttpServiceProvider to centralize the bindings then. Additionally, what are your thoughts on making the array of Middlewares associative to prevent issues like the one I described above (where any kind of removal, replacement, insertabove, etc is essentially impossible with the changes you proposed and implemented)

askvortsov1:
Closing in favor of https://github.com/flarum/core/pull/2095

PR #2010, opened 2020-02-19
Remove Zend compatability bridge (Beta 13)

### LINKED ISSUE #1964: Remove Backwards compatibility layer for Zend/Laminas
After https://github.com/flarum/core/pull/1963 Flarum will start using the Laminas namespace in beta.12 instead of Zend, as a temp solution to backwards compatibility the zend-bridge was [added to composer](https://github.com/flarum/core/blob/master/composer.json#L52). This bridge should be removed in beta.13 as beta.12 deprecates the Zend namespace.

askvortsov1:
**Fixes #1964**

Trivial commit to remove zend-laminas compatability bridge for beta 13

I don't think this has any unintended consequences?

**Confirmed**

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -49,7 +49,6 @@
         "laminas/laminas-diactoros": "^1.8.4",
         "laminas/laminas-httphandlerrunner": "^1.0",
         "laminas/laminas-stratigility": "^3.0",
-        "laminas/laminas-zendframework-bridge": "^1.0",
         "league/flysystem": "^1.0.11",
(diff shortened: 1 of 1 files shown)


tankerkiller125:
There will be no unintended consequences of doing this in beta.13, beta.12 is being used a transition period for extension developers to switch to Laminas. If they haven't done it by beta.13 that's not going to be our issue. The Flarum core and official extensions are already switched over to Laminas in beta.12

PR #2014, opened 2020-02-20
Drop support for PHP 7.1

### LINKED ISSUE #1988: Require PHP 7.2+
Let's give everyone a two months heads-up, and then dump PHP 7.1 (does not even receive security fixes anymore) for beta.13.

Related:
- https://github.com/wikimedia/less.php/pull/17#issuecomment-580886403
- This would enable further Laravel upgrades. Any LTS we can fix our eyes on?

askvortsov1:
**Fixes #1988 **

- Drop support for PHP 7.1
- Use v1 of wikimedia/less (more up to date)

- Wikimedia less is planning to release v3.0.0 shortly (https://github.com/wikimedia/less.php/pull/26) to clear up their versioning. I think we shouldn't merge till then, and update our composer.json when they do.
- ~~In https://github.com/flarum/core/issues/1988#issuecomment-583945173, Matt suggested upgrading Laravel to 6.x. When I tried doing this, our automated tests would fail when ran in github actions (although not on my local machine). Do we want to look into this update and compatibility with our setup? If so, here, or in a second PR?~~

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,7 +8,7 @@ jobs:
 
     strategy:
       matrix:
-        php: [7.1, 7.2, 7.3, 7.4]
+        php: [7.2, 7.3, 7.4]
         service: ['mysql:5.7', mariadb]
         prefix: ['', flarum_]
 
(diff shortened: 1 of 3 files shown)


datitisev:
No to Laravel v6 - that will be a separate PR, most likely created by a core member as it requires changes to Flarum's internals.

askvortsov1:
> I'd prefer not to delay this by waiting for the LESS library - can you remove the change to that library so that we can merge this now?

Done!

PR #2015, opened 2020-02-21
Getting user permissions shouldn't stop on the first non-null value

### LINKED ISSUE #1832: Applying the "reply to discussions" permission to a tag may break the "lock" mode

askvortsov1:
**Fixes #1832**

Permissions shouldn't depend on extension boot order. Right now, they are determined by getting the first non-null Policy response. I propose evaluating all policy responses, discounting any nulls, and returning "false" if ANY policies return false. If all that return non-null return true, the permission would be granted, otherwise, the procedure to handle null permission responses is unchanged.

- Are there better ways to handle this?

**Confirmed**

### DIFF
--- a/src/User/UserServiceProvider.php
+++ b/src/User/UserServiceProvider.php
@@ -64,12 +64,17 @@ class UserServiceProvider extends AbstractServiceProvider
             // Fire an event so that core and extension policies can hook into
             // this permission query and explicitly grant or deny the
             // permission.
-            $allowed = $this->app->make('events')->until(
+            $allowed = $this->app->make('events')->fire(
                 new GetPermission($actor, $ability, $model)
             );
 
-            if (! is_null($allowed)) {
-                return $allowed;
+            foreach ($allowed as $key => $value) {
+                if (is_null($value)) {
+                    unset($allowed[$key]);
+                }
+            }
+            if (! empty($allowed)) {
+                return ! in_array(false, $allowed, true);
             }
 
             // If no policy covered this permission query, we will only grant
--- a/src/User/UserServiceProvider.php
+++ b/src/User/UserServiceProvider.php
@@ -64,17 +64,29 @@ class UserServiceProvider extends AbstractServiceProvider
             // Fire an event so that core and extension policies can hook into
             // this permission query and explicitly grant or deny the
             // permission.
-            $allowed = $this->app->make('events')->fire(
+            $handler = $this->app->make('events');
+
+            $allowed = $handler->fire(
                 new GetPermission($actor, $ability, $model)
             );
 
+            if (count($allowed) != count($handler->getListeners(GetPermission::class))) {
+                // This means that at least one handler returned false, causing
+                // laravel to return before firing all events. This means that the
+                // permission has been denied.
+                return false;
+            }
+
             foreach ($allowed as $key => $value) {
+                // Remove any null values so that the following empty statement works properly
                 if (is_null($value)) {
                     unset($allowed[$key]);
                 }
             }
+
             if (! empty($allowed)) {
-                return ! in_array(false, $allowed, true);
+                // Grant the permission if any handlers explicitly return true
+                return in_array(true, $allowed, true);
             }
 
             // If no policy covered this permission query, we will only grant


datitisev:
As far as I can tell, this does *not* fix #1832 - I can still reproduce the error using the steps provided in master & this PR.

askvortsov1:
Odd, but will investigate further. Possible that I wasn't following the reproduction steps properly.

datitisev:
@askvortsov1 When you create the discussion, you are also supposed to lock it - I'm assuming you forgot to as it wasn't mentioned in the steps, just shown in the screenshot.

matteocontrini:
I fixed the steps, but I don't think that was the issue. If he forgot to lock he could have replied successfully, and I understand that he couldn't reply (which means issue fixed)

askvortsov1:
Fixed, sorry for the confusion

luceos:
The solution is the code seems overly complex, convoluted. Maybe this could better be handled by our own implementation of a Dispatcher which would generalize the functionality?

askvortsov1:
I don't know if writing a whole new Dispatcher is a good idea (I don't see too many use cases). IMO, the more code we can outsource to libraries and dependencies, the more maintainable flarum becomes. Although I suppose we could extend Laravel's to make it add false to the list of responses before terminating?

As to the complexity, I added an explanation of how this works to the comments and to the commit description, but I do agree that it's a little more hacky than featury.

Regardless, I feel that this (or something closely like it) is the best solution to this issue. If a permission is explicitly denied at any point, the user must not be granted it, otherwise we risk running into significantly more serious security issues. Load order simply should not be a factor here.

askvortsov1:
I don't know if writing a whole new Dispatcher is a good idea (I don't see too many use cases). IMO, the more code we can outsource to libraries and dependencies, the more maintainable flarum becomes. Although I suppose we could extend Laravel's to make it add false to the list of responses before terminating?

As to the complexity, I added an explanation of how this works to the comments and to the commit description, but I do agree that it's a little more hacky than featury.

Regardless, I feel that this (or something closely like it) is the best solution to this issue. If a permission is explicitly denied at any point, the user must not be granted it, otherwise we risk running into significantly more serious security issues. Load order simply should not be a factor here.

askvortsov1:
> The solution in the code seems overly complex, convoluted. Maybe this could better be handled by our own implementation of a Dispatcher which would generalize the functionality?

On second thought, since until is used for this purpose in multiple places in the codebase, maybe it would be good to build our own dispatcher extending off of Laravels where we could support this across the board.

franzliedke:
I'm also doubtful, to be honest. IMO, this requires a deep dive into the permission layer, some new test cases for existing behavior, and a more generic fix in that layer. Will have to do some reading before I can suggest something concrete, though. :speak_no_evil:

askvortsov1:
Doubtful about what part exactly? This PR, or modifying a dispatcher, or both? We use dispatcher->until() throughout multiple parts of the code, and imo that's an issue because an explicit denial shouldn't be overridden regardless of which listener returns it. At least for permissions, not sure about the other uses of until

askvortsov1:
> I'm also doubtful, to be honest. IMO, this requires a deep dive into the permission layer, some new test cases for existing behavior, and a more generic fix in that layer. Will have to do some reading before I can suggest something concrete, though.

Doubtful about our own dispatcher, or about this proposed solution? The permission system is fairly simple, imo. For a given permission, we set off an event, and get results from all the listeners.

The problem is that right now, we just go with the first non-null response. This is a significant issue because for permissions, an explicit false should always override an explicit true.

Ideally, I'd like for the dispatcher to include false with the list of responses. That way, we can check for explicit falses, then explicit trues. Unfortunately, it doesn't do that. So the way I see it, we're left with 3 options:

1. Override the dispatcher to include false in the returned responses, and go with the solution above
2. Use some other mechanism to see if an explicit false was returned, like comparing the number of received responses with the number of should-have-been-received responses (as I'm doing in this PR).
3. Rework all uses of until() completely into extenders. It doesn't respect the order in which extensions are called, which is an issue.

To be honest, I'd prefer option 3. In addition to the fact that we don't want our extensions using event listeners (kind of the whole point of the extenders), returning the first result is, imo, just not a good idea for all its current use cases. 

As a side note, this particular issue should be fixed with the planned Policy extender anyway :D. I'll get that done for Beta 13.

Here's an outline of everything that needs to be done if we're going with option 3:

0. Permissions and policies (this).

1. Model relationships (planned in Models Extender) (Also, why do we limit it to only 1 custom relationship right now lol?)
https://github.com/flarum/core/blob/0d208dc4437c0cc551c556d2c2cc7c12bda369ae/src/Database/AbstractModel.php#L140

2. Serialization Relationships (planned in API Serializer Extender)
https://github.com/flarum/core/blob/0d208dc4437c0cc551c556d2c2cc7c12bda369ae/src/Api/Serializer/AbstractSerializer.php#L131

3. Whether Discussions and Posts are private
https://github.com/flarum/core/blob/d492579638fb52dafbfe65f1f36a95eb6047f7f3/src/Post/Post.php#L102
https://github.com/flarum/core/blob/8877bf97c4b42d8093146039d652262f21c5717d/src/Discussion/Discussion.php#L116

4. Checking for flooding
https://github.com/flarum/core/blob/b91e9032849446d86f6bec0dc1fa6f6cf0146c8a/src/Post/Floodgate.php#L51

5. Getting users' display names
https://github.com/flarum/core/blob/d492579638fb52dafbfe65f1f36a95eb6047f7f3/src/User/User.php#L312

6. Checking passwords
https://github.com/flarum/core/blob/d492579638fb52dafbfe65f1f36a95eb6047f7f3/src/User/User.php#L323

PR #2017, opened 2020-02-21
Middleware extender test

tankerkiller125:
**Fixes #0000**

Adds a test to test the Middleware Extender is working properly

* Is there any cleaner ways anyone can think of to do this?

**Confirmed**


franzliedke:
Thank you very much! I just added a small test case *without* any extenders that shows what the default behavior is, and - in comparison - which part the extender changes.

PR #2018, opened 2020-02-22
Fix permission icon width in admin area

### LINKED ISSUE #2016: Make permission icons fixed width
## Bug Report

**Current Behavior**

Differently sized FA icons cause text to be misaligned on the permissions grid:

**Expected Behavior**

Apply class `fa-fw` to give all icons a fixed width, to prevent this misalignment.

This could be a default class which icons have in the Permissions Grid, in addition to the icon class itself defined in the extension's JS.

**Screenshots**

## After

**Environment**
- Website URL: (local)
- Webserver: apache
- Hosting environment: local (xampp)
- PHP version: 7.3.14

tankerkiller125:
**Fixes #2016**

Uses the Font Awesome fixed width helper to make all icons the same width so that text aligns

* Is this the best way of doing this? Or is there some other way that I missed?

**Screenshot**

The screenshots in #2016 clearly show the diffrence

**Confirmed**

### DIFF
--- a/js/src/admin/components/PermissionGrid.js
+++ b/js/src/admin/components/PermissionGrid.js
@@ -44,7 +44,7 @@ export default class PermissionGrid extends Component {
             </tr>
             {section.children.map(child => (
               <tr className="PermissionGrid-child">
-                <th>{icon(child.icon)}{child.label}</th>
+                <th>{icon(child.icon, {className: 'fa-fw'})}{child.label}</th>
                 {permissionCells(child)}
                 <td/>
               </tr>


davwheat:
I'm definitely not a reviewer, but I think this is the best way to go about it.

There's no real reason that an extension would want to make their permission icon not be fixed width, as far as I can think of.

tankerkiller125:
the only thing I can possibly think of is if an extension for some reason uses it's own icon library

askvortsov1:
> the only thing I can possibly think of is if an extension for some reason uses it's own icon library

In that case, the added CSS class should have no effect, I think this is fine!

davwheat:
Brilliant!

franzliedke:
> the only thing I can possibly think of is if an extension for some reason uses it's own icon library

This was a good point, so I [fixed this slightly differently](https://github.com/flarum/core/commit/922e29466814ecb3968158badb7ead92f4fafc2f) by applying the fixed width to our existing CSS for these icons (along with some additional cosmetics). This should be completely independent of icon library (or future FontAwesome upgrades).

Thanks for bringing it up @davwheat and thanks for the PR @tankerkiller125 !

PR #2020, opened 2020-02-22
Only translations for enabled extensions should be loaded from language pack

### LINKED ISSUE #1837: Only translations for enabled extensions should be loaded from language pack

rob006:
fixes #1837
replaces #1954

This is continuation of #1954. I fixed condition for detecting core files (`validation.yml` and `core.yml`) and tested this locally - it is working as expected.

### DIFF
--- a/src/Extend/LanguagePack.php
+++ b/src/Extend/LanguagePack.php
@@ -11,6 +11,7 @@ namespace Flarum\Extend;
 
 use DirectoryIterator;
 use Flarum\Extension\Extension;
+use Flarum\Extension\ExtensionManager;
 use Flarum\Locale\LocaleManager;
 use Illuminate\Contracts\Container\Container;
 use InvalidArgumentException;
@@ -18,6 +19,11 @@ use RuntimeException;
 
 class LanguagePack implements ExtenderInterface, LifecycleInterface
 {
+    private const CORE_LOCALE_FILES = [
+        'core',
+        'validation',
+    ];
+
     private $path;
 
     /**
@@ -49,13 +55,13 @@ class LanguagePack implements ExtenderInterface, LifecycleInterface
 
         $container->resolving(
             LocaleManager::class,
-            function (LocaleManager $locales) use ($extension, $locale, $title) {
-                $this->registerLocale($locales, $extension, $locale, $title);
+            function (LocaleManager $locales) use ($extension, $locale, $title, $container) {
+                $this->registerLocale($container, $locales, $extension, $locale, $title);
             }
         );
     }
 
-    private function registerLocale(LocaleManager $locales, Extension $extension, $locale, $title)
+    private function registerLocale(Container $container, LocaleManager $locales, Extension $extension, $locale, $title)
     {
         $locales->addLocale($locale, $title);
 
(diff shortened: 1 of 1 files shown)


rob006:
OK, I found one problem. Extensions that use translations but does not register `Locales` extender, does not trigger locale cache refresh after enabling/disabling the extension - user needs to refresh cache manually to load translations for these extensions. This affects only core extensions - they have their translations in https://github.com/flarum/lang-english/tree/master/locale. Every other extension has bundled English translation and register it using `Locales` extender.

I'm not sure how to proceed with this. Personally, I would just move these translation files to extensions, like in every other 3rd party extension - it seems to be more consistent and pragmatic (you don't need to open 2 PRs every time you add a new setting to extension, and you can release extensions and language pack independently).

askvortsov1:
Is there a reason why core extensions don't use Locales? @core team

tankerkiller125:
I'm personally in favor of moving translations to their respective extensions.

askvortsov1:
I would agree, that should make development a bit less complex imo.

askvortsov1:
> I'm personally in favor of moving translations to their respective extensions.

Should this be split into a separate issue?

rob006:
> I'm personally in favor of moving translations to their respective extensions.

Another reason to do this is that English fallback (#1961) will not work for these extensions, since there will be no English translations available (unless you have English language pack installed and enabled).

clarkwinkelmann:
Suggestion: call the Locale extender in each of the core extensions even if they have no locales files so that they clear the cache. That way we can ship this bug-free without having to make an immediate discussion on whether we move language files out of the English pack.

Maybe we need an additional method on the Locale extender so that we can register an empty locale file list without having to create an empty `locales` folder.

franzliedke:
We just talked about this in our dev meeting.

- We agree with the approach.
- We want to discuss translations for bundled extensions in a separate issue (#2090) in time for the stable release.
- For bundled extensions, the problem currently isn't solved. IMO, that's not a huge problem. We can ignore that (as #2090 may solve it, too). or go with Clark's approach.

---

> Maybe we need an additional method on the Locale extender so that we can register an empty locale file list without having to create an empty locales folder.

@clarkwinkelmann I think that won't be necessary. For the current code in the extender, *any* directory without YAML files should do - and it might even work with non-existing directories. :grin:

tomsgad:
> We just talked about this in our dev meeting.
> 
> * We agree with the approach.
> * We want to discuss translations for bundled extensions in a separate issue (#2090) in time for the stable release.
> * For bundled extensions, the problem currently isn't solved. IMO, that's not a huge problem. We can ignore that (as #2090 may solve it, too). or go with Clark's approach.
> 
> > Maybe we need an additional method on the Locale extender so that we can register an empty locale file list without having to create an empty locales folder.
> 
> @clarkwinkelmann I think that won't be necessary. For the current code in the extender, _any_ directory without YAML files should do - and it might even work with non-existing directories. 😁

I am still getting the same error, I have just added beta-12, all I get is keys e.g, core.forum.discussion_list.empty_text. Any suggestions on how I can fix this for now?

askvortsov1:
> I am still getting the same error, I have just added beta-12, all I get is keys e.g, core.forum.discussion_list.empty_text. Any suggestions on how I can fix this for now?

This is most likely unrelated to this pull request, which is good, since you probably don't need to wait for Beta 13 to fix it! I suspect that you're also the one who replied on https://discuss.flarum.org/d/19684-missing-translations-flarum-shows-translation-id based on the profile picture and timing. I'll respond there as discuss is a better place for support.

franzliedke:
Possible follow-up tasks for interested people (hint hint @rob006 @clarkwinkelmann @askvortsov1):

**Updates for bundled extensions**
If we want to fix this for the bundled extensions (I don't care too much), feel free to send PRs to all bundled extensions (*except for the English language pack*) with the following:

1. A new file `locale/.keep`.
2. A new line `new Extend\Locales(__DIR__ . '/locale'),` in the extension's `extend.php` script.

**Documentation update**
The *convention* on how to name translation files is now a stricter requirement, if I get this right. So we should mention this [in the documentation](https://flarum.org/docs/extend/i18n.html#registering-locales).

Issue #1655, opened 2018-04-17
"Send test email" button


luceos:
There is a way to reset your (admin) password without doing a full new reinstall.

Use the forgot password feature. In case the mail driver isn't fully working edit the `settings.mail_driver` value to `log`. Then run the password forgotten action again and check `storage/logs/flarum.log` it will hold the full email contents.

franzliedke:
Another good follow-up to #1169.

askvortsov1:
The above 2 PRs should address both the test mail functionality and the console password reset functionality as requested in the discussion above.

PR #2021, opened 2020-02-23
Added console command to reset password for a user

askvortsov1:
**Fixes https://github.com/flarum/core/issues/1655#issuecomment-394183409**

Added a console command to reset anyone's password. If a forum admin finds themselves logged out with a forgotten password and a broken email configuration, they shouldn't need to reinstall Flarum. Yes, it's also important to work on preventing broken email configuration, but there should be a failsafe.

- Do we want to add commands for creating an admin user and creating a user?

**Confirmed**

- [x] Backend changes: manually tested.

### DIFF
--- a/src/User/Console/SetPasswordCommand.php
+++ b/src/User/Console/SetPasswordCommand.php
@@ -15,7 +15,7 @@ use Flarum\User\UserRepository;
 use Flarum\User\UserValidator;
 use Illuminate\Validation\ValidationException;
 
-class ResetPasswordCommand extends AbstractCommand
+class SetPasswordCommand extends AbstractCommand
 {
     use AskQuestionTrait;
 
@@ -40,8 +40,8 @@ class ResetPasswordCommand extends AbstractCommand
     protected function configure()
     {
         $this
-            ->setName('password:reset')
-            ->setDescription('Reset a user\'s password');
+            ->setName('user:set_password')
+            ->setDescription('Set a user\'s password');
     }
 
     /**
--- a/src/User/Console/SetPasswordCommand.php
+++ b/src/User/Console/SetPasswordCommand.php
@@ -23,9 +23,6 @@ class SetPasswordCommand extends AbstractCommand
 
     protected $userValidator;
 
-    /**
-     * @param UserRepository $userRepository
-     */
     public function __construct(UserRepository $userRepository, UserValidator $userValidator)
     {
         $this->userRepository = $userRepository;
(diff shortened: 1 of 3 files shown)


franzliedke:
@flarum/core Do we want this? I *did* suggest it in #1655, but we never really discussed / voted on it.

luceos:
I would prefer one, this one needs a bit of work though (eg use the actual Validator). Aside from that, looking great already.

askvortsov1:
I took a lot of the code from the admin user creation on installation. Is there an abstraction somewhere I can use here?

luceos:
The UserValidator. You can resolve that validator from the container to reuse the password validation.

askvortsov1:
> The UserValidator. You can resolve that validator from the container to reuse the password validation.

Done!

askvortsov1:
Updated to use new trait, depends on (https://github.com/flarum/core/pull/2068)

clarkwinkelmann:
Wouldn't something like `password:set` and `SetPasswordCommand` make more sense for the naming ? To me "reset" is associated to the password reset email. The current naming would indicate to me that it either sends the password reset, or generates a password reset link. 

Generating a reset link would also be a valid way of solving this and wouldn't require managing the password input here. Having the command generate a reset link could also help admins manually send the link to a user for which emails are not working.

askvortsov1:
I envisioned 'reset' because the password is already set, and thus, changing it is resetting it. I would be happy with changing it to password:set, although I would like to hear the common consensus first. I think the link generation approach is a bit unnecessarily convoluted though, but am open to considering it as well.

datitisev:
Would it make sense to put this commands under a user:  namespace? I feel like nothing else would be added to a password namespace, so maybe user:password makes more sense (or something similar)?

askvortsov1:
I would be fine with changing it to user:setPassword, I agree since we might want to create more user related commands (as might extensions)

franzliedke:
Agreed about the `user` namespace, but please, no camelCase in the console. :see_no_evil:

askvortsov1:
Fixed to use user namespace and changed the name, also cleaned up StyleCi, so this should be good to go once (#2068) is merged.

src/User/Console/SetPasswordCommand.php
```diff
@@ -0,0 +1,96 @@
...
+    use AskQuestionTrait;
+
+    protected $userRepository;
+
+    protected $userValidator;
+
+    /**
+     * @param UserRepository $userRepository
+     */
+    public function __construct(UserRepository $userRepository, UserValidator $userValidator)
```
luceos, line 29: If the class is type hinted already, there's no need to add a phpdoc. Because - as you can see - it makes code maintenance harder; you are missing one argument already 👍 

```suggestion
    public function __construct(UserRepository $userRepository, UserValidator $userValidator)
```

askvortsov1:
Done, thanks a lot!

luceos:
Can you confirm AskQuestionTrait is working, the tests seem to fail on that @askvortsov1 .

askvortsov1:
It's working, but it's split up into a different PR (2068), so that'd need to be merged first

PR #2023, opened 2020-02-24
Send Test Mail Feature

### LINKED ISSUE #1655: "Send test email" button

askvortsov1:
**Fixes #1655**

- Added Admin endpoint for sending a test email
- Added frontend support for sending test emails
- Added TemporarySettingsRepository to encapsulate settings for temporary use (not saved in db)

- Should the endpoint be moved to API?
- Are there better ways to handle settings encapsulations than the TemporarySettingsRepository
- Code style?

**Confirmed**


franzliedke:
Some high-level things before I dive deeper into the code:

1. Can we see a screenshot, please? :pray: 
2. What about using the current user's email address? That would let us get rid of quite a bit of complexity surrounding that extra field, and frankly, it's probably the address you're gonna enter anyways.

src/Admin/Controller/SendTestMailController.php
```diff
@@ -0,0 +1,106 @@
...
+
+class SendTestMailController implements RequestHandlerInterface
+{
+    use AssertPermissionTrait;
+
+    protected $app;
+
+    protected $translator;
+
+    public function __construct(Application $app, TranslatorInterface $translator)
```
franzliedke: Please typehint the `Container` contract instead. You're only using the container methods, as far as I can tell, so that's a bit more decoupled.

(You'll receive the same instance in the end, but this class doesn't need to know that. :wink:)

src/Admin/routes.php
```diff
@@ -16,4 +17,9 @@
         'index',
         $route->toAdmin()
     );
+    $map->post(
+        '/testmail',
```
franzliedke: My OCD would prefer `/mail/test`, to be honest. :grin: 

Maybe we should change the settings endpoint to `/mail/settings` then? :thinking:

franzliedke: Oh wait, this isn't the API? Interesting. You *are* returning JSON here, but it's not a typical JSON:API endpoint.

src/Admin/Controller/SendTestMailController.php
```diff
@@ -0,0 +1,106 @@
...
+    {
+        $this->app = $app;
+        $this->translator = $translator;
+    }
+
+    public function handle(ServerRequestInterface $request): ResponseInterface
```
franzliedke: We could probably drastically simplify this method with the following approach:
- Assume settings are saved.
- *Only* the test recipient is then needed in the payload (depending on the discussion about using the admin's email). It's not a settings field like all the others, it's only needed for this little transaction.
- You can now just inject the driver, ask it whether it `canSend()` for error handling, and get the transport / mailer from it.

What do you think?

franzliedke:
Oh, one more thing... question for @flarum/core:
Depending on how we implement this, this test mail would either be sent via the queue (if a real queue is enabled) or ... not. I would think in this one case we actually want synchronous sending, right (as the queue would keep retrying and wouldn't give us direct feedback).

src/Admin/Controller/SendTestMailController.php
askvortsov1: While that would be cleaner code, I don't think we should be saving potentially invalid settings. To be honest, I would prefer if the "Save Settings" button was disabled until a successful test email has been sent, that way it's impossible to add in incorrect config. Thoughts?

src/Admin/routes.php
askvortsov1: I'm going to move this to the API, it doesn't make sense to have JSON Endpoints in admin.

askvortsov1:
Good idea about using the email address. And yeah I considered doing it through the queue, but we would need it to be synchronous, which is why so much logic is recreated.

src/Admin/Controller/SendTestMailController.php
askvortsov1: Done

js/src/admin/components/MailPage.js
```diff
@@ -163,34 +162,55 @@ export default class MailPage extends Page {
     return this.fields.some(key => this.values[key]() !== app.data.settings[key]);
   }
 
+  sendingTestEmail() {
+    return this.sendingTest;
+  }
+
```
datitisev: We use Mithril props for this. `this.sendingTest = m.prop(false);`, then update it with `this.sendingTest(true)`.

askvortsov1:
Thanks, will update! I'm not super familiar with mithril so I appreciate the advice!

luceos:
If you use dispatchNow just use the Bus instead of the queue. The bus has dispatchNow to fire off the job immediately.

askvortsov1:
My point is, I don't think we should save incorrect settings, so while saving the settings reduces the amount of code, it's also un intuitive to the user and might break the system. That's why I don't want to call into the app's mailer. Right now, I'm essentially reconstructing a SwiftMailer and just using that, were y'all thinking of a different approach?

clarkwinkelmann:
How do other platforms do it ?

I was under the impression most systems first save the email info, and then offer a way to test the currently saved configuration.

IMO Sending the test email without saving the settings first could leave people believing everything is okay and closing the window without saving the settings they just entered.

If you use the saved credentials, you're testing that the currently saved configuration is valid. Which I think is the most important.

askvortsov1:
I modelled this off of what I remember while using Discourse. I also think that since the send test email button is before the send button, that should be clear to users. I could add additional text clarifying. From a UX perspective, I'd prefer to be able to test send until it works, then save.

IMO it's better not to save credentials unless we know theyre valid. Yes it would make the code cleaner, but if our goal is to minimize incorrect configuration. Tbh, I'd prefer to go as far as disabling the save settings button until a test email has been sent

clarkwinkelmann:
I think saving credentials that are not yet valid is a perfectly valid use case, so requiring sending a test email would be problematic for that use case. Also from a development standpoint I might be offline while I configure email, this shouldn't prevent saving some settings.

Usually when entering new email settings you don't yet have a valid setup. So saving or not saving the potentially invalid credentials wouldn't change a thing :man_shrugging: 

The only case where I might want to test before saving is when changing from one email provider to another. Seeing this makes the code more complex, I'd recommend not including that in core. An extension could provide a way to test email credentials without saving them if someone needs that feature. The test email button could simply be grayed out until you save credentials.

My opinion of course. I would love to hear what other think.

Random thought: how often do people enter **wrong** credentials ? It's been years since I had email problems. I always copy-paste the credentials from my provider and it just works.

franzliedke:
I totally agree with Clark. Also consider that the "Test Email" button can be useful to debug / confirm temporary problems with sending emails (rate limiting etc.) - no need for changing any settings there.

> The test email button could simply be grayed out until you save credentials.

Exactly, I was going to suggest that as well. We already do the opposite for the "Save" button IIRC - it is grayed out until at least one value has been changed.

franzliedke:
P.S.: My suggestion was not primarily about simplifying the code - it's how I've always envisioned that feature. Sorry for not being clearer in the ticket.

askvortsov1:
Ok, sounds good, I'll rework this

askvortsov1:
> P.S.: My suggestion was not primarily about simplifying the code - it's how I've always envisioned that feature. Sorry for not being clearer in the ticket.

Done!

luceos:
One nitpick. Not sure what the others think. Would it make more sense to use route test/mail instead, so that we can share this path for other logic? Or maybe even confirm-settings/mail?

askvortsov1:
I think mail/test (what's being proposed here) only makes sense if we change mail-settings to mail/settings (which I think I have in this PR) (which on second thought I should split into a second PR, I'll do that later today). I'm not sure about confirm-settings/mail, since it's not really a confirmation (i see that more as a user action), but I'm fine with both test/mail and mail/test

datitisev:
I feel like it'd be better to have them as `/mail/settings` and `/mail/test`, not the other way around...

franzliedke:

js/src/admin/components/MailPage.js
```diff
@@ -149,10 +169,56 @@ export default class MailPage extends Page {
     return this.fields.some((key) => this.values[key]() !== app.data.settings[key]);
   }
 
+  sendTestEmail() {
+    if (this.saving || this.sendingTest()) return;
+
+    this.sendingTest = m.prop(true);
+    const settings = {};
+
+    this.fields.forEach((key) => (settings[key] = this.values[key]()));
```
clarkwinkelmann, line 178: If I understand correctly, the saved settings from the database are now used to send the test. So this means this code is no longer needed (?)

js/src/admin/components/MailPage.js
```diff
@@ -149,10 +169,56 @@ export default class MailPage extends Page {
...
+        app.alerts.show(
+          new Alert({
+            type: 'success',
+            children: app.translator.trans('core.admin.email.send_test_mail_success'),
+          })
+        );
+      })
+      .catch((error) => {
+        this.sendingTest = m.prop(false);
+        const response = JSON.parse(error.responseText)['message'];
```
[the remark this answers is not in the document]
clarkwinkelmann: And in case the message isn't there, I suggest re-throwing the error so that the normal Flarum error handler can kick in

js/src/admin/components/MailPage.js
```diff
@@ -11,6 +11,7 @@ export default class MailPage extends Page {
     super.init();
 
     this.saving = false;
+    this.sendingTest = m.prop(false);
```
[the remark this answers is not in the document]
datitisev: Just as an FYI, Mithril still has them in `mithril/stream`, as an optional feature.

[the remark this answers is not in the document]
clarkwinkelmann: I know, I was trying to keep my comment short :innocent:

js/src/admin/components/MailPage.js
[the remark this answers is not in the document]
datitisev: Doesn't `error` have a `response` field for JSON? And if not - we need a try/catch here anyway to make sure to throw the original error and not the JSON parsing error.

askvortsov1:
Thanks for the review, now simplified further!

js/src/admin/components/MailPage.js
```diff
@@ -121,11 +122,29 @@ export default class MailPage extends Page {
                 ],
               })}
 
-            {Button.component({
-              type: 'submit',
-              className: 'Button Button--primary',
-              children: app.translator.trans('core.admin.email.submit_button'),
-              disabled: !this.changed(),
+            {FieldSet.component({
```
clarkwinkelmann, line 125: Nitpick: Could we just use `<FieldSet></FieldSet>` here and get rid a bunch of braces ?

src/Api/Serializer/MailSettingsSerializer.php
```diff
@@ -17,7 +17,7 @@ class MailSettingsSerializer extends AbstractSerializer
     /**
      * {@inheritdoc}
      */
-    protected $type = 'mail-settings';
+    protected $type = 'mail/settings';
```
clarkwinkelmann, line 20: Was this change here before ? Doesn't it impact other code that has not been updated ?

askvortsov1: ack I thought I had removed that change, sorry. Shouldn't impact code, but definitely should be its own PR. Reverted,

js/src/admin/components/MailPage.js
askvortsov1: Fixed throughout, thanks!

clarkwinkelmann:
Oops I see you've taken my suggestions to the max. I actually think it's a bit too much now :sweat_smile: 

- Regarding the FieldSet, I meant that one FieldSet that only had a `children` attribute. I think changing everything in f6592346cee98b9f2aaac26cce49c17b6da80528 is too much, it doesn't help readability very much for the FieldSets with many props
- Regarding the API path changes, I think it's fine to do it in this PR. I was just wondering whether the change in the serializer was intentional. The serializer type doesn't have to match with the API path exactly. I was finding it a bit odd to have `mail/settings` as the serializer type (we've not used a slash anywhere else yet). The URL could be `mail/settings` while the serializer stays as `mail-settings` ?

js/src/admin/components/MailPage.js
```diff
@@ -68,65 +69,69 @@ export default class MailPage extends Page {
...
+                  <Select
+                    value={this.values.mail_driver()}
+                    options={Object.keys(this.driverFields).reduce((memo, val) => ({ ...memo, [val]: val }), {})}
+                    onchange={this.values.mail_driver}
+                  />
+                </label>
+              </div>
+            </FieldSet>
+
+            {this.status.sending
```
datitisev, line 94: This could be `!this.status.sending && Alert.component()`

js/src/admin/components/MailPage.js
```diff
@@ -68,65 +69,69 @@ export default class MailPage extends Page {
...
+              ? ''
+              : Alert.component({
+                  children: app.translator.trans('core.admin.email.not_sending_message'),
+                  dismissible: false,
+                })}
+
+            {fieldKeys.length > 0 ? (
```
datitisev, line 101: This could be `{!!fieldKeys.length && (...)}` (not sure if the `!!` is needed to make sure the `0` doesn't render - probably not, feel free to test.

PR #2026, opened 2020-02-27
Unfuzz Times

### LINKED ISSUE #815: Unfuzz some fuzzy times (e.g. >30 days, >1 year)
As a result of [this discussion on discuss.flarum](https://discuss.flarum.org/d/1776-display-full-date-in-post-header-of-posts-older-than-30-days). 

Quoting @tobscure:

> <30 days = relative time ago
> <1 year = Day Month, no year
> [_greater than_ symbol here because Markdown is a PITA]1 year = Day Month Year

This prevents ambiguous dates like Oct, '15, which can be misconstrued (especially in North America) as 15 October, instead of October 2015.

askvortsov1:
**Fixes #815**

Improved dates:
< 30 days ago: Relative
30 days ago < DATE < 1 year: MMM D (if US), D MMM (if not US)
> 1 year: MMM D, YYYY (if US), D MMM YYYY (if not US)

- Should we change months to non-abbreviated?

**Confirmed**

### DIFF
--- a/js/src/common/utils/humanTime.js
+++ b/js/src/common/utils/humanTime.js
@@ -18,26 +18,15 @@ export default function humanTime(time) {
 
   const day = 864e5;
   const diff = m.diff(moment());
-  const userLang = (navigator.language || navigator.userLanguage).toLowerCase();
   let ago = null;
 
   // If this date was more than a month ago, we'll show the name of the month
   // in the string. If it wasn't this year, we'll show the year as well.
   if (diff < -30 * day) {
     if (m.year() === moment().year()) {
-      // By default, moment's en locale uses MDY. If language is english and the user's browser is not US or CA, DMY should be used instead.
-      if (moment.locale() == 'en' && (userLang == 'en-us' || userLang == 'en-ca')) {
-        ago = m.format('MMM D');
-      } else {
-        ago = m.format('D MMM');
-      }
+      ago = m.format('D MMM');
     } else {
-      // By default, moment's en locale uses MDY. If the user's browser is not US or CA, DMY should be used instead.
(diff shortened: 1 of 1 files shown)


tankerkiller125:
Do we know if the US is the only country that formats time as Month Day Year? Are there any other countries that use this format that we should be aware of?

askvortsov1:
It seems like most countries don't have a clear pattern between either, hence the default to the more internationally acceptable D MMM YYYY. 

One case that we might want to add in is YYYY MMM D for asian countries, however, as I don't have direct experience / knowledge, it might be best to ask around our community and confirm before adding that in.

datitisev:
What about using localized times, so we keep the logic in `moment`?

image

askvortsov1:
> What about using localized times, so we keep the logic in `moment`?
> 
> image

I think we already have that in all places where we display a time?

askvortsov1:
SO for some reason my commits are registering as being from 2021 (?????), but the latest version now uses ll

datitisev:
i am so confused

image

askvortsov1:
Just to clarify, the sum effect of these time traveling commits is to change 1 line of code to use 'll' instead of 'MMM YY' for posts older than a year

franzliedke:
> SO for some reason my commits are registering as being from 2021 (?????), but the latest version now uses ll

Haha, maybe GitHub tries to mess with you when your PR mentions "date" or "time"? :rofl:

askvortsov1:
I'll make a new PR to avoid the date confusion. Essentially my clock was off because I was resting how the dates displayed

askvortsov1:
Closing in favor of #2034

PR #2027, opened 2020-02-27
Add Updated Gambit

askvortsov1:
**Fixes Part Of flarum/issue-archive#418**

- Add Updated gambit

**Confirmed**

### DIFF
--- a/src/Discussion/Search/Gambit/UpdatedGambit.php
+++ b/src/Discussion/Search/Gambit/UpdatedGambit.php
@@ -1,0 +1,43 @@
+<?php
+
+/*
+ * This file is part of Flarum.
+ *
+ * For detailed copyright and license information, please view the
+ * LICENSE file that was distributed with this source code.
+ */
+
+namespace Flarum\Discussion\Search\Gambit;
+
+use Flarum\Discussion\Search\DiscussionSearch;
+use Flarum\Search\AbstractRegexGambit;
+use Flarum\Search\AbstractSearch;
+use LogicException;
+
+class UpdatedGambit extends AbstractRegexGambit
+{
+    /**
+     * {@inheritdoc}
+     */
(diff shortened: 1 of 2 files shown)


src/Discussion/Search/Gambit/UpdatedGambit.php
```diff
@@ -0,0 +1,43 @@
...
+        if (! $search instanceof DiscussionSearch) {
+            throw new LogicException('This gambit can only be applied on a DiscussionSearch');
+        }
+
+        // If we've just been provided with a single YYYY-MM-DD date, then find
+        // discussions that were started on that exact date. But if we've been
+        // provided with a YYYY-MM-DD..YYYY-MM-DD range, then find discussions
+        // that were last updated during that period.
+        if (empty($matches[3])) {
+            $search->getQuery()->whereDate('last_posted_at', $negate ? '!=' : '=', $matches[1]);
```
franzliedke: Can we ignore (or completely prevent) the negation flag here? I don't really see any use-case, but more importantly, if I am not mistaken, negating would mean that the index on `last_posted_at` could not be used.

franzliedke (commented):
Because you're so productive... :grin: 

Can I ask you for a test case for this one? We already have a few integration tests for the search endpoint, if I am not mistaken.

src/Discussion/Search/Gambit/UpdatedGambit.php
askvortsov1: This is pretty much copied from the CreatedGambit. I can def do that, but should I ignore it there as well?

askvortsov1:
> Because you're so productive... 
> 
> Can I ask you for a test case for this one? We already have a few integration tests for the search endpoint, if I am not mistaken.

Absolutely! But before I do that, can we decide whether or not we want this gambit as per https://github.com/flarum/core/issues/256#issuecomment-591916018?

datitisev:
Not stale, just not a priority.

askvortsov1:
Let's keep this around for now

PR #2036, opened 2020-02-29
Fixed extenders added after app initialized not working

### LINKED ISSUE #2035: Integration Testcase: adding extenders after app has been initialized doesn't work.
## Bug Report

**Current Behavior**
Extenders are only applied when the app is being initialized for the first time in the testcase. This means that if extenders are added after any database calls have been made, everything breaks

askvortsov1:
**Fixes #2035 **

If extenders are added in a test case after the app has been initalized, extend the app with them. Right now, they do nothing, which makes it impossible to do anything in the database prior to initializing the app.

### DIFF
--- a/tests/integration/TestCase.php
+++ b/tests/integration/TestCase.php
@@ -57,7 +57,11 @@ abstract class TestCase extends \PHPUnit\Framework\TestCase
 
     protected function extend(ExtenderInterface $extender)
     {
-        $this->extenders[] = $extender;
+        if (is_null($this->app)) {
+            $this->extenders[] = $extender;
+        } else {
+            $extender->extend($this->app->getContainer());
+        }
     }
 
     /**


franzliedke:
The problem is that some extenders (current or future) may have no effect when the app is already booted (e.g. when a binding has already been resolved from the container, and injected elsewhere before being able to extend it), so I'd prefer to avoid this.

For your concrete problem in testing mail extenders, you should be able to circumvent this by moving the `prepareDatabase()` call (essentially a one-liner) to each test case, and not using it in `setUp()`. A bit of duplication, but it avoids the potentially surprising behavior explained above.

(Of course, we could also register the extender in `setUp()`, but then we couldn't do the baseline case of showing what is the default behavior without extenders, which would be sad IMO.)

If the problem comes up again and the workaround gets more annoying, we can revisit this.

askvortsov1:
@franzliedke this was the first solution I tried, but I found it repetitive and unintuitive. Im fine with this being closed for now, considering it's an implementation detail, but I think we need to make our test cases as easy as possible to create and manage, considering that our test suite is somewhat sparse atm.

askvortsov1:
@franzliedke to follow up on this, I want to add a documentation section explaining this. I was also thinking of reinstantiating "app" if it is already instantiated in the extend method, that would preserve the functionality of this request without the issue you identified above.

askvortsov1:
Superceded by https://github.com/flarum/core/pull/2052.

PR #2039, opened 2020-02-29
Improved UI of Switch with loading indicator

### LINKED ISSUE #1491: Loading state of Switch is hard to see
## Bug Report

**Current Behavior**
After changing a setting using a switch at settings, the loader that warns the user the setting is applying appears over the switch instead of appearing next to it, creating a bad looking UI.

**Steps to Reproduce**
1. Go to settings
2. Click on change one of the options with a switch
3. See error

**Expected Behavior**
The loader should appear next to the switch instead of over it.

**Screenshots**

Please don't mind the screenshots, I was able to take these even hard because of **the speed of Flarum** 😉 (the loader was going away immediately)

**Environment**
Flarum Community
https://discuss.flarum.org

**Possible Solution**
Move loader next to the switch.

askvortsov1:
**Fixes #1491**

Moved loading indicator outside of switch slider on settings page to improve ui

Seems trivial.

**Screenshot**
https://i.imgur.com/6A8pGfZ.png
https://i.imgur.com/X4q7dnP.png

**Confirmed**

### DIFF
--- a/less/common/Checkbox.less
+++ b/less/common/Checkbox.less
@@ -67,7 +67,7 @@
     color: @muted-more-color !important;
   }
 }
-.loading.Checkbox--switch {
+.Checkbox--switch.loading {
   color: @muted-more-color !important;
 
   &.on .Checkbox-display {
--- a/less/common/Checkbox.less
+++ b/less/common/Checkbox.less
@@ -67,7 +67,7 @@
     color: @muted-more-color !important;
   }
 }
-.Checkbox.loading.Checkbox--switch {
+.loading.Checkbox--switch {
   color: @muted-more-color !important;
 
   &.on .Checkbox-display {
--- a/less/common/Checkbox.less
+++ b/less/common/Checkbox.less
@@ -3,7 +3,7 @@
   cursor: pointer;
   margin: 0;
 
-  & input[type=checkbox] {
+  & input[type="checkbox"] {
     display: none;
   }
 }
@@ -32,7 +32,7 @@
   .transition(background-color 0.2s);
 
   .on& {
-    background: #58A400;
+    background: #58a400;
   }
 
   &:before {
@@ -48,7 +48,7 @@
     }
   }
   &:before {
-    content: ' ';
+    content: " ";
     background: @body-bg;
     border-radius: 11px;
     box-shadow: 0 2px 4px @shadow-color;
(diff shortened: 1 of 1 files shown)


datitisev:
I don't think this is the fix we want(ed) to go with. Apart from content jumping around horizontally when the loading indicator appears - which was mentioned by Franz, I think Franz's suggestion is a better way to tackle this.

> Since I would like to avoid page content jumping around (which might happen with the indicator appearing next to the toggle button), what about adding a semi-transparent overlay to the button while it is saving the state, with the indicator on top? That way, it would be more visible...
> \- https://github.com/flarum/core/issues/1491#issuecomment-402843307

askvortsov1:
Not 100% sure what you mean by 'jumping around horizontally': at least from my testing, the indicator doesn't displace anything, it just shows up. I'm not sure that making that switch (more) transparent would result in a more pleasant UI

datitisev:
If the switch has items before it, it will not look well. I think what Franz was thinking of was having the loading indicator on top of the whole component with a semi-translucent background.

askvortsov1:
Ah yeah that'd do it. So we could do Franz's idea with an overlay and the indicator in the middle. Alternatively, we could also move the indicator to the other side of the switch when it's going from on to off so that it's always on a white background.

datitisev:
I think that'd get a bit confusing. Putting the loading over the whole thing is probably the best idea.

askvortsov1:
Updated, how's this?

https://youtu.be/H2Yzu5U8574

datitisev:
Still a bit hard to read IMO, the background might still be too transparent.

askvortsov1:
Just to confirm, are you proposing more or less transparency on the switch?

datitisev:
Uh I'd say more transparent? The loading indicator isn't very visible right now.

askvortsov1:
@datitisev  how's this: https://youtu.be/hYxWwHgtd6k

datitisev:
I think that's good? It may still be hard to view for some people, but I'd rather have other members of the team comment their thoughts before making you do more work.

tankerkiller125:
I'm happy with the latest revision, it looks much better, the only thing I could possible think to improve is make the actual loading wheel just a tad darker maybe?

askvortsov1:
How's this: https://youtu.be/4JJUz0MC6YQ? Too dark?

datitisev:
@askvortsov1 Looks fine to me. How does it look in dark mode?

image

askvortsov1:
Good catch, thanks! How's this: https://youtu.be/xeuNbixI-sU?

tankerkiller125:
I'm happy with it now

PR #2044, opened 2020-03-03
Deprecate AssertPermissionTrait

### LINKED ISSUE #1320: Move AssertPermissionTrait methods directly onto the User model
```php
$actor->assertGuest();
$actor->assertAdmin();
$actor->assertRegistered();
$actor->assertCan('createGroup'); // via AuthenticableTrait
```

askvortsov1:
**Fixes #1320 **

Deprecate the AssertPermissionTrait. We should merge this in the next release to give extension authors time to adjust prior to stable.

**Confirmed**

### DIFF
--- a/src/Api/Controller/ListNotificationsController.php
+++ b/src/Api/Controller/ListNotificationsController.php
@@ -13,14 +13,11 @@ use Flarum\Api\Serializer\NotificationSerializer;
 use Flarum\Discussion\Discussion;
 use Flarum\Http\UrlGenerator;
 use Flarum\Notification\NotificationRepository;
-use Flarum\User\AssertPermissionTrait;
 use Psr\Http\Message\ServerRequestInterface;
 use Tobscure\JsonApi\Document;
 
 class ListNotificationsController extends AbstractListController
 {
-    use AssertPermissionTrait;
-
     /**
      * {@inheritdoc}
      */
@@ -67,7 +64,7 @@ class ListNotificationsController extends AbstractListController
     {
         $actor = $request->getAttribute('actor');
 
-        $this->assertRegistered($actor);
+        $actor->assertRegistered();
 
         $actor->markNotificationsAsRead()->save();
 
--- a/src/Api/Controller/SendConfirmationEmailController.php
+++ b/src/Api/Controller/SendConfirmationEmailController.php
@@ -12,7 +12,6 @@ namespace Flarum\Api\Controller;
 use Flarum\Http\UrlGenerator;
 use Flarum\Mail\Job\SendRawEmailJob;
 use Flarum\Settings\SettingsRepositoryInterface;
-use Flarum\User\AssertPermissionTrait;
 use Flarum\User\EmailToken;
 use Flarum\User\Exception\PermissionDeniedException;
 use Illuminate\Contracts\Queue\Queue;
@@ -25,8 +24,6 @@ use Symfony\Component\Translation\TranslatorInterface;
 
 class SendConfirmationEmailController implements RequestHandlerInterface
 {
-    use AssertPermissionTrait;
-
     /**
      * @var SettingsRepositoryInterface
      */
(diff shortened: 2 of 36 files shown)


src/Group/Command/CreateGroupHandler.php
```diff
@@ -47,8 +45,8 @@ public function handle(CreateGroup $command)
         $actor = $command->actor;
         $data = $command->data;
 
-        $this->assertRegistered($actor);
-        $this->assertCan($actor, 'createGroup');
+        $actor->assertRegistered();
+        $actor->assertCan('createGroup');
```
clarkwinkelmann, line 49: it's odd to have two checks. there should be single policy for this.

clarkwinkelmann:
What were the motivations behind this change ? Why get rid of `AssertPermissionTrait` ? This puts even more code into the `User` model :thinking:

src/Group/Command/CreateGroupHandler.php
franzliedke: It's intentional, in order to provide different status codes depending on the exact situation. See commit 3b5691ee2807bcc917b89f86ba99f4a4725095dc.

franzliedke:
I believe the idea is encapsulation. These methods all operate on the user/actor objects, so it makes sense that they are owned by the `User` class.

Yes, this adds more code to `User`, but as I argue above, it kinda belongs there - to the part of it that is concerned with "actor" stuff. The class inherits from the Eloquent model god class, so this is merely a drop in the bucket. :laughing: 

On a more serious note. I could see us reducing or at least separating the responsibilities by introducing an *interface* for actor-related things. That would help in e.g. at least decoupling `Guest` from all the Eloquent stuff. That might help simplify certain tests.

src/Api/Controller/ListNotificationsController.php
```diff
@@ -67,7 +64,7 @@ protected function data(ServerRequestInterface $request, Document $document)
     {
         $actor = $request->getAttribute('actor');
 
-        $this->assertRegistered($actor);
+        $request->getAttribute('actor')->assertRegistered();
```
franzliedke, line 67: This can use the `$actor` variable.

src/Api/Controller/SendConfirmationEmailController.php
```diff
@@ -69,7 +66,7 @@ public function handle(ServerRequestInterface $request): ResponseInterface
         $id = Arr::get($request->getQueryParams(), 'id');
         $actor = $request->getAttribute('actor');
 
-        $this->assertRegistered($actor);
+        $request->getAttribute('actor')->assertRegistered();
```
franzliedke, line 69: This can use the `$actor` variable.

luceos:
> Yes, this adds more code to User, but as I argue above, it kinda belongs there

@franzliedke can't we move this out of the User class into a Concerns/trait please? It's what I proposed in the user preferences as well and it would help keep Users clean.

franzliedke:
@luceos We could. *Personally*, I rarely see the benefit in extracting a trait if it's not for the purpose of sharing code. In this case, we would just try to conceal the fact that the class is too big / doing too much. On the downside, I tend to be surprised by traits, because I often forget to consider them (or simply overlook them) when searching for a method's implementation.

Not too strong an opinion, curious to hear what others think.

askvortsov1:
I also see traits as a mechanism for sharing functionality among unrelated classes moreso than a way to split up big classes :+1:

franzliedke:
@askvortsov1 That did not go well. :scream:

askvortsov1:
@franzliedke here we go :D :see_no_evil:

src/Api/Controller/SendTestMailController.php
```diff
@@ -39,7 +36,7 @@ public function __construct(Container $container, Mailer $mailer, TranslatorInte
     public function handle(ServerRequestInterface $request): ResponseInterface
     {
         $actor = $request->getAttribute('actor');
-        $this->assertAdmin($actor);
+        $actor->assertAdmin($actor);
```
franzliedke, line 39: That param should be removed. :wink:

franzliedke:
Any volunteers for applying this change to all bundled extensions? :grin:

tankerkiller125:
@franzliedke I'll take care of it this weekend once I get done meeting the dog I might be adopting.

PR #2046, opened 2020-03-03
Cleaned up discussion idWithSlug handling

### LINKED ISSUE #1360: Generate and store discussion slugs on server-side
As per https://github.com/flarum/core/pull/1351#pullrequestreview-95053052

> General question, though: should we maybe move this calculation onto a further attribute computed in the Discussion model on the server-side? Then we don't need to re-create the logic twice.

askvortsov1:
**Fixes #1360**

- Moved all discussion idWithSlug generation logic to backend.

- Should we still send the raw slug itself to the frontend?

**Confirmed**

### DIFF
--- a/views/frontend/content/index.blade.php
+++ b/views/frontend/content/index.blade.php
@@ -5,21 +5,19 @@
 
     <ul>
         @foreach ($apiDocument->data as $discussion)
-            <li>
-                <a href="{{ $url->to('forum')->route('discussion', [
-                    'id' => $discussion->id . (trim($discussion->attributes->slug) ? '-' . $discussion->attributes->slug : '')
-                ]) }}">
-                    {{ $discussion->attributes->title }}
-                </a>
-            </li>
+        <li>
+            <a href="{{ $url->to('forum')->route('discussion', ['id' => $discussion->attributes->idWithSlug]) }}">
+                {{ $discussion->attributes->title }}
+            </a>
+        </li>
         @endforeach
     </ul>
 
     @if (isset($apiDocument->links->prev))
-        <a href="{{ $url->to('forum')->route('index') }}?page={{ $page - 1 }}">&laquo; {{ $translator->trans('core.views.index.previous_page_button') }}</a>
+    <a href="{{ $url->to('forum')->route('index') }}?page={{ $page - 1 }}">&laquo; {{ $translator->trans('core.views.index.previous_page_button') }}</a>
     @endif
 
     @if (isset($apiDocument->links->next))
-        <a href="{{ $url->to('forum')->route('index') }}?page={{ $page + 1 }}">{{ $translator->trans('core.views.index.next_page_button') }} &raquo;</a>
+    <a href="{{ $url->to('forum')->route('index') }}?page={{ $page + 1 }}">{{ $translator->trans('core.views.index.next_page_button') }} &raquo;</a>
     @endif
 </div>
--- a/views/frontend/content/index.blade.php
+++ b/views/frontend/content/index.blade.php
@@ -14,10 +14,10 @@
     </ul>
 
     @if (isset($apiDocument->links->prev))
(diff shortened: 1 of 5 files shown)


franzliedke:

askvortsov1:
Keeping this logic on the backend could be useful in preparation for transliteration changes.

askvortsov1:
There's one question I have before merging: If we use this as a base for transliteration drivers in discussion slugs, perhaps it should be called something different in the API ('fullSlug', perhaps?) As @franzliedke pointed out in #2049, the REST API is something that we are somewhat commited to support.

tankerkiller125:
@askvortsov1 I'm happy with `fullSlug`

views/frontend/content/index.blade.php
```diff
@@ -5,21 +5,19 @@
...
-        <a href="{{ $url->to('forum')->route('index') }}?page={{ $page - 1 }}">&laquo; {{ $translator->trans('core.views.index.previous_page_button') }}</a>
+    <a href="{{ $url->to('forum')->route('index') }}?page={{ $page - 1 }}">&laquo; {{ $translator->trans('core.views.index.previous_page_button') }}</a>
     @endif
 
     @if (isset($apiDocument->links->next))
-        <a href="{{ $url->to('forum')->route('index') }}?page={{ $page + 1 }}">{{ $translator->trans('core.views.index.next_page_button') }} &raquo;</a>
+    <a href="{{ $url->to('forum')->route('index') }}?page={{ $page + 1 }}">{{ $translator->trans('core.views.index.next_page_button') }} &raquo;</a>
     @endif
-</div>
+</div>
```
clarkwinkelmann, line 23: You removed the ending newline.

Also are those indentation changes above an intended change?

askvortsov1: Fixed, sorry.

askvortsov1:
> You want to rename `idWithSlug` into `fullSlug` ? I think `idWithSlug` makes more sense here. It clearly says what it is, and also sounds like it would be a generated value.

@clarkwinkelmann  For this particular instance, idWithSlug makes more sense. But if we want to return one value for transliteration, naming it 'fullSlug' might be better, especially if a transliteration driver uses just the ID

askvortsov1:
@clarkwinkelmann But in the same way, if the transliteration strategy does not include a slug at all, wouldn't idWithSlug also be incorrect?

tankerkiller125:
Another answer to the name of the object could be `fullPath` since technically we are setting the path.

clarkwinkelmann:
@askvortsov1 I think this needs discussing in a new issue. Right now routing is fully ID based, and the ID must always be the first part of the URL. Slug is only there for SEO and is irrelevant to routing.

If you want to propose a change in how discussion routing works, a new issue would be best I think.

If we're only discussing names without a change in how it works, maybe one could be called the "minimal" URL (just id) and the other the "canonical" or "friendly" URL (id + slug) ? Maybe `friendlyId` ?

askvortsov1:
Sorry, I don't think I'm wording this well. I recognize that this idWithSlug is only there for SEO, but if we want to have it vary depending on the transliteration scheme, imo we should call it something generic enough that its not tied to any given implementation. I apologize for all the bikeshedding here, I just think that when we choose a name that goes in core's public REST API, we should support it. As per https://en.wikipedia.org/wiki/Clean_URL, perhaps 'cleanId' (or 'friendlyId' as you suggest) could be the way to go here.

clarkwinkelmann:
IMO, the word "id" needs to be in it, because the URL structure is `/resources/<id>`. I fear "path" might give the wrong impression that it could contain the full path and not just the last ID part.

So yeah, maybe one of `friendlyId`, `cleanId`, or maybe `seoId` ?

Even though the Wikipedia page makes it very clear what "clean URL" mean, I feel like it sounds like `cleanId` would be a shorter version of `id`, but in our case it's quite the opposite.

askvortsov1:
@flarum/core Do we have a preference for friendlyId or seoId?

tankerkiller125:
@askvortsov1 between those two I vote `seoId`

askvortsov1:
Superceded by https://github.com/flarum/core/pull/2456

PR #2047, opened 2020-03-03
Add index page title, add mechanism to clear title from defaultRoute.

### LINKED ISSUE #495: Add titles for pages that lack them
Some pages show nothing but the forum name in the title bar. The following titles can be added:
- "All Discussions" for the All Discussions page
- "Tags" for the Tags page
- The tag name when a tag-specific discussion list is displayed

In addition, a routine is needed to omit the title of whatever page is set as the site homepage.

askvortsov1:
**Fixes #495**

Add index page title, add mechanism to clear title from defaultRoute.

**Confirmed**

### DIFF
--- a/js/src/common/Application.js
+++ b/js/src/common/Application.js
@@ -211,11 +211,7 @@ export default class Application {
   }
 
   updateTitle() {
-    if (m.route() == '/') {
-      document.title = this.forum.attribute('title');
-    } else {
-      document.title = (this.titleCount ? `(${this.titleCount}) ` : '') + (this.title ? this.title + ' - ' : '') + this.forum.attribute('title');
-    }
+    document.title = (this.titleCount ? `(${this.titleCount}) ` : '') + (this.title && m.route() !== '/' ? this.title + ' - ' : '') + this.forum.attribute('title');
   }
 
   /**
--- a/js/src/common/Application.js
+++ b/js/src/common/Application.js
@@ -211,8 +211,10 @@ export default class Application {
   }
 
   updateTitle() {
-    document.title =
-      (this.titleCount ? `(${this.titleCount}) ` : '') + (this.title && m.route() !== '/' ? this.title + ' - ' : '') + this.forum.attribute('title');
+    const count = this.titleCount ? `(${this.titleCount}) ` : '';
+    const pageTitleWithSeparator = this.title && m.route() !== '/' ? this.title + ' - ' : '';
+    const title = this.forum.attribute('title');
+    document.title = count + pageTitleWithSeparator + title;
   }
 
   /**
--- a/js/src/common/Application.js
+++ b/js/src/common/Application.js
@@ -211,7 +211,8 @@ export default class Application {
   }
 
   updateTitle() {
-    document.title = (this.titleCount ? `(${this.titleCount}) ` : '') + (this.title && m.route() !== '/' ? this.title + ' - ' : '') + this.forum.attribute('title');
+    document.title =
+      (this.titleCount ? `(${this.titleCount}) ` : '') + (this.title && m.route() !== '/' ? this.title + ' - ' : '') + this.forum.attribute('title');
   }
 
   /**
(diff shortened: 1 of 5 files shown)


franzliedke:

clarkwinkelmann:
I believe this PR only changes the behavior on the javascript side, the title on the PHP/preload side will still include the page title?

This does not need to be the same PR, but changing one and not the other wouldn't make sense.

askvortsov1:
Done, all good points. Thanks!

Also, now that Document has access to request, perhaps we should just use the current url for canonical url (in a separate PR)? That's often what we do anyway...

src/Frontend/Document.php
```diff
@@ -173,7 +180,9 @@ protected function makeView(): View
      */
     protected function makeTitle(): string
     {
-        return ($this->title ? $this->title.' - ' : '').Arr::get($this->forumApiDocument, 'data.attributes.title');
+        $onHomePage = rtrim($this->request->getAttribute('originalUri')->getPath(), '/') === '';
```
clarkwinkelmann, line 183: Does `originalUri` means Flarum in a subdomain will still show the title for its index page?

askvortsov1: Considering that in https://github.com/askvortsov1/core/blob/58244a131ebf07ed6085052ce112ec77ad114e8a/src/Foundation/InstalledApp.php#L58-L67, "/" maps to the base path, I'm assuming that's not a player with OriginalMessages as well? (since that's just calling getUri)

clarkwinkelmann: Oh right, the originalUri will be the original before the router, but after the basepath. So it should be the same value for `->getUri()` and `->getAttribute('originalUri')` when under the `flarum.forum.handler` router ?

In this case why not simply use `getUri()` as most documents will be in the forum router anyway?

Or is it about the admin homepage? Do we want to handle that page like another homepage or like a different non-homepage?

askvortsov1: originalUri is more explicit, but then again, I'm not sure that the admin dashboard would ever be set as the homepage on any reasonable forum... getUrl would probably be cleaner tbh.

js/src/common/Application.js
```diff
@@ -211,7 +211,8 @@ export default class Application {
   }
 
   updateTitle() {
-    document.title = (this.titleCount ? `(${this.titleCount}) ` : '') + (this.title ? this.title + ' - ' : '') + this.forum.attribute('title');
+    document.title =
+      (this.titleCount ? `(${this.titleCount}) ` : '') + (this.title && m.route() !== '/' ? this.title + ' - ' : '') + this.forum.attribute('title');
```
datitisev, line 215: Perhaps we want to make this easier to read?

Add `count`, `separator`, and `title` variables that replace these parentheses, and just add them together `count + separator + title` when setting to document.title.

datitisev: You know what's funny? I just realized we could use the class list util (or classnames lob) to simplify this code further. Obv were not gonna do it, but... that's funny.

src/Frontend/Document.php
```diff
@@ -131,14 +132,20 @@ class Document implements Renderable
...
+     * @var Request
+     */
+    protected $request;
+
     /**
      * @param Factory $view
      * @param array $forumApiDocument
      */
-    public function __construct(Factory $view, array $forumApiDocument)
+    public function __construct(Factory $view, array $forumApiDocument, Request $request)
```
datitisev, line 144: Could this be considered a breaking change?

datitisev: Never mind, looks like it's only initialized in one place.

clarkwinkelmann: Well it would be a breaking change if an extension decided to extend it... Though I don't see why an extension would try to extend that class and I don't think anyone does it :sweat_smile:

datitisev: Yeah, I didn't look at other uses of Frontend\Document until after commenting, which is why I resolved the conversation.

PR #2048, opened 2020-03-04
WIP: Custom ServerRequest Subclass

### LINKED ISSUE #869: Switch to a custom subclass of ServerRequest

``` php
// etc.
```

``` php
namespace Flarum\Http;

{
    public function getActor();
    public function withActor();
    public function getSession();
    // etc.
}
```

askvortsov1:
**Fixes #869**

Switch to custom subclass of ServerRequest

WIP

**Confirmed**

### DIFF
--- a/src/Http/ServerRequest.php
+++ b/src/Http/ServerRequest.php
@@ -42,26 +42,26 @@ class ServerRequest extends BaseServerRequest
 
     public function withBypassFloodgate(bool $bypassFloodgate)
     {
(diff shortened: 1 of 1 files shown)


askvortsov1:
1st thing: Are there any flarum-specific attributes that I forgot / should not have included?

franzliedke:
I think I'd rather hold off on this exact approach, as explained in https://github.com/flarum/core/issues/869#issuecomment-612838885. Please let me know what you think.

PR #1948, opened 2019-12-03
add meta description

xmugenx:
**Fixes #0000**

**Changes proposed in this pull request:**
add setDescription to Application.js for meta description.

**Confirmed**

- [ x ] Frontend changes: tested on a local Flarum installation.
- [ x ] Backend changes: tests are green (run `composer test`).

**Required changes:**

- [ x ] Related core extension PRs: flarum/tags

PR #2049, opened 2020-03-04
Added excerpt method to Discussion, and meta description to description page.

askvortsov1:
**Cleans up #1948**

- Added length-configurable exerpt method to discussion (could be used in projects like flarum summaries)
- Added meta description to description page 

**Confirmed**

### DIFF
--- a/src/Api/Serializer/BasicDiscussionSerializer.php
+++ b/src/Api/Serializer/BasicDiscussionSerializer.php
@@ -36,6 +36,7 @@ class BasicDiscussionSerializer extends AbstractSerializer
         return [
             'title' => $discussion->title,
             'slug'  => $discussion->slug,
+            'excerpt' => $discussion->excerpt(),
         ];
     }
 
--- a/src/Discussion/Discussion.php
+++ b/src/Discussion/Discussion.php
@@ -454,4 +454,18 @@ class Discussion extends AbstractModel
         $this->attributes['title'] = $title;
         $this->slug = Str::slug($title);
     }
+
+    /**
+     * Generates a brief excerpt based off of the first post
+     *
+     * @param int length (Defaults to 160)
+     * @return string
+     */
+    public function excerpt($length = 160)
+    {
+        $relevantPost = $this->mostRelevantPost ?: $this->firstPost ?: $this->comments->first();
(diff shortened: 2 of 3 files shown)


franzliedke:
Please fix the failing test - it's probably not set up completely, now that we expose more content in the API...

askvortsov1:
> Please fix the failing test - it's probably not set up completely, now that we expose more content in the API...

Done, I failed to include support for discussions with no posts in my original PR. Fixed now.

franzliedke:
> discussions with no posts 

Is that a thing that's going to happen, though? :thinking: Or simply bad setup in our tests?

src/Api/Serializer/BasicDiscussionSerializer.php
```diff
@@ -36,6 +36,7 @@ protected function getDefaultAttributes($discussion)
         return [
             'title' => $discussion->title,
             'slug'  => $discussion->slug,
+            'excerpt' => $discussion->excerpt(),
```
franzliedke, line 39: What I hadn't considered when [suggesting to encapsulate this in the model](https://github.com/flarum/core/pull/1948#issuecomment-578299506) is that we now expose this in the API, even though it's sort of an implementation detail. And I strongly dislike adding things to APIs unless we have a direct use-case for them - you are bound to support them forever.

Thoughts?

askvortsov1:
Even if it happens, imo it shouldn't break our system, so I set it to return an empty string for the excerpt in that case.

src/Api/Serializer/BasicDiscussionSerializer.php
askvortsov1: While I generally agree with the sentiment, and also think that bloating core's API can be problematic, I also have a hard time imagining a situation where discussions cease to have text and this is no longer relevant.

I don't suppose there's any other way we could get the description instance itself into Forum\Content\Document?

askvortsov1: Just thought of something while trying to figure out #2177: The Tag Content pulls in an instance of the tag: https://github.com/askvortsov1/tags/blob/285622d51999d012e368d9d82a0e8f5bbe4777ef/src/Content/Tag.php#L44-L71

We could do the same here, then we don't need add anything to the API at all.

PR #2052, opened 2020-03-06
Add Authenticated Test Case utility

askvortsov1:
**Organized Workaround for  #2035**

Provides a superclass for integration tests that require authentication on requests. Supports both admin and non-admin permissions. Compliant with https://github.com/flarum/core/pull/2036#issuecomment-595734221. Makes integration tests a LOT more easier. Also, probable base for moving the Controller integration tests onto using requests, not directly calling the controller.

### DIFF
--- a/tests/integration/AuthenticatedTestCase.php
+++ b/tests/integration/AuthenticatedTestCase.php
@@ -1,0 +1,73 @@
+<?php
+
+/*
+ * This file is part of Flarum.
+ *
+ * For detailed copyright and license information, please view the
+ * LICENSE file that was distributed with this source code.
+ */
+
+namespace Flarum\Tests\integration;
+
+use Carbon\Carbon;
+use Flarum\Api\ApiKey;
+use Illuminate\Support\Str;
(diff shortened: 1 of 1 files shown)


franzliedke:
Squash-merged this to master with its StyleCI fixes.

Please note I will likely make some tweaks to this, so your PRs building on top of this might have to change a little as well. Will sum up those changes here in a bit.

franzliedke:
@askvortsov1 Done. Please check out commit 25f772c1ea36053c9ce78df6fbc419eceae0f1b6.

askvortsov1:
Awesome, thanks! I'll update my PRs. 

@franzliedke I also created a similar system in https://github.com/flarum/core/pull/2057/files. After consideration I definitely need to split that into a separate PR, but should I also move that into a trait?

franzliedke:
@askvortsov1 I think having those parent classes for different *types* of tests (console in this case) makes sense, no trait needed there. I just wanted to avoid deeper type hierarchies than that.

Thanks for asking!

PR #2056, opened 2020-03-08
WIP Policy Refactor, Extender

askvortsov1:
Foundation for fixing #1832
Part of #1891

- Ask that policies return ->allow, ->deny, ->forceAllow, ->forceDeny instead of true/false to increase flexibility
- Improve evaluation of whether a user has the ability to do something on an instance to avoid order-dependence.
- Rename getPermission (->checkAbility), scopeModelVisibility (->scopeQueryListener) methods, as well as the methods scopeModelVisibility calls (->scopeQuery, ->scopeQueryPerAbility), for purposes of clarify.
- Restructure old scopeModelVisibility calls down to only 2 methods
- Refactor how policies are registered by moving them into the User service provider, making an Extender more viable
- Removed the GroupServiceProvider as it no longer does anything
- Deprecate everything that's no longer used
- Add Policy extender
- Added integration tests for policy extender (dependent on https://github.com/flarum/core/pull/2052)

- I don't like that the gate->before callback now requires us to app->make ALL the policies every time we want to check a permission. The same applies to the part where we listen to ScopeModelVisibility. Should I add a singleton that contains resolved instances of all policies?
- Are the new names what we want?
- Do we want to get rid of ScopeModelVisibility?

**TODO**
If this is accepted and merged, bundled extensions will need to be updated.

**Confirmed**


askvortsov1:
Not stale.

src/User/AbstractPolicy.php
```diff
@@ -30,8 +101,7 @@ public function subscribe(Dispatcher $events)
     }
 
     /**
-     * @param GetPermission $event
-     * @return bool|void
+     * @deprecated in favor of checkAbility
      */
     public function getPermission(GetPermission $event)
     {
```
SychO9, line 107: If this is deprecated, why don't we just call `checkAbility` from here, instead of having duplicated blocks of code.

src/User/AbstractPolicy.php
```diff
@@ -53,7 +123,7 @@ public function getPermission(GetPermission $event)
     }
 
     /**
-     * @param ScopeModelVisibility $event
+     * @deprecated
      */
     public function scopeModelVisibility(ScopeModelVisibility $event)
     {
```
SychO9, line 129: same here, why not just call `scopeQueryListener`

src/User/UserServiceProvider.php
```diff
@@ -61,6 +74,26 @@ protected function registerAvatarsFilesystem()
...
+            $evaluationCriteria = [
+                AbstractPolicy::$FORCE_DENY => false,
+                AbstractPolicy::$FORCE_ALLOW => true,
+                AbstractPolicy::$DENY => false,
+                AbstractPolicy::$ALLOW => true,
+            ];
+
+            $results = [];
+            foreach ($this->app->make('flarum.policies') as $policy) {
+                $results[] = $this->app->make($policy)->checkAbility($actor, $ability, $model);
```
SychO9, line 86: Before this refactor, when policies were listeners, they were resolved only once.

How about we resolve them all in an array before making the gate and then we use that array of all resolved policies inside.

src/User/UserServiceProvider.php
```diff
@@ -92,9 +127,12 @@ public function boot()
         $events->listen(EmailChangeRequested::class, EmailConfirmationMailer::class);
 
         $events->subscribe(UserMetadataUpdater::class);
-        $events->subscribe(UserPolicy::class);
 
         $events->listen(ConfigureUserPreferences::class, [$this, 'configureUserPreferences']);
+
+        foreach ($this->app->make('flarum.policies') as $policy) {
+            $events->listen(ScopeModelVisibility::class, [$this->app->make($policy), 'scopeQueryListener']);
```
SychO9, line 134: same here, let's resolve all policies once in an array before making the gate and use that.

SychO9 (changes requested):

PR #2058, opened 2020-03-09
Add extension management console commands

askvortsov1:
There should be an ability to manage extensions programatically.

Our goal for Flarum Core is to provide a base for extensions to work with. As such, it makes sense to try to build infrastructure that makes working with extensions easier.

This will also result in a lot of happy users who don't need to manually disable every single extension before upgrading going forward.

Depends on (https://github.com/flarum/core/pull/2068)

Added console commands for:
- Enabling an extension
- Enabling all extensions
- Disabling an extension
- Disabling all extensions

- Anything I missed?

**Confirmed**


luceos:
I actually use enable all in several forums. Especially helpful if you are used to DTAP. Not having looked at code, I think the idea is awesome. One thing we have to watch for is that we provide too many commands natively.

askvortsov1:
I can add it in. Considering the other extension safety checks I'm planning on adding, it shouldnt be too much of an issue imo.

Also, while I recognize that we want to avoid bloat, IMO this is something that we want in core, because then automated systems can assume that it is always available. If it were an extension, well, then there'd need to be a programmatic way of enabling that extension, and then we're back to square one

luceos:
Here's a programmatic way of enabling extensions:

https://discuss.flarum.org/d/21402-enable-a-extension-from-cli/2

Safety check can be to ask for confirmation or ignore that when the `--force` flag is used; this is the standard for Laravel commands as well.

askvortsov1:
Ah I see. Although I still feel like this is something we should offer in core. I could add a safety check to have users confirm (I'm assuming for disableAll and enableAll they'd confirm every single extension?)

askvortsov1:
> I actually use enable all in several forums. Especially helpful if you are used to DTAP.

Added! What's DTAP just out of curiosity?

> Safety check can be to ask for confirmation or ignore that when the `--force` flag is used; this is the standard for Laravel commands as well.

I like this feature a lot. Only thing is, we're going to need to use interactive questions (->ask, ->secret, ->validationError), which currently aren't included in the Abstract Command (but exist in https://github.com/flarum/core/blob/master/src/Install/Console/UserDataProvider.php, as well as https://github.com/flarum/core/pull/2021/files#).

Since these are likely to be used across a lot of commands, I want to provide native support. @luceos Should I add these to AbstractCommand (better imo), or set up a trait for them?

luceos:
A trait would work better

DTAP means development, testing, acceptance, production. It's a reference to multi stage development.

askvortsov1:
@luceos  I added confirmation (bypassable by -y or --yes). Is this what you envisioned?

Also, just to clarify, this depends on https://github.com/flarum/core/pull/2068

luceos:
Why not force? If that's the convention, why run your own show? Just curious on the flow of thought.. 😇

askvortsov1:
> Why not force? If that's the convention, why run your own show? Just curious on the flow of thought..

I'm not sure that force is the convention for what this does. For example, in cp, force means overwriting without prompting. -y or --yes or similar forms seem to be the convention for package managers:
https://askubuntu.com/questions/523962/how-to-install-a-package-with-apt-without-the-do-you-want-to-continue-y-n-p
https://dnf.readthedocs.io/en/latest/command_ref.html

If I'm wrong I'm not particularly attached to either solution. --assume-yes might be better than --yes tbh

datitisev:
I think the convention, at least based on what `apt` does, is that `-y` bypasses confirmation as confirm, right? Unless I'm remembering `apt` wrong.

luceos:
I wasn't talking about the convention for linux distros, but for Laravel apps.

datitisev:
Oh, that makes sense. But I'd think end users have more experience with running apt comma D's to install dependencies for the forum than Laravel applications.

luceos:
Some feedback after having reviewed the PR:

- Why two commands, one for single and one for all? Those could probably merged into one, you could even make the `extension` argument an array option?
- I see unused code in the ExtensionEnableAllCommand (`$enabled =`).
- The extension argument doesn't explain what format one needs to use, `flarum/tags` vs `tags` or `fof/upload` vs `fof-upload`? Ps it's the vendor path, so `fof/upload`.
- Bundled extensions are usually called first party extensions. Not sure whether we want to change that terminology here, what term are we using in our docs?

askvortsov1:
Agree with all except the first statement. One of the reasons for having this is that it's a royal pain to manage extensions when you have a lot of them, especially if you need to disable them to update flarum. 

If we want to condense it down to one command, perhaps we could accept an asterisk as an argument, which would just take the role of 'all'?

luceos:
> If we want to condense it down to one command, perhaps we could accept an asterisk as an argument, which would just take the role of 'all'?

That was exactly what I was thinking of, sorry for not stating that ;) Alternatively an `--all` flag is also possible.

askvortsov1:
> * Why two commands, one for single and one for all? Those could probably merged into one, you could even make the `extension` argument an array option?

Merged, I agree this is a lot better. And extensions has been changed to an array

> * I see unused code in the ExtensionEnableAllCommand (`$enabled =`).

Fixed!

> * The extension argument doesn't explain what format one needs to use, `flarum/tags` vs `tags` or `fof/upload` vs `fof-upload`? Ps it's the vendor path, so `fof/upload`.

Actually, it's the ID format, updated :grinning:  

> * Bundled extensions are usually called first party extensions. Not sure whether we want to change that terminology here, what term are we using in our docs?

As per https://flarum.org/docs/extend/#core-vs-extensions, we call them bundled

askvortsov1:
Agree that comments are extra, I mainly had them in to space out code to slightly increase readibility and forgot to remove.

As to merging the two, while they look similar, I think there's enough small distinctions that a combined version could be messier than what we have now

src/Extension/Console/ExtensionDisableCommand.php
```diff
@@ -0,0 +1,125 @@
...
+     */
+    protected function configure()
+    {
+        $this
+            ->setName('extensions:disable')
+            ->setDescription('Disable an extension')
+            ->addArgument(
+                'extensions',
+                InputArgument::IS_ARRAY | InputArgument::OPTIONAL,
+                'The extensions to disable. You must provide this in extension-id format, that is: VENDOR_NAME-EXTENSION_NAME, based on the composer package, where "flarum-" or "flarum-ext-" is stripped from the extension name. For example, "flarum\tags" has an id of "flarum-tags", and "some-programmer/flarum-extension" has an id of "some-programmer-extension"'
```
clarkwinkelmann, line 48: Do we need such an extensive description ? We should just document somewhere what a "Flarum extension ID" is (maybe we already do?) then assume the user is familiar with that definition.

src/Extension/Console/ExtensionDisableCommand.php
```diff
@@ -60,20 +72,54 @@ protected function configure()
...
+    }
+
+    protected function getExtensionsToDisable()
+    {
+        $extensionIds = [];
+        if ($this->input->getOption('all')) {
+            // All Extensions
+            $extensionIds = $this->allEnabledExtensions();
+        } elseif ($this->input->getOption('all-except-bundled')) {
+            // All exc
```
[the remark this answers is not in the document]
clarkwinkelmann: Agree about the comments. I think all comments could be removed in this PR.

src/Extension/Console/ExtensionEnableCommand.php
```diff
@@ -0,0 +1,124 @@
...
+        foreach ($this->input->getArgument('extensions') as $extension) {
+            if (! $this->extensions->getExtension($extension)) {
+                $this->error("Extension: $extension is not installed.");
+            } elseif (! $extensionIds->contains($extension)) {
+                $extensionIds[] = $extension;
+            }
+        }
+
+        $extensionIds = $extensionIds->filter(function ($extension) {
+            return ! $this->extensions->isEnabled($extension) ?: $this->info("Extension: $extension is already enabled, ignoring") && false;
```
clarkwinkelmann, line 119: Would it be more readable to do:

```
if ($this->extensions->isEnabled($extension)) {
    $this->info("Extension: $extension is already enabled, ignoring");
    return false;
} else {
    return true;
}
```

One-liners are great on the moment, but I think that `&& false` will be confusing to future code readers. Also it's a bit unclear on first read what the `!` negates.

Same suggestion for all other one-liner conditions in the two files.

askvortsov1: Absolutely agree, that's a lot better!

src/Extension/Console/ExtensionDisableCommand.php
askvortsov1: Yeah good point.

src/Extension/Console/ExtensionDisableCommand.php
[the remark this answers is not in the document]
askvortsov1: Done

src/Extension/Console/ExtensionDisableCommand.php
```diff
@@ -97,27 +97,41 @@ protected function allEnabledExtensions()
...
             } elseif (! in_array($extension, $extensionIds)) {
                 $extensionIds[] = $extension;
             }
         }
 
+        if ($this->input->getOption('all')) {
+            $extensionIds = $this->allEnabledExtensions();
+        } elseif ($this->input->getOption('all-except-bundled')) {
+            $extensionIds = array_merge($extensionIds, array_filter($this->allEnabledExtensions(), function ($extension) use ($extensionIds) {
+                if (substr($extension, 0, 6) === 'flarum' && ! in_array($extension, $extensionIds)) {
```
clarkwinkelmann, line 117: I would suggest adding an `$onlyCore` parameter to `$this->allEnabledExtensions()` and doing the check there against the actual vendor.

I think there's a good chance to find vendor names that start with "flarum" or even "flarum-" in the future. We already have https://packagist.org/packages/flarumite/

Not sure how that would play with the info message though.

src/Extension/Console/ExtensionEnableCommand.php
```diff
@@ -90,33 +94,43 @@ protected function allExtensions()
...
             }
         }
 
-        $extensionIds = $extensionIds->filter(function ($extension) {
-            return ! $this->extensions->isEnabled($extension) ?: $this->info("Extension: $extension is already enabled, ignoring") && false;
+        if ($this->input->getOption('all')) {
+            $extensionIds = $this->allExtensions();
+        } elseif ($this->input->getOption('include-bundled')) {
+            $extensionIds = array_merge($extensionIds, array_filter($this->allExtensions(), function ($extension) use ($extensionIds) {
+                if (substr($extension, 0, 6) !== 'flarum' && ! in_array($extension, $extensionIds)) {
```
clarkwinkelmann, line 116: Same thing here. Too much collision risks with this check IMO

src/Extension/Console/ExtensionDisableCommand.php
askvortsov1: Do we somewhere store a list of core extensions? I'm not sure how we could really validate this.

src/Extension/Console/ExtensionEnableCommand.php
askvortsov1: Same point as above, but I'd definitely add a "-" after "flarum" in the check

src/Extension/Console/ExtensionDisableCommand.php
clarkwinkelmann: I'd say a core extension is an extension whose vendor is `flarum`. And this can't be checked against the extension ID because we lose the separation between vendor and package. This should be checked against the extension object.

askvortsov1: I'm not sure how that fixes the issue though: if an extension's vendor is `flarum`, it's ID will start with `flarum-`, and vice versa.

clarkwinkelmann: The problem is that `flarum-something` is a valid vendor name that someone could register or even only use locally, and there's no way to tell it apart from the actual `flarum` vendor if we are not reading the `vendor` attribute of the extension.

Why we remove the separation between vendor and package in the extension ID would be a question for another issue... Keeping it would give a bit more flexibility for situations like these and completely avoid the possibility for two extensions having the same ID

askvortsov1: That's a good point, I hadn't thought of that.

luceos:
I've been battling to look at this PR for several reasons. Most of the thoughts that came to mind weren't that optimistic, so I'll try to keep this to the point. You said that you worked on a legacy code base which made you respect the Flarum code base even more. This is crucial in what I'm going to say.

I dislike the commands logic to `getExtensionsToEnable` and `getExtensionsToDisable`, both methods are unnecessarily convoluted. They don't look clean. Maybe we need to spice up the Extension Manager to make it responsible of understanding the difference between core and **community** extensions. If we could make that logic clean and simple it would mean the world in this PR.

askvortsov1:
I'm going to draft this for now. I think the question of revamping the extension manager is a big one worthy of it's own discussion (what's the benefit of shortening to an extension ID instead of just using the package name), and we still need to come to a decision on #2068

luceos:
I think the extension id was shortened in previous releases to be used inside the `app.initializers` array. However that convention has been dropped for a while and should work with the actual package name as well.

clarkwinkelmann:
Shortened ID is also used to expose all javascript exports to the global `flarum.extensions` object, and to store the list of enabled extensions.

I do agree it would make sense to rely on the Composer package name in as many places as possible, in particular for the list of enabled extensions, which then means we can rely on that for all extension management commands.

askvortsov1:
I think this is better done in https://discuss.flarum.org/d/25246-extiverse-helper, as that doesn't depend on the app being able to boot (which is a major limitation here)

clarkwinkelmann:
Could we maybe create an issue to discuss this?

There were many things discussed here and I think there are still some very valid discussions to be had on the subject. A PR definitely isn't the best place for that until we make a definite choice.

askvortsov1:
> Could we maybe create an issue to discuss this?
> 
> There were many things discussed here and I think there are still some very valid discussions to be had on the subject. A PR definitely isn't the best place for that until we make a definite choice.

I think the main applicable discussion here is changing extension IDs to be the full composer package name. Any others?

PR #2059, opened 2020-03-11
SSO Overhaul, Authentication Extenders

### LINKED ISSUE #223: Login via ResponseFactory (used for external auth) doesn't generate LoggedIn event

### LINKED ISSUE #236: OAuth2 Login/Registration Flow
## Bug Report

**Current Behavior**
The current behavior relies on email addresses to compare user authentication, this means that if a user uses an OAuth2 account that doesn't use the same email as a previous account that user ends up with two accounts.

**Steps to Reproduce**
1. Register an account the normal way (using one email)
2. Logout
3. Login using an OAuth2 provider that uses a different email address

**Expected Behavior**
We should do this exactly the same way all other systems do this as described:

##### User Has Account Already (E-Mail Matches)
1. User chooses to sign in using OAuth2 for the first time
2. Server detects that the email from OAuth2 matches that of a current user
3. Server request password for the found user account
4. User enters the password correctly
5. Server connects the accounts and never request a password going forward

##### User Has Account Already (E-Mail Doesn't Match)

1. User chooses to sign in using OAuth2 for the first time
2. Server can not find a pre-existing account
3. Server displays a prompt asking if user wants to sign in using existing account or register a new one
3. User chooses to sign into account
4. User successfully enters credentials
5. server makes the association and never prompts for forum credentials again when using OAuth2

##### User Does Not Have Account

1. User chooses to sign in using OAuth2
2. Server can not find pre-existing account
3. Server displays a prompt asking if user wants to sign in using existing account or register new one
4. User chooses to register new account
5. Server request required information not provided by OAuth2
6. User completes sign up
7. Server makes association between the newly registered account and the OAuth2 provider

**Screenshots**

**Environment**
- Flarum version: 0.1.0-beta.9
- Website URL: ANY flarum site (https://discuss.flarum.org)
- Webserver: [e.g. apache, nginx] ANY
- Hosting environment: [e.g. shared, vps] ANY
- PHP version: ANY
- Browser: [e.g. chrome 67, safari 11] ANY

**Possible Solution**
From my understanding from a [forum post I made](https://discuss.flarum.org/d/21014-oauth2-authentication-flow-signup-login-needs-work) the code to store additional data from OAuth2 providers has already been implemented. It should be possible to I believe that this is mostly a UI change to make this change function. I don't know exactly what needs to happen in order to make this function in the backend or for that matter if this can be done using a plugin.

### LINKED ISSUE #285: User setting to connect existing account with auth provider

### LINKED ISSUE #332: Using SSO authentication to login to an existing un/pw account
I created an administrative user authenticated with a local password; later, I logged it out, and selected to login using github.

I was immediately logged in to my administrator account.

This means if the upstream services do not verify email addresses sufficiently it is possible to impersonate/use the account of a local administrative user, etc.

This has been an area where facebook in particular has had a history of security bugs (e.g. myopera accounts not requiring email address verification).

If a user account has a password, it should be necessary to enter the password to allow access from facebook, IMO.

### LINKED ISSUE #380: Disable standard registration

### LINKED ISSUE #2043: Authentication Revamp
Right now, we have a pretty decent foundation for various authentication methods, but I think a bit of infrastructural revamp is warranted. I propose the following:

- Keep authentication within the Forum service provider, so that the various views can take full advantage of the frontend
- Providing an extender for authentication methods. This would support managing authentication drivers, as well as pre/post login, logout, and registration middleware
- Creating an interface for authentication drivers (they would define login, registration, logout, and possibly other things (e.g. resetting passwords).
- making the current username password system an instance of these drivers
- cleaning up code so that login and registration via SSO isn't dependent on a login modal being open while not disturbing the convenient system.
- Similarly to the above point, standardize an abstract class of "SSO Auth Providers".  as an implementation of the authentication driver instance. Right now, there's a lot of code duplication across all the SSO systems that we could easily manage centrally.
- adding a section to the user profile page where users can connect and disconnect login providers
- adding a section to the admin panel where forum admin can enable / disable authentication providers (and possibly configure other settings, like disabling sign-up and requiring email validation). Require that at least 1 be enabled at all time though.
- Add a console command to disable custom authentication drivers and reenable the username/password driver so that admins don't get locked out of their forum if SSO breaks. 

And as a very eventual wishlist feature, offer a way to test SSO providers from the admin portal before enabling them.

Considering that our goal is to provide a stable API for extensions and that a large portion of extensions have to do with various SSO providers, I think this is something (or at least components of it) that warrants being included in Stable. It would also provide an elegant solution to https://github.com/flarum/core/issues/641, https://github.com/flarum/core/issues/1858, https://github.com/flarum/core/issues/897, and others.

askvortsov1:
Fixes flarum/framework#2043 
Fixes Part of flarum/framework#1891 
Fixes flarum/issue-archive#380 
Fixes flarum/issue-archive#236 
Fixes flarum/issue-archive#332
Fixes flarum/issue-archive#223 
Fixes flarum/issue-archive#285 
- probably more

- [x] Deprecate disorganized components of old authentication system
- [x] Centralize system for extensions to register new authentication systems
- [x] Add extender to add an sso driver
- [x] Sso registration respects disabled signups.
- [x] Add new admin pane, allowing forum admins to:
   - [x] Enable/disable authentication providers
   - [x] Enable/disable signups (move from permissions pane)
   - [x] Enable/disable username/password registration
- [x] Add new user settings page, allowing users to link and unlink external auth providers from their account
   - [x] Don't allow users to unlink sso providers if username/password authentication is disabled, and the account in question only has 1 linked provider.
- [x] Add a console command to re-enable username/password login (if SSO breaks we don't want to strand forum admin).
- [x] Add 'trust emails' feature. If a user is not logged in, and SSO returns the email of an existing user, an error message will be shown if this SSO Driver doesn't have 'trust emails' enabled in the admin pane. If trust emails is enabled for this driver, the SSO provider will be linked to the user's account.
- [x] Automatically display login buttons based off of meta method of driver.
- [x] Rewrite bundled extensions to use this new system
   - [x] Auth Github
   - [x] Auth Facebook
   - [x] Auth Twitter
- [x] Add integration tests for SSO extender methods/functionality test (relies on flarum/framework#2052)
- [x] Extensive manual tests

TO ADD AFTER STABLE
- Ability to test SSO providers in admin panel before enabling (shouldn't be too difficult)
- Add (non-accidential) support for linking multiple identifiers from a provider to a flarum user.
- Manage button meta appearance from auth settings pane
- Manage provider settings through auth settings pane?
- Add extender method, full support for multifactor authentication.

- How should we standardize nomenclature? IE sso vs auth vs external auth? driver vs provider? Right now, it's kind of a mess of all these.
- Does the sso flow described in SsoController accurately describe what we want? If not, what should we change/improve?
- Any things I missed?

**Screencasts**
Disable username/password auth:
https://youtu.be/kN4XT8BwUIQ

Disabling all sso providers force enables user/pass auth
https://youtu.be/OvS1jxHj5QQ

Disabling signup:
https://youtu.be/TnwrDFgCqnM

Auth Panel controls buttons, Registration works
https://youtu.be/EQchPk7oQuc

Linking/unlinking accounts works
https://youtu.be/teHYWFbTzQY

Can't unlink last provider if username/password auth disabled
https://youtu.be/YcMkBNWORgY

SSO registration respects disabled signups:
https://youtu.be/UCYJWyJB07I

**Confirmed**


tankerkiller125:
> How should we standardize nomenclature? IE sso vs auth vs external auth? driver vs provider? Right now, it's kind of a mess of all these.

The nomenclature is pretty simple for me at least. 

SSO is Representative of the whole sign in experience through 3rd party companies/services

Auth to me at least describes any way of logging in to the system (AKA are you authorized to access this resource)

External auth, I think is something we can get rid of, either that or we need to replace SSO with it depending on if we want to be technically accurate or we want it to be user friendly (SSO being technically accurate)

When it comes to driver vs provider the way I see it is that the driver is the implementation between flarum and the 3rd party company/service. Provider to me at least means the 3rd party company/service itself.

> Does the sso flow described in SsoController accurately describe what we want? If not, what should we change/improve?

The flow looks correct to me, but I would have to test it myself to get a full understanding of how it works.

> Should the "disable username/password auth" slider be disabled if no SSO providers are enabled? I'm also leaning in this direction.

Yes, I'd rather not let people break their site, even if admins can bypass it (which it appears they can?) I just think it would create a support headache if we don't disable the slider if there are no SSO provider enabled, further I think that the slider should automatically slide to be turned off if the user disables all of the 3rd party providers.

Also before you go and update all the bundled 3rd party providers to in theory support this system I would recommend you wait until a decision is made on flarum/framework#2006

src/Forum/Controller/SsoController.php
```diff
@@ -98,6 +98,12 @@ public function handle(Request $request): ResponseInterface
                 }
             }
 
+            if (!$actor->isGuest() && in_array($driverId, $actor->linkedProviders())) {
+                // This user has already linked an account from this provider. We don't currently support multiple
```
tankerkiller125: Is this because we're still linking providers by email address? Personally I would prefer to see that the PR also includes a switch to identify users via a relationship between the Flarum user ID and the provider user ID with email strictly as a backup during the transition period.

askvortsov1: Not exactly. The actual "linkage" is done via an identifier that the SsoDriver would return back (as part of the SsoResponse wrapper class). That identifier COULD be an email, but it could also be any other persistent id.

Email Address is still used to identify if an incoming ssoResponse might be linked to an existing user (which we should keep imo). However, this PR also adds support to linking accounts that don't necessarily share an email (see line 124). 

That means that theoretically, a user could link an account that doesn't share an email with their flarum account through their user portal, log out, and log in (same provider, different account on provider, where that account shares an email with their flarum account). Without this check, that new account would also be linked, and now they'd have multiple accounts on one provider linking to one flarum account.

We could allow this, but under the UI in my PR, there isn't a mechanism to unlink a specific identifier from a provider, nor to add another identifier from the same provider. 

This might be a feature, not a bug: the system should still function properly with this, and I can't think of security issues it could cause. If it's something we want to support, I can remove this check.

tankerkiller125: I think for simplicity for now we can wait on multi-identity single provider sso until after stable. I just wanted to verify how this is actually implemented so that we had the option to do so easily in the future.

askvortsov1:
>SSO is Representative of the whole sign in experience through 3rd party companies/services
>
>Auth to me at least describes any way of logging in to the system (AKA are you authorized to access this resource)
>
>External auth, I think is something we can get rid of, either that or we need to replace SSO with it depending on if we want to be technically accurate or we want it to be user friendly (SSO being technically accurate)
>
>When it comes to driver vs provider the way I see it is that the driver is the implementation between flarum and the 3rd party company/service. Provider to me at least means the 3rd party company/service itself.

I agree with this interpretation, I'll standardize this. I'm also going to rename LoginProvider (and the associated table) to SsoProvider, as that more accurately describes what it does (username password login doesn't use LoginProviders afaik.

>Yes, I'd rather not let people break their site, even if admins can bypass it (which it appears they can?) I just think it would create a support headache if we don't disable the slider if there are no SSO provider enabled, further I think that the slider should automatically slide to be turned off if the user disables all of the 3rd party providers.

Done.

> Also before you go and update all the bundled 3rd party providers to in theory support this system I would recommend you wait until a decision is made on flarum/framework#2006

One of my main reasons to do this is to give the core dev team an easier way to test the new functionality when reviewing this PR. Besides, even if the extensions are getting moved to FoF, might as well make their lives easier too :D

src/Install/Steps/WriteSettings.php
```diff
@@ -62,6 +62,7 @@ private function getDefaults()
             'custom_less' => '',
             'default_locale' => 'en',
             'default_route' => '/all',
+            'enable_user_pass_auth' => '1',
```
clarkwinkelmann: I would prefer the word "password" to be written in full in the setting name. "user" could be dropped.

Something like "enable_password_auth" or even "enable_password_authentication".

I feel like shortening the name can only lead to confusion and doesn't provide any real performance gain.

src/Api/Serializer/ForumSerializer.php
```diff
@@ -60,6 +60,14 @@ public function getId($model)
      */
     protected function getDefaultAttributes($model)
     {
+        $ssoProviders = array_map(function ($provider) {
+            return $this->app->make($provider)->meta();
+        }, $this->app->make('flarum.auth.supported_drivers'));
+
+        $ssoProviders = array_filter($ssoProviders, function ($provider) {
+            return $this->settings->get('auth_driver_enabled_'.$provider);
```
clarkwinkelmann: It would be better to check whether the provider is enabled **before** instantiating it through the container.

Wouldn't this logic be better placed elsewhere ? Surely it's not the only place that must obtain the list of enabled drivers ? I might have missed another place.

src/Api/Controller/DeleteSsoProviderController.php
```diff
@@ -0,0 +1,67 @@
...
+
+        if ($actor->isGuest()) {
+            throw new NotAuthenticatedException;
+        }
+
+        if (! in_array($provider, $actor->ssoProviderNames())) {
+            throw new RouteNotFoundException;
+        }
+
+        if (!$this->settings->get('enable_user_pass_auth', true) && count($actor->ssoProviderNames()) == 1) {
```
clarkwinkelmann: Please use `===` wherever possible

less/common/Checkbox.less
```diff
@@ -54,6 +54,12 @@
     box-shadow: 0 2px 4px @shadow-color;
   }
 }
+.Checkbox--switch.disabled {
+    color: @muted-more-color !important;
```
clarkwinkelmann: Can't we do without `!important` ?

js/src/forum/components/SsoGrid.js
```diff
@@ -0,0 +1,146 @@
...
+      </table>
+    );
+  }
+
+  providerLinked(provider) {
+    return app.session.user.ssoProviders().includes(provider);
+  }
+
+  disableUnlinking(provider) {
+    return this.providerLinked(provider) && !app.forum.attribute('enableUserPassAuth') && app.session.user.ssoProviders().length == 1;
```
clarkwinkelmann: I'd prefer `===`.

js/src/forum/components/SsoGrid.js
```diff
@@ -0,0 +1,146 @@
...
+          control.props.state = false;
+          control.loading = false;
+          app.session.user.data.attributes.ssoProviders = app.session.user.ssoProviders().filter(item => item != provider);
+          for (const input in this.inputs) {
+            this.inputs[input].props.disabled = this.disableUnlinking(input);
+          }
+          m.redraw();
+        })
+        .catch(() => {
+          control.loading = false;
```
clarkwinkelmann: Any error is completely silenced here I believe.

js/src/forum/components/LogInButtons.js
```diff
@@ -20,6 +21,28 @@ export default class LogInButtons extends Component {
...
+    const providers = app.forum.data.attributes.ssoProviders;
+
+    for (const provider in providers) {
+      const className = `Button LogInButton--${provider}`;
+      const path = `/auth/${provider}`;
+      var style = "";
+      style += providers[provider].buttonColor ? `background-color: ${providers[provider].buttonColor};` : '';
+      style += providers[provider].buttonTextColor ? `color: ${providers[provider].buttonTextColor}` : '';
+      items.add(provider,
+        <LogInButton
```
clarkwinkelmann: I have not really seen how it looks like for extensions after this change, but I feel they loose some control over their login buttons. They are restricted to only the attributes hard-coded here but might want to customize others.

js/src/forum/components/LogInButtons.js
```diff
@@ -20,6 +21,28 @@ export default class LogInButtons extends Component {
...
   items() {
-    return new ItemList();
+    const items = new ItemList();
+
+    const providers = app.forum.data.attributes.ssoProviders;
+
+    for (const provider in providers) {
+      const className = `Button LogInButton--${provider}`;
+      const path = `/auth/${provider}`;
+      var style = "";
```
clarkwinkelmann: Please use an object for styles. (`style = {}`, `style.beckgroundColor = ''`, ...). This makes it easier for other exensions to override parts of the styles. Also it might be more performant for Mithril because it doesn't need to override the whole style, it can only update the properties that changed.

And replace that `var` with `let` :stuck_out_tongue: (or `const` if you replace with an object).

js/src/admin/components/AuthPage.js
```diff
@@ -0,0 +1,207 @@
...
+export default class AuthPage extends Page {
+  init() {
+    super.init();
+
+    this.saving = false;
+    this.refresh();
+  }
+
+  refresh() {
+    this.loading = true;
```
clarkwinkelmann: an `m.redraw()` call is probably needed here, otherwise it doesn't refresh until `this.loading` is set back to `false` at the end of the function.

Does the loading indicator make any sense though ? It's a synchronous operation and I don't see what could be holding the process enough to necessitate a loading indicator.

js/src/admin/components/AuthPage.js
```diff
@@ -0,0 +1,207 @@
...
+    this.fields.forEach(key => this.values[key] = m.prop(settings[key]));
+
+    for (const driver in this.drivers) {
+      this.driverFields.forEach(field => {
+        const fieldName = this.driverFieldKey(field.name, driver);
+
+        this.fields.push(fieldName);
+        this.values[fieldName] = m.prop(settings[fieldName]);
+
+        this.driverInputs[fieldName] = new Checkbox({
```
clarkwinkelmann: Component instances should be avoided as much as possible. Try using `Checkbox.component()` in the view where it's needed. This syntax will likely no longer work once we update to the newer Mithril.

clarkwinkelmann (changes requested):
I read through the code and here are some comments I have.

Also I noticed some files don't have the ending newline. Not sure if our automated tools will fix that formatting or not.

src/Forum/Console/EnableUserPassAuthCommand.php
```diff
@@ -0,0 +1,53 @@
...
+ * For detailed copyright and license information, please view the
+ * LICENSE file that was distributed with this source code.
+ */
+
+namespace Flarum\Forum\Console;
+
+use Flarum\Console\AbstractCommand;
+use Flarum\Settings\SettingsRepositoryInterface;
+
+class EnableUserPassAuthCommand extends AbstractCommand
```
clarkwinkelmann: Oh and I forgot one thing.

I don't think it makes sense to create this command. We might as well have a single command to edit any setting, and explain in the documentation how to re-enable password login if necessary.

src/Install/Steps/WriteSettings.php
askvortsov1: I've changed it to enable_password_auth throughout, I agree that's a lot clearer. Thanks!

src/Forum/Console/EnableUserPassAuthCommand.php
askvortsov1: Removed, I agree that'd be a better approach (although probably best for a separate PR). Were you envisioning that users input json encoded strings directly?

js/src/forum/components/SsoGrid.js
askvortsov1: Done!

src/Api/Controller/DeleteSsoProviderController.php
askvortsov1: Done!

js/src/forum/components/LogInButtons.js
askvortsov1: That's true, I wanted to make it easier for extensions, and these are the settings that every auth extension uses so far. If extensions wanted to go beyond this, couldn't they just extend the LogIn Buttons itemlist?

js/src/forum/components/LogInButtons.js
askvortsov1: Done, didn't know that was an option, thanks!

js/src/forum/components/SsoGrid.js
askvortsov1: I can confirm that this isn't the case: the red alert popup will still show up.

less/common/Checkbox.less
askvortsov1: Done, thanks!

js/src/admin/components/AuthPage.js
askvortsov1: True, fixed by removing the loading indicator.

src/Api/Serializer/ForumSerializer.php
askvortsov1: I changed it to filter before making. This is, actually, the only place where that's used, so IMO it's alright to keep it here until we'd need it elsewhere.

js/src/admin/components/AuthPage.js
askvortsov1: Fixed, although when I tried to do this with the Alert, it didn't work so I didn't include that. Is that something we've encountered before?

clarkwinkelmann: `Alert` is different. An instance has to be passed to `app.alerts.show` so while it also inherits from Component, it's not used in the same way as components injected in a `view` method.

Eldenroot:
@askvortsov1 you are on fire :D I just wish that Flarum team would be faster in merging and releasing new versions...

Littlegolden:
Hope to be able to disable password and email registration.
But the login window still retains these forms.

askvortsov1:
@Littlegolden could you clarify what the use case for this would be? This seems like a niche feature that would moreso belong in an extension than as a part of core

Littlegolden:
I noticed that after turning off the auth button in [this demo video](https://youtu.be/kN4XT8BwUIQ), the email + password login method was closed for both the login and registration windows.

I want to be able to turn off the `email + password` login method only during registration. Generally speaking, many spamers have a lot of spam emails. They use these email registration forums to post advertisements or other spam content, but third-party logins are largely restricted The behavior of the spoilers, especially those social media (such as WeChat and Telegram) that require a mobile phone number to register, so only allowing third-party logins can largely avoid spam encounters on the forum.

askvortsov1:
@flarum/core should the disable standard auth switch be split into one to disable login and one to disable registration?

askvortsov1:
Also: one thing I've noticed on my own site is that the "Sign Up" modal title can be confusing if the user is coming back from an SSO login and just needs to enter a username/email/etc. What if, if there's a registration token, we set the title of the modal to something like "Finish Registration" or "Complete Registration" or "One Last Step..."

js/src/admin/components/AuthPage.js
```diff
@@ -0,0 +1,215 @@
...
+  }
+
+  toggle(key) {
+    this.values[key](!this.values[key]());
+
+    this.driverInputs[key].props.state = this.values[key]();
+
+    if (this.allSsoDisabled()) {
+      this.values['enable_password_auth'](true);
+      m.redraw();
```
clarkwinkelmann, line 146: This `m.redraw()` is unnecessary as there's another just below outside of the `if`

js/src/admin/components/AuthPage.js
```diff
@@ -0,0 +1,215 @@
...
+          state: this.values[fieldName](),
+          onchange: () => this.toggle(fieldName)
+        });
+      });
+    }
+
+    if (this.allSsoDisabled() && !this.values['enable_password_auth']()) {
+      this.values['enable_password_auth'](true);
+      this.saveSettings();
+      m.redraw();
```
clarkwinkelmann, line 49: This `m.redraw()` is unnecessary as there's another just below outside of the `if`

js/src/admin/components/AuthPage.js
```diff
@@ -0,0 +1,215 @@
...
+    };
+
+    for (const driver in this.drivers) {
+      this.driverFields.forEach(field => {
+        const fieldName = this.driverFieldKey(field.name, driver);
+
+        this.fields.push(fieldName);
+        this.values[fieldName] = m.prop(settings[fieldName]);
+
+        this.driverInputs[fieldName] = Checkbox.component({
```
clarkwinkelmann, line 39: This is still not the intended way to use components. A component should be rendered as part of `view()` and not stored https://mithril.js.org/components.html#avoid-creating-component-instances-outside-views

You can keep this loop to initiate `this.fields` and `this.values`, but `this.driverInputs` should not be used and the checkbox should be created directly in the `view()` method.

js/src/forum/components/LogInModal.js
```diff
@@ -61,11 +61,12 @@ export default class LogInModal extends Modal {
 
   body() {
     return [
-      <LogInButtons/>,
+      <LogInButtons />,
 
-      <div className="Form Form--centered">
-        {this.fields().toArray()}
-      </div>
+      app.forum.attribute('enablePasswordAuth') ?
```
clarkwinkelmann, line 66: Wouldn't it be better to add the condition inside of `fields()` ? And return an empty fields array when password login is disabled.

That way extensions that want to insert custom fields into the login modal by `extend()`ing the `fields` ItemList can still do it even if password auth is disabled.

js/src/forum/components/SsoGrid.js
```diff
@@ -0,0 +1,147 @@
...
+     * that represent them.
+     *
+     * @type {Object}
+     */
+    this.inputs = {};
+
+    // For each of the Sso type-method combinations, create and store a
+    // new checkbox component instance, which we will render in the view.
+    this.providers.forEach(provider => {
+      this.inputs[provider.name] = Checkbox.component({
```
clarkwinkelmann, line 32: Same comment as with the Checkbox in `AuthPage.js`. It should be created inside of `view()` https://mithril.js.org/components.html#avoid-creating-component-instances-outside-views

js/src/forum/components/SsoGrid.js
```diff
@@ -0,0 +1,147 @@
...
+      m.redraw();
+
+      app.request({
+        method: 'DELETE',
+        url: app.forum.attribute('apiUrl') + '/auth/' + provider
+      })
+        .then(() => {
+          control.props.state = false;
+          control.loading = false;
+          app.session.user.data.attributes.ssoProviders = app.session.user.ssoProviders().filter(item => item != provider);
```
clarkwinkelmann, line 95: Strict comparison ? `!==`

js/src/forum/components/SsoGrid.js
```diff
@@ -0,0 +1,147 @@
...
+      const $window = $(window);
+
+      window.open(app.forum.attribute('baseUrl') + '/auth/' + provider, 'logInPopup',
+        `width=${width},` +
+        `height=${height},` +
+        `top=${$window.height() / 2 - height / 2},` +
+        `left=${$window.width() / 2 - width / 2},` +
+        'status=no,scrollbars=yes,resizable=no');
+      control.loading = false;
+      m.redraw();
```
clarkwinkelmann, line 117: The two `m.redraw()` in `if` and `else` could be moved outside of the condition block and replaced by a single call at the end of the `toggle()` method.

Maybe the redraw can be completely removed because Mithril is supposed to already redraw after an `onchange` event like the one that's used to call `toggle`.

src/Forum/Auth/SsoResponse.php
```diff
@@ -0,0 +1,175 @@
...
+        return $this;
+    }
+
+    /**
+     * @param string $username
+     * @return $this
+     */
+    public function suggestUsername(string $username): self
+    {
+        $username = preg_replace('/[^a-z0-9-_]/i', '', $username);
```
clarkwinkelmann, line 151: How are we going to make this extendable ?

Can/should we re-use the regex from the UserValidator ?

clarkwinkelmann (changes requested):

js/src/admin/components/AuthPage.js
askvortsov1: Fixed

js/src/admin/components/AuthPage.js
askvortsov1: Fixed

js/src/admin/components/AuthPage.js
askvortsov1: Done!

js/src/forum/components/LogInModal.js
askvortsov1: Done, made the same fix for signup modal

js/src/forum/components/SsoGrid.js
askvortsov1: Done!

js/src/forum/components/SsoGrid.js
askvortsov1: Done!

js/src/forum/components/SsoGrid.js
askvortsov1: The ones in the if statement need to be there since it's an asynchronous operation. For the redraw in else(), we could remove it, but then if the user closes the window it'll just keep on loading. I'm actually not 100% sure whether it should spin if a new provider is being linked.

src/Forum/Auth/SsoResponse.php
askvortsov1: I would be fine running the username itself through the user validator, but the question is, what to do if it doesn't validate? Right now I'm thinking we just return an empty string, but should there be some notification/error message? Personally don't think so since it's suggested, but open to other ideas.

askvortsov1:
Thanks for the review Clark! I've left a few follow-up comments. Also, any thoughts on the 2 things I addressed above (changing modal title when token present, and splitting password auth switch to allow login but not signup)

src/Forum/Auth/SsoResponse.php
clarkwinkelmann: Yeah that was just an idea, we could extract the regex from the validation rules, then use that regex like you did to filter out characters.

But then if an extension changes the validator rules and no longer uses a regex, this will break :thinking:

js/src/forum/components/SsoGrid.js
clarkwinkelmann: Oops I misspoke. The other redraw I saw was line 78, which is before the condition. If you move the redraw from line 78 to the end of the `toggle()` method, it will take care of the one you removed as well.

franzliedke:

tankerkiller125 (approved):

dsevillamartin (changes requested):
This PR seems to have commits that shouldn't be here. Not sure what happened - perhaps you rebased your branch and the existing commits have different SHAs so they are new?

267 files cannot be changed by this PR 🙂

tankerkiller125 (changes requested):
My original review was a case of not paying close enough attention to which PR I was on

BoneTM:
why this stuff stopped

PR #2067, opened 2020-03-14
Make profile urls work with usernames only and not user IDs

### LINKED ISSUE #1356: Usernames cannot be numeric

clarkwinkelmann:
**Fixes #1356**

Drop support for ID in profile urls. Only accept username.

This is based on one of the proposal discussed in the issue. However I'm not sure how to handle errors, as "not found" is now an empty list instead of a native 404.

Also the fact that `apiDocument` is now a list could break some extensions that read the payload directly. I'm thinking of the SEO extension which I have not tested in combination with this.

**Confirmed**

**Required changes:**

### DIFF
--- a/js/src/forum/components/UserPage.js
+++ b/js/src/forum/components/UserPage.js
@@ -7,6 +7,7 @@ import SelectDropdown from '../../common/components/SelectDropdown';
 import LinkButton from '../../common/components/LinkButton';
 import Separator from '../../common/components/Separator';
 import listItems from '../../common/helpers/listItems';
+import Alert from '../../common/components/Alert';
 
 /**
  * The `UserPage` component shows a user's profile. It can be extended to show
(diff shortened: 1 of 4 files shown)


askvortsov1:
This should be rebased on master to fix the database issue in CI/CD tests.

askvortsov1 (changes requested):
I think the biggest issue here is the usage of the list endpoint to fetch a single object. Could we create a second route, connecting to the same endpoint, which would have the same path but accept a string username instead of an integer id?

Alternatively, could we set up the endpoint so that we can pass a 'byUsername' argument (perhaps via a GET parameter)?

clarkwinkelmann:
@askvortsov1 that was the discussion taking place in #1356 . I don't find any of the solutions perfectly adequate, but this one was the one I preferred. I'm still not sure if it's the best one to go with.

I'll take care of the rebase.

clarkwinkelmann:
Rebased!

I still don't like this approach very much, but it's the one I like the best from all the available options :grimacing:

askvortsov1:
I don't see this breaking anything for extensions, or setting up anything that could be broken for extensions, so I'm fine with having this solution (as you said, it's the best of what's been proposed so far) until if/when we come up with something better.

askvortsov1 (approved):

js/src/forum/components/UserPage.js
```diff
@@ -94,14 +95,33 @@ export default class UserPage extends Page {
...
-      if ((user.username().toLowerCase() === lowercaseUsername || user.id() === username) && user.joinTime()) {
+      if ((user.username().toLowerCase() === lowercaseUsername) && user.joinTime()) {
         this.show(user);
         return true;
       }
     });
 
     if (!this.user) {
-      app.store.find('users', username).then(this.show.bind(this));
+      app.request({
```
datitisev, line 105: This `app.request()` can be replaced with `app.store.find('users', { filter: ... })`

franzliedke:

js/src/forum/components/UserPage.js
clarkwinkelmann: Changed!

clarkwinkelmann:
I'm closing this for now. I still don't like the approach enough.

PR #2068, opened 2020-03-14
Add AskQuestionTrait for console commands

askvortsov1:
**Fixes #0000**

Various console commands are likely to need to ask for input within the command. So:
- The ->ask() and ->secret() utilities have been added in a new trait, so that they are accessible to console commands.
- A ->confirm() utility has been added to require a user to confirm something (y/N)

**Confirmed**

### DIFF
--- a/src/Console/AskQuestionTrait.php
+++ b/src/Console/AskQuestionTrait.php
@@ -46,7 +46,7 @@ trait AskQuestionTrait
     protected function confirm($question)
     {
         while (true) {
-            $confirmation = $this->ask($question . ' (y/N)');
+            $confirmation = $this->ask($question.' (y/N)');
 
             if (strtolower($confirmation) === 'y') {
                 return true;
--- a/src/Console/AskQuestionTrait.php
+++ b/src/Console/AskQuestionTrait.php
@@ -47,7 +47,7 @@ trait AskQuestionTrait
     protected function confirm($question)
     {
         while (true) {
-            $confirmation = $this->ask($question.' (y/N)');
+            $confirmation = $this->ask($question.' (y/n)');
 
             if (strtolower($confirmation) === 'y') {
                 return true;
--- a/src/Console/AskQuestionTrait.php
+++ b/src/Console/AskQuestionTrait.php
@@ -42,4 +42,19 @@ trait AskQuestionTrait
 
         return $this->questionHelper()->ask($this->input, $this->output, $question);
     }
+
+    protected function confirm($question)
+    {
+        while (true) {
+            $confirmation = $this->ask($question . ' (y/N)');
+
+            if (strtolower($confirmation) === 'y') {
+                return true;
+            } elseif (strtolower($confirmation) === 'n') {
+                return false;
+            } else {
+                $this->error('Invalid Response');
+            }
+        }
+    }
 }
--- a/src/Console/AskQuestionTrait.php
+++ b/src/Console/AskQuestionTrait.php
@@ -24,6 +24,7 @@ trait AskQuestionTrait
         if (is_null($this->questionHelper)) {
             $this->questionHelper = new QuestionHelper;
         }
+
         return $this->questionHelper;
     }
 
--- a/src/Console/AskQuestionTrait.php
+++ b/src/Console/AskQuestionTrait.php
@@ -1,0 +1,45 @@
+<?php
+
+/*
+ * This file is part of Flarum.
+ *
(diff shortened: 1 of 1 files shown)


askvortsov1:
Ah didn't notice Symfony had the confirmation question option. As for the rest of the methods in this trait, they actually use Symfony's system. The trait is intended as less of a functionality adding system, and moreso to reduce code reuse. The methods themselves were taken from the installation command's CLI data provider.

askvortsov1:
After looking at Symfony's ConfirmationQuestion, it seems that Symfony doesn't repeat until a vaid input is entered (which I thought would be a better behavior here), and as such, I'd prefer to keep my implementation, unless there's strong preference to do it the way Symfony does. 

Just to clarify my previous comment: I anticipate that with the console extender, there will be more interest and ability for extensions to add console commands, so adding this utility to provide the question helper and some basic question utilities should make people's lives easier, and reduce code reuse.

src/Console/AskQuestionTrait.php
```diff
@@ -0,0 +1,61 @@
...
+
+        $question->setHidden(true)->setHiddenFallback(true);
+
+        return $this->questionHelper()->ask($this->input, $this->output, $question);
+    }
+
+    protected function confirm($question)
+    {
+        while (true) {
+            $confirmation = $this->ask($question.' (y/N)');
```
franzliedke, line 50: This seems to suggest that `n` is the default. But the subsequent code would just report "Invalid Response", if I hit Enter immediately, right?

franzliedke:
When you say "reduce code reuse", do you mean code *duplication*? Reuse generally seems like a valuable goal to me. :wink:

src/Console/AskQuestionTrait.php
askvortsov1: I don't think it does, my goal here is that unless the user enters y or n, it would keep asking. I know that I've hit the wrong key in these before by accident, resulting in something that wasn't the result I wanted, which is why I'm proposing to do it this way.

askvortsov1:
> When you say "reduce code reuse", do you mean code _duplication_?

Indeed!

src/Console/AskQuestionTrait.php
franzliedke: That's a fine approach, but why is the `N` upper-case and the `y` lower-case?

askvortsov1: Not sure tbh, thought that was convention. Ill standardize.

src/Console/AskQuestionTrait.php
```diff
@@ -0,0 +1,61 @@
...
+
+namespace Flarum\Console;
+
+use Symfony\Component\Console\Helper\QuestionHelper;
+use Symfony\Component\Console\Question\Question;
+
+trait AskQuestionTrait
+{
+    /**
+     * @var QuestionHelper
```
clarkwinkelmann, line 18: Is this kind of docblock useful at all?

src/Console/AskQuestionTrait.php
```diff
@@ -0,0 +1,61 @@
...
+    protected function secret($question)
+    {
+        $question = new Question("<question>$question</question> ");
+
+        $question->setHidden(true)->setHiddenFallback(true);
+
+        return $this->questionHelper()->ask($this->input, $this->output, $question);
+    }
+
+    protected function confirm($question)
```
clarkwinkelmann, line 47: I notice there's no option to give a default here? It seems Symfony's own `ConfirmationQuestion` has a way to give a default.

Then the default option is usually put in uppercase in the text (example `(Y/n)`)

Not sure if we want that?

askvortsov1:
> Seeing Laravel is based on Symfony's commands, wouldn't it make sense to use Laravel's own console package here? Their own base command is super comprehensive and already contains all those helpers https://github.com/laravel/framework/blob/7.x/src/Illuminate/Console/Concerns/InteractsWithIO.php

I.... uh did not know this exists... I'm all for using it, but we don't currently use the illuminate/console package. Is that something we'd be comfortable adding?

askvortsov1:
Let's keep this open for now.

tankerkiller125:
Like clark, I'm also in favor of using the illuminate/console package for commands. Will be better for extension devs and is already comprehensive for the future.

askvortsov1:
Let's do that then. It's needed anyways if we want to bring scheduled commands into core.

PR #2074, opened 2020-03-18
WIP Access tokens and session management

clarkwinkelmann:
Work in progress

**Screenshot**

Absolutely not final at all:

image

**Confirmed**

**Required changes:**


tankerkiller125:
Based on that screenshot (I know it's not final) but I would recommend saying Expires (instead of Lifetime) and then show a proper date or human readable time till (eg. in 30 minutes)

askvortsov1:
1. This is awesome. Between this and the sso thing there's gonna be a LOT more user control over auth and login. 

2. Agree with expires

3. For the token, one valid use case is if users can use it for Flarum's REST APIs. I think that there should be a setting whether or not to share tokens with users (perhaps in the authentication page?), and if so, the token shouldnt be immediately displayed until the user hits a display button to reduce the risk of leakage due to screen recording / onlookers

clarkwinkelmann:
So my idea is as follow:

I refactored the session to be token-based. That way revoking a token means revoking a session, and will work with any session driver.

Logging out revokes the token that was used for the session.

I added a nullable `title` to tokens. The idea is that named tokens are user-created tokens, while unnamed tokens are those created by Flarum ("session tokens").

I will add a button that allows the user to create a token. A name will be required.

I will hide the token value from the UI. For Flarum-created tokens, I will show them in an "active sessions" layout similar to apps like for example Telegram:

image

I don't plan to offer any way to visually access the token value for session tokens. For user-created tokens, I want to make a UI similar to all developer dashboards, with an option to reveal the token, for example like Mailgun:

image

The expiration time will most likely not be shown. In particular if we implement the suggestion from #2075 where tokens would become either "expirable" or "non-expirable". In this case user-created tokens would be non-expirable, and sessions could show a label showing if it's either a "remember me" session or a session that will expire after no activity.

A button will allow terminating all other sessions at once. That feature will get linked to the password change as well, so that changing passwords clears all active sessions.

tankerkiller125:
This sounds like a perfect UI to me and it's pretty standard so I like it.

askvortsov1:
Agreed with Matt, I like these proposed changes a lot!

js/src/common/Application.js
```diff
@@ -81,6 +82,7 @@ export default class Application {
    * @public
    */
   store = new Store({
+    'access-tokens': AccessToken,
```
askvortsov1: Isn't the convention for JS variables and attributes to use camelCase?

js/src/common/models/AccessToken.js
```diff
@@ -0,0 +1,19 @@
+import Model from '../Model';
+
+export default class AccessToken extends Model {
+  apiEndpoint() {
```
askvortsov1: Is there a reason why we don't want to just call the model type 'tokens'?

js/src/forum/components/SettingsPage.js
```diff
@@ -60,6 +61,14 @@ export default class SettingsPage extends UserPage {
       })
     );
 
+    items.add('tokens',
```
askvortsov1: I feel like this section should go alongside email/password/sso settings, not 100% sure though.

js/src/forum/components/SettingsPage.js
```diff
@@ -142,4 +151,19 @@ export default class SettingsPage extends UserPage {
 
     return items;
   }
+
+  /**
+   * Build an item list for the user's tokens controls.
+   *
+   * @return {ItemList}
+   */
+  tokensItems() {
```
askvortsov1: why not tokenItems?

js/src/forum/components/TokensManager.js
```diff
@@ -0,0 +1,62 @@
...
+    }).then(payload => {
+      this.tokens = app.store.pushPayload(payload);
+      m.redraw();
+    });
+
+    // TODO: pagination
+  }
+
+  view() {
+    if (this.tokens === null) {
```
askvortsov1: Why not use a this.loading attribute?

js/src/forum/components/TokensManager.js
```diff
@@ -0,0 +1,62 @@
...
+      return LoadingIndicator.component();
+    }
+
+    console.log(this.tokens);
```
askvortsov1: Leftover statement from debugging? :D

askvortsov1 (changes requested):
Should there be some kind of optional link between access tokens and login providers? That way, on linking/unlinking a login provider, all access tokens relating to that login provider could be automatically removed.

js/src/common/Application.js
datitisev: I think this is `access-tokens` because of the `type` field in the serializer.

askvortsov1:
Are you planning on adding in the user-created token UI/mechanisms in this PR, or as an add-on after this refactor is completed?

js/src/forum/components/TokensManager.js
clarkwinkelmann: Yep this will be removed :innocent:

js/src/forum/components/TokensManager.js
clarkwinkelmann: That's how I have usually been doing it for resource loading. null = not fetched, [] = no results, [...] = results.

I usually only use an additional loading variable when an action is being performed after clicking a button.

js/src/forum/components/SettingsPage.js
clarkwinkelmann: The other entry with a name was also plural (`notificationsItems`) so that's why I used plural as well. But I agree this sounds a bit odd.

js/src/forum/components/SettingsPage.js
clarkwinkelmann: Still WIP. we can change that later

js/src/common/models/AccessToken.js
clarkwinkelmann: Yes. Because there's also `email_tokens`, `password_tokens` and `registration_tokens`. I feel like calling it `token` is just a recipe for confusion.

js/src/common/Application.js
clarkwinkelmann: Yes that's the type, and the prefix is because there are multiple kinds of tokens (see the other comment). None of the other tokens have a serializer (yet) though.

It will be the first two-word serializer type used in Flarum core, so we have some room on how to name them. In my extensions I have usually used the dash-syntax for such types, like `catchthefish-fishes`, but of course we could also use camelCase.

clarkwinkelmann:
> Should there be some kind of optional link between access tokens and login providers?

I feel like this would become overly complex. But this could be added, even by an extension.

> Are you planning on adding in the user-created token UI/mechanisms in this PR, or as an add-on after this refactor is completed?

I was thinking of adding it here, because I will have to design the tokens UI anyway so it's best I integrate it together.

js/src/common/Application.js
datitisev: I think camelCase would be best here - I have used hyphens in the past as well, and usually am confused as to how I should be naming them.

clarkwinkelmann: The JSON:API spec does not make any recommendation on this https://jsonapi.org/format/#document-resource-object-identification

**However**, the type is usually matched to the url endpoint. I don't really like the idea of having a camelCase endpoint url...

datitisev: I don't think it should match the URL endpoint. Everything returned by the API right now is camelCase (model attributes), so adding hyphens really doesn't make sense.

js/src/forum/components/TokensManager.js
askvortsov1: Hmm it might make sense to standardize this across other views, super low priority tho

js/src/common/models/AccessToken.js
askvortsov1: Agreed. In that case, should we change the API controller path?

js/src/common/Application.js
askvortsov1: Agree with David's point. But if we're going for word separators in an endpoint, I believe that snake case is the convention?

franzliedke: We do already have a two-word serializer, type `mail-settings`. Let's just stick to that, everything else seems like bikeshedding to me. :wink:

js/src/forum/components/SettingsPage.js
```diff
@@ -60,6 +61,14 @@ export default class SettingsPage extends UserPage {
       })
     );
 
+    items.add('tokens',
+      FieldSet.component({
```
KyrneDev, line 65: Let's use JSX here

js/src/forum/components/TokensManager.js
```diff
@@ -0,0 +1,62 @@
...
+
+    return m('table', [
```
KyrneDev, line 32: Why not JSX?

clarkwinkelmann: It's all WIP right now, I'll fix the code style later.

I'm used to not using JSX for reasons I have already expressed elsewhere, but I'll adapt to match the Flarum code style.

askvortsov1:
Closing for inactivity

PR #2080, opened 2020-03-21
Fix Buggy Modals

### LINKED ISSUE #1504: Start a Discussion Tag Popup closes in IE 11

### LINKED ISSUE #1813: Modal is not showing
## Bug Report

**Current Behavior**

Whenever you open the edit tags modal, it hides itself. This is caused by the missing class `in`. This makes editing tags a pain.

**Steps to Reproduce**
1. Go to discuss.flarum.org, open a discussion
2. Click on edit tags
3. Modal hides

The behavior is sporadic, but just now it happened to me constantly on discuss using firefox.

the-turk:
**Fixes #1504**
**Fixes #1813**

First commit was proposing to change margin values for `@tablet-up` breakpoint as seen on [Bootstrap stylesheets](https://github.com/twbs/bootstrap/blob/68b0d231a13201eb14acd3dc84e51543d16e5f7e/dist/css/bootstrap.css#L6053). Evethough it fixes the issue, i wasn't happy with it and did some further research. Eventually, i noticed that we're calling `onready()` method of `ModalManager` component twice [Line 33 & 57]. First call waits for transitions to be complete and the second one seems useless. Made some conversation with @datitisev on Discord and now we're here.

* Did i miss something?

I know this one is hard to catch since it seems sporadic but try to test it with developer tools docked to bottom of the browser (make it bigger until you succeed).

Note that zooming out from the page before clicking on the "Choose Tags" button also fixes the issue ([as stated by @sometao](https://discuss.flarum.org/d/22788-opening-choose-tags-from-the-new-post-form-is-buggy/11)) ~~so it should be something related with viewport heights~~.

I look forward to your reviews and I need to fix this since it effects some of my modal related extensions.

**Screenshot**

https://vimeo.com/399523797

**Confirmed**

### DIFF
--- a/js/src/common/components/ModalManager.js
+++ b/js/src/common/components/ModalManager.js
@@ -54,7 +54,6 @@ export default class ModalManager extends Component {
     m.redraw(true);
 
     this.$().modal({backdrop: this.component.isDismissible() ? true : 'static'}).modal('show');
-    this.onready();
   }
 
   /**
--- a/less/common/Modal.less
+++ b/less/common/Modal.less
@@ -177,7 +177,7 @@
 
 @media @tablet-up {
   .Modal {
-    margin: 30px auto;
+    margin: 120px auto;
   }
   .Modal-close {
     position: absolute;
--- a/less/common/Modal.less
+++ b/less/common/Modal.less
@@ -177,7 +177,7 @@
 
 @media @tablet-up {
   .Modal {
-    margin: 120px auto;
+    margin: 30px auto;
   }
   .Modal-close {
     position: absolute;


the-turk:
This doesn't fix extension settings modal's issue.

> People who encounters with this issue may see that the same thing happens if they enable an extension which has settings modal (it'll try to show the extension's settings modal after enabling it but it'll fail eventually).

Here are some quotes from @datitisev:

> both timeout of `setTimeout(() => this.extensionSettings[enabled](), 0);` or `requestAnimationFrame(() => this.extensionSettings[enabled]())` work

> only thing is that it blinks with full "brightness" before opening the modal, but not much we can do about that

the-turk:
Some modal issues is still a mystery, i had the same missing class problem while developing the Diff extension and i end up with adding it manually.

Maybe we should accept this PR as a solution for the tag selection modal issue (or we can just call it "a fix for double `onready()` call") and check other modals (like extension settings) after the frontend rewrite.

luceos:
Thanks for this fix. I just merged and deployed it, but seemingly it doesn't fix the issue with the edit tags modal 😞

the-turk:
> 
> 
> Thanks for this fix. I just merged and deployed it, but seemingly it doesn't fix the issue with the edit tags modal 😞

@luceos I can confirm that this PR fixed both modals

luceos:
@the-turk my apologies, it seems I had been too fast with updating discuss. Compilation of the dist files hadn't been completed yet, I can confirm the PR works, thank you!

PR #2082, opened 2020-03-24
(#1236) user preferences refactoring

### LINKED ISSUE #1236: Database naming changes

luceos:
**Continued from #1236**

This is the second attempt to refactor the user (notification) preferences.

- I've decided to split up the retained and deprecated logic into traits for easier review and future maintainability. 
- Added a test to confirm functionality is retained.
- Added two extenders, one for registering a NotificationChannel and another for UserPreferences, offering the public scopes required by extensions.

There's still some work required with the tests, I am unsure how to handle them in our current set up 😞 

**Confirmed**

**Required changes:**


src/Extend/NotificationChannel.php
```diff
@@ -0,0 +1,41 @@
...
+    private $channel;
+
+    private $enabled = [];
+
+    public function __construct(string $channel)
+    {
+        $this->channel = $channel;
+    }
+
+    public function extend(Container $container, Extension $extension = null)
```
askvortsov1: Super tiny thing, but I think it might be better to have the extend method after the enabled method. That way, code relating to the API isn't split up by the implementation.

src/Extend/UserPreferences.php
```diff
@@ -0,0 +1,29 @@
...
+use Illuminate\Contracts\Container\Container;
+
+class UserPreferences implements ExtenderInterface
+{
+    public function extend(Container $container, Extension $extension = null)
+    {
+        // There's nothing here as the logic is contained in the `add()` method directly.
+    }
+
+    public function add(string $key, callable $transformer = null, $default = null)
```
askvortsov1: See comment above about rearranging method order.

src/User/Concerns/DeprecatedUserNotificationPreferences.php
```diff
@@ -0,0 +1,36 @@
...
+
+trait DeprecatedUserNotificationPreferences
+{
+    /**
+     * Encode an array of preferences for storage in the database.
+     *
+     * @param mixed $value
+     * @deprecated 0.1.0-beta.13: `users.preferences` is no longer used.
+     */
+    public function setPreferencesAttribute($value)
```
askvortsov1: If they're being deprecated, shouldn't the methods in this class still contain an implementation so they don't break dependent code until we remove them in beta 14?

src/User/Concerns/DeprecatedUserNotificationPreferences.php
```diff
@@ -0,0 +1,36 @@
...
+    /**
+     * Get the key for a preference which flags whether or not the user will
+     * receive a notification for $type via $method.
+     *
+     * @param string $type
+     * @param string $method
+     * @return string
+     * @deprecated 0.1.0-beta.13: `users.preferences` is no longer used, use \Flarum\User\NotificationPreference.
+     */
+    public static function getNotificationPreferenceKey($type, $method)
```
askvortsov1: See comment above about including logic in deprecated code until we remove it.

src/User/Concerns/UserPreferences.php
```diff
@@ -0,0 +1,87 @@
+<?php
+
+/*
+ * This file is part of Flarum.
+ *
+ * For detailed copyright and license information, please view the
+ * LICENSE file that was distributed with this source code.
+ */
+
+namespace Flarum\User\Concerns;
```
askvortsov1: Tiny thing, but not sure how I feel about this namespace naming.

src/User/Concerns/UserPreferences.php
```diff
@@ -0,0 +1,87 @@
...
+     * @param string $value
+     * @return array
+     */
+    public function getPreferencesAttribute($value)
+    {
+        $defaults = array_map(function ($value) {
+            return $value['default'];
+        }, static::$preferences);
+
+        $user = Arr::only($this->notificationPreferences->toArray(), array_keys(static::$preferences));
```
askvortsov1: maybe this should be called $userPreferences instead of $user?

src/User/Concerns/UserPreferences.php
```diff
@@ -0,0 +1,87 @@
...
+    /**
+     * Get the value of a preference for this user.
+     *
+     * @param string $key
+     * @param mixed $default
+     * @return mixed
+     */
+    public function getPreference($key, $default = null)
+    {
+        return $this->$key ?? $default;
```
askvortsov1: Should this take into account defaults defined in $preferences?

src/User/Concerns/UserPreferences.php
```diff
@@ -0,0 +1,87 @@
...
+     * @param mixed $value
+     */
+    public function setPreference($key, $value)
+    {
+        $preference = static::$preferences[$key];
+
+        // If a user preference is registered, transform the value.
+        if ($preference) {
+            $value = $value === null ? $preference['default'] : $value;
+            $value = $preference['transformer']($value);
```
askvortsov1: Should there be a check for a null transformer here? Since we allow that in the extender

src/User/Concerns/UserPreferences.php
```diff
@@ -0,0 +1,87 @@
...
+    /**
+     * An array of registered user preferences. Each preference is defined with
+     * a key, and its value is an array containing the following keys:.
+     *
+     * - transformer: a callback that confines the value of the preference
+     * - default: a default value if the preference isn't set
+     *
+     * @var array
+     */
+    protected static $preferences = [];
```
askvortsov1: Is there a reason why you don't include disclose_online and follow_after_reply here? If they're the default values...

migrations/2019_09_28_222801_migrate_users_table_preferences_columns.php
```diff
@@ -0,0 +1,52 @@
...
+        $db = $builder->getConnection();
+
+        $db->table('users')
+            ->select(['id', 'preferences'])
+            ->whereNotNull('preferences')
+            ->orderBy('id')
+            ->each(function ($user) use ($db) {
+                collect(json_decode($user->preferences ?? '{}'))
+                    ->each(function ($value, $key) use ($user, $db) {
+                        if (in_array($key, ['discloseOnline', 'followAfterReply'])) {
```
askvortsov1: Should this also include preferences defined in extensions?

src/Extend/UserPreferences.php
```diff
@@ -0,0 +1,29 @@
...
+ * LICENSE file that was distributed with this source code.
+ */
+
+namespace Flarum\Extend;
+
+use Flarum\Extension\Extension;
+use Flarum\User\User;
+use Illuminate\Contracts\Container\Container;
+
+class UserPreferences implements ExtenderInterface
```
askvortsov1: Should we have a method for changing the default value/transformer of a preference? Or perhaps that can come after we figure out extension boot order.

src/User/NotificationPreference.php
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * @property int $user_id
+ * @property string $type
+ * @property string $channel
+ * @property bool $enabled
+ */
+class NotificationPreference extends AbstractModel
+{
+    protected static $channels = [];
```
askvortsov1: Should this include 'alert' and 'email'?

src/User/NotificationPreference.php
```diff
@@ -0,0 +1,75 @@
...
+
+    public function user()
+    {
+        return $this->belongsTo(User::class);
+    }
+
+    public static function addChannel(string $channel, array $defaults = [])
```
askvortsov1: IMO 'defaults' is a bit ambiguous, maybe something more like 'defaultEnabledTypes'?

tests/integration/extenders/NotificationChannelTest.php
```diff
@@ -0,0 +1,47 @@
...
+
+    private function add_channel()
+    {
+        $this->extend(new NotificationChannel('test'));
+    }
+
+    /**
+     * @test
+     */
+    public function can_add_notification_channel()
```
askvortsov1: This doesn't actually work: setUp instantiates the app before extenders are registered, meaning that the extenders don't actually need to do anything. To check that the extender registers without error, you'll need to separate out the database operations into their own method and call that explicitly after adding the extender.

tests/integration/extenders/NotificationChannelTest.php
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+namespace Flarum\Tests\integration\extenders;
+
+use Flarum\Extend\NotificationChannel;
+use Flarum\Tests\integration\RetrievesAuthorizedUsers;
+use Flarum\Tests\integration\TestCase;
+use Flarum\User\User;
+
+class NotificationChannelTest extends TestCase
```
askvortsov1: Following in the convention for extender tests so far, I'd like to see at least tests for:

User::getPreferences does not contain a channel if it has not been added via an extender
A given type is disabled/null (what should it be by the way if it hasn't been registered). Also should there be a way to register types? Anyway, a given type is disabled/null if it hasn't been enabled/disabled.
User::getPreferences DOES contain a channel if it has been added via an extender
A given type is enabled for a user that hasn't overriden that if it has been enabled by default via an extender
A user can override a default type preference by setting their own preference.

tests/integration/extenders/UserPreferencesTest.php
```diff
@@ -0,0 +1,67 @@
...
+use Flarum\Tests\integration\RetrievesAuthorizedUsers;
+use Flarum\Tests\integration\TestCase;
+use Flarum\User\User;
+use Illuminate\Support\Arr;
+
+class UserPreferencesTest extends TestCase
+{
+    use RetrievesAuthorizedUsers;
+
+    public function setUp()
```
askvortsov1: See comment above about setUp

src/User/NotificationPreference.php
```diff
@@ -0,0 +1,75 @@
...
+            ->get();
+
+        if ($channel && $type) {
+            return $saved->first();
+        }
+
+        return $saved;
+    }
+
+    public static function setNotificationPreference(User $user, string $channel, string $type, bool $enabled = true)
```
askvortsov1: Why is enabled assumed to be true?

src/Extend/UserPreferences.php
askvortsov1: We should document that this does NOT add in database columns

migrations/2019_09_28_222800_create_users_table_preferences_columns.php
```diff
@@ -0,0 +1,27 @@
...
+ * For detailed copyright and license information, please view the
+ * LICENSE file that was distributed with this source code.
+ */
+
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Schema\Builder;
+
+return [
+    'up' => function (Builder $schema) {
+        $schema->table('users', function (Blueprint $table) {
```
askvortsov1: Why don't we have a dedicated table for user preferences like with notification preferences? I'm not sure how I feel about extensions modifying the user table this much, it might lead to unexpected conflicts.

tests/integration/extenders/UserPreferencesTest.php
```diff
@@ -0,0 +1,67 @@
...
+
+namespace Flarum\Tests\integration\extenders;
+
+use Flarum\Extend\UserPreferences;
+use Flarum\Tests\integration\RetrievesAuthorizedUsers;
+use Flarum\Tests\integration\TestCase;
+use Flarum\User\User;
+use Illuminate\Support\Arr;
+
+class UserPreferencesTest extends TestCase
```
askvortsov1: This also needs a default test case, as well as a few more case tests, I'll try to think of good tests to include

askvortsov1:
How does the ability to set default enabled types for a channel play with registration of types in ConfigureNotificationTypes (which also has enabledByDefault) (which I think is good variable naming and could also be used here)?

askvortsov1:
This should probably also be rebased on master to fix the automated testing.

askvortsov1:
I think we should coordinate a bit on these extenders. I think that we can probably rename the UserPreferences extender to just User (to include more functionality), although I don't feel so strongly.

@luceos I'm a bit more interested in the notification extender. Right now, we have plans to make a Notification Type extender (to register blueprints for notification types). I'm not exactly sure how this should play with your NotificationChannel extender (I'm also not seeing what exactly the notification channel would do, there doesn't seem to be an option to register a driver that would execute when the notification is fired).

I would propose that we have a notification extender that looks as follows:

`class Notification implements ExtenderInterface
{
    public function type($blueprint, $serializer, array $channelsEnabledByDefault = []);
    public function channel($identifier, $driver, array $typesEnabledByDefault = []);
}`

If a particular notification option is enabled by either a channel or a type, it would be enabled.

How would you feel about this layout?

askvortsov1:
One major concern I have about this is requiring that, in order for new preferences to be added, a new column must be added to the users table. I'm sure there was some reasoning behind that, but I am unsure why this is better than the alternative (creating a new table for user preferences, and json-encoding the values).  From a scalability perspective, I feel like a dedicated table (like we're doing with notification preferences) would scale better, and have less damaging column naming conflicts. This approach would also break all dependent extensions and require them to make new migrations. I know that we could always change it again later, but I'm still unsure of the benefit of this approach.

luceos:
I've also made a case about the reasoning behind this change. But we decided back then to move forward and I still think we should enact these changes as soon as possible. If we come across to many problems in the future, we can always modify this logic for 0.2.

luceos:
I think it makes more sense to refactor user preferences as a whole beyond 0.1. I'll create an issue for that and assign it to the 0.2 milestone for now.

PR #2088, opened 2020-03-27
Fix calculation of post position on page

### LINKED ISSUE #2086: Link to post scrolls to the previous post
**Current Behavior**
Opening a Link to direct post, say 5th post, scrolls to the 4th post.

**Steps to Reproduce**
1. Go to https://discuss.flarum.org/d/23184-flarum-categories/5
2. Address bar will show /4 not /5.  4th post would also be scrolled .

**Possible Solution**
Creating a pull request :) . Should it be in beta branch ? docs say latest stable branch. Apologies, this is my first attempt at collaborating with git.

gyaaniguy:
**Fixes #2086**

Changes Formula for calculating the current post, which is used to scroll to the post. Current implementation scrolled to the previous post, and not the linked post.

Open this link: https://discuss.flarum.org/d/9033-telegram-login-and-notifications-by-flagrow/16 . Note the address bar . The address bar should not change to /15

**Confirmed**

btw: I think I screwed up my previous pull requests. Hope I am doing it right this time. Would love to get some guidance if I am doing it wrong.

### DIFF
--- a/js/src/forum/components/PostStream.js
+++ b/js/src/forum/components/PostStream.js
@@ -476,7 +476,7 @@ class PostStream extends Component {
       const top = $item.offset().top;
       const height = $item.outerHeight(true);
 
-      if (top + height > scrollTop) {
+      if (top + height > scrollTop + 2) {
         if (!startNumber) {
           startNumber = endNumber = $item.data('number');
         }
--- a/js/src/forum/components/PostStream.js
+++ b/js/src/forum/components/PostStream.js
@@ -475,7 +475,10 @@ class PostStream extends Component {
       const $item = $(this);
       const top = $item.offset().top;
       const height = $item.outerHeight(true);
-
+      
+      // jQuery offset() and outHeight() may return value in decimals.
+      // Instead of 5, may return 5.8. TO offset this, we add +2 to the
+      // value of scrollTop.
       if (top + height > scrollTop + 2) {
         if (!startNumber) {
           startNumber = endNumber = $item.data('number');


js/src/forum/components/PostStream.js
```diff
@@ -476,7 +476,7 @@ class PostStream extends Component {
       const top = $item.offset().top;
       const height = $item.outerHeight(true);
 
-      if (top + height > scrollTop) {
+      if (top + height > scrollTop + 2) {
```
askvortsov1: Could you explain/document where the number 2 is coming from?

askvortsov1:
I can confirm that this fixes the issue, have left a comment about clarifying a value.

js/src/forum/components/PostStream.js
gyaaniguy: Sure. jQuery offset() and outerheight() can return value in fractions.
Instead of 5 , it may return 5.9. Since this happens twice, increasing value of scrolltop by 2, nullifies this.

askvortsov1: That makes sense, thanks! I think it would be good to have that explanation as a comment above this block of code so that it's instantly understandable to developers in the future.

gyaaniguy: Most welcome! I have added a commit with a comment in code. Please let me know if its okay.

askvortsov1: Awesome!

the-turk:
I'm not convinced to be honest. How's int 2 fixes offset caused by a random fractional value? Why is 2? What's the math behind this?

Also, what will happen to this line if you decide to merge this PR:

https://github.com/flarum/core/blob/6b3d6349174d0cd70b84cd8a4cced3017122aae7/js/src/forum/components/PostStream.js#L484

It also calculating things using `scrollTop`

gyaaniguy:
The fractions increase the value of the variables 'top' , 'height' . What should be 3 , is currently 3.3 . This results in top+height being higher than what it should be , by a max value of .99 + .99 . Both can have fractions of .99 .
Another, potentially cleaner way, would be to round the fractions:
```
const top = floor($item.offset().top);
const height = floor($item.outerHeight(true));
```
Thoughts ? 
If I do make this change, should I simply make this as another commit ?

tankerkiller125:
If that change still functions properly then I would say that overall it's cleaner. And yes you would just make a new commit to the new branch and this PR will automatically get updated.

gyaaniguy:
hmm. floor() didn't have desired result. It seems +2 was just a hacky way. 
I researched deeper. The crux of the problem is that, when a post link is shown, couple of pixels of the previous post are visible in the viewport. I am not sure exactly why some pixels sneak below the top sticky bar...

the-turk:
This approach clearly will not work all the time. If top + height ends up with .20 + .10, you will still be off by adding 2.  Furthermore, .99 + .99 makes 1.98 and you're still off by 0.02. I think this is not a proper way for solving the issue.

gyaaniguy:
@the-turk I added 2px because 1.98 < 2 . That said, I also think this fractions idea may well be incorrect.

I digged in more: Opening a normal internal link, leaves 0px above the link. See here: https://en.wikipedia.org/wiki/Wikipedia:About#About_Wikipedia
On flarum, we can see some gap over an internal link. See the few pixels above the border and below the sticky header: https://discuss.flarum.org/d/9033-telegram-login-and-notifications-by-flagrow/16 .

The Maths: 
scrollTop: distance from top to target element.
top: distance from top to current element
height: height of current element
When height + top > scrolltop, then the current element is the target element.
This is fine. But in flarum, the extra pixels above cause the scrollTop value to reduce. So the height+top of previous element becomes larger than scrolltop... Thoughts ?

w-4:
In my flarum install I've had this for the calculatePosition function for a longer time (I think probably because of the same issue):

```js
 calculatePosition() {
    const marginTop = this.getMarginTop();
    const $window = $(window);
    const scrollTop = $window.scrollTop();
    const viewportTop = scrollTop + marginTop;
    const viewportHeight = $window.height() - marginTop;

    let startNumber;
    let endNumber;

    this.$('.PostStream-item').each(function() {
      const $item = $(this);
      const top = $item.offset().top;
      const height = $item.outerHeight(true);

      // If this item is above the top of the viewport, skip to the next
      // one. If it's below the bottom of the viewport, break out of the
      // loop.
      if (top + height < viewportTop) {
        return true;
      }
      if (top > viewportTop + viewportHeight) {
        return false;
      }

      let number = $item.data('number')

      if (!startNumber) {
        startNumber = endNumber = number;
      }

      if (number) endNumber = number;
    });

    if (startNumber) {
      this.trigger('positionChanged', startNumber || 1, endNumber);
    }
  }
```

here's the version of the pull request for reference:
```js
 calculatePosition() {
    const marginTop = this.getMarginTop();
    const $window = $(window);
    const viewportHeight = $window.height() - marginTop;
    const scrollTop = $window.scrollTop() + marginTop;
    let startNumber;
    let endNumber;

    this.$('.PostStream-item').each(function() {
      const $item = $(this);
      const top = $item.offset().top;
      const height = $item.outerHeight(true);
      
      // jQuery offset() and outHeight() may return value in decimals.
      // Instead of 5, may return 5.8. TO offset this, we add +2 to the
      // value of scrollTop.
      if (top + height > scrollTop + 2) {
        if (!startNumber) {
          startNumber = endNumber = $item.data('number');
        }

        if (top + height < scrollTop + viewportHeight) {
          if ($item.data('number')) {
            endNumber = $item.data('number');
          }
        } else return false;
      }
    });

    if (startNumber) {
      this.trigger('positionChanged', startNumber || 1, endNumber);
    }
  }
```

askvortsov1:
Alternatively to adding 2, could we just use floor()?

datitisev:
@askvortsov1 Wouldn't the problem be solved with .ceil()? If we are adding more whole numbers because of the number perhaps being a bit short of the next one, I assume that's what would fix the issue. 🤷

gyaaniguy:
To be honest, I am unsure of the fraction thing being the issue here at all. I'll explore this a bit more.

askvortsov1:
@gyaaniguy any updates on this?

gyaaniguy:
Even though I have a pretty good idea, I am unable to find exactly what's wrong here. Have tried a lot, gonna have to call it a quits.

askvortsov1:
Thank you for trying! I'm going to close this for now, we can always reopen it later if this issue emerges again.

Issue #2054, opened 2020-03-06
Transform deleting user alert into modal with extendable details


datitisev:
Currently, when deleting a user, we use a JavaScript `alert()` and simply mention that the posts won't be deleted (https://github.com/flarum/mentions/pull/41).

We will want to transform this alert to a JS Modal that is easily extensible to allow extensions to add their own info about what of their data will be deleted and/or kept when deleting a user.

E.g. Moderation notes (fof/moderator-notes) are deleted when the user they are for is deleted, but the notes that a deleted user created are kept. The extension would add this info to the Modal that appears when a user is deleted.

PR #2089, opened 2020-03-27
Add confirmation dialog when hiding and deleting comments (#1966)

### LINKED ISSUE #1966: Ask for confirmation when deleting a post
## Feature Request

I've found that quite a few users on my forum (soft)delete posts by mistake and sometimes don't even realize they do that.

I think this could be improved by asking for confirmation when deleting/hiding a post. I don't think it would be annoying, since deleting a post is not something you do very often.

It could also be an option, actually, depending on how people feel this issue.

timas130:
**Closes flarum/framework#1966**

- Added `ConfirmModal` for creating simple confirmation modals.

My shitcoding, please fix it.

**Screenshot**
flarumFeaturev2

**Confirmed**

- [x] No backend changes.

**Required changes:**

- [x] Related documentation PR: (Add docs for `ConfirmModal` component?)
- [x] Related core extension PRs: (flarum/lang-english#162)

### DIFF
--- a/js/src/forum/components/ConfirmModal.js
+++ b/js/src/forum/components/ConfirmModal.js
@@ -3,29 +3,20 @@ import Button from "../../common/components/Button";
 
 /**
  * A basic confirm modal.
+ * The special props are:
+ * * `translation` Translation path for some of the elements. It'll be used like this:
+ *   `${translation}.title`
+ * * `save` Callback called when pressing positive button.
  */
 export default class ConfirmModal extends Modal {
-  /**
-   * Initialize the modal.
-   * @param {String} title Modal title
-   * @param {String} message Message, shown above buttons
-   * @param {String} positiveButton Text to show on positive button
-   * @param {String} negativeButton Text to show on negative button
-   * @param {Function} callback Callback called when 
-   */
-  init(title, message, positiveButton, negativeButton, callback) {
-    super.init();
+  constructor(props, children) {
+    super(props, children);
 
-    this.titleProp = title;
-    this.message = message;
-    this.positiveButton = positiveButton;
-    this.negativeButton = negativeButton;
-    this.positiveCallback = callback;
     this.loading = false;
   }
 
   title() {
-    return this.titleProp;
+    return app.translator.trans(`${this.props.translation}.title`);
   }
 
   className() {
@@ -36,11 +27,11 @@ export default class ConfirmModal extends Modal {
     return (
       <div className="Modal-body">
         <div className="Form Form--centered">
-          <p>{this.message}</p>
+          <p>{app.translator.trans(`${this.props.translation}.message`)}</p>
           <div className="Form-group">
             {Button.component({
               className: 'Button Button--primary Button--block',
-              children: this.positiveButton,
+              children: app.translator.trans(`${this.props.translation}.positive`),
               type: 'submit',
               loading: this.loading
             })}
@@ -48,7 +39,7 @@ export default class ConfirmModal extends Modal {
           <div className="Form-group">
             {Button.component({
               className: 'Button Button--block',
-              children: this.negativeButton,
+              children: app.translator.trans(`${this.props.translation}.negative`),
               onclick: this.hide.bind(this),  
               disabled: this.loading // Disable negative button when loading
             })}
@@ -60,7 +51,7 @@ export default class ConfirmModal extends Modal {
 
   onsubmit(e) {
     this.loading = true;
-    this.positiveCallback().then((() => {
+    this.props.save().then((() => {
       this.loading = false;
       this.hide();
     }).bind(this));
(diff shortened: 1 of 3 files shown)


js/src/forum/components/ConfirmModal.js
```diff
@@ -0,0 +1,70 @@
...
+    this.loading = false;
+  }
+
+  title() {
+    return this.titleProp;
+  }
+
+  className() {
+    return 'ConfirmModal Modal--small';
+  }
```
timas130: A question: should I add the `ConfirmModal` class? It doesn't have any additional less code.

luceos: Yes it should have that class. Even though we don't run changes it on, doesn't mean webmasters won't 👍

js/src/forum/utils/PostControls.js
```diff
@@ -128,14 +129,22 @@ export default {
...
+      app.translator.trans('core.forum.post_controls.hide_post_modal.message'),
+      app.translator.trans('core.forum.post_controls.hide_post_modal.positive'),
+      app.translator.trans('core.forum.post_controls.hide_post_modal.negative'),
+      () => {
+        this.pushAttributes({ hiddenAt: new Date(), hiddenUser: app.session.user });
+  
+        return this.save({ isHidden: true }).then(() => m.redraw());
+      }
+    );
+    app.modal.show(confirmModal);
```
datitisev: This is not how we initialize modals in Flarum (at least in core). Apart from that, there's no reason to be passing translator messages.

```suggestion
    app.modal.show(
	    new ConfirmModal({
	    	action: 'hide',
	    	save: () => {
		    	this.pushAttributes({ hiddenAt: new Date(), hiddenUser: app.session.user });
		  
		        return this.save({ isHidden: true }).then(() => m.redraw());
	        }
	    })
    );
```

The translated messages should be translated directly in the children, not passed as properties.
To get this `save` callback, you can do `this.props.save()` (or another name if you rename it), and the action with `this.props.action` (for example).

Same in the `deleteAction` below.

This can then be used as 
```jsx
Button.component({
	className: 'Button Button--block',
	children: app.translator.trans(`core.forum.post_controls.${action}_post_modal.negative`),
    onclick: this.hide.bind(this),  
    disabled: this.loading // Disable negative button when loading
})
```

dsevillamartin (changes requested):

js/src/forum/utils/PostControls.js
timas130: > Apart from that, there's no reason to be passing translator messages.

Ok, but what if it will be used outside `post_controls`? Is it ok to pass translation path like that:
```js
new ConfirmModal({
  action: 'hide',
  translation: 'core.forum.post_controls.hide_post_modal',
  // ...
});
```

datitisev: Oh, I didn't realize the Modal was supposed to be kind of a global utility, I see. Uh, not sure what the other devs' thoughts are on this, but in that case I guess you would have to pass multiple translations. _Or_ you could have a function that returns the translation key like
```js
(action) => `core.forum.post.controls.{action}_post_modal`
```
but this is probably overcomplicating it.

timas130: I didn't understand your idea with that function, because it'll be the same as the first message there. Anyway, I'll commit it like I said before, because I also realized it would be also great to have it in discussion hiding/deleting. It will look like this for now:
```js
hideAction() {
  app.modal.show(new ConfirmModal({
    translation: "core.forum.discussion_controls.hide_discussion_modal",
    save: () => {
      this.pushAttributes({ hiddenAt: new Date(), hiddenUser: app.session.user });
  
      return this.save({ isHidden: true });
    }
  }));
}
```

datitisev:
Hm, not sure if that's how we want the translations to be exactly. I'll wait for someone else from our team to comment instead of giving more feedback on that.

franzliedke:
Hi @timas130, thanks so much for this PR!!!

We just had a dev meeting where we talked about this and other PRs. Since there are still a few outstanding discussions (and we currently have a large queue of PRs to tackle, and want to avoid getting sidetracked in our focus on the first stable release), we want to make the following suggestion:

1. Please update this PR to only add new confirmation dialogs in the places where we don't have them yet, based on the simple, native-`confirm`-based approach that is already used in `DiscussionControls.prototype.deleteAction`. That would fix flarum/framework#1966, which we agree is quite important!
2. Please extract the introduction of the new, pretty confirmation modal into a new PR, where its implementation, design and general approach can be discussed. We already have an issue (#2054) for this topic, which is currently planned for a later release.

To be clear: we're suggesting this to get more of these nice changes in rather soon, while trying to avoid too much discussion about features that we hadn't planned to tackle yet. That doesn't mean we cannot merge a good PR for the other issue if we find the time, but at least we can ensure that we can all enjoy the urgent fixes (#1966).

franzliedke:

franzliedke:
@timas130 We integrated the changes for post deletion with a simple native JS alert in flarum/framework#2135, to fix the bug.

Feel free to submit your changes, rebased on master, in a new PR to fix flarum/issue-archive#221.

PR #2098, opened 2020-03-29
Refactor formatter extender

askvortsov1:
**Fixes part of #1891**

- Deprecated all events involved with Formatter
- Refactor ->configure() method on extender not to use events
- Add extender methods for ->render() and ->parse()
- Add integration tests for all of this

How do we feel about the method naming for ->render() and ->parse()? I feel it could be better but all my other ideas were clunky.

**Confirmed**

### DIFF
--- a/src/Formatter/Formatter.php
+++ b/src/Formatter/Formatter.php
@@ -79,7 +79,7 @@ class Formatter
     {
         $parser = $this->getParser($context);
 
-        // Deprecated in beta 13, remove in beta 14
+        // Deprecated in beta 15, remove in beta 16
         $this->events->dispatch(new Parsing($parser, $context, $text));
 
         foreach ($this->parsingCallbacks as $callback) {
@@ -101,7 +101,7 @@ class Formatter
     {
         $renderer = $this->getRenderer();
 
-        // Deprecated in beta 13, remove in beta 14
+        // Deprecated in beta 15, remove in beta 16
         $this->events->dispatch(new Rendering($renderer, $context, $xml, $request));
 
         foreach ($this->renderingCallbacks as $callback) {
@@ -153,7 +153,7 @@ class Formatter
         $configurator->Autolink;
         $configurator->tags->onDuplicate('replace');
 
-        // Deprecated in beta 13, remove in beta 14
+        // Deprecated in beta 15, remove in beta 16
         $this->events->dispatch(new Configuring($configurator));
 
         foreach ($this->configurationCallbacks as $callback) {
--- a/src/Formatter/Formatter.php
+++ b/src/Formatter/Formatter.php
@@ -26,21 +26,6 @@ class Formatter
 
     protected $renderingCallbacks = [];
 
-    public function addConfigurationCallback($callback)
-    {
-        $this->configurationCallbacks[] = $callback;
-    }
-
-    public function addParsingCallback($callback)
-    {
-        $this->parsingCallbacks[] = $callback;
-    }
-
-    public function addRenderingCallback($callback)
-    {
-        $this->renderingCallbacks[] = $callback;
-    }
-
     /**
      * @var Repository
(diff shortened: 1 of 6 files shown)


tankerkiller125:
Parse and render are pretty standard naming conventions when it comes to text formatting systems so I'm personally happy with that name.

clarkwinkelmann:
I'm not sure if discussions happened elsewhere but reading this PR I have trouble finding benefits to the callbacks solution. They seem to just clutter the code compared to the previous solution.

To quote @franzliedke in #1891 

> I think the more focused public API and the greater freedom to make internal changes without having to change the API for extensions is absolutely worth it. Once we have the public API, we can get rid of the events as implementation detail in many cases, which should pretty much reduce the bloat by the same amount we increased it.

Why not start with the extender + unit tests, and see later what we want to do with that event ?

askvortsov1:
> I'm not sure if discussions happened elsewhere but reading this PR I have trouble finding benefits to the callbacks solution. They seem to just clutter the code compared to the previous solution.

As far as I understand, the goal of the extender system is to have people using this more structured/rigorous public API of extenders instead of events in cases where events aren't actually events.

However, then we're faced with the issue of how to handle events that are actually domain events, like Created, Deleted, Saving, LoggedIn, etc. While we could set up an extender for those, it'd be very cluttered, and wouldn't scale to domain events outside of Flarum (plenty of extensions register their own models, and we'd be forcing them to create their own extenders for those domain events, which would be messy, and to quote Franz,

> P.S.: I want extensions to mostly use extenders, not build their own.

So the solution we came up with recently was to keep domain events and listeners for them, through a dedicated Event->listen() extender. But if we have that, people might continue using it to listen into the non-domain events. So currently, the idea is to deprecate all of those and replace them with non-event-based extenders, and when that's stable, refactor the system to reduce overhead.

Here's an issue discussing how the extender idea came about: https://github.com/flarum/core/issues/851

src/Formatter/Formatter.php
```diff
@@ -20,6 +20,27 @@
...
+
+    public function addParsingCallback($callback)
+    {
+        $this->parsingCallbacks[] = $callback;
+    }
+
+    public function addRenderingCallback($callback)
+    {
+        $this->renderingCallbacks[] = $callback;
+    }
```
franzliedke, line 42: Please move these methods down to the other methods. :pray:

tests/integration/extenders/FormatterTest.php
```diff
@@ -0,0 +1,147 @@
...
+     */
+    public function custom_formatter_config_works_if_added_with_closure()
+    {
+        $this->extend((new Extend\Formatter)->configure(function ($config) {
+            $config->BBCodes->addFromRepository('B');
+        }));
+
+        // Needed so that the config is refreshed, this is usually handled by lifecyclehooks.
+        $formatter = $this->getFormatter();
+        $formatter->flush();
```
franzliedke, line 45: Can you expand on why this is necessary? This sounds like state from one test is leaking over to the next. We should try to get each test to always start with a completely flushed formatter (i.e. cache in this case, it seems).

src/Formatter/Formatter.php
askvortsov1: Done!

askvortsov1:
Shoo stale bot!

tankerkiller125 (approved):
Other than the depreciation messages needing updating the code looks OK to me. I have not tested this locally yet.

clarkwinkelmann (changes requested):
Looking good overall!

There are two more comments to update in `Formatter/Formatter.php` about beta 13.

I think the extender docblocks should use `callable|string` for the `$callable` instead of just `callable`, like we do in `Extend\Frontend::content` for example

clarkwinkelmann (approved):

PR #2099, opened 2020-03-29
Install prettier for consistent JS styling

franzliedke:
This sets up [prettier](https://prettier.io/), the "opinionated formatter" for JavaScript code. This is widely considered to be the most frictionless solution for consistent JavaScript styling, because it not only acts as linter, but can automatically format the code "correctly". I mostly care about consistency, not a particular code style, so I like this approach as well.

This PR configures Prettier v2 based on what @datitisev has experimented with in his frontend rewrite branch - with one important difference: we stick to the current tab width of two spaces, because that will keep the already large diffs smaller. Changing this (and other mostly subjective configuration options) should IMO be delayed to after the frontend rewrite has been done. My focus is on keeping diffs and changes small, reduce the amount of conflicts and so on.

The last commit was me running prettier on just a few files, so you can get a feeling of what the diff will look like for most of our files. I only did it for a few, so that the diff is mostly about the configuration and workflow changes.

Once we have approved this change and have solved the questions below, I will run Prettier once for all files in our codebase and then merge. (Might delay this to doing it on the master branch after merge.)

Open questions:
- How should developers integrate this into their workflow? Most IDEs support scripts to be executed after saving, but do we also want e.g. Git hooks that prevent commits without properly formatted JS?
- Git hooks can be disabled (they have to be enabled manually, as far as I know), so we also need a server-side enforcement. GitHub Actions or another form of continuous integration, basically.

**Required changes:**

  - [ ] Update contributor documentation

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,31 @@
+name: Lint code
+
+on:
+  push:
+    paths:
+      - 'js/src/**'
+  pull_request:
+    paths:
+      - 'js/src/**'
+
+jobs:
+  prettier:
+    runs-on: ubuntu-latest
+
+    name: Lint JS code with Prettier
+
+    steps:
+      - uses: actions/checkout@master
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: "12"
+
+      - name: Install JS dependencies
+        run: npm ci
+        working-directory: ./js
+
+      - name: Check JS code for formatting
+        run: node_modules/.bin/prettier --check src
+        working-directory: ./js
--- a/js/.prettierrc.json
+++ b/js/.prettierrc.json
@@ -1,0 +1,6 @@
+{
+  "printWidth": 150,
+  "singleQuote": true,
+  "tabWidth": 2,
+  "trailingComma": "es5"
+}
--- a/js/package-lock.json
+++ b/js/package-lock.json
@@ -3699,6 +3699,12 @@
       "resolved": "https://registry.npmjs.org/posix-character-classes/-/posix-character-classes-0.1.1.tgz",
       "integrity": "sha1-AerA/jta9xoqbAL+q7jB/vfgDqs="
     },
+    "prettier": {
+      "version": "2.0.2",
+      "resolved": "https://registry.npmjs.org/prettier/-/prettier-2.0.2.tgz",
+      "integrity": "sha512-5xJQIPT8BraI7ZnaDwSbu5zLrB6vvi8hVV58yHQ+QK64qrY40dULy0HSRlQ2/2IdzeBpjhDkqdcFBnFeDEMVdg==",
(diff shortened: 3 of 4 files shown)


datitisev:
We could format the code in the same commit (or a separate one) as building JS.

IDEs like VSCode and PhpStorm have plugins for Prettier.

I wouldn't *prevent* committing unformatted code, I'd automatically format it before3 committing so it's an automatic process. There's a package called [husky](https://www.npmjs.com/package/husky) that facilitates git hooks.

We could perhaps also have a commit status for properly formatted code, so it shows in PRs that they need to format their code (if they didn't allow pushing to the PR).

askvortsov1:
Agreed about having a check on pull request. Is there any way we could go a step beyond that, and have a bot automatically format and add that to any pull requests?

datitisev:
@askvortsov1 I suggested having that as well, might not've been clear. That doesn't work for all PRs if the author doesn't let us push to their branch - so that's why I suggested the check in case it's not supported (we could try/do both).

askvortsov1:
Ah my bad. Could we, instead, have a bot that makes a commit to master every time we merge a PR, fixing style issues?

datitisev:
Yes, which is why I said we could maybe format it in the same commit as building JS or separate, and automatically push the correct formatting.

tankerkiller125:
I love this idea, I should also note that their is a plugin for PHP as well, not sure if that's something we would want or not but just figured I'd point it out.

datitisev:
If I remember correctly, the PHP plugin isn't completely stable (and isn't by prettier themselves). Either way, we already have StyleCI so that's not needed.

franzliedke:
Not the biggest fan of auto-commit bots.

I'll do two things for now:
- Configure Husky for Git hooks.
- Add a CI step that fails when applying prettier yields changes, so that contributors are warned about it in PRs. Fixing it means running one console command, so I think that's a reasonable requirement. :wink: 

Those of us pushing to master will have to make sure they have the Git hooks set up, that should cover all cases,.

datitisev:
@franzliedke Sounds good! Just make sure to only run prettier if any changes occur to JS - that way we can avoid random style fixes if the JS wasn't touched.

datitisev:
I see that you added a `format-check` NPM command but don't use it in the GitHub workflow. Any reason as to that?

franzliedke:
> I see that you added a `format-check` NPM command but don't use it in the GitHub workflow. Any reason as to that?

@datitisev Good eye! I did originally use that, but `npm run` adds some extra ugly (and potentially confusing) output when the underlying script fails with a non-zero exit status. To avoid that, I used the `bin` file directly in the GitHub workflow.

> Also - I feel like the check name should be something like "JavaScript / Lint" or "Check Formatting" - it's a rather long name right now.

Good idea, I [shortened it now](https://github.com/flarum/core/commit/1592cd1013f2307e882267ebc179474164d3a0fd).

PR #2100, opened 2020-03-29
Added Model extender, integration tests

askvortsov1:
**Fixes part of #1891**

- Add Model extender with relationship method (decided not to go with one method-per-relationship as this is simpler, more flexible, and easier to operate). Also added configureDates and configureDefaultAttributes methods, which also take callables.
- Deprecated GetModelRelationship, ConfigureModelDefaultAttributes, ConfigureModelDates events
- Added integration tests for model relationship extender

Anything I'm missing?

**Confirmed**

### DIFF
--- a/src/Database/AbstractModel.php
+++ b/src/Database/AbstractModel.php
@@ -47,6 +47,12 @@ abstract class AbstractModel extends Eloquent
      */
     protected $afterDeleteCallbacks = [];
 
+    public static $customRelations = [];
+
+    public static $dateAttributes = [];
+
+    public static $defaults = [];
+
     /**
      * {@inheritdoc}
      */
@@ -68,58 +74,20 @@ abstract class AbstractModel extends Eloquent
     }
 
     /**
-     * An array of custom relationships defined by extensions.
-     */
-    protected static $customRelations = [];
-
-    protected static $dateCallbacks = [];
-
-    protected static $defaultAttributeCallbacks = [];
-
-    public static function addCustomRelation(string $from, string $name, $relation)
-    {
-        if (! array_key_exists($from, static::$customRelations)) {
-            static::$customRelations[$from] = [];
-        }
-
-        static::$customRelations[$from][$name] = $relation;
-    }
-
-    public static function addDateCallback(string $model, $callback)
-    {
-        if (! array_key_exists($model, static::$dateCallbacks)) {
-            static::$dateCallbacks[$model] = [];
-        }
-
-        static::$dateCallbacks[$model][] = $callback;
-    }
-
-    public static function addDefaultAttributeCallback(string $model, $callback)
-    {
-        if (! array_key_exists($model, static::$defaultAttributeCallbacks)) {
-            static::$defaultAttributeCallbacks[$model] = [];
-        }
-
-        static::$defaultAttributeCallbacks[$model][] = $callback;
-    }
-
-    /**
      * {@inheritdoc}
      */
     public function __construct(array $attributes = [])
     {
-        $defaults = [];
+        $this->attributes = Arr::get(static::$defaults, static::class, []);
 
         // Deprecated in beta 13, remove in beta 14.
         static::$dispatcher->dispatch(
-            new ConfigureModelDefaultAttributes($this, $defaults)
+            new ConfigureModelDefaultAttributes($this, $this->attributes)
         );
 
-        foreach (Arr::get(static::$defaultAttributeCallbacks, static::class, []) as $callback) {
-            $defaults = $callback($defaults);
-        }
-
-        $this->attributes = $defaults;
+        $this->attributes = array_map(function ($item) {
+            return is_callable($item) ? $item() : $item;
+        }, $this->attributes);
 
         parent::__construct($attributes);
     }
@@ -131,23 +99,11 @@ abstract class AbstractModel extends Eloquent
      */
     public function getDates()
     {
-        static $dates = [];
+        static::$dispatcher->dispatch(
+            new ConfigureModelDates($this, $this->dates)
+        );
 
-        $class = get_class($this);
-
-        if (! isset($dates[$class])) {
-            static::$dispatcher->dispatch(
-                new ConfigureModelDates($this, $this->dates)
-            );
-
-            $dates[$class] = $this->dates;
-        }
-
-        foreach (Arr::get(static::$dateCallbacks, static::class, []) as $callback) {
-            $dates[$class] = $callback($dates[$class]);
-        }
-
-        return $dates[$class];
+        return array_merge($this->dates, Arr::get(static::$dateAttributes, static::class, []));
     }
 
     /**
@@ -185,7 +141,7 @@ abstract class AbstractModel extends Eloquent
      */
     protected function getCustomRelation($name)
     {
-        $relation = Arr::get(Arr::get(static::$customRelations, static::class, []), $name, null);
+        $relation = Arr::get(static::$customRelations, static::class.".$name", null);
 
         if (! is_null($relation)) {
             return $relation($this);
--- a/src/Database/AbstractModel.php
+++ b/src/Database/AbstractModel.php
@@ -78,7 +78,7 @@ abstract class AbstractModel extends Eloquent
 
     public static function addCustomRelation(string $from, string $name, $relation)
     {
-        if (!array_key_exists($from, static::$customRelations)) {
+        if (! array_key_exists($from, static::$customRelations)) {
             static::$customRelations[$from] = [];
         }
 
@@ -87,7 +87,7 @@ abstract class AbstractModel extends Eloquent
 
     public static function addDateCallback(string $model, $callback)
     {
-        if (!array_key_exists($model, static::$dateCallbacks)) {
+        if (! array_key_exists($model, static::$dateCallbacks)) {
             static::$dateCallbacks[$model] = [];
         }
 
@@ -96,7 +96,7 @@ abstract class AbstractModel extends Eloquent
 
     public static function addDefaultAttributeCallback(string $model, $callback)
     {
-        if (!array_key_exists($model, static::$defaultAttributeCallbacks)) {
+        if (! array_key_exists($model, static::$defaultAttributeCallbacks)) {
             static::$defaultAttributeCallbacks[$model] = [];
         }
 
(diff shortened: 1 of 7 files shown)


tests/integration/extenders/ModelTest.php
```diff
@@ -0,0 +1,213 @@
...
+                $this->normalUser(),
+            ]
+        ]);
+    }
+
+    protected function tearDown()
+    {
+        if (! is_null($discussion = Discussion::find(1))) {
+            $discussion->delete();
+        }
```
franzliedke, line 39: You can achieve the same (completely independently from other tests that might run *before this one*) instead by adding the following line to the `prepareDatabase()` call:

~~~php
'discussions' => [],
~~~

Any table that is referenced in that call will be completely emptied before being filled with the provided rows - in this case, we would only empty the table.

tests/integration/extenders/ModelTest.php
```diff
@@ -0,0 +1,205 @@
...
+
+        $this->assertNotEquals('Custom Default', $user->name_singular);
+    }
+
+    /**
+     * @test
+     */
+    public function custom_date_doesnt_exist_if_not_set()
+    {
+        $post = new Post;
```
franzliedke, line 161: Please run `$this->app()` in this test (and some of the other ones) as well, for maximum comparability. :grin:

tests/integration/extenders/ModelTest.php
```diff
@@ -0,0 +1,205 @@
...
+
+        $user = new User;
+
+        $this->assertNotEquals('Custom Default', $user->name_singular);
+    }
+
+    /**
+     * @test
+     */
+    public function custom_date_doesnt_exist_if_not_set()
```
franzliedke, line 159: This description doesn't seem to match the implementation. :thinking:

tests/integration/extenders/ModelTest.php
```diff
@@ -0,0 +1,205 @@
...
+    {
+        $post = new Post;
+
+        $this->assertContains('hidden_at', $post->getDates());
+    }
+
+    /**
+     * @test
+     */
+    public function custom_date_works_if_set()
```
franzliedke, line 169: Please expand this name to describe what is actually tested here...

src/Extend/Model.php
```diff
@@ -0,0 +1,104 @@
...
+     * @param callable $callable
+     *
+     * The callable can be a closure or invokable class, and should accept:
+     * - \Flarum\User\User $user: the user in question.
+     * - array $dateAttributes: an array of attributes which are processed as dates.
+     *
+     * The callable should return:
+     * - array $dateAttributes: an array of attributes which are processed as dates.
+     */
+    public function configureDates(callable $callable)
```
franzliedke, line 43: Honestly, I don't see the use-case for full-blown manipulation currently (and also not in our existing - bundled - extensions). Can we stick to `date($attributeName)` for now?

src/Extend/Model.php
```diff
@@ -0,0 +1,104 @@
...
+     * @param callable $callable
+     *
+     * The callable can be a closure or invokable class, and should accept:
+     * - \Flarum\User\User $user: the user in question.
+     * - array[string] $dateAttributes: an associative array of attributes => default values.
+     *
+     * The callable should return:
+     * - array[string] $dateAttributes: an associative array of attributes => default values.
+     */
+    public function configureDefaultAttributes(callable $callable)
```
franzliedke, line 62: Same here, I would prefer `defaultAttribute($name, $value)` here. `$value` could either be a static value (for simple scalars) or a callable. 

Naming: What do you think about `defaultValue(...)` instead?

src/Extend/Model.php
```diff
@@ -0,0 +1,104 @@
...
+    public function relationship(string $name, callable $callable)
+    {
+        $this->relationships[$name] = $callable;
+
+        return $this;
+    }
+
+    public function extend(Container $container, Extension $extension = null)
+    {
+        $container->extend('flarum.model.customRelations', function ($existingRelations) {
```
franzliedke, line 92: Sorry for my OCD, but I think we don't (yet?) have upper-case in our IoC bindings anywhere.

I want... I need... this to stay that way!!!!111!!1one1!!! :stuck_out_tongue_winking_eye:

src/Database/DatabaseServiceProvider.php
```diff
@@ -54,6 +54,18 @@ public function register()
 
         $this->app->alias(ConnectionInterface::class, 'db.connection');
         $this->app->alias(ConnectionInterface::class, 'flarum.db');
+
+        $this->app->singleton('flarum.model.customRelations', function () {
+            return [];
+        });
```
franzliedke, line 60: Out of interest: does `extend()` (in the extenders) also work if you use `$this->app->instance('...', [])` here?

src/Database/DatabaseServiceProvider.php
```diff
@@ -63,5 +75,35 @@ public function boot()
...
+        }
+
+        foreach ($this->app->make('flarum.model.dateCallbacks') as $modelClass => $callbacks) {
+            foreach ($callbacks as $callback) {
+                if (is_string($callback)) {
+                    $callback = $this->app->make($callback);
+                }
+
+                AbstractModel::addDateCallback($modelClass, $callback);
+            }
```
franzliedke, line 96: If we decide to switch to the non-callback solution which I proposed in the other comment, this could just assign directly to the (public) property on the model.

I don't think we should allow injection from the container into the callbacks (this would only affect default attributes, as per my proposal), that seems like overkill for this use case.

src/Extend/Model.php
```diff
@@ -0,0 +1,104 @@
...
+     * @param callable $callable
+     *
+     * The callable can be a closure or invokable class, and should accept:
+     * - $instance: An instance of this model.
+     *
+     * The callable should return:
+     * - $relationship: A Laravel Relationship object. See relevant methods of models
+     *                  like \Flarum\User\User for examples of how relationships should be returned.
+     */
+    public function relationship(string $name, callable $callable)
```
franzliedke, line 83: For these ones, I think that accepting the arguments here directly makes using these extenders much nicer.

We'd have two options:
~~~php
->relationship('posts', 'hasMany', ...)

->hasMany('posts', ...)
~~~

Personally, I prefer the second option, as that's even less typing and the list is not very likely to get any bigger. If you experiment a bit with [variadic arguments](https://www.php.net/manual/en/functions.arguments.php#functions.variable-arg-list), I guess you won't even have to explicitly pass on all the different arguments for each type. (Except for listing them in the docblock, I guess.)

franzliedke: This would further reduce the boilerplate for certain relationship types, by the way, because we would not always to specify [lots of default values just to set the last argument](https://github.com/flarum/mentions/blob/9ac08e302d01849a6252fe4339c8eb1b415a12b3/src/Listener/AddPostMentionedByRelationship.php#L40), which the extender could derive from the first argument (the relationship name).

src/Extend/Model.php
askvortsov1: Agreed, I put it like this to make it maximally compatible with the old system, but I've been leaning towards this too.

src/Extend/Model.php
askvortsov1: good point

src/Database/DatabaseServiceProvider.php
askvortsov1: Did not know that existed, interesting! Not sure, can try to test though.

src/Database/DatabaseServiceProvider.php
askvortsov1: Agreed

src/Extend/Model.php
askvortsov1: I put it to just `->default($attribute, $value)` for now

src/Extend/Model.php
askvortsov1: No longer exists!

src/Database/DatabaseServiceProvider.php
askvortsov1: No longer exists, so irrelevant

src/Database/DatabaseServiceProvider.php
askvortsov1: All are done

src/Extend/Model.php
askvortsov1: I'm going to keep the main relationship one that accepts a callback for edge cases, and provide specific ones (that use the relationship one) for the main cases.

askvortsov1:
For the Default Attributes one, do we want to allow extension developers to use stuff like settings there? We could pass a predetermined number of things (translator, settings, URL Generator), and have them use those for the default attributes.

franzliedke:
Not for now, I'd say. We can consider that when the use-case pops up. Until then it feels just like wild guessing. :wink:

src/Extend/Model.php
askvortsov1: Done.

src/Database/AbstractModel.php
```diff
@@ -89,19 +99,11 @@ public function __construct(array $attributes = [])
...
-            );
-
-            $dates[$class] = $this->dates;
-        }
+        static::$dispatcher->dispatch(
+            new ConfigureModelDates($this, $this->dates)
+        );
 
-        return $dates[$class];
+        return array_merge($this->dates, Arr::get(static::$dateAttributes, static::class, []));
```
franzliedke, line 106: Should we keep the cache semantics of the previous implementation - so that the event dispatch and the array merge only happens once per class per request?

(Not sure how often this is called internally... it might not be relevant any longer when the event dispatching is gone.)

franzliedke: I'll take care of this!

franzliedke: Played around with it. Static caching would cause problems in tests (uh oh), and I'm not too worried about it once the event dispatching is gone, so I'm leaving this as is.

clarkwinkelmann:
I'm only catching up now with those features.

I understand you added a way to specify new date attributes on the model, but not casts ? I'm pretty sure casts would be useful. Should we make an issue for that ?

askvortsov1:
The goal here is to move to the extenders as public API (no functionality was changed), but I agree that would be useful.

PR #2101, opened 2020-03-29
Added post extender with type method, deprecated ConfigurePostTypes

askvortsov1:
**Fixes part of  #1891 , #2418**

- Added post types extender
- Added integration tests
- Deprecated ConfigurePostTypes

**Confirmed**


askvortsov1:
Not stale

src/Extend/Post.php
```diff
@@ -0,0 +1,35 @@
...
+    public function type($postType)
+    {
+        ActualPost::setModel($postType::$type, $postType);
+
+        return $this;
+    }
+
+    public function extend(Container $container, Extension $extension = null)
+    {
+        // Nothing happens here, type does all the work.
```
clarkwinkelmann, line 33: There's probably not much of a difference, but wouldn't it still be better to store posts in an array and call `setModel` from here?

I believe this might delay the resolving of the classname until extensions actually boot? Not sure how that would be useful but resolving as late as possible is always nice.

src/Extend/Post.php
```diff
@@ -0,0 +1,35 @@
...
+ * This file is part of Flarum.
+ *
+ * For detailed copyright and license information, please view the
+ * LICENSE file that was distributed with this source code.
+ */
+
+namespace Flarum\Extend;
+
+use Flarum\Extension\Extension;
+use Flarum\Post\Post as ActualPost;
```
clarkwinkelmann, line 13: Wouldn't it make more sense to alias this as `PostModel`?

src/Extend/Post.php
```diff
@@ -0,0 +1,35 @@
...
+{
+    /**
+     * Register a new post type. This is generally done for custom 'event posts',
+     * such as those that appear when a discussion is renamed.
+     *
+     * @param string $postType: The ::class attribute of the custom Post type that is being added.
+     */
+    public function type($postType)
+    {
+        ActualPost::setModel($postType::$type, $postType);
```
clarkwinkelmann, line 26: Not actually something here, but maybe we could take the opportunity to add parameter type to `setModel` ? (string and string)

clarkwinkelmann (commented):
Looking good. A number of small comments.

clarkwinkelmann (approved):

SychO9 (commented):
Not a blocker for this PR but for consistency's sake we could also rename `registerPostTypes` to `setPostTypes` and make it protected.

tankerkiller125 (approved):

PR #2102, opened 2020-03-29
Validator extender

askvortsov1:
**Fixes part of #1891**

- Add Validator extender with configure method
- Add integration tests
- Deprecate Validating event

I slightly deviated from the previously discussed spec on this one, as I think that having a configure method that runs the validator through a callable gives us and developers more flexibility than simply the ability to add rules. Similarly, there are extensions that depend on this type of capabilities.

Should there be a dedicated method to run the ruleset through a callable as well? I'm not sure if that is necessary (this PR supports that), but it might be a nice shortcut.

**Confirmed**

### DIFF
--- a/src/Extend/Validator.php
+++ b/src/Extend/Validator.php
@@ -32,12 +32,14 @@ class Validator implements ExtenderInterface
 
     public function extend(Container $container, Extension $extension = null)
     {
-        foreach ($this->configurationCallbacks as $callback) {
-            if (is_string($callback)) {
-                $callback = $container->make($callback);
-            }
+        $container->resolving($this->validator, function ($validator, $container) {
+            foreach ($this->configurationCallbacks as $callback) {
+                if (is_string($callback)) {
+                    $callback = $container->make($callback);
+                }
 
-            AbstractValidator::addConfiguration($this->validator, $callback);
-        }
+                $validator->addConfiguration($callback);
+            }
+        });
     }
 }
--- a/src/Extend/Validator.php
+++ b/src/Extend/Validator.php
@@ -1,0 +1,43 @@
+<?php
+
+/*
+ * This file is part of Flarum.
+ *
+ * For detailed copyright and license information, please view the
+ * LICENSE file that was distributed with this source code.
+ */
+
+namespace Flarum\Extend;
+
+use Flarum\Extension\Extension;
+use Flarum\Foundation\AbstractValidator;
+use Illuminate\Contracts\Container\Container;
+
+class Validator implements ExtenderInterface
+{
+    private $configurationCallbacks = [];
+    private $validator;
+
+    public function __construct($validator)
+    {
+        $this->validator = $validator;
+    }
+
+    public function configure($callback)
+    {
+        $this->configurationCallbacks[] = $callback;
+
+        return $this;
+    }
+
+    public function extend(Container $container, Extension $extension = null)
+    {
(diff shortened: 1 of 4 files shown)


askvortsov1:
Now, for validator extender I've looked through the code, and it's a tad bit tricker. The most commonly used things that SHOULD be explicit on the API are:

```
->addRule($attribute, $rule)
->removeRule($attribute, $rule) // Rule supports wildcard pattern matching

// For the following 2, I propose this system so that we force extension devs to translate any custom error messages they want to add.
->addMessage($rule, $messageTranslationKey)
->removeMessage($rule, $messageTranslationKey)

->addExtension($rule, $extension)
```

These alone should cover more than 99.9% of use case; however, there are other public methods on the validator API that, theoretically speaking, should be supported. Do we want to keep the generic "configure" method as well, or support every single one of those methods, or just support some?

Also, there's cases like this: https://github.com/branchzero/flarum-ext-qqcaptcha/blob/b043e38792614425d83f6085209c0b7486c1e57f/src/Listener/AddValidatorRule.php that use settings in deciding validation rules/extenders, what should be done there?

franzliedke:
> Similarly, there are extensions that depend on this type of capabilities.

Could you provide a few examples / links? It's hard to think about this in the abstract.

askvortsov1:
> Could you provide a few examples / links? It's hard to think about this in the abstract.
https://github.com/FriendsOfFlarum/doorman/blob/bb428bcc6aa2e2f30d91fbe7361d7e7e3c19249c/src/Listeners/AddValidatorRule.php
https://github.com/FriendsOfFlarum/recaptcha/blob/b9df60ff5604c61e621b479c88823855212564f0/src/Listeners/AddValidatorRule.php
https://github.com/clarkwinkelmann/flarum-ext-passwordless/blob/a772005219b4005c8a4105c50e3f137093c90918/src/Listeners/DontRequirePasswordOnSignUp.php

Really, the only/biggest question is what to do with validator extensions, as they will often require DI)

askvortsov1:
Not stale

clarkwinkelmann (changes requested):
The deprecation messages will need to be updated (beta13 => beta14)

Regarding the code of the test, I'm not sure whether we want to explain what it does or simplify it? I believe it's copied from one of my example and the reason it's so complex is to keep the additional password rules that might have been made by other extensions. I think having a comment reminding that would be useful for future readers.

Or we could also use a shorter code that blindly replaces the rules without regard for other extensions. This might be something legitimately used in the local `extend.php` when the list of extensions is well known. Something like: (untested)

```php
        $this->extend((new Extend\Validator(UserValidator::class))->configure(function ($validator) {
                $validator->setRules([
                    'password' => [
                        'required',
                        'min:8'
                    ]
                ] + $validator->getRules());
            }
        }));
```

I think it would still make sense to pass the Flarum validator object to the callback. It could be the second param instead of the first like it was with the event. Sometimes values are stored on that object, for example the user instance on `UserValidator`.

There's also the not-exactly-related-but-somewhat-close matter of #2398 where an event might or might not make sense (?)

clarkwinkelmann (approved):
I have not tested locally, but this looks good :+1: 

Should the tests cover calling a callback/class that takes a single argument? This is something that should work in PHP? One of the two (callback or class) could omit the second argument on purpose (with a comment explaining it's omitted)

tankerkiller125 (approved):

Issue #1355, opened 2018-01-31
Revamp search API


tobscure:
Plan for new search system:

Search driver (eg. Algolia) is called upon to update its index each time a post/discussion is changed
(Each post is a separate object in the index which contains not only post content but also its discussion's title/tags/other metadata)

q = foo "bar baz" (author:toby OR author:franz) is:following

1. Parse `q` into an abstracted query object . eg. https://github.com/pimcore/search-query-parser 
2. Loop through each `Term` in the query object, allow gambits to match and convert term into a `Where` object to be processed by search driver
3. Extensions may also populate the query object to enact permissions, eg. add a `Where` to restrict results to only tags that the user can see
4. Search driver builds its vendor-specific query by inspecting query object's `Terms`/`Wheres`, returns list of discussion IDs

However, this is complex and will take a while to implement, so it would be wise to keep it slated for 0.2 (as per current roadmap). In the meantime, we are doing #1339

franzliedke:
After continually jumping back and forth between performance issues and little bugs in our search implementation, it has become clear that we need to make progress on our driver-based search infrastructure (#509). This would then allow for swapping out a default, naive implementation (hand-maintained index, similar to FluxBB; MySQL fulltext search; etc.) for more advanced drivers based on e.g. Elasticsearch or Algolia.

I was asked to describe my thoughts on the initial architecture in a new ticket. But we already have this one, so I'll add my thoughts here.

---

### Laravel Scout

Multiple people brought up using [Laravel Scout](https://laravel.com/docs/6.x/scout), as it incorporates a driver infrastructure for searches on Eloquent models. This *might* be possible to use, but I am somewhat skeptical for several reasons:

- We want to allow extensions to change things at runtime.
- Tight coupling to Eloquent models / queries (possibly not compatible with Toby's approach outlined above)
- We want to define interfaces based on Flarum's functional requirements. Laravel's APIs usually change more often.

What might be possible is using Scout's [engines](https://github.com/laravel/scout/tree/c145b4dd752ccdda7855344981dd1815ee0ae746/src/Engines) (i.e. drivers) so that we have less work in building out the adapters to the various backends. We would still have to take care of the integration into Eloquent / our search infrastructure manually, though. Definitely worth a shot when we get to the implementation phase.

---

### Requirements

These are the filters / gambits supported by our current search layer...

- Searching users:
  - email
  - group name

- Searching discussions:
  - author name
  - tag
  - creation date (exact date, date range)
  - is hidden
[...]

PR #2104, opened 2020-03-30
Search Refactor and Extenders

### LINKED ISSUE #2071: Guest users can't load more posts on profiles
## Bug Report

**Current Behavior**
When browsing forums as a guest, the "Load More" button for posts is not shown on user profiles, however if you log in and scroll to the bottom of posts on a profile, the option is shown. 

Occurs on discuss.flarum.org

askvortsov1:
**Fixes part of flarum/framework#1891 **
Fixes flarum/framework#2071 
Moves towards flarum/issue-archive#286, flarum/issue-archive#443, probably others
Implements my ideas from https://github.com/flarum/core/issues/1355#issuecomment-605503004

- [x] Bring UserSearcher and DiscussionSearcher under a common, extendable system
- [x] Refactor ListPostsController to use this same system
- [x] Add a search extender and integration tests
- [x] Replace searching events with "search mutator" callbacks

- Should each model's gambit registration be moved to the appropriate service provider?
- Should the explicit ID list used in ListPostsController be accomodated by the AbstractSearcher? If so, should other filters apply here, or should only the ID list be used?
- Should we change how we accept search queries? e.g. sending them as JSON instead of CSV to begin with? For example, filter parameters could be sent separately.
- Do we need to support compat for ConfigurePostsQuery no longer being functional? It's going to be a pain/very hacky to get it to work with the refactor, and I was unable to find any extensions other than tags and mentions (which we have control over) using them? Given the work needed, and potential issues involved in, supporting it, and considering that the impact would be extremely minimal, I think this would be a justifiable breaking change.
- Franz, does this more-or-less move towards your vision for the search API?

**Confirmed**


franzliedke:

askvortsov1:
I updated the extender name to FlarumSearch, so that we can use Search for the extender that will manage search drivers. That way, we can merge this, and proceed with further PRs as the discussion in https://github.com/flarum/core/issues/1355 evolves.

franzliedke:
We'll revisit this for Beta 15. I like that this makes the `ConfigurePostTypes` event obsolete, as far as I can tell.

askvortsov1:
We might want to consider moving whereVisibleTo checks after gambits to ensure that filtering doesn't accidentially expose content that should not be exposed. If that has performance implications on the full text search, I think we should call gambits and full text search separately with `whereVisibleTo` in between.

Also, if we want to keep the Searching events, we should provide them with the full Eloquent builder, not just the query builder.

askvortsov1:
Aside from comments on implementation, the main thing we need to figure out here is whether we want to make any changes to these APIs so that https://github.com/flarum/core/issues/1355 minimizes breaking changes. Some ideas:

- Should we pass in the full Eloquent query object to gambits?
- Should we get a full list of gambits that will be run before they are actually run, so that gambits can be aware of other gambits?
- Could we add an object/array to AbstractSearch that gambits could modify to add data about what they're filtering, and which could be assembled/applied in the full text gambit (as suggested in https://github.com/flarum/core/issues/1355#issuecomment-719970809)
- Would operations on other search drivers use a different extender, or the same one? If so, we should rename this extender to `Search`, and initialize it with an argument, which is a string identifier for the driver in question.
- We might want to separate the "full text gambit" from the rest of the gambits, and treat it as its own operation. Or at least rename it since "full text" is an implementation name for MySQL?
- We might want to provide the full text gambit with info about sort, offset, limit, etc if those are (at least partially) handled by the search engine being used.

luceos (changes requested):
This PR is so massive, it's insane. This needs to be well tested.

- deprecated messages need to be updated (beta 13 => beta 15)
- issue has to be created to remove the hacky temporary workaround

I left one comment where you've used raw queries. Raw queries are exceptionally risky, try to refrain from using them too much.

askvortsov1:
Since we've changed our mind on how we want to approach this a bit, I'm going to close it, but then use pieces from it in other PRs.

PR #2108, opened 2020-03-31
Scroll to edited post or inform the user

### LINKED ISSUE #634: Editing a post from another page
Not sure whether this has been reported, but I experienced the following on the discuss forum;
- Reply in a thread, opens the composer, browse away to the forum index for instance, finish the message and post; "view" button pops up -- OK
- Edit an existing post of yours, opens the composer, browse away to the forum index for instance, finish editing and post; NO pop up..

I would say it would be very useful to also have a "view" button pop up in these circumstances?

the-turk:
**Fixes #634**

Pretty much copy/pasted from [ReplyComposer](https://github.com/flarum/core/blob/02ceed4fedb5eb62ae95c38306a62a3e18f22793/js/src/forum/components/ReplyComposer.js#L74).

> If we're currently viewing the discussion which post edit was made in, then we can scroll to the post. Otherwise, we'll create an alert message to inform the user that their edit has been made, containing a button which will transition to their edited post when clicked.

* Translations.
* Did I miss something?

**Confirmed**

**Required changes:**

- [x] Related core extension PRs: flarum/lang-english#163

### DIFF
--- a/js/src/forum/components/EditPostComposer.js
+++ b/js/src/forum/components/EditPostComposer.js
@@ -1,4 +1,6 @@
 import ComposerBody from './ComposerBody';
+import Alert from '../../common/components/Alert';
+import Button from '../../common/components/Button';
 import icon from '../../common/helpers/icon';
 
 function minimizeComposerIfFullScreen(e) {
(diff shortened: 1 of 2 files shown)


franzliedke:
@the-turk Thanks for another nice PR! Keep 'em coming. :+1: 

As you can see, we have a lot of open PRs currently. Our top goal right now is to keep our focus on the stable release, and [our planned path to get there](https://github.com/flarum/core/milestones).

As this issue (#634) was originally planned for a later release, it might take a while until we get around to further reviews (and merging it). Now that so much work has already been done, that does not mean we cannot merge a good PR if we find the time, but I wanted to explain what we're trying to keep in mind when choosing what to review next, in order to avoid confusion or disappointment. We are trying hard not to get sidetracked in our focus, and our resources are limited.

Thanks for your understanding - and your contribution! :hugs:

the-turk:
Thanks @franzliedke, I appreciate that! I'm aware of your priorities and hard work. I just wanted to lighten the load on y'all in my free time and these are just my humble contributions. I'll leave it here for future reference but I should say that I'm not happy with using the same code multiple times either, feel free to close it if you've got better ideas to add the feature. Thanks again. 🤜

franzliedke:

the-turk:
Did I do this right? 🙈

datitisev:
@the-turk The lock file comes from `npm install`.

This is basically the same logic as the reply alert, right? Not sure if any of it can be moved into a common place so there's less duplication, if that's true.

the-turk:
hey @datitisev, that's true

> 
> 
> I should say that I'm not happy with using the same code multiple times either, feel free to close it if you've got better ideas to add the feature. Thanks again. 🤜

datitisev:
Not sure if we want to simply stuff in this PR or make it separate. 🤷 Either works for me.

askvortsov1:
For this case in particular, I'm fine having the code structure repeated. It's only really used twice, if we have to add it again, then we can try to abstract it out. Additionally, there's enough differences between the two that abstracting out the logic would likely create a messier codebase, not a cleaner one

datitisev:
We never merged the locale strings for this PR...

PR #2110, opened 2020-03-31
User Extender (prepareGroups functionality)

askvortsov1:
**Fixes #0000**

- Added User extender with addGroupProcessor method
- Added integration tests
- Deprecated PrepareUserGroups

Are we ok with the naming?

**Confirmed**

### DIFF
--- a/src/Event/PrepareUserGroups.php
+++ b/src/Event/PrepareUserGroups.php
@@ -12,6 +12,7 @@ namespace Flarum\Event;
 use Flarum\User\User;
 
 /**
+ * @deprecated beta 13, remove in beta 14. Use the User extender instead.
  * The `PrepareUserGroups` event.
  */
 class PrepareUserGroups
--- a/src/Extend/User.php
+++ b/src/Extend/User.php
@@ -10,7 +10,6 @@
 namespace Flarum\Extend;
 
 use Flarum\Extension\Extension;
-use Flarum\User\User as ActualUser;
 use Illuminate\Contracts\Container\Container;
 
 class User implements ExtenderInterface
--- a/src/Extend/User.php
+++ b/src/Extend/User.php
@@ -54,18 +54,12 @@ class User implements ExtenderInterface
 
     public function extend(Container $container, Extension $extension = null)
     {
-        // Display name config
         $container->extend('flarum.user.display_name.supported_drivers', function ($existingDrivers) {
             return array_merge($existingDrivers, $this->displayNameDrivers);
         });
 
-        // Group processor config
-        foreach ($this->groupProcessors as $callback) {
-            if (is_string($callback)) {
-                $callback = $container->make($callback);
-            }
-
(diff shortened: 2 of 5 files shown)


src/Extend/User.php
```diff
@@ -0,0 +1,46 @@
...
+     * @param callable $callable
+     *
+     * The callable can be a closure or invokable class, and should accept:
+     * - \Flarum\User\User $user: the user in question.
+     * - array $groupIds: an array of ids for the groups the user belongs to.
+     *
+     * The callable should return:
+     * - array $groupIds: an array of ids for the groups the user belongs to.
+     */
+    public function addGroupProcessor(callable $callable)
```
franzliedke, line 33: `addGroupProcessor` again sounds awkwardly close to our internal implementation details. What about just `groups()`? Since we're passing a callback anyway, this allows for all kinds of manipulation, so we're golden.

Since the callbacks are not named, there isn't going to be any removal anyway, so by convention we wouldn't use the `add()` prefix here.

src/User/UserServiceProvider.php
```diff
@@ -82,6 +86,14 @@ public function boot()
             return false;
         });
 
+        foreach ($this->app->make('flarum.user.groupProcessors') as $callback) {
```
franzliedke, line 89: AFAIK, we use `snake_case` for multi-word services (at least in one place I found). :wink:

tests/integration/extenders/UserTest.php
```diff
@@ -0,0 +1,62 @@
...
+        $this->prepareDatabase([
+            'users' => [
+                $this->normalUser(),
+            ],
+        ]);
+    }
+
+    public function tearDown()
+    {
+        User::find(2)->delete();
```
franzliedke, line 32: This shouldn't be necessary with the way `prepareDatabase()` works?

src/Extend/User.php
franzliedke: I just read the docblock, ahem. :see_no_evil: 

Since this is just for calculating permissions, I guess `groups()` would be a bit too generic and could confuse developers. Hmm... `permissionGroups()`? Other ideas?

src/User/UserServiceProvider.php
tankerkiller125: @franzliedke I think I've seen snake case and dot notated in the code, we might want to work on truly standardizing that (like we are working towards with permissions)

franzliedke: The dot notation is fine, I was only referring to the last part.

src/Event/PrepareUserGroups.php
```diff
@@ -12,6 +12,7 @@
 use Flarum\User\User;
 
 /**
+ * @deprecated beta 13, remove in beta 14. Use the User extender instead.
```
franzliedke, line 15: These numbers are not correct (anymore :see_no_evil: ). Will fix this in master.

PR #2111, opened 2020-04-01
Fix extensions page layout

### LINKED ISSUE #2029: Extensions page on Firefox is bugged out
**Browser: Firefox 73**
image

**Browser: Chromium (Vivaldi)**
image

the-turk:
**Fixes #2029**

`break-inside: avoid-column;` is [not supported on Firefox](https://caniuse.com/#feat=mdn-css_properties_break-inside_multicol_context_column), yet they have `page-break-inside` property to prevent breaks in **columns**, as well as pages. Also, `-webkit-column-break-inside` should be used for WebKit-based browsers (like Safari). See [MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/break-inside) for further information.

There are also couple of work-arounds I've found on [StackOverflow](https://stackoverflow.com/a/7785711) but it seems like the `page-break-inside` property fixes our issue without needing any of them.

**Confirmed**

### DIFF
--- a/less/admin/ExtensionsPage.less
+++ b/less/admin/ExtensionsPage.less
@@ -28,7 +28,9 @@
   .clearfix();
 
   > li {
+    -webkit-column-break-inside: avoid;
     break-inside: avoid-column;
+    page-break-inside: avoid;
     text-align: left;
     position: relative;
     border-radius: 4px;


luceos:
Thank you @the-turk that's extremely useful 👏

PR #2113, opened 2020-04-02
Fix Mods being able to edit Admins, Split permissions

### LINKED ISSUE #1965: Users with edit user permission can make themselves admins
## Bug Report

**Current Behavior**
If a moderator is assigned with the permission to edit users, the user is also able to make itself admin by editing and assigning the administrator role to itself.

**Expected Behavior**
Only administrators should be able to assign or remove the admin role on a user.

**Environment**

tankerkiller125:
**Fixes flarum/framework#1965 **

- Currently only prevents non-admins from editing admin
- Splits the edit.user permission into three (edit.username, edit.credentials, edit.groups)

- Is this the best way of doing this?
- Do we want to limit what shows up on the edit screen to only what users have permissions to do?

**Screenshot**

image

**Confirmed**

**Required changes:**
- [x] Related core extension PRs: (Remove if irrelevant)
  - flarum/english#164

### DIFF
--- a/js/src/forum/components/EditUserModal.js
+++ b/js/src/forum/components/EditUserModal.js
@@ -61,7 +61,7 @@ export default class EditUserModal extends Modal {
       );
     }
 
-    if (app.session.user !== this.props.user && app.session.user.canEditCredentials()) {
+    if (app.session.user.canEditCredentials()) {
       items.add(
         'email',
         <div className="Form-group">
--- a/js/src/forum/components/EditUserModal.js
+++ b/js/src/forum/components/EditUserModal.js
@@ -46,20 +46,24 @@ export default class EditUserModal extends Modal {
   fields() {
     const items = new ItemList();
 
-    items.add(
-      'username',
-      <div className="Form-group">
+    if (app.session.user.canEditUsername()) {
+      items.add('username', <div className="Form-group">
         <label>{app.translator.trans('core.forum.edit_user.username_heading')}</label>
-        <input className="FormControl" placeholder={extractText(app.translator.trans('core.forum.edit_user.username_label'))} bidi={this.username} />
-      </div>,
-      40
-    );
+        <input className="FormControl" placeholder={extractText(app.translator.trans('core.forum.edit_user.username_label'))}
+             bidi={this.username} />
+      </div>, 40);
+    }
 
-    if (app.session.user !== this.props.user) {
-      items.add(
-        'email',
-        <div className="Form-group">
-          <label>{app.translator.trans('core.forum.edit_user.email_heading')}</label>
+    // The first part prevents silent account hijacking
+    if (app.session.user !== this.props.user && app.session.user.canEditCredentials()) {
+      items.add('email', <div className="Form-group">
+        <label>{app.translator.trans('core.forum.edit_user.email_heading')}</label>
+        <div>
+          <input className="FormControl"
+                 placeholder={extractText(app.translator.trans('core.forum.edit_user.email_label'))}
+                 bidi={this.email}/>
+        </div>
+        {!this.isEmailConfirmed() ? (
           <div>
             <input className="FormControl" placeholder={extractText(app.translator.trans('core.forum.edit_user.email_label'))} bidi={this.email} />
           </div>
@@ -131,22 +135,35 @@ export default class EditUserModal extends Modal {
               </label>
             ))}
         </div>
-      </div>,
-      10
-    );
+      </div>, 20);
+    }
 
-    items.add(
-      'submit',
-      <div className="Form-group">
-        {Button.component({
-          className: 'Button Button--primary',
-          type: 'submit',
-          loading: this.loading,
-          children: app.translator.trans('core.forum.edit_user.submit_button'),
-        })}
-      </div>,
-      -10
-    );
+    if (app.session.user.canEditGroups()) {
+      items.add('groups', <div className="Form-group EditUserModal-groups">
+        <label>{app.translator.trans('core.forum.edit_user.groups_heading')}</label>
+        <div>
+          {Object.keys(this.groups)
+            .map(id => app.store.getById('groups', id))
+            .map(group => (
+              <label className="checkbox">
+                <input type="checkbox"
+                     bidi={this.groups[group.id()]}
+                     disabled={this.props.user.id() === '1' && group.id() === Group.ADMINISTRATOR_ID} />
+                {GroupBadge.component({group, label: ''})} {group.nameSingular()}
+              </label>
+            ))}
+        </div>
+      </div>, 10);
+    }
+
+    items.add('submit', <div className="Form-group">
+      {Button.component({
+        className: 'Button Button--primary',
+        type: 'submit',
+        loading: this.loading,
+        children: app.translator.trans('core.forum.edit_user.submit_button')
+      })}
+    </div>, -10);
 
     return items;
   }
--- a/js/src/forum/components/EditUserModal.js
+++ b/js/src/forum/components/EditUserModal.js
@@ -61,7 +61,6 @@ export default class EditUserModal extends Modal {
       );
     }
 
-    // The first part prevents silent account hijacking
     if (app.session.user !== this.props.user && app.session.user.canEditCredentials()) {
       items.add(
         'email',
--- a/js/src/forum/components/EditUserModal.js
+++ b/js/src/forum/components/EditUserModal.js
@@ -47,23 +47,26 @@ export default class EditUserModal extends Modal {
     const items = new ItemList();
 
     if (app.session.user.canEditUsername()) {
-      items.add('username', <div className="Form-group">
-        <label>{app.translator.trans('core.forum.edit_user.username_heading')}</label>
-        <input className="FormControl" placeholder={extractText(app.translator.trans('core.forum.edit_user.username_label'))}
-             bidi={this.username} />
-      </div>, 40);
+      items.add(
+        'username',
+        <div className="Form-group">
+          <label>{app.translator.trans('core.forum.edit_user.username_heading')}</label>
+          <input
+            className="FormControl"
+            placeholder={extractText(app.translator.trans('core.forum.edit_user.username_label'))}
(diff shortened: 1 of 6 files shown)


luceos (changes requested):
For some reason I cannot add comments to the Serializer. 

```
$canEdit = $gate->allows('edit.credentials') || $gate->allows('edit.groups') || $gate->allows('edit.username') ? true : false;
```
The last `?:` is superfluous. I also highly doubt it does what you expect. The condition applies only to the last gate allow method, not all three.

Why exactly are there no new policy methods, are these magically applied? Would you have kept the `user` as an argument to the `allows` method and created the methods on the `UserPolicy` you could have moved the `EditUserHandler` change to the Policy to enforce it globally, I think that's the better solution here..

tankerkiller125:
Thanks for the feedback @luceos I'll work on implementing your suggestions, I think the problem stems from my lack of experience working with Flarums permission systems and not know exactly how policies got applied.

askvortsov1 (changes requested):
In theory this could work without a policy: if no policy governs a permission, there'll be a check as to whether the user belongs to a group that has that permission. 

However, I agree that a policy could be useful here, as that'd be a better place to put the `if user->isAdmin() => assertAdmin($actor)` logic, so that we keep the permission layer separate from the editing logic.

src/User/Command/EditUserHandler.php
```diff
@@ -80,6 +80,9 @@ public function handle(EditUser $command)
                     $validate['email'] = $attributes['email'];
                 }
             } else {
+                if ($attributes['email'] !== $user->email && $user->isAdmin()) {
+                    $this->assertAdmin($actor);
+                }
                 $this->assertPermission($canEdit);
```
askvortsov1, line 86: Shouldn't this be changed to reflect the appropriate new permission?

askvortsov1 (changes requested):

src/User/Command/EditUserHandler.php
```diff
@@ -80,6 +80,9 @@ public function handle(EditUser $command)
                     $validate['email'] = $attributes['email'];
                 }
             } else {
+                if ($attributes['email'] !== $user->email && $user->isAdmin()) {
```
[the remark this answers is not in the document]
tankerkiller125: The commit is based on the information from clarks idea https://github.com/flarum/core/issues/1965#issuecomment-584365887 of which he specficially states that "edit credentials" be the one that's admin only. Plus with a permission split edit groups in theory would be reserved for admins only. But I can see your point.

src/User/Command/EditUserHandler.php
```diff
@@ -90,6 +93,9 @@ public function handle(EditUser $command)
         }
 
         if (isset($attributes['password'])) {
+            if ($user->isAdmin()) {
+                $this->assertAdmin($actor);
+            }
             $this->assertPermission($canEdit);
```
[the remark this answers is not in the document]
tankerkiller125: That is not what that does, `$user` is the user that is being edited, `$actor` is the person doing the editing, this check only applies if the `$user` is an admin. For mods, and users it does not check.

src/Api/Serializer/UserSerializer.php
```diff
@@ -36,7 +36,7 @@ protected function getDefaultAttributes($user)
 
         $gate = $this->gate->forUser($this->actor);
 
-        $canEdit = $gate->allows('edit', $user);
+        $canEdit = $gate->allows('edit.credentials') || $gate->allows('edit.groups') || $gate->allows('edit.username') ? true : false;
```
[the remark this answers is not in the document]
tankerkiller125: This will hopefully be removed in the refactor, however the frontend uses the `canEdit` json attribute to determine if the edit option should appear in the dropdown for user editing. Unless you think that splitting that out makes more sense?

src/User/Command/EditUserHandler.php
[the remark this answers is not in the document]
luceos: The reason this should be moved to a policy is to guarantee that any extension that runs a permission check, won't allow editing admins. Are you able to push these into the Policy or would you prefer some help with that?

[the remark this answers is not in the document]
tankerkiller125: @luceos I'll work on doing that, didn't get a chance to do it this weekend, I may however have plenty of questions on discord.

src/Api/Serializer/UserSerializer.php
```diff
@@ -36,7 +36,7 @@ protected function getDefaultAttributes($user)
 
         $gate = $this->gate->forUser($this->actor);
 
-        $canEdit = $gate->allows('edit', $user);
```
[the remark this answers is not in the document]
clarkwinkelmann: I've been wondering about this as well. I'm not sure why we use a `$gate` object and not `->can()` directly. Maybe there was a performance reason behind this ?

In extensions I always use `$this->actor->can()`.

clarkwinkelmann:
I agree with the existing comments. Please ping me again and I'll do an actual review when more code is pushed.

tankerkiller125:
@clarkwinkelmann I've refactored it almost completely

franzliedke:

tankerkiller125:
Rebase was completed

franzliedke:
I'd like to discuss the naming convention for the discussions in our next meeting, that's why I'm not merging this yet.

js/src/forum/components/EditUserModal.js
```diff
@@ -46,16 +46,23 @@ export default class EditUserModal extends Modal {
...
+            placeholder={extractText(app.translator.trans('core.forum.edit_user.username_label'))}
+            bidi={this.username}
+          />
+        </div>,
+        40
+      );
+    }
 
-    if (app.session.user !== this.props.user) {
+    // The first part prevents silent account hijacking
```
clarkwinkelmann, line 64: I don't understand what the account hijacking would be here.

You mean you'd want to prevent an admin from changing their password without sending themselves a reset email ? If so this should be controlled at the Policy level and blocked in the API as well. Just adding this check in the frontend does not prevent the bad actor from directly calling the API endpoint.

But I'd argue that even implementing this in the backend makes little sense. If you hacked into an admin account and want to discretely change the password, you could just as well change the password of another admin account, enter that other account, then change the password of your original hacked account.

Or you could just change their email first, which triggers no notification on the current email, then change password once a new email has been verified.

A better measure against account hijacking could be to send emails on changes. Like "Your email has been changed to <x>" to the old email and "Your password has been changed by an administrator" when an admin changes emails. But that would be another feature that could even be delegated to an extension.

tankerkiller125: This was pre-existing based on the way I read the code I understood this as a way to prevent someone who hijacked an admin/moderator accounts from being able to change that admins/moderators account without the moderator from knowing. I completely agree that doing it in the fronted only isn't that great. We could remove that part all together.

askvortsov1: Clark brings up a good point that there's really no way to prevent hijacking if someone gets into an admin account. I think the strategy of notifying on email change is a good idea, but yeah I would side with removing that check and making sure the proper events are dispatched so that extensions can tune in.

tankerkiller125: For now I've removed the comment, if I remove the check it allows admin and moderators to change their own password and email via the edit interface instead of settings. I'm not sure if that's something we want. (I do understand that via API it's still technically possible)

clarkwinkelmann: I feel like it's even more confusing without a comment, as the check isn't really self-explanatory.

I really think this test should be removed from the frontend.

As for adding it to the backend instead, I don't think it makes sense either, as described in my first post.

tankerkiller125: Removed

askvortsov1:
@franzliedke when you get back from vacation, what's the verdict on permission naming?

clarkwinkelmann (approved):

js/src/forum/components/EditUserModal.js
clarkwinkelmann: Actually re-reading the code I think I know why that check was there in the first place. I don't think it had anything to do with security.

By forcing a user to use the email change feature, this makes sure they set a valid email.

If an administrator manually changes their own email via the edit user modal and put a wrong email by mistake, they might lock themselves out of their account.

I still prefer we get rid of the test (as has been done). If you have that power over other users, I think you can responsibly use it on your own account.

A text that recommends using the email change feature over manually setting an email could be a good idea maybe?

askvortsov1 (approved):
I think Clark brings up a good point here: https://github.com/flarum/core/pull/2113#discussion_r441119334, a warning message would be nice. Other than that, the only thing left is Franz's decision on the ability naming.

js/src/forum/components/EditUserModal.js
tankerkiller125: @clarkwinkelmann Good idea, I'll see if I can get that working tonight maybe after work.

clarkwinkelmann: If I may suggest, we could ship this PR without any special warning or feature and open an issue to keep track of that other possibility.

Actually I'd be very much in favor of changing the UI so that when you edit the email you get two options: either request an email change on the behalf of the user (default) or force-change the email (not default, with warning text)

tankerkiller125: I agree, I'll create a new issue for the proposal and we can ship this PR without it

askvortsov1 (changes requested):
I think we should change it to camel case, and then switch back to dot separated later if we decide that to be the new convention. This also needs to be updated for mithril 2.

One major concern as I thought back to this: we currently have extensions using `$actor->can('edit', $user)`, including our own nicknames. If we remove `edit` without a BC layer, we could run into issues. I think we should re-think this split, keeping `user.edit` as the main user edit permission, and requiring individual permissions (like `user.editGroups`) for higher level things.

tankerkiller125:
Closing in favor of flarum/framework#2620

Issue #2055, opened 2020-03-07
Upgrade Illuminate components to 6.x


franzliedke:
Let's adopt the LTS release. See previous attempt in #1930.

#### Challenges

- [x] Application contract has gathered even more cruft, let's decouple (#2117, #2142, #2155)
- [x] Dispatcher::fire has become Dispatcher::dispatch (f4ab6f4b1f5fb2f254b7fb65fd924ff65c3c1993)
  - impacts extensions, can we add a BC layer?
- [x] Gate contract now has several unnecessary methods, consider alternatives (#2181)
- [ ] Translator contract added more methods, is the contract needed?

franzliedke:
Status update: I'm almost done with getting rid of Laravel's `Application` contract. That will take us to version 5.8.

The `Gate` and `Translator` stuff are changing in L6 if I am not mistaken.

askvortsov1:
I'm likely misunderstanding here, but when I was working on the policy/model visibility scoping refactor, it seemed to me that we're really only using a very small part of Gate. Do we want to keep using it at all? It seems like we could make a gate-inspired system with much less overhead

franzliedke:
That's a good point, but I would rather see that happen with or after the policy rework. No need to block the Laravel upgrade with such a big change to such a central place.

luceos:
Yeah let's keep the changes for the upgrade to a minimum. @franzliedke Gate and Translator indeed where blocking the upgrade, I'd suggest implementing the required changes and exchange them with simpler components if possible before stable.

franzliedke:
Made a little more progress with this. Moving this to Beta.14 now.

A status update:
- For the upgrade to Laravel 5.8, the biggest remaining blocker is avoiding Laravel's huge `Application` contract. I made small steps here with the goal of injecting it in far fewer places.
- I have prepared a commit that splits up the `Application`, most importantly:
  - ignore the `Application` contract, we don't need it
  - stop inheriting from and instead inject a `Container` instance
  - separate paths / URLs from the config derived from `config.php`, from the application bootstrap stuff
  - get rid of some of the fluff, such as deferred service providers and other Laravel-specific things
- I will PR this commit (an improved version of it, really) at the beginning of the B.14 cycle, so that we have enough time to find / encounter issues caused by this change.
- The Gate change (for Laravel 6) might become obsolete with the policy changes in #2056.

luceos:
[...]

PR #2117, opened 2020-04-10
Simplify uploads, avoid Application contract

franzliedke:
**Prepares for #2055**

Laravel 5.8 further increased the size of the `Application` contract. Reimplementing all these methods just becomes pointless. The problem has been recognized over at Laravel, but not yet fully solved. So, instead of reimplementing all these methods, we should instead try to avoid using this contract.

This is one part of reducing the usages - in the various places where we dealt with accepting / deleting file uploads, we injected our `Application` class just to build a path. Instead, let's inject filesystem instances, which are configured to point to a certain directory elsewhere (in service providers). Additional benefit: this makes it easier to inject e.g. a S3 filesystem where scaling capabilities are needed / local filesystems are not writable.

Additional changes:
- No longer use the `MountManager` for these uploads. It only added complexity (by moving temporary files around) and will not be available in the next major release of Flysystem.
- Stop copying files to a temporary location before moving them again.
- For the avatar upload, write custom validation logic instead of wrapping a PSR-7 file object in a Symfony file object just to be compatible with Laravel's validation layer.
- Upgrade `intervention/image` to support passing PSR upload streams. (Very likely, users have already updated to this newer version, as the old constraint allowed it, but we should be explicit for correctness' sake.)

**Confirmed**

### DIFF
--- a/src/Settings/SettingsServiceProvider.php
+++ b/src/Settings/SettingsServiceProvider.php
@@ -9,8 +9,15 @@
 
 namespace Flarum\Settings;
 
+use Flarum\Api\Controller\DeleteFaviconController;
+use Flarum\Api\Controller\DeleteLogoController;
+use Flarum\Api\Controller\UploadFaviconController;
+use Flarum\Api\Controller\UploadLogoController;
 use Flarum\Foundation\AbstractServiceProvider;
+use Illuminate\Contracts\Container\Container;
+use Illuminate\Contracts\Filesystem\Factory;
 use Illuminate\Database\ConnectionInterface;
+use League\Flysystem\FilesystemInterface;
 
 class SettingsServiceProvider extends AbstractServiceProvider
 {
@@ -28,5 +35,18 @@ class SettingsServiceProvider extends AbstractServiceProvider
         });
 
         $this->app->alias(SettingsRepositoryInterface::class, 'flarum.settings');
+
+        $assets = function (Container $app) {
+            return $app->make(Factory::class)->disk('flarum-assets')->getDriver();
+        };
+
+        $this->app->when([
+            DeleteFaviconController::class,
+            DeleteLogoController::class,
+            UploadFaviconController::class,
+            UploadLogoController::class,
+        ])
+            ->needs(FilesystemInterface::class)
+            ->give($assets);
     }
 }
--- a/composer.json
+++ b/composer.json
@@ -56,7 +56,7 @@
         "illuminate/support": "5.7.*",
         "illuminate/validation": "5.7.*",
         "illuminate/view": "5.7.*",
-        "intervention/image": "^2.3.0",
+        "intervention/image": "^2.5.0",
         "laminas/laminas-diactoros": "^1.8.4",
         "laminas/laminas-httphandlerrunner": "^1.0",
         "laminas/laminas-stratigility": "^3.0",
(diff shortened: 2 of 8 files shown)


luceos:
@franzliedke I haven't reviewed the code yet, but I'm worried about:

> Stop copying files to a temporary location before moving them again.

My recommendation is to use the storage folder for storing temporary files. This prevents abuse as soon as we can by preventing unsafe files from ever appearing in a public path. This step is also need to confirm mime types, validate file sizes etc.

franzliedke:
> My recommendation is to use the storage folder for storing temporary files. This prevents abuse as soon as we can by preventing unsafe files from ever appearing in a public path. This step is also need to confirm mime types, validate file sizes etc.

We're operating on files directly wherever PHP stores them. The Intervention library then operates on the streams / in-memory, as far as I can tell.

This topic reminds me of [this fun old fix](https://github.com/flarum/core/commit/cde8dd0dc44d6f94a3627a659cb809a9cecd4022), which was only necessary because we were instantiating a `Local` filesystem to use the `MountManager` (which checked whether the entire `/tmp` dir was readable). Since we're not doing that anymore, we could now resort back to using the system's global temp dir, but as I've tried to explain above, this should not be necessary any longer.

> I've also noticed you added some security validation measures for the avatar, @imorland recently pushed changes into fof/upload that have been inspired by a security colleague, maybe we can re-use that code?

I only reproduced what Laravel's validation layer did for us.

> Aside from that the approach makes sense to reduce use of Application. Is there a way to bind our paths as a Paths class instead of using the Application?

Yes, I was considering that as one of my next steps. (I did not do this here, because I preferred injecting the filesystems, because that means we are less coupled to local filesystem paths, potentially.)

src/Settings/SettingsServiceProvider.php
```diff
@@ -28,5 +35,18 @@ public function register()
         });
 
         $this->app->alias(SettingsRepositoryInterface::class, 'flarum.settings');
+
+        $assets = function (Container $app) {
+            return $app->make(Factory::class)->disk('flarum-assets')->getDriver();
+        };
+
+        $this->app->when([
```
askvortsov1, line 43: I'm not 100% sure how I feel about having to enumerate every possible use of this interface here. Instead, what if we defined an AssetsFilesystemInterface, which extended FilesystemInterface, but had no body. Then, we could bind that to the $assets made above as a singleton, which should enable extensions to inject it, and gain access to assets?

franzliedke: Yeah, I can relate. For now (and especially since this is about the `Application` contract right now) I can live with this. But we can (and should) revisit this when we tackle extensibility / scaling issues for the filesystem.

franzliedke: Maybe this is actually a legit use-case for one of Illuminate's manager classes. /cc @luceos 

Probably not 100% the same use-cases, but at least we want to manage multiple "disks".

askvortsov1: I'm good with implementing this as a temporary solution, and refactoring later to allow file system driver extendability :+1:

franzliedke:
> Should we do something similar for the storage folder (which we might want to rename tbh). It's less urgent as it's not public, but as far as I know, that's the only other filesystem structure we might want to abstracify.

Yes, similar changes will follow for other places using `Application` or one of the helpers relying on it.

src/Settings/SettingsServiceProvider.php
luceos: Agreeing with Sasha and yourself. Keep changes to a minimum and iterate to fulfill our future needs. This to me is okay.

PR #2123, opened 2020-04-15
CDN acceleration

Littlegolden:
Use jsDelivr to increase the loading speed of `jquery.min.js` in China.
The original file from `jquery.com` can't be loaded in China.

**Confirmed**

### DIFF
--- a/views/install/install.php
+++ b/views/install/install.php
@@ -66,7 +66,7 @@
   </div>
 </form>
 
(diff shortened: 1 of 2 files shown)


askvortsov1:
Thank you for the PR! However, considering that we're going to switch away from JQuery within this release, I'm going to close this. We'll try to take geographical considerations into account when loading whatever library we replace jquery with.

PR #2125, opened 2020-04-15
[Frontend Rewrite] Add select component

askvortsov1:
Add select component.

I want to start my contributions to the frontend rewrite incrementally, as I'm not super familiar with the system. Does this look about right?


js/src/common/components/Select.tsx
```diff
@@ -0,0 +1,64 @@
+import Component, { ComponentProps } from "../Component";
+import icon from "../helpers/icon";
+
+export interface SelectProps extends ComponentProps {
+    /**u
```
datitisev, line 5: `u` 👀 

```suggestion
    /**
```

js/src/common/components/Select.tsx
```diff
@@ -0,0 +1,64 @@
...
+    /**
+     * A callback to run when the selected value is changed.
+     */
+    onchange?: Function;
+
+    /**
+     * An object of options
+     */
+    options: {
+        [index: string]: string;
```
datitisev, line 19: Let's use `key` instead of `index` - index is targetted more for arrays, and the rest of instances where we typehint objects like this use `key`.

js/src/common/components/Select.tsx
```diff
@@ -0,0 +1,64 @@
...
+}
+
+/**
+ * The `Select` component displays a <select> input, surrounded with some extra
+ * elements for styling. It accepts the following props:
+ *
+ * - `options` A map of option values to labels.
+ * - `onchange` A callback to run when the selected value is changed.
+ * - `value` The value of the selected option.
+ * - `disabled` Disabled state for the input.
```
datitisev, line 35: Let's move these descriptions to the prop descriptions above in `SelectProps`. You already have comment documentation there, just replace them with this 🙁

js/src/common/components/Select.tsx
askvortsov1: no u

js/src/common/components/Select.tsx
askvortsov1: yeah good idea

js/src/common/components/Select.tsx
askvortsov1: Done, good point

askvortsov1:
Done, thanks for the review!

datitisev:
Is the addition of the `onchange` method simply to avoid ` onchange={onchange ? m.withAttr('value', onchange.bind(this)) : undefined}`?

askvortsov1:
I've been trying to increase code style consistency a bit, for the on change stuff, that's how it's done in Checkbox. Not sure if we want to separate a process to standardize stuff a bit independently of the rewrite

datitisev:
Looks fine to me, was just wondering about the change.

franzliedke:
Please squash-merge these, to avoid making rebases / merges of the rewrite branch even more complicated. :pray:

datitisev:
@franzliedke Was planning on squashing 👍

PR #2129, opened 2020-04-19
Hide permission groups

### LINKED ISSUE #845: Add a way to hide certain groups

askvortsov1:
Fixes #845. Replaces #1967.

> - added permission
> - changed serializer
> - created column
> - modified policy


franzliedke:
Is this simply a re-do of #1967, completely replacing the other one?

askvortsov1:
Just a few tweaks on top of it to get the formatting right

js/src/admin/components/PermissionGrid.js
```diff
@@ -112,6 +112,16 @@ export default class PermissionGrid extends Component {
       100
     );
 
+    items.add(
+      'viewHiddenGroups',
+      {
+        icon: 'fas fa-users',
+        label: app.translator.trans('core.admin.permissions.view_hidden_groups_list_label'),
```
franzliedke, line 119:
```suggestion
        label: app.translator.trans('core.admin.permissions.view_hidden_groups_label'),
```

migrations/2020_01_14_130500_change_permission_groups_add_is_hidden.php
```diff
@@ -0,0 +1,14 @@
...
+ * This file is part of Flarum.
+ *
+ * For detailed copyright and license information, please view the
+ * LICENSE file that was distributed with this source code.
+ */
+
+use Flarum\Database\Migration;
+
+return Migration::addColumns('groups', [
+    'is_hidden' => ['boolean', 'default' => 0]
```
franzliedke, line 13:
```suggestion
    'is_hidden' => ['boolean', 'default' => false]
```

At least that's how we do it elsewhere. :wink:

migrations/2020_01_14_130500_change_permission_groups_add_is_hidden.php
```diff
@@ -0,0 +1,14 @@
+<?php
```
franzliedke, line 1: Please change the filename to have a recent number. (Not 100% sure how well we deal with adding migrations with timestamps older than the currently highest one.) Not super relevant for normal beta upgrades, but possibly for developers / bleeding edge platforms.

tests/integration/api/groups/ListTest.php
```diff
@@ -26,6 +48,35 @@ public function shows_index_for_guest()
         $this->assertEquals(200, $response->getStatusCode());
         $data = json_decode($response->getBody()->getContents(), true);
 
+        $this->assertEquals(Group::where('is_hidden', 0)->count(), count($data['data']));
```
franzliedke, line 51: Could we improve this (and the other test) to check *which* groups are returned? E.g. map the JSON to an array of IDs and match the one here to `[1]` and the other one to `[1, 10]` (using `assertEqualsCanonicalizing`, see [this StackOverflow answer](https://stackoverflow.com/a/28189403))?

src/Api/Serializer/BasicUserSerializer.php
```diff
@@ -45,6 +45,10 @@ protected function getDefaultAttributes($user)
      */
     protected function groups($user)
     {
-        return $this->hasMany($user, GroupSerializer::class);
+        if ($this->getActor()->can('viewHiddenGroups')) {
+            return $this->hasMany($user, GroupSerializer::class);
+        }
+
+        return $this->hasMany($user, GroupSerializer::class, 'visibleGroups');
```
franzliedke, line 52: I haven't seen this approach in other serializers. Do we have a way to use `whereVisibleTo()` here as well? :thinking:

askvortsov1: I don't think so: we're returning a tobscure Relationship object, not a query.

franzliedke:
I squash-merged this into 6e8884f190395b5a06fb8ebee4f8bb4e3cb8c467, but did so right at the beginning of today's GitHub outage, so apparently the PR did not get closed. We're done here! :clap: :clinking_glasses:

PR #2132, opened 2020-04-24
Model extender inheritance fix

askvortsov1:
Support influencing subclasses via model extender.

### DIFF
--- a/tests/integration/extenders/ModelTest.php
+++ b/tests/integration/extenders/ModelTest.php
@@ -10,7 +10,6 @@
 namespace Flarum\Tests\integration\extenders;
 
 use Carbon\Carbon;
-use Flarum\Database\AbstractModel;
 use Flarum\Discussion\Discussion;
 use Flarum\Extend;
 use Flarum\Group\Group;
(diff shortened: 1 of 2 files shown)


tests/integration/extenders/ModelTest.php
```diff
@@ -167,6 +168,66 @@ public function custom_relationship_is_inherited_to_child_classes()
...
     }
 
+    /**
+     * @test
+     */
+    public function custom_relationship_prioritizes_child_classes_within_2_parent_classes()
+    {
+        $this->extend(
+            (new Extend\Model(AbstractModel::class))
+                ->belongsTo('ancestor', User::class, 'user_id'),
```
franzliedke, line 178: What is the idea behind this scenario? I see no use-case for extending this abstract base class. And as long as no use-case pops up, I'd rather not define how we want to (and then have to) behave.

The `Post` / `CommentPost` scenario you added below makes sense. :+1:

tests/integration/extenders/ModelTest.php
```diff
@@ -261,6 +322,31 @@ public function custom_default_attribute_is_inherited_to_child_classes()
...
     }
 
+    /**
+     * @test
+     */
+    public function custom_default_attribute_inheritance_prioritizes_child_class()
+    {
+        $this->extend(
+            (new Extend\Model(AbstractModel::class))
+                ->default('answer', 'dont do this'),
```
franzliedke, line 332: Same here. This scenario is useful with the `Post` / `CommentPost` combination, but IMO we don't need the `AbstractModel` example.

askvortsov1: Clarified the use case

PR #2134, opened 2020-04-24
View Extender (add namespace)

askvortsov1:
**Fixes part of #1891**

Add view extender with addNamespace method. I called it this instead of just ->namespace() in case we want to support prependNamespace, replaceNamespace, etc

Do we want to include replaceNamespace here? I think that might be useful. EDIT: Not yet. Flarum's namespaces are loaded after extenders do their thing, so it wouldn't work there, hence, we should hold off on that for now.

**Confirmed**


franzliedke:
What do you think about naming the class `ViewNamespace` and the method `add()`? I just went through all existing extenders and this would be more in line with them. All of them have very concise method names, and `add()` is only used when other operations like `remove()` exist as well.

Hmm... :thinking: ... do we really need other operations? We barely use the concept of namespaces in core, so there's nothing there to replace. And I don't want to suggest removing / replacing other extensions' namespaces is a good thing (or important enough to deal with in one of the bundled extenders).

tests/integration/extenders/ViewTest.php
```diff
@@ -0,0 +1,39 @@
...
+
+class ViewTest extends TestCase
+{
+    /**
+     * @test
+     */
+    public function custom_view_namespace_does_not_exist_by_default()
+    {
+        $this->expectException(\InvalidArgumentException::class);
+        $this->assertEquals('<html><body>Hello World!</body></html>', trim($this->app()->getContainer()->make(Factory::class)->make('integration.test::test')->render()));
```
franzliedke, line 24: I assume you're going for symmetry with the test case below here, but it confuses me that there is an assertion after the `expectException()` call. That primarily signals to me that we expect this assertion to hold true.

Which call here is going to throw the `InvalidArgumentException`? Is it `Factory::make()`? In that case, let's only call this method, please. Or is it the `assertEquals()`? (I would assume that throws a custom exception class.)

If you want more symmetry, you could do the `$this->app()...->render()` on one line in both tests, and in the second one store the result in a variable and then `assertEquals()` on that in an extra line.

tests/integration/resources/views/test.blade.php
```diff
@@ -0,0 +1 @@
+<html><body>Hello World!</body></html>
```
franzliedke, line 1: Could we move this to `tests/fixtures`? :grin:

src/Extend/View.php
```diff
@@ -0,0 +1,43 @@
...
+namespace Flarum\Extend;
+
+use Flarum\Extension\Extension;
+use Illuminate\Contracts\Container\Container;
+use Illuminate\Contracts\View\Factory;
+
+class View implements ExtenderInterface
+{
+    private $addNamespaces = [];
+    private $replaceNamespaces = [];
```
franzliedke, line 19: Seems like a leftover.

src/Extend/View.php
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * Register a new namespace of laravel views.
```
franzliedke, line 22: Let's use this opportunity to add more explanation about what a **L**aravel namespace is, and how the namespace can then be referenced in other views or controllers. :wink:

src/Extend/View.php
```diff
@@ -0,0 +1,43 @@
...
+     *
+     * @param  string  $namespace
+     * @param  string|array  $hints
```
franzliedke, line 25: What does "hint" mean in this context? Isn't this a path (or multiple)?

askvortsov1: I just copied this from the laravel docblocks

src/Extend/ViewNamespace.php
```diff
@@ -0,0 +1,51 @@
...
+    public function add($namespace, $hints)
+    {
+        $this->adds[$namespace] = $hints;
+
+        return $this;
+    }
+
+    public function extend(Container $container, Extension $extension = null)
+    {
+        $factory = $container->make(Factory::class);
```
franzliedke, line 45: Please `extend` the view factory - no need to resolve it on every request.

franzliedke: Ah, screw it. I can take care of it.

franzliedke:
I took care of the lazy extending (requested in https://github.com/flarum/core/pull/2134#discussion_r456342233) and the renaming (discussed in https://github.com/flarum/core/pull/2134#issuecomment-656621297) in master. 

Thanks for the PR! :raised_hands:

PR #2139, opened 2020-04-29
common: button typings improvements

askvortsov1:
Button typing improvements, and exporting LinkButtonProps.

### DIFF
--- a/js/src/common/components/Button.tsx
+++ b/js/src/common/components/Button.tsx
@@ -5,46 +5,64 @@ import extractText from '../utils/extractText';
 import LoadingIndicator from './LoadingIndicator';
 
 export interface ButtonProps extends ComponentProps {
+    /**
+     * A tooltip for the button.
+     */
     title?: string;
+
+    /**
+     * An html type attribute for the button.
+     */
     type?: string;
+
+    /**
+     * The name of the icon class. If specified, the button will be given a
+     * 'has-icon' class name.
+     */
     icon?: string;
 
+    /**
+     * Whether or not the button should be in a disabled loading state.
+     */
     loading?: boolean;
+
+    /**
+     * Whether or not the button is disabled. If truthy, the button
+     * will be given a 'disabled' class name, and any `onclick` handler will be
+     * removed.
+     */
     disabled?: boolean;
+
+    /**
+     * A callback to run when the button is clicked.
+     */
     onclick?: Function;
 }
 
 /**
  * The `Button` component defines an element which, when clicked, performs an
- * action. The button may have the following special props:
- *
- * - `icon` The name of the icon class. If specified, the button will be given a
- *   'has-icon' class name.
- * - `disabled` Whether or not the button is disabled. If truthy, the button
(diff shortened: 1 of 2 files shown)


js/src/common/components/Button.tsx
```diff
@@ -5,46 +5,64 @@ import extractText from '../utils/extractText';
...
- * - `loading` Whether or not the button should be in a disabled loading state.
- *
- * All other props will be assigned as attributes on the button element.
+ * action.
  *
  * Note that a Button has no default class names. This is because a Button can
  * be used to represent any generic clickable control, like a menu item.
  */
-export default class Button<T extends ButtonProps = ButtonProps> extends Component<T> {
+export default class Button<T extends ButtonProps = ButtonProps> extends Component<ButtonProps> {
```
datitisev, line 49: What's the reason why this was change? Afaik, classes that extend `Button` and pass their own props won't have those props typings available in `this.props` - right now it looks like `T` is pointless.

js/src/common/components/Button.tsx
```diff
@@ -5,46 +5,64 @@ import extractText from '../utils/extractText';
...
     view() {
-        const { children, ...attrs } = this.props;
+        const attrs: ButtonProps = { ...this.props };
```
datitisev, line 51: This *could* use `T`, not necessary though because you are only using it in `view`, but perhaps for consistency? Doesn't matter to me, just some thoughts.

js/src/common/components/LinkButton.tsx
```diff
@@ -1,24 +1,24 @@
...
- *
- * All of the props accepted by `Button`, plus:
- *
- * - `active` Whether or not the page that this button links to is currently
- *   active.
- * - `href` The URL to link to. If the current URL `m.route()` matches this,
- *   the `active` prop will automatically be set to true.
  */
-export default class LinkButton extends Button<LinkButtonProps> {
+export default class LinkButton<T extends LinkButtonProps> extends Button<LinkButtonProps> {
```
datitisev, line 21: If you do this you need to set a default and pass `T` to `Button` instead of `LinkButtonProps`. Read my first comment.

js/src/common/components/Button.tsx
askvortsov1: So are you saying we keep this as T? I'm somewhat new to generics, so apologies if I'm coming off really stupid

askvortsov1:
I have (hopefully) fixed the weird generics stuff

datitisev:
@askvortsov1 Lemme try to make you understand the generic class stuff.

If you are passing `ButtonProps` to button instead of `T`, it doesn't matter what class is passed to `LinkButton<?>` as the one that goes to `this.props` is the one passed to the Component class. By passing `ButtonProps` you are doing `ButtonProps -> Button -> Component` instead of `<T extends ButtonProps> -> Button -> Component`, removing the point of having `T` in `LinkButton` because it does anything there.

Now that I've typed this, I can tell I didn't make much sense either, but hopefully this helps.

askvortsov1:
Thank you, that actually cleared things up a lot for me!

PR #2141, opened 2020-04-30
[Frontend Rewrite]: All admin pages

askvortsov1:
I've set up all the admin pages to use TS/Mithril v2


datitisev:
Issues

- Figure out that annoying routing issue
	>  Any route after the hashbang (e.g. /admin#!/fdksjafkldjflksajfdla) gets changed to /admin#!/ if it doesn't correspond to a valid route, but anything that doesn't include the hashbang (/admin#, /admin, /admin#!, etc) doesn't get its URL changed. The page content gets shown properly which makes me think that m.route is partially working correctly, but the URL doesn't get changed
- Fix up the typing for AdminLinkButton once the button typing improvements PR gets merged (lmk if there's anything else that needs to be done there)

datitisev:
`/admin#!` does get its url change - the only issue here is that `/admin` doesn't get the `#`.

Not sure why your examples have `#!/` - none of my URLs have that...?

Currently, in Flarum beta 12, any unknown route redirects to `/admin#` so that's fine.

Tested it out, looks to work great. Only thing is that "Please wait" modals can be dismissed - but that's not part of your code so don't fix it here.

askvortsov1:
To clarify that a bit, I was initially testing with '#!' as the prefix, but later changed that to '#', to be more consistent with what we have now. The issue above still applies, but just pretend the '!' isn't there.

PR #2142, opened 2020-05-01
Split up Application and Container

franzliedke:
**Refs #2055**

- Stop trying to implement Laravel's Application contract, which
  has no value for us.
- Stop inheriting from the Container, injecting one works equally
  well and does not clutter up the interfaces.
- Create a new Paths object that encapsulates a little logic (may
  attract more behavior later) and which can be injected instead
  of the entire Application class when collaborators only want to
  get access to the paths.
- Inject the Paths collection instead of unwrapping it again, for
  better encapsulation.

This brings us one step closer toward upgrading our Laravel
components (#2055), because we no longer need to adopt the changes
to the Application contract.

- Any places I may have missed?

**Confirmed**

### DIFF
--- a/src/Admin/AdminServiceProvider.php
+++ b/src/Admin/AdminServiceProvider.php
@@ -12,7 +12,6 @@ namespace Flarum\Admin;
 use Flarum\Extension\Event\Disabled;
 use Flarum\Extension\Event\Enabled;
 use Flarum\Foundation\AbstractServiceProvider;
-use Flarum\Foundation\Application;
 use Flarum\Foundation\ErrorHandling\Registry;
 use Flarum\Foundation\ErrorHandling\Reporter;
 use Flarum\Foundation\ErrorHandling\ViewFormatter;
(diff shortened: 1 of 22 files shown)


askvortsov1:
Largely looks good! Do we want to depreciate the path methods on Application? Or are we keeping them to support the PHP helpers

franzliedke:
> Do we want to depreciate the path methods on Application? Or are we keeping them to support the PHP helpers

@askvortsov1 I want to deprecate them for beta.14. And the helpers as well, if I don't hit any obstacles.

But before I can do so, I will need to do a bunch of cleanup in core and extensions, wherever paths are retrieved. I also want to try extracting a class for the "config", similar to what I did for the paths.

But that's all not particularly urgent right now. I just wanted to remove this roadblock for the Laravel upgrade.

askvortsov1:
Sounds like a good plan! Another good step (and great to see more removal of unused code!)

PR #2145, opened 2020-05-05
fix search input value disappearing on page reload

w-4:
**Fixes the following bug:**
going to https://discuss.flarum.org/?q=abc
the search input shows no value if accessed through the link, or when searching something and reloading the page.

remove the empty string in m.prop, so the below code in the view function can run.
```js
   // Initialize search input value in the view rather than the constructor so
    // that we have access to app.current.
    if (typeof this.value() === 'undefined') {
      this.value(currentSearch || '');
    }
```

**Confirmed**

### DIFF
--- a/js/src/forum/components/Search.js
+++ b/js/src/forum/components/Search.js
@@ -24,7 +24,7 @@ export default class Search extends Component {
      *
      * @type {Function}
      */
-    this.value = m.prop('');
+    this.value = m.prop();
 
     /**
      * Whether or not the search input has focus.


askvortsov1:
Thank you for the PR! Looking back at the file history, it looks like the value was not set to undefined intentionally (https://github.com/flarum/core/commit/aaab2cc86e49825531831a3abca7487f5751f524). That being said, this should also be resolved as a side effect of the changes in #2151, so I'm going to close this.

franzliedke:
(I added this to the milestone so that we don't forget about this bug being fixed in the changelog.)

Issue #2144, opened 2020-05-05
Frontend Needs Proper State Management


askvortsov1:
## Architecture

**Is your feature request related to a problem? Please describe.**
Right now, our frontend uses somewhat of a haphazard state management / propogation system. We have:

- Local state (is a page loading or not)
- Global state (all the discussions that have been retrieved from the database

This means that the displayed UI can be modified by:

1. Modifying parts of the component instances directly (which is bad)
2. Modifying the components' local states (which is slightly better, but still bad)
3. Modifying the global state, and having changes propogate during redraw (this is considered the best and is most consistent with unidirectional data flow).

**Describe the solution you'd like**
This should probably be done gradually, one step at a time. Radical change is probably not needed, however, the following steps should be individually taken:

1. Components should NOT be modified by externally calling methods on an instance of said virtual component. This means getting rid of app.current, app.modal.component, and any other similar systems. Instead, modifications should be done to the global state, and components should be redrawn as appropriate after said state modifications.
2. Adopt a central state management paradigm. Right now, we have a mutable global state, which isn't ideal. We should try to gradually convert to something more robust.  [Meiosis](https://meiosis.js.org/) looks like it could be a simple and elegant solution, although there are plenty of options (Redux and Flux being the big ones)
2. Starting with the most extendable areas, try to convert miscategorized local state to global state when appropriate. 

In my opinion, (1) should be included in #1821. (2) should be aimed for before stable. (3) can be done after stable

**Benefits**
If done properly, this should:

- Make the frontend more extendable
- Increase robustness, as things will be changing only when they should be changing
- Make code easier to update/change without worrying about breaking everything
- Respect better practices
- Make systems more testable

askvortsov1:
## Part 1

In a sequence of PRs, I'm going to convert:

app.current
app.search
app.modal
app.alerts
app.cache.discussionList

away from components. That is, the values stored in app.* will all be states and methods relating to changing states, not component instances. Those changes can be reviewed individually, and will make the frontend rewrite easier actually, as the mechanism for changing state should still work with Mithril v2. Considering that the rewrite would depend on this, I will make this a top priority, and try to get the PRs out within the next 2 days. Considering a general lack of inderdependence, I'm hoping that each of these can be its own PR, which should make reviewing easier.

Just to clarify again why this is necessary: aside from promoting good practice, we're either going to need to go this route, or use hacky workarounds in the rewrite which will come back to bite us later.

clarkwinkelmann:
[...]

PR #2150, opened 2020-05-08
Move Discussion List State into its own class

askvortsov1:
**Fixes Part Of #2144**

- Separate the state of the discussion list out into it's own class (copied over David's code from the mithril/ts rewrite).
- Always keep an instance of discussion list state active; refresh instead of setting it null

Anything I missed?

**Confirmed**

EDIT THIS BEFORE MERGING: https://github.com/flarum/core/pull/2151#pullrequestreview-417226123

### DIFF
--- a/js/src/forum/ForumApplication.js
+++ b/js/src/forum/ForumApplication.js
@@ -77,8 +77,8 @@ export default class ForumApplication extends Application {
    * @inheritdoc
    */
   cache = {
-    discussionList: new DiscussionListState()
-  }
+    discussionList: new DiscussionListState(),
+  };
 
   constructor() {
     super();
--- a/js/src/forum/ForumApplication.js
+++ b/js/src/forum/ForumApplication.js
@@ -78,10 +78,12 @@ export default class ForumApplication extends Application {
 
     routes(this);
 
-    this.discussionList = new DiscussionListState({ forumApp: this });
+    this.discussions = new DiscussionListState({ forumApp: this });
 
-    // Deprecated, remove in beta 15. BC layer.
-    this.cache.discussionList = this.discussionList;
+    /**
+     * @deprecated beta 14, remove in beta 15.
+     */
+    this.cache.discussionList = this.discussions;
   }
 
   /**
--- a/js/src/forum/ForumApplication.js
+++ b/js/src/forum/ForumApplication.js
@@ -73,11 +73,6 @@ export default class ForumApplication extends Application {
    */
   search = new GlobalSearchState();
 
-  /*
-   * @inheritdoc
-   */
-  discussionList;
-
   constructor() {
     super();
 
--- a/js/src/forum/ForumApplication.js
+++ b/js/src/forum/ForumApplication.js
@@ -76,14 +76,15 @@ export default class ForumApplication extends Application {
   /*
    * @inheritdoc
    */
-  cache = {
-    discussionList: new DiscussionListState(),
-  };
+  discussionList = new DiscussionListState();
 
   constructor() {
     super();
 
     routes(this);
+
+    // Deprecated, remove in beta 15. BC layer.
+    this.cache.discussionList = this.discussionList;
   }
 
   /**
--- a/js/src/forum/ForumApplication.js
+++ b/js/src/forum/ForumApplication.js
@@ -73,6 +73,13 @@ export default class ForumApplication extends Application {
    */
   search = new GlobalSearchState();
 
+  /*
+   * @inheritdoc
+   */
+  cache = {
+    discussionList: new DiscussionListState()
+  }
+
   constructor() {
     super();
 
@@ -104,7 +111,6 @@ export default class ForumApplication extends Application {
 
     this.pane = new Pane(document.getElementById('app'));
     this.composer = m.mount(document.getElementById('composer'), Composer.component());
-    this.cache.discussionList = new DiscussionListState();
 
     m.route.mode = 'pathname';
     super.mount(this.forum.attribute('basePath'));
--- a/js/src/forum/ForumApplication.js
+++ b/js/src/forum/ForumApplication.js
@@ -78,6 +78,12 @@ export default class ForumApplication extends Application {
 
     routes(this);
 
+    /**
+     * An object which controls the state of the cached discussion list, which
+     * is used in the index page and the slideout pane.
+     *
+     * @type {DiscussionListState}
+     */
     this.discussions = new DiscussionListState({ forumApp: this });
 
     /**
--- a/js/src/forum/ForumApplication.js
+++ b/js/src/forum/ForumApplication.js
@@ -15,6 +15,7 @@ import Application from '../common/Application';
 import Navigation from '../common/components/Navigation';
 import NotificationListState from './states/NotificationListState';
 import GlobalSearchState from './states/GlobalSearchState';
+import DiscussionListState from './state/DiscussionListState';
 
 export default class ForumApplication extends Application {
   /**
@@ -103,6 +104,8 @@ export default class ForumApplication extends Application {
 
     this.pane = new Pane(document.getElementById('app'));
     this.composer = m.mount(document.getElementById('composer'), Composer.component());
+    this.cache.discussionList = new DiscussionListState();
+    this.cache.discussionList.refresh();
 
     m.route.mode = 'pathname';
     super.mount(this.forum.attribute('basePath'));
--- a/js/src/forum/ForumApplication.js
+++ b/js/src/forum/ForumApplication.js
@@ -105,7 +105,6 @@ export default class ForumApplication extends Application {
     this.pane = new Pane(document.getElementById('app'));
     this.composer = m.mount(document.getElementById('composer'), Composer.component());
     this.cache.discussionList = new DiscussionListState();
-    this.cache.discussionList.refresh();
 
     m.route.mode = 'pathname';
     super.mount(this.forum.attribute('basePath'));
--- a/js/src/forum/ForumApplication.js
+++ b/js/src/forum/ForumApplication.js
@@ -76,12 +76,14 @@ export default class ForumApplication extends Application {
   /*
    * @inheritdoc
    */
-  discussionList = new DiscussionListState();
+  discussionList;
 
   constructor() {
     super();
 
     routes(this);
+
+    this.discussionList = new DiscussionListState({ forumApp: this });
 
     // Deprecated, remove in beta 15. BC layer.
     this.cache.discussionList = this.discussionList;
--- a/js/src/forum/components/DiscussionList.js
+++ b/js/src/forum/components/DiscussionList.js
@@ -34,7 +34,7 @@ export default class DiscussionList extends Component {
       });
     }
 
-    if (state.hasDiscussions() && !state.isLoading()) {
+    if (!state.hasDiscussions() && !state.isLoading()) {
       const text = app.translator.trans('core.forum.discussion_list.empty_text');
       return <div className="DiscussionList">{Placeholder.component({ text })}</div>;
     }
--- a/js/src/forum/components/DiscussionList.js
+++ b/js/src/forum/components/DiscussionList.js
@@ -40,7 +40,7 @@ export default class DiscussionList extends Component {
     }
 
     return (
-      <div className={'DiscussionList' + (state.params.q ? ' DiscussionList--searchResults' : '')}>
+      <div className={'DiscussionList' + (state.isSearchResults() ? ' DiscussionList--searchResults' : '')}>
         <ul className="DiscussionList-discussions">
           {state.discussions.map((discussion) => {
             return (
--- a/js/src/forum/components/DiscussionList.js
+++ b/js/src/forum/components/DiscussionList.js
@@ -11,10 +11,11 @@ import Placeholder from '../../common/components/Placeholder';
  *
  * - `params` A map of parameters used to construct a refined parameter object
  *   to send along in the API request to get discussion results.
+ * - `state` A DiscussionListState object that represents the discussion lists's state.
  */
 export default class DiscussionList extends Component {
   init() {
-    this.state = this.props.state || new DiscussionListState();
+    this.state = this.props.state;
   }
 
   view() {
--- a/js/src/forum/components/DiscussionList.js
+++ b/js/src/forum/components/DiscussionList.js
@@ -34,7 +34,7 @@ export default class DiscussionList extends Component {
       });
     }
 
-    if (!state.hasDiscussions() && !state.isLoading()) {
+    if (state.empty()) {
       const text = app.translator.trans('core.forum.discussion_list.empty_text');
       return <div className="DiscussionList">{Placeholder.component({ text })}</div>;
     }
--- a/js/src/forum/components/DiscussionList.js
+++ b/js/src/forum/components/DiscussionList.js
@@ -21,10 +21,10 @@ export default class DiscussionList extends Component {
   view() {
     const state = this.state;
 
-    const params = state.params;
+    const params = state.getParams();
     let loading;
 
-    if (state.loading) {
+    if (state.isLoading()) {
       loading = LoadingIndicator.component();
     } else if (state.moreResults) {
       loading = Button.component({
@@ -34,7 +34,7 @@ export default class DiscussionList extends Component {
       });
     }
 
-    if (state.discussions.length === 0 && !state.loading) {
+    if (state.hasDiscussions() && !state.isLoading()) {
       const text = app.translator.trans('core.forum.discussion_list.empty_text');
       return <div className="DiscussionList">{Placeholder.component({ text })}</div>;
     }
--- a/js/src/forum/components/DiscussionList.js
+++ b/js/src/forum/components/DiscussionList.js
@@ -14,53 +14,34 @@ import Placeholder from '../../common/components/Placeholder';
  */
 export default class DiscussionList extends Component {
   init() {
-    /**
-     * Whether or not discussion results are loading.
-     *
-     * @type {Boolean}
-     */
-    this.loading = true;
-
-    /**
-     * Whether or not there are more results that can be loaded.
-     *
-     * @type {Boolean}
-     */
-    this.moreResults = false;
-
-    /**
-     * The discussions in the discussion list.
-     *
-     * @type {Discussion[]}
-     */
-    this.discussions = [];
-
-    this.refresh();
+    this.state = this.props.state || new DiscussionListState();
   }
 
   view() {
-    const params = this.props.params;
+    const state = this.state;
+
+    const params = state.params;
     let loading;
 
-    if (this.loading) {
+    if (state.loading) {
       loading = LoadingIndicator.component();
-    } else if (this.moreResults) {
+    } else if (state.moreResults) {
       loading = Button.component({
         children: app.translator.trans('core.forum.discussion_list.load_more_button'),
         className: 'Button',
-        onclick: this.loadMore.bind(this),
+        onclick: state.loadMore.bind(state),
       });
     }
 
-    if (this.discussions.length === 0 && !this.loading) {
+    if (state.discussions.length === 0 && !state.loading) {
       const text = app.translator.trans('core.forum.discussion_list.empty_text');
       return <div className="DiscussionList">{Placeholder.component({ text })}</div>;
     }
 
     return (
-      <div className={'DiscussionList' + (this.props.params.q ? ' DiscussionList--searchResults' : '')}>
+      <div className={'DiscussionList' + (state.params.q ? ' DiscussionList--searchResults' : '')}>
         <ul className="DiscussionList-discussions">
-          {this.discussions.map((discussion) => {
+          {state.discussions.map((discussion) => {
             return (
               <li key={discussion.id()} data-id={discussion.id()}>
                 {DiscussionListItem.component({ discussion, params })}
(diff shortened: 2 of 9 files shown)


js/src/forum/ForumApplication.js
```diff
@@ -94,6 +95,8 @@ export default class ForumApplication extends Application {
 
     this.pane = new Pane(document.getElementById('app'));
     this.composer = m.mount(document.getElementById('composer'), Composer.component());
+    this.cache.discussionList = new DiscussionListState();
+    this.cache.discussionList.refresh();
```
datitisev, line 99: You are calling .refresh() twice here - once in the constructor, and the second time here.

You should *not* be refreshing the discussion list in the mounting anyway. That's a pointless API call if they are not anywhere with the discussion list visible.

askvortsov1: Why pointless: wouldn't it just preload all the discussions? What would be a better place for it?

I suppose we could always take out the refresh method from the constructor if we want to preserve the state instance throughout the lifetime of app?

datitisev: Because that's an API call that will occur on every page load. If you are not in the IndexPage, and are not planning to go there, it's an extra call that shouldn't be occurring.

Also - it doesn't only affect the IndexPage. It'd be preloading the default discussion list, and if the current page is a tag discussion list, the response of that API call would be discarded (as it doesn't have the query params and all).

We shouldn't be making API calls when the resources requested aren't needed by the user.

askvortsov1: true, fixed.

js/src/forum/components/IndexPage.js
```diff
@@ -33,26 +33,27 @@ export default class IndexPage extends Page {
...
+        if (app.cache.discussionList.params[key] !== params[key]) {
+          app.cache.discussionList.clear();
           return true;
         }
       });
     }
 
-    if (!app.cache.discussionList) {
-      app.cache.discussionList = new DiscussionList({ params });
+    if (!app.cache.discussionList.hasDiscussions()) {
```
datitisev, line 54: Problem here - if the forum has no discussions, it will keep refreshing it every time the init method runs. That could potentially be an issue somewhere else ?

askvortsov1: I see this as being more of a positive though: every time someone goes to the index page, it checks if new discussions have been added

datitisev: This is a functionality change though. Not sure if we want that .

askvortsov1: I don't see any negatives to it. We'd need a bit of a hackier system to restore previous functionality, but that doesn't really bring any benefits: very rarely will forums have no discussions at all, and in the small case that they do, making it easier to see that new discussions have been added is a positive

js/src/forum/ForumApplication.js
```diff
@@ -94,6 +95,7 @@ export default class ForumApplication extends Application {
 
     this.pane = new Pane(document.getElementById('app'));
     this.composer = m.mount(document.getElementById('composer'), Composer.component());
+    this.cache.discussionList = new DiscussionListState();
```
datitisev, line 98: Perhaps this should go in a separate method? I don't feel like it belongs in `mount()`, that's for mounting stuff to the DOM.

Also - not entirely sure it belongs in the Application class at all; not sure where else it would go though.

askvortsov1: Good point, since refresh is no longer in the constructor, I moved it out of the method.

Unless we want to add another global variable, I'd say we can keep all this in app for now, and then change it around when we get to Part 2 of the state refactor (which would be after the TS/Mithril 2 upgrades)

js/src/forum/components/IndexPage.js
askvortsov1: If there's no significant objections, can we resolve this for now?

datitisev: Sure, if you've tested it extensively.

js/src/forum/ForumApplication.js
```diff
@@ -63,6 +64,13 @@ export default class ForumApplication extends Application {
    */
   history = new History();
 
+  /**
+   * @inheritdoc
+   */
+  cache = {
+    discussionList: new DiscussionListState(),
+  };
```
franzliedke, line 72: Sorry if this was discussed elsewhere, I did not find anything about it in #2144... Why is this nested under the `cache`, unlike the various other new state objects?

js/src/forum/components/DiscussionList.js
```diff
@@ -14,53 +14,34 @@ import Placeholder from '../../common/components/Placeholder';
...
-
-    /**
-     * The discussions in the discussion list.
-     *
-     * @type {Discussion[]}
-     */
-    this.discussions = [];
-
-    this.refresh();
+    this.state = this.props.state || new DiscussionListState();
```
franzliedke, line 17: Again, defensive coding that probably isn't needed... especially not once we have Typescript.

js/src/forum/components/DiscussionList.js
```diff
@@ -14,53 +14,34 @@ import Placeholder from '../../common/components/Placeholder';
...
       loading = Button.component({
         children: app.translator.trans('core.forum.discussion_list.load_more_button'),
         className: 'Button',
-        onclick: this.loadMore.bind(this),
+        onclick: state.loadMore.bind(state),
       });
     }
 
-    if (this.discussions.length === 0 && !this.loading) {
+    if (state.discussions.length === 0 && !state.loading) {
```
franzliedke, line 36: Encapsulate plz :grin:

js/src/forum/components/IndexPage.js
```diff
@@ -33,26 +33,27 @@ export default class IndexPage extends Page {
...
+    if (app.cache.discussionList.hasDiscussions()) {
       // Compare the requested parameters (sort, search query) to the ones that
       // are currently present in the cached discussion list. If they differ, we
       // will clear the cache and set up a new discussion list component with
       // the new parameters.
       Object.keys(params).some((key) => {
-        if (app.cache.discussionList.props.params[key] !== params[key]) {
-          app.cache.discussionList = null;
+        if (app.cache.discussionList.params[key] !== params[key]) {
+          app.cache.discussionList.clear();
```
franzliedke, line 48: The comment above is now incorrect.

js/src/forum/state/DiscussionListState.js
```diff
@@ -0,0 +1,153 @@
...
+
+    if (preloadedDiscussions) {
+      return Promise.resolve(preloadedDiscussions);
+    }
+
+    const params = this.requestParams();
+    params.page = { offset };
+    params.include = params.include.join(',');
+
+    return app.store.find('discussions', params);
```
franzliedke, line 103: Do we want to pass in the `app` (or store, or whatever dependencies there are) to the state as well? :thinking: That would make it easier to test them, potentially... :thinking:

js/src/forum/ForumApplication.js
askvortsov1: Exclusively for BC reasons, to not break stuff that calls app.cache.discussionList. Although tbh, we could very well store it as a direct property of app, and have app.cache.discussionList just be a reference.

js/src/forum/utils/DiscussionControls.js
```diff
@@ -230,11 +230,8 @@ export default {
       }
 
       return this.delete().then(() => {
-        // If there is a discussion list in the cache, remove this discussion.
-        if (app.cache.discussionList) {
-          app.cache.discussionList.removeDiscussion(this);
-          m.redraw();
-        }
+        app.cache.discussionList.removeDiscussion(this);
+        m.redraw();
```
franzliedke, line 234: We're being quite inconsistent with where we trigger redraws - sometimes within the state, sometimes in places like this. But I'll bring up that question in the ticket, it's more of a general point of discussion for me.

js/src/forum/ForumApplication.js
franzliedke: Okay, I guess the other changes made this more clear - you probably want to reduce breakage, because other places were already using this... I guess it's still broken for most usages, though, so renaming might be an option for consistency? Or is there another semantic difference you see between the other states and the "cache"?

js/src/forum/utils/DiscussionControls.js
franzliedke: This is one of the places I was referring to in https://github.com/flarum/core/issues/2144#issuecomment-632957701. 

Potentially, I could see `removeDiscussion()`, `addDiscussion()`, `clear()` and even `setParams()` all trigger redraws - that would slightly change behavior in some places.

js/src/forum/ForumApplication.js
askvortsov1: On the bright side, it isn't really broken because all the public api methods from the component exist on the state. Moved, BC layer added.

js/src/forum/state/DiscussionListState.js
askvortsov1: Isn't that what's done here? App is globally available

js/src/forum/utils/DiscussionControls.js
askvortsov1: Add to all except setParams: usually, setParams would be followed by refresh, which already triggers redraw.

js/src/forum/state/DiscussionListState.js
```diff
@@ -0,0 +1,169 @@
...
+   * Are there discussions stored in the discussion list state?
+   */
+  hasDiscussions() {
+    return this.discussions.length > 0;
+  }
+
+  /**
+   * Are discussions currently being loaded?
+   */
+  isLoading() {
```
datitisev, line 166: I don't think having this method is necessary.

askvortsov1: Probably not, we need to decide how much we actuallt want to encapsulate. One benefit of over encapsulation is that when we eventually get rid of these states and move the data out into a central object, we can provide full backwards compatability.

datitisev: Good point. Hm.

js/src/forum/state/DiscussionListState.js
franzliedke: That's the opposite of passing it in. We're accessing a global variable, I was suggesting to do dependency injection.

js/src/forum/utils/DiscussionControls.js
franzliedke: But isn't the whole point of these state objects that triggering write operations on the state automatically results in a redraw, so that all other places that depend on this state (which the caller is not necessarily aware about) are updated as well?

In other words, if the call has to be followed by another call, why not combine the two? (There might be edge cases or semantics I am missing here...)

js/src/forum/state/DiscussionListState.js
askvortsov1: Redux, Meiosis, and most other state management methodologies are based around having a single, global state object. Now, typically the main reason for wanting to avoid globals is that anything, anywhere can modify them, leading to unintended consequences. That's where the main point of these methodologies comes in: the modification of the global state is controlled by the methodology. For instance, with Meiosis, http://meiosis.js.org/tutorial/01-introduction.html

js/src/forum/utils/DiscussionControls.js
askvortsov1: Good point on this one, not sure why I split it up this way. I added in that setParams calls refresh afterwards. Alternatively, we could refactor it into one method so that if params are provided to refresh, it will set them.

askvortsov1: Done!

askvortsov1: Put into 1 method.

js/src/forum/components/DiscussionPage.js
```diff
@@ -37,7 +38,7 @@ export default class DiscussionPage extends Page {
     // page, then we don't want Mithril to redraw the whole page – if it did,
     // then the pane would which would be slow and would cause problems with
     // event handlers.
-    if (app.cache.discussionList) {
+    if (app.cache.discussionList.hasDiscussions()) {
```
franzliedke, line 41: This seems like slightly different behavior - "the discussion list being loaded" (i.e. present) is a different condition from "the discussion list having discussions". An empty list can be loaded as well, right?

js/src/forum/components/DiscussionPage.js
```diff
@@ -90,9 +91,9 @@ export default class DiscussionPage extends Page {
 
     return (
       <div className="DiscussionPage">
-        {app.cache.discussionList ? (
+        {app.cache.discussionList.hasDiscussions() ? (
           <div className="DiscussionPage-list" config={this.configPane.bind(this)}>
-            {!$('.App-navigation').is(':visible') ? app.cache.discussionList.render() : ''}
+            {!$('.App-navigation').is(':visible') && <DiscussionList state={app.cache.discussionList} />}
```
franzliedke, line 96: Interesting little change in capabilities: the component is no longer dynamic. 

@datitisev @askvortsov1 What about having the component class be overridable via prop, with `DiscussionList` being the default? Open for arguments...

js/src/forum/components/IndexPage.js
```diff
@@ -33,26 +33,25 @@ export default class IndexPage extends Page {
...
           return true;
         }
       });
     }
 
-    if (!app.cache.discussionList) {
-      app.cache.discussionList = new DiscussionList({ params });
+    if (!app.cache.discussionList.hasDiscussions()) {
+      app.cache.discussionList.refresh(true, params);
```
franzliedke, line 54: Not a big fan of boolean arguments - they hide intent at the call sites (e.g. here). What do you think? Typical alternatives:
- option arguments
- two methods with more explicit names (one can call each other in the implementation, obviously)

js/src/forum/components/DiscussionPage.js
franzliedke: I think this was brought up by David in another place before, so I might be misguided pointing this out again here.

js/src/forum/state/DiscussionListState.js
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  /**
+   * Clear and reload the discussion list.
+   */
+  refresh(clear = true, params) {
+    if (clear) {
+      this.loading = true;
+      this.discussions = [];
+    }
```
franzliedke, line 72: Should the `loading` flag be set regardless of the value of `clear`?

js/src/forum/components/DiscussionPage.js
datitisev: Correct, I brought it up @ https://github.com/flarum/core/pull/2150#discussion_r422225696.

js/src/forum/components/DiscussionPage.js
datitisev: With _the component is no longer dynamic_ do you mean that other extensions can't replace the class like they could set `app.cache.discussionList` before? Not sure if any ext uses that, but if you want to keep the functionality, i'd be fine with that.

js/src/forum/components/IndexPage.js
datitisev: Does _option arguments_ mean merge the two parameters into an options object, which would identify the options better by the object keys?

js/src/forum/state/DiscussionListState.js
franzliedke: I understand that concept, and the states and stores being a central place makes sense. I was discussing the method of accessing this central place: by calling on a global name (`app.store`) or by passing it in to the constructor (dependency injection). The latter not only makes tests easier, but it also allows passing in a different object with the same interface, should the need ever arise (and part of our extensibility philosophy is not knowing what use-cases might come up :laughing: ).

askvortsov1: Alright, I see your point here. I'll change it so that if `app` is needed by the states, it gets passed in on initialization (or should `app.store` be passed in?)

js/src/forum/ForumApplication.js
```diff
@@ -63,10 +64,18 @@ export default class ForumApplication extends Application {
...
+   */
+  discussionList = new DiscussionListState();
+
   constructor() {
     super();
 
     routes(this);
+
+    // Deprecated, remove in beta 15. BC layer.
+    this.cache.discussionList = this.discussionList;
```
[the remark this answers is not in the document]
askvortsov1: So this is a note-to-self for later, not something actionable, right? If so, I'll add that to the BC layer comment?

js/src/forum/components/IndexPage.js
```diff
@@ -33,26 +33,25 @@ export default class IndexPage extends Page {
...
       // Compare the requested parameters (sort, search query) to the ones that
       // are currently present in the cached discussion list. If they differ, we
-      // will clear the cache and set up a new discussion list component with
-      // the new parameters.
+      // will clear the cache update the parameters.
       Object.keys(params).some((key) => {
-        if (app.cache.discussionList.props.params[key] !== params[key]) {
-          app.cache.discussionList = null;
+        if (app.cache.discussionList.getParams()[key] !== params[key]) {
+          app.cache.discussionList.clear();
```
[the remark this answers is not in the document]
askvortsov1: In general, we'd like to redraw when clearing, but in this case, it's true that we get an unnecessary redraw. Since `refresh` has `clear` as a parameter, instead of clearing if this condition is met, we can set a constant which we will pass into that refresh call.

js/src/forum/state/DiscussionListState.js
askvortsov1: After some more thought, I'm going to side with passing in the entire app. Pretty much every piece of the app could be reused in some way or another.

js/src/forum/components/DiscussionPage.js
askvortsov1: Sure, but can we make that an attribute of DiscussionListState?

js/src/forum/components/IndexPage.js
askvortsov1: We initially had 2 methods, but setParams isn't really ever used independently. I'd prefer an options object if it's down to those 2 choices.

js/src/forum/state/DiscussionListState.js
franzliedke: Okay, thanks. No strong opinion there right now. We can revisit this when we have states everywhere, and see how it looks then / what patterns emerge / whether we want to refactor towards another solution.

franzliedke: I definitely like where this is going already.

js/src/forum/ForumApplication.js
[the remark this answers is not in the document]
franzliedke: Just thinking out loud, I guess. :yum: 

I think we'll remember when removing the BC stuff - either that, or quickly notice everything is broken. :laughing:

js/src/forum/components/IndexPage.js
franzliedke: Not necessarily both arguments in the options hash, it could also be `refresh(params, {clear: true})`.

js/src/forum/components/DiscussionPage.js
franzliedke: I thought (and was hoping) we're trying to keep states and components clearly separated?

askvortsov1: Well, for modals and alerts, the class of the component is included in the state, right? Because the state of the application is that there is currently a modal with these attrs and this class?

franzliedke: Ah right, a reference to the class should be fine.

js/src/forum/components/DiscussionPage.js
franzliedke: Ah, right. This is the reverse condition, though, and here we now have an unhandled case of "results loaded, but none found"...

js/src/forum/components/DiscussionList.js
```diff
@@ -11,56 +11,38 @@ import Placeholder from '../../common/components/Placeholder';
...
 
-    if (this.discussions.length === 0 && !this.loading) {
+    if (!state.hasDiscussions() && !state.isLoading()) {
       const text = app.translator.trans('core.forum.discussion_list.empty_text');
       return <div className="DiscussionList">{Placeholder.component({ text })}</div>;
     }
 
     return (
-      <div className={'DiscussionList' + (this.props.params.q ? ' DiscussionList--searchResults' : '')}>
+      <div className={'DiscussionList' + (state.params.q ? ' DiscussionList--searchResults' : '')}>
```
franzliedke, line 43: This one looks like we're breaking encapsulation. (We're using `getParams()` elsewhere. But really, I'd rather see a method that has a name that explains the semantics of what we're doing here - after all, the `q` param already has special meaning within the state class, IIRC.)

js/src/forum/components/DiscussionPage.js
askvortsov1: I don't see any benefit to showing the slideout when there's no discussions?

js/src/forum/components/DiscussionList.js
askvortsov1: Agreed, and done: this is clearer

js/src/forum/components/DiscussionList.js
```diff
@@ -11,56 +11,38 @@ import Placeholder from '../../common/components/Placeholder';
...
 
-    if (this.discussions.length === 0 && !this.loading) {
+    if (state.empty()) {
       const text = app.translator.trans('core.forum.discussion_list.empty_text');
       return <div className="DiscussionList">{Placeholder.component({ text })}</div>;
     }
 
     return (
-      <div className={'DiscussionList' + (this.props.params.q ? ' DiscussionList--searchResults' : '')}>
+      <div className={'DiscussionList' + (state.isSearchResults() ? ' DiscussionList--searchResults' : '')}>
```
datitisev, line 43: `hasSearchResults`? I don't know how something is a search results.

js/src/forum/components/DiscussionPage.js
```diff
@@ -49,6 +50,7 @@ export default class DiscussionPage extends Page {
     app.history.push('discussion');
 
     this.bodyClass = 'App--discussion';
+    this.discussionListClass = DiscussionList;
```
datitisev, line 53: What does this do?

datitisev:
**Breaking changes:**

- Moved methods & properties from component to state class
  - Accessible via `app.discussions` instead of `app.cache.discussionList`
  - Has helper methods to encapsulate possible future breaking changes
- DiscussionList component requires state prop
- Changed behavior, show discussion list pane if there are discussions in the discussion list
- Maybe a few more I missed

js/src/forum/components/DiscussionList.js
franzliedke: `isFiltered`?

datitisev: Perhaps `searched`? `isSearching`? `isSearch`? Either works yeah.

js/src/forum/components/DiscussionPage.js
franzliedke: That was a leftover an intermediate step which we decided to revert again. Good catch, removed in master.

PR #2151, opened 2020-05-08
Remove app.search instance, cache app.cache.searched

askvortsov1:
**Fixes Part Of #2144**

**Changes**
See title. There's no reason for the app.search component to be saved other than to cache `searched` (array of previous searches), so I just moved that array out to `app.cache.searched`

Do we even need `app.search` to be accessible? Looking at the code, there's 0 public API for it, and I don't understand why we'd need it to be saved in app? Can we just put the sources back into Search, and initialize it like any other component?

EDIT: Looks like this is to cache previous search queries (good idea). I'm going to move that into the global app.

**Confirmed**


askvortsov1:
@clarkwinkelmann as a followup to https://github.com/flarum/core/issues/2144#issuecomment-626013131, what would you like to move out into a state class? Right now, I'm envisioning:

- `this.getCurrentSearch()`
- `this.value`
- `sources` (not sure about this one tbh)

js/src/forum/components/Search.js
```diff
@@ -158,7 +151,7 @@ export default class Search extends Component {
 
         clearTimeout(search.searchTimeout);
         search.searchTimeout = setTimeout(() => {
-          if (search.searched.indexOf(query) !== -1) return;
+          if (app.cache.searched.indexOf(query) !== -1) return;
```
datitisev, line 154: This is something we want to avoid. We should be passing the searched state through a prop, not directly accessing `app.cache`. If another extension uses Search (e.g. fof/merge-discussions), they'd have to completely rewrite the config method to not use `app.cache.searched`. It should be passed as the state prop/attribute or something like that.

askvortsov1: Good point, I'll modify this to be `this.searched = this.props.searched || []`

datitisev: In the state? Or in the component? Do we want this available to all? 🤷

askvortsov1: Done

askvortsov1:
Decided to leave value and sources for now, as no reason to move them

js/src/forum/components/IndexPage.js
```diff
@@ -36,7 +36,7 @@ export default class IndexPage extends Page {
       app.cache.discussionList = null;
     }
 
-    const params = this.params();
+    const params = app.search.params();
```
franzliedke, line 39: Just for clarification: In other (lower-level) components, we wouldn't want to access `app` within the component, right? And instead have these dependencies passed in as props?

Here it's okay because page components don't get any props passed in, and we're at the top level anyway?

js/src/forum/states/SearchState.js
```diff
@@ -0,0 +1,73 @@
...
+
+  /**
+   * Redirect to the index page using the given sort parameter.
+   *
+   * @param {String} sort
+   */
+  changeSort(sort) {
+    const params = this.params();
+
+    if (sort === Object.keys(app.cache.discussionList.sortMap())[0]) {
```
franzliedke, line 52: I suppose this will need changing after #2150 is merged?

js/src/forum/components/Search.js
```diff
@@ -217,7 +214,7 @@ export default class Search extends Component {
     this.value('');
 
     if (this.getCurrentSearch()) {
-      app.current.clearSearch();
+      app.search.clearSearch();
```
franzliedke, line 217: This should work as well (while being more consistent), right?

```suggestion
      this.state.clearSearch();
```

js/src/forum/components/Search.js
```diff
@@ -19,6 +20,8 @@ import UsersSearchSource from './UsersSearchSource';
  */
 export default class Search extends Component {
   init() {
+    this.state = this.props.state || new SearchState();
```
franzliedke, line 23: Is the fallback just defensive coding or is this going to happen in known code? If not, let's just require it? (Should also be added to the component's docblock, I guess.)

js/src/forum/components/Search.js
```diff
@@ -70,7 +66,7 @@ export default class Search extends Component {
 
     // Initialize search input value in the view rather than the constructor so
     // that we have access to app.current.
-    if (typeof this.value() === 'undefined') {
+    if (!this.value()) {
```
franzliedke, line 69: The comment above may no longer be correct?

js/src/forum/components/Search.js
```diff
@@ -158,7 +155,7 @@ export default class Search extends Component {
 
         clearTimeout(search.searchTimeout);
         search.searchTimeout = setTimeout(() => {
-          if (search.searched.indexOf(query) !== -1) return;
+          if (state.cachedSearches.indexOf(query) !== -1) return;
```
franzliedke, line 158: Now that we're no longer operating on `this`, but a dependency instead, it might be worth encapsulating this logic in some nicely named methods? E.g. `state.isCached()` and (for the next change) `state.cache()`.

js/src/forum/components/IndexPage.js
askvortsov1: Yes. But the problem isn't that they are accessing app: that in of itself is fine. We want it to be extensible. No one's gonna instantiate an instance of Index Page somewhere random, but the lower level components might be reused by extensions in unexpected ways.

franzliedke: > the problem isn't that they are accessing app: that in of itself is fine.

Well, it would break the unidirectional data flow we're envisioning, right?

js/src/forum/components/Search.js
askvortsov1: Tbh, I think it might be better to move the entire method into SearchState, along with the value prop.

js/src/forum/components/Search.js
askvortsov1: Doesn't happen anywhere I know of; and agree, explicit is better than implicit. As to docblock, yes on adding that, but that'll be Typescript-ized soon enough, hopefully.

js/src/forum/components/Search.js
askvortsov1: uh nope, not at all.

js/src/forum/components/Search.js
askvortsov1: yeah I'd agree with that.

js/src/forum/states/SearchState.js
askvortsov1: Yep!

askvortsov1:
@franzliedke revisiting this now, Im actually starting to think that this might not be enough of a change. Feels like `value` should definitely go to the state, and the clear method. 

Sources I'm torn on: I don't want inheritance trees off of these states since I view them as an intermediate step between our current situation and http://meiosis.js.org/, so I'd prefer that people extend Search. But then again, perhaps additional sources should be provided as a parameter when instantiating a SearchState? Kinda torn on this one.

js/src/forum/components/Search.js
franzliedke: Including the Mithril stuff (the prop and the `redraw()`)? Do we want rendering / framework stuff in those states?

js/src/forum/states/SearchState.js
franzliedke: To play devil's advocate, how do we track these cross-dependencies?

askvortsov1: We don't. Probably should, but I don't know of a way that would be too great. I suppose TODO comments in the code, or editing the description of each issue could work?

js/src/forum/components/Search.js
askvortsov1: This is a difficult question, but I'm leaning towards yes on this one. My strictest no-no would be including any jsx in state methods (not the biggest fan of storing JSX in states as in #2163, but that's not AS critical). The way I see it, every state method is either involved in:

- Changing state data
- Reacting to changes in state data by setting off animations / redraws.
- Presenting the public API for components to interact with other components (also bad, at no point should any component have access to an instance of any other component (although we can use events/pass bound functions to circumvent some of the limitations))

So redraws are fine, as is direct DOM manipulation via JQuery (although when we CAN, I'd prefer to still have that in component classes, especially if it's a big method).

askvortsov1: One other thing I'll note is that I view states as an intermediate step on the way to a proper state management methodology, at which point the data in states and the public API of manipulating state would be further split off. But one step at a time.

js/src/forum/states/SearchState.js
franzliedke: I guess we can leave a TODO comment in the PR discussion of #2150, then merge this one here first.

franzliedke:
Inheritance trees? Additional sources? Can you expand on those? :wink: 

Is that anything being done by extensions currently? If not, let's worry about it now...

js/src/forum/components/Search.js
askvortsov1: Don't read this / reply here, I posted an expanded version in the main 'Frontend Needs Better State' discussion

askvortsov1:
I don't want people making new classes inheriting off of the states. For sources, I meant search sources: currently a property of the Search component, but could be better in Search State tbh

js/src/forum/components/Search.js
franzliedke: I will comment on the specific issue I commented on. I can live with redraws in states for now, but would prefer trying to avoid props in there. (Then again, these props are just setters that trigger redraws, if I understand correctly?)

datitisev:
> I don't want people making new classes inheriting off of the states

To give context, I assume this is referring to what I brought up today internally about my AlertState classes that act like components (but can be extended) in https://github.com/flarum/core/commit/4c8af1cdf876b9bd3770f76c33eb6c9ac80608f0 (my frontend rewrite branch).

These AlertState instances are passed to `app.alerts.show`, can be extended like components, and contain component & JSX logic.

Sasha doesn't like this approach. I understand the term "state" here isn't good.

> My problem with that [implementation] is that it's not moving us towards our eventual goal of using a sensible state management system. I view all these 'State' classes as intemediaries until we're able to adopt a proper state management pattern like http://meiosis.js.org/
> \- @askvortsov1

js/src/forum/components/IndexPage.js
askvortsov1: Not quite: My main goal is that the rendered content of the app can be considered the output of a somewhat pure, deterministic function, the input of which is the current state.

This means that if a component wants to change another component, it would do so by changing the state, and not by changing the component directly.

The only element I'm unsure of (and one we're gonna need to figure out) is transitional animations / direct DOM manipulations via JQuery, but we need to take this one step at a time.

js/src/forum/components/Search.js
askvortsov1: I moved value into state (and de-prop-ized it), but am going to keep clear() in the component, as Byobu needs to override that. Besides, it has to do with component functionality, not global search functionality.

js/src/forum/components/Search.js
```diff
@@ -272,6 +237,19 @@ export default class Search extends Component {
     return $item;
   }
 
+  /**
+   * Clear the search input and the current controller's active search.
+   */
```
franzliedke, line 242: The docblock is not correct any longer.

js/src/forum/states/SearchState.js
```diff
@@ -0,0 +1,100 @@
...
+
+  /**
+   * Return the current search query, if any. This is implemented to activate
+   * the search box in the header.
+   *
+   * @see Search
+   * @return {String}
+   */
+  getCurrentSearch() {
+    return this.params().q;
```
franzliedke, line 53: Previous call sites checked for the controller supporting search, right? So we have a slight change in behavior - in theory, other places having a `q` query param would now be interpreted as having a current search, too. Not sure if this is a practical problem. :thinking:

js/src/forum/components/Search.js
askvortsov1: Just to quickly expand on this confusing wording:

- the `value` attribute of `Search.js` has been moved into `SearchState`
- Since a mithril prop is no longer necessary for it to function properly, it now uses a regular old attribute.
- Byobu overrides the functionality of the `clear` method in https://github.com/FriendsOfFlarum/byobu/blob/6ad1252790aa01329c01d6a5d2ed3c458b32cc67/js/src/forum/components/RecipientSearch.js#L129. If we want to keep this flexibility without encouraging people to make classes extending SeachState, `clear` needs to be in the Search.js class.

js/src/forum/states/SearchState.js
askvortsov1: Now that value is stored in SearchState, we could always just return that?

js/src/forum/components/Search.js
```diff
@@ -211,16 +189,10 @@ export default class Search extends Component {
...
    */
   clear() {
-    this.value('');
-
-    if (this.getCurrentSearch()) {
-      app.current.clearSearch();
-    } else {
-      m.redraw();
-    }
+    this.clear();
```
franzliedke, line 195: That's not going to go well. :stuck_out_tongue_winking_eye:

askvortsov1: :facepalm:

js/src/forum/states/SearchState.js
franzliedke: That won't work - `getCurrentSearch()` is the fallback value called in `getValue()`, so you'd have an infinite loop there.

That said, what about my question? We don't check whether the current page / controller supports search anymore, right?

askvortsov1: > I do not like the search PR
> This code is more opaque than tar
> This jumbled madness breaks my mind,
> and while an answer we shall find,
> what that shall be, I can't forsee
> I do not like this, Franz Liedke.

But on a serious note, we don't but we could just move that check here and return null if the page isn't searchable?

askvortsov1: I'll be honest, I don't really like the naming here. I think it needs to be clearer to indicate that getCurrentSearch is whatever was searched on the last page refresh, and getValue is the current value of the search input. I feel like there's way too many methods involved in search and it could probably be clearer. Not sure where to go with it though.

franzliedke:
For documentation purposes:

- We introduced a new `SearchState` base class that can be used as a prop (for the `Search` component) for simple search interfaces (with a local cache) that don't have global side effects.
- The new `GlobalSearchState` is used for the global discussion search, inherits the storage and adds global side effects (taking the initial value from the URL and reflecting changes there as well as triggering redraws) and a few methods for sorting.

I hope you like this @datitisev :hugs:

js/src/forum/ForumApplication.js
```diff
@@ -68,9 +68,11 @@ export default class ForumApplication extends Application {
    * An object which stores previously searched queries and provides convenient
    * tools for retrieving and managing search values.
    *
-   * @type {SearchState}
+   * @type {HeaderSearchState}
    */
-  search = new SearchState();
+  search = new HeaderSearchState();
+
+  searchResults = false;
```
datitisev: This isn't used anywhere. I assume this was before you added the static `hasSearchResults` property to page ?

datitisev: I am an idiot and was looking at an individual commit. Ignore me.

datitisev:
@franzliedke Thanks :+1: 🤗 ❤️ 

**Breaking changes:**
- Many methods removed from `Search` component
  - Moved to state that needs to be stored & passed to component
  - Use `SearchState` class for basic search interfaces - `GlobalSearchState` applies mainly to `IndexPage` & header search
  - Some methods have been renamed
    - `getCurrentSearch` -> `getInitialSearch`
- Instead of using `searching` method in Page components where we want searching to apply (e.g. IndexPage), set static property `hasSearchResults` to true
- Perhaps a few more, depending on what your code does with Search exactly

PR #2154, opened 2020-05-08
Upgrade to Laravel 5.8

franzliedke:
**Tackles #2055**

Now that the biggest hurdle for a Laravel 5.8 upgrade has been removed in #2142, we can make the first step and upgrade to Laravel 5.8.

- I had missed [one change](https://laravel.com/docs/5.8/upgrade#container-generators) in 5.8 that required changing the type signature of the `HandleErrors` constructor. (That's because `Container::tagged()` now returns an iterator instead of an array, which is kinda cool because the error reporters will only be resolved when actually needed, i.e. when an error occurs.)
- The `env()` method was used in the test setup script. That's one of the global helpers from `illuminate/support`, but apparently no longer works without some other dependencies. Oh well, `getenv()` works just fine.

**Confirmed**

**Required changes:**

*None*


franzliedke:
Nice, legitimately failing tests. :+1: Will check these out later.

franzliedke:
Fixed. :muscle:

Issue #487, opened 2015-09-15
Setting open_basedir can cause installation problems


pauljherring:
Example situation:

Whole domain present at: /var/www/vhosts/example.com/httpdocs and served as http://example.com
Subdomain present in subdirectory at /var/www/vhosts/example.com/flarum and served as http://flarum.example.com

php.ini set to have the parent directory open:
`open_basedir=/var/www/vhosts/example.com/:/tmp`

Installation to /var/www/vhosts/example.com/flarum fails - in the instance experienced - around 
…/­vendor/­s9e/­text-formatter/­src/­Plugins/­BBCodes/­Configurator/­Repository.php30

See http://discuss.flarum.org/d/977-invalid-repository-file for discovery process.

logixtreeCode:
Faced the same problem. 
Problematic file : /vendor/flarum/core/src/Frontend/Compiler/JsCompiler.php
You can simply adjust the sys_get_temp_dir() to getcwd();. Since we are simply creating a temporary file then deleting it. It won't cause any harm. Here is the adjusted code. 
 
```
        // Add a comment to the end of our file to point to the sourcemap
        // we just constructed. We will then write the JS file, save the
        // map to a temporary location, and then move it to the asset dir.
        $output[] = '//# sourceMappingURL='.$this->assetsDir->url($mapFile);

        $this->assetsDir->put($file, implode("\n", $output));
        $mapTemp = tempnam(getcwd(), $mapFile);
        //print_r($mapTemp);die;
        $map->save($mapTemp);
        $this->assetsDir->put($mapFile, file_get_contents($mapTemp));
        @unlink($mapTemp);

        return true;
```

In my case, I had the open_basedir setup perfectly but php was not picking up the local values. But above case sorted out the concern. 

Regards,
Team LogixTree

luceos:
Fixed in https://github.com/flarum/core/commit/54660ebd6311f9ea142f1b573263d0d907400786

PR #2155, opened 2020-05-08
Clean up usages / deprecate path helpers

franzliedke:
**Relates to #2055**

- Deprecate the global path helpers, e.g. `public_path()`
- Deprecate the path helpers on the `Application` class, e.g. `Application::publicPath()`

All of these are left-overs from our deeper Laravel integration. The recommended way to access these is to use the `Paths` class, which is injectable as a proper dependency.

All remaining changes are cleanups of places where we used the helpers ourselves - some of them quite obscure (e.g. in `WritablePaths`).

Any downsides to removing these? I haven't yet found a case, where refactoring away from the global helpers did not result in an improvement.

**Confirmed**

**Required changes:**

- [ ] Related documentation PR: TODO
- No changes to core extensions necessary

### DIFF
--- a/src/Formatter/FormatterServiceProvider.php
+++ b/src/Formatter/FormatterServiceProvider.php
@@ -10,6 +10,7 @@
 namespace Flarum\Formatter;
 
 use Flarum\Foundation\AbstractServiceProvider;
+use Flarum\Foundation\Paths;
 use Illuminate\Cache\Repository;
 use Illuminate\Contracts\Container\Container;
 
@@ -24,7 +25,7 @@ class FormatterServiceProvider extends AbstractServiceProvider
             return new Formatter(
                 new Repository($container->make('cache.filestore')),
                 $container->make('events'),
-                $this->app['flarum']->storagePath().'/formatter'
+                $this->app[Paths::class]->storage.'/formatter'
             );
         });
 
--- a/src/Frontend/Compiler/JsCompiler.php
+++ b/src/Frontend/Compiler/JsCompiler.php
@@ -50,16 +50,12 @@ class JsCompiler extends RevisionCompiler
         }
 
         // Add a comment to the end of our file to point to the sourcemap
-        // we just constructed. We will then write the JS file, save the
-        // map to a temporary location, and then move it to the asset dir.
+        // we just constructed. We will then store the JS file and the map
+        // in our asset directory.
         $output[] = '//# sourceMappingURL='.$this->assetsDir->url($mapFile);
 
         $this->assetsDir->put($file, implode("\n", $output));
-
-        $mapTemp = @tempnam(storage_path('tmp'), $mapFile);
-        $map->save($mapTemp);
-        $this->assetsDir->put($mapFile, file_get_contents($mapTemp));
-        @unlink($mapTemp);
+        $this->assetsDir->put($mapFile, json_encode($map, JSON_UNESCAPED_SLASHES));
 
         return true;
     }
--- a/src/Locale/LocaleServiceProvider.php
+++ b/src/Locale/LocaleServiceProvider.php
@@ -11,6 +11,7 @@ namespace Flarum\Locale;
 
 use Flarum\Foundation\AbstractServiceProvider;
 use Flarum\Foundation\Event\ClearingCache;
+use Flarum\Foundation\Paths;
 use Flarum\Settings\SettingsRepositoryInterface;
 use Illuminate\Contracts\Events\Dispatcher;
 use Illuminate\Contracts\Translation\Translator as TranslatorContract;
@@ -73,6 +74,6 @@ class LocaleServiceProvider extends AbstractServiceProvider
 
     private function getCacheDir(): string
     {
-        return $this->app['flarum']->storagePath().'/locale';
+        return $this->app[Paths::class]->storage.'/locale';
     }
 }
--- a/src/Queue/QueueServiceProvider.php
+++ b/src/Queue/QueueServiceProvider.php
@@ -13,6 +13,7 @@ use Flarum\Console\Event\Configuring;
 use Flarum\Foundation\AbstractServiceProvider;
 use Flarum\Foundation\ErrorHandling\Registry;
 use Flarum\Foundation\ErrorHandling\Reporter;
+use Flarum\Foundation\Paths;
 use Illuminate\Contracts\Debug\ExceptionHandler as ExceptionHandling;
 use Illuminate\Contracts\Queue\Factory;
 use Illuminate\Contracts\Queue\Queue;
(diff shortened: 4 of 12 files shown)


src/Frontend/Compiler/JsCompiler.php
```diff
@@ -56,7 +56,7 @@ protected function save(string $file, array $sources): bool
 
         $this->assetsDir->put($file, implode("\n", $output));
 
-        $mapTemp = @tempnam(storage_path('tmp'), $mapFile);
```
askvortsov1, line 59: Any particular reason for switching out the storage dir we include in flarum/flarum for php's default tmp dir? Not opposed, just curious

src/Formatter/FormatterServiceProvider.php
```diff
@@ -24,7 +25,7 @@ public function register()
             return new Formatter(
                 new Repository($container->make('cache.filestore')),
                 $container->make('events'),
-                $this->app['flarum']->storagePath().'/formatter'
+                $this->app[Paths::class]->storage.'/formatter'
```
askvortsov1, line 28: Why not $container->make? Also not insisting on anything, moreso curious as to the difference.

franzliedke: I probably didn't bother changing it, because that's what the old code used. :thinking: 

I typically prefer `make()` when I write new code, because that one is part of the `Container` *contract*, while the array offset stuff comes from the concrete `Container` class.

src/Frontend/Compiler/JsCompiler.php
franzliedke: This was the faster solution compared to trying to inject `Paths` - and the sys temp dir does the trick. :shrug:

src/Formatter/FormatterServiceProvider.php
luceos: Agreed, these should use the `make()` method instead. Good spot Sasha.
```suggestion
                $this->app->make(Paths::class)->storage.'/formatter'
```

src/Frontend/Compiler/JsCompiler.php
luceos: We cannot rely on sys_get_temp_dir() on shared hosting. In addition the explanation for `tempnam` is that it wraps around `sys_get_temp_dir`:

> Creates a file with a unique filename, with access permission set to 0600, in the specified directory. If the directory does not exist or is not writable, tempnam() may generate a file in the system's temporary directory, and return the full path to that file, including its name. 

I think the compiler has to rely on the storage directory. Let's revert this change because it will break shared hosting installations and badly configured environments where a tmp sys dir isn't writeable.

src/Locale/LocaleServiceProvider.php
```diff
@@ -73,6 +74,6 @@ private function getDefaultLocale(): string
 
     private function getCacheDir(): string
     {
-        return $this->app['flarum']->storagePath().'/locale';
+        return $this->app[Paths::class]->storage.'/locale';
```
luceos, line 77:
```suggestion
        return $this->app->make(Paths::class)->storage.'/locale';
```

src/Queue/QueueServiceProvider.php
```diff
@@ -70,7 +71,7 @@ public function register()
         // Override the Laravel native Listener, so that we can ignore the environment
         // option and force the binary to flarum.
         $this->app->singleton(QueueListener::class, function ($app) {
-            return new Listener($app->basePath());
+            return new Listener($app[Paths::class]->base);
```
luceos, line 74:
```suggestion
            return new Listener($app->make(Paths::class)->base);
```

src/Frontend/Compiler/JsCompiler.php
franzliedke: I tend to think the key word here is "badly configured environments", not breaking them. :thinking: 

For reference: https://github.com/flarum/core/issues487

src/Formatter/FormatterServiceProvider.php
franzliedke: I don't think this change belongs here. I agree with the proposal, but if we want to do it, we should do it for all occurrences (and this PR doesn't touch all of them). Feel free to take care of that in a separate PR. :wink:

src/Frontend/Compiler/JsCompiler.php
franzliedke: Potentially also relevant: [my comment here](https://github.com/flarum/core/pull/2117#issuecomment-611979392). Not sure if this is what your comment is about, @luceos. Any sources or pointers would be helpful.

askvortsov1: > For reference: #487

It seems like the precedent here is using the flarum storage dir? If I understand https://github.com/flarum/core/pull/2117#issuecomment-611979392 correctly and its no longer relevant, I'm fine with this approach.

franzliedke: I'm not 100% sure, which is why I was hoping for clarification from @luceos.

luceos: My comments are based on experience with upload. Using a system tmp directory doesn't always guarantee success.

franzliedke: Sorry to be a bit thick - I was hoping you could expose on the "using" part. Above, I linked [to a comment of mine](https://github.com/flarum/core/pull/2117#issuecomment-611979392) that has an example of "using" the system tmp directory that was only problematic due to the specific implementation (checking for access to the *entire directory*, but not files inside). So do we have a detailed explanation somewhere, or a place to link to? :pray: 

I fully believe you there are other problems, but I would love to read about them...

luceos: I browsed the net for evidence of what I said, but found nothing. The upload repository hasn't held any reference to the function involved. So.. it seems I have been mistaken, sorry for wasting your time with this.

franzliedke: @luceos Oh, haha, I was about to say I am ready to change back to storage dir. And now I found this: https://github.com/PHPOffice/PHPWord/issues/310 :laughing:

luceos: Ooh. Great find! That's exactly what I was looking for 😂

The safest bet is always to work within the confines of the web space, especially in shared hosting environments where isolation matters.

A drawback is that we need to take care of the cleanup during failed uploads.. Something a catch should be able to handle.

franzliedke: Oh, look what I found, haha: https://github.com/flarum/core/commit/54660ebd6311f9ea142f1b573263d0d907400786 and #487. We already had all the information here, sorry for the confusion.

Anyway, it turns out the tmp file should not even be necessary, it was just circumventing the fact that the sourcemap library does not have a method for getting the file contents without writing it.

I'll push that fix soon-ish.

franzliedke: Pushed it finally. :fireworks:

askvortsov1:
Looks good to merge IMO

PR #2156, opened 2020-05-09
Clean up app.current, app.previous in JS

askvortsov1:
- Removed unnecessary app.current.discussion in Post.js
- Rework `app.current`, `app.previous` as per Plan 2 below
- Consolidate Admin Page.js and Forum Page.js into one class, with ForumPage extending


askvortsov1:
For fixing up app.current and app.previous, the main issue isn't the component itself (as Clark suggested, we can just store the class instead of the instance). The problem is data that's stored in the instances that needs to be accessed. I'm torn between 3 approaches:

1. Putting everything separately (I don't really like this)
```
app.current = T extends Page
app.currentPagePostStream = ...
app.currentPage = ...
```

2. Putting everything together

```
app.current = {
  page: IndexPage
  data: { postStream: ..., user: ..., etc: ...}
}
```

3. Putting everything kinda together

```
app.current = IndexPage
app.currentData = {postStream: ..., user: ..., etc: ...}
```

The same would be mirrored for `app.previous`. That being said, we could take some of the culprits out (cache `PostStream`'s state like we do with `DiscussionListState` in #2150, but that wouldn't be as flexible as ideal. We could also pass down the relevant data through a long line of props, but that would be a pain.

askvortsov1:
For this one, David proposed (and I agree) that we should put current, currentData, previous, and previousData into a PageState object, instead of modifying directly.

datitisev:
Not sure what you mean with your comment exactly, or if I didn't make myself clear, but `current` and `currentData` would be _one_ PageState instance, and `previous` & `previousData` would be another - we'd still have `app.current` and `app.previous` separate.

js/src/forum/components/DiscussionPage.js
```diff
@@ -41,7 +42,7 @@ export default class DiscussionPage extends Page {
       app.pane.enable();
       app.pane.hide();
 
-      if (app.previous instanceof DiscussionPage) {
+      if (app.previous && app.previous.subclassOf(DiscussionPage)) {
```
franzliedke, line 45:
```suggestion
      if (subclassOf(app.previous, DiscussionPage)) {
```

js/src/forum/components/Search.js
```diff
@@ -193,6 +193,7 @@ export default class Search extends Component {
   getCurrentSearch() {
     return app.current && typeof app.current.searching === 'function' && app.current.searching();
   }
+  // THIS IS NOT CHANGED AS ITS REMOVED IN THE SEARCH PR
```
franzliedke, line 196: Up for debate in https://github.com/flarum/core/pull/2151/files#r434902043

js/src/forum/components/DiscussionPage.js
franzliedke: Oh wait, you really did add that method to the `PageState` class. Never mind my suggestion then.

Would it be possible to ensure that `app.current` and `app.previous` are *always* set (at least to some kind of [null object](https://en.wikipedia.org/wiki/Null_object_pattern)) so that we can omit these long conditionals?

js/src/forum/components/DiscussionPage.js
```diff
@@ -199,6 +200,11 @@ export default class DiscussionPage extends Page {
     this.stream = new PostStream({ discussion, includedPosts });
     this.stream.on('positionChanged', this.positionChanged.bind(this));
     this.stream.goToNumber(m.route.param('near') || (includedPosts[0] && includedPosts[0].number()), true);
+
+    app.current.setData({
+      discussion: discussion,
+      stream: this.stream,
+    });
```
franzliedke, line 207: What do these calls do? They don't seem to replicate any previously existing logic...

js/src/common/components/Page.js
```diff
@@ -1,4 +1,5 @@
 import Component from '../../common/Component';
+import PageState from '../../forum/states/PageState';
```
franzliedke, line 2: It feels wrong for a component in `common` to import something from a more specific module such as `forum`. So should `PageState` be moved to `common` as well?

js/src/forum/ForumApplication.js
```diff
@@ -137,7 +138,7 @@ export default class ForumApplication extends Application {
    * @return {Boolean}
    */
   viewingDiscussion(discussion) {
-    return this.current instanceof DiscussionPage && this.current.discussion === discussion;
+    return subclassOf(this.current, DiscussionPage) && this.currentData.discussion === discussion;
```
franzliedke, line 141: This seems to have been forgotten in a refactoring step?

js/src/forum/components/DiscussionPage.js
franzliedke: That said, the `subclassOf` import is no longer needed. Here and elsewhere...

js/src/forum/components/Page.js
```diff
@@ -1,33 +1,14 @@
...
+import Page from '../../common/components/Page';
 
 /**
- * The `Page` component
+ * The `ForumPage` component
  *
  * @abstract
  */
-export default class Page extends Component {
+export default class ForumPage extends Page {
```
franzliedke, line 8: The file would need to be renamed as well?

js/src/forum/components/Page.js
```diff
@@ -1,33 +1,14 @@
...
   init() {
-    app.previous = app.current;
-    app.current = this;
+    super.init();
```
franzliedke, line 10: Why are we still overriding this method (or this class, for that matter)? The `Page` class in `common` has the `app.modal.close()` line as well.

franzliedke: The two classes were equal before (except for the `app.drawer.hide()`, which may have been an oversight), so combining them into one may make sense. On the other hand, it seems to be a BC break for no particularly important reason, and it seems at least possible that we'd want to have separate base classes for admin and forum pages at some point in the future...

franzliedke:
I agree with @datitisev's suggestion, for the record.

js/src/forum/components/Page.js
askvortsov1: I meant to put app.drawer.hide, not app.modal.close. That's my bad. As to pages, unless we're planning on adoping a radically different structure, (no modals and no PageState on admin), I'd say having one central class would be good. To keep BC, we could add an AdminPage class that extends off of Page and doesn't do anything

js/src/forum/components/DiscussionPage.js
askvortsov1: There's plenty of references to app.current.stream throughout. That was previously available by default, but not now.

js/src/forum/components/Search.js
franzliedke: This comment is now obsolete. :muscle:

js/src/common/components/ModalManager.js
```diff
@@ -43,8 +43,6 @@ export default class ModalManager extends Component {
     this.showing = true;
     this.component = component;
 
-    if (app.current) app.current.retain = true;
```
franzliedke, line 46: We just assume this was a performance optimization, right?

js/src/forum/components/Post.js
```diff
@@ -125,7 +126,7 @@ export default class Post extends Component {
       classes.push('Post--by-actor');
     }
 
-    if (user && app.current.discussion && app.current.discussion.attribute('startUserId') == user.id()) {
+    if (user && user == discussion.user()) {
```
franzliedke, line 129: If I am not mistaken, the previous code worked even when the discussion's user object has not been loaded. Can we guarantee that?

franzliedke: Same for navigating from the post to its `discussion()` - do we reverse-fill that relationship?

js/src/common/Application.js
```diff
@@ -115,6 +116,12 @@ export default class Application {
    */
   requestError = null;
 
+  /**
+   * Initialize app current and app previous to null objects
+   */
```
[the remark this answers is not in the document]
franzliedke: I'll take care of this one. :grin:

franzliedke:
Last but not least, I wonder whether we should retain the old API using JavaScript proxies (accessing the properties in "data" as if they were properties on the page state object). Would be less of a BC break, and a nicer API. For the latter, we could at least switch to something like `get('stream')` over `getData().stream` - that feels a bit too much like reaching into the depths of an object that doesn't belong to me, IMO.

But really, I would like to introduce a real concept / lifecycle for page components at some point - maybe we can just expose all their props (after they initialize themselves - AFAIK they don't receive props from the router) instead of an arbitrary collection of keys on some pages, and none on others. But not now, don't worry. :wink:

js/src/admin/compat.js
```diff
@@ -36,12 +36,12 @@ export default Object.assign(compat, {
   'components/SessionDropdown': SessionDropdown,
   'components/HeaderPrimary': HeaderPrimary,
   'components/AppearancePage': AppearancePage,
-  'components/Page': Page,
   'components/StatusWidget': StatusWidget,
   'components/HeaderSecondary': HeaderSecondary,
   'components/SettingsModal': SettingsModal,
   'components/DashboardWidget': DashboardWidget,
   'components/AddExtensionModal': AddExtensionModal,
+  'components/AdminPage': AdminPage,
```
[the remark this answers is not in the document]
askvortsov1: I'd be alright with changing each of them to 'components/Page' if that's what you mean? Or would you like the JS files to be changed as well?

js/src/forum/components/UserPage.js
```diff
@@ -71,6 +71,8 @@ export default class UserPage extends Page {
   show(user) {
     this.user = user;
 
+    app.current.setData({ user });
```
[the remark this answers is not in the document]
askvortsov1: https://github.com/flarum/core/blob/b82504b4b1b6b1acb7447d8ca9b3af3a7f49629a/js/src/forum/utils/UserControls.js#L115

But you eliminated that need in your refactor of matches, so I'll remove that line :D

js/src/common/components/ModalManager.js
askvortsov1: Yeah I guess... This isn't available in Mithril 2 anyway, and removing it didn't seem to change anything.

js/src/forum/components/Post.js
askvortsov1: > If I am not mistaken, the previous code worked even when the discussion's user object has not been loaded. Can we guarantee that?

Not sure, we could change this back though? It worked for me when I tested locally.



> Same for navigating from the post to its `discussion()` - do we reverse-fill that relationship?

Where would this be?

askvortsov1:
> Last but not least, I wonder whether we should retain the old API using JavaScript proxies (accessing the properties in "data" as if they were properties on the page state object).

Did not know that was a thing, Javascript, you crazy language! But yeah if we can put in a BC layer I'd be in favor.

> For the latter, we could at least switch to something like `get('stream')` over `getData().stream` - that feels a bit too much like reaching into the depths of an object that doesn't belong to me, IMO.

I like this more as well. Pushed a new commit adding it. 


> But really, I would like to introduce a real concept / lifecycle for page components at some point - maybe we can just expose all their props (after they initialize themselves - AFAIK they don't receive props from the router) instead of an arbitrary collection of keys on some pages, and none on others. But not now, don't worry.

100% agree!

js/src/forum/components/Post.js
franzliedke: Does it lazy-load the data? :thinking:

js/src/forum/components/UserPage.js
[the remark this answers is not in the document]
franzliedke: Wait a second, I just refactored that line, I just missed that occurrence. That means we do need the `set` call.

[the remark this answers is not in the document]
askvortsov1: Added back in

js/src/forum/components/Post.js
franzliedke: Talked it through. Resolved.

Issue #353, opened 2015-08-29
Hardening HTTP Headers


franzliedke:
Probably a good idea... here's a nice checklist: https://scotthelme.co.uk/hardening-your-http-response-headers/

d1a8lo24:
I may be wrong but most of the information concerning http headers on the link your provided is for server administrators to apply on the server side and not really via an application since most of the configurations there have to do with how a server response and replies. Most of the rules and configurations there are using the htaccess file to overwrite standard headers and when it comes to playing with the htaccess file many servers will respond differently or will just give a 500 error for miss configuration.

And not only that but most apply to either virtual servers or dedicated servers where you have more control, when it comes to share hosting that's where problems may happen since some of the configurations may not be available. Some share hosts don't even allow for you to have an htaccess file or a php.ini file, to turn on or off some of the configurations.

The reason for this is because most people don't really know what they are doing so instead of making their site more secure they end up making it less secure or to the point of overloading because of miss configuration

And yes you can have some control of certain headers with PHP but they still rely on the rules of the server so even though you're sending a particular header if the server decides not to send it or to allow to overwrite it then it won't be applied.. 

I wouldn't put this as a feature or start playing on how an application should send headers since it will become more of a headache but it should be something people should have as a reference no matter what type of site they are running is always good to know what you can apply for "security" on a server that allows it..

But then again I may be wrong, is not like I read the whole thing but just by browsing I did notice a lot of configurations I have done on my server via htaccess, again it was not something that many of my applications required or could be apply through them but because is good standards to follow when you're running a server.

franzliedke:
I think things like the Frame-Options headers (or Content-Security-Policy, as it's now called if I'm not mistaken) are pretty useful (e.g. for preventing clickjacking attacks). We'll be careful about what we consider. :)

d1a8lo24:
Yes those are headers that need to be applied via the htaccess file as a user or the server response header configurations file or similar by the server admin or host server admins.
Not via the application because of the limitations of many share hosts which can end up giving errors of miss configuration since many of them may not allow certain settings.

Most content security headers are use for CDNs and calls for REST requests and API requests as far as clickjacking this is not just a concern for Flarum this is for any site out there.

As I said before is should not be a feature but a reference somewhere in the installation guide or best practises.

Looking at other applications out there you can see that none of them will try to do this not because they can't but because not every server out there is capable of applying such settings via the application.

[...]

PR #2159, opened 2020-05-09
Fixes #2157, Explicitly set SameSite value for cookies

### LINKED ISSUE #2157: Flarum gives a cookie warning in console
Flarum gives the following warning in v1.0-beta.13:

image

*Cookie “flarum_session” will be soon rejected because it has the “sameSite” attribute set to “none” or an invalid value, without the “secure” attribute. To know more about the “sameSite“ attribute, read https://developer.mozilla.org/docs/Web/HTTP/Cookies*

tankerkiller125:
Also contains an update for the cookie library dependency

**Fixes #2157 **

Updates the cookie library and explicitly sets the same site value

* Is Lax the value we want to set? None requires HTTPs and Strict would prevent cookie use on a primary/sub-domain

**Confirmed**

### DIFF
--- a/src/Http/CookieFactory.php
+++ b/src/Http/CookieFactory.php
@@ -45,6 +45,13 @@ class CookieFactory
     protected $secure;
 
     /**
+     * Same Site cookie value
+     *
+     * @var string
+     */
+    protected $samesite;
+
+    /**
      * @param Application $app
      */
     public function __construct(Application $app)
@@ -57,6 +64,7 @@ class CookieFactory
         $this->path = $app->config('cookie.path', Arr::get($url, 'path') ?: '/');
         $this->domain = $app->config('cookie.domain');
         $this->secure = $app->config('cookie.secure', Arr::get($url, 'scheme') === 'https');
+        $this->samesite = $app->config('cookie.samesite', 'lax');
     }
 
     /**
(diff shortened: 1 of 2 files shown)


askvortsov1:
As per https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite, Lax seems to be the default value, so I think lax is best for now (this way we just avoid browsers complaining about no explicit setting). 

We can consider hardening this, as well as the stuff in #353 and https://discuss.flarum.org/d/22642-security-roadmap, later on. For now, :+1:

askvortsov1:
Only one thought: is there any place where we could add a unit test to confirm that the cookie returns samesite=lax as expected?

rob006:
Wouldn't that break some use-cases for https://github.com/flarum/embed?

askvortsov1:
I'd still like to see a unit test for this, but otherwise looks great!

tankerkiller125:
@askvortsov1 Any suggestions as to where in the testing this would fit? Not sure if we would classify it as a unit test or integration test.

askvortsov1:
It seems from the cookie library code that the actual cookies are resolved by calling toString on them ( (string) $cookie)), so perhaps in a unit test, we could assert that the proper samesite value is present in the resolved cookie string? I wouldn't mind an integration test as well, but I'd hold off on that until we have more of other stuff we can test.

src/Http/CookieFactory.php
```diff
@@ -85,6 +94,15 @@ public function make($name, $value = null, $maxAge = null)
...
         }
 
+        // Explicitly set SameSite value, use sensible default if no value provided
+        if ($this->samesite === 'strict') {
+            $cookie = $cookie->withSameSite(SameSite::strict());
+        } elseif ($this->samesite === 'none') {
+            $cookie = $cookie->withSameSite(SameSite::none());
+        } else {
+            $cookie = $cookie->withSameSite(SameSite::lax());
+        }
```
luceos, line 104: How about this? If an incorrect value is given the code will immediately error out as well.

```suggestion
        $cookie = $cookie->withSameSite(forward_static_call([SameSite, $this->samesite ?? 'lax']));
```

Also setting a default in line 67 then isn't required, just keep it null so that we know it isn't set at all because you already define a fallback here.

tankerkiller125: I'm happy with this, wasn't even ware that forward_static_call was a thing, great feature to know about for things like this though.

tankerkiller125:
Got an error with the example, code from luceos. Actually found an even nicer way of doing it I think but I need to get my unit test working properly first on my laptop. If it does work it will be `SameSite::{$this->samesite ?? 'lax'}()` which I think is much easier to understand and looks better. (Supported syntax in 7.0 and up)

Issue #1821, opened 2019-07-12
Frontend Framework Rewrite


datitisev:
* [x] Part 1 of state changes in #2144, as per the discussion in the same issue. This will reduce the work needed in the later mithril 2.0 upgrade, and increase code quality. Rework:
  * [x]  `app.current`, `app.previous`, cleanup Page classes #2156 @askvortsov1
  * [x]  `app.search`: #2151 @askvortsov1
  * [x]  `app.cache.discussionList`: #2150 @askvortsov1
  * [x]  `app.alerts` #2163 @askvortsov1
  * [x]  `app.modal` #2162 @askvortsov1
  * [x]  `app.composer` #2161 @askvortsov1
  * [x]  extract state from PostStream #2160 @askvortsov1
  * [x]  Cleanup widget classes #2164 @askvortsov1
  * [x]  NotificationGrid storing switches #2183  @askvortsov1
  * [x]  `CommentPost` stores `PostUser` #2184 @askvortsov1
  * [x]  NotificationPage/NotificationDropdown/NotificationList #2185 @askvortsov1
* [x] Upgrade / fix bundled extensions after state changes
* [x] Configure Typescript
* [x] Upgrade core to Mithril 2.0 (#2255, based off #2126)
* [x] Upgrade all bundled extensions to Mithril 2.0
  - [x] akismet - https://github.com/flarum/akismet/pull/15
  - [x] approval - https://github.com/flarum/approval/pull/21
  - [x] auth-facebook - obsolete with #2006
  - [x] auth-github - obsolete with #2006
  - [x] auth-twitter - obsolete with #2006
  - [x] bbcode - no JS, no PR necessary
  - [x] embed - https://github.com/flarum/embed/pull/10
  - [x] emoji - https://github.com/flarum/emoji/pull/29
  - [x] flags - https://github.com/flarum/flags/pull/29
  - [x] likes - https://github.com/flarum/likes/pull/22
  - [x] lock - https://github.com/flarum/lock/pull/23
  - [x] markdown - https://github.com/flarum/markdown/pull/19
  - [x] mentions - https://github.com/flarum/mentions/pull/52
  - [x] pusher - https://github.com/flarum/pusher/pull/23
  - [x] statistics - https://github.com/flarum/statistics/pull/12
  - [x] sticky - https://github.com/flarum/sticky/pull/19
  - [x] subscriptions - https://github.com/flarum/subscriptions/pull/32
[...]

PR #2160, opened 2020-05-10
Extract PostStream state

### LINKED ISSUE #2001: PostStream scroll bug
## Bug Report

When refreshing the discussion-page in chrome sometimes it jumps 10 number forward or back to a different post. 

image

Setting a timeout of 200ms for the scrollToNumber fixed the issue for me.

### LINKED ISSUE #2086: Link to post scrolls to the previous post

askvortsov1:
**Fixes Part of #1821, #2144**
Fixes #2001 
Fixes #2086

Extract PostStream state out of PostStream, PostStreamScrubber.

- Can we get rid of the positionChanged event sent out in PostStream?

**Confirmed**

### DIFF
--- a/js/src/forum/components/DiscussionPage.js
+++ b/js/src/forum/components/DiscussionPage.js
@@ -32,20 +32,7 @@ export default class DiscussionPage extends Page {
      */
     this.near = m.route.param('near') || 0;
 
-    const preloadedDiscussion = app.preloadedApiDocument();
-    if (preloadedDiscussion) {
-      // We must wrap this in a setTimeout because if we are mounting this
-      // component for the first time on page load, then any calls to m.redraw
-      // will be ineffective and thus any configs (scroll code) will be run
-      // before stuff is drawn to the page.
-      setTimeout(this.show.bind(this, preloadedDiscussion), 0);
-    } else {
-      const params = this.requestParams();
-
-      app.store.find('discussions', m.route.param('id').split('-')[0], params).then(this.show.bind(this));
-    }
-
-    m.lazyRedraw();
+    this.load();
 
     // If the discussion list has been loaded, then we'll enable the pane (and
     // hide it by default). Also, if we've just come from another discussion
@@ -138,6 +125,26 @@ export default class DiscussionPage extends Page {
     if (this.discussion) {
       app.setTitle(this.discussion.title());
     }
+  }
+
+  /**
+   * Load the discussion from the API or use the preloaded one.
+   */
+  load() {
+    const preloadedDiscussion = app.preloadedApiDocument();
+    if (preloadedDiscussion) {
+      // We must wrap this in a setTimeout because if we are mounting this
+      // component for the first time on page load, then any calls to m.redraw
+      // will be ineffective and thus any configs (scroll code) will be run
+      // before stuff is drawn to the page.
+      setTimeout(this.show.bind(this, preloadedDiscussion), 0);
+    } else {
+      const params = this.requestParams();
+
+      app.store.find('discussions', m.route.param('id').split('-')[0], params).then(this.show.bind(this));
+    }
+
+    m.lazyRedraw();
   }
 
   /**
--- a/js/src/forum/components/DiscussionPage.js
+++ b/js/src/forum/components/DiscussionPage.js
@@ -109,7 +109,7 @@ export default class DiscussionPage extends Page {
                     <ul>{listItems(this.sidebarItems().toArray())}</ul>
                   </nav>
                   <div className="DiscussionPage-stream">
-                    {PostStream.component({ state: this.stream, positionHandler: this.positionChanged.bind(this) })}
+                    {PostStream.component({ discussion, state: this.stream, positionHandler: this.positionChanged.bind(this) })}
                   </div>
                 </div>,
               ]
@@ -267,6 +267,7 @@ export default class DiscussionPage extends Page {
     items.add(
       'scrubber',
       PostStreamScrubber.component({
+        discussion: this.discussion,
         state: this.stream,
         className: 'App-titleControl',
       }),
--- a/js/src/forum/components/DiscussionPage.js
+++ b/js/src/forum/components/DiscussionPage.js
@@ -30,9 +30,22 @@ export default class DiscussionPage extends Page {
      *
      * @type {Integer}
      */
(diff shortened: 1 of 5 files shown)


js/src/forum/components/DiscussionPage.js
```diff
@@ -88,6 +89,24 @@ export default class DiscussionPage extends Page {
   view() {
     const discussion = this.discussion;
 
+    let content;
+
+    if (discussion) {
+      const stream = new PostStream({ state: this.streamState });
+      stream.on('positionChanged', this.positionChanged.bind(this));
```
datitisev, line 96: Won't this add the event handler oh every redraw?

js/src/forum/components/DiscussionPage.js
```diff
@@ -88,6 +89,24 @@ export default class DiscussionPage extends Page {
   view() {
     const discussion = this.discussion;
 
+    let content;
+
+    if (discussion) {
+      const stream = new PostStream({ state: this.streamState });
```
datitisev, line 95: I don't understand the point of a state class if it's not accessible at all.

I still see stream.render()...  Isn't the point of the states to avoid stuff like this?

js/src/forum/components/RenameDiscussionModal.js
```diff
@@ -57,7 +57,7 @@ export default class RenameDiscussionModal extends Modal {
         .save({ title })
         .then(() => {
           if (app.viewingDiscussion(this.discussion)) {
-            app.current.stream.update();
+            app.current.streamState.update();
```
datitisev, line 60: Are we still storing the stream then?

js/src/forum/components/DiscussionPage.js
askvortsov1: It will, I wasn't sure of a better way to do this. Ideally, that event would not be coming from the component class at all, but I wasn't sure of the best way to decomponentize it.

js/src/forum/components/DiscussionPage.js
askvortsov1: The state class IS accessible, through app.current.streamState. It'll be moved out to a more appropriate location in the app.current refactor.

As per this part, if we can figure out how to get rid of the annoying event coming from PostStream, we can safely redo this with the utmost best practices.

js/src/forum/components/RenameDiscussionModal.js
askvortsov1: No, we're storing the state.

datitisev: Then why not just keep it as `app.current.stream`?

js/src/forum/components/DiscussionPage.js
datitisev: Oh, I see. This is the component. Yeah, the point of switching to states is also to avoid doing `component.render()`. Why can't we pass the event handler below through a prop instead of an event?

Also - the positionChanged method doesn't refer to DiscussionPage at all - it could perhaps be moved into the post stream state ? Not sure what the best approach is here in terms of extensibility.

js/src/forum/components/RenameDiscussionModal.js
askvortsov1: Good point; we'll probably eventually move all the states into app.state or something with the current values acting as an alias, but till then, no benefit in unnecesssary breaks.

js/src/forum/components/DiscussionPage.js
askvortsov1: I passed the event handler for now. Moving around positionChanged isn't necessary right now, I'd rather do that later. It doesn't add THAT much clutter, and I'd prefer to keep stuff that changes url/title/etc in the page class for now.

js/src/forum/components/DiscussionPage.js
askvortsov1: No longer relevant as it's passed via a prop now.

clarkwinkelmann:
I'm mostly ok with this solution, but I'm thinking whether it wouldn't make sense to maybe keep the stream as a single class?

If we simply made the stream into a non-component class, that has just a `render()` method that outputs vnodes. The only thing that would need moving are the stuff in `config()`. Those could become a start and stop method that are called from the DiscussionPage component. And the things in `init` would move to the constructor instead. It would no longer be a component, but just a helper class to generate the output of the DiscussionPage.

It would make for a way simpler diff, and in my opinion this is a valid way to use Mithril (very similar to what I was suggesting with AlertManager or ModalManager).

clarkwinkelmann:
Could you avoid force-pushing during the review phase? It makes it completely impossible to see what was changed since last review :sweat_smile: :sweat_smile: :sweat_smile:

askvortsov1:
Sorry, that was a rebase to avoid conflicts with other merged things.

js/src/forum/states/PostStreamState.js
```diff
@@ -0,0 +1,376 @@
...
+  /**
+   * Reset the stream so that a specific range of posts is displayed. If a range
+   * is not specified, the first page of posts will be displayed.
+   *
+   * @param {Integer} [start]
+   * @param {Integer} [end]
+   */
+  reset(start, end) {
+    this.visibleStart = start || 0;
+    this.visibleEnd = this.sanitizeIndex(end || this.constructor.loadCount);
```
w-4, line 139: Since doing a major rewrite, might want to also change visibleEnd to:
```
this.visibleEnd = this.sanitizeIndex(end || this.constructor.loadCount - 1)
```

Logic:
Suppose loadCount is 1, visibleStart = 0 and visibleEnd should be the same post 0.

Would need to also adjust all the code that references visibleEnd - I can send you a pull request for that if you want.

clarkwinkelmann:
I'm really not sure how to handle review on this. The diff is very complicated to read.

If it works as before I'd say go ahead and we can always fix stuff later.

PR #2161, opened 2020-05-11
Extract Composer state

### LINKED ISSUE #1821: Frontend Framework Rewrite

askvortsov1:
**See #1821, #2144**

- Move state and public API of Composer, ComposerBody (+ inheriting classes), and TextEditor into non-component ComposerState class.
- Currently working on getting tags and mentions to work with these changes
- It seems that with the BC layer, the Tags and Mentions bundled extensions (which are the ones that most modify the composer/related stuff) are minimally broken, if at all.

- Should the direct DOM manipulations in ComposerState be moved back into the component classes and linked with events?
- Should the opposite be done?
- Am I being excessive with use of events?
- Should we provide some helper methods as a public api for ComposerState.fields?
- How else can we make these changes minimally destructive to extensions? One idea I had was that if app.composer.load is passed a component instance, we could try to manually extract out the class and state, and proceed as usual in a temporary BC layer. The same could be done for modals and alerts.

**Confirmed**


js/src/forum/components/ReplyComposer.js
```diff
@@ -68,10 +58,16 @@ export default class ReplyComposer extends ComposerBody {
    * @return {Object}
    */
   data() {
-    return {
-      content: this.content(),
-      relationships: { discussion: this.props.discussion },
-    };
+    const data = super.data();
+    data.relationships = { discussion: this.props.discussion };
```
datitisev, line 62: What if `data.relationships` isn't an empty object? You are deleting relationship data then.

js/src/common/utils/subclassOf.js
```diff
@@ -0,0 +1,6 @@
+/**
+ * Check if class A is a subclass of class B.
+ */
+export default function subclassOf(A, B) {
+  return A == B || A.prototype instanceof B;
```
datitisev, line 5: I think we want to check that `A` isn't falsy here before continuing, to reduce code duplication & random errors that originate here.

```suggestion
  return A && (A == B || A.prototype instanceof B);
  ```
```

askvortsov1: Good point!

js/src/forum/components/ReplyComposer.js
askvortsov1: Not necessarily. super.data() returns a new object, not a saved one. We could do `const data = {...super.data()}` instead though to be totally safe in case of future changes.

askvortsov1:
> I think we should have a ComposerState `key` or something (like my rewrite Alert state) and use that as an attribute or class instead of a generic `js-TextEditor` because

I wouldn't be opposed to this, but iirc, the concern was adding unnecessary bloat.

> 1. What's the point of adding `js-TextEditor` to `TextEditor`? They both point to the same elements and could target multiple ones

The goal is to make it clear that js-TextEditor concerns javascript/functionality, not style. So someone could feasibly make changes to / remove / use elsewhere the TextEditor class, but not so much for TextEditor. I would probably prefer a `TextEditor-u8dsa7f98a`-esque ID though, as you suggest above with your comment about the key. @clarkwinkelmann iirc you didn't like the key idea, what are your thoughts on this?

js/src/forum/components/ReplyComposer.js
datitisev: If something extends the Composer `data` method, and returns relationship data there, you are getting rid of it. That's what I mean.

askvortsov1: That seems to be the default current behavior though? We could try to merge the objects with `Object.assign` perhaps?

datitisev: I don't think Object.assign works that way. The current behavior doesn't have a parent data model that you are calling, so if you start to call it, you should use everything it returns _OR_ make it known that the relationship data is going to be thrown out. This could cause a lot of wasted time on debugging if you just replace the relationships.

askvortsov1: Would merging the relationships objects be good?

datitisev: Hm, perhaps assigning data.relationships would work. Or `if (!data.relationships) data.relationships = {}`.

js/src/forum/ForumApplication.js
```diff
@@ -124,9 +131,10 @@ export default class ForumApplication extends Application {
    */
   composingReplyTo(discussion) {
     return (
-      this.composer.component instanceof ReplyComposer &&
-      this.composer.component.props.discussion === discussion &&
-      this.composer.position !== Composer.PositionEnum.HIDDEN
```
askvortsov1, line 129: Move this to ComposerState

js/src/forum/components/CommentPost.js
```diff
@@ -72,7 +73,7 @@ export default class CommentPost extends Post {
   }
 
   isEditing() {
-    return app.composer.component instanceof EditPostComposer && app.composer.component.props.post === this.props.post;
+    return app.composer.bodyClass && subclassOf(app.composer.bodyClass, EditPostComposer) && app.composer.bodyProps.post === this.props.post;
```
franzliedke, line 76: We could encapsulate this as something like `app.composer.bodyIs(componentClass, props)`.

askvortsov1:
Should we wrap bodyClass and props within a JS object like we're doing with alerts and modals?

datitisev:
> Should we wrap bodyClass and props within a JS object like we're doing with alerts and modals?

Can you elaborate? Not sure what you mean here.

askvortsov1:
Well, we wrapped modalClass and modalAttrs into a ModalState class? Perhaps we should do something similar here for consistency so that composers are shown by passing in a ComposerBodyState? (but with different naming)

franzliedke:
That `.js-TextEditor` worries me, too - not so much because of the naming (that `.js-` convention is a pretty wide-spread convention with the meaning that @askvortsov1 explained), but rather because we're reaching across all kinds of boundaries to a DOM element owned by a completely different part of our app.

Unfortunately, I have no particularly good alternative in mind right now. The problem with JavaScript UIs is that there are so many different *types* of states. VDOM state (components and props), animation state, actual DOM state, cursor/focus state etc.

And for some of these, it doesn't seem necessary to keep VDOM and DOM state in sync at all times, e.g. when the user is typing lots of words in the composer.

Not sure where I'm going with this...

askvortsov1:
The best alternative idea I have for the js-TextEditor is providing a unique ID. That would make ownership clearer?

js/src/forum/components/ReplyComposer.js
askvortsov1: Improved, is this better?

datitisev:
Can we not just use the `ref` attribute and set it to a state property? Something like `ref={el => this.state.el = el}`? (I can't remember if this is exactly how it works)

askvortsov1:
> Can we not just use the `ref` attribute and set it to a state property? Something like `ref={el => this.state.el = el}`? (I can't remember if this is exactly how it works)

I think this might be a react/vue thing, doesn't seem to do anything. We could, however, pass the texteditor dom element to the state in the config method?

js/src/forum/components/DiscussionComposer.js
```diff
@@ -21,7 +21,8 @@ export default class DiscussionComposer extends ComposerBody {
      *
      * @type {Function}
      */
-    this.title = m.prop('');
+    this.state.fields.title = m.prop(this.state.fields.title ? this.state.fields.title() : '');
```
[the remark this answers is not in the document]
askvortsov1: The thinking here was mainly that this way, the title stored in the state and the title stored here would always be identical, and setters would be 2x lighter.

js/src/forum/components/ComposerBody.js
```diff
@@ -22,45 +22,53 @@ import ItemList from '../../common/utils/ItemList';
...
       <div className={'ComposerBody ' + (this.props.className || '')}>
         {avatar(this.props.user, { className: 'ComposerBody-avatar' })}
         <div className="ComposerBody-content">
           <ul className="ComposerBody-header">{listItems(this.headerItems().toArray())}</ul>
-          <div className="ComposerBody-editor">{this.editor.render()}</div>
+          <div className="ComposerBody-editor">
+            {TextEditor.component({
+              submitLabel: this.props.submitLabel,
+              placeholder: this.props.placeholder,
+              disabled: this.loading | this.disabled,
```
franzliedke, line 65: Is that the operator we want? Or should it be `||`?

datitisev: Looks like both work. Preferably `||` because that's the behavior we want.

- `false | false` => `0`
- `false | true` => `1`
- `true | true` => `1`

js/src/forum/states/ComposerState.js
```diff
@@ -0,0 +1,341 @@
...
+
+  /**
+   * Determine whether we are on mobile.
+   *
+   * @return {Boolean}
+   * @public
+   */
+  onMobile() {
+    // 767 is the mobile screen cutoff defined in the less variables file
+    return window.innerWidth <= 767;
```
w-4, line 208: It's possible to expose the variable in less:
```css
:root {
 --phone: @phone;
}   
```

I think it'd be also better to move that method into global state, since it's something other components might use aswell. 

Inside ForumApplication mount:
```js
const styles = getComputedStyle(document.documentElement)
const phone = styles.getPropertyValue('--phone')

this.isMobile = function () {
   return window.matchMedia(phone).matches
}
```

askvortsov1: That's a really good point! This particular code is actually something we weren't happy with during internal review, and I like what you're proposing a lot more! @franzliedke how would you feel about this?

datitisev: I forgot about CSS variables when we discussed this! They should be supported by every browser that supports our JS syntax (see https://caniuse.com/#feat=css-variables).

franzliedke: @w-4 Great idea, thank you very much! And :+1: for daring to venture into this code review, we appreciate it.

I've applied this idea, with a twist: instead of exposing and then matching the media query, I now expose a CSS variable that is changed in CSS based on the media query and returns the mode. Also added this to the `Application` base class.

PR #2162, opened 2020-05-12
Extract modal state

askvortsov1:
**Fixes Part of #1821, #2144**

- Move modal state out from modal manager component

- Anything I missed?

**Confirmed**


js/src/common/states/ModalState.js
```diff
@@ -0,0 +1,94 @@
+import evented from '../utils/evented';
+
+import Modal from '../components/Modal';
+
+class ModalState {
```
clarkwinkelmann, line 5: For consistency with the Alert stuff, I'd call this `ModalManagerState`

js/src/common/states/ModalState.js
```diff
@@ -0,0 +1,94 @@
...
+      throw new Error('The ModalManager component can only show Modals');
+    }
+
+    clearTimeout(this.hideTimeout);
+
+    this.showing = true;
+    this.modalClass = modalClass;
+    this.modalProps = modalProps;
+
+    if (app.current) app.current.retain = true;
```
clarkwinkelmann, line 27: This will have to go, but I think this will be taken care of at a later stage

js/src/common/components/Modal.js
```diff
@@ -9,13 +9,25 @@ import Button from './Button';
...
+    this.state = this.props.state;
+
     /**
      * An alert component to show below the header.
      *
      * @type {Alert}
      */
     this.alert = null;
+
+    this.state.modalOnHide = this.onhide.bind(this);
```
clarkwinkelmann, line 28: Since `this.state` is evented, couldn't we just use listeners here ?

js/src/common/states/ModalState.js
askvortsov1: I'll remove it now, as it's no longer necessary as a result of the state change.

askvortsov1:
Should app.modal.show require a ModalState object? It would be more in line with what we're doing for Alerts, but would also require a bunch more imports...

datitisev:
Not sure if I like having `getClass` and `getAttrs` methods (instead of directly accessing them), but we could perhaps add a method to the ModalState to check for the class prototype instance of (as it's used in a few places).

askvortsov1:
In later versions, I anticipate "ModalState" staying relatively constant, just that it wouldn't be stored in "ModalManagerState". That being said, the only change I'd make at this point is rename it to just "Modal" and put it in an `instances` directory (or some other naming) as per Franz's comment, don't have the link to that though.

js/src/common/states/ModalManagerState.js
```diff
@@ -0,0 +1,88 @@
...
+   *
+   * @param {Modal} component
+   * @public
+   */
+  show(modalClass, attrs) {
+    const modal = new ModalState(modalClass, attrs);
+
+    if (!(modal.getClass().prototype instanceof Modal)) {
+      throw new Error('The ModalManager component can only show Modals');
+    }
```
franzliedke, line 22: Is that supposed to intentionally break old code (expecting to call `show` on the `ModalManager` component), to make developers aware they need to upgrade?

js/src/common/states/ModalState.js
```diff
@@ -0,0 +1,14 @@
...
+  }
+
+  getAttrs() {
+    return this.attrs;
+  }
+
+  getClass() {
+    return this.modalClass;
+  }
+}
```
franzliedke, line 14: Hmm, this is exactly the same as the class (and the same purpose) as the `PageState` in #2156. And very, very similar to the `AlertState` in #2163.

What is this really? Maybe a "frozen component"?

askvortsov1: It's to allow different component classes for different modals. wdym by frozen component?

js/src/common/states/ModalManagerState.js
askvortsov1: This was in Modal Manager's show before. That being said, we could either do this, put in a BC layer (if a modal component is sent in, package it's constructor and props in a modal POJO), or both?

clarkwinkelmann: I think we should check there's a nice error message for the breaking change, like "must pass a class and not an instance". We mostly agreed not trying to do BC I believe.

js/src/common/components/Modal.js
```diff
@@ -57,7 +70,7 @@ export default class Modal extends Component {
    *
    * @return {Boolean}
    */
-  isDismissible() {
+  static isDismissible() {
```
askvortsov1, line 73: If this is static, why make it a method at all and not just a property?

js/src/common/states/ModalManagerState.js
clarkwinkelmann: After doing a bit of digging it seems there is no "real" way of checking whether a variable contains an object or a class given how javascript works.

The `protoype instanceof` check above will work against both an object instance or the class itself.

I really think we need an error message for instances as this will be common with code updates.

It seems like we could check for the presence of one of the known methods.

So I'd suggest:

    if (!(type.prototype instanceof Modal)) {
      throw new Error('The ModalManager can only show Modals');
    }
    if (type.init) {
      throw new Error('The type parameter must be a modal class, not a modal instance');
    }

js/src/common/states/ModalManagerState.js
```diff
@@ -0,0 +1,43 @@
...
+  constructor() {
+    this.modal = null;
+  }
+
+  /**
+   * Show a modal dialog.
+   *
+   * @public
+   */
+  show(type, attrs) {
```
clarkwinkelmann, line 13: If you agree with my comment about renaming `type` to `component` or `componentClass` in https://github.com/flarum/core/pull/2163#discussion_r446678767 I think we should use an identical naming convention here.

datitisev: I assume the parameters here are switched because `type` in AlertManager#show is optional, right?

clarkwinkelmann: I noticed that as well. Here, `attrs` is optional. Whether with Alerts, `type` is optional.

So I think switching them makes the most sense.

js/src/common/components/Modal.js
franzliedke: Because it was a method before and allow(ed) for custom logic. :shrug: 

Not sure if that was used anywhere... too lazy to come up with arbitrary use-cases. :laughing:

datitisev: You could perhaps make it dismissible only after an input was filled out and/or form submitted. We can't do that with a static method though, don't have access to the current instance data...

js/src/common/states/ModalManagerState.js
franzliedke: I don't understand your suggestion here, @clarkwinkelmann, and your response seems to be missing a word. :thinking: 

Are you saying that having them switched (when compared with alerts) makes sense, or are you suggesting to switch them again?

js/src/common/components/Modal.js
franzliedke: Fair point. Should be okay then.

js/src/common/states/ModalManagerState.js
clarkwinkelmann: Sorry for the confusion. I think the current version (switched) is good. We place the required arguments first and optional arguments second.

The requested change of this thread was only to rename, not reorder.

PR #2163, opened 2020-05-12
Extract Alert State, AlertManagerState

### LINKED ISSUE #1821: Frontend Framework Rewrite

askvortsov1:
**See #1821, #2144**

- Extract Alert State for each alert
- Extract AlertManagerState for alerts in general
- Mostly a copy paste of work done by David in the draft mithril rewrite branch.

- Anything I missed?
- Should we be concerned that we're storing component instances in the controls? We could take care of that as well (store tuples of classes, props) if needed

**Confirmed**


js/src/common/Application.js
```diff
@@ -325,20 +331,20 @@ export default class Application {
...
             </Button>,
           ],
-        });
+        };
 
         try {
           options.errorHandler(error);
         } catch (error) {
-          this.alerts.show(error.alert);
+          this.alerts.show(error.alertProps);
```
clarkwinkelmann, line 347: I think there might be a missing assignment to `error.alertKey` here

js/src/common/states/AlertManagerState.js
```diff
@@ -0,0 +1,44 @@
...
+
+  /**
+   * Dismiss an alert.
+   */
+  dismiss(keyOrState) {
+    if (!keyOrState) return;
+
+    const key = keyOrState instanceof AlertState ? keyOrState.key : keyOrState;
+
+    let index = this.activeAlerts.indexOf(this.activeAlerts.filter((a) => a.key == key)[0]);
```
clarkwinkelmann, line 28: You could use `.findIndex` here

js/src/forum/components/ChangeEmailModal.js
```diff
@@ -121,8 +121,8 @@ export default class ChangeEmailModal extends Modal {
   }
 
   onerror(error) {
-    if (error.status === 401) {
-      error.alert.props.children = app.translator.trans('core.forum.change_email.incorrect_password_message');
+    if (error.status === 403) {
```
clarkwinkelmann, line 124: Why is the error code changing here? Was it incorrect?

js/src/common/components/AlertManager.js
```diff
@@ -7,20 +7,16 @@ import Alert from './Alert';
...
   }
 
   view() {
     return (
       <div className="AlertManager">
-        {this.components.map((component) => (
-          <div className="AlertManager-alert">{component}</div>
+        {this.state.activeAlerts.map((alert) => (
+          <div className="AlertManager-alert">
+            <Alert state={alert} ondismiss={this.state.dismiss.bind(this.state, alert.key)} />
```
clarkwinkelmann, line 18: Why not pass the props directly as `<Alert {...props} />` ? That would also remove the need for the additional code inside of `Alert.initProps`.

js/src/common/states/AlertState.js
```diff
@@ -0,0 +1,6 @@
+export default class AlertState {
+  constructor(props = {}, key = Date.now()) {
```
clarkwinkelmann, line 2: How safe is it to use the date for a unique identifier ? Isn't there a risk of collision if an extension opens two alerts with two consecutive lines of code ?

Wouldn't it be safer to use an incremental ID (the counter value could be in the manager state), or one of the random methods of javascript ?

js/src/forum/components/ChangeEmailModal.js
askvortsov1: Yep, 403 is what's returned when the password is wrong. On second thought, the backend should probably return 401 as that's a more accurate description of the issue. Fixed.

js/src/common/components/AlertManager.js
askvortsov1: This wouldn't work because we'd need to do a deep clone (when I try, props.children comes up as an empty array), but I can simplify down the initProps

askvortsov1:
> I still think it might be a good idea to give the ability to customize the Alert component. What about adding a `component` property to `AlertState`, which defaults to `Alert`. Your code can already handle passing a custom `AlertState` to `show()`, so all someone would need to customize the component, is to create the `AlertState` object themselves. This wouldn't overload the `show()` method in any way.

After a few other changes to simplify the code/interface, it doesn't slot in 100% as cleanly, so I just put it in as an optional param to alerts.show(). How's this look?

src/Api/Controller/UpdateUserController.php
```diff
@@ -62,7 +62,7 @@ protected function data(ServerRequestInterface $request, Document $document)
             $password = Arr::get($request->getParsedBody(), 'meta.password');
 
             if (! $actor->checkPassword($password)) {
-                throw new PermissionDeniedException;
+                throw new NotAuthenticatedException;
```
datitisev, line 65: This should not be in this PR?

askvortsov1: Ack I thought I put that into a different branch...

askvortsov1:
> I don't like that `alertAttrs` and `alertKey` are saved separately.

Why not? The 2 aren't inherently related. IMO, the mechanism of keeping track of the currently active alerts should be in AlertManagerState, which is what is done here.

> This also means that extensions that extend Alert (https://github.com/FriendsOfFlarum/terms/blob/572025438f6795e96ce9e8b854a1c31f69b485d1/js/src/forum/components/UpdateAlert.js) will have to do `app.alerts.show({ children: <extended alert>.component() })` which seems like a waste.

Not necessarily: you can pass in a class subclassing Alert into `show`, so the structure remains constant. We might want to switch the order of attrs and Alert around though so that it matches what we have for Modal.

> Is it because we don't want to have DOM in the states? If so, i think we should move past that because we have other state PRs directly modifying the DOM, so there's no point in differentiating that.

As of right now, I view the states as 2 things:

- A bucket for data
- A public API for modifying other components.

In the future, when we adopt a proper state changing mechanism such as redux, these roles will be further split up. However, for now, our priority is Mithril 2.0 and not storing components, so this intermediate step is fine. For alerts specifically, ideally I'd like to have a class for each type of alert (like we do with Modals), which is where the children and control elements would be held. However, that's not worth delaying the rewrite over.

My concern isn't avoiding anything to do with DOM in states. My concern is when states start to act like components in terms of containing JSX / business logic. I'd like to restrict states to dealing with data and animating stuff around.

> I much prefer keeping the old syntax of `app.alerts.show(Alert.component // new AlertState({ ... })`.

I don't think that passing the component instance directly was supported in the other PR, just the attrs dict or the state object?

js/src/common/Application.js
```diff
@@ -367,8 +373,8 @@ export default class Application {
    * @param {string[]} [formattedError]
    * @private
    */
-  showDebug(error, formattedError) {
-    this.alerts.dismiss(this.requestError.alert);
+  showDebug(error) {
```
datitisev, line 376: `formattedError` should not have been removed. It is still used in the same method.

datitisev:
> I don't think that passing the component instance directly was supported in the other PR, just the attrs dict or the state object?

It was supported, just not required (I think). This was to allow classes extending AlertState to be passed.

> A bucket for data

If the state is a bucket for data, why wouldn't you pass it as you pass the attribute data? That way you have the whole thing and can store it if necessary (why, I don't know, but in case it's needed for some reason).

The key is referencing a specific alert state. Why not just have the alert state stored with the key, instead of storing them separately when you are saving them in things like `requestError`?

> Why not? The 2 aren't inherently related. IMO, the mechanism of keeping track of the currently active alerts should be in AlertManagerState, which is what is done here.

What do you mean, not inherently related?  They both relate to a specific alert state?

datitisev:
I think we should get the opinions of other extension creators on how the data should be stored/accessible. Might shed some new light on this. I know it's not exactly important, but accessing the alert data should be through `requestError.alert` and not have two separate keys for references to the same object, essentially.

js/src/common/Application.js
askvortsov1: Shoddy merge conflict resolution on my part, sorry about that.

askvortsov1:
> It was supported, just not required (I think). This was to allow classes extending AlertState to be passed.

What's the benefit of extending AlertState? If we keep it minimal, there's not really a need.



> If the state is a bucket for data, why wouldn't you pass it as you pass the attribute data? That way you have the whole thing and can store it if necessary (why, I don't know, but in case it's needed for some reason).

My initial idea was to pass the entire state into Alert, iirc @clarkwinkelmann recommended passing just the `state.attrs`, as that's the only part of the AlertState that's used.

> What do you mean, not inherently related? They both relate to a specific alert state?

My initial thought in separating was to make things simpler, but storing the key in the state is just as good tbh. Any objections about going back to that mechanism?

askvortsov1:
Here's what I'm thinking in terms of next steps:

- Show accepts an AlertState instance. 
- Dismiss accepts a key.
- AlertState stores  attrs, Alert class (defaults to Alert), and the key
- AlertStates are stored in AlertManagerState via an object of keys => alertStates

askvortsov1:
As a side note, what if we required that the children props of Alert be a string (in that case, perhaps rename it to title), and if controls/other children are needed, we created a subclass of Alert? I feel that would be cleaner state-wise.

datitisev:
Don't think that's necessary _or_ a good idea. Would create extra classes for basically nothing (like our many confirmation modals 🙃)

franzliedke:
> As of right now, I view the states as 2 things:
>
> * A bucket for data
> * A public API for modifying other components.

This part I have my doubts about. `AlertState` seems to fall into the former category, `AlertManagerState` into the latter. IMO, naming these concepts differently would help a lot in understanding / teaching these concepts.

The "states", as I see them, should be of the second variant: application-wide containers (with a bit of behavior) that accept events (method calls, really) that then trigger actions / data changes across various places of the UI, followed by a redraw.

That `Alert` class is just that, an object representing an alert. A POJO (Plain Old Javascript Object). It does get confusing, because we also have an `Alert` component, but that's kinda already solved by our directory structure (even before we make larger changes to that).

---

One of my favorite programming quotes, somewhat relevant:

> So much complexity in software comes from trying to make one thing do two things.
> – Ryan Singer

askvortsov1:
My only concern would be importing Alert and Alert, but I suppose we can just use import aliases? Should POJO alert still be in states directory?

franzliedke:
I wouldn't see it in the `states` directory, no. It would be a simple value object, and should not do some of the things (change routes, trigger redraws etc.) that "states" are allowed to.

franzliedke:
Out of interest: what's the idea behind the latest commit ("Encapsulate ...")? There doesn't seem to be much encapsulation going on, especially considering that you added both getters *and* setters (in their most trivial form) to these attributes..

askvortsov1:
> Out of interest: what's the idea behind the latest commit ("Encapsulate ...")? There doesn't seem to be much encapsulation going on, especially considering that you added both getters _and_ setters (in their most trivial form) to these attributes.

While I hate to be the introductory college java course where encapsulation is taught as religion, I do think it's warranted here: if we continue as planned, the attributes might not be accessible as attributes in the future.

js/src/common/states/AlertManagerState.js
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  getActiveAlerts() {
+    return this.activeAlerts;
+  }
+
+  /**
+   * Show an Alert in the alerts area.
+   */
+  show(attrs, type) {
```
clarkwinkelmann, line 14: May I suggest renaming `type` with `component` or even `componentClass` ?

I find it confusing to have both a parameter and an attr/prop named `type`. People might confuse the two.

franzliedke: People only supply the value for the argument, though - argument names aren't part of the public API.

Are you suggesting to change the property of the alert POJO as well? If not, I don't mind changing the parameter name back - something like `componentClass` was the original proposal anyway, IIRC.

clarkwinkelmann: I'd say both the argument and the POJO attribute to reduce confusion to the maximum.

The attribute of the POJO should indeed never be visible to the developers but for our own sanity I'd argue it would make sense to call it `componentClass` so there's no ambiguity when reviewing the code.

The name of the argument might not be part of the public API per say but it might be visible in IDEs that show autocompletion and documentation, even more so once we have proper typescript definitions shipped.

franzliedke: I don't care enough, whatever is clearer to you guys. :wink:

Issue #1934, opened 2019-11-15
Some email notifications are forced into English


Pherinos:
## Bug Report

I have identified some PHP files where the text is not translatable because forced into English in the code.

### **Files**

- "_[New Post]_ " in vendor/flarum/subscriptions/src/Notification/NewPostBlueprint.php

```
 66     /**
 67      * {@inheritdoc}
 68      */
 69     public function getEmailSubject()
 70     {
 71         return '[New Post] '.$this->post->discussion->title;
 72     }

```

- " _mentioned you in_ " in vendor/flarum/mentions/src/Notification/UserMentionedBlueprint.php

```
64     /**
 65      * {@inheritdoc}
 66      */
 67     public function getEmailSubject()
 68     {
 69         return "{$this->post->user->display_name} mentioned you in {$this->post->discussion->title}";
 70     }

```

- " _replied to your post in_ " in vendor/flarum/mentions/src/Notification/PostMentionedBlueprint.php

```
 72     /**
 73      * {@inheritdoc}
 74      */
 75     public function getEmailSubject()
 76     {
 77         return "{$this->reply->user->display_name} replied to your post in {$this->post->discussion->title}";
 78     }

```
### **Environment**
- Flarum version: 0.1.0-beta.10
- Website URL: https://latoilemysterieuse.a2co.org
- Webserver: Apache
- Hosting environment: private physical server
- PHP version: 7.3.11

```
Flarum core 0.1.0-beta.10
PHP version: 7.3.11
Loaded extensions: Core, date, libxml, openssl, pcre, zlib, bz2, calendar, ctype, hash, filter, ftp, gettext, gmp, SPL, iconv, pcntl, readline, Reflection, session, standard, SimpleXML, sockets, mbstring, tokenizer, xml, mysqlnd, bcmath, curl, dba, dom, enchant, fileinfo, gd, imagick, imap, intl, json, ldap, exif, mysqli, odbc, PDO, pdo_mysql, PDO_ODBC, pdo_pgsql, pdo_sqlite, pgsql, Phar, posix, pspell, redis, soap, sodium, sqlite3, sysvmsg, sysvsem, sysvshm, tidy, xmlreader, xmlrpc, xmlwriter, xsl, zip, ionCube Loader, Zend OPcache
+---------------------------------+-----------------+--------+
| Flarum Extensions               |                 |        |
+---------------------------------+-----------------+--------+
| ID                              | Version         | Commit |
+---------------------------------+-----------------+--------+
[...]

PR #2169, opened 2020-05-17
Inject Url Generator and Translator Interface into notification mailer

askvortsov1:
**Allows Fix For #1934**

Allows us to get rid of https://github.com/flarum/mentions/blob/e054158ae5917d4edadb3c996e4c9ae992ab78e0/views/emails/postMentioned.blade.php#L5
moving one step closer to deprecating helpers

- Inject an instance of translator into notification views
- Inject an instance of UrlGenerator into notification views

What else should we inject? SettingsRepository came to mind, as well as possibly an optional associative array of arguments provided when calling `$mailer->send` One important thing to note is that we want to avoid an anti-pattern of business logic in the email views, so I'd stick to the essentials for now.

**Confirmed**

### DIFF
--- a/src/Notification/NotificationMailer.php
+++ b/src/Notification/NotificationMailer.php
@@ -9,9 +9,11 @@
 
 namespace Flarum\Notification;
 
+use Flarum\Http\UrlGenerator;
 use Flarum\User\User;
 use Illuminate\Contracts\Mail\Mailer;
 use Illuminate\Mail\Message;
+use Symfony\Component\Translation\TranslatorInterface;
 
 class NotificationMailer
 {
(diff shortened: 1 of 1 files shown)


franzliedke:
@askvortsov1 Aren't these already available in views thanks to this?

https://github.com/flarum/core/blob/3b39c212e0fef7522e7d541a9214ff3817138d5d/src/Frontend/FrontendServiceProvider.php#L68-L71

askvortsov1:
Isn't that only for page views / views returned as the result of the frontend?

franzliedke:
If the mailer component uses the view component as well, it should be likewise affected.

askvortsov1:
You are 100% correct. Sorry, I should have checked before implementing it. Reverted in https://github.com/flarum/core/commit/4ee6d6fd885bc98febd5f9a9ac23e93d60a4c987

franzliedke:
It was very easy to miss. Stumbled across those lines while changing something completely unrelated, just after having seen this PR a couple of hours earlier. :shrug:

askvortsov1:
Although I gotta say, does make me curious why these weren't translated in the first place...

franzliedke:
Probably because it feels pointless to have a file only consisting of `{{ trans() }}` calls. There might have been thoughts about providing complete files in language packs. But as soon as HTML gets involved, that doesn't work well. :shrug:  So I'm glad we're moving forward here.

Issue #2091, opened 2020-03-28
Replacement options for dispatcher->until() 


askvortsov1:
## Functionality Discussion

**Is your feature request related to a problem? Please describe.**
Certain extendable features currently use ->until() with. While this has some benefits, there are also several significant issues, namely:
1. We are looking to transition away from non-domain events in favor of our more robust extender API
2. Only the first extension to return a non-null value is considered. This has security implications, as in https://github.com/flarum/core/issues/1832. In my opinion, permissions should NOT depend on extension boot order (which is a separate issue).
3. There is no way for an extension to override a value returned by another extension. 

I've compiled a list of all instances where ->until() is used here: https://github.com/flarum/core/pull/2015#issuecomment-592825650

**Describe the solution you'd like**
I have come up with 2 possible ideas this far. In my opinion, at this stage, the actual implementation is less important than deciding how to handle this situation.

1. Consider ALL responses returned by extensions. This is inspired by Symfony's voter approach. When requesting this replacment to a `->until()` check, the code would indicate an 'evaluation mode'. Depending on the mode, return:
   a. `true` if any true values are present
   b. `false` if any false values are present
   c. whichever of `true` or `false` has more returned, with `true` returned if there is a tie
   d. whichever of `true` or `false` has more returned, with `false` returned if there is a tie

2. Extensions can return one of 4 options:
   a. $this->allow()
   b. $this->allow()
   a. $this->forceAllowallow()
   a. $this->forceDeny()

Any $this->forceDeny()s returned would override, if not present, any $this->forceAllow()s would override, if not present any $this->allow()s would override, and so on. This is my preferred approach as it's intuitive and flexible. One potential implementation is included in https://github.com/flarum/core/pull/2056

EDIT:

Some time after writing this, I realized that this would not necessarily work for ->until() methods that don't return a boolean (such as GetDisplayName). For those cases, I think that the extender should be refactored to use a driver, and a driver should be selectable in the admin configuration, like we do with mail drivers.

askvortsov1:
Franz and I had a very productive discussion today, here are my takeaways for how to procees:

0. Permissions and Policies.

[...]

PR #2170, opened 2020-05-17
Move floodgate to middleware, add extender + integration tests

askvortsov1:
**Part of #2091, #1891 **

Depends on https://github.com/flarum/core/pull/2425

- Move Floodgate out to a middleware, it now affects all API requests
- Convert current post throttling protections into flood checkers under new system
- Add floodgate extender
- Add integration tests for flooding on discussions / posts + for the extender itself

- How do we feel about the mechanism for registering flood checkers to routes / methods as done here?
- How do we feel about how returned values are handled?

NOTE: The changes to MailTest are here to prevent this PR breaking those tests due to DB contamination. We really need to wipe the DB between every test class...
Note followup: if #2172 is merged before this, I'll remove those changes.

**Confirmed**


src/Api/ApiServiceProvider.php
```diff
@@ -42,6 +42,32 @@ public function register()
...
+                [
+                    'paths' => ['*'],
+                    'methods' => ['GET', 'POST', 'PUT', 'PATCH', 'DELETE'],
+                    'callback' => function ($actor, $request) {
+                        if ($request->getAttribute('bypassFloodgate')) {
+                            return false;
+                        }
+                    }
+                ]
+            ];
```
datitisev, line 56: Should this perhaps be class instances instead? I don't particularly like the idea of using array objects for this, to be completely honest.

src/Post/PostServiceProvider.php
```diff
@@ -9,11 +9,36 @@
...
 class PostServiceProvider extends AbstractServiceProvider
 {
+    /**
+     * {@inheritdoc}
+     */
+    public function register()
+    {
+        $this->app->extend('flarum.api.floodgates', function ($floodgates) {
+            $floodgates['createPostFloodgate'] = [
+                'paths' => ['/api/posts', '/api/discussions'],
```
datitisev, line 25: i think this should use the route _names_ instead, as the paths could potentially change.

src/Extend/Floodgate.php
```diff
@@ -0,0 +1,73 @@
...
+     * @param string[] $methods: An array of methods on which to apply this checker.
+     *                           Methods should be uppercase, wildcards are NOT supported.
+     * @param string[] $callback: A closure or invokable class that contains the logic of the flood gate.
+     *                            It should return one of:
+     *                              - `false`: This marks the request as NOT flooding. It overrides all other outputs
+     *                              - `true`: This marks the request as flooding.
+     *                            All other outputs will be ignored.
+     * @return self
+     */
+    public function set(string $name, array $paths, array $methods, callable $callback)
```
datitisev, line 36: These arrays should allow for strings to be passed instead of one-element arrays (which would require no type hinting as PHP doesn't support string typings).

Or, continuing off my class idea, do something like

```php


(new Extend\Floodgate())
  ->add(
    Floodgate::build($name)
      ->routes($route1, $route2)
      ->methods($method1, $method2)
      ->handler($callback)
  )
```

src/Api/ApiServiceProvider.php
askvortsov1: I don't have strong feelings either way, arrays felt simpler to start with.

src/Post/PostServiceProvider.php
askvortsov1: That would require us to check that the route exists and map the path to a route, which only happens in DispatchRoutes. While I'd agree it'd be nicer, the extra boilerplate might not be worth it.

src/Extend/Floodgate.php
askvortsov1: Mixed feelings on the API format. I like us having the freedom to assemble / store the actual flood gate objects as we see fit, that allows us to change things without breaking it later on.

src/Post/PostServiceProvider.php
datitisev: I'm not talking about nicer. An extension could change it and then the floodgate would become useless.

src/Extend/Floodgate.php
datitisev: > allows us to change things without breaking it later on.

I feel like storing them as an array would make them break in the future more? If you want to change something, the array values would need to change. With this helper function, you can modify the internal code to adapt to the internally breaking changes ?

src/Post/PostServiceProvider.php
askvortsov1: Then the same should probably be done for the CSRF extender. We'd probably also want to turn the path -> route retrieval logic into it's own system.

datitisev: Hm, perhaps.

askvortsov1: After looking more at the code, I'm not sure how feasible this is if we want to use middleware. We would effectively need to run DispatchRoutes twice. The best way that comes to my mind of doing this is:

- Run DispatchRoutes somewhere in the middle of the middleware stack.
- Instead of directly running the controller, save it as a request attribute, alongside params and the route name
- At the end of the middleware stack, run it through the controller.

The end result is a bit messier and less direct. It also loses the advantage of rate-limiting spammers who indiscriminately check every endpoint they can, meaning that the site could be probed for which endpoints do / do not exist moreso than now.

src/Extend/Floodgate.php
askvortsov1: We also don't have any other extenders that take an object. Although I do agree that at least internally, storing in an object makes more sense than storing in an array.

src/Post/PostServiceProvider.php
askvortsov1: Implemented in https://github.com/flarum/core/pull/2425.

src/Extend/Floodgate.php
askvortsov1: But to follow up, the internal storage is just that, an implementation detail that we can tweak at any time since it's private API

src/Api/ApiServiceProvider.php
luceos: I agree with @datitisev, even though I had mixed feelings. I think it makes more sense to make code strict than vague.

src/Extend/Floodgate.php
luceos: I think I agree with @datitisev ; question on my end is what the `Floodgate` is (not the extender I hope) and what `build` returns. Maybe an instance of `Throttle`? Do we have examples from eg Laravel on how they handle this?

askvortsov1:
I've been thinking a bit more about this, and a few more questions have come up:

- Do we actually need to take care of route restrictions, or should the throttling callbacks themselves decide which routes to apply themselves to? One benefit of the latter is that a route restriction could be applied to all routes at once (E.G for 'bypassFloodgate'). But on the other hand, it's more work for floodgate checkers, and is less effective: if we bind floodgate checkers directly to routes, we don't have to cycle through ALL checkers on every request, just the ones for that route. Pehaps we could allow a single asterisk string instead of the array of routes for checkers to be applied on all routes?


- Should throttling apply only to API routes (as is done in this PR), or to all routes? One possible use case for doing this for all routes is throttling LoginController, although that IS throttled through the API since it works by calling CreateTokenController

askvortsov1:
@luceos @datitisev I updated this to drastically simplify it: throttlers are now just regular old callbacks, and are responsible for deciding which routes they do / don't want to be called on themselves (this way, wildcard routes are supported in the rare instances where needed without us having to explicitly support it.

src/Extend/ThrottleApi.php
```diff
@@ -0,0 +1,65 @@
...
+     *
+     * @param string $name: The name of the throttler.
+     * @param string[] $callback: A closure or invokable class that contains the logic of the throttler.
+     *                            It should return one of:
+     *                              - `false`: This marks the request as NOT flooding. It overrides all other outputs
+     *                              - `true`: This marks the request as flooding.
+     *                            All other outputs will be ignored.
+     * @return self
+     */
+    public function set(string $name, callable $callback)
```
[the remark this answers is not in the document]
SychO9: would also be nice to mention that the callback accepts the $request as an argument.

SychO9 (changes requested):

SychO9 (approved):
Looks good to me! Tested locally as well.

tankerkiller125 (approved):

askvortsov1:
@clarkwinkelmann wanted to ping you on this one as well given your thoughts on #2438: This one has all variables contained as properties (or public getters) of the request, but do we want to pass them in as individual arguments for additional safety (since, for instance, we might change how Actor is registered with the request object)?

clarkwinkelmann (approved):
Looking at the code, this looks good.

I don't think we should worry about splitting the actor and such into additional parameters at this time. We can always make the change later. If we make changes to request attributes, don't forget we also break user-created middlewares and controllers, so it would have a huge impact independently from what we do here.

Not for this PR - but do we want to add a default throttle to the API, with maybe a default rate for GET/HEAD and a different one for the other methods?

Should we make this feature also available to the forum frontend? It would be great for restricting login and password reset.

askvortsov1:
> Not for this PR - but do we want to add a default throttle to the API, with maybe a default rate for GET/HEAD and a different one for the other methods?

Hmm that's bordering on extension territory IMO


> Should we make this feature also available to the forum frontend? It would be great for restricting login and password reset.

Login and Password Reset are both implemented by the Forum views using our simulated API client to send requests to backend views. This isn't the case for email confirmations, although IMO we should make it so.

Then, we could have the ThrottleAPI middleware run on ApiClient requests (which shouldn't be that difficult, we just need to simulate a middleware pipe in the client), and voila, we have throttling!

I think that would make sense as a separate PR though.

PR #2172, opened 2020-05-18
Ensure consistent DB state for integration tests

askvortsov1:
- Clear the database before each test class
- Cache the database connection instance to avoid 'Too Many Connections' 
- Adjust tests accordingly.

- Is there a cleaner way to do this? I don't like that teardown is static but setup isnt.
- Can we truncate ALL tables instead of hardcoding the ones we want to truncate in?

**Confirmed**

### DIFF
--- a/tests/integration/TestCase.php
+++ b/tests/integration/TestCase.php
@@ -27,6 +27,29 @@ abstract class TestCase extends \PHPUnit\Framework\TestCase
      */
     protected $app;
 
+    // Used to run non-static setUp only once.
+    protected $initialized = false;
+
+    protected function setUp()
+    {
+        if (!$this->initialized) {
+            // Clear some essential tables. If it's used in more than 1 test classes, it should be here.
+            $this->prepareDatabase([
+                'users' => [],
+                'groups' => [],
+                'group_user' => [],
+                'group_permission' => [],
+                'discussions' => [],
+                'posts' => [],
+                'settings' => [],
+            ]);
+            // Unset app so that extenders can be resolved
+            $this->app = null;
+        }
+
+        $this->initialized = true;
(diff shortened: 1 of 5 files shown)


franzliedke:
Mmh, I'd rather not go this route. Instead, let's do something I have wanted to do for a long time and wrap each test in a transaction, so that we can avoid the very slow truncation mechanism. 
This tremendously improves the speed of the integration tests, which is always nice.

I've done so in the [`fl/tests-in-transaction` branch](https://github.com/flarum/core/tree/fl/tests-in-transaction), but that requires a bit more work: this is the old problem of test scenarios extending the app after the app already being resolved (because of DB insertions).

I feel this approach would provide us with super simple assumptions for each test, without the need for targeting specific database tables, like in this PR:
- We start with a basic, completely-set-up Flarum install provided by the `setup.php` script.
- Each test can add anything it wants or needs to the database.
- After each test, the database is reset to its previous clean state.

I believe this would solve the problems you are trying to tackle here - though I can't be completely sure, because the description is somewhat vague on the *why*. :wink:

askvortsov1:
I am a big fan of this, and look forward to the PR! Using transactions is definitely much better than this haphazard approach. 

> but that requires a bit more work: this is the old problem of test scenarios extending the app after the app already being resolved (because of DB insertions).

@franzliedke Do you mean that you'd like to solve that issue properly in your proposed changes? Tbh, one somewhat simple approach could be that if, when calling `$this->extend`, `$this->app` is not null, set it to  null so that it would be regenerated next time. This would probably slow down the tests a bit, but would solve the problem without causing an inconsistent state.

That being said, I'm going to close this PR as what you're proposing is much better.

PR #2174, opened 2020-05-18
Use drivers for display names, add display name extender

askvortsov1:
**Part of #2091, #1891 **

- Deprecate GetDisplayName event
- Use drivers for display name
- Add extender for adding display name driver
- Add integration tests for extender
- Add frontend ui for 

- Is there a better place to put the driver UI than the basics page?

**Confirmed**


luceos:
There's a bug in this code @askvortsov1 ; the property is `$displayNameDrivers`, but it is nowhere used in the User extender.

askvortsov1:
Hmm, odd that the tests still worked... Nevertheless, fixed in aaebd35

PR #2176, opened 2020-05-19
Refactor password checker, add extender

askvortsov1:
**Part of #1891**

- Deprecate CheckingPassword event
- Use callbacks for registering password checkers
- Add extender for adding/removing password checkers
- Add integration tests for password checkers

- Anything I missed?

**Confirmed**


src/User/User.php
```diff
@@ -320,13 +333,28 @@ public function getDisplayNameAttribute()
...
-        return static::$hasher->check($password, $this->password);
+        $valid = false;
+
+        foreach (static::$passwordCheckers as $checker) {
+            $result = $checker($this, $password);
+
+            if ($result === false) {
+                return false;
+            } elseif ($result === true) {
+                $valid = true;
```
luceos, line 353: What's the benefit of not immediately returning false here. In any future iteration the value will only be set to true if true anyway. Allowing more iteration will only cost us performance here.

src/User/UserServiceProvider.php
```diff
@@ -82,6 +96,16 @@ public function boot()
...
+        $passwordCheckers = array_map(function ($checker) {
+            if (is_string($checker)) {
+                $checker = $this->app->make($checker);
+            }
+
+            return $checker;
+        }, $this->app->make('flarum.user.password_checkers'));
+
+        User::setPasswordCheckers($passwordCheckers);
+
```
luceos, line 108: Newline isn't really necessary here?

src/User/UserServiceProvider.php
```diff
@@ -82,6 +96,16 @@ public function boot()
             return false;
         });
 
+        $passwordCheckers = array_map(function ($checker) {
+            if (is_string($checker)) {
+                $checker = $this->app->make($checker);
+            }
+
+            return $checker;
+        }, $this->app->make('flarum.user.password_checkers'));
```
luceos, line 105: I've seen similar logic before, can't we make a helper function that loops and resolves automatically?

askvortsov1: We could, but considering the need for imports, I don't know how much better it would be. Also unsure where we would put it. Regardless, I'd rather revisit this after we finish extenders.

src/User/User.php
askvortsov1: Do you mean returning true? It's because an output of false overrides any outputs of true, regardless of order of evaluation. Unless order is a feature (as with middleware), I'm trying to maximally avoid order of evaluation mattering. 

As such, an explicit false, denies, an explicit true allows as long as there are no explicit denials, and no trues means deny.

src/Extend/User.php
```diff
@@ -0,0 +1,63 @@
...
+        return $this;
+    }
+
+    /**
+     * Remove a password checker.
+     *
+     * @param string $identifier: The unique identifier of the password checker to remove.
+     * @return self
+     */
+    public function removePasswordChecker(string $identifier)
```
franzliedke, line 44: Is there a known use-case for this currently? If not, I'd avoid it for now...

franzliedke:
On a higher level: This feels a bit too oriented around the old event, but I'd like us to focus on *use cases* with the extenders. What are we (extension authors) hoping to achieve when using this extender?

Doesn't changing the password checker usually mean we'd need to change the hasher as well? We might need an abstraction around this combination of actions: persisting and checking passwords. (PHP's password hashing abstractions also adds a rehash check to the mix.)

askvortsov1:
Well, I'd agree with that point. This reminds me most of Django's authentication backends: https://docs.djangoproject.com/en/3.0/topics/auth/customizing/. We could expand the scope of the extender to entail that?

askvortsov1:
Shoo!

askvortsov1:
@clarkwinkelmann The only examples of real-life use of this event are in your extensions, so congrats, you're hereby subpoenaed as an expert witness :stuck_out_tongue:!

I agree with Franz that the password checker and hasher are related concepts. But from what I'm seeing in https://github.com/clarkwinkelmann/flarum-ext-passwordless/blob/230ef8a5602bb4a72a25a42f50823d142d44f932/src/Listeners/CheckPassword.php and https://github.com/migratetoflarum/old-passwords/blob/d6b78d2ac6cd609d41e5df1af69c548bcda920f7/src/Listeners/CheckPassword.php, it's not _quite_ that direct. IMO, we have a few options:

- Add methods to the extender to change the password hasher, this might be a good idea regardless (but what if multiple extensions do it at once?)
- Combine the hasher and password checker into a "password backend" driver. Not sure I like this idea though
- Go a step beyond, and substitute the whole "checking password" method with a Django-esque auth driver. But honestly that's pretty similar to what we have now.

src/User/User.php
clarkwinkelmann: So if I understand correctly, we need at least one return to be `=== true` and none `=== false` for `checkPassword` to return `true`.

I'm assuming it's so checks can return `null` without impacting the result if they have neither success nor failure.

src/Extend/Auth.php
```diff
@@ -0,0 +1,63 @@
...
+class Auth implements ExtenderInterface
+{
+    private $addPasswordCheckers = [];
+    private $removePasswordCheckers = [];
+
+    /**
+     * Add a new password checker.
+     *
+     * @param string $identifier: Unique identifier for password checker.
+     * @param string $passwordChecker: A closure or invokable class that contains the logic of the password checker.
```
clarkwinkelmann, line 24: Needs `callable|string` docblock

src/User/UserServiceProvider.php
```diff
@@ -84,6 +98,15 @@ public function boot()
             User::addGroupProcessor($callback);
         }
 
+        $passwordCheckers = array_map(function ($checker) {
+            if (is_string($checker)) {
+                $checker = $this->app->make($checker);
```
clarkwinkelmann, line 103: I'm concerned that this gets resolved every time Flarum boots, even though password validation is something that happens very rarely during the lifetime of the app. I'd advocate a solution similar to https://github.com/flarum/core/pull/2423#discussion_r517028321

clarkwinkelmann (commented):
Indeed, seems like I'm the only one to have attempted to extend this.

One use case is definitely replacing the default hasher. But this is probably a rare use case, since the default is very sensible.

On the hasher front, I'd suggest the following, if that's not already the case: allow customizing the hasher of the Laravel Hash object with a config value (assuming this is customizable with a constructor parameter), and have the hasher in the container (that's already the case).

This offers a lot of flexibility and an extension can always replace the hasher in the container if they wish so. No need to build any extender or configuration UI for the password hasher IMO.

---

The other use case is the one I have used in these two occasions: allowing additional things to be used as password, which aren't in the `password` column, and possibly not in the database at all.

I think the proposed solution in this PR looks good. I don't think we really *need* priority to be customizable. Ability to add and remove should be enough. Having the default check be in the stack also allows extensions to use this path if they want to replace the default hasher with something that's not plug-and-play into the Laravel Hasher.

src/Extend/Auth.php
askvortsov1: For the hasher, couldn't an extension just override `$container->make('hash')` via a ServiceProvider extender?

src/User/UserServiceProvider.php
askvortsov1: I'll update this when #2423 is merged

clarkwinkelmann (commented):
Any reason `ContainerUtil::wrapCallback` is called in the provider and not in the extender? Are we consistent across extenders with that way of doing things?

clarkwinkelmann:
@askvortsov1 was something holding this back?

Should I do a final review now or wait for something else?

askvortsov1:
I don't believe so, should be good to go

askvortsov1:
Merge conflicts resolved, this should be good to review.

clarkwinkelmann (changes requested):
I have not tested locally, but it's looking all good to me.

I'll try to update my old-passwords extension before release to actually test it with the new system once this is merged to master.

I think you will want to update the deprecation comment before merging? It's the only thing I noticed.

src/User/Event/CheckingPassword.php
```diff
@@ -11,6 +11,9 @@
 
 use Flarum\User\User;
 
+/**
+ * @deprecated beta 14, remove in beta 15. Use User extender instead.
```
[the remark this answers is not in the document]
clarkwinkelmann: Also, I think the name of the extender in the comment is wrong?

clarkwinkelmann (approved):

tankerkiller125 (approved):

src/User/User.php
askvortsov1: To clarify, checkers should return as follows:

- If the password is valid, return true
- If not, return null
- If the request must be blocked, return false

This is what's currently used in Clark's extensions. Before I merge this in, is this schema fine with everyone?

clarkwinkelmann: Maybe we could offer CONST aliases just like we did for policies? I'm ok with the raw value, but maybe it would be a bit easier in the documentation if each of these option had a human readable constant associated with it?

askvortsov1: Hmm, I'm not sure. They'd have to be properties of the User object. More importantly though, I have no idea what to call the `null` one. `User::PASSWORD_VALID` and `User::PASSWORD_INVALID` make sense, but maybe `User::PASSWORD_SKIP`?

clarkwinkelmann: Not sure :upside_down_face: 

Maybe I'd call them something including AUTHORIZE, FAIL_AUTHORIZATION and something with PASS/SKIP/NEXT?

clarkwinkelmann: Also, technically you don't really need to return `null` ever. It's also what happens when you don't return anything so there might be a need for a constant.

askvortsov1: I think that this might make things more confusing to be completely honest. Let's keep it as is for now; if we see confusion in the developer community we can always add aliases later I think?

> Also, technically you don't really need to return null ever. It's also what happens when you don't return anything so there might not be a need for a constant.

True, I'll try to clarify this in the docblock

Issue #2092, opened 2020-03-28
Refactoring Policies


askvortsov1:
## Functionality Discussion

**Is your feature request related to a problem? Please describe.**
Policies are currently designed in a way that is confusing, and does not work well with our proposed extenders system, as ScopeModelVisibility depends on a non-domain event.

I've started on some of these in https://github.com/flarum/core/pull/2056.

**Describe the solution you'd like**
There's a lot of components to this, but I would propose the following:

1. Split off visibility scoping into a second abstract class from AbstractPolicy. The way I see it, AbstractPolicy applies to whether or not a user has a permission on a given instance/class of a model. ScopeModelVisibility restricts a query to records that a user can see, optionally based off of some action. It seems like these 2 were grouped together for sake of convenience, but similar methods (`->view()` vs `->find()`), support for custom methods (based on the permission in question), and significant differences in functionality make this a suboptimal design IMO.

2. in AbstractPolicy, rename `getPermission` to `canPerform`, or similar. Permissions are something concrete that are attached to groups and stored in the database. Policy checks allow logic to determine whether a user can perform a somewhat more abstract action, such as `view`, `edit`, or something custom like `overrideAdminSettings` (made that one up but it's possible to implement). `canPerform` would be more descriptive, or something like `canPerformActionOnInstance` (although that one is clunky).

3. Rename pretty much everything on the new ScopeModelVisibility-related abstract class. I don't like `find`, because it's not a searching sysgtem, and it's not looking for individual instances. I'd prefer that devs, when extending the abstract model, define:
- $this->scopeQuery(): now called `find`, this scopes a query for viewing
- $this->scopeQueryForAction(): now called `findACTION_NAME`, where a different method is defined for each action. I think these should be consolidated into one method, where the action is provided with a string, then devs could conditionally apply filters depending on the action in question. Alternatively, `$this->scopeQueryForACTION_NAME` could also work.

4. Move AbstractPolicy away from events in implementation. One serious issue with the current implementation can be found here: https://github.com/flarum/core/issues/2091. A potential fix could be found here: https://github.com/flarum/core/pull/2056

5. Move AbstractQueryVisibilityScoper away from events in implementation. This removed yet another non-domain event, which is part of our path to stable. I propose the following:

- We create a QueryVisibilityScoperManager class. This is passed to the AbstractModel class as a static instance variable, similar to how we pass the event dispatcher
- We create an extender which adds QueryVisiblityScopers defined in extensions to this manager
[...]

PR #2181, opened 2020-05-24
Get rid of Laravel Gate contract

askvortsov1:
**Refs #2055 #2092 **

As of right now, the only piece of the gate we actually use is the `before` callback in UserServiceProvider. This PR replaces the gate with that callback, freeing us from the Laravel Gate contract that we don't use anyway. Unless I made a mistake somewhere, no functionality has actually been changed.

Next steps:
- #2186 
- Implement the extender-able, namespaced policy system discussed in #2092.

- What here, if anything, do we need to provide a BC layer for? $user->can has been around for a while now, and Im assuming that is being used instead of directly calling the gate.

**Confirmed**

### DIFF
--- a/src/Api/Serializer/DiscussionSerializer.php
+++ b/src/Api/Serializer/DiscussionSerializer.php
@@ -24,10 +24,10 @@ class DiscussionSerializer extends BasicDiscussionSerializer
             'createdAt'         => $this->formatDate($discussion->created_at),
             'lastPostedAt'      => $this->formatDate($discussion->last_posted_at),
             'lastPostNumber'    => (int) $discussion->last_post_number,
-            'canReply'          => $this->user->can('reply', $discussion),
-            'canRename'         => $this->user->can('rename', $discussion),
-            'canDelete'         => $this->user->can('delete', $discussion),
-            'canHide'           => $this->user->can('hide', $discussion)
+            'canReply'          => $this->actor->can('reply', $discussion),
+            'canRename'         => $this->actor->can('rename', $discussion),
+            'canDelete'         => $this->actor->can('delete', $discussion),
+            'canHide'           => $this->actor->can('hide', $discussion)
         ];
 
         if ($discussion->hidden_at) {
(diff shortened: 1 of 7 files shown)


src/Api/Serializer/DiscussionSerializer.php
```diff
@@ -10,40 +10,24 @@
...
             'lastPostedAt'      => $this->formatDate($discussion->last_posted_at),
             'lastPostNumber'    => (int) $discussion->last_post_number,
-            'canReply'          => $gate->allows('reply', $discussion),
-            'canRename'         => $gate->allows('rename', $discussion),
-            'canDelete'         => $gate->allows('delete', $discussion),
-            'canHide'           => $gate->allows('hide', $discussion)
+            'canReply'          => $this->actor->can('reply', $discussion),
+            'canRename'         => $this->actor->can('rename', $discussion),
+            'canDelete'         => $this->actor->can('delete', $discussion),
+            'canHide'           => $this->actor->can('hide', $discussion)
```
franzliedke, line 30: How often do we call the `can` method throughout the codebase? I wonder if we could get rid of it in most / all places, in favor of injecting the `Gate` and passing the actor instance to the `allows()` method - that would let us avoid the static member on the `User` class.

Not sure if it's worth it...

src/Discussion/DiscussionPolicy.php
```diff
@@ -29,25 +28,18 @@ class DiscussionPolicy extends AbstractPolicy
...
      */
     protected $events;
 
     /**
      * @param SettingsRepositoryInterface $settings
-     * @param Gate $gate
      * @param Dispatcher $events
      */
-    public function __construct(SettingsRepositoryInterface $settings, Gate $gate, Dispatcher $events)
+    public function __construct(SettingsRepositoryInterface $settings, Dispatcher $events)
```
franzliedke, line 40: This was just a forgotten dependency, apparently?

src/Api/Serializer/DiscussionSerializer.php
askvortsov1: I don't think it is: extensions use $actor->can a lot, it would be a big breaking change with little benefit

src/Discussion/DiscussionPolicy.php
askvortsov1: yep, probably from when we started using $actor->can instead.

PR #2182, opened 2020-05-24
Fix installer removing URL port

### LINKED ISSUE #2152: Can't set port for Flarum Host
## Bug Report

**Current Behavior**
When doing an install you can not set a port for the Flarum Hostname (eg http://localhost:8080 won't work)

**Steps to Reproduce**
1. Create a fresh install
2. During the install command or menu use a port in the Hostname area
3. See problem

**Expected Behavior**
Should allow ports here, currently the port gets removed from the input

tankerkiller125:
**Fixes #2152 **

Fixes the Installer so that a port can be used in the URL

Is there any way to simplify this even more? 

**Confirmed**

### DIFF
--- a/src/Install/BaseUrl.php
+++ b/src/Install/BaseUrl.php
@@ -71,7 +71,7 @@ final class BaseUrl
             $path = substr($parseUrl['path'], 0, strrpos($parseUrl['path'], '/'));
         }
 
-        $port = isset($parseUrl['port']) ? ':' . $parseUrl['port'] : '';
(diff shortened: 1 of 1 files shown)


franzliedke:
@tankerkiller125 Can you send a follow-up PR that adds test for this feature? The `BaseUrl` class is already unit-tested.

tankerkiller125:
@franzliedke Sure, I get started on that tonight once I finish up on my College assignment.

PR #2183, opened 2020-05-24
Don't store checkbox instances in NotificationGrid

askvortsov1:
**Refs #1821 #2144**

- Don't store checkbox instances in NotificationGrid
- Use props for Checkbox/Switch loading state instead of attribute so it can be passed in / modified

- The code in `view` is very repetetive. Can we somehow put the key into a variable?
- I don't like that preferenceSaver is directly changing the component's props like that. But I also don't think it's a priority (we're not storing an instance of the component), and I don't know exactly what to do with it (we could have a register of all the fields / preferences as an extensible method, create an array of loadingState in `init` of SettingsPage, but that would still require extensions that add preferences to include a reference to that loadingState, breaking their loading indicators) 

**Confirmed**

### DIFF
--- a/js/src/forum/components/NotificationGrid.js
+++ b/js/src/forum/components/NotificationGrid.js
@@ -25,7 +25,7 @@ export default class NotificationGrid extends Component {
      *
      * @type {Object}
      */
-    this.loadingState = {};
+    this.loading = {};
 
     /**
      * Information about the available notification types.
@@ -62,7 +62,7 @@ export default class NotificationGrid extends Component {
                   <td className="NotificationGrid-checkbox">
                     {Checkbox.component({
                       state: !!this.props.user.preferences()[key],
-                      loading: this.loadingState[key],
+                      loading: this.loading[key],
                       disabled: !(key in this.props.user.preferences()),
                       onchange: () => this.toggle([key]),
                     })}
@@ -107,14 +107,14 @@ export default class NotificationGrid extends Component {
     const enabled = !preferences[keys[0]];
 
     keys.forEach((key) => {
-      this.loadingState[key] = true;
+      this.loading[key] = true;
       preferences[key] = enabled;
     });
 
     m.redraw();
 
     user.save({ preferences }).then(() => {
-      keys.forEach((key) => (this.loadingState[key] = false));
+      keys.forEach((key) => (this.loading[key] = false));
 
       m.redraw();
     });
--- a/js/src/forum/components/NotificationGrid.js
+++ b/js/src/forum/components/NotificationGrid.js
@@ -36,6 +36,8 @@ export default class NotificationGrid extends Component {
   }
 
   view() {
+    const preferences = this.props.user.preferences();
+
     return (
       <table className="NotificationGrid">
         <thead>
@@ -61,9 +63,9 @@ export default class NotificationGrid extends Component {
                 return (
                   <td className="NotificationGrid-checkbox">
                     {Checkbox.component({
-                      state: !!this.props.user.preferences()[key],
+                      state: !!preferences[key],
                       loading: this.loading[key],
-                      disabled: !(key in this.props.user.preferences()),
+                      disabled: !(key in preferences),
                       onchange: () => this.toggle([key]),
                     })}
                   </td>
--- a/js/src/forum/components/NotificationGrid.js
+++ b/js/src/forum/components/NotificationGrid.js
@@ -55,16 +55,18 @@ export default class NotificationGrid extends Component {
               <td className="NotificationGrid-groupToggle" onclick={this.toggleType.bind(this, type.name)}>
                 {icon(type.icon)} {type.label}
               </td>
-              {this.methods.map((method) => (
-                <td className="NotificationGrid-checkbox">
+              {this.methods.map((method) => {
+                const key = this.preferenceKey(type.name, method.name);
+
+                return <td className="NotificationGrid-checkbox">
                   {Checkbox.component({
-                    state: !!this.props.user.preferences()[this.preferenceKey(type.name, method.name)],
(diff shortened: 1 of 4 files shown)


js/src/forum/components/NotificationGrid.js
```diff
@@ -21,37 +21,22 @@ export default class NotificationGrid extends Component {
...
-          state: !!preference,
-          disabled: typeof preference === 'undefined',
-          onchange: () => this.toggle([key]),
-        });
-      });
-    });
   }
 
   view() {
+    console.log('hi');
```
SychO9, line 39: Hello 😆 

forgotten debugging code ?

js/src/forum/components/NotificationGrid.js
```diff
@@ -133,7 +123,7 @@ export default class NotificationGrid extends Component {
    * @param {String} method
    */
   toggleMethod(method) {
-    const keys = this.types.map((type) => this.preferenceKey(type.name, method)).filter((key) => !this.inputs[key].props.disabled);
+    const keys = this.types.map((type) => this.preferenceKey(type.name, method)).filter((key) => key in this.props.user.preferences());
```
datitisev, line 126: I'm not sure if babel correctly transpiles this for older browsers.

Then again, I don't know what browsers we want to support anymore 🤔

js/src/forum/components/NotificationGrid.js
askvortsov1: Indeed :D Thanks for the review!

js/src/forum/components/NotificationGrid.js
askvortsov1: As per https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in, looks like it's supported widely?

datitisev: Oh. I never use it and assumed it to be newer syntax 🤔. Ignore me.

franzliedke:
> In particular the preferenceSaver thing WILL need change anyway to work with Mithril 2.

@clarkwinkelmann Can you expand on that? It seems to contradict what @askvortsov1 said in the PR description:

> I don't like that preferenceSaver is directly changing the component's props like that. But I also don't think it's a priority (we're not storing an instance of the component), ...

I'm probably just misunderstanding something, though.

clarkwinkelmann:
The `component` variable is a component instance, which we won't be able to pass once we stop using component instances.

Currently the way that instance is obtained is through the second parameter of `onchange`. I'm not sure what Mithril did in the latest versions but I doubt the component instance is still passed https://github.com/flarum/core/blob/638d6d79ca25adde174ce755a05b80cb26d0f5ee/js/src/forum/components/SettingsPage.js#L142

An alternative could be to pass the `props` object, but then it requires saving an instance of the `props` object. So I would recommend storing an attribute for the loading state of each individual checkbox.

askvortsov1:
Argh that's a good point, I hadn't thought of that... Considering that pretty much every preference is updated on change, perhaps preferenceLoadingState could be separated out as a property of the current user along with the preferences themselves?

askvortsov1:
> So I would recommend storing an attribute for the loading state of each individual checkbox.

I've been thinking about this, but can't come up with a good place to store the preference loading states:

1. We could do it individually as streams, and provide those to the preference saver
2. We could store it individually and have each component be in charge of changing its own loading state. We should probably also then separate out the preferenceSaver into a helper/util. We could also bind the preferenceSaver to the component so the preferenceSaver can edit `this.loading`
3. We could store it in the Settings page via an extensible itemlist of keys (all initialized to false), and have the preference saver manipulate it from there.
4. We could store it with the User model.

My favorite here is probably 2, but I would also be alright with 3. Not a fan of 1 or 4.

clarkwinkelmann:
@askvortsov1 yes my thought was option 2.

> We should probably also then separate out the preferenceSaver into a helper/util

I don't think that's necessary. It's already a wrapper around the very apt `this.user.savePreferences()` function. I don't think another helper method is necessary. Let's leave extensions the task of managing their loading states. They can just copy what we did in core here anyway.

I'd suggest applying my two proposed code samples from the in-code comments and then I'm ok with the PR.

js/src/forum/components/SettingsPage.js
```diff
@@ -116,11 +116,11 @@ export default class SettingsPage extends UserPage {
    */
   preferenceSaver(key) {
     return (value, component) => {
-      if (component) component.loading = true;
+      if (component) component.props.loading = true;
```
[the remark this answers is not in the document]
askvortsov1: It feels like this would be a lot of code repetition though?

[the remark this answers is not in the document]
clarkwinkelmann: I don't feed like it is. Some extensions might even want to use a bit of a different control, or a dropdown, or something else, in which case they'd need to rewrite it anyway.

In Flarum I believe we only have two such fields. This one, and the "automatically follow discussions I reply to". I don't know many extensions that add settings, but we could do a quick survey to find out.

I think a bit of repetition here is okay if this avoid a function that tries to do everything, including modifying arbitrary state variables related to other inputs.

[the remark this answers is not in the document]
askvortsov1: There's also the notification grid but yeah I see your point.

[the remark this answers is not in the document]
clarkwinkelmann: It seems like you've already implemented a solution for the notification grid, no? No further changes are required there I believe.

It's only the individual checkboxes below the notification grid that still use these references to the component for their loading state.

datitisev:
**Breaking changes:**

- `NotificationGrid` `inputs` no longer exists
- `Checkbox` `loading` class property is now a `loading` component prop

PR #2184, opened 2020-05-24
Don't store PostUser instance in CommentPost

askvortsov1:
**Refs #1821 #2144**

- Don't store PostUser instance in commentPost
- Eliminate unnecessary cardVisible attribute of PostUser

I am concerned about the `this.subtree.check(() => this.isEditing());` in `CommentPost.js`: I'm assuming we're dropping that in Mithril 2.0, but when I do so here, there's a delay when editing before the preview starts to show (for some reason, config isn't immediately called).

**Confirmed**

### DIFF
--- a/js/src/forum/components/CommentPost.js
+++ b/js/src/forum/components/CommentPost.js
@@ -33,11 +33,7 @@ export default class CommentPost extends Post {
 
     // Create an instance of the component that displays the post's author so
     // that we can force the post to rerender when the user card is shown.
-    this.postUser = new PostUser({ post: this.props.post });
-    this.subtree.check(
-      () => this.postUser.cardVisible,
-      () => this.isEditing()
-    );
+    this.subtree.check(() => this.isEditing());
   }
 
   content() {
@@ -131,7 +127,7 @@ export default class CommentPost extends Post {
     const post = this.props.post;
     const props = { post };
 
-    items.add('user', this.postUser.render(), 100);
+    items.add('user', PostUser.component({ post: this.props.post }), 100);
     items.add('meta', PostMeta.component(props));
 
     if (post.isEdited() && !post.isHidden()) {
--- a/js/src/forum/components/CommentPost.js
+++ b/js/src/forum/components/CommentPost.js
@@ -31,8 +31,6 @@ export default class CommentPost extends Post {
      */
     this.revealContent = false;
 
-    // Create an instance of the component that displays the post's author so
-    // that we can force the post to rerender when the user card is shown.
     this.subtree.check(() => this.isEditing());
   }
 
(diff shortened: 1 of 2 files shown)


datitisev:
Subtree retainer will still exist in Mithril 2.x - it's actual behavior may be a bit different. Can you perhaps debug further why there is a delay? Don't want to change that behavior.

js/src/forum/components/CommentPost.js
```diff
@@ -131,7 +127,7 @@ export default class CommentPost extends Post {
     const post = this.props.post;
     const props = { post };
 
-    items.add('user', this.postUser.render(), 100);
+    items.add('user', PostUser.component({ post: this.props.post }), 100);
```
clarkwinkelmann, line 130: `{ post: this.props.post }` could be shortened to `{ post }` (itself short for `{ post: post }`) because the value is already available in a variable named `post`

clarkwinkelmann:
I'm unsure what the consequences of this PR are without actually testing it.

Does the behavior really doesn't change?

It seems like this was required to ensure proper redraw after the transition completes.

js/src/forum/components/CommentPost.js
clarkwinkelmann: Also, there's even that `props` variable there that could be re-used.

But I'd actually suggest dropping that `props` variable because it could introduce unwanted effects to re-use a reference to that variable. `{post}` could simply be used everywhere a `props` variable is used right now.

franzliedke:
If I understand correctly, this PR removes the `cardVisible` attribute because that seems to not be needed when looking at the code. So far, so good.

However, the comment above the subtree retainer's `check()` call in the `CommentPost` component clearly states that we want the comment post to re-render whenever the user card is toggled. So that's why that otherwise unnecessary attribute was there.

So, for me the question is: why did we want or need the redraw?

And if we decide we want to keep the redraw logic, how would we implement this without storing a reference to the component? I guess we'd need to promote the `cardVisible` thing (*per post*) to a "state" on a higher level in the component hierarchy? Feels a bit weird indeed... :thinking:

askvortsov1:
I managed to track this down all the way to https://github.com/flarum/core/blob/b68a4711dceb303dfc9b9b47cf9ae45b18d470b5/js/forum/src/components/post-comment.js

Before that, we had
https://github.com/flarum/core/blob/6f67b8c247841cc516bd0c26c0c6edd75a5ebf88/ember/forum/app/components/discussion/post-comment.js
which was back when we used ember

I can only assume that the UI has changed since then and that retain is no longer necessary, as removing it did not cause any issues for me when testing

datitisev:
I think this PR broke the UserCard.

JIg1xatlU1

franzliedke:
@datitisev Can you explain what exactly you are doing in that screen recording? Clicks? Hovers? Magic incantations?

datitisev:
@franzliedke It's a simple hover. The UserCard is rendered in the DOM (though invisible) and is on top of the other elements. Trying to click the timestamp clicks the username link instead.

PR #2185, opened 2020-05-24
Extract NotificationList state

askvortsov1:
**Refs #1821 #2144**

- Extract NotificationListState into its own state class

- Anything I missed?

**Confirmed**

### DIFF
--- a/js/src/forum/states/NotificationListState.js
+++ b/js/src/forum/states/NotificationListState.js
@@ -1,5 +1,7 @@
 export default class NotificationListState {
-  constructor() {
+  constructor(app) {
+    this.app = app;
+
     this.notificationPages = [];
 
     /**
@@ -34,7 +36,7 @@ export default class NotificationListState {
    * been loaded.
    */
   load() {
-    if (app.session.user.newNotificationCount()) {
+    if (this.app.session.user.newNotificationCount()) {
       this.notificationPages = [];
     }
 
@@ -42,7 +44,7 @@ export default class NotificationListState {
       return;
     }
 
-    app.session.user.pushAttributes({ newNotificationCount: 0 });
+    this.app.session.user.pushAttributes({ newNotificationCount: 0 });
 
     this.loadMore();
   }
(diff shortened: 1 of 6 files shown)


datitisev:
Why are we passing the app global? I don't understand the purpose of this. If it is to allow for extensibility, wouldn't that require the extension developer to create their own whole Application class anyway with the data that this state requires?

askvortsov1:
@datitisev As per the discussion in https://github.com/flarum/core/pull/2150#discussion_r429495569

datitisev:
Hm I see. It doesn't seem right to me.

franzliedke:
@datitisev Would you prefer to see the actual dependencies (e.g. the `store` and the `session`) being passed in instead of the entire app?

datitisev:
I think so, yeah.

js/src/forum/states/NotificationListState.js
```diff
@@ -0,0 +1,104 @@
...
+    this.moreResults = false;
+  }
+
+  getNotificationPages() {
+    return this.notificationPages;
+  }
+
+  isLoading() {
+    return this.loading;
+  }
```
franzliedke, line 28: Same discussion as [brought up by @datitisev here](https://github.com/flarum/core/pull/2150#discussion_r429545622)... do we want those methods, or advertise direct access to the properties?

Methods are slightly better in terms of encapsulation (reading the values could be restricted to the state class itself), so I'd favor that?

The docblocks seem to suggest the properties are still part of the official API? Is that because of backwards compatibility?

askvortsov1:
> Instead, moving it from the `NotificationList` to the `NotificationsDropdown` would be enough - the dropdown component is the "highest" component that needs this data (no siblings or parents need it), and it could pass relevant data down to children (the `NotificationList` component) as props where they need it.

I would agree with this.

js/src/forum/states/NotificationListState.js
askvortsov1: Docblocks are a result of me copy-pasting stuff. But yeah, I'm in favor of encapsulating here since we're not planning on always keeping the data in these state classes.

askvortsov1:
> I would agree with this.

Actually, scratch that: NotificationsPage also uses it.

datitisev:
Breaking changes: `app.cache.notifications` is now `app.notifications.getNotificationPages()`.

PR #2187, opened 2020-05-27
Fix Paths test failing on Windows

SychO9:
**Fixes #0000**

The PathsTest fails on windows because of the directory separator, it's a backslash in windows.

**Confirmed**

### DIFF
--- a/tests/unit/Foundation/PathsTest.php
+++ b/tests/unit/Foundation/PathsTest.php
@@ -48,7 +48,7 @@ class PathsTest extends TestCase
             'storage' => '/var/www/flarum/storage',
         ]);
 
-        $this->assertEquals('/var/www/flarum'.DIRECTORY_SEPARATOR.'vendor', $paths->vendor);
+        $this->assertEquals('/var/www/flarum/vendor', $paths->vendor);
     }
 
     /** @test */
--- a/tests/unit/Foundation/PathsTest.php
+++ b/tests/unit/Foundation/PathsTest.php
@@ -48,7 +48,7 @@ class PathsTest extends TestCase
             'storage' => '/var/www/flarum/storage',
         ]);
 
-        $this->assertEquals('/var/www/flarum/vendor', $paths->vendor);
+        $this->assertEquals('/var/www/flarum'.DIRECTORY_SEPARATOR.'vendor', $paths->vendor);
     }
 
     /** @test */
--- a/src/Foundation/Paths.php
+++ b/src/Foundation/Paths.php
@@ -34,7 +34,7 @@ class Paths
         }, $paths);
 
         // Assume a standard Composer directory structure unless specified
-        $this->paths['vendor'] = $this->vendor ?? $this->base.DIRECTORY_SEPARATOR.'vendor';
+        $this->paths['vendor'] = $this->vendor ?? $this->base.'/vendor';
     }
 
     public function __get($name): ?string


tests/unit/Foundation/PathsTest.php
```diff
@@ -48,7 +48,7 @@ public function it_derives_the_vendor_dir_from_the_base_path()
             'storage' => '/var/www/flarum/storage',
         ]);
 
-        $this->assertEquals('/var/www/flarum/vendor', $paths->vendor);
+        $this->assertEquals('/var/www/flarum'.DIRECTORY_SEPARATOR.'vendor', $paths->vendor);
```
tankerkiller125, line 51: Why is only one of the slashes replaced? Shouldn't all of them be replaced?

datitisev: image

🤷

franzliedke: Because of this: https://github.com/flarum/core/blob/3c87f800ddaa048d2c19ca296c8f1da9ffb3fbfa/src/Foundation/Paths.php#L37

I mean, we could also change the source code to just use a slash? That should still work on Windows, right?

franzliedke:
Nice catch :grin:

tests/unit/Foundation/PathsTest.php
SychO9: I would guess that it uses the defined paths above and appends to it, hence why only the last slash is the OS's slash type.

franzliedke: Yes, but [it should not be necessary](https://stackoverflow.com/a/38428899/192741). So could we just change `Paths.php` to use a slash instead? Or does anyone see a downside to that?

SychO9: Oh I hadn't seen your reply until now, yeah windows works fine with either type of slashes so I don't think there'll be any problem with changing it.

askvortsov1: I'm fine with either, but if we can have consistent slashes everywhere, that would be preferable as long as it doesn't break anything. Which I'm asuming it doesn't since it currently works with both types of slashes, right?

SychO9: only case scenarios where it would be a problem is if you were using php functions/constants to get a path, and then for example for w/e reason exploding that path with `DIRECTORY_SEPARATOR` that wouldn't be a place to change to a forward slash since php would in windows return a path with backslashes.

askvortsov1: But wouldnt that issue still apply with the slashes that are already in the path?

SychO9: how so ?

askvortsov1: Well, we already have UNIX-style slashes in the `/var/www/flarum` component, so if you're trying to split on windows with `DIRECTORY_SEPARATOR`, it would fail anyway, right?

SychO9: yeah it would because the path is predefined, it would only properly work with the constant if the path was returned from a php function or a php constant such `__DIR__` because it would use the OS slash type.

franzliedke: > Well, we already have UNIX-style slashes in the `/var/www/flarum` component.

That's just example data in our test case, though. [These](https://github.com/flarum/flarum/blob/3aeffd15aac9ef98dd44dc53eff960245ab0d247/site.php#L46-L50) would be realistic values, currently.

askvortsov1: Then we might as well go all out and make sure there's no literal slashes in the code. Or is that too extreme?

franzliedke: I don't think it's necessary - and it would just make the code more verbose.

But I may be missing a case where the difference is relevant.

askvortsov1: So how do you want to proceed?

franzliedke: [This was my proposal.](https://github.com/flarum/core/pull/2187#discussion_r431410375)

SychO9:
let me know if you would rather I rebase into one single commit

askvortsov1:
That's probably not necessary, we generally merge PRs via squash. Thanks though!

Issue #2093, opened 2020-03-28
Extension Dependency Management Revamp


askvortsov1:
## Functionality Discussion

Related Discussions:
https://github.com/flarum/core/issues/1318
https://github.com/flarum/core/issues/1039

**Is your feature request related to a problem? Please describe.**
Flarum assumes that if an extension installs, everything is perfectly fine with it. Additionally, it loads extensions in the order they were enabled. Neither of these are particularly robust. IMO, the following issues should be fixed:

1. If an extension depends on another flarum extension, it should not be enable-able until its flarum dependencies are enabled.
2. If an extension has flarum extensions depending on it, it should not be disable-able until its flarum dependencies are disabled
3. If, during app boot, an enabled extension has a missing flarum dependency, it should not be loaded, and should be marked as disabled (this will not result in an inconsistent state as we don't even TRY to load it).
4. Enabled extensions should be loaded in such an order that an extension is loaded AFTER its enabled dependencies have been loaded.
5. Required PHP versions and php extension versions should be obeyed before loading extensions.

**Describe the solution you'd like**
One way that has been proposed is getting dependencies from the composer.json file. I don't like this for several reasons:

1. It assumes that the extension is 'compiled' (well more like assembled) on the same machine/environment that the forum is running on.
2. It assumes that an extension is dependent on all flarum extensions it lists in requires. I have written extensions that 'require' other extensions while not actually depending on their functionality.
3. The composer.json `require` list does NOT indicate whether something is a flarum extension. This means that during load, we would need to open and check the composer.json file of every single package required by an extension to make a list of flarum extensions that extension depends on
4. This further ties us to composer, in a manner where I don't think that's necessary.

I would like to propose that we build in our own method for dependency management. In addition to the extend.php file, we require a flarum-dependencies.php (or flarum-dependencies.json) file, which would list:

- PHP version requirements
- PHP extension dependencies and version requirements.
- Flarum extension dependencies and version requirements.
- Flarum extension conflicts and version requirements.

If any of these are omitted, they would be ignored.

This would give us a LOT more robustness in our extension management system.

askvortsov1:
Closing in favor of https://github.com/flarum/core/issues/1318, we can revisit this if we decide that we want to do runtime dependency checks.

PR #2188, opened 2020-05-28
Basic Extension Dependency Support

### LINKED ISSUE #1318: Allow optional dependency / one-of-several dependency requirements to control extension boot order
Currently extensions are booted in the order that they are enabled (in both backend and frontend).

This should be modulated so an extension cannot be booted before any of the extensions it `require`s.

askvortsov1:
**Refs #2093**
**Refs #1318**

- Don't enable an extension if its dependencies are not enabled
- Don't disable an extension if its dependencies are not disabled

**Confirmed**

- [x] Frontend changes: tested locally

**Screenshots**
Error when disabling with dependent: https://i.imgur.com/X2npKFt.png
Error when enabling with missing dependencies: https://i.imgur.com/rioQMXc.png

Lang English PR: https://github.com/flarum/lang-english/pull/172


franzliedke:
Why would we need a separate field for declaring extensions we depend on? Composer already has `require`... :thinking: 

Anyway, I won't look at this in more detail until we've made more progress with the frontend rewrite, sorry.

askvortsov1:
> Why would we need a separate field for declaring extensions we depend on? Composer already has require ... 🤔 

Composers require doesn't tell us what is and isn't a flarum extension, so if something isn't installed, we don't know if it's needed. Also, we'd need to do a lot more unnecessary computations to scan every extensions composer json whenever we add an extension. Overall this will just result in a clearer codebase imo. 

> Anyway, I won't look at this in more detail until we've made more progress with the frontend rewrite, sorry.

Sorry I should have made it clearer in the description: this is very much a "I wanna code something useful but don't want to add to Franz's workload so I'm gonna explore this cool idea" PR.

franzliedke:
> Composers require doesn't tell us what is and isn't a flarum extension, so if something isn't installed, we don't know if it's needed. Also, we'd need to do a lot more unnecessary computations to scan every extensions composer json whenever we add an extension.

Ah, but we [have the information about *all Composer-installed packages* in memory already](https://github.com/flarum/core/blob/3b1f5ca07b879e1ccc2466bac8fbf343d2359e75/src/Extension/ExtensionManager.php#L80-L81), just before the `Extension` objects are instantiated. That includes `require` keys as well as the `type` key (which needs to be checked to equal `flarum-extension`).

askvortsov1:
I see your point, I still think the code would be a bit messy. But let's table this for now, and revisit after frontend / extenders?

askvortsov1:
It now uses dependencies from require, and I was able to do it without significantly messing up big O complexity or code convolution!

luceos:
What's the memory footprint increase of this? I'm not going to approve this PR unless we know the downside in terms of performance. Having said so, I'm not sure this is the right approach. This code is executed on each request, even though it is only needed during the installation of an extension. I would vote to move this to a composer lifecycle hook (post update/install cmd) instead if possible. That would be my first attempt.

askvortsov1:
The process for figuring out which extensions are dependent on which has an O(n) time complexity, and a O(n) space complexity (for creating the hash table of php package names to composer.json files, ie $associativeInstalled).

The process of checking whether dependencies are violated have an O(n * m) time compexity (n being number of extensions, m being average number of dependencies per extension). This is, however, inconsequential as it's only run when enabling / disabling an extension, not with every request.

The resolution of extension dependency order has an O(n + v) time complexity (as it's just Kahn's algorithm), with n being the number of extensions, and v being the number of vertices in the dependency graph (ie the sum of each extension's number of dependencies).

None of these are especially high, and all of these scale well. That being said, this code doesn't just check for whether something should be enabled/disabled, but also ensures that extensions are booted in the proper order, every time. That's why I think it should run on every request.

Also, does PHP really boot up the entire system on every request? There isn't a way to boot the application and have it await requests as they come in as a process running in the background?

tankerkiller125:
@askvortsov1 Standard PHP does not have a way to keep the application running, there are of course PHP work arounds for this using special libraries and stuff but that's not for Flarum to get into. Of course any production environment will have OPCache or APC cache enabled which does store the compiled PHP files so they execute faster. Once PHP 8 is out we'll also get to see how JIT does with performance.

src/Extension/ExtensionManager.php
```diff
@@ -296,7 +338,9 @@ public function getEnabledExtensions()
      */
     public function extend(Container $app)
     {
-        foreach ($this->getEnabledExtensions() as $extension) {
+        $enabled = static::resolveExtensionOrder($this->getEnabledExtensions());
```
clarkwinkelmann, line 341: What about doing this whenever an extension is enabled, and save the new order in the database? That way there's no need to do any computation when Flarum boots.

clarkwinkelmann (commented):
I see a bunch of unrelated commits which I guess come from different PRs. It doesn't help to figure out what's part of this PR... Two things above.

js/src/admin/components/ExtensionsPage.js
```diff
@@ -131,4 +132,18 @@ export default class ExtensionsPage extends Page {
 
     app.modal.show(LoadingModal);
   }
+
+  onerror(e) {
+    app.modal.close();
+
+    const error = JSON.parse(e.responseText).errors[0];
+
+    app.alerts.show(
```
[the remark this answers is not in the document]
askvortsov1: How exactly would that work? I don't believe there are already validaiton error messages with this particular text (or close to it). We're not passing a string in the backend, we're passing an array of extension data.

src/Extension/ExtensionManager.php
askvortsov1: I suppose we could... But this way, it'll always be right (even if someone messes with the database), and decouples the storage of enabled extensions from the order in which they're booted. Plus, topological sort runs really fast, it'd hardly be a bottleneck compared to other aspects of the site.

askvortsov1:
> I see a bunch of unrelated commits which I guess come from different PRs. It doesn't help to figure out what's part of this PR...

Yeah that was a messy rebase... sorry about that. All cleaned up now!

src/Extension/ExtensionManager.php
tankerkiller125: I'm going to have to agree with Sasha here, although it's rare that the database gets modified directly it does happen sometimes. especially during some of our more complex troubleshooting processes.

luceos: And I am going to agree with @clarkwinkelmann here. This is all superfluous code that, by itself, might not seriously increase loading times (first-byte) of our forum. But each and every of these kind of decisions will impede performance over time. This scenario, with this code, is especially true with forums that have a ton of extensions - and trust me that will sadly be the reality.

Honestly I would even make this a post update/install hook by saving some kind of **order** map of all **installed** extensions to a file or the database. Then during extension enable/disable retrieve this order and store the order of enabled extensions to the database. If anyone decides to manually override that order from the database, they have good reason to do so and don't particularly would hope to see something override that order magically..

tankerkiller125: Further review into this, I'm going to agree with @luceos and @clarkwinkelmann, I also agree with luceos idea of making this a post update/install hook for composer to store the order map. Further the database edits by people are incredibly rare and if someone did do it on accident (or purpose) we could simply inform them to run the same command that composer does to resolve it.

askvortsov1: I spent some more time thinking about this, and just the dependency checks on enabling / disabling will be a HUGE improvement over what we have now (and much, much simpler to review). And that should also result in a working order.

I'll remove the algorithms stuff and move it over to the extension management commands PR (we'll need it if we're trying to enable/disable extensions in a working order). No need to get everything done in a single PR.

luceos: Are you saying this PR will only contain logic to check whether depending extensions are installed? Sounds like a smart move. The composer hook should be quite simple too, but needs some work.

askvortsov1: Yep! That way, we can split this into separate PRs and have the composer hook later.

As noted earlier somewhere, due to our current system, if we have this check in place before enabling/disabling extensions, that'll prevent the vast majority of issues (since load order currently depends on order of enabling, and this way, you won't be able to enable anything.

The composer hook stuff would be in a separate PR, definitely not aiming for beta 14.

src/Extension/ExtensionManager.php
```diff
@@ -101,6 +109,11 @@ public function getExtensions()
 
                 $extensions->put($extension->getId(), $extension);
             }
+
```
luceos, line 112: Instead of seeding $installedSet above in an expensive loop, why net assign it here after the loop by using the Collection pluck method:

```suggestion
$installedSet = $this->extensions->pluck('name')->toArray();
```

luceos (commented):
I left two comments. Somehow I think this isn't the best solution, but we can work on that when we tackle the dependency logic from within the composer hook.

src/Extension/ExtensionManager.php
```diff
@@ -165,6 +190,22 @@ public function disable($name)
 
         $extension = $this->getExtension($name);
 
+        $dependentExtensions = [];
+
+        foreach ($this->getEnabledExtensions() as $possibleDependent) {
+            foreach ($possibleDependent->extensionDependencies as $dependency) {
+                if ($dependency === $extension->getId()) {
+                    $dependentExtensions[] = $possibleDependent->getId();
+                    // No need to cycle through the rest of this extension's dependencies.
```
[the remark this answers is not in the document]
askvortsov1: I need to improve that message. Essentially, we're finding all the extensions dependent on the extension we want to disable. So we check all enabled extensions, and for each one, we check its dependencies. If one of those is the extension we're disabling, we move onto the next enabled extension. That way, we can give a list of ALL dependent extensions in the validation message

src/Extension/ExtensionManager.php
askvortsov1: I think the cost would be the same (as I'm assuming pluck does a loop under the surface), but it would be cleaner code.

askvortsov1:
Are there any particular concerns you have about this strategy? I agree that we need more robust actual ordering of extensions (this PR doesn't really fix that, nor does it claim to rigorously). But that's what we'll revisit in later PRs.

Either way, we'll need validation checks to explicitly block disabling an extension that others are dependent on. And I think this is as good a way to accomplish that as anything else.

src/Extension/ExtensionManager.php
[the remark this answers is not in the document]
luceos: Got it, missed that.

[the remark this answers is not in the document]
askvortsov1: I will improve that message though to avoid confusion in the future. Rule of thumb: if the purpose isn't obvious, needs better commenting :D

src/Extension/ExtensionManager.php
tankerkiller125: The pluck logic calls `Arr::pluck()` which is the following https://github.com/laravel/framework/blob/6.x/src/Illuminate/Support/Arr.php#L384 which does indeed do a loop. However I think using pluck is better regardless as it makes our code look a bit cleaner.

askvortsov1: Actually looking at this again, Arr::pluck would create an array of values. What I'm doing here uses associative arrays to create a hashset, which will have constant lookup time, compared to an array, where we'd have to iterate through that list of installed extensions, every time we check a dependency.

That being said, I do want to change it to store true instead of null as values, that seems to be convention as described by https://stackoverflow.com/a/57166943.

luceos: You adopted the collection usage, but still defined the same variable above and use it to fill it unnecessarily above. Can you clarify what approach will be final? My preference is using a collection:

- There's no reason to use a hashset, because there's no chance that the array is assigned a duplicate value. The extensions iterated over aren't duplicate.
- We already know the extensions installed so we can simply re-use that set to retrieve the names.

Not sure this is stubborn me, feel free to let me know what you have in mind 👍

askvortsov1: I'm not pushing to use a hashset because of duplicate values, I want to do so for performance reasons. It takes constant time to check if an element is in a hashset, but the entire array must be iterated (although under the surface) to check if an element is in an unsorted array / collection.

luceos: Okay that makes sense. Then I agree with your code.

luceos (approved):

src/Extension/ExtensionManager.php
```diff
@@ -109,16 +99,17 @@ public function getExtensions()
                 $extension->setInstalled(true);
                 $extension->setVersion(Arr::get($package, 'version'));
 
-                $extensions->put($extension->getId(), $extension);
+                // We use the composer package name as the key so that:
+                //   1. We don't have naming collisions
+                //   2. We can use it for constant time lookups in `calculateDependencies`
+                $extensions->put(Arr::get($package, 'name'), $extension);
```
luceos, line 105: I am not entirely sure this causes issues with extensions using the Extension Manager, @datitisev and @clarkwinkelmann do you know of any extensions that will be impacted by this?

I'm okay with this solution, but we do need to clearly mention this in the upgrade notes for devs.

luceos (dismissed):

tankerkiller125 (approved):
Other than the note from luceos I'm good with these changes

luceos:
We just discovered on discord that the change to use the full package name as key is a breaking change and will impact the getExtension method amongst other. So reverted my approval.

src/Extension/Exception/DependentExtensionsExceptionHandler.php
```diff
@@ -0,0 +1,34 @@
...
+ * LICENSE file that was distributed with this source code.
+ */
+
+namespace Flarum\Extension\Exception;
+
+use Flarum\Foundation\ErrorHandling\HandledError;
+
+class DependentExtensionsExceptionHandler
+{
+    public function handle(DependentExtensionsException $e): HandledError
```
clarkwinkelmann, line 16: Do we have a habit of shortening `$exception` to `$e` ? I personally don't like that trend very much.

clarkwinkelmann (changes requested):
I see I have been mentioned previously, but I can't find the message that was notified by email. In the current state of the PR, I'm not seeing any issue with other extensions.

I was about to comment on the enabled**Ids` thing, but I see a commit was just pushed fixing it.

src/Extension/Exception/DependentExtensionsExceptionHandler.php
askvortsov1: That seems to generally be the trend, although we have a few `$error`s as well

js/src/admin/components/ExtensionsPage.js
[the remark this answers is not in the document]
clarkwinkelmann: I think the fact a separate HTTP error code and exception is used prevents us from doing what I described. I'm marking this resolved.

src/Extension/Exception/DependentExtensionsExceptionHandler.php
clarkwinkelmann: Ok that's fine too

clarkwinkelmann:
When the commits stop flowing in I might approve :stuck_out_tongue:

clarkwinkelmann (approved):
I shared some ideas regarding the error message on Discord, suggesting to maybe show title together with ID. But the current implementation is fine for the time being.

Issue #1991, opened 2020-02-09
Add unique slug to groups


askvortsov1:
## Feature Request

**Is your feature request related to a problem? Please describe.**
Right now, groups are difficult to manage at scale, programatically. It's possible to sync groups via IDs, but the ID is the only unique field of groups. The group name is not unique, which means it's impossible to query the API for a particular group. Therefore, if I want to build a system that sets groups from an external source, I need to hardcode, or manually insert the IDs of each group.

**Describe the solution you'd like**
I propose that a unique "slug" field be added to groups. This would give a human readable (and manageable) lookup system for groups. For existing groups, the slug could be generated as a combo of the group name and id ( ex. admin-1), and further configurable in the admin panel.

**Justify why this feature belongs in Flarum's core, rather than in a third-party extension**
It's a change to groups themselves, so core would be a natural place for it. Additionally, it would be very useful to developers who want to extend groups functionality. The difficulty of managing groups at scale is a well known issue, and this would help move towards alleviating it.

tankerkiller125:
Keep for now, probably needs some discussion.

clarkwinkelmann:
Do we have additional use case for this? Most problems can be solved via the ID.

I do like the ID of a slug. But for forums that might handle many groups, I feel like this should be a nullable property.

The slug would make sense if it ended up in URLs somewhere. Is that likely or could the rare extension in need of this provide their own slug column?

askvortsov1:
Someone could make a discussion that adds pages to groups, but it feels like user-directory might have that covered. My main concern is provisioning: if a group gets deleted by accident and recreated, the ID changes. Also, there isn't anything logical controlling the ID. But I do see your point that this isn't necessary for core (and an extension could definitely provide a slug column)

tankerkiller125:
One of the big things I could see this solving is inter forum sync, Notably I know that @askvortsov1 has the SSO Sync Extension that is able to sync groups. The problem however is that someone trying to do SSO between forums would need to create a mapping of groups between all the forums that they perform SSO on. Even if those groups have the same names. With slugs this would be drastically simplified since they would only need to make sure that all the forums have the same slugs.

askvortsov1:
Yeah that explains what I was trying to say a lot better, thanks Matt! And one negative of adding it via an extension is that unlike many other things, if 2 extensions both try to add this column there can be some kinda nasty collisions (or conditional migrations which I think are bad practice)

PR #2192, opened 2020-06-07
Group Gambit Improvements

askvortsov1:
**Needed for https://github.com/FriendsOfFlarum/user-directory/pull/39**

- Add ID to fields searched in group gambit. This is needed to search for groups that have a space in their names. When we add in group slugs as per #1991, this should be less necessary, but good to have for now.
- Use joins instead of looping in group gambit

**Confirmed**

### DIFF
--- a/src/User/Search/Gambit/GroupGambit.php
+++ b/src/User/Search/Gambit/GroupGambit.php
@@ -47,7 +47,7 @@ class GroupGambit extends AbstractRegexGambit
 
         $groupIdentifiers = $this->extractGroupIdentifiers($matches);
 
-        $ids = Group::whereIn('name_singular', $groupIdentifiers)
+        $userIds = Group::whereIn('name_singular', $groupIdentifiers)
             ->orWhereIn('name_plural', $groupIdentifiers)
             ->orWhereIn('id', $groupIdentifiers)
             ->whereVisibleTo($search->getActor())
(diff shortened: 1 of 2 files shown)


src/User/Search/Gambit/GroupGambit.php
```diff
@@ -44,16 +45,14 @@ protected function conditions(AbstractSearch $search, array $matches, $negate)
...
 
-        // TODO: Use a JOIN instead (and don't forget to remove the findByName() method again)
-        $ids = [];
-        foreach ($groupNames as $name) {
-            $group = $this->groups->findByName($name);
-            if ($group && count($group->users)) {
-                $ids = array_merge($ids, $group->users->pluck('id')->all());
-            }
-        }
+        $ids = Group::whereIn('name_singular', $groupIdentifiers)
```
clarkwinkelmann, line 50: I'd suggest taking the opportunity to rename this variable to `$userIds` to improve readability.

askvortsov1:
> I'm globally ok with the approach, but I'm thinking it might be more appropriate to actually use is_numeric to make the difference between a name and ID.

Is the goal to avoid groups named "123"? Personally I'd prefer if we could query with a slug instead of an ID, but there's currently no slug mechanism. Maybe I should make a group slug PR first, and then when that's merged we can use that here?

clarkwinkelmann:
I think the slug thing needs more discussion before it happens, so I'd suggest making the ID the primary way to use the gambit for now, so that extensions like user-directory can start using it starting next beta even if we don't decide on anything for the slug before then.

PR #2195, opened 2020-06-11
Fix discussion unreadCount could be higher than commentCount

w-4:
**Fixes an issue**

Happens when posts are deleted.


datitisev:
We may want a comment to clarify what this does and/or why it's needed.

askvortsov1:
Done @datitisev

PR #2206, opened 2020-06-21
TypeScript infrastructure

datitisev:
Will probably want to include some shims
<details>
<summary>Old example shims in WIP rewrite branch</summary>

https://github.com/flarum/core/blob/4c8af1cdf876b9bd3770f76c33eb6c9ac80608f0/js/webpack-flarum-shims.d.ts#L7-L17

https://github.com/flarum/core/blob/4c8af1cdf876b9bd3770f76c33eb6c9ac80608f0/js/shims.d.ts#L3-L7
</details>

---

- Webpack config is temporarily in this PR to avoid weird NPM garbage & publishing buggy flarum-webpack-config release
  - See https://github.com/flarum/flarum-webpack-config/pull/3

PR #2207, opened 2020-06-21
TypeScript Initial Conversion

datitisev:
See https://github.com/flarum/core/pull/2206.

Currently PRs _into_ `ds/typescript` branch so TypeScript infrastructure (#2206) is available for building.

Merge after state PRs are done.


franzliedke:
> Merge after state PRs are done.

Can you expand on that? I don't see any potential for conflicts right now...

askvortsov1:
I'd also be in favor of starting to merge TS stuff not, the two don't really conflict

datitisev:
@franzliedke This PR is basically whatever TS is converted before state PRs are merged. Not necessarily because of conflicts, but to only have one PR for the initial batch that isn't very complicated and goes file-by-file.

franzliedke:
This batch already looks great, and very manageable in size. :smiley:

js/src/common/utils/extract.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * The `extract` utility deletes a property from an object and returns its
+ * value.
+ *
+ * @param object The object that owns the property
+ * @param property The name of the property to extract
+ * @return The value of the property
+ */
+export default function extract<T, K extends keyof T>(object: T, property: K): T[K] {
```
franzliedke, line 9: Generics... crazy stuff. :scream:

js/src/common/utils/RequestError.ts
```diff
@@ -1,5 +1,14 @@
 export default class RequestError {
-  constructor(status, responseText, options, xhr) {
+  status: string;
+  options: object;
+  xhr: XMLHttpRequest;
+
+  responseText: string | null;
```
franzliedke, line 6: How is this different from `responseText?: string;`?

js/src/common/utils/ItemList.ts
```diff
@@ -10,23 +14,15 @@ class Item {
...
-     *
-     * @type {Object}
-     * @public
-     */
-    this.items = {};
-  }
+  /**
+   * The items in the list
+   */
+  items: { [x: string]: Item } = {};
```
franzliedke, line 20: I assume the name `x` is a required part of this syntax, which I assume to mean that all keys in the object need to be of type `string` and mapping to a value of `Item`?

js/src/common/utils/extract.ts
datitisev: Yep! This was actually new to me as well when I wrote this. It basically means that when you do `extract({ id: 2 }, 'id')`:
- `T` is the type of `{ id: number }`
- `K` is the name of a property of T, so `'id'`
- `T[K]` is the type of the property `T[K]`, or `object[property]`, in this case `number`

Hopefully this makes sense & I explained it correctly.

js/src/common/utils/ItemList.ts
```diff
@@ -1,5 +1,9 @@
 class Item {
-  constructor(content, priority) {
+  content: any;
+  priority: number;
```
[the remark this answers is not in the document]
datitisev: Not sure 🤔 good catch.

js/src/common/utils/ItemList.ts
datitisev: Yep. That `x` could be `key` instead.

js/src/common/utils/RequestError.ts
datitisev: `?:` means this type or undefined. Undefined isn't treated the same as null.

Here we are saying responseText *will* be defined, just may be null.

matteocontrini: I think this is true only if the strict flag is enabled though, right? It's quite likely that I'm missing something, but I don't see it being set anywhere 🤔

datitisev: 🤷 There _is_ a difference between undefined & null, so I think we should make it clear that it'll be defined but possibly null.

matteocontrini: Yes, but what I meant is that without strict mode you can still assign `null` to any variable or field even without explicitly writing that. To disallow this behaviour you need strict mode, which is usually recommended as a default configuration anyway.

datitisev: Ah I see what you mean. Not sure if we want strict mode because, due to the nature of core, most stuff isn't declared until the initializers run, and even then stuff like app.forum isn't set. It'll be set in most instances, but we can't really declare those.

js/src/common/utils/ItemList.ts
franzliedke: Indeed, [the name is only for readability](https://basarat.gitbook.io/typescript/type-system/index-signatures#declaring-an-index-signature). I'm changing it to `key` now.

js/src/common/utils/RequestError.ts
franzliedke: I'd love to try out strict mode, to be honest. But we can do that when we get to TypeScript-y-fying the application class.

js/src/common/utils/ItemList.ts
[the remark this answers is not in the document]
franzliedke: Apparently possible without strict mode: https://www.typescriptlang.org/docs/handbook/functions.html#optional-and-default-parameters

PR #2209, opened 2020-06-24
Fixed user card issue

askvortsov1:
Fixes the issue brought up in https://github.com/flarum/core/pull/2184#issuecomment-648184305

Before, the card would only appear in the DOM when it was visible. Now it always appears. Therefore, when it's not in view, it's z-index should be less than that of the rest of the page so that hovering over where it will pop up doesn't trigger it

Tested locally

### DIFF
--- a/less/forum/Post.less
+++ b/less/forum/Post.less
@@ -64,7 +64,7 @@
     position: absolute;
     top: -10px;
     left: -100px;
-    z-index: -@zindex-dropdown;
+    z-index: @zindex-dropdown;
     .transition(~"opacity 0.2s, transform 0.2s");
     transform: scale(0.95);
     transform-origin: left top;
@@ -73,7 +73,6 @@
     &.in {
       transform: scale(1);
       opacity: 1;
-      z-index: @zindex-dropdown;
     }
   }
 }
--- a/less/forum/Post.less
+++ b/less/forum/Post.less
@@ -64,7 +64,7 @@
     position: absolute;
     top: -10px;
     left: -100px;
-    z-index: @zindex-dropdown;
+    z-index: -@zindex-dropdown;
     .transition(~"opacity 0.2s, transform 0.2s");
     transform: scale(0.95);
     transform-origin: left top;
@@ -73,6 +73,7 @@
     &.in {
       transform: scale(1);
       opacity: 1;
+      z-index: @zindex-dropdown;
     }
   }
 }


tankerkiller125 (approved):

datitisev:
This behavior is technically still different than previously. Why don’t we use the same code from before?

askvortsov1:
How's the behavior different than before? The card will only trigger when hovering over the appropriate places, will sustain its appearance while hovering within the card, and will disappear when the cursor is moved outside the card's boundaries.

And we can't use the old code because we can't store the card component

datitisev:
With the old code, was the user card in the DOM at all when not showing?

askvortsov1:
Nope, but I'm not sure the benefit of that approach: since we're only using jquery to determine whether it should be showing, we might as well stay consistent with that

clarkwinkelmann:
I'm confused by the approach taken in #2184 and now this...

Actually the UserCard instance was never saved. An intermediary variable was used for the `.component()` call in `view`, but this isn't a case of saving component instances. No change was actually necessary.

My suggestion would be actually reverting #2184 and see later if a change is necessary.

If we still want to go forward with the idea of always rendering the card (I don't like this solution), why not use a solution that relies on `display: none` instead of changing the z-index ? This will also avoid users being able to TAB their way into the card.

askvortsov1:
I tried going at this wirh not including the card in the DOM, but thats an absolute mess when it comes to coordinating the redraw.

With display none, it works, but the css transition fails

clarkwinkelmann:
Right, I got confused between the PostUser and UserCard. I thought we didn't render the PostUser until the first mouseover. But that was incorrect. We were already rendering all PostUser components all the time already. However the PostUser component was the one holding the state about whether to show the UserCard inside of it.

Wouldn't it make sense to move the now-removed `cardVisible` variable from `PostUser` to `CommentPost`, and then pass it down to `PostUser` as a prop to preserve the original behavior?

It seems like it's what Franz tried to say in https://github.com/flarum/core/pull/2184#issuecomment-638481492 with "moving up the hierarchy" but this was not brought up again

askvortsov1:
I dont think `CommentPost` needs cardVisible status at all tbh: that could probably all be contained within PostUser. Within the time I set aside for the task, I was unable to figure out how to trigger redraws at proper times so that the UserCard would be displayed:

EDIT: Actually, I think I just figured out why the cardVisible property was in the subtree retainer: if we set it to only redraw when the editing status changes, then it explicitly won't redraw when the usercard becomes visible. The issue with moving the cardVisible attribute up to CommentPost is that then we can't modify it in PostUser.

clarkwinkelmann:
@askvortsov1 on `transitionend` event in `PostUser`, you could call an `onhide` method that must be passed as prop. Then use that to update the `cardVisible` propery in `CommentPost`?

askvortsov1:
Closing as https://github.com/flarum/core/pull/2224 is better

PR #2210, opened 2020-06-27
Add check to register state of '0' as false for checkboxes

### LINKED ISSUE #1395: Make the Switch/Checkbox components compatible with raw settings values
When using the Switch or Checkbox in a settings modal, saving works fine (sending `true` or `false` to the API)

However when showing the value on page load, the raw settings use values `'0'` or `'1'`, which both evaluate to `true`.

The solution is either to cast the setting at the API level, which you don't usually do in the admin because you pull raw values from the full list of settings, or do it in the front-end with a solution like `state: [true, '1'].indexOf(this.setting('mysetting'))`. Both solutions are very verbose and don't make the code more readable.

Of course storing type information along with settings would solve this issue #1312

But in the meantime maybe the component should be made to accept `'0'` as an "off" value ?

Might be related to #1367 

This follows a discussion on the Flarum Discord.

### LINKED ISSUE #1478: switcher for show language dropdown not correctly displayed

askvortsov1:
**Fixes flarum/framework#1395**
**Fixes flarum/framework#1478**

- If the state supplied to a checkbox is '0', change it to false
- Remove unnecessary conversion in AppearancePage

- Do we still want this workaround? I'm in favor because we might not get to flarum/issue-archive#299 for a while

**Confirmed**

### DIFF
--- a/js/src/admin/components/AppearancePage.js
+++ b/js/src/admin/components/AppearancePage.js
@@ -13,8 +13,8 @@ export default class AppearancePage extends Page {
 
     this.primaryColor = m.prop(app.data.settings.theme_primary_color);
     this.secondaryColor = m.prop(app.data.settings.theme_secondary_color);
-    this.darkMode = m.prop(app.data.settings.theme_dark_mode === '1');
-    this.coloredHeader = m.prop(app.data.settings.theme_colored_header === '1');
+    this.darkMode = m.prop(app.data.settings.theme_dark_mode);
+    this.coloredHeader = m.prop(app.data.settings.theme_colored_header);
   }
 
   view() {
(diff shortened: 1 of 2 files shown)


tankerkiller125 (approved):
I'm fine with this work around for now

clarkwinkelmann (approved):

PR #2211, opened 2020-06-27
Use lifecycle interface for frontend extender

### LINKED ISSUE #1935: Apply lifecycle hooks to Frontend extender (see flarum/core#1581)
Move it.. Needs info.

#1581

askvortsov1:
**Fixes #1935**

- Use lifecycle hooks to recompile frontend assets on extension enable/disable
- Only recompile frontend assets if the extension included css or js

**Confirmed**

### DIFF
--- a/src/Extend/Frontend.php
+++ b/src/Extend/Frontend.php
@@ -80,7 +80,7 @@ class Frontend implements ExtenderInterface, LifecycleInterface
             return;
         }
 
-        $abstract = 'flarum.assets.'.$this->frontend;
+        $abstract = 'flarum.assets.' . $this->frontend;
 
         $container->resolving($abstract, function (Assets $assets) use ($moduleName) {
             if ($this->js) {
@@ -104,7 +104,7 @@ class Frontend implements ExtenderInterface, LifecycleInterface
             }
         });
 
-        if (! $container->bound($abstract)) {
+        if (!$container->bound($abstract)) {
             $container->bind($abstract, function (Container $container) {
                 return $container->make('flarum.assets.factory')($this->frontend);
             });
@@ -114,12 +114,8 @@ class Frontend implements ExtenderInterface, LifecycleInterface
 
             $events->listen(
                 ClearingCache::class,
-                function () use ($container, $abstract) {
-                    $recompile = new RecompileFrontendAssets(
-                        $container->make($abstract),
-                        $container->make(LocaleManager::class)
-                    );
-                    $recompile->flush();
+                function () use ($container) {
+                    $this->recompile($container);
                 }
             );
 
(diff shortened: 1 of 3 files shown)


src/Extend/Frontend.php
```diff
@@ -185,4 +183,26 @@ private function getModuleName(?Extension $extension): string
...
     }
+
+    public function onEnable(Container $container, Extension $extension)
+    {
+        if (! empty($this->js) || ! empty($this->css)) {
+            $recompile = new RecompileFrontendAssets(
+                $container->make('flarum.assets.'.$this->frontend),
+                $container->make(LocaleManager::class)
+            );
+            $recompile->flush();
```
franzliedke, line 194: Could we extract this block into a private method on the extender, please? It's now duplicated very often in this file.

franzliedke:
Oh, if you have the time... would you be willing to spend a bit on experimenting with integration tests for this extender?

askvortsov1:
> Oh, if you have the time... would you be willing to spend a bit on experimenting with integration tests for this extender?

Can we put this into a separate issue and PR?

franzliedke:
It seems that this causes problems, as [discovered in flarum/embed](https://github.com/flarum/embed/pull/10#issuecomment-692885376).

And it's definitely a change in behavior: the previous code listened for the extension `Enabled` / `Disabled` events, which is fired when *any* extension is enabled or disabled. Now, we run this code in the lifecycle hooks, which are only fired when the extension that uses this extension is enabled or disabled.

@askvortsov1 Was that intended?

@clarkwinkelmann Not 100% sure how this maps to the behavior you were observing

clarkwinkelmann:
Yeah looking at this code I have no idea how this matches with the issue I was experiencing, but if say so it would seem a plausible source for the issue.

I'm not really sure what this PR did. Was it needed at all? As you say this changes the behavior from Flarum listening for extensions being enabled to extensions triggering something in Flarum via an extender they might or might not use.

clarkwinkelmann:
Ok so here's the problem. It's also breaking my Wordpress extension.

What was happening before:

- Whenever any extension is enabled or disabled, we clear the compiled assets of all frontends currently in use on the forum

What is happening after this PR:

- Whenever an extension is enabled or disabled, the assets for the frontends used by this extension are cleared

Clearing the assets for `flarum.frontend.forum` and `flarum.frontend.admin` was done through `ForumServiceProvider` and `AdminServiceProvider` respectively and they would be cleared for all extensions, even if they are not registering any asset in that frontend. I think this makes sense because other extensions could modify their assets based on which other extensions are enabled.

The code that was in `Extend\Frontend` was only inside of `!$container->bound($abstract)` and was only made to handle new frontends that extensions might define, like we do in the Embed extension and like I do in the Wordpress extension.

The reason for the error is that previously, the assets would be cleared when any **other** extension was enabled/disabled, so this meant the extension that created any given frontend would always be active when their frontend assets were cleared.

However with the new logic using `onEnable`, Flarum tries clearing the assets for the extension itself in the API request that enabled it, and in that request only the lifecycle extenders run, not the normal extenders. Which means `extend` never gets called, which means the frontend isn't available in the container.

I suggest we revert all of this. The behavior radically changed from what it was before and I see no benefit for the new system.

There's actually one thing that we should maybe consider an issue with the old behavior, but I'll open a separate issue for that if I can confirm it's really a problem. If an extension is the only one using a custom frontend name, that frontend won't be cleared when the extension itself is enabled. Usually the frontend will have been cleared when the extension was disabled, but if it was disabled by editing the database, enabling it again won't clear the existing assets it seems.

askvortsov1:
I agree about reverting it. @flarum/core any objections?

PR #2213, opened 2020-06-27
Drop user bio column

askvortsov1:
**Fixes #1280 *

DO NOT MERGE UNTIL STABLE RELEASE

PR #2215, opened 2020-06-28
Don't create user bio column

### LINKED ISSUE #1280: Remove user bio column from database

askvortsov1:
Fixes #1280 
Replaces #2213 

See https://github.com/flarum/core/issues/1280#issuecomment-650655900

### DIFF
--- a/migrations/2015_02_24_000000_create_users_table.php
+++ b/migrations/2015_02_24_000000_create_users_table.php
@@ -18,7 +18,6 @@ return Migration::createTable(
         $table->string('email', 150)->unique();
         $table->boolean('is_activated')->default(0);
         $table->string('password', 100);
-        $table->text('bio')->nullable();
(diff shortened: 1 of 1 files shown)


clarkwinkelmann:
That's kind of an illicit move in the migration world :stuck_out_tongue: 

As long as we coordinate this with the bio extension so that no one ends up with no column at all, I'm ok with the approach.

askvortsov1:
Although it's not the best practice, I feel like this is a better and cleaner option than my first approach. But yeah, might be nice to see fof adding a conditional migration that adds the column if it doesn't exist before merging this.

askvortsov1:
Fof change opened in https://github.com/FriendsOfFlarum/user-bio/pull/19/files

PR #2219, opened 2020-06-29
Switch from 'moment' to 'dayjs'

datitisev:
Almost exactly the same as https://github.com/flarum/core/commit/d81cceed7d4642b61ad19fe332c15a73aa5c9b98.
I changed the `User.prototype.isOnline` method to not use greater than and instead use the `dayjs` query `isAfter`.

- `forum.js`: 361 KB ----> 318 KB (-12%)
    - `forum.js.map`: 1.5 MB ----> 1.3 MB (-13%)
- `admin.js`: 280 KB ----> 240 KB (-14%)
    - `admin.js.map`: 1.2 MB ----> 956 KB (-26%)

Total bundle size reduction by 85 KB across forum & admin (excluding source maps).

`moment` will still mostly work, it now is the same as `dayjs`.

The only big breaking change I think is that `moment.duration` is no longer a thing.

In PostStream:
image

<details>
<summary>Source map comparison before & after</summary>

Before:
image

After:
image
</details>


datitisev:
The moment config in language packs should continue functioning for the most part.

With no additional changes, Dutch language pack working just fine.

image

datitisev:
Looks like there is [an official plugin](https://day.js.org/docs/en/plugin/duration) for duration. Do we want/need this?

It'd add ~4 KB back.

askvortsov1:
Are we using duration?

datitisev:
@askvortsov1 We were for time lapsed. Some other extensions use it as well. Not sure if we want it because it may have differences in some locales... the function I replaced it with in `PostStream` basically does the same thing as far as I can tell.

askvortsov1:
Alright I'm fine not adding it then :+1:

js/src/common/utils/humanTime.js
```diff
@@ -6,8 +6,8 @@
  * @return {String}
  */
 export default function humanTime(time) {
-  let m = moment(time);
-  const now = moment();
+  let m = dayjs(time);
```
askvortsov1, line 9: Hmm actually one super quick nitpick: perhaps this should be called `d` instead of `m`?

datitisev: It really doesn't matter. It can be changed.

tankerkiller125: m could be considered a backwards compatible thing I think (if extensions use it), but I agree that it could be confusing long term.

js/webpack.config.js
```diff
@@ -1,12 +1,8 @@
 const config = require('flarum-webpack-config');
-const webpack = require('webpack');
```
askvortsov1, line 2: Does day need to be added here?

franzliedke:
I've replaced `moment` with `dayjs` in the two bundled extensions (statistics, suspend) that used it. We should consider deprecating the `moment` alias for stable.

PR #2225, opened 2020-07-02
Fix discussions user page

askvortsov1:


datitisev:
What exactly does this fix?

askvortsov1:
The discussions user page is currently broken on master, since all the logic for fetching discussions from the DB was moved to discussion list state

PR #2236, opened 2020-07-21
Fixes wrong IP address when using a reverse proxy

x7airworker:
Two new **optional** config values: 
- `reverse_proxy.enabled` will enable ip address finding of a user using a reverse proxy at `\Http\Middleware\ProxyAddress`
- `reverse_proxy.allowed` controls the allowed reverse proxy IPs used for checking.

Changed `Arr::get($request->getServerParams(), 'REMOTE_ADDR', '127.0.0.1');` to `$request->getAttribute('ipAddress');` to use the new attribute in the codebase.

- The Middleware is added in` \Flarum\Foundation\InstalledApp` and im unsure if this is the correct/best position.

**Confirmed**

- [x] Backend changes: 2 failures (WritablePathsTest which is not related to the changes I made).

**Required changes:**

- [x] Related documentation PR: TODO (Would be good to document the optional config values.)

### DIFF
--- a/src/Api/ApiServiceProvider.php
+++ b/src/Api/ApiServiceProvider.php
@@ -67,7 +67,7 @@ class ApiServiceProvider extends AbstractServiceProvider
         });
 
         $this->app->bind('flarum.api.proxy_middleware', function () {
-            $config = $this->app->get('flarum.config');
+            $config = $this->app->make('flarum.config');
 
             return new HttpMiddleware\ProxyAddress(
                 Arr::get($config, 'reverse_proxy.enabled', false),
--- a/src/Api/ApiServiceProvider.php
+++ b/src/Api/ApiServiceProvider.php
@@ -66,7 +66,7 @@ class ApiServiceProvider extends AbstractServiceProvider
             );
         });
 
-        $this->app->bind('flarum.api.proxy_middleware', function() {
+        $this->app->bind('flarum.api.proxy_middleware', function () {
             $config = $this->app->get('flarum.config');
 
             return new HttpMiddleware\ProxyAddress(
--- a/src/Api/ApiServiceProvider.php
+++ b/src/Api/ApiServiceProvider.php
@@ -22,6 +22,7 @@ use Flarum\Http\Middleware as HttpMiddleware;
 use Flarum\Http\RouteCollection;
 use Flarum\Http\RouteHandlerFactory;
 use Flarum\Http\UrlGenerator;
+use Illuminate\Support\Arr;
 use Laminas\Stratigility\MiddlewarePipe;
 
 class ApiServiceProvider extends AbstractServiceProvider
@@ -45,6 +46,7 @@ class ApiServiceProvider extends AbstractServiceProvider
         $this->app->singleton('flarum.api.middleware', function () {
             return [
                 'flarum.api.error_handler',
+                'flarum.api.proxy_middleware',
                 HttpMiddleware\ParseJsonBody::class,
                 Middleware\FakeHttpMethods::class,
                 HttpMiddleware\StartSession::class,
@@ -61,6 +63,15 @@ class ApiServiceProvider extends AbstractServiceProvider
                 $this->app->make(Registry::class),
                 new JsonApiFormatter($this->app['flarum']->inDebugMode()),
                 $this->app->tagged(Reporter::class)
+            );
+        });
+
+        $this->app->bind('flarum.api.proxy_middleware', function() {
+            $config = $this->app->get('flarum.config');
+
+            return new HttpMiddleware\ProxyAddress(
+                Arr::get($config, 'reverse_proxy.enabled', false),
+                Arr::get($config, 'reverse_proxy.allowed', ['127.0.0.1'])
             );
         });
 
--- a/src/Foundation/InstalledApp.php
+++ b/src/Foundation/InstalledApp.php
@@ -59,12 +59,6 @@ class InstalledApp implements AppInterface
         $pipe = new MiddlewarePipe;
 
         $pipe->pipe(new BasePath($this->basePath()));
-        $pipe->pipe(
-            new ProxyAddress(
-                Arr::get($this->config, 'reverse_proxy.enabled', false),
-                Arr::get($this->config, 'reverse_proxy.allowed', ['127.0.0.1'])
-            )
-        );
         $pipe->pipe(new OriginalMessages);
         $pipe->pipe(
             new BasePathRouter([
--- a/src/Foundation/InstalledApp.php
+++ b/src/Foundation/InstalledApp.php
@@ -11,11 +11,9 @@ namespace Flarum\Foundation;
 
 use Flarum\Foundation\Console\InfoCommand;
 use Flarum\Http\Middleware\DispatchRoute;
-use Flarum\Http\Middleware\ProxyAddress;
 use Flarum\Settings\SettingsRepositoryInterface;
 use Illuminate\Console\Command;
 use Illuminate\Contracts\Container\Container;
-use Illuminate\Support\Arr;
 use Laminas\Stratigility\Middleware\OriginalMessages;
 use Laminas\Stratigility\MiddlewarePipe;
 use Middlewares\BasePath;
--- a/src/Foundation/InstalledApp.php
+++ b/src/Foundation/InstalledApp.php
@@ -11,9 +11,11 @@ namespace Flarum\Foundation;
 
 use Flarum\Foundation\Console\InfoCommand;
 use Flarum\Http\Middleware\DispatchRoute;
+use Flarum\Http\Middleware\ProxyAddress;
 use Flarum\Settings\SettingsRepositoryInterface;
 use Illuminate\Console\Command;
 use Illuminate\Contracts\Container\Container;
+use Illuminate\Support\Arr;
 use Laminas\Stratigility\Middleware\OriginalMessages;
 use Laminas\Stratigility\MiddlewarePipe;
 use Middlewares\BasePath;
@@ -57,6 +59,11 @@ class InstalledApp implements AppInterface
         $pipe = new MiddlewarePipe;
 
         $pipe->pipe(new BasePath($this->basePath()));
+        $pipe->pipe(new ProxyAddress(
+                Arr::get($this->config, 'reverse_proxy.enabled', false),
+                Arr::get($this->config, 'reverse_proxy.allowed', ['127.0.0.1'])
+            )
+        );
         $pipe->pipe(new OriginalMessages);
         $pipe->pipe(
             new BasePathRouter([
--- a/src/Foundation/InstalledApp.php
+++ b/src/Foundation/InstalledApp.php
@@ -59,7 +59,8 @@ class InstalledApp implements AppInterface
         $pipe = new MiddlewarePipe;
 
         $pipe->pipe(new BasePath($this->basePath()));
-        $pipe->pipe(new ProxyAddress(
+        $pipe->pipe(
+            new ProxyAddress(
                 Arr::get($this->config, 'reverse_proxy.enabled', false),
                 Arr::get($this->config, 'reverse_proxy.allowed', ['127.0.0.1'])
             )
--- a/src/Http/Exception/ProxyNotAllowedException.php
+++ b/src/Http/Exception/ProxyNotAllowedException.php
@@ -1,0 +1,21 @@
+<?php
+
+/*
+ * This file is part of Flarum.
+ *
+ * For detailed copyright and license information, please view the
+ * LICENSE file that was distributed with this source code.
+ */
+
+namespace Flarum\Http\Exception;
+
+use Exception;
+use Flarum\Foundation\KnownError;
+
+class ProxyNotAllowedException extends Exception implements KnownError
+{
+    public function getType(): string
+    {
+        return 'reverse_proxy_not_allowed';
+    }
+}
--- a/src/Http/Middleware/ProxyAddress.php
+++ b/src/Http/Middleware/ProxyAddress.php
@@ -61,7 +61,7 @@ class ProxyAddress implements Middleware
                 $ipAddress = Arr::get($serverParams, 'HTTP_CLIENT_IP', $ipAddress);
                 $ipAddress = Arr::get($serverParams, 'X_PROXYUSER_IP', $ipAddress);
             } else {
-                throw new ProxyNotAllowedException("The used proxy isn't allowed to connect!");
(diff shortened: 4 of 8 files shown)


src/Foundation/InstalledApp.php
```diff
@@ -57,6 +59,12 @@ public function getRequestHandler()
         $pipe = new MiddlewarePipe;
 
         $pipe->pipe(new BasePath($this->basePath()));
+        $pipe->pipe(
```
tankerkiller125, line 62: I would prefer to see the Middleware changed made in https://github.com/flarum/core/blob/master/src/Api/ApiServiceProvider.php and https://github.com/flarum/core/blob/master/src/Forum/ForumServiceProvider.php following a similiar method that the error handler Middleware is implemented to ensure that it can be overridded if needed by extension devs.

You can see how the error handler is implemented in the ApiServiceProvider file on line 59 (actual call/binding) and 47 (adding to middleware array), The ForumServiceProvider has similar logic (possibly even the same logic)

x7airworker: I definitely agree with this change.
Do I also need to include the middleware in
https://github.com/flarum/core/blob/master/src/Admin/AdminServiceProvider.php ?

tankerkiller125: I would say yes to ensure that IPs are tracked properly for that, I don't think core uses IPs in anyway for admin, but I know a couple extensions might. So I'd include it there too.

src/Api/ApiServiceProvider.php
```diff
@@ -64,6 +66,15 @@ public function register()
             );
         });
 
+        $this->app->bind('flarum.api.proxy_middleware', function () {
+            $config = $this->app->get('flarum.config');
```
askvortsov1, line 70: Shouldn't this be $this->app->make, so that we can fully use DI?

src/Http/Exception/ProxyNotAllowedException.php
```diff
@@ -0,0 +1,21 @@
...
+ * For detailed copyright and license information, please view the
+ * LICENSE file that was distributed with this source code.
+ */
+
+namespace Flarum\Http\Exception;
+
+use Exception;
+use Flarum\Foundation\KnownError;
+
+class ProxyNotAllowedException extends Exception implements KnownError
```
askvortsov1, line 15: We might want to have a custom error page for this, so that it's clear why the forum can't be accessed (otherwise it'd be the standard "an error has occured" page)

src/Http/Middleware/ProxyAddress.php
```diff
@@ -0,0 +1,59 @@
...
+
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
+    {
+        $ipAddress = Arr::get($request->getServerParams(), 'REMOTE_ADDR', '127.0.0.1');
+
+        if ($this->enabled) {
+            if (in_array($ipAddress, $this->allowedAddresses)) {
+                // standard header for proxies, see: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For
+                $ipAddress = Arr::get($request->getServerParams(), 'X_FORWARDED_FOR', $ipAddress);
+                $ipAddress = Arr::get($request->getServerParams(), 'HTTP_CLIENT_IP', $ipAddress);
```
askvortsov1, line 49: Is there a reason why these are in the order they are in?

src/Http/Middleware/ProxyAddress.php
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
+    {
+        $ipAddress = Arr::get($request->getServerParams(), 'REMOTE_ADDR', '127.0.0.1');
+
+        if ($this->enabled) {
+            if (in_array($ipAddress, $this->allowedAddresses)) {
+                // standard header for proxies, see: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For
+                $ipAddress = Arr::get($request->getServerParams(), 'X_FORWARDED_FOR', $ipAddress);
```
askvortsov1, line 48: Do we want to make this list of headers extensible? I'm leaning towards no, but would like to hear @clarkwinkelmann @franzliedke's thoughts

askvortsov1:
Overall looks good, left a few comments. Also, I'm not seeing any failing test cases (re: PR description), so we're good on that front :D

src/Http/Middleware/ProxyAddress.php
x7airworker: I've used the order "cave bro" used in his basic implementation.

src/Api/ApiServiceProvider.php
x7airworker: Idk, but get works fine. What would be the benefit of using make?

src/Http/Exception/ProxyNotAllowedException.php
x7airworker: A normal user wouldn't ever get this error if the server configuration is correct, so I think no custom error page is needed. 
This is a just a security mechanism.

x7airworker:
> Overall looks good, left a few comments. Also, I'm not seeing any failing test cases (re: PR description), so we're good on that front :D

I think the failed test cases had to do with my environment as I've written in the description.

src/Http/Middleware/ProxyAddress.php
x7airworker: Because "X_FORWARDED_FOR" is a standard header used by all leading web servers for proxies, I think that an extensible headers list is not needed.

src/Api/ApiServiceProvider.php
askvortsov1: Looking at https://github.com/illuminate/container/blob/b5cfeb3b9e9c0e12f663b4101865762d14f795ee/Container.php#L639, make allows for additional parameters (although they aren't needed here), but both ultimately call resolve. That being said, let's stick with make for consisteny purposes since that's what's used across the codebase.

src/Http/Middleware/ProxyAddress.php
askvortsov1: That's a fair point (and anyone who needs a different usage can remove/replace the middleware),

src/Http/Middleware/ProxyAddress.php
x7airworker: Also found a more appropriate source: https://serverfault.com/a/920060

askvortsov1: Might it also be worth adding X-ProxyUser-Ip as per the linked documentation (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For)? Also (asking as I haven't dealt with this myself), do you know of any big proxies that might set multiple of these headers?

src/Http/Middleware/ProxyAddress.php
```diff
@@ -0,0 +1,59 @@
...
+                $ipAddress = Arr::get($request->getServerParams(), 'HTTP_CLIENT_IP', $ipAddress);
+            } else {
+                throw new ProxyNotAllowedException("The used proxy isn't allowed to connect!");
```
askvortsov1, line 51: We generally like to avoid hardcoding text in any particular language where possible. In this case though, is the exception message really neccessary? Injecting a translator instance seems... excessive when this exception is only raised in one instance.

src/Http/Exception/ProxyNotAllowedException.php
askvortsov1: That... is absolutely correct...

src/Http/Middleware/ProxyAddress.php
```diff
@@ -0,0 +1,59 @@
...
+     */
+    public function __construct($enabled, $allowedAddresses)
+    {
+        $this->enabled = $enabled;
+        $this->allowedAddresses = $allowedAddresses;
+    }
+
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
+    {
+        $ipAddress = Arr::get($request->getServerParams(), 'REMOTE_ADDR', '127.0.0.1');
```
askvortsov1, line 43: Can we be confident that reverse proxies would always either set this to 127.0.0.1 (if set at all) or leave it blank? This question also largely comes from inexperience working with reverse proxies, but it would be nice to avoid unnecessary issues.

src/Http/Middleware/ProxyAddress.php
```diff
@@ -0,0 +1,59 @@
...
+
+        if ($this->enabled) {
+            if (in_array($ipAddress, $this->allowedAddresses)) {
```
askvortsov1, line 46: Perhaps we should support wildcards here?

src/Http/Middleware/ProxyAddress.php
x7airworker: While I agree that the "Forwarded" header is the new standard and we should probably support it, I have a few concerns:
1. The header has a key-value map which would we need to parse. Maybe PHP has that already built-in?
2. I doubt that proxies will ever stop sending headers that were used for this purpose before, just because of support for older applications.

x7airworker: Yeah we sure can add this header.

src/Http/Middleware/ProxyAddress.php
tankerkiller125: The only service/Proxy that I can think of that set's multiple headers at the same time is Cloudflare, but that's the X-Forwarded-For header and then a custom Cloudflare-IP header if I remember correctly. Officially there are only 3 proxy headers that are standard/matter for IPs and those are:

- Forwarded (not part of the HTTP standards, but is well known and used commonly)
- X-Forwarded-For
- Via

src/Http/Middleware/ProxyAddress.php
tankerkiller125: Technically the `REMOTE_ADDR` will never be blank, it will either contain the IP of the client requesting data, or it will contain the IP of the proxy server forwarding the request. It's important that this be the first in the options so that it can be overridden with a correct proxy value if a proxy is used. If this value where blank then there would be a MUCH larger issue with the server.

src/Http/Middleware/ProxyAddress.php
tankerkiller125: I actually agree, wildcards would be great, even better if we could find a way to support CIDR notation easily as something like `169.254.56.*` might include multiple ISPs, services, or companies. A CIDR is much more accurate.

(Note: Example IP is non-routable and safe for example use)

src/Http/Middleware/ProxyAddress.php
x7airworker: > We generally like to avoid hardcoding text in any particular language where possible. In this case though, is the exception message really neccessary? Injecting a translator instance seems... excessive when this exception is only raised in one instance.

The exception message is not really neccessary, the exception name itself might be enough.

src/Http/Middleware/ProxyAddress.php
askvortsov1: As long as we support some form of wildcard so that if a reverse proxy uses a non-small set of possible options for REMOTE_ADDR we can support that, we should be fine.

src/Http/Middleware/ProxyAddress.php
```diff
@@ -0,0 +1,71 @@
...
+    public function __construct($enabled, $allowedAddresses)
+    {
+        $this->enabled = $enabled;
+        $this->allowedAddresses = $allowedAddresses;
+    }
+
+    private function wildcardMatch(string $ipAddress): bool
+    {
+        foreach ($this->allowedAddresses as $allowedAddress) {
+            if (fnmatch($allowedAddress, $ipAddress)) {
```
tankerkiller125, line 44: My only concern with `fnmatch` is that it's unavailable for Windows systems. I know that Windows is a very, very tiny part of our community but I think it's still something we would want to be able to support.

askvortsov1: We already use fnmatch for csrf checks though.

askvortsov1: Also, as per https://www.php.net/manual/en/function.fnmatch.php, it should be available?

x7airworker: I did develop and test this function on Windows with XAMPP, so it definitely works.

src/Http/Middleware/ProxyAddress.php
askvortsov1: Yeah let's go that route, thanks!

askvortsov1: @x7airworker I've been thinking some more about this, and I'm still a bit uneasy with hardcoding the headers so definitively (that stackoverflow link you shared also had some alternative options). Maybe it might be worth allowing a list of headers to be provided in the config.php, but defaulting to these 3? Of course, if the Forwarded RFC becomes standard, this will need to be revisited, but I don't think that's something we should worry about right now.

Aside from the exception message, I think that's my last comment on this, after that it should be good to merge. Thank you so much for implementing this, and I definitely hope to see more of your PRs in the future :grin:

src/Http/Middleware/ProxyAddress.php
tankerkiller125: AH my fault, I read the warning incorrectly, I thought it said that it was available to non-POSIX compliant systems except Windows, reading it again I realize I had that backwards.

src/Http/Middleware/ProxyAddress.php
x7airworker: Right now I've implemented the biggest 3, if we've missed something, someone will sometime open an issue about some edge case header.
And the middleware can still be overwritten by extensions.

askvortsov1: Alright, that's fair! Not something worth bikeshedding about :D

tankerkiller125:
@franzliedke the problem I had with putting the middleware at the highest level is that the extender for middleware does not have the ability to overwrite middleware in IntalledApp I don't think.

franzliedke:
I don't see a big need for extensibility here, personally. It's security-relevant and the middleware can be disabled by changing the config.

But honestly, that is the least of my concerns, just an implementation detail.

askvortsov1:
Considering our backlog for beta 14, it might be best to revert it, and encourage @x7airworker to open a new PR that takes into account Franz's comments above. I don't want to make this something that might hold back the release.

tankerkiller125:
I agree @askvortsov1 if you want to rollback I think that's a good decision at the moment. I would but I'm at work :(

PR #2243, opened 2020-07-24
Upgrade to Laravel 6, finally!

### LINKED ISSUE #2055: Upgrade Illuminate components to 6.x

franzliedke:
**Fixes #2055**

This was easier than [anticipated](https://github.com/flarum/core/issues/2055#issuecomment-646889234). :smile: 
Luckily, we are only using the translator using the interface from Symfony. We only need to fulfill Laravel's contract to be able to pass something into its validator factory.

- Update Illuminate components to current LTS release (v6)
- Fix a few remaining global helper calls (removed in v6) and replace them by static calls
- Unify typehints for dependency injection of translator instances
- Satisfy updated `Translator` contract (from Laravel) in our implementation - only needed for the validation component

**Confirmed**


franzliedke:
@tankerkiller125 offered to take care of replacing `array_get` (and possibly other old helpers?) with the class-based variant in the bundled extensions. That's a blocker for merging this PR, so we can leave that to him. :grin:

askvortsov1:
On the contrary, perhaps it would be better to merge this PR, break bundled extensions on dev/master, and that way, cleaning up bundled extensions would be easier?

franzliedke:
It's a pretty easy find-and-replace operation (and safer than relying on hitting all code paths with clicking through the app or relying on non-existent tests).

tankerkiller125:
@askvortsov1 I'll have all the core extensions updated this weekend anyway and I'll be committing directly to master, (no point in PRing a simple helper function to class function PR) so honestly a couple days at most wait to merge isn't that bad.

src/Locale/Translator.php
```diff
@@ -17,6 +17,16 @@ class Translator extends BaseTranslator implements TranslatorContract
 {
     const REFERENCE_REGEX = '/^=>\s*([a-z0-9_\-\.]+)$/i';
 
+    public function get($key, array $replace = [], $locale = null)
```
askvortsov1, line 20: Can we do this the other way around, and have the logic in Laravel's new methods, and have the old methods reference the new ones?

franzliedke: The "old" ones come from the Symfony base implementation.

askvortsov1: The reason I bring this up is that if we eventually want to comply with the laravel contract, we might as well move the logic to laravel's methods. If we want to eventually drop the laravel translation contract, this is fine though.

tankerkiller125:
Core extensions updated

src/Locale/Translator.php
franzliedke: I don't get what you mean, sorry. `trans()` and `transChoice()` are implemented in Symfony's translator, and we don't overwrite these. And since we're mostly/only using Symfony's interface throughout the app, I'd rather keep their implementation as well. These two methods are only needed as a bridge to be able to use this Symfony-based translator in Laravel's validation component.

PR #2246, opened 2020-07-28
Fix optional parameters in url generator

### LINKED ISSUE #1972: Optional route parameters cannot be filled via the url builder
## Bug Report

**Current Behavior**

    app(UrlGenerator::class)->to('forum')->route('discussion', ['id' => 1]);
    // http://flarum.localhost/d/1

    app(UrlGenerator::class)->to('forum')->route('discussion', ['id' => 1, 'near' => 5]);
    // http://flarum.localhost/d/1

**Expected Behavior**

    app(UrlGenerator::class)->to('forum')->route('discussion', ['id' => 1]);
    // http://flarum.localhost/d/1

    app(UrlGenerator::class)->to('forum')->route('discussion', ['id' => 1, 'near' => 5]);
    // http://flarum.localhost/d/1/5

**Possible Solution**

This feature can probably be implemented without too many troubles inside `RouteCollection`.

Currently all variations of the route (first with a single parameter, then with two parameters) are parsed by `fast-route`, and added to both the routing and the `$reverse` array.

The limitation is that we currently always take the first route matching the name from the `$reverse`, whereas the one with the optional parameter is likely in second place.

https://github.com/flarum/core/blob/d492579638fb52dafbfe65f1f36a95eb6047f7f3/src/Http/RouteCollection.php#L91

We should probably try replacing the parameters in each of the `$this->reverse[$name]` matches, then return the one where we were able to replace the most tokens.

Or alternatively, index the reverse array based not only on name, but also on the parameter names or count in order to pick the correct reverse route in a single access.

Once implemented, we should use the route builder in places where it's currently skipped for this very reason https://github.com/flarum/subscriptions/blob/master/views/emails/newPost.blade.php#L6

**Additional Context**

Discussed on the forum here https://discuss.flarum.org/d/22573-how-to-navigate-to-a-particular-post-in-flarum

askvortsov1:
**Fixes #1972**

- Select the most fitting paths option instead of the first one to support using url generator with optional parameters
- Add unit tests for this functionality

- Any shortcomings in the algorithm I used?
- Any code style recommendations? The variable naming I feel a bit iffy on

**Confirmed**

- [x[ Backend changes: tests are green (run `composer test`).

### DIFF
--- a/src/Http/RouteCollection.php
+++ b/src/Http/RouteCollection.php
@@ -11,6 +11,7 @@ namespace Flarum\Http;
 
 use FastRoute\DataGenerator;
 use FastRoute\RouteParser;
+use Illuminate\Support\Arr;
 
 class RouteCollection
 {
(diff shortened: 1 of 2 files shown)


src/Http/RouteCollection.php
```diff
@@ -88,7 +88,23 @@ protected function fixPathPart(&$part, $key, array $parameters)
...
         if (isset($this->reverse[$name])) {
+            $curr_max_count = 0;
             $parts = $this->reverse[$name][0];
+
+            // For a given route name, we want to choose the option that best matches the given parameters.
+            // Each routing option is an array of parts. Each part is either a constant string
+            // (which we don't care about here), or an array where the first element is the parameter name
+            // and the second element is a regex into which the parameter value is inserted, if the parameter matches.
+            foreach ($this->reverse[$name] as $parts_option) {
+                for ($i = 0; $i < count($parts_option); $i++) {
```
clarkwinkelmann, line 99: Wouldn't `foreach ($parts_option => $i => $part)` replace the two lines here?

src/Http/RouteCollection.php
```diff
@@ -88,7 +88,23 @@ protected function fixPathPart(&$part, $key, array $parameters)
...
+                    $part = $parts_option[$i];
+                    if (is_array($part) && in_array($part[0], array_keys($parameters)) && $i > $curr_max_count) {
```
clarkwinkelmann, line 101: Maybe replace `in_array($part[0], array_keys($parameters))` with `array_key_exists($part[0], $parameters))` or even the shorter `Arr::exists($parameters, $part[0])` ?

tankerkiller125:
I'll wait to review until the things clark pointed out are fixed. Like clark I also have difficulty reading the actual implementation so some clarification of variables would be great.

askvortsov1:
Oh boy this code is a mess, not sure what I was thinking when I wrote it... I did a bit of cleanup, that was embarrasing...

Issue #1847, opened 2019-08-13
Unable to access request actor in error handler


datitisev:
Introduced in #1843 which introduces an alternative instead of middleware for error handling.

> The "problem" here is that PSR-7 requests and responses are immutable. The ErrorHandler middleware basically passes on the request it received from the outside world to inner middleware. One of these middlewares extracts cookies, headers etc. from the response to determine the actor, which is then stored as a "request attribute" on the request. Once you add a new request attribute, you're creating a new request instance, though - which means the request instance known by the ErrorHandler is unchanged. This was a design decision when PSR-7 was created.

**TL;DR**: Request object passed to error handler is not the same instance as the one passed to middleware. Actor is unavailable.

franzliedke:
My plan is the following:
- Add a simple, single middleware *before* the error handler middleware that injects a `ActorReference` (optionally with a better name) into the request as an attribute. Because that reference is an object, it will be the same across middleware even when they receive a different request instance.
- As the name suggests (does it?), the `ActorReference` is mutable. Later middleware (e.g. `AuthenticateWithSession`) can call `setActor()` on the reference once they do authentication.
- From that point on, all middleware - most importantly the error handler - can access the actor from the reference, even if it was added later in the stack. (Note: For early middleware like the error handler, the actor will only be available after getting a response from other middleware, not when passing a request in to the stack.)

@datitisev Does this make sense? It should solve your problem, right?

At first I thought: "Woah, this is just a huge hack". But it could turn out to be quite useful, as this could even be used to inject an actor into Flarum e.g. when building integrations with an existing website's authentication system.

datitisev:
@franzliedke As far as I can tell, yes, it makes sense and should work. Thanks for writing out the plan 🙂.

askvortsov1:
This works well with https://github.com/flarum/core/issues/869#issuecomment-612838885. We can keep both systems (actor as direct attribute, and ActorReference) at the start, then deprecate actor as direct attribute, while encouraging people to use `RequestUtil::actorFromRequest($request)` as pseudo-encapsulation.

PR #2249, opened 2020-07-29
Handle 401, 403, and 404 errors in frontend

### LINKED ISSUE #206: Handle 404, 403, 410, 401 exceptions in frontend

### LINKED ISSUE #2214: Handle 404, 403, 410, 401 exceptions in frontend
#Feature Request

**Replaces #1305**
**Replaces #654**

Instead of trying to hack in a search box or login field, we should take advantage of the great frontend we have now. 

The primary challenge here is getting the request and it's attributes (actor, locale, etc) to the error handler. This doesn't work by default because $request->withAttribute clones and returns a new ServerRequest object. However, we can be confident that any exceptions that fall under this category will be thrown in the DispatchRoutes level of the middleware. I've come up with 3 potential solutions to the above challenge:

1. In DispatchRoutes.php, store the request instance as it is received somewhere globally accessible, then fetch it from ErrorHandler if an exception does occur.
2. Add another ErrorHandler middleware layer right before DispatchRoutes. The request supplied here will have actor, locale, etc set
3. Re-engineer ServerRequest->withAttribute so that it modifies itself instead of returning a new object
4. Pass in the request to exceptions (this isn't really even an option because most places where exceptions are thrown don't have access to the request object.

To be honest, I'm not a huge fan of any of these (and would love to consider alternatives), but (2) seems like the best out of them.

askvortsov1:
**Fixes flarum/issue-archive#206**

Handle 401, 403, and 404 errors in frontend (see commits for specific granular changes)

- See GH review comments

**Confirmed**


src/Http/Content/NotAuthenticated.php
```diff
@@ -0,0 +1,24 @@
...
+
+use Flarum\Frontend\Document;
+use Psr\Http\Message\ServerRequestInterface as Request;
+
+class NotAuthenticated
+{
+    public function __invoke(Document $document, Request $request)
+    {
+        $document->title = 'Not Authenticated';
+        $document->payload['errorCode'] = 103;
```
franzliedke, line 20:
```suggestion
        $document->payload['errorCode'] = 401;
```

src/Http/Middleware/HandleErrors.php
```diff
@@ -56,6 +62,24 @@ public function process(Request $request, Handler $handler): Response
...
+                    if ($e instanceof $errorClass) {
+                        $handled = true;
+                        break;
+                    }
+                }
+
+                if (! $handled) {
+                    throw $e;
+                }
+            }
```
franzliedke, line 82: This seems to re-implement logic similar to what the registry already does, and the meaning of the additional parameter doesn't seem very intuitive.

js/src/forum/components/ErrorPage.js
```diff
@@ -0,0 +1,13 @@
+import Page from '../../common/components/Page';
+
+export default class NotAuthenticatedErrorPage extends Page {
```
franzliedke, line 3: This doesn't match the filename.

js/src/forum/routes.js
```diff
@@ -24,6 +25,8 @@ export default function (app) {
 
     settings: { path: '/settings', component: SettingsPage.component() },
     notifications: { path: '/notifications', component: NotificationsPage.component() },
+
+    '400': { path: '/:400', component: ErrorPage.component() },
```
franzliedke, line 29: Why 400? This also seems to be pre-Mithril-2?

src/Foundation/ErrorHandling/FrontendFormatter.php
```diff
@@ -0,0 +1,71 @@
...
+    /**
+     * @var SettingsRepositoryInterface
+     */
+    protected $settings;
+
+    public function __construct(Container $container, TranslatorInterface $translator, SettingsRepositoryInterface $settings)
+    {
+        $this->container = $container;
+        $this->translator = $translator;
+        $this->settings = $settings;
```
franzliedke, line 54: These aren't used.

src/Foundation/ErrorHandling/FrontendFormatter.php
```diff
@@ -0,0 +1,71 @@
...
+ * A formatter for turning caught exceptions into "pretty" HTML error pages.
+ *
+ * For certain known error types, we display pages with dedicated information
+ * relevant to this class of error, e.g. a page with a search form for HTTP 404
+ * "Not Found" errors. We look for templates in the `views/error` directory.
+ *
+ * If no specific template exists, a generic "Something went wrong" page will be
+ * displayed, optionally enriched with a more specific error message if found in
+ * the translation files.
+ */
```
franzliedke, line 32: Given that this behaves differently than the `ViewFormatter`, it should likely have a different docblock. :wink:

src/Foundation/ErrorHandling/FrontendFormatter.php
```diff
@@ -0,0 +1,71 @@
...
+    public function __construct(Container $container, TranslatorInterface $translator, SettingsRepositoryInterface $settings)
+    {
+        $this->container = $container;
+        $this->translator = $translator;
+        $this->settings = $settings;
+    }
+
+    public function format(HandledError $error, Request $request): Response
+    {
+        $frontend = $this->container->make('flarum.frontend.forum');
```
franzliedke, line 59: Since this is already explicitly instantiated in the service provider, it would be preferable to inject the actual dependency, instead of the container. This just hides dependencies.

franzliedke (commented):
I already expressed my doubts about this approach on the issue, but I'd like to add a few general concerns:

- Again, first and foremost, I can only recommend to avoid complexity like this on error pages - from painful experience.
- This would render the forum frontend for any 404, even on API and admin, if I am not mistaken.
- "Model not found" errors must result in 404 status codes, too, to prevent e.g. search engines indexing error pages, and be good internet citizens - if I understand correctly, you would check this in the frontend now, at which point you could not set the status code anymore.

askvortsov1:
https://github.com/flarum/core/issues/1847 is a MUCH better way to get the actor, without needing to create a second error handler. Thanks Franz!

askvortsov1:
> * This would render the forum frontend for any 404, even on API and admin, if I am not mistaken.

No API errors will be handled like this, since it's not on the API middleware stack. And neither will admin 404s, as ALL requests to admin routes are sent to the admin frontend, defaulting to the dashboard page if the route isn't found. 

> "Model not found" errors must result in 404 status codes, too, to prevent e.g. search engines indexing error pages, and be good internet citizens - if I understand correctly, you would check this in the frontend now, at which point you could not set the status code anymore.

I don't think so? At least from my testing, it returns the page with a 404 response code.

> Again, first and foremost, I can only recommend to avoid complexity like this on error pages - from painful experience.

We wouldn't be doing it on proper "error" pages, only on 4xxs, where nothing has gone wrong, it's just that the client's input is in some way invalid.

askvortsov1:
Other TODO:

- Once flarum/framework#2703  and flarum/framework#2410 are merged, we should switch to using one ErrorHandler
- I think we should add a `handleInFrontend` method to the `HandledError` class, and provide the list of errors to be handled in the frontend to `Registry`.
- We might want to provide both FrontendFormatter and ViewFormatter to `HandleErrors`? Although that'd be inconsistent with what's needed for API.
- Alternatively, we could merge FrontendFormatter and ViewFormatter into the same class
- Design the error page.

src/Foundation/ErrorServiceProvider.php
```diff
@@ -57,6 +63,15 @@ public function register()
             ];
         });
 
+        $this->container->singleton('flarum.error.contents', function () {
+            return [
+                NotAuthenticatedException::class => NotAuthenticated::class,
```
askvortsov1, line 68: Should these be codes or status strings? That might be cleaner for exceptions added by extensions.

src/Foundation/ErrorHandling/Middleware/ExecuteToFrontend.php
```diff
@@ -0,0 +1,39 @@
...
+
+namespace Flarum\Foundation\ErrorHandling\Middleware;
+
+use Flarum\Http\RouteHandlerFactory;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\MiddlewareInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+
+class ExecuteErrorToFrontend implements MiddlewareInterface
```
askvortsov1: Is this too verbose?

src/Foundation/ErrorHandling/Middleware/ExecuteErrorToFrontend.php
```diff
@@ -0,0 +1,37 @@
...
+
+namespace Flarum\Foundation\ErrorHandling\Middleware;
+
+use Flarum\Http\RouteHandlerFactory;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\MiddlewareInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+
+class ExecuteErrorToFrontend implements MiddlewareInterface
```
askvortsov1, line 18: Is this too verbose? `ExecuteError`?

src/Foundation/ErrorHandling/HandledError.php
```diff
@@ -65,6 +67,11 @@ public function shouldBeReported(): bool
         return $this->type === 'unknown';
     }
 
+    public function contentClass()
+    {
+        return $this->frontendContentClass;
```
askvortsov1, line 72: Not sure I like the naming, especially of the prop

src/Forum/ForumServiceProvider.php
```diff
@@ -124,6 +129,30 @@ public function register()
...
+        $this->container->when(FrontendFormatter::class)
+        ->needs(MiddlewarePipeInterface::class)
+        ->give(function() {
+            $middleware = $this->container->make('flarum.forum.middleware');
+
+            $pipe = new MiddlewarePipe;
+
+            foreach ($middleware as $middlewareClass) {
+                if (!in_array($middlewareClass, [
+                    'flarum.forum.error_handler',
```
askvortsov1, line 141: Anything else we want to exclude?

askvortsov1, line 141: We can't include the error handler to avoid a circular resolution dependency

js/src/common/components/ErrorPage.js
```diff
@@ -0,0 +1,40 @@
...
+      404: '/',
+    };
+
+    const link = links[this.attrs.errorCode];
+
+    return (
+      <div className="ErrorPage">
+        <div className="container">
+          <h2>{this.title}</h2>
+          <p>{app.translator.trans(`core.forum.error.${this.attrs.errorCode}_text`)}</p>
```
askvortsov1, line 28: @frontend folks, any ideas for making this look... not horrible?

davwheat (changes requested):
If I go to a 404 page, click return to forum, then the back button in my browser, I end up with a broken error page:

image

---

We might be able to fix this by using `history.replaceState()` to store the info about the error as part of the browsing history state.

https://developer.mozilla.org/en-US/docs/Web/API/History/replaceState#examples

src/Foundation/ErrorHandling/Middleware/ExecuteErrorToFrontend.php
davwheat: Verbosity === self-documentation in my eyes! )

js/src/common/components/ErrorPage.js
davwheat: My first idea would be bigger text. 13px is fine for forum posts and dense pages, but an error page needs to be padded out in some form, I think.

We don't want to do too much theming as that should be up to forum maintainers, though.

PR #2250, opened 2020-07-29
Fix ui of deleting discussion from user discussions page

### LINKED ISSUE #2233: [Frontend] UI does't get refreshed if a discussion is deleted from the Discussions tab of the user profile
## Bug Report

**Current Behavior**
If a user that has the rights to permanently delete discussions and does this from the Discussions tab of a user profile, the UI doesn't get refreshed and the Discussion stays on screen.

**Steps to Reproduce**
1. Login in as admin
2. Create a new discussion
3. Got to own profile, then on the Discussions tab
4. Use the dropdown menu to delete the discussion created in step 2
5. Use the dropdown menu to "Delete Forever"
6. The discussions stays on screen until the page ins reloaded.

**Expected Behavior**
In step 6, the discussion should not be visible on the screen anymore.

**Environment**
- Flarum version: 0.1.0-beta.13
- PHP version: 7.2.24
- Browser: Chrome 83.0.4103.116

askvortsov1:
**Fixes #2233**

If we are on a discussions user page when we delete a discussion, it should be removed from that page's discussion list.

Seems relatively trivial to be honest. Tested locally.

**Confirmed**

### DIFF
--- a/js/src/forum/utils/DiscussionControls.js
+++ b/js/src/forum/utils/DiscussionControls.js
@@ -6,6 +6,7 @@ import Separator from '../../common/components/Separator';
 import RenameDiscussionModal from '../components/RenameDiscussionModal';
 import ItemList from '../../common/utils/ItemList';
 import extractText from '../../common/utils/extractText';
+import DiscussionsUserPage from '../components/DiscussionsUserPage';
 
 /**
  * The `DiscussionControls` utility constructs a list of buttons for a
@@ -227,7 +228,13 @@ export default {
         app.history.back();
       }
 
-      return this.delete().then(() => app.discussions.removeDiscussion(this));
+      return this.delete().then(() => {
+        app.discussions.removeDiscussion(this);
+
+        if (app.current.matches(DiscussionsUserPage)) {
+          app.current.get('discussions').removeDiscussion(this);
+        }
+      });
     }
   },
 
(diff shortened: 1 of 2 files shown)


js/src/forum/utils/DiscussionControls.js
```diff
@@ -227,7 +228,13 @@ export default {
         app.history.back();
       }
 
-      return this.delete().then(() => app.discussions.removeDiscussion(this));
+      return this.delete().then(() => {
+        app.discussions.removeDiscussion(this);
+
+        if (app.current.matches(DiscussionsUserPage)) {
```
franzliedke, line 234: Code like this (basically an `instanceof` check) always sets off alarm bells for me. `DiscussionControls` is a generic component, and you are coupling it to one of its current concrete usages.

You are also subtly changing behavior here, if I am not mistaken. The old code returned a promise that resolved to the result of `removeDiscussion` - now there's no more return value. That points to the proposed solution, though.

Please try to find a way to implement this with `deleteAction().then(...)` directly inside the `DiscussionsUserPage` component.

askvortsov1: The problem here is that deleteAction is never called from the DiscussionsUserPage; that page isn't even aware of it. In fact, all that page does is instantiate a DiscussionList and a DiscussionListState. The preferable solution would be some way to get the discussion list state instance to the discussion controls, but that also comes with it's own issues.

clarkwinkelmann:
Wouldn't this be better solved with some kind of event system?

Broadcast which discussion has been deleted, and have each list update its state.

I don't like the fact that the user discussion list becomes a "global" variable just for this.

Thinking from an extension standpoint: how would an extension achieve the same if they somehow have another discussion list? I think we should prefer an approach that will work with any kind of discussion list added by any core or extension.

PR #2252, opened 2020-08-03
Fix less build error.

corvofeng:
**Screenshot**

Fix less build error.
image

**Confirmed**

### DIFF
--- a/less/common/scaffolding.less
+++ b/less/common/scaffolding.less
@@ -3,10 +3,10 @@
 :root {
   --flarum-screen: none;
 
-  @media @phone { --flarum-screen: phone }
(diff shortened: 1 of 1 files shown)


datitisev:
We should probably have semicolons there, but why are you trying to compile the Flarum LESS code using webpack or something similar? There's no need for that, the PHP compiles it.

corvofeng:
@datitisev  @tankerkiller125 
I'd like to rewrite this project with `golang`, so I use webpack to build js and css files, and the `flarum/core` is a submodule in my project. After I updated to the latest source, I found this error.

Hope this error can be resolved soon.

PR #2255, opened 2020-08-05
Mithril 2 update

dsevillamartin:
Supersedes flarum/framework#2126.

Some minor TypeScript changes will be here because it makes the development easier. Don't yell at us.

Decisions made thus far:

- mithril 2.0 does not rerun oninit hooks for page components when a change in route results in the same page. There is general agreement that we want to keep / allow the old behavior. The current implementation is a well documented patch in patchMithril.js, but we are happy to consider alternatives. 
- Children is no longer passed to AlertManagerState through attrs; it is now the first argument (attrs is second, componentClass isthird). We might want to switch the order to have attrs before children, but I feel like children is more likely to be provided.
- onunload methods have been merged with context.onunload (from config) into onremove. I am not sure if this is the right way to do this, but it seems to make sense and hasn't broken anything so far
- Instead of returning a retain vnode, subtree retainer now returns a boolean value, which can be used in onbeforeupdate.
- in many cases, we pass a "config" attr to some html element instead of creating a custom component for it. Since there are now multiple life cycle hooks, not just config, passing multiple of those would be inconvenient. If multiple hooks, or logic more than a few lines, needs to be passed as attrs to a component, I am creating a new component for it (e.g. CommentPostPreview, ReplyPlaceholderPreview)
- despite the consensus in the previous team meeting to not store this.attrs, on implementation, that has turned out more difficult than expected, as we'd need to pass attrs through long chains of sometimes more than 5 methods. We have reverted to using this.attrs, and are also keeping initAttrs (although thst is NOT static anymore)
- we are no longer using attrs.children to pass in children. We've even added a warning error message to deter against this.
- In some cases, attrs would be passed to an html vnode, which are distinct from the attrs passed to the component. Previously, these were managed via an attrs method. However, since we now use this.attrs, that will not work. We are renaming those methods to elementAttrs.
- mapRoutes no longer sets a routeName attr on components
- All frontend deprecated code has been removed, except for various warnings/errors designed to prevent wrong inputs.

References to props in comments have not been changed; they will be removed when we introduce attr typing in the typescript part of the rewrite.

Some issues that need fixes
- https://github.com/flarum/core/pull/2255/commits/e0fce1c02d58f115151e80d81d6d045912caa31e is messy... Do we want to abstractify these components, and bring them out of this file?

Extracted fixes:
- https://github.com/flarum/core/pull/2262
- https://github.com/flarum/core/pull/2263
- https://github.com/flarum/core/pull/2264


askvortsov1:
Decisions made thus far:

- mithril 2.0 does not rerun oninit hooks for page components when a change in route results in the same page. There is general agreement that we want to keep / allow the old behavior. The current implementation is a well documented patch in patchMithril.js, but we are happy to consider alternatives. 
- onunload methods have been merged with context.onunload (from config) into onremove. I am not sure if this is the right way to do this, but it seems to make sense and hasn't broken anything so far
- in many cases, we pass a "config" attr to some html element instead of creating a custom component for it. Since there are now multiple life cycle hooks, not just config, passing multiple of those would be inconvenient. If multiple hooks, or logic more than a few lines, needs to be passed as attrs to a component, I am creating a new component for it (e.g. CommentPostPreview, ReplyPlaceholderPreview)
- despite the consensus in the previous team meeting to not store this.attrs, on implementation, that has turned out more difficult than expected, as we'd need to pass attrs through long chains of sometimes more than 5 methods. We have reverted to using this.attrs, and are also keeping initAttrs (although thst is NOT static anymore)
- we are no longer using attrs.children to pass in children. We've even added a warning error message to deter against this.
- In some cases, attrs would be passed to an html vnode, which are distinct from the attrs passed to the component. Previously, these were managed via an attrs method. However, since we now use this.attrs, that will not work. We are renaming those methods to elementAttrs.

js/src/forum/components/SignUpModal.js
```diff
@@ -73,8 +73,8 @@ export default class SignUpModal extends Modal {
           type="text"
           placeholder={extractText(app.translator.trans('core.forum.sign_up.username_placeholder'))}
           value={this.username()}
-          onchange={m.withAttr('value', this.username)}
-          disabled={this.loading || this.isProvided('username')}
+          bidi={this.username}
```
franzliedke, line 76: Again, `value` can be removed, I think. Same for several other occurrences in this file.

js/src/forum/components/SignUpModal.js
```diff
@@ -73,8 +73,8 @@ export default class SignUpModal extends Modal {
...
+          disabled={this.loading || this.isProvided('username', this.attrs)}
```
franzliedke, line 77: That second parameter (`this.attrs`) is not necessary and has not been added to the method (same a few lines below).

js/src/common/utils/withAttr.ts
```diff
@@ -0,0 +1,4 @@
+export default (key: string, cb: Function) =>
```
franzliedke, line 1: Document this, please.

js/src/common/utils/extractText.js
```diff
@@ -8,7 +8,7 @@ export default function extractText(vdom) {
   if (vdom instanceof Array) {
     return vdom.map((element) => extractText(element)).join('');
   } else if (typeof vdom === 'object' && vdom !== null) {
-    return extractText(vdom.children);
+    return vdom.children ? extractText(vdom.children) : vdom.text;
```
franzliedke, line 11: Just for my understanding: Does Mithril VDOM now have a special case for text-only nodes?

js/src/common/utils/SubtreeRetainer.js
```diff
@@ -32,7 +32,7 @@ export default class SubtreeRetainer {
    * @return {Object|false}
    * @public
    */
-  retain() {
+  needsRebuild() {
```
franzliedke, line 35: Please rename / change the docblock example accordingly.

franzliedke (changes requested):
Found a bunch of stuff already. Haven't looked at all files yet, though.

js/src/forum/components/ForgotPasswordModal.js
```diff
@@ -65,17 +64,19 @@ export default class ForgotPasswordModal extends Modal {
               type="email"
               placeholder={extractText(app.translator.trans('core.forum.forgot_password.email_placeholder'))}
               value={this.email()}
-              onchange={m.withAttr('value', this.email)}
+              bidi={this.email}
```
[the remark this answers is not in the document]
datitisev: Yes. We removed value from some places, but it seems we didn't get them all.

js/src/forum/components/Notification.js
```diff
@@ -8,26 +8,25 @@ import Button from '../../common/components/Button';
...
       <a
         className={'Notification Notification--' + notification.contentType() + ' ' + (!notification.isRead() ? 'unread' : '')}
-        href={href}
-        config={function (element, isInitialized) {
-          if (href.indexOf('://') === -1) m.route.apply(this, arguments);
-
-          if (!isInitialized) $(element).click(this.markAsRead.bind(this));
-        }}
+        {...linkAttrs}
+        onclick={this.markAsRead.bind(this)}
```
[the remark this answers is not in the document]
datitisev: Not sure. I think the performance impact is minimal, so that shouldn't matter? We could move it to an oncreate, but I don't see why that was needed in the first place either.

js/src/forum/components/NotificationList.js
```diff
@@ -66,7 +63,7 @@ export default class NotificationList extends Component {
                   return (
                     <div className="NotificationGroup">
                       {group.discussion ? (
-                        <a className="NotificationGroup-header" href={app.route.discussion(group.discussion)} config={m.route}>
+                        <a className="NotificationGroup-header" route={app.route.discussion(group.discussion)}>
```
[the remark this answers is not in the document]
datitisev: Yes. The `route` parameter is replaced with `href` in `patchMithril`, and the tag `a` is replaced with `m.route.Link` (which internally uses `a` by default).

js/src/forum/components/Post.js
```diff
@@ -27,61 +29,62 @@ export default class Post extends Component {
...
   }
 
-  config(isInitialized) {
+  onbeforeupdate(vnode) {
+    super.onbeforeupdate(vnode);
+
+    return this.subtree.needsRebuild();
+  }
+
+  onupdate(vnode) {
```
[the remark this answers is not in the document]
datitisev: We have only been calling the base on `oninit`, `oncreate` and `onbeforeupdate`, as we use those for updating the stored attributes/props. There's is no parent `onupdate` in this case, I think. So yeah, the vnode parameter can be removed.

js/src/forum/utils/UserControls.js
```diff
@@ -22,10 +22,10 @@ export default {
     const items = new ItemList();
 
     ['user', 'moderation', 'destructive'].forEach((section) => {
-      const controls = this[section + 'Controls'](user, context).toArray();
+      const controls = this[section](user, context).toArray();
```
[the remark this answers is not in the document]
datitisev: Ah, I see what you meant with non-mithril related changes now. This was a suggestion from https://github.com/flarum/core/issues/872#issuecomment-198682653, but I guess it should go in a different PR.

js/src/common/components/ConfirmDocumentUnload.js
```diff
@@ -17,21 +17,23 @@ import Component from '../Component';
...
 
-    $(window).on('beforeunload', handler);
+    $(window).on('beforeunload', this.handler.bind(this));
+  }
 
-    context.onunload = () => {
-      $(window).off('beforeunload', handler);
-    };
+  onremove(vnode) {
+    $(window).off('beforeunload', this.handler.bind(this));
```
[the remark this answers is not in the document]
datitisev: Could you elaborate on what you mean here?

js/src/admin/AdminApplication.js
```diff
@@ -27,13 +27,16 @@ export default class AdminApplication extends Application {
...
-    m.mount(document.getElementById('admin-navigation'), AdminNav.component());
+    m.mount(document.getElementById('app-navigation'), { view: () => Navigation.component({ className: 'App-backControl', drawer: true }) });
+    m.mount(document.getElementById('header-navigation'), Navigation);
+    m.mount(document.getElementById('header-primary'), HeaderPrimary);
+    m.mount(document.getElementById('header-secondary'), HeaderSecondary);
+    m.mount(document.getElementById('admin-navigation'), AdminNav);
+
+    if (!document.location.hash) document.location.hash = '#/';
+
+    m.route.prefix = '#';
```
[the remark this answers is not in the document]
datitisev: That's if we want the new default behavior. Using `#!` would change the routing in admin from `/admin#/appearance` to `/admin#!/appearance`.

js/src/admin/AdminApplication.js
```diff
@@ -27,13 +27,16 @@ export default class AdminApplication extends Application {
...
-    m.mount(document.getElementById('header-primary'), HeaderPrimary.component());
-    m.mount(document.getElementById('header-secondary'), HeaderSecondary.component());
-    m.mount(document.getElementById('admin-navigation'), AdminNav.component());
+    m.mount(document.getElementById('app-navigation'), { view: () => Navigation.component({ className: 'App-backControl', drawer: true }) });
+    m.mount(document.getElementById('header-navigation'), Navigation);
+    m.mount(document.getElementById('header-primary'), HeaderPrimary);
+    m.mount(document.getElementById('header-secondary'), HeaderSecondary);
+    m.mount(document.getElementById('admin-navigation'), AdminNav);
+
+    if (!document.location.hash) document.location.hash = '#/';
```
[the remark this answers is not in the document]
datitisev: I think I encountered this issue too. With mithril 0.2.x, mithril redirects to the route with the hash automatically. With 2.x, it does not do that, so it doesn't display the home route in the admin.

js/src/common/utils/extractText.js
datitisev: Yes. If it's text only, it'll be found in `vnode.text` as opposed to being an array of a string in `vnode.children` (I think that's what it previously was, anyway)

js/src/common/components/Dropdown.js
```diff
@@ -104,15 +104,15 @@ export default class Dropdown extends Component {
    * @return {*}
    * @protected
    */
-  getButtonContent() {
+  getButtonContent(children) {
```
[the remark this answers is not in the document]
askvortsov1: It's not used here, but it is used in subclasses. SplitDropdown, for instance.

js/src/common/components/Modal.js
```diff
@@ -14,26 +14,20 @@ export default class Modal extends Component {
...
+   */
+  alertAttrs = null;
 
-    this.props.onshow(() => this.onready());
+  oncreate(vnode) {
+    super.oncreate(vnode);
 
-    context.onunload = () => {
-      this.props.onhide();
-    };
```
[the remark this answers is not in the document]
askvortsov1: Actually, this one is deliberate: it fixes an issue introduced in the state PR where you can't open a modal if a modal is already open.

js/src/common/components/ModalManager.js
```diff
@@ -20,22 +16,23 @@ export default class ModalManager extends Component {
     );
   }
 
-  config(isInitialized, context) {
-    if (isInitialized) return;
+  onupdate(vnode) {
+    if (this.$('Modal') && !this.attrs.state.modal) {
+      this.animateHide();
+    }
+  }
```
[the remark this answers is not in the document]
askvortsov1: This comes to fix the issue mentioned above. Should be split into a separate PR though...

js/src/common/components/Placeholder.js
```diff
@@ -9,10 +9,10 @@ import Component from '../Component';
  * - `text`
  */
 export default class Placeholder extends Component {
-  view() {
+  view(vnode) {
```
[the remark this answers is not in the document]
askvortsov1: not neccessary, I'd be fine removing it.

js/src/admin/AdminApplication.js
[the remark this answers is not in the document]
askvortsov1: Yep, this was discovered in the first mithril rewrite attempt.

js/src/common/utils/extractText.js
askvortsov1: Which is very annoying... But it is what it is, I suppose. I think we should just document when we need to hack around it (like in UsersSearchSource.

js/src/common/components/ConfirmDocumentUnload.js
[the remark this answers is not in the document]
franzliedke: This is not removing the handler attached in `oncreate`. `Function.prototype.bind` creates a new bound function on every call.

js/src/common/components/Dropdown.js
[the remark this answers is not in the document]
franzliedke: Saw that later and forgot to edit the comment. It doesn't go down to `getButtonContent()`, though, IIRC. I think `SplitDropdown` only needs the children in `getButtonContent()`.

js/src/common/components/Modal.js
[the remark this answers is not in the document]
franzliedke: Okay, extract it, please, with a detailed explanation.

js/src/common/components/ConfirmDocumentUnload.js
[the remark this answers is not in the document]
datitisev: Ah, that's true. We might want to do `this.handler = this.handler.bind(this)` to resolve this.

js/src/forum/components/Notification.js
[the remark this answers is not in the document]
askvortsov1: Or a readability one. Either way, this is cleaner.

js/src/forum/components/PostEdited.js
```diff
@@ -32,7 +34,7 @@ export default class PostEdited extends Component {
     );
   }
 
-  config(isInitialized) {
+  onupdate(vnode) {
```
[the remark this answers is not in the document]
askvortsov1: In this case no, the tooltip wouldn't be there to destroy on create. I'll check other places though.

js/src/forum/components/UsersSearchSource.js
```diff
@@ -43,11 +43,17 @@ export default class UsersSearchResults {
       <li className="Dropdown-header">{app.translator.trans('core.forum.search.users_heading')}</li>,
       results.map((user) => {
         const name = username(user);
+
+        if (!name.children) {
+          name.children = [name.text];
+          delete name.text;
+        }
```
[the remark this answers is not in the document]
askvortsov1: This is now removed, best to forget it now...

js/src/forum/components/Composer.js
```diff
@@ -64,12 +66,10 @@ export default class Composer extends Component {
...
       this.prevPosition = this.state.position;
     }
+  }
 
-    if (isInitialized) return;
-
-    // Since this component is a part of the global UI that persists between
-    // routes, we will flag the DOM to be retained across route changes.
-    context.retain = true;
+  oncreate(vnode) {
```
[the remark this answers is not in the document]
askvortsov1: Why? It currently animates in just fine... Besides, since this.prevPosition is set to this.state.position oninit, it wouldn't be applicable?

js/src/common/components/Dropdown.js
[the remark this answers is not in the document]
askvortsov1: SelectDropdown is the one that uses it actually.

js/src/common/components/Modal.js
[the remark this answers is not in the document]
askvortsov1: https://github.com/flarum/core/pull/2263

js/src/forum/components/NotificationList.js
```diff
@@ -97,27 +94,34 @@ export default class NotificationList extends Component {
...
-      const scrollTop = $scrollParent.scrollTop();
-      const viewportHeight = $scrollParent.height();
-      const contentTop = $scrollParent === $notifications ? 0 : $notifications.offset().top;
-      const contentHeight = $notifications[0].scrollHeight;
+    $scrollParent.off('scroll', this.scrollHandler.bind(this));
+  }
+
+  scrollHandler() {
+    const state = this.attrs.state;
+    const scrollTop = $scrollParent.scrollTop();
```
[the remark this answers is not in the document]
askvortsov1: Fixed by moving it up a few lines for now, see your other comment for more discussion on this.

js/src/forum/components/NotificationList.js
```diff
@@ -97,27 +94,34 @@ export default class NotificationList extends Component {
     );
   }
 
-  config(isInitialized, context) {
-    if (isInitialized) return;
+  oncreate(vnode) {
+    super.oncreate(vnode);
+
+    const $notifications = this.$('.NotificationList-content');
+    const $scrollParent = $notifications.css('overflow') === 'auto' ? $notifications : $(window);
```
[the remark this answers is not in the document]
askvortsov1: I'm not sure I like explicitly tying our business logic here to the physical DOM. I'm also not sure if this is even allowed, seeing as the actual DOM can, at times, be regenerated to keep up with Mithril. This isn't a problem with SuperTextArea in Composer since that's reset on every single call to view(), but it might  be here. All for avoiding unnecesary repetition, but I don't think that's the way.

js/src/forum/components/UserPage.js
```diff
@@ -1,12 +1,12 @@
 import Page from '../../common/components/Page';
 import ItemList from '../../common/utils/ItemList';
-import affixSidebar from '../utils/affixSidebar';
 import UserCard from './UserCard';
 import LoadingIndicator from '../../common/components/LoadingIndicator';
 import SelectDropdown from '../../common/components/SelectDropdown';
 import LinkButton from '../../common/components/LinkButton';
 import Separator from '../../common/components/Separator';
 import listItems from '../../common/helpers/listItems';
+import UserPageSidebar from './AffixedSidebar';
```
clarkwinkelmann, line 9: Why rename the component in the import?

js/src/forum/components/AffixedSidebar.js
```diff
@@ -1,16 +1,12 @@
...
- * @param {Object} context
- */
-export default function affixSidebar(element, isInitialized, context) {
-  if (isInitialized) return;
-
-  const onresize = () => {
-    const $sidebar = $(element);
+import Component from '../../common/Component';
+
+export default class AffixedSidebar extends Component {
```
clarkwinkelmann, line 3: I seem to remember that the affix behavior was no longer working in the current Flarum version. So this change also fixes it?

js/src/forum/components/TextEditor.js
```diff
@@ -47,26 +53,28 @@ export default class TextEditor extends Component {
...
+    $(vnode.dom).bind('keydown', 'ctrl+return', handler);
 
-    this.props.composer.editor = new SuperTextarea(element);
+    this.attrs.composer.editor = new SuperTextarea(vnode.dom);
   }
 
+  /**
+   * Here for extensibility
+   */
+  textareaOnupdate(vnode) {}
```
clarkwinkelmann, line 71: Is this actually needed ? The `textarea` is a direct child of `TextEditor`. Extensions could simply extend `TextEditor.onupdate` and access `vnode.dom.querySelector('textarea')` from there. Since there are no conditionals  in `view` it's guaranteed to exist when the component's lifecycle events run.

We could also do the same above and it would save us one `bind`.

clarkwinkelmann (changes requested):
I'd like to suggest we create a util method or Mithril patch for `m.route.set('/', null, { state: { key: Date.now() } });`. It's not obvious what it does and it would be much simpler to have it inside of a named method instead of adding a comment everywhere. Something like `setRouteWithForcedRefresh()`

I have read through the changes in this PR last week and again now and I think it's still not an exhaustive review but I think I agree with the vast majority of changes.

js/src/forum/components/AffixedSidebar.js
askvortsov1: Maybe I don't understand what affixed sidebar does then? Seems to work for me on discuss.flarum.org on the user page

js/src/forum/components/UserPage.js
askvortsov1: Typo on my part

js/src/forum/components/TextEditor.js
askvortsov1: Good point, done!

js/src/forum/components/AffixedSidebar.js
clarkwinkelmann: Nevermind, I misremembered. The code I was thinking of is on the TagsPage and wasn't doing anything since the homepage didn't use affix.

https://github.com/flarum/tags/blob/1c6dfdd0ff6ad2bfb8c68baa2ce160f7e926c74f/js/src/forum/components/TagsPage.js#L26

I see it has been removed in the Mithril2 PR for Tags, so all is good. Neither index nor tags page use the affix behavior, that's good.

askvortsov1:
> I'd like to suggest we create a util method or Mithril patch for `m.route.set('/', null, { state: { key: Date.now() } });`. It's not obvious what it does and it would be much simpler to have it inside of a named method instead of adding a comment everywhere. Something like `setRouteWithForcedRefresh()`

That's a great idea! Also allows us to document this centrally. Done!

clarkwinkelmann (approved):
All good regarding my previous comments. Not sure if I should do another read-through before the end of the week or if it's enough to wait on others to complete their own review?

askvortsov1 (approved):
Ooo 2 approving reviews, does that mean we can merge :rofl: 
(just to clarify, no merge until everyone agrees or Franz overrides. This approval is largely symbolic in that I think thisr implementation is sufficient to not diminish extensibility, and does not introduce bugs (that I have been able to find))

dsevillamartin (commented):
We still have some `lazyRedraw`s left behind - see https://github.com/flarum/core/blob/4c4c148c6e3fe6c951941762a685ab42f2329453/js/src/common/Model.js#L183

js/src/forum/components/NotificationList.js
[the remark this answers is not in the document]
franzliedke: It's the same concept you're applying for the `boundScrollHandler` - that one is an event handler, this one is an element.

I don't think this is business logic at all, it's DOM state management. What you attach manually in `oncreate`, you also need to detach in `onremove`.

> seeing as the actual DOM can, at times, be regenerated to keep up with Mithril.

For such a simple DOM structure, Mithril will re-use the DOM nodes, it only re-creates the VDOM repeatedly.

Another small problem with recalculating this in `onremove` is that you could theoretically end up with a different `$scrollParent`, which doesn't have the scroll handler. Since this is about removing a particular handler instance from a particular DOM node, IMO it can be stored. Anyway, too many words. :speak_no_evil:

js/src/forum/components/Composer.js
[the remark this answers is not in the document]
franzliedke: I see how the `if` branch won't be executed in `oncreate` (as `this.prevPosition` will probably be undefined), but what about the `else` branch? Without this, `this.prevPosition` won't be set on creation.

js/src/admin/AdminApplication.js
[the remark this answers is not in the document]
franzliedke: Add a comment, please? :pray:

js/src/common/utils/extractText.js
franzliedke: Is the `else` case still relevant then? I would assume that can be removed then?

js/src/common/utils/SubtreeRetainer.js
franzliedke: Also, we could implement this with `this.callbacks.some(...)` now. :grin: 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

js/src/forum/components/Composer.js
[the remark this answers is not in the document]
askvortsov1: Well, `this.prevPosition` is set on `oninit`, and `animatePositionChange` isn't relevant, because the state's position is always hidden on creation (since this is the "composer manager", and not the composer body.

js/src/forum/components/NotificationList.js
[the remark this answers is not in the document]
askvortsov1: For some reason, I thought you had been talking about the `state` state. Like `NotificationListState`. Sorry, that's a misunderstanding on my part. Fixed.

js/src/common/utils/extractText.js
askvortsov1: No, because sometimes it'll still put text in a text vnode, which still uses children. For instance, if you run `m('p', [m('a', 'Hello!'), 'world']);` in browser console.

tankerkiller125 (approved):

js/src/forum/components/DiscussionListItem.js
```diff
@@ -98,22 +97,21 @@ export default class DiscussionListItem extends Component {
...
+            route={user ? app.route.user(user) : '#'}
             className="DiscussionListItem-author"
             title={extractText(
               app.translator.trans('core.forum.discussion_list.started_text', { user: user, ago: humanTime(discussion.createdAt()) })
             )}
-            config={function (element) {
-              $(element).tooltip({ placement: 'right' });
-              m.route.apply(this, arguments);
+            oncreate={function (vnode) {
+              $(vnode.dom).tooltip({ placement: 'right' });
```
franzliedke, line 106: This was previously called on `onupdate` as well. Did that simply have no effect?

js/src/common/components/LinkButton.js
```diff
@@ -13,15 +13,17 @@ import Button from './Button';
...
   }
 
-  view() {
-    const vdom = super.view();
+  view(vnode) {
+    const vdom = super.view(vnode);
 
-    vdom.tag = 'a';
+    vdom.tag = m.route.Link;
+    vdom.attrs.active = String(vdom.attrs.active);
```
franzliedke, line 26: I don't understand what effect this is supposed to have. I must be missing something completely, but there's nothing in the parent class (and neither was in old code) that does something to this attribute, as far as I can tell.

js/src/forum/components/CommentPost.js
```diff
@@ -41,48 +40,44 @@ export default class CommentPost extends Post {
...
     // necessary because TextFormatter outputs them for e.g. syntax highlighting.
-    if (context.contentHtml !== contentHtml) {
+    if (vnode.dom.contentHtml !== contentHtml) {
       this.$('.Post-body script').each(function () {
         eval.call(window, $(this).text());
       });
     }
 
-    context.contentHtml = contentHtml;
+    vnode.dom.contentHtml = contentHtml;
```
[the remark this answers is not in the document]
datitisev: Hm yeah not sure why that's being stored there now. Maybe because of `context`? Though to be honest I never understood what `context` is. It might have been the VDOM.

js/src/common/components/LinkButton.js
askvortsov1: Tbh, I don't recall making this change. @datitisev do you remember why we might have added this?

js/src/forum/components/CommentPost.js
```diff
@@ -41,48 +40,44 @@ export default class CommentPost extends Post {
 
     this.subtree.check(
       () => this.cardVisible,
-      () => this.isEditing()
+      () => this.isEditing(),
+      () => this.revealContent
```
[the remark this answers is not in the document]
askvortsov1: Pretty much, yeah

js/src/forum/components/CommentPost.js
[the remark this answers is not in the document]
askvortsov1: Yeah, this was moved over because I didn't understand what context did (and why).

js/src/forum/components/PostEdited.js
```diff
@@ -32,7 +34,7 @@ export default class PostEdited extends Component {
     );
   }
 
-  config(isInitialized) {
+  onupdate() {
```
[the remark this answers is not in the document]
askvortsov1: Good point, fixed!

js/src/forum/components/DiscussionListItem.js
askvortsov1: Well, once we've created the tooltip, it remains on the DOM element until the DOM element is removed, right? (in which case oncreate would run again if its readdded). There's no need to re-update that if we're not changing the text (which we're not).

franzliedke: Thanks. I wasn't familiar with how the tooltip JS works. :+1:

js/src/forum/components/Composer.js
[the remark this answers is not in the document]
franzliedke: Ah, I missed your point about `this.prevPosition` being initialized. More importantly, though, `oncreate` does not need any of the height animation stuff, because the composer is always hidden on first render.

js/src/common/components/SelectDropdown.js
```diff
@@ -21,7 +39,7 @@ export default class SelectDropdown extends Dropdown {
   }
 
   getButtonContent(children) {
-    const activeChild = children.filter((child) => child.attrs.active)[0];
+    const activeChild = children.filter(isActive)[0];
```
datitisev, line 42: We can now use `Array.prototype.find()` because we're screwing IE. 🎉

datitisev: We can now use `Array.prototype.find()` because we're screwing IE. 🎉

datitisev: Lol! https://github.com/flarum/core/pull/2255/commits/ab9d72d860787f9b89fbab80d5e90ba05566c459

js/src/forum/components/PostStreamScrubber.js
```diff
@@ -29,26 +31,26 @@ export default class PostStreamScrubber extends Component {
...
-      if (context.oldStyle) {
-        $element.stop(true).css(context.oldStyle).animate(newStyle);
+      if (vnode.oldStyle) {
+        $element.stop(true).css(vnode.oldStyle).animate(newStyle);
       } else {
         $element.css(newStyle);
       }
 
-      context.oldStyle = newStyle;
+      vnode.oldStyle = newStyle;
```
franzliedke, line 50: Shouldn't this be stored on `vnode.state` and not `vnode` itself? According to https://mithril.js.org/vnodes.html#structure, `state` is persisted. The `vnode` is recreated on every render, I would assume.

js/src/forum/components/PostStream.js
```diff
@@ -73,7 +75,7 @@ export default class PostStream extends Component {
       }
 
       return (
-        <div className="PostStream-item" {...attrs}>
+        <div className="PostStream-item fade-in" {...attrs}>
```
franzliedke, line 78: This seems to be a left-over, right? We extracted the CSS changes to master in #2262, but even that doesn't even have this CSS class... :shrug:

js/src/forum/components/PostStream.js
```diff
@@ -368,6 +374,10 @@ export default class PostStream extends Component {
    * @param {jQuery} $item
    */
   flashItem($item) {
-    $item.addClass('flash').one('animationend webkitAnimationEnd', () => $item.removeClass('flash'));
+    $item.addClass('flash').on('animationend webkitAnimationEnd', () => {
+      if (event.animationName === 'fadeIn') {
+        $item.removeClass('flash');
+      }
+    });
```
franzliedke, line 381: Same here. This change was added in commit 1e95bb2a, but not extracted to master in #2262 and not reverted in commit 4872ddc5.

Is this needed or not? If so, what does it do? Should this be in master?

js/src/forum/components/PostStreamScrubber.js
```diff
@@ -133,6 +136,15 @@ export default class PostStreamScrubber extends Component {
       .on('mouseup touchend', (this.handlers.onmouseup = this.onmouseup.bind(this)));
 
     setTimeout(() => this.scrollListener.start());
+
+    this.updateScrubberValues({ animate: true, forceHeightChange: true });
```
franzliedke, line 140: This does not wait for the `loadPromise`, as does `onupdate`. I'd expect that to be even more relevant on first render, but may be wrong...

franzliedke (commented):
A few post stream questions for @askvortsov1

js/src/forum/components/PostStream.js
askvortsov1: SInce we now use CSS for everything, it ensures that we don't remove irrelevant animations (as animationend will run for any animation). And yes, it should be in master.

js/src/forum/components/PostStreamScrubber.js
askvortsov1: Interesting, iirc, this was more of a "it works this way" implementation. That begs the question: Should other calls wait for the promise? If so, should that be a part of the updateScrubberValues method itself?

franzliedke: Let's not try figuring that out here. :wink:  Since it's a time-based thing, it may also be a problem only under certain conditions, or never. I just noticed it's technically a change in behavior.

js/src/forum/components/PostStream.js
franzliedke: I removed it here. Please feel to send a bugfix PR if this is still a problem in master.

js/src/common/helpers/listItems.js
```diff
@@ -30,21 +30,29 @@ export default function listItems(items) {
...
-    const active = item.component && item.component.isActive && item.component.isActive(item.props);
-    const className = item.props ? item.props.itemClassName : item.itemClassName;
+    const isListItem = item.tag && item.tag.isListItem;
+    const active = item.tag && item.tag.isActive && item.tag.isActive(item.attrs);
+    const className = (item.attrs && item.attrs.itemClassName) || item.itemClassName;
 
     if (isListItem) {
       item.attrs = item.attrs || {};
       item.attrs.key = item.attrs.key || item.itemName;
+      item.key = item.attrs.key;
```
franzliedke, line 40: Do we need `item.attrs.key` or `item.key`? Probably not both, right?

js/src/common/helpers/listItems.js
```diff
@@ -30,21 +30,29 @@ export default function listItems(items) {
...
-      <li className={classList([item.itemName ? 'item-' + item.itemName : '', className, active ? 'active' : ''])} key={item.itemName}>
+      <li
+        className={classList([className, item.itemName && `item-${item.itemName}`, active && 'active'])}
+        key={(item.attrs && item.attrs.key) || item.itemName}
+      >
         {item}
       </li>
     );
+
+    node.state = node.state || {};
```
franzliedke, line 54: Why is this needed? If there's no comment, it's hard to know (or remember) why.

js/src/common/components/LinkButton.js
franzliedke: Will hopefully be cleaned up as part of #2265.

js/src/common/helpers/listItems.js
askvortsov1: @datitisev

js/src/forum/components/PostStreamScrubber.js
askvortsov1: Moved to https://github.com/flarum/core/issues/2297

js/src/common/helpers/listItems.js
askvortsov1: Iirs, we're modifying it after it's been created. So if we want to change the key, we should change both the actual key and the attrs (for consistency)

js/src/common/Application.js
```diff
@@ -434,9 +432,19 @@ export default class Application {
...
+    const route = this.routes[name];
+
+    if (!route) throw new Error(`Route '${name}' does not exist`);
+
+    const url = route.path.replace(/:([^\/]+)/g, (m, key) => extract(params, key));
+
+    // Remove falsy values in params to avoid having urls like '/?sort&q'
+    for (const key in params) {
+      if (params.hasOwnProperty(key) && !params[key]) delete params[key];
+    }
```
franzliedke, line 444: Why wasn't this previously necessary, by the way?

js/src/common/helpers/listItems.js
franzliedke: Would prefer to get rid of it, or have a comment explaining what's the point of this.

js/src/common/Application.js
datitisev: I think because Mithril's generate query string utility probably got updated to include empty parameters.

js/src/common/helpers/listItems.js
datitisev: 🤔 I don't know... test if things work before removing it though.

askvortsov1: Tested without it, didn't seem to break anything. Goodbye!

js/src/common/helpers/listItems.js
askvortsov1: To elaborate, from my understanding of mithril, if `vnode.key` is set, `vnode.attrs.key` should also be set.

franzliedke (approved):

franzliedke:
Congrats! :tada:

PR #2261, opened 2020-08-14
Fix number abbreviation when the number is n-thousand

matteocontrini:

This commit fixes the method `abbreviateNumber` so that it behaves as stated in the JSDoc.

Previously, an input of `1234` would have produced `1K`. With this change, the output will be `1.2K`.

/

**Screenshot**

/

**Confirmed**


askvortsov1:
We'll merge this after the rewrite is merged.

askvortsov1:
Actually, as there's no conflict, might as well merge this now.

PR #2262, opened 2020-08-14
Extract a few changes from the Mithril 2 upgrade

franzliedke:
Just a few small changes / bug fixes that were discovered when working on the update to Mithril 2.0 (#2255), apparently, but are unrelated.


askvortsov1:
Can we move the Modal changes (removing onunload, adding functionally equivalent code in ModalManager.onupdate) here too?

PR #2263, opened 2020-08-16
Fix opening modals from other modals.

askvortsov1:
**DO NOT MERGE WHEN APPROVED, REROUTE TO MASTER AFTER THE REWRITE IS MERGED**

While seemingly correct, an onremove method in Modal that triggers animateHide is problematic, because if one modal is opened from another, the one currently open will be removed from the DOM, triggering animateHide, and closing the new modal.

To compensate, an onupdate method now closes a modal if one is open but shouldn't be; this supports the functionality of the old method when the modal is closed not from the modal instance itself (e.g. app.modal.close())

This is not ideal, but necessary. We should consider eventually expanding the modal system to support showing multiple modals at the same time (stacked over each other). Then, we can move this back to individual modals.

### DIFF
--- a/js/src/common/components/ModalManager.js
+++ b/js/src/common/components/ModalManager.js
@@ -17,7 +17,7 @@ export default class ModalManager extends Component {
   }
 
   onupdate() {
-    if (this.$('Modal') && !this.attrs.state.modal) {
+    if (this.$('.Modal') && !this.attrs.state.modal) {
       this.animateHide();
     }
   }
--- a/js/src/common/components/ModalManager.js
+++ b/js/src/common/components/ModalManager.js
@@ -16,6 +16,12 @@ export default class ModalManager extends Component {
     );
   }
 
+  onupdate() {
+    if (this.$('Modal') && !this.attrs.state.modal) {
+      this.animateHide();
+    }
+  }
+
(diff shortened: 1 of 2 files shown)


js/src/common/components/ModalManager.js
```diff
@@ -16,6 +16,12 @@ export default class ModalManager extends Component {
     );
   }
 
+  onupdate() {
+    if (this.$('Modal') && !this.attrs.state.modal) {
```
clarkwinkelmann, line 20: Are you sure about this? Shouldn't it be `.Modal` ?

askvortsov1: Should be... Probably I made a typo when extracting it from the main PR, thanks for the catch!

clarkwinkelmann: So... was it an error or not? I see you did not change it

askvortsov1: Now I'm confused... Code editor says I pushed that commit...

askvortsov1: Turns out I pushed to the wrong remote... Good job me! :see_no_evil:

clarkwinkelmann (approved):
All good then!

PR #2264, opened 2020-08-16
Revert to using state.needsScroll for triggering scrolls with PostStream

askvortsov1:
While more pleasant from an FSM standpoint, comparing the current targetPost to the previous one does not work if goToNumber is called twice in a row for the same post. For instance, if a user clicks the mentions link to a post twice, post stream breaks.

### DIFF
--- a/js/src/forum/components/PostStream.js
+++ b/js/src/forum/components/PostStream.js
@@ -121,15 +121,10 @@ export default class PostStream extends Component {
    * Start scrolling, if appropriate, to a newly-targeted post.
    */
   triggerScroll() {
-    if (!this.attrs.targetPost) return;
+    if (!this.attrs.targetPost || !this.stream.needsScroll) return;
 
-    const oldTarget = this.prevTarget;
     const newTarget = this.attrs.targetPost;
-
-    if (oldTarget) {
-      if ('number' in oldTarget && oldTarget.number === newTarget.number) return;
-      if ('index' in oldTarget && oldTarget.index === newTarget.index) return;
-    }
+    this.stream.needsScroll = false;
 
     if ('number' in newTarget) {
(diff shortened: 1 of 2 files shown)


clarkwinkelmann:
I'm not familiar with this issue. You say it's restoring a previous behavior? Where/when was it changed?

askvortsov1:
Sorry, "revert" is poor wording. This system was used towards the beginning of the Post Stream State PR, but replaced with the current one as that PR was developed. So this was never in master, but it was in that PR.

datitisev:
I don't entirely know what this fixes and why/how things were changed so I won't approve, but this seems to be fine to me.

askvortsov1:
Under the current system, we store the last post we scrolled to. If, on update, the current scroll "target" is different, we execute a scroll.

This is problematic because if we try to scroll to the same post twice in a row (or to a post we opened the page on), it won't work the second time around.

So, I propose that we directly set a flag that indicates that a scroll is needed, and use that instead. Which fixes the issue.

js/src/forum/components/PostStream.js
```diff
@@ -121,24 +121,17 @@ export default class PostStream extends Component {
    * Start scrolling, if appropriate, to a newly-targeted post.
    */
   triggerScroll() {
-    if (!this.attrs.targetPost) return;
+    if (!this.attrs.targetPost || !this.stream.needsScroll) return;
```
clarkwinkelmann, line 124: You might even be able to drop `!this.attrs.targetPost` if you know it's always set together with `needsScroll

clarkwinkelmann (approved):
Ok that new description of that this does makes more sense :+1: 

Haven't tested, but looks good.

datitisev (approved):

Issue #2266, opened 2020-08-17
[Frontend Rewrite] Bring back current routeName


w-4:
Often times it's useful to get the route name of the current Page. (index, tag, ..)
I don't think it's possible anymore after https://github.com/flarum/core/pull/2156. 
Could be easily re-implemented: 
For mithril v.0.2 inside Page init():
```js
  app.current = new PageState(this.constructor, { routeName: this.props.routeName })
```
Would make it possible to get it through `app.current.get('routeName').`

askvortsov1:
Hmm, I don't know if this exact solution will work, since the attr being set was removed in https://github.com/flarum/core/pull/2255/commits/ccc77dd789de0f8a79802fb1ad8a3dba322fc519. Although my choice of commit title was definitely poor (since this isn't obsolete, really), we now provide component classes, not component instances to mithil v2. I'm not sure that we can set this attr on the page components directly... The first thing that comes to mind is keeping track of routeNames somewhere parallel, but that doesn't seem particularly robust. Any thoughts @clarkwinkelmann @datitisev

datitisev:
I feel like it'd make sense if Application knew the current route name, as it has the current and previous route data. If not, perhaps app.history ?

askvortsov1:
Problem is, once we generate a route with app.route, there's nothing tying that to the route name. We could, of course, use a hashmap to store routes generated by `app.route` and map them back to their original route name, but that could grow unweildy pretty quickly.

clarkwinkelmann:
Can we maybe use a route render method to achieve that?

```js
  for (const key in routes) {
    const route = routes[key];

    map[basePath + route.path] = {
      render() {
        return m(route.component, {routeName: key});
      }
    };
  }
```

askvortsov1:
Oooo I like that! Not sure if there's any detriment to always using route resolvers though.... But then again, that was my plan for https://github.com/flarum/core/pull/2249 too!

> EDIT: but it won't be available from the global scope though...

Couldn't we just add that to PageState in oninit? Or somewhere?

clarkwinkelmann:
Yep, I've edited myself multiple times... I think it would be perfectly reasonable to set that value in the pagestate or elsewhere on the `app` object from `Page.oninit`, and the value can be cascaded from a route resolved like shown in my snippet.

askvortsov1:
@clarkwinkelmann @datitisev is that something we want in the main PR, or as a separate one on top of that?

[...]

PR #2275, opened 2020-08-27
Allow extensions to use route resolvers

askvortsov1:
Depends on https://github.com/flarum/core/pull/2378.

See https://github.com/flarum/core/issues/2266.

Mithril Route Resolvers are new features in Mithril 2 that allow some logic to be inserted before a route is handled. Essentially, a quick layer before the component: https://mithril.js.org/route.html#routeresolver. Essentially, its 2 methods allow for hacking in:

- Before the page is first rendered (so we could run authentication or use a different component, for instance)
- Before every redraw while on a page

Currently, Route Resolvers are used to pass in the Flarum route name as an attr to the component handling the route. We will also be using them to process 403 and 404 errors in the frontend (instead of having dedicated blade pages).

The current implementation of mapRoutes automatically wraps all components in a route resolver. This is nice, but it doesn't allow extensions to use custom route resolvers in their routes. As such, this PR provides the following options for mapRoute:

- the `resolver` key can be used to provide an INSTANCE of a route resolver. This instance should define which component should be used, and hardcode the route name to be passed into it. This instance will be used without any modifications by Flarum.
- The `resolverClass` key AND `component` key can be used to provide a CLASS that will be used to instantiate a route resolver, to be used instead of Flarum's default one, as well as the component to use. It's constructor should take 2 arguments: `(component, routeName)`. We recommend extending the `flarum/utils/DefaultResolver` class.
- The `component` key can be used alone to provide a component. This will result in the default behavior.

We also use a custom route resolver for DiscussionPage so that routing from one post to another in the same discussion results in a scroll, not a full page rerender.


clarkwinkelmann:
My suggestion would actually be that we add validation for `route.component` and only accept components there.

Then we can add a second property, like `route.resolver` which will be checked to be a valid resolver.

I see there's an interface for `RouteResolver` we could use https://github.com/MithrilJS/mithril.d.ts/blob/84a87fefe18868c1b56a0a9a09efc7443a1fbd0f/index.d.ts#L58

While it might be helpful to provide an extendable base resolver, we need to also make sure devs can register resolvers that don't extend that base resolver, since there would otherwise be no way to have only `onmatch` or only `render`

datitisev:
@clarkwinkelmann Those are TS typings, we can't use them to check in JS. They most likely don't translate to anything in JS.

clarkwinkelmann:
Yes, I know. I was just saying we should use the typings since that code is TS already.

The part about checking the parameters will have to use basic javascript and not interfaces.

askvortsov1:
> My suggestion would actually be that we add validation for `route.component` and only accept components there.
> 
> Then we can add a second property, like `route.resolver` which will be checked to be a valid resolver.

I would be alright with that. The only question is, how do we want to handle providing routeName to a custom resolver? I'd say we should expect a setRouteName method on all custom resolvers then?

clarkwinkelmann:
We don't necessarily need to pass the routeName to the resolver. The resolver could also hard-code its routeName in the case of custom resolvers.

I'd say we add a check in the `Page` component to log a warning if no `attrs.routeName` is passed, and let extensions do as they please to pass it in their custom resolvers. If they extend our base resolver, they could re-use most of the logic.

Something else we could do is check whether the `route.resolver` parameter is a class. If it's a class (or alternatively, that it extends our class), we call `new` and pass the same parameters as we do to our DefaultResolver. If the parameter isn't a class but just a POJO, we pass it as-it, assuming the routeName is already hard-coded or stored in there.

askvortsov1:
@clarkwinkelmann how's this?

clarkwinkelmann (approved):
I'm not sure that `resolverClass` feature is really needed, but otherwise looking good.

Functionnaly I guess it's the same but why use `onmatch` and not return `m(this.component, [...])` in `render()` like previously?

askvortsov1:
> Functionnaly I guess it's the same but why use `onmatch` and not return `m(this.component, [...])` in `render()` like previously?

By doing that, we'd be effectively killing off the ability to use/modify onmatch. It's a useful feature, and I'm planning to use it in https://github.com/flarum/core/pull/2249 to show error pages when relevant (passing the error details in through the payload).

franzliedke:
Could the PR description get a few more details on the actual use-case here? What is the intention here? It's a bit difficult sifting through so much discussion about temporary, now-obsolete changes in the huge Mithril branch.

askvortsov1:
> Could the PR description get a few more details on the actual use-case here? What is the intention here? It's a bit difficult sifting through so much discussion about temporary, now-obsolete changes in the huge Mithril branch.

Done, and rebased

franzliedke:
I still find it hard to see actual concrete use-cases, to be honest. The only concrete ones seems to be the "frontend error handling" one (#2249), and I've raised my concerns about that in the relevant issue and PR.

That said, I would first recommend to spend time on actual abstractions for pages / routes, to get hold of the hacks that were introduced around routing in the upgrade to Mithril 2 and then revisit this if there are actual extension needs - we might not need the resolvers in core anymore, or have better extension points to offer at that time.

askvortsov1:
Resolvers are a mithril 2 feature, and even if we can't come up with use cases now, I'm not sure why we shouldn't try to support the feature itself, especially when it's not difficult to do so

datitisev (approved):
I think this is good (???)

askvortsov1:
> I still find it hard to see actual concrete use-cases, to be honest. The only concrete ones seems to be the "frontend error handling" one (#2249), and I've raised my concerns about that in the relevant issue and PR.
> 
> That said, I would first recommend to spend time on actual abstractions for pages / routes, to get hold of the hacks that were introduced around routing in the upgrade to Mithril 2 and then revisit this if there are actual extension needs - we might not need the resolvers in core anymore, or have better extension points to offer at that time.

Could you go a bit more into detail about your vision for page/route abstraction? We need to decide whether we want to include this in b14

clarkwinkelmann:
I don't think this is an urgent matter, I'd recommend keeping this for the next release cycle.

It would be nice to have at least one (community?) extension making use of this feature to validate its implementation.

js/src/common/resolvers/DefaultResolver.ts
```diff
@@ -0,0 +1,34 @@
...
+  makeKey() {
+    return this.routeName + JSON.stringify(m.route.param());
+  }
+
+  onmatch(args, requestedPath, route) {
+    return this.component;
+  }
+
+  render(vnode) {
+    return { ...vnode, routeName: this.routeName, key: this.makeKey() };
```
w-4, line 32: We need to wrap the object into an array for the key to work properly I believe. An easy way to debug this is to include a `console.log('init')` inside `Page.onint` and go from the 'All discussions' page to a tag page and see if the oninit was run.

w-4 (changes requested):

js/src/forum/resolver/DiscussionPageResolver.ts
```diff
@@ -0,0 +1,40 @@
...
+    }
+    if ('id' in params && params.id.includes('-')) {
+      params.id = params.id.split('-')[0];
+    }
+    return this.routeName.replace('.near', '') + JSON.stringify(params);
+  }
+
+  onmatch(args, requestedPath, route) {
+    const sameDiscussion = m.route.param('id') && args.id && m.route.param('id').split('-')[0] === args.id.split('-')[0];
+    if (route.startsWith('/d/:id') && sameDiscussion) {
```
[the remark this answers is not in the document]
askvortsov1: Near might not be in args if we're linking to the first post in a discussion, for which near isn't used.
startsWith shouldn't result in issues in a subfolder strategy, as routes are relative to the base route. That being said, I changed it to `includes`.

datitisev (approved):
The code looks clean and organized, and nothing else is added apart from the resolver for Discussion so :+1:

Not tested locally.

js/src/forum/resolver/DiscussionPageResolver.ts
```diff
@@ -0,0 +1,47 @@
...
+    const params = { ...m.route.param() };
+    if ('near' in params) {
+      delete params.near;
+    }
+    params.id = getDiscussionIdFromSlug(params.id);
+    return this.routeName.replace('.near', '') + JSON.stringify(params);
+  }
+
+  onmatch(args, requestedPath, route) {
+    if (route.includes('/d/:id') && getDiscussionIdFromSlug(args.id) === getDiscussionIdFromSlug(m.route.param('id'))) {
```
w-4, line 30: What would happen if we remove the route includes check? Isn't the `onmatch` of the DiscussionPageResolver only run for discussion routes?

askvortsov1: So we need to check 2 parts: that we are currently on a DiscussionPage, and that the page we are going to is a DiscussionPage. That we are in `onmatch` of the DiscussionPageResolver gives us the first bit, but we still need to ensure the second. That's why there's no `app.current.matches(DiscussionPage)` here.

w-4: I see what you mean, but I'm not sure that this check would work.
 
1. From IndexPage to DiscussionPage : DiscussionPageResolver onmatch is run and route includes '/d/:id'
2. From DiscussionPage to DiscussionPage : same, route includes '/d/:id'
3. From DiscussionPage to IndexPage : DiscussionPageResolver onmatch is not run, instead we run the DefaultResolver onmatch.

We can remove the route.includes statement because it would be always true when DiscussionPageResolver is run. 
If we want to check whether we're going from DiscussionPage to DiscussionPage (to avoid the highly unlikely edge case that the previous page has also an id param and the id's match), using `app.current.matches(DiscussionPage)` should work.  

Here's the console.log statement that I used in onmatch to debug this:
```js    
console.log({
      route,
      requestedPath,
      id: getDiscussionIdFromSlug(args.id),
      oldId: getDiscussionIdFromSlug(m.route.param('id')),
      getRoute: m.route.get(),
      matchesAppCurrent: app.current.matches(DiscussionPage),
    });
`

askvortsov1: Of course... onmatch runs for the resolver of the route we are going to, not of the route we are currently on :facepalming:

PR #2276, opened 2020-08-27
Prevent discussions being displayed twice

### LINKED ISSUE #2268: [Frontend Rewrite] DiscussionList double results bug
## Bug Report

Introduced in https://github.com/flarum/core/pull/2150

**Current Behavior**
When a user navigates away (to a different tag for example) while the results are loading, it will lead to both the previous and new results being added to app.discussions. 

```js
  refresh({ clear = true } = {}) {
    this.loading = true;

    if (clear) {
      this.clear();
    }

    return this.loadResults().then(
      (results) => {
        this.parseResults(results);
      },
      () => {
        this.loading = false;
        m.redraw();
      }
    );
  }

```

**Possible Solution**

Add an additional `this.discussions = []` before calling parseResults. Not perfect, since results are parsed twice still, but in most cases it wouldn't be visually noticeable.

askvortsov1:
Fixes #2268

### DIFF
--- a/js/src/forum/states/DiscussionListState.js
+++ b/js/src/forum/states/DiscussionListState.js
@@ -88,6 +88,11 @@ export default class DiscussionListState {
 
     return this.loadResults().then(
       (results) => {
+        // If `this.loading` is false, that means that a prior request loaded while this one was in process.
+        // As such, before we add on the discussion items we just loaded in, we want to clear the list again.
+        if (!this.loading) {
+          this.discussions = [];
+        }
         this.parseResults(results);
       },
       () => {
--- a/js/src/forum/states/DiscussionListState.js
+++ b/js/src/forum/states/DiscussionListState.js
@@ -90,8 +90,8 @@ export default class DiscussionListState {
       (results) => {
         // If `this.loading` is false, that means that a prior request loaded while this one was in process.
         // As such, before we add on the discussion items we just loaded in, we want to clear the list again.
-        if (!this.loading) {
-          this.discussions = [];
+        if (clear && !this.loading) {
+          this.clear();
         }
         this.parseResults(results);
       },
--- a/js/src/forum/states/DiscussionListState.js
+++ b/js/src/forum/states/DiscussionListState.js
@@ -88,11 +88,9 @@ export default class DiscussionListState {
 
     return this.loadResults().then(
       (results) => {
-        // If `this.loading` is false, that means that a prior request loaded while this one was in process.
-        // As such, before we add on the discussion items we just loaded in, we want to clear the list again.
-        if (clear && !this.loading) {
-          this.clear();
-        }
+        // Just in case we had 2 requests going at the same time, let's clear our cached list to ensure that
+        // we only take into account the latest set of data.
+        this.clear();
         this.parseResults(results);
       },
       () => {


clarkwinkelmann:
Can't we just read the `clear` flag again? Or does the issue also happens in situations where `clear` is `false`?

Also it would be neat to re-use the existing `this.clear();` method, but I'm not sure whether this triggers two redraws or if Mithril would automatically delay the redraw in `this.clear();` until the second redraw.

In any case, I think a comment would make sense above that additional code, mentioning the "race condition".

askvortsov1:
How's this?

clarkwinkelmann (approved):
sounds good to me. We could even get rid of the `!this.loading` at this point and just do a precautionary clear every time.

w-4:
What's the use case of calling refresh with `{ clear: false }` ? I couldn't find it being called like this anywhere.

askvortsov1:
> What's the use case of calling refresh with `{ clear: false }` ? I couldn't find it being called like this anywhere.

To be honest, I don't completely remember. I looked through the relevant PR, and the closest I found was this: https://github.com/flarum/core/pull/2150#discussion_r431223384. My best guess would be that the old version had separate `refresh`, `refreshParams`, and `clear` calls (or the equivalent) from the `IndexPage` component, and in each iteration of PR review, we wanted to preserve options. By the end, we were left with something that didn't make sense, but was a logical transition from the previous iteration. 

It could almost definitely be removed to simplify the method (and accordingly that ugly method signature), but I'd prefer to do so in a separate PR after we merge the rewrite.

js/src/forum/states/DiscussionListState.js
```diff
@@ -88,6 +88,11 @@ export default class DiscussionListState {
 
     return this.loadResults().then(
       (results) => {
+        // If `this.loading` is false, that means that a prior request loaded while this one was in process.
+        // As such, before we add on the discussion items we just loaded in, we want to clear the list again.
+        if (clear && !this.loading) {
```
w-4, line 93: I'd keep the `clear: false` option and just remove the check on this line to always clear the discussions. This way calling `refresh({clear: false})` could refresh the `DiscussionList` without the loading animation. Could be quite useful for when you unfollow a discussion, the `DiscussionList` could be updated in a smooth way by calling this method.

askvortsov1: I like that approach! Done

js/src/forum/states/DiscussionListState.js
```diff
@@ -88,6 +88,9 @@ export default class DiscussionListState {
 
     return this.loadResults().then(
       (results) => {
+        // Just in case we had 2 requests going at the same time, let's clear our cached list to ensure that
+        // we only take into account the latest set of data.
+        this.clear();
```
w-4, line 93: Clearing without the m.redraw() here would be better. I tested locally that this leads to an additional redraw in mithril v0.2

askvortsov1: I'm fine changing it back to directly setting this.discussions without the `clear` method. @clarkwinkelmann how would you feel about reverting that change?

franzliedke (commented):
Since we're at the state level here, I'd expect the fix to revolve around ignoring the results of requests when another request has been triggered before the first one returned.

askvortsov1:
So effectively, don't even send requests if we've already sent one until that one has completed? To me it feels like later requests should take precedence (otherwise we get cases where refresh is called and inexplicably does nothing: if refresh is called while still loading that means that we switched to another page (while still loading), and the user would expect that the second page be rendered

datitisev:
> don't even send requests if we've already sent one until that one has completed

We definitely *do not* want to do this. We can store a randomly generated key / request count and check if it is the same when the request completes / errors - if it's not, ignore. If it is the same, show error and/or update discussion list.

askvortsov1:
To be honest, I like what we have now most from a user perspective.

w-4:
When I tested it in mihtril v0, this error only seemed to occur when `m.redraw.strategy('diff')` was set to diff. 
On nightly flarum it seems to occur regularly (see screenshot below).. I'm not sure what changed, but if it's not possible to do it with some mithril setting, another solution could be to cancel all previous requests on route change manually. 
image

js/src/forum/states/DiscussionListState.js
clarkwinkelmann: Is my input still needed on this? Can you clarify? Revert something from which commit/PR?

The force pushes seem to have removed every context possible. I seem to remember the original PR was more than a one-line change.

askvortsov1:
Closing in favor of #2319, which is more robust

PR #2279, opened 2020-08-29
Fix composer header hidden by mobile browser

### LINKED ISSUE #1670: Reply composer window sometimes outside the edge of the screen, on mobile
## Bug Report

**Current Behavior**
When posting a new reply using Chrome for Android, it might happen that part of the reply box is hidden behind the Chrome UI.

**Steps to Reproduce**
1. Open any discussion on Chrome for Android
2. Scroll to the bottom of the page, and make sure that the Chrome header (the one containing the URL of the page) is hidden (it goes away when you scroll down, it reappears when you scroll up)
3. Press "Write a Reply..."
4. The reply "popup" is shown, but the title of the popup and the "send" button are hidden by the Chrome UI. A manual scroll is required to post the reply

**Expected Behavior**
The reply box/popup/textarea should be fully visible by default.

**Screenshots**

 

**Environment**
- Flarum version: 0.1.0-beta8
- Website URL: https://discuss.flarum.org
- Browser: Chrome 70.0.3538.110

**Additional Context**
I think this happens with other browsers and operating systems as well, and I'm not sure if there's a workaround for it. It's certainly annoying and a bit confusing for new users.

SychO9:
**Fixes #1670**

Fixes the position of the composer to the screen, you can test this in https://beta.flarum.site, just make sure you add the relevant CSS from the admin section.
```css
@media (max-width: 767px) {
    .Composer:not(.minimized) {
        position: fixed;
    }
}
```

My only concern is that looking at git blame, the composer's position was made absolute in 2015 to fix a bug with iOs 8/9, but there doesn't seem to be any mention of what that bug exactly was, so I cannot tell if this could potentially bring back that bug.

Relevant commit: https://github.com/flarum/core/commit/e6e2cdd3e98be6268f4dd7fb370d366f565ae72e

**Screenshot**
The issue has images of the bug fixed by this PR.

**Confirmed**
- ~[ ] Backend changes: tests are green (run `composer test`).~

### DIFF
--- a/less/forum/Composer.less
+++ b/less/forum/Composer.less
@@ -114,7 +114,7 @@
     background: @body-bg;
 
     &:not(.minimized) {
-      position: absolute;
+      position: fixed;
       top: 0;
       height: 350px !important;
       padding-top: @header-height-phone;
--- a/js/src/forum/components/Composer.js
+++ b/js/src/forum/components/Composer.js
@@ -270,7 +270,7 @@ export default class Composer extends Component {
     this.animateHeightChange().then(() => this.focus());
 
     if (app.screen() === 'phone') {
-      this.$().css('top', $(window).scrollTop());
+      this.$().css('top', 0);
       this.showBackdrop();
     }
   }


less/forum/Composer.less
```diff
@@ -114,7 +114,7 @@
     background: @body-bg;
 
     &:not(.minimized) {
-      position: absolute;
```
askvortsov1, line 117: My understanding is that this is intentional: we want people to be able to scroll up and down in a discussion while working on their reply. That being said, my understanding might be incorrect.

SychO9: oh, if that's the case then we wouldn't want to change this then. From the commit I mentioned above https://github.com/flarum/core/commit/e6e2cdd3e98be6268f4dd7fb370d366f565ae72e my understanding is that it was only done to fix a certain bug.

SychO9: Also, users can always minimize the composer to take a look at the discussion, so I'm not sure if scrolling up and down from there is really needed

askvortsov1: Not inherently opposed to changing, but let's revisit this in the next release cycle and hopefully have some discussion on it

zerosonesfun:
Original solution above, using fixed, for me completely hides composer now. 

This may be best. This ensures the composer is at the bottom, similar to how it is on desktop. It makes the most sense. Composer box adjacent to your virtual keyboard. "But, we want people to be able to scroll and see behind it," some say. 1) You can still scroll up and see other posts. 2) After typing enough content you start to see your new post behind the post box without scrolling. 3) There is the eyeball 👁 preview button. People should use that to see their post before posting. There is very limited real estate on mobile with the keyboard expanded. We should no longer play the “must see everything behind it” game when it comes to mobile. 

~~~

@media (max-width: 767px) {
    .Composer:not(.minimized) {
      top: initial !important;
      bottom: 0 !important;
        
    }
}
~~~

Looks like this:

matteocontrini:
@zerosonesfun I think there might be the risk that the composer gets hidden by the browser header on small screens, right?

SychO9:
@zerosonesfun did you only apply the LESS changes or did you also apply JS changes ?

zerosonesfun:
> @zerosonesfun did you only apply the LESS changes or did you also apply JS changes (i.e: all changes done by this PR) ?

Sorry, no I didn’t do a pull request. Just commenting that the original pull request probably won’t work.

zerosonesfun:
> @zerosonesfun I think there might be the risk that the composer gets hidden by the browser header on small screens, right?

Maybe. But, that’s a risk even if it is left as-is isn’t it? You could also make the composer height a little smaller. It’s close to 250px as-is. It could be changed to 200px or 220px. 

Another way of looking at this is: Unless a better fix can be found relatively quickly... Should Flarum be a good experience on super tiny screens or a good experience on very popular iPhone screens?

SychO9:
> Sorry, no I didn’t do a pull request. Just commenting that the original pull request probably won’t work.

It's hard to tell if you haven't also applied the js changes if this is a bug with the PR or not:
> Original solution above, using fixed, for me completely hides composer now.

---

> This may be best. This ensures the composer is at the bottom, similar to how it is on desktop.

That looks nice on your mobile screen, but how does it behave on smaller mobile screens where there is no space for both the keyboard and the composer ? I imagine you wouldn't be able to see the title field anymore.

Also, fixed positioning fixes a problem where the composer is stuck at a specific position in the screen, creating issues when closing the keyboard and other scenarios: https://discuss.flarum.org/d/24831-ui-bug-mobile-postdiscussion-is-not-on-the-screen-sometime

> Another way of looking at this is: Unless a better fix can be found relatively quickly.. Should Flarum be a good experience on super tiny screens or a good experience on very popular iPhone screens?

I don't understand what you mean, we are not forced to choose between both, this pull request fixes all issues.

zerosonesfun:
> > Sorry, no I didn’t do a pull request. Just commenting that the original pull request probably won’t work.
> 
> It's hard to tell if you haven't also applied the js changes if this is a bug with the PR or not:
> 
> > Original solution above, using fixed, for me completely hides composer now.
> 
> > This may be best. This ensures the composer is at the bottom, similar to how it is on desktop.
> 
> That looks nice on your mobile screen, but how does it behave on smaller mobile screens where there is no space for both the keyboard and the composer ? I imagine you wouldn't be able to see the title field anymore.
> 
> Also, fixed positioning fixes a problem where the composer is stuck at a specific position in the screen, creating issues when closing the keyboard and other scenarios: https://discuss.flarum.org/d/24831-ui-bug-mobile-postdiscussion-is-not-on-the-screen-sometime
> 
> > Another way of looking at this is: Unless a better fix can be found relatively quickly.. Should Flarum be a good experience on super tiny screens or a good experience on very popular iPhone screens?
> 
> I don't understand what you mean, we are not forced to choose between both, this pull request fixes all issues.

I didn’t look at the pull request. I didn’t realize it included JavaScript too. Sounds great! At least I have a CSS patch for my sites for now which seems to work. I’ll probably do some tiny screen tests later to see what all gets hidden. 

I like the pop up... but, sometimes I wonder if there’s a better fully responsive way. Because it’s not fully responsive right? That’s why we’re having the screen size issues?

SychO9:
> I like the pop up... but, sometimes I wonder if there’s a better fully responsive way. Because it’s not fully responsive right? That’s why we’re having the screen size issues? And having to rely on JavaScript to position it well?

That's what this pull request does, removes the need to position it with javascript, and fixes it to the screen instead. Currently the behavior is an absolute and a set position with JS, so that one can see the discussion behind while scrolling (but there is no need for that since you can minimize the composer anyway). The composer element itself is responsive because it adapts to a better view on smaller screens, it's just that because of the absolute position (changed to fixed in this PR) it creates the 2 linked to issues.

zerosonesfun:
> > I like the pop up... but, sometimes I wonder if there’s a better fully responsive way. Because it’s not fully responsive right? That’s why we’re having the screen size issues? And having to rely on JavaScript to position it well?
> 
> That's what this pull request does, removes the need to position it with javascript, and fixes it to the screen instead. Currently the behavior is an absolute and a set position with JS, so that one can see the discussion behind while scrolling (but there is no need for that since you can minimize the composer anyway). The composer element itself is responsive because it adapts to a better view on smaller screens, it's just that because of the absolute position (changed to fixed in this PR) it creates the 2 linked to issues.

Ok, thanks for explaining! I didn’t get it at first. I’ll hold my breath and cross my figures that all of the devs agree with this pull request for beta 16. I would think it needs to be in beta 16 to ensure full testing before stable.

SychO9:
Well thank you for testing and looking into this, it's really appreciated!

Also if you can't test by making the JS changes (it's more tedious to do so I know) you can use this, I don't think (not sure) it works well when composer is minimized, but it at least allows you to test when the composer is open on mobile :) without having to touch the JS
```less
@media (max-width: 767px) {
    .Composer:not(.minimized) {
        position: fixed;
        top: 0 !important;
    }
}
```

zerosonesfun:
> Well thank you for testing and looking into this, it's really appreciated!
> 
> Also if you can't test by making the JS changes (it's more tedious to do so I know) you can use this, I don't think (not sure) it works well when composer is minimized, but it at least allows you to test when the composer is open on mobile :) without having to touch the JS
> 
> ```css-less
> @media (max-width: 767px) {
>     .Composer:not(.minimized) {
>         position: fixed;
>         top: 0 !important;
>     }
> }
> ```

This works great! So, I guess the idea behind having it stuck to the top instead of stuck to the bottom has to do with previewing and tiny screens? My brain is like... hmm, having it at the top is fine. But, comment boxes are usually at the bottom of UIs, no? Regardless I’m going this this CSS now at my sites until there’s something permanent. I’m sure the JS change works too by the way.

SychO9:
> So, I guess the idea behind having it stuck to the top instead of stuck to the bottom has to do with previewing and tiny screens? My brain is like... hmm, having it at the top is fine. But, comment boxes are usually at the bottom of UIs, no?

having it at the bottom does look and behave better imo, but although I haven't tested, I am pretty sure if it were fixed to the bottom, smaller mobile screens would have a portion of the composer element hidden when the keyboard is open, because there wouldn't be enough space for both, so it just feels safer to fix it at the top.

askvortsov1:
> I am pretty sure if it were fixed to the bottom, smaller mobile screens would have a portion of the composer element hidden when the keyboard is open, because there wouldn't be enough space for both, so it just feels safer to fix it at the top.

If there isn't enough room for both, wouldnt it not matter if the composer is at the top or bottom, since the lack of space implies that all space below/above the composer would be taken up, and the composer and keyboard would be the only 2 elements on the page?

SychO9:
I guess I'm just assuming it would be better if the keyboard hid a part of the composer's bottom, then if the keyboard pushed the composer up to hide it's top :shrug:

askvortsov1 (approved):
This makes sense to me

tankerkiller125 (approved):

Issue #2284, opened 2020-09-03
IndexPage should not be scrolled on mobile devices when switching tags


iPurpl3x:
## Bug Report

**Current Behavior**
When you switch tags, the `config()` method of `IndexPage` scroll the page intelligently to keep that tags list (for navigation) always on the same position. This behavior only makes sense on desktop, where there is a side-bar for navigation.

**Steps to Reproduce**
1. Create 2 tags, one with no description, one with a long description
2. On a mobile device (or small browser window) go to the forum and switch from All Discussions to the first tag (the one with no description)
3. Switch to the other tag that has a long description. The page is scrolled and you cannot see the title of the current tag in the Hero element.

**Expected Behavior**
The automatic scroll behavior should not be applied on when on a mobile device.

**Environment**
- Flarum version: 0.1.0-beta.13
- Website URL: https://beta.flarum.site/
- PHP version: 7.3
- Browser: Firefox 81

askvortsov1:
Possibly the same as https://github.com/flarum/core/issues/2284?

iPurpl3x:
> Possibly the same as #2284?

🤔 ... this is #2284

askvortsov1:
Lol good job me! I meant https://github.com/flarum/core/issues/1951, but couldn't decide which one to mark as the duplicate, and so ended up making that very... obvious statement.

PR #2288, opened 2020-09-06
Don't scroll IndexPage oncreate if not on desktop

askvortsov1:
**Fixes https://github.com/flarum/core/issues/2284**

Only scroll IndexPage in oncreate if on desktop or larger.

Do we still need this with the Mithril 2 update changes?

**Confirmed**


datitisev:
@askvortsov1 Yes, we still need this with the mithril changes. However, we need this on the `onupdate` too with Mithril 2 because it's the same component instance when changing tags.

askvortsov1:
Hmm, but in that case, there's probably other oncreate stuff that we should move into that onbeforeupdate block...

PR #2291, opened 2020-09-08
Fix issue where posts API doesn't return the right amount of posts

### LINKED ISSUE #2071: Guest users can't load more posts on profiles

w-4:
**Fixes #2071**

Check for permissions when getting the post Ids. 

- [x] Tested locally

### DIFF
--- a/src/Api/Controller/ListPostsController.php
+++ b/src/Api/Controller/ListPostsController.php
@@ -108,12 +108,13 @@ class ListPostsController extends AbstractListController
      */
     private function getPostIds(ServerRequestInterface $request)
     {
+        $actor = $request->getAttribute('actor');
         $filter = $this->extractFilter($request);
         $sort = $this->extractSort($request);
         $limit = $this->extractLimit($request);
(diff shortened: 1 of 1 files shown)


askvortsov1 (approved):
Thank you very much!

datitisev (approved):
Test locally and it works. Thank you so much! I think I'll want Franz or luceos to take a look before merging, but this LGTM.

tankerkiller125 (approved):

Pi-Bouf (approved):

luceos (approved):

luceos:
We need to identify where we use this method and prevent the whereVisibleTo to be added after that.

clarkwinkelmann:
This sounds like a good fix.

> We need to identify where we use this method

You mean `getPostIds`? That one is private and used only from there.

Issue #245, opened 2015-08-27
Decide on testing strategy


tobscure:
With my limited knowledge, here's what I'm thinking:

* Unit tests for backend classes. I think phpspec is the preferred tool here – we already have one spec thanks to @franzliedke.
* Functional/integration tests for the JSON-API. (Not sure what tool would be best for this?)
* Functional/integration tests for the extension API. (Ditto?)
* Unit tests for front-end utils, components, helpers, and models. (Probably using QUnit?)
* Acceptance tests for the whole thing.

Thoughts?

franzliedke:
From my perspective, the most important things are **unit tests** for important classes (which we'll have to go looking for) and **acceptance tests** (because those cover the other things).

Luceos:
I see there was a help wanted. If you'd like me to contribute with this I'd love to create the tests for the php side of things (laravel, eloquent, user creation etc).

franzliedke:
Woah, I just played around with Codeception, and acceptance testing for the whole stack amounts to a whole lot of work. Therefore, I'd say we focus on the following for now:
- more PhpSpec unit tests
- integration (?) tests for the front-end components, one-by-one (cause that's where the magic is at)

We probably should add a few acceptance tests for the critical routes (such as registration and posting maybe), but limit it to a few...

What do you think?

tobscure:
Sounds good to me. I'll work on setting up a front-end testing framework, you focus on the backend? :)

tobscure:
What about acceptance testing the JSON-API? That would be easier than acceptance testing our whole stack, and is quite important since our app isn't the only one using the JSON-API.

@Luceos Sorry, didn't see your comment above. I'm sure @franzliedke would greatly appreciate your help with backend testing!

franzliedke:
Yeah, I'll have a look at how we can do JSON-API testing.

@Luceos If you ever feel bored enough to help with tests, you could look at testing some of the central base classes, e.g. the `Action` class. ;)

Luceos:
I'll take a look at that asap @franzliedke , i have more experience with travis though, so you've decided for codeception, I'll read up into that then.

franzliedke:
[...]

PR #2298, opened 2020-09-19
Jest Testing Configuration

### LINKED ISSUE #424: Set up frontend testing framework

askvortsov1:
Fixes https://github.com/flarum/core/issues/245

This is planned for the beta 15 release cycle.

This PR sets up the jest framework for frontend unit tests. In order to avoid hacking in babel config, we're making core js an es6 module. Only change for now is renaming webpack config (which will no longer be necessary when webpack 5 comes out).

REMAINING WORK:
- Typescript support is currently problematic: when `export default` is used, ts-jest sees this as exporting an object with a key called default. Also, remove the console.log (I included this for now so reviewers can see what's up)
- Revert https://github.com/flarum/core/pull/2298/commits/783f36c3b38ce121c02cc15b7f00aaf6c384dbf0 before merging (we can keep for now while we're testing)


franzliedke (commented):
Once #2311 is merged, we should probably add another badge to the README in this PR here. :grin:

franzliedke (approved):
Merge it, merge it! Tests are a multiplier. 👏🏼 😉

davwheat:
> This is planned for the beta 15 release cycle.

*Sweats profusely*

askvortsov1:
Ah, I forgot about this one for a bit. The whole ".default" requirement for TypeScript is problematic and confusingly inconsistent though. However, I suspect (and really hope!) that's caused by a config issue. Will dig deeper.

davwheat:
What's the default issue?

askvortsov1:
> Typescript support is currently problematic: when export default is used, ts-jest sees this as exporting an object with a key called default. Also, remove the console.log (I included this for now so reviewers can see what's up)

See the code: for the typescript test case, we need to import, then call `.default` on the imported stuff

davwheat:
Hmmm... In my recent project I didn't need to do this. I think Jest supports TS properly now.

askvortsov1:
> Hmmm... In my recent project I didn't need to do this. I think Jest supports TS properly now.
> 
> https://github.com/davwheat/let-me-really-google-that-for-you/blob/main/__tests__/ErrorBoundary.test.js

I'm making a new branch, that will be cleaner.

askvortsov1:
> Hmmm... In my recent project I didn't need to do this. I think Jest supports TS properly now.
> 
> https://github.com/davwheat/let-me-really-google-that-for-you/blob/main/__tests__/ErrorBoundary.test.js

Just one problem with that. From https://jestjs.io/docs/getting-started#using-typescript:

> However, there are some caveats to using TypeScript with Babel. Because TypeScript support in Babel is purely transpilation, Jest will not type-check your tests as they are run. If you want that, you can use ts-jest instead, or just run the TypeScript compiler tsc separately (or as part of your build process).

davwheat:
We should probably use `tsc`. There are some pretty big issues with core surrounding type-checking last time I tried it which we should probably fix.

Issue #2297, opened 2020-09-18
[Frontend Rewrite] PostStream Scroll errors ($(...).offset() is undefined)


askvortsov1:
See https://github.com/flarum/embed/pull/10#issuecomment-692893291

askvortsov1:
Also, https://github.com/flarum/core/pull/2255#discussion_r491036135

PR #2299, opened 2020-09-20
[Frontend Rewrite] PostStream Improvements

askvortsov1:
Fixes **#2297**

When merging, don't squash

See commit titles, descriptions.

One outstanding issue with PostStream is that sometimes, gotoIndex will take you to the end of the loaded range, in which case it'll start over. I'm not exactly sure why this is happening, but it occasionally (slightly) throws off the scroll. That being said, I doubt anyone is sitting there, clicking the scrubber like a maniac, which is practically needed for this to show up.

### DIFF
--- a/js/src/forum/components/DiscussionPage.js
+++ b/js/src/forum/components/DiscussionPage.js
@@ -160,10 +160,8 @@ export default class DiscussionPage extends Page {
    * @param {Discussion} discussion
    */
   show(discussion) {
-    this.discussion = discussion;
-
     app.history.push('discussion', discussion.title());
-    app.setTitle(this.discussion.title());
+    app.setTitle(discussion.title());
     app.setTitleCount(0);
 
     // When the API responds with a discussion, it will also include a number of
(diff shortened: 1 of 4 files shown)


clarkwinkelmann (approved):
If you say that works I'm going to believe you on that because I haven't time to test right now :sweat_smile: 

Does this also help with the `flarum/embed` issue or is it completely irrelevant?

askvortsov1:
It should, the issue with embed is just PostStream being wonky

js/src/forum/components/DiscussionPage.js
```diff
@@ -194,10 +192,12 @@ export default class DiscussionPage extends Page {
     // posts we want to display. Tell the stream to scroll down and highlight
     // the specific post that was routed to.
     this.stream = new PostStreamState(discussion, includedPosts);
-    this.stream.goToNumber(m.route.param('near') || (includedPosts[0] && includedPosts[0].number()), true);
+    this.stream.goToNumber(m.route.param('near') || (includedPosts[0] && includedPosts[0].number()), true).then(() => {
+      this.discussion = discussion;
```
datitisev, line 196: Why are we setting this.discussion later now? What does this affect/solve?

askvortsov1: Since the PostStream and PostStreamScrubber components are only shown in view when `this.discussion` is set, this way we set `this.discussion` only after the initial set of posts has loaded, so that there's less time for race conditions / stuff to go wrong with PostStreamScrubber animation glitches.

datitisev (approved):

js/src/forum/components/PostStream.js
```diff
@@ -351,12 +352,32 @@ export default class PostStream extends Component {
...
+      // first post was requested). In some cases, we may have scrolled to
+      // the end of the available post range, in which case, the next range
+      // of posts will be loaded in. However, in those cases, the post we
+      // requested won't exist, so scrolling to it would cause an error.
+      // Accordingly, we start by checking that it's offset is defined.
+      const offset = $(`.PostStream-item[data-index=${index}]`).offset();
+      if (index === 0) {
+        $(window).scrollTop(0);
+      } else if (offset) {
+        $(window).scrollTop($(`.PostStream-item[data-index=${index}]`).offset().top - this.getMarginTop());
```
franzliedke, line 378: You could use the `offset` variable here, I guess. :wink:

PR #2302, opened 2020-09-21
Replace $app->url() with url-generated link to index

askvortsov1:
The `url` method of Container was removed in #2142.  Works locally.

### DIFF
--- a/views/error/not_found.blade.php
+++ b/views/error/not_found.blade.php
@@ -5,7 +5,7 @@
     {{ $message }}
   </p>
   <p>
-    <a href="{{ $url->to('forum')->route('index') }}">
+    <a href="{{ $url->to('forum')->base() }}">
       {{ $translator->trans('core.views.error.not_found_return_link', ['{forum}' => $settings->get('forum_title')]) }}
     </a>
   </p>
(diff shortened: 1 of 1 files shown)


tankerkiller125 (approved):

clarkwinkelmann (changes requested):
I think everywhere else we make a link to the homepage we use `->to('forum')->base()` https://github.com/flarum/core/blob/df8f73bd3d656e073dcfff5bddd8e7692ca19162/src/Api/Serializer/ForumSerializer.php#L67

Or an alternative could be to use the `default` named route https://github.com/flarum/core/blob/2431df5602e218d77388cec0019d13a473912be0/src/Forum/ForumServiceProvider.php#L207

Using `index` would create a link to `/all` when a different homepage has been selected I believe.

askvortsov1:
Good catch, thanks!

clarkwinkelmann (approved):

franzliedke:
Removed this from the milestone as this is just a regression from another change and not relevant for the changelog or anything like that, IMO. I hope you don't mind.

askvortsov1:
Thanks! I only added it so that we don't forget to fix it before release. The milestone will need a comb-over before release anyways

PR #2304, opened 2020-09-24
Run Backend Tests in Transactions

askvortsov1:
Integration tests are now run in transactions. This has been separated into well-described commits, I recommend looking through those for more info on specifics of changes. Some highlights:

- populateDatabase no longer actually modifies the database, it just adds data to an object of seed data that will be applied after the transaction begins during app boot. This means:
  - populateDatabase no longer triggers app boot, so we can get rid of all the `prepDb` stuff and just use setUp
  - populateDatabase does nothing once the app is booted, so we need to use $this->database directly for those changes
- We now rely on the state of the installed Flarum site (with regards to default user (admin), groups, and permissions), so those do not need to be specified in per-test seed data
- A trait that enables purging of cached settings has been added. @tankerkiller125 never again shall we face this extremely annoying bug!
- Seed data for fulltext search test cases is inserted into (and removed from) the database outside of a transaction in a special case, I don't expect to see this anywhere else so there's no util for it.


tests/integration/extenders/ModelTest.php
```diff
@@ -200,6 +203,7 @@ public function custom_relationship_is_inherited_to_child_classes()
     public function custom_relationship_prioritizes_child_classes_within_2_parent_classes()
     {
         $this->extend(
+            // FIXME: Second has no effect
```
SychO9, line 206: this is no longer the case I believe ?

tests/integration/extenders/ModelTest.php
```diff
@@ -200,6 +203,7 @@ public function custom_relationship_is_inherited_to_child_classes()
...
             (new Extend\Model(Post::class))
                 ->belongsTo('ancestor', User::class, 'user_id'),
             (new Extend\Model(AbstractEventPost::class))
```
SychO9, line 209: This is no longer the case I believe ?

SychO9 (commented):
I was wondering why the display name driver test was failing for me locally despite clearing the settings cache. Makes sense now, I missed the fact that it is statically set when the app is booted.

It all looks good to me, nicely done!

tests/integration/extenders/ModelTest.php
askvortsov1: Confirmed and cleaned up, thanks!

SychO9 (approved):
LGTM

tankerkiller125 (approved):
Looks good code wise to me, I haven't tested locally but CI is passing. Given the massive changes here maybe one more set of eyeballs before merging?

luceos (approved):

PR #2212, opened 2020-06-27
Permissions Cleanup: Core

askvortsov1:
**Fixes #1680**

**Changes proposed in this pull request:**
- Rename viewDiscussions to viewForum
- Rename viewUserList to searchUsers
- Revert https://github.com/flarum/core/commit/40e4c0acddd5c29dec322e6588ecd84a89544fa4
- Add migrations to update permission keys
- Fix an improperly configured test

**Reviewers should focus on:**
This is marked as draft until we agree upon https://github.com/flarum/core/issues/2202

**Confirmed**

- [x] Frontend changes: tested on a local Flarum installation.
- [x] Backend changes: tests are green (run `composer test`).

Issue #1680, opened 2018-12-04
Guests can see user profile regardless of the permission "view user list"


MichaelBelgium:
## Bug Report
Related to #1676 and https://discuss.flarum.org/d/17835-the-requested-resource-was-not-found-on-refresh and cuz toby asked me to 😛 

**Current Behavior**
Guests can view user profiles regardless of the permission "view user list".  It only triggers the permission on refreshing the page of a user or when directly navigating to a user profile url.

**Steps to Reproduce**
1. Login and go the the permission page in the admin section
2. Set the permission "View user list" to "Members"
3. Log out 
4. Navigate to a discussion
5. Click on someone's name 
6. You can view their profile.

Optional:
7. Refresh the page or navigate directly to a user profile and see it working nonetheless, but displaying a vague error.

**Expected Behavior**
* A more clear error message after step 7. You get a `The requested resource was not found` and not some kind of `Permission denied` error
* Guests can't browse through a discussion to a user profile if the permission is set to members only

**Environment**
- Flarum version: beta 8
- Website URL: local
- Webserver: nginx
- Hosting environment: Laravel homestead
- PHP version: 7.2.9
- Browser: latest version of chrome

```
Flarum core 0.1.0-beta.8
PHP version: 7.2.9-1+ubuntu18.04.1+deb.sury.org+1
Loaded extensions: Core, date, libxml, openssl, pcre, zlib, filter, hash, pcntl, Reflection, SPL, sodium, session, standard, mysqlnd, PDO, xml, bcmath, calendar, ctype, curl, dom, mbstring, fileinfo, ftp, gd, gettext, iconv, igbinary, imap, intl, json, ldap, exif, msgpack, mysqli, pdo_mysql, pdo_pgsql, pdo_sqlite, pgsql, Phar, posix, readline, shmop, SimpleXML, soap, sockets, sqlite3, sysvmsg, sysvsem, sysvshm, tokenizer, wddx, xmlreader, xmlwriter, xsl, zip, memcached, blackfire, Zend OPcache, xdebug
+------------------------------+-------------------+------------------------------------------+
| Flarum Extensions            |                   |                                          |
+------------------------------+-------------------+------------------------------------------+
| ID                           | Version           | Commit                                   |
+------------------------------+-------------------+------------------------------------------+
| flarum-emoji                 | v0.1.0-beta.8     |                                          |
| flarum-lang-english          | v0.1.0-beta.8     |                                          |
[...]

PR #2305, opened 2020-09-24
Revert https://github.com/flarum/core/pull/1536

askvortsov1:
**Refs #1680**. The rest will be fixed in https://github.com/flarum/core/pull/2212, but we can get this particular bug fixed in time for the next release.

Essentially, among our permission scheme, we have the `viewDiscussions` and `viewUserList` permissions. The first actually serves as a generic access permission for the entire forum, while the second is intended for the LIST users API. IE, `viewUserList` controls whether the list of users can be searched. This is further explained in https://github.com/flarum/core/issues/1680#issuecomment-444286686.

This PR fixes the second part of that comment, so that people stop seeing errors when refreshing the page after going to a user profile. This means that the `viewUserList` permission can now be used as intended, and unnecessary user searching can be restricted.

The rest (renaming permissions) will be done in https://github.com/flarum/core/pull/2212, but we need to discuss (and agree upon) a permission naming convention first.

### DIFF
--- a/src/User/UserPolicy.php
+++ b/src/User/UserPolicy.php
@@ -36,7 +36,7 @@ class UserPolicy extends AbstractPolicy
      */
     public function find(User $actor, Builder $query)
     {
-        if ($actor->cannot('viewUserList')) {
+        if ($actor->cannot('viewDiscussions')) {
             if ($actor->isGuest()) {
                 $query->whereRaw('FALSE');
             } else {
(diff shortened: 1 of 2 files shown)


tankerkiller125 (approved):

clarkwinkelmann:
Context? All the linked stuff seems to be from 2018?

askvortsov1:
Sorry about that, I've updated the PR description.

clarkwinkelmann (approved):
Ok, makes more sense.

Maybe I wouldn't have called this a revert since it's so long ago, but I agree with the change. I didn't test locally though.

PR #2308, opened 2020-09-24
Changes methods and properties from private to protected

luceos:

The goal of this PR is to offer increased flexibility for integrators and
custom solutions in skeleton modifications.

### DIFF
--- a/src/Foundation/InstalledApp.php
+++ b/src/Foundation/InstalledApp.php
@@ -70,12 +70,12 @@ class InstalledApp implements AppInterface
         return $pipe;
     }
 
-    private function inMaintenanceMode(): bool
+    protected function inMaintenanceMode(): bool
     {
         return $this->config['offline'] ?? false;
     }
 
-    private function needsUpdate(): bool
+    protected function needsUpdate(): bool
     {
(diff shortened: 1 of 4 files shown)


askvortsov1 (approved):

tankerkiller125 (approved):

luceos:
That was fast 🙈

tankerkiller125:
@luceos I watch Github like a hawk these days for minor improvements I can approve easily. I'm working on those review chops.

franzliedke:
Hey, these interfaces are small. And most of these classes could be considered internal. But hey. :wink:

franzliedke:
For example, #2271 will remove at least one of those methods, which may also break your overrides unexpectedly. All I'm saying is you will have to watch this space for changes anyway, and I don't consider these internals to be part of the API that's protected from BC breaks.

luceos:
I do get it. We talked about this before. Someone who overrides those classes does so at their own risk and expertise. Not providing this ability seriously hampers extensibility and forces someone to fully copy complete classes or sometimes multiple. That shouldn't be our goal.

The idea of the skeleton is that one can create their own fairly simple. Using private on all classes that the skeleton uses seriously hampers the flexibility we supposedly provide. Thats all.

PR #2309, opened 2020-09-24
Various TypeScript improvements

askvortsov1:
See commit titles/descriptions

### DIFF
--- a/js/shims.d.ts
+++ b/js/shims.d.ts
@@ -1,6 +1,5 @@
 // Mithril
-import * as Mithril from 'mithril';
-import Stream from 'mithril/stream';
+import Mithril from 'mithril';
 
 // Other third-party libs
 import * as _dayjs from 'dayjs';
@@ -29,7 +28,7 @@ export as namespace Mithril;
  */
 declare global {
   const $: typeof _$;
-  const m: m;
+  const m: Mithril.Static;
   const dayjs: typeof _dayjs;
 }
 
--- a/js/shims.d.ts
+++ b/js/shims.d.ts
@@ -10,13 +10,6 @@ import * as _$ from 'jquery';
 import Application from './src/common/Application';
 
 /**
- * Helpers that flarum/core patches into Mithril
- */
-interface m extends Mithril.Static {
-  stream: typeof Stream;
-}
-
-/**
  * Export Mithril typings globally.
  *
  * This lets us use these typings without an extra import everywhere we use
--- a/js/shims.d.ts
+++ b/js/shims.d.ts
@@ -13,7 +13,7 @@ import Application from './src/common/Application';
  * Helpers that flarum/core patches into Mithril
  */
 interface m extends Mithril.Static {
-  prop: typeof Stream;
+  stream: typeof Stream;
 }
 
 /**
--- a/js/src/common/Component.ts
+++ b/js/src/common/Component.ts
@@ -32,7 +32,7 @@ export interface ComponentAttrs extends Mithril.Attributes {}
  *
  * @see https://mithril.js.org/components.html
  */
-export default abstract class Component<T extends ComponentAttrs> implements Mithril.ClassComponent<T> {
+export default abstract class Component<T extends ComponentAttrs = ComponentAttrs> implements Mithril.ClassComponent<T> {
   /**
    * The root DOM element for the component.
    */
--- a/js/src/common/Component.ts
+++ b/js/src/common/Component.ts
@@ -3,11 +3,7 @@ import * as Mithril from 'mithril';
 let deprecatedPropsWarned = false;
 let deprecatedInitPropsWarned = false;
 
-export type ComponentAttrs = {
-  className?: string;
-
-  [key: string]: any;
-};
+export interface ComponentAttrs extends Mithril.Attributes {}
 
 /**
  * The `Component` class defines a user interface 'building block'. A component
@@ -36,7 +32,7 @@ export type ComponentAttrs = {
  *
  * @see https://mithril.js.org/components.html
  */
-export default abstract class Component<T extends ComponentAttrs = any> implements Mithril.ClassComponent<T> {
+export default abstract class Component<T extends ComponentAttrs> implements Mithril.ClassComponent<T> {
   /**
    * The root DOM element for the component.
    */
--- a/js/src/common/components/Alert.tsx
+++ b/js/src/common/components/Alert.tsx
@@ -1,31 +1,33 @@
-import Component from '../Component';
+import Component, { ComponentAttrs } from '../Component';
 import Button from './Button';
 import listItems from '../helpers/listItems';
 import extract from '../utils/extract';
+import Mithril from 'mithril';
+
+export interface AlertAttrs extends ComponentAttrs {
+  /** The type of alert this is. Will be used to give the alert a class name of `Alert--{type}`. */
+  type: string;
+  /** An array of controls to show in the alert. */
+  controls: Mithril.Children;
+  /** Whether or not the alert can be dismissed. */
+  dismissible: boolean;
+  /** A callback to run when the alert is dismissed */
+  ondismiss: Function;
+}
 
 /**
  * The `Alert` component represents an alert box, which contains a message,
  * some controls, and may be dismissible.
- *
- * ### Attrs
- *
- * - `type` The type of alert this is. Will be used to give the alert a class
- *   name of `Alert--{type}`.
- * - `controls` An array of controls to show in the alert.
- * - `dismissible` Whether or not the alert can be dismissed.
- * - `ondismiss` A callback to run when the alert is dismissed.
- *
- * All other attrs will be assigned as attributes on the DOM element.
  */
-export default class Alert extends Component {
-  view(vnode) {
+export default class Alert<T extends AlertAttrs> extends Component<T> {
+  view(vnode: Mithril.Vnode) {
     const attrs = Object.assign({}, this.attrs);
 
     const type = extract(attrs, 'type');
     attrs.className = 'Alert Alert--' + type + ' ' + (attrs.className || '');
 
     const content = extract(attrs, 'content') || vnode.children;
-    const controls = extract(attrs, 'controls') || [];
+    const controls = (extract(attrs, 'controls') || []) as Mithril.ChildArray;
 
     // If the alert is meant to be dismissible (which is the case by default),
     // then we will create a dismiss button to append as the final control in
--- a/js/src/common/components/Alert.tsx
+++ b/js/src/common/components/Alert.tsx
@@ -6,20 +6,20 @@ import Mithril from 'mithril';
 
 export interface AlertAttrs extends ComponentAttrs {
   /** The type of alert this is. Will be used to give the alert a class name of `Alert--{type}`. */
-  type: string;
+  type?: string;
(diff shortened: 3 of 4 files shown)


js/src/common/Component.ts
```diff
@@ -33,7 +29,7 @@ export type ComponentAttrs = {
  *
  * @see https://mithril.js.org/components.html
  */
-export default abstract class Component<T extends ComponentAttrs = any> implements Mithril.ClassComponent<T> {
+export default abstract class Component<T extends ComponentAttrs> implements Mithril.ClassComponent<T> {
```
datitisev, line 32: Why remove `= any`? To be honest, it *should* be `= ComponentAttrs` instead, so that TS linters (IDEs, etc) do not complain when a component doesn't pass its own interface.

datitisev (commented):
Everything is `public` by default, I don't think there's any point in adding `public` everywhere - just `protected` or `private` (which won't convert into protected/private JS, it'll just warn).

js/src/common/Component.ts
askvortsov1: I thought it wasn't necessary tbh. But yeah that's a good point, I'll add that back in

askvortsov1:
> Everything is `public` by default, I don't think there's any point in adding `public` everywhere - just `protected` or `private` (which won't convert into protected/private JS, it'll just warn).

I like having it there, if only to explicitly specify the difference when looking at code. Open to changing my mind on this one though

js/src/common/Component.ts
```diff
@@ -1,10 +1,6 @@
 import * as Mithril from 'mithril';
 
-export type ComponentAttrs = {
-  className?: string;
-
-  [key: string]: any;
-};
+export interface ComponentAttrs extends Mithril.Attributes {}
```
franzliedke, line 3: What's the benefit of exporting this (and using it further below) over using `Mithril.Attributes` directly?

js/src/common/components/Alert.tsx
```diff
@@ -1,31 +1,33 @@
-import Component from '../Component';
+import Component, { ComponentAttrs } from '../Component';
 import Button from './Button';
 import listItems from '../helpers/listItems';
 import extract from '../utils/extract';
+import Mithril from 'mithril';
```
franzliedke, line 5: So do we need to import this or not? IIRC from the TypeScript discussions in the mithril-2 branch, I was told that we're exposing this namespace in the shims so that it does not need to be imported.

franzliedke: Same in other file.

js/src/common/states/AlertManagerState.ts
```diff
@@ -1,30 +1,41 @@
...
    * Show an Alert in the alerts area.
+   *
+   * @returns The alert's ID, which can be used to dismiss the alert.
    */
-  show(arg1, arg2, arg3) {
+  public show(children: Mithril.Children): number;
+  public show(attrs: AlertAttrs, children: Mithril.Children): number;
+  public show(componentClass: Alert, attrs: AlertAttrs, children: Mithril.Children): number;
+
+  public show(arg1: Mithril.Children | AlertAttrs | Alert, arg2?: AlertAttrs | Mithril.Children, arg3?: Mithril.Children) {
```
franzliedke, line 27: If I understand [the docs](https://www.typescriptlang.org/docs/handbook/functions.html#overloads) correctly, this last signature isn't actually callable (otherwise, it would currently allow passing three arguments of type `Mithril.Children`). Therefore, we can improve readability a bit by doing `arg1: any, arg2: any, arg3: any`. (This would probably require a few typecasts below.) What do you think?

js/src/common/states/AlertManagerState.ts
```diff
@@ -1,30 +1,41 @@
...
 
   /**
    * Show an Alert in the alerts area.
+   *
+   * @returns The alert's ID, which can be used to dismiss the alert.
    */
-  show(arg1, arg2, arg3) {
+  public show(children: Mithril.Children): number;
+  public show(attrs: AlertAttrs, children: Mithril.Children): number;
+  public show(componentClass: Alert, attrs: AlertAttrs, children: Mithril.Children): number;
```
franzliedke, line 25: Could we return an explicit `AlertIdentifier` type here (that can extend from `number`, assuming this is possible in TS), to hide that implementation detail a little more? (And signal that this is an opaque value, not to be used except for memorizing the alert in question.)

franzliedke (changes requested):
Nice, I like where this is going!

franzliedke:
I tend to agree with @datitisev about `public`. Typescript already adds a lot of boilerplate, so seeing something that's plain modern JS is a nice feeling. :laughing:

js/src/common/components/Alert.tsx
askvortsov1: I think this is the typing library, not the actual library. @datitisev knows more about this though

js/src/common/Component.ts
askvortsov1: We import Component in every file anyway, so that decreases the import lines we need by 1. Also, it gives us better flexibility to change things later.

js/src/common/states/AlertManagerState.ts
askvortsov1: Done, good point

js/src/common/states/AlertManagerState.ts
askvortsov1: Done, typecasts were necessary anyways

js/src/common/components/Alert.tsx
franzliedke: But you are importing this only for the types, as far as I can tell?

js/src/common/states/AlertManagerState.ts
```diff
@@ -1,30 +1,48 @@
...
+  attrs: AlertAttrs;
+  children: Mithril.Children;
+}
 
 export default class AlertManagerState {
-  constructor() {
-    this.activeAlerts = {};
-    this.alertId = 0;
-  }
+  protected activeAlerts: { [ids: string]: AlertState } = {};
```
franzliedke, line 16: Shouldn't the `ids` be of type `number`, or better `AlertIdentifier`?

js/src/common/components/Alert.tsx
```diff
@@ -1,31 +1,33 @@
...
- *   name of `Alert--{type}`.
- * - `controls` An array of controls to show in the alert.
- * - `dismissible` Whether or not the alert can be dismissed.
- * - `ondismiss` A callback to run when the alert is dismissed.
- *
- * All other attrs will be assigned as attributes on the DOM element.
  */
-export default class Alert extends Component {
-  view(vnode) {
+export default class Alert<T extends AlertAttrs = AlertAttrs> extends Component<T> {
```
franzliedke, line 22: This is probably just me not knowing TypeScript well enough, but would this work as well?

```suggestion
export default class Alert extends Component<AlertAttrs> {
```

franzliedke: Especially seeing that `T` is not used anywhere within the class.

js/src/common/states/AlertManagerState.ts
datitisev: Also, `ids` here should be singular - you're typing each individual object key as one, not combined. Not that it matters, but you know 😛

js/src/common/components/Alert.tsx
datitisev: This would work until a class tries to extend Alert and pass its own ComponentAttrs.

js/src/common/components/Alert.tsx
datitisev: I don't entirely remember anything 👼 

Will have to check in IDE to see what it complains about.

js/src/common/components/Alert.tsx
franzliedke: Hmm, okay, but even the base interfaces are extensible with any custom keys, right? But I see the point...

datitisev: > Hmm, okay, but even the base interfaces are extensible with any custom keys, right? But I see the point...

Not sure what you mean by this. We can pass things to `Component<>` because we have basically the same `<T extends ... = ..>` but with `ComponentAttrs`. We need to explicitly allow this in Component classes, else if they pass something it won't be registered as component props because we're already passing e.g. AlertAttrs ``and that can't be changed.

franzliedke: I meant that `ComponentAttrs` (and therefore `AlertAttrs`) already allows any key in attrs, beyond the well-defined ones. So the usefulness of custom typings for the subclasses' props is limited.

datitisev: The usefulness is to type the props themselves so when you use the class it tells you what the props should be & their type, and then when accessing them from within the class too.

js/shims.d.ts
```diff
@@ -13,7 +13,7 @@ import Application from './src/common/Application';
  * Helpers that flarum/core patches into Mithril
  */
 interface m extends Mithril.Static {
-  prop: typeof Stream;
+  stream: typeof Stream;
```
clarkwinkelmann, line 16: Is this still needed after #2316 ? I think `m.prop` is only there for BC with beta13 and `m.stream` is only there for BC with master extensions that haven't been updated yet? So this should no longer be needed with beta14?

js/src/common/states/AlertManagerState.ts
askvortsov1: According to TS docs, it has to be either number or string. I guess number is better here

js/shims.d.ts
askvortsov1: yeah that can be removed

clarkwinkelmann (approved):

js/shims.d.ts
```diff
@@ -9,13 +9,6 @@ import * as _$ from 'jquery';
 // Globals from flarum/core
 import Application from './src/common/Application';
 
-/**
- * Helpers that flarum/core patches into Mithril
- */
-interface m extends Mithril.Static {
-  prop: typeof Stream;
-}
```
franzliedke, line 17: I guess you should do `m: Mithril.Static` in the `declare global` selection further below then?

js/src/common/components/Alert.tsx
franzliedke: Should be tested out, I guess?

franzliedke (changes requested):

js/shims.d.ts
datitisev: Not sure if that works. I had a hard time getting `m` global to properly work... if `m: Mithril.Static` doesn't work we can keep `interface m extends Mithril.Static {}` just fine I think.

js/src/common/components/Alert.tsx
askvortsov1: @datitisev any luck?

datitisev: I have not tested code.

askvortsov1: After some discussion, we've concluded that importing this in core is not an issue, as mithril is used in other places anyway. The shim is used for recognition of the global `m`, but we still need to type stuff that's not part of that global function.

There might be issues with extensions and typescript, but that is unrelated to this PR, and should be revisited later.

datitisev (approved):

PR #2310, opened 2020-09-24
Add in BC layer for props, initProps, m.withAttr, and m.prop

askvortsov1:


js/src/common/Component.ts
```diff
@@ -133,4 +133,31 @@ export default abstract class Component<T extends ComponentAttrs = any> implemen
...
+  // BEGIN DEPRECATED MITHRIL 2 BC LAYER
+
+  /**
+   * Initialize the component's attrs.
+   *
+   * This can be used to assign default values for missing, optional attrs.
+   *
+   * @deprecated, use initAttrs instead.
+   */
+  protected static initProps<T>(attrs: T): void {}
```
franzliedke, line 146: I am missing code that calls this method when `initAttrs()` does not exist. (Depending on whether and how this can be done with TypeScript, this method stub may then not be needed.)

franzliedke (changes requested):

js/src/common/Component.ts
askvortsov1: Yeah, just realized this doesn't actually do anything :facepalm:. Fixed!

franzliedke (approved):

tankerkiller125 (approved):

PR #2313, opened 2020-09-25
Add warnings to Mithril 2 BC layer

askvortsov1:


datitisev (changes requested):
You need to add checks so that you don't have a warning on *every* use of those methods - because with the current changes you WILL crash the console.

The warnings should only appear once.

askvortsov1:
That is a good point, although I suppose it would get the point across that an upgrade is needed :rofl:. I don't really see a good way of doing this outside of creating a global boolean flag for each of these warnings...

datitisev:
Not global. A `let` at the beginning of the file for each warning is completely fine.

franzliedke (approved):

tankerkiller125 (approved):

Issue #2303, opened 2020-09-23
Investigate Modal State Inconsistencies when closing via `this.hide`


askvortsov1:
See https://github.com/flarum/tags/pull/93#discussion_r491291402

franzliedke:
Really, the best way you can clean this up for real is by stopping to use Bootstrap's modal JS (the CSS is nice for styling). Mixing DOM-based state and VDOM stuff is bound to be endlessly confusing.

franzliedke:
To be clear: if the proposed fixes fix the end-user behavior, then that's fine for now. But IMO, cleaning this up will greatly help with maintenance in the future.

askvortsov1:
We're planning to eliminate bootstrap in later releases, we'll either adopt a smaller custom library, or code in custom modal logic (which might make sense given how (relatively) small it is

franzliedke:
The biggest problem in this particular area is the huge and very confusing mismatch between DOM-based and VDOM-based approaches to JS. This is independent from using or not using Bootstrap for styling certain things.

PR #2314, opened 2020-09-25
Mithril 2 modal fixes

askvortsov1:
This PR:

- Reverts the hacky #2263 
- Makes `this.hide` call the global `modal.close`, fixing #2303 
- Uses `onbeforeremove` for closing animation instead of `onremove`, and only when we actually need to close the modal
- Ensures that `readyCallback` is called on modals opened from other modals.

### DIFF
--- a/js/src/common/components/Modal.js
+++ b/js/src/common/components/Modal.js
@@ -27,6 +27,10 @@ export default class Modal extends Component {
     this.attrs.onshow(() => this.onready());
   }
 
+  onremove() {
+    this.attrs.onhide();
+  }
+
   view() {
     if (this.alertAttrs) {
       this.alertAttrs.dismissible = false;
--- a/js/src/common/components/Modal.js
+++ b/js/src/common/components/Modal.js
@@ -31,7 +31,7 @@ export default class Modal extends Component {
     // If the global modal state currently contains a modal,
     // we've just opened up a new one, and accordingly,
     // we don't need to show a hide animation.
-    if (!app.modal.modal) {
+    if (!this.attrs.state.modal) {
       this.attrs.onhide();
     }
   }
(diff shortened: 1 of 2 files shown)


js/src/common/components/ModalManager.js
```diff
@@ -34,6 +35,13 @@ export default class ModalManager extends Component {
   animateShow(readyCallback) {
     const dismissible = !!this.attrs.state.modal.componentClass.isDismissible;
 
+    // If we are opening this modal while another modal is already open,
+    // the shown event will not run, because the modal is already open.
+    // So, we need to manually trigger the readyCallback.
+    if (this.$().hasClass('in')) {
+      readyCallback();
```
franzliedke, line 42: Should this also have an early return?

js/src/common/components/Modal.js
```diff
@@ -103,7 +112,7 @@ export default class Modal extends Component {
    * Hide the modal.
    */
   hide() {
-    this.attrs.onhide();
+    this.attrs.state.close();
```
franzliedke, line 115: The naming and usage of `hide`, `onhide` and `close` is now very irritating, IMO.

js/src/common/components/ModalManager.js
askvortsov1: Good point!

js/src/common/components/Modal.js
askvortsov1: Agree, but not that much we can do. Renamed onhide to animateHide

clarkwinkelmann (approved):

js/src/common/components/Modal.js
clarkwinkelmann: Shouldn't `onshow` also be renamed in that case? It looks inconsistent in `ModalManager` line 17 now.

datitisev (approved):
Seems a bit hacky, but the modal manager was bound to be 🤔

PR #2316, opened 2020-09-25
Put m.stream in flarum/utils/stream

askvortsov1:


clarkwinkelmann:
I'm a bit confused with the capitalization. It seems the import path is lowercase, but the actual method/object is called `Stream` with uppercase in Mithril? https://mithril.js.org/stream.html

If so I think we should use the same capitalization as Mithril for our object. I see this PR already exports it as default with uppercase, but then everywhere the default is imported as lowercase.

Alternatively I wouldn't be against calling it `prop`. But I suppose it's best to keep the naming as consistent as possible with Mithril.

clarkwinkelmann (approved):

tankerkiller125 (approved):

PR #2319, opened 2020-09-27
Abort requests on route change

w-4:
**Fixes https://github.com/flarum/core/issues/2268**

See: https://github.com/flarum/core/pull/2276#issuecomment-698977900

Just a basic implementation. Sorry for the large diff, it's due to prettier formatting (the change is in the last line).

### DIFF
--- a/js/src/common/Application.js
+++ b/js/src/common/Application.js
@@ -160,12 +160,12 @@ export default class Application {
   titleCount = 0;
 
   /**
-   * An object that keeps track of active GET requests.
+   * A Map object that keeps track of active GET requests.
    *
    * @type {Object}
    * @public
    */
-  activeRequests = {};
+  activeRequests = new Map();
 
   load(payload) {
     this.data = payload;
@@ -309,7 +309,7 @@ export default class Application {
     // when the user changes routes while the request is still loading.
     const requestId = +new Date();
     if (options.method === 'GET') {
-      extend(options, 'config', (result, xhr) => (this.activeRequests[requestId] = xhr));
+      extend(options, 'config', (result, xhr) => this.activeRequests.set(requestId, xhr));
     }
 
     // When we deserialize JSON data, if for some reason the server has provided
@@ -427,7 +427,7 @@ export default class Application {
           return Promise.reject(error);
         }
       )
-      .finally(() => delete this.activeRequests[requestId]);
+      .finally(() => this.activeRequests.delete(requestId));
   }
 
   /**
--- a/js/src/common/Application.js
+++ b/js/src/common/Application.js
@@ -305,9 +305,12 @@ export default class Application {
       options.method = 'POST';
     }
 
-    //Save a reference to the request object, that can be used to abort the request when the user changes routes while the request is still loading.
+    // Save a reference to the request object, that can be used to abort the request
+    // when the user changes routes while the request is still loading.
     const requestId = +new Date();
-    if (options.method === 'GET') extend(options, 'config', (result, xhr) => (this.activeRequests[requestId] = xhr));
+    if (options.method === 'GET') {
+      extend(options, 'config', (result, xhr) => (this.activeRequests[requestId] = xhr));
+    }
 
     // When we deserialize JSON data, if for some reason the server has provided
     // a dud response, we don't want the application to crash. We'll show an
--- a/js/src/common/Application.js
+++ b/js/src/common/Application.js
@@ -308,7 +308,7 @@ export default class Application {
     // Save a reference to the request object, that can be used to abort the request
     // when the user changes routes while the request is still loading.
     const requestId = +new Date();
-    if (options.method === 'GET') {
+    if (options.abortOnPageChange) {
       extend(options, 'config', (result, xhr) => this.activeRequests.set(requestId, xhr));
     }
 
--- a/js/src/common/Application.js
+++ b/js/src/common/Application.js
@@ -160,7 +160,7 @@ export default class Application {
   titleCount = 0;
 
   /**
-   * A Map object that keeps track of active GET requests.
+   * A Map object that keeps track of all active requests.
    *
    * @type {Object}
    * @public
@@ -305,12 +305,10 @@ export default class Application {
       options.method = 'POST';
     }
 
-    // Save a reference to the request object, that can be used to abort the request
-    // when the user changes routes while the request is still loading.
-    const requestId = +new Date();
-    if (options.abortOnPageChange) {
-      extend(options, 'config', (result, xhr) => this.activeRequests.set(requestId, xhr));
-    }
+    // Save a reference to the request object, that can be used to later
+    // abort the request.
+    const requestId = options.requestId || +new Date();
+    extend(options, 'config', (result, xhr) => this.activeRequests.set(requestId, xhr));
 
     // When we deserialize JSON data, if for some reason the server has provided
     // a dud response, we don't want the application to crash. We'll show an
--- a/js/src/common/Application.js
+++ b/js/src/common/Application.js
@@ -306,8 +306,8 @@ export default class Application {
       options.method = 'POST';
     }
 
-    // Save a reference to the request object, that can be used to later
-    // abort the request.
+    // Save a reference to the xhr object, that can be used to later
+    // abort the request by calling this.activeRequests.abort(requestId)
     const requestId = options.requestId || +new Date();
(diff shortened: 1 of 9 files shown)


js/src/common/Application.js
```diff
@@ -297,6 +305,10 @@ export default class Application {
       options.method = 'POST';
     }
 
+    //Save a reference to the request object, that can be used to abort the request when the user changes routes while the request is still loading.
```
askvortsov1, line 308: Extremely minor nitpick, but can we add a space before the start of comment text and split the comment over 2 lines?

Also for the conditional below, can we wrap that in braces?

js/src/common/components/Page.js
```diff
@@ -34,6 +34,8 @@ export default class Page extends Component {
    */
   onNewRoute() {
     app.current.set('routeName', this.attrs.routeName);
+    //abort all previous requests to prevent bugs
+    for (let request in app.activeRequests) request.abort();
```
askvortsov1, line 38: When the request is aborted, how does that affect the `.then` handling above? Would that count as an error, or would neither of those 2 blocks run? I wasn't able to find that bit in Mithril documentation

w-4: Tested that neither will run

askvortsov1 (approved):
Thank you very much!

clarkwinkelmann:
I feel like this has a way broader impact than just fixing the referenced issue.

What about extensions that trigger requests that aren't tied to data on a page? Like a chat widget, or an analytics script?

Also what happens to cancelled requests? Is an error message displayed? Are they silently dismissed? How will an extension know its request has been cancelled?

askvortsov1:
Thats a good point. Perhaps it should be an opt-in system? Like an abortOnPageChange flag in request options?

clarkwinkelmann:
If the only use case are requests triggered from `oninit`, then I think this should be either a special flag, or a dedicated request method that proxies to the full request method. Maybe even make it part of `Page` if that's the only context it will be called from?

askvortsov1:
Since the lifecycle of these methods is controlled by `Page`, I like the idea of the `Page` component being responsible for aborting them when irrelevant; that could be called in `onremove` and in `onNewRoute`.

w-4:
> I feel like this has a way broader impact than just fixing the referenced issue.
> 
> What about extensions that trigger requests that aren't tied to data on a page? Like a chat widget, or an analytics script?


Good point. 
All page requests use the `app.store.find` method to request data as far as I can tell. I think it's probably best to abort all of those on route change for core. If not for bugs, also for increased performance on slow networks. 
How about an option `abortOnRouteChange` that is passed to the request method, and that we set to default to true in `app.store.find`?

> 
> Also what happens to cancelled requests? Is an error message displayed? Are they silently dismissed? How will an extension know its request has been cancelled?

Yep, cancelled requests are silently dismissed.

w-4:
How about this guys? ( draft, that's not ready for merge yet )

askvortsov1 (commented):
I like where this is headed, but I think we should plan on beta 15 for this, so that we have enough time to discuss implications in full. The immediate discussion list issue should be fixed by #2322

js/src/common/components/Page.js
```diff
@@ -33,6 +30,9 @@ export default class Page extends Component {
    * adjust the current route name.
    */
   onNewRoute() {
+    app.previous = app.current;
+    app.previous.abortRequests();
+    app.current = new PageState(this.constructor);
```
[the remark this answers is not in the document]
w-4: yes, maybe it can be cherry-pick merged for beta 14.

js/src/common/states/PageState.js
```diff
@@ -30,4 +31,32 @@ export default class PageState {
...
+
+  /**
+   * Aborts all ongoing page GET requests.
+   */
+  abortRequests() {
+    for (let requestId of this.activeRequestIds) {
+      const request = app.activeRequests.get(requestId);
+      if (request) {
+        request.abort();
+        app.activeRequests.delete(requestId);
```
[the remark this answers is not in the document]
w-4: There's now a helper class `ActiveRequests` so it's a little better structured. It's possible to use the abort functionality in other components and extensions by passing a `requestId` in the request options, and calling `app.activeRequest.abort(requestId)` to abort the request if needed.

js/src/common/utils/ActiveRequests.ts
```diff
@@ -0,0 +1,8 @@
+export default class ActiveRequests extends Map {
```
askvortsov1, line 1: Could we have a docblock for this?

js/src/common/Application.js
```diff
@@ -297,6 +306,11 @@ export default class Application {
       options.method = 'POST';
     }
 
+    // Save a reference to the xhr object, that can be used to later
+    // abort the request by calling this.activeRequests.abort(requestId)
+    const requestId = options.requestId || +new Date();
+    extend(options, 'config', (result, xhr) => this.activeRequests.set(requestId, xhr));
```
askvortsov1, line 312: Do we still need these changes?

js/src/common/components/Page.js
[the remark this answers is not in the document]
askvortsov1: Done in https://github.com/flarum/core/pull/2328, couldn't cherry-pick unfortunately as that commit had other stuff in it

clarkwinkelmann:
I like this feature, but I still wonder. How useful is it actually?

Because in most cases, I don't think it will make sense to cancel requests. Once the requests are already being processed, whether they finish or not shouldn't have a huge impact no? If the component is out of scope, the results will be dismissed.

The global discussion list thing really stands out, because it's a global object, but we want to refresh it between some page loads.

Most of the other situations that have been handled here won't actually benefit much from the change I believe? Only advantage I can see is for very large requests to be cancelled and save up on bandwidth. But that would only apply to very rare situations, and would only benefit very low resources devices.

js/src/common/Application.js
w-4: Yes, this part is needed to save a reference to the XHR object.

w-4:
> I like this feature, but I still wonder. How useful is it actually?
> 
> Because in most cases, I don't think it will make sense to cancel requests. Once the requests are already being processed, whether they finish or not shouldn't have a huge impact no? If the component is out of scope, the results will be dismissed.
> 
> The global discussion list thing really stands out, because it's a global object, but we want to refresh it between some page loads.
> 
> Most of the other situations that have been handled here won't actually benefit much from the change I believe? Only advantage I can see is for very large requests to be cancelled and save up on bandwidth. But that would only apply to very rare situations, and would only benefit very low resources devices.

Correct. The only real noticeable change is in the discussion list, also because it's a common behavior to navigate away while it's loading - f.e. when you want to go from the index route straight to a child tag.

askvortsov1:
As much as I like the general idea, I'm not sure this is necessary anymore, now that the discussion page issue has been fixed. I think we should close this for now, and reopen it if a use case more clearly emerges.

tankerkiller125 (approved):

tankerkiller125 (changes requested):
After looking deeper, I think I might agree with Sasha that these changes might not be needed anymore. Plus there are a bunch of conflicts that need to be rebased before this could be merged.

js/src/common/states/PageState.js
```diff
@@ -30,4 +31,32 @@ export default class PageState {
   set(key, value) {
     this.data[key] = value;
   }
+
+  /**
+   * A wrapper around the Store find method.
+   *
+   */
+  findInStore(type, id, query = {}, options = {}) {
```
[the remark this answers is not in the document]
davwheat: Maybe call it `findPendingXhrRequestsInStore`, unless that's a bit too long?

js/src/common/Application.js
```diff
@@ -159,6 +160,14 @@ export default class Application {
   title = '';
   titleCount = 0;
 
+  /**
+   * A Map object that keeps track of all active requests.
+   *
+   * @type {Object}
```
davwheat, line 166:
```suggestion
```

davwheat:
To be honest, I do like this implementation (just by looking at it on my phone). It seems well thought through and put together, as well as efficient.

We would need to recreate these changes or rebase onto master, but it does look like a good implementation to me.

askvortsov1:
This would probably go better on the discussion, but one of my concerns is that implementing this will mask incorrect / unideal implementations elsewhere (as with the problem that this PR was opened to solve).

PR #2321, opened 2020-09-28
Improve developer experience by forcing LF line endings

w-4:
Force LF line endings when checking out files. The default on git for windows is to checkout with CRLF line endings and to commit with LF line endings. Because of the husky prettier hook, after each commit it reformats all files, showing up as changes.


askvortsov1 (commented):
Makes sense to me! Although I suppose we'd probably want an ending newline. That's about as minor a nitpick as possible though :D

askvortsov1 (approved):

tankerkiller125 (approved):

PR #2322, opened 2020-09-28
Improve DiscussionListState refresh method

w-4:
See: https://github.com/flarum/core/pull/2276#issuecomment-690859579
and https://github.com/flarum/core/pull/2276#discussion_r486778011

**Confirmed**

### DIFF
--- a/js/src/forum/states/DiscussionListState.js
+++ b/js/src/forum/states/DiscussionListState.js
@@ -77,14 +77,14 @@ export default class DiscussionListState {
   }
 
   /**
-   * Clear and reload the discussion list. Passing the option { clear: false } will clear
-   * discussions only after new data has been received. This can be used to refresh
-   * discussions without loading animations.
+   * Clear and reload the discussion list. Passing the option `deferClear: true`
+   * will clear discussions only after new data has been received.
+   * This can be used to refresh discussions without loading animations.
    */
-  refresh({ clear = true } = {}) {
+  refresh({ deferClear = false } = {}) {
     this.loading = true;
 
-    if (clear) {
+    if (!deferClear) {
       this.clear();
     }
 
--- a/js/src/forum/states/DiscussionListState.js
+++ b/js/src/forum/states/DiscussionListState.js
@@ -90,7 +90,7 @@ export default class DiscussionListState {
 
     return this.loadResults().then(
       (results) => {
-        // We don't use the clear method, since it would lead to an additional redraw.
+        // Same effect as this.clear(), but without an additional redraw.
         this.discussions = [];
         this.parseResults(results);
       },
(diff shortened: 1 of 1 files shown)


askvortsov1:
So effectively, this adds in that option you were talking about where we can clear without the animations? (or I suppose an alternate way of looking it would be that it actually makes clear = false useful). I like it!

askvortsov1 (approved):

askvortsov1:
One quick thought: maybe it would be better to use `animate` or something like that instead of `clear`? Since we're always clearing under this system

askvortsov1 (changes requested):
Now that I look at it again, let's rename the "clear" method to "animate" or "redraw" or something, as the discussion list will ALWAYS be cleared

askvortsov1 (approved):

clarkwinkelmann:
Is/was the `clear` parameter used anywhere in core or community extensions? To know whether any other change is needed and whether we need backward compatibility for this.

askvortsov1:
Nope, it was introduced during the stream extraction, but left over as a useless side product. This PR makes it useful (in addition to fixing the referenced bug)

js/src/forum/states/DiscussionListState.js
```diff
@@ -77,17 +77,21 @@ export default class DiscussionListState {
...
     this.loading = true;
 
-    if (clear) {
+    if (!deferClear) {
       this.clear();
     }
 
     return this.loadResults().then(
       (results) => {
+        // We don't use the clear method, since it would lead to an additional redraw.
```
clarkwinkelmann, line 93: Just one last suggestion, I'd make this comment say something like "Same effect as this.clear(), but without an additional redraw"

I find the current phrasing a bit confusing.

Otherwise, nothing else to add.

clarkwinkelmann (changes requested):

js/src/forum/states/DiscussionListState.js
```diff
@@ -77,17 +77,23 @@ export default class DiscussionListState {
...
-    if (clear) {
+    if (!deferClear) {
       this.clear();
     }
 
     return this.loadResults().then(
       (results) => {
+        // Same effect as this.clear(), but without an additional redraw.
+        // This ensures that any changes made while waiting on this request
+        // are ignored. Otherwise, we could get duplicate discussions.
```
clarkwinkelmann, line 95: I'm unsure about that "duplication" statement. This comment is inside of the promise callback, after everything has been done. There's no waiting at this point, no?

The reason to skip a redraw is just for performance and to maybe prevent a flicker?

clarkwinkelmann (changes requested):

js/src/forum/states/DiscussionListState.js
askvortsov1: Well, this is the part that actually fixes the issue where changes to the discussion list while the refresh request is loading result in duplicate things.

It's explaining why we clear it, not why we clear it without `this.clear`. I'll reword this.

clarkwinkelmann (approved):
Actually, now that I re-read it, I'm not sure why I found it so confusing in my last comment.

All good!

PR #2323, opened 2020-09-28
Mark keys for Config required only with InstalledSite

luceos:
Fixes an issue with the $_SERVER object having no value in cli, this is used in the cli install command.

### DIFF
--- a/src/Foundation/Site.php
+++ b/src/Foundation/Site.php
@@ -25,6 +25,8 @@ class Site
         date_default_timezone_set('UTC');
 
         if (! static::hasConfigFile($paths->base)) {
+            // Instantiate site instance for new installations,
+            // fallback to localhost for validation of Config for instance in CLI.
             return new UninstalledSite(
                 $paths,
                 Arr::get($_SERVER, 'REQUEST_URI', 'http://localhost')
--- a/src/Foundation/Site.php
+++ b/src/Foundation/Site.php
@@ -25,7 +25,10 @@ class Site
         date_default_timezone_set('UTC');
 
         if (! static::hasConfigFile($paths->base)) {
-            return new UninstalledSite($paths, $_SERVER['REQUEST_URI']);
+            return new UninstalledSite(
+                $paths,
+                Arr::get($_SERVER, 'REQUEST_URI', 'http://localhost')
+            );
         }
 
         return (


askvortsov1:
Unit tests need to be adjusted to explicitly require the url key in the test that checks for it.

src/Foundation/Config.php
```diff
@@ -42,7 +40,7 @@ public function inMaintenanceMode(): bool
         return $this->data['offline'] ?? false;
     }
 
-    private function requireKeys(...$keys)
+    public function requireKeys(...$keys)
```
askvortsov1, line 43: Not sure how I feel about mutating the singleton... I would prefer that we pass in required keys to the constructor, or make this method return a clone of the config object

src/Foundation/Site.php
```diff
@@ -25,7 +25,10 @@ public static function fromPaths(array $paths)
         date_default_timezone_set('UTC');
 
         if (! static::hasConfigFile($paths->base)) {
-            return new UninstalledSite($paths, $_SERVER['REQUEST_URI']);
+            return new UninstalledSite(
+                $paths,
+                Arr::get($_SERVER, 'REQUEST_URI')
```
askvortsov1, line 30: Is this change necessary for this PR?

askvortsov1 (changes requested):

src/Foundation/Site.php
luceos: Without this change the CLI still fails, because the original error is that $_SERVER does not contain that key:

```
PHP Notice:  Undefined index: REQUEST_URI in /home/forge/nightly.flarum.site/nightly/vendor/flarum/core/src/Foundation/Site.php on line 28
```

This is **the** fix. But because in the code the URL cannot be NULL we needed deeper changes.

src/Foundation/Config.php
luceos: Agreed, will update and fix tests in one swoop.

luceos: This won't work. The construct of this class is fired from the site.php which is too easily accessible from the outside. Either we drop outside mutation of the required keys from outside (demanding them based on a flag perhaps $installed as argument) or we use the above change.

src/Foundation/Site.php
askvortsov1: Makes sense. Let's keep that then!

src/Foundation/Config.php
askvortsov1: In this case, let's temporarily remove requiring _any_ keys, and revisit validation later? The previous config object didn't have external validation, so adding it now isn't necessary (plus, the choice of 'url' as the main attribute to enforce needs more discussion in retrospect)

luceos: Patched tests, removed method.

src/Foundation/Site.php
franzliedke, line 30: What do we need the value for at runtime for the installer? Since this is for the `UninstalledSite`, we could just provide a bogus value when `$_SERVER['REQUEST_URI']` isn't set?

I'd find that cleaner in this case.

franzliedke: When reviewing the old code, it looks like the old version must have had `url` set to `null` in the CLI, so a bogus value should work fine and be less invasive?

askvortsov1: Hmm, generally I'm not the biggest fan of bogus values, but in this case, we could just put in the nightly site's literal url I suppose? And since we'd expect a url to be provided on CLI install, I suppose we wouldn't run into this issue for actual usage?

franzliedke: I would literally put a bogus value there - something that signals that this is intentionally bogus, because irrelevant. `doesnotmatter.tld` or `bogus.url` maybe?

franzliedke: A comment can help explain the choice, of course.

askvortsov1: Wait, in the core source code, or in the nightly site config?

franzliedke: Core source code. When we pass a base URL to the `UninstalledSite` in `Site.php`.

askvortsov1: Ah alright I get what you're getting at.

Small followup that I should have thought of earlier: is there a reason why we require `url`, and only `url`?

franzliedke: Good question. Haven't thought that true, I guess. :wink:

franzliedke: Or actually... it's the one that's used far more than any of the other ones, but obviously the other ones must be used somewhere, too.

askvortsov1: Tbh, that's mostly why I preferred being able to pass in required keys per-site, or maybe even a flag. How do you feel about the mutation-based approach that @luceos  had in the original PR?

franzliedke: Meh. At that point, we could easily do two classes again.

src/Foundation/Site.php
```diff
@@ -25,7 +25,10 @@ public static function fromPaths(array $paths)
         date_default_timezone_set('UTC');
 
         if (! static::hasConfigFile($paths->base)) {
-            return new UninstalledSite($paths, $_SERVER['REQUEST_URI']);
+            return new UninstalledSite(
+                $paths,
+                Arr::get($_SERVER, 'REQUEST_URI', 'http://localhost')
```
askvortsov1, line 30: I like `bogus.url` a bit more as it's clearer that this doesn't actually mean anything. But either way, we'd want a comment explaining that this is only needed for validation at this stage, and doesn't actually do anything

askvortsov1 (changes requested):

askvortsov1 (approved):

tankerkiller125 (approved):

PR #2324, opened 2020-09-28
Fix mobile controls gesture on discussion deletion/restoration

### LINKED ISSUE #1427: Mobile slide gesture breaks on Delete
On the mobile view of Flarum, swiping left on a discussion will open the discussion options menu. Choosing delete, and then performing the swipe left option again will hide the discussion from view (the display block remains, though empty) and fail to bring up the discussion options again. 

I haven't experienced this with any option other than delete.

SychO9:
**Fixes #1427**

Because the `Slidable` class was always added on creation, it was lost every time the class list changed (in this case when the discussion was hidden/unhidden which added/removed `DiscussionListItem--hidden` class). So by determining the `Slidable` class's presence in `elementAttrs()` method, it guarantees it always properly set. 

..

**Confirmed**


askvortsov1 (approved):
Makes sense to me, thank you!

datitisev (approved):

PR #2328, opened 2020-09-30
Minor improvements to onNewRoute

askvortsov1:
In this PR:

- We move calls to create/modify `app.previous` and `app.current` from `oninit` to `onNewRoute` (inspired by https://github.com/flarum/core/pull/2319#discussion_r497053049)
- We call onNewRoute in `onbeforeupdate` for DiscussionPage and UserPage as well as in IndexPage (so that several visits to DiscussionPage in a row don't count as one "PageState".

This also sets us up for possible changes like https://github.com/flarum/core/pull/2275#issuecomment-701032151

### DIFF
--- a/js/src/common/components/Page.js
+++ b/js/src/common/components/Page.js
@@ -31,9 +31,7 @@ export default class Page extends Component {
    */
   onNewRoute() {
     app.previous = app.current;
-    app.current = new PageState(this.constructor);
-
-    app.current.set('routeName', this.attrs.routeName);
+    app.current = new PageState(this.constructor, { routeName: this.attrs.routeName });
   }
 
   oncreate(vnode) {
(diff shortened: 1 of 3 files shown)


clarkwinkelmann:
On its own, does this diff introduce any change of behavior?

Does it change anything in regard to the navigation history?

Does it make sense to call `app.current.set('routeName', this.attrs.routeName);` if it's called everytime a new `PageState` is created? Shouldn't this be passed as a constructor argument then?

askvortsov1:
Not to navigation history, but if you go from one discussions page to another discussions page, app.current and app.previous will more accurately reflect that, instead of acting like we never left the page.

And yeah, those can be condensed into one argument. Will do.

tankerkiller125 (approved):

clarkwinkelmann (approved):

PR #2351, opened 2020-10-03
Drop support for unused utils

askvortsov1:
The `mixin` and `evented` utils aren't used, and removing them has been proposed since https://github.com/flarum/core/issues/872#issuecomment-198682653.


clarkwinkelmann:
:warning: :warning: :warning: `mixin` is used by all extensions implementing models but not using the new `Model` JS extenders, for example https://github.com/FriendsOfFlarum/terms/blob/master/js/src/common/models/Policy.js#L5

41 extensions total are using it in their latest release (though some are abandoned) https://query.flarum.dev/q/1bee4cae-d79a-44e2-adc8-5ce60579c2f2

I'm unable to find any extension using `evented` though.

How about deprecating them and adding warnings, like we did with other methods ? We can then remove them one version later once we're certain.

datitisev:
>  implementing models but not using the new Model JS extenders, 

The Model JS extenders have nothing to do with the `mixin` use. `mixin` is being used to extend the Model class - instead of directly extending the modal class like in core.

The Model JS extenders are used to extend existing models, I'm pretty sure.

**Also**, flarum/tags uses it - https://github.com/flarum/tags/blob/6d79900aed44f26e827a679d4819f82285d09ffd/js/src/common/models/Tag.js#L5

clarkwinkelmann:
> The Model JS extenders have nothing to do with the mixin use

Oops, my bad. So, we can't really deprecate/remove it, unless we decide on another recommended solution.

> Also, flarum/tags uses it

Yeah, I noticed that as well in the output from Query :sweat_smile:

askvortsov1:
Why couldn't extensions just extend the Model class?

clarkwinkelmann:
There might be some historical reason, maybe something related to or predating the ES6 classes?

I am able to successfully replace

```js
export default class MyModel extends mixin(Model, {
    title: Model.attribute('title'),
    isHidden: Model.attribute('isHidden'),
    hiddenAt: Model.attribute('hiddenAt', Model.transformDate),

    user: Model.hasOne('user'),

    badges() {
        const items = new ItemList();

        if (this.isHidden()) {
            items.add('hidden', Badge.component({
                type: 'hidden',
                icon: 'fas fa-trash',
                label: app.translator.trans('my-translation'),
            }));
        }

        return items;
    },
}) {
    apiEndpoint() {
        return '/custom-endpoint' + (this.exists ? '/' + this.data.id : '');
    }
}
```

with

```js
export default class MyModel extends Model {
    title = Model.attribute('title');
    isHidden = Model.attribute('isHidden');
    hiddenAt = Model.attribute('hiddenAt', Model.transformDate);

    user = Model.hasOne('user');

    badges() {
        const items = new ItemList();

        if (this.isHidden()) {
            items.add('hidden', Badge.component({
                type: 'hidden',
                icon: 'fas fa-trash',
                label: app.translator.trans('my-translation'),
            }));
        }

        return items;
    }

    apiEndpoint() {
        return '/custom-endpoint' + (this.exists ? '/' + this.data.id : '');
    }
}
```

And my extension still compiles and runs successfully. So maybe it really has no advantage over ES6 extend?

We could start by changing it in core extensions and see how it goes on master. Then decide if we want to deprecate it in the next beta.

datitisev:
Yes, it can be replaced with a class extending Model just fine. That's what I did in my rewrite branch.

datitisev:
I don't know how I managed to close this. My bad.

clarkwinkelmann:
Uh it's another of those PR-discussions.

I think we have mostly agreed to not remove them anytime soon, so should we close the PR?

We can open another issue to discuss whether we want to keep them, maybe even one issue per method.

The method that's completely unused, I'd be in favor to deprecate for next beta.

`mixin` we should probably keep for now and discuss the remaining use cases. I've been removing it from every extension I work on so the number of usages must be going down since this PR was created.

I remember discussing elsewhere what there was a difference regarding the object `prototype` when using `mixin` vs something else, but I cannot find it anymore and I don't remember if it was compared to ES6 `extend` or something else.

PR #2352, opened 2020-10-03
Remove depreciated code in QueueServiceProvider

smartclash:
Apparently the `Configuring` event was depreciated and eventually removed. But traces of it's references were not. This PR aims to removes them.

### DIFF
--- a/src/Queue/QueueServiceProvider.php
+++ b/src/Queue/QueueServiceProvider.php
@@ -63,8 +63,7 @@ class QueueServiceProvider extends AbstractServiceProvider
             return new Worker(
                 new HackyManagerForWorker($app[Factory::class]),
                 $app['events'],
-                $app[ExceptionHandling::class],
-                function () { return false; }
+                $app[ExceptionHandling::class]
             );
         });
 
--- a/src/Queue/QueueServiceProvider.php
+++ b/src/Queue/QueueServiceProvider.php
@@ -63,7 +63,8 @@ class QueueServiceProvider extends AbstractServiceProvider
             return new Worker(
                 new HackyManagerForWorker($app[Factory::class]),
                 $app['events'],
-                $app[ExceptionHandling::class]
+                $app[ExceptionHandling::class],
+                function () { return false; }
             );
         });
 
--- a/src/Queue/QueueServiceProvider.php
+++ b/src/Queue/QueueServiceProvider.php
@@ -103,6 +103,22 @@ class QueueServiceProvider extends AbstractServiceProvider
         $this->app->alias(Factory::class, 'queue');
         $this->app->alias(Worker::class, 'queue.worker');
         $this->app->alias(Listener::class, 'queue.listener');
+
+        $this->registerCommands();
+    }
+
+    private function registerCommands()
+    {
+        $queue = $this->app->make(Queue::class);
+
+        // There is no need to have the queue commands when using the sync driver.
+        if ($queue instanceof SyncQueue) {
+            return;
+        }
+
+        $this->app->extend('flarum.console.commands', function ($existingCommands) {
+            return array_merge($existingCommands, $this->commands);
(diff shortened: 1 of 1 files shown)


src/Queue/QueueServiceProvider.php
```diff
@@ -64,7 +63,8 @@ public function register()
             return new Worker(
                 new HackyManagerForWorker($app[Factory::class]),
                 $app['events'],
-                $app[ExceptionHandling::class]
+                $app[ExceptionHandling::class],
+                function () { return false; }
```
askvortsov1, line 67: What is this parameter for? Would be nice to see a comment at least

src/Queue/QueueServiceProvider.php
```diff
@@ -104,24 +104,6 @@ public function __call($name, $arguments)
         $this->app->alias(Factory::class, 'queue');
         $this->app->alias(Worker::class, 'queue.worker');
         $this->app->alias(Listener::class, 'queue.listener');
-
-        $this->registerCommands();
-    }
-
-    protected function registerCommands()
```
askvortsov1, line 111: Does this need to be replaced with something? @luceos you're a bit more familiar with the queue system than I am

askvortsov1 (changes requested):

src/Queue/QueueServiceProvider.php
datitisev: Yes, the commands need to be registered still.

datitisev:
@tankerkiller125 mentioned on Discord that this would be preferred as 2 separate PRs. That would probably be better.

Just mentioning this as there's already feedback submitted in the form of review.

smartclash:
I'll split this PR into two separate parts - first thing when I get back to my PC tomorrow ;)

src/Queue/QueueServiceProvider.php
smartclash: Made it as a separate PR, and I've also commented what that does. https://github.com/flarum/core/pull/2357

smartclash:
This PR is now split into 2 parts. 

#2357 now adds the missing parameter. I've also made the changes requested in that PR for that specific problem.

src/Queue/QueueServiceProvider.php
smartclash: I've "tried" registering the commands. Let me know if the approach is okay or not :)

src/Queue/QueueServiceProvider.php
```diff
@@ -108,19 +107,17 @@ public function __call($name, $arguments)
...
+        $queue = $this->app->make(Queue::class);
 
-            // There is no need to have the queue commands when using the sync driver.
-            if ($queue instanceof SyncQueue) {
-                return;
-            }
+        // There is no need to have the queue commands when using the sync driver.
+        if ($queue instanceof SyncQueue) {
+            return;
+        }
```
luceos, line 117: Can you reset the changes you made please so that the PR only reflects the changes. Otherwise we will merge changes that do nothing but rewrite the same code, thank you.

luceos:
@smartclash Karan sorry to be such a bother, I was looking at the Queue implementation and noticed there was more work needed to patch it up for the upcoming release. Especially while running the console commands with the NULL driver I found several additional issues to take care of.

- For instance we've upgrade our Illuminate dependency which enables us to remove the HackyQueueManager previously added. 
- The WorkCommand failed because it was unable to resolve the maintenance flag from the Container as we've moved that to its dedicated class Config.
- The Worker needed an additional argument in relation to that same maintenance flag.

I pivoted this work into its' own branch which pretty much obsoletes your work. I'm extremely sorry to void your work on this, although I did adopt your method of extending the commands binding.

smartclash:
Sure @luceos, no worries. I am happy to help this project, always :)

PR #2353, opened 2020-10-03
Converted Utils to Typescript

visibleajay:
Converted some utils files from js to ts. Issue #2353 

**Fixes #0000**
Converted few independent utils js files to ts.

Change few independent js utils files to ts.

have ts-ignore at one place, if possible would like to know the specific type to make it work. 
Have also used a generic type Object for routes in `mapRoutes.ts`, would like to make it a specific type.

**Screenshot**

**Confirmed**

**Required changes:**


js/src/common/utils/extractText.ts
```diff
@@ -1,15 +1,23 @@
 /**
  * Extract the text nodes from a virtual element.
  *
- * @param {VirtualElement} vdom
- * @return {String}
+ * @param {IVirtualElement} vdom
+ * @return {string}
  */
-export default function extractText(vdom) {
+export default function extractText(vdom: IVirtualElement): string {
```
askvortsov1, line 7: We want to use Mithril types for vdom, I believe we're using those from https://github.com/MithrilJS/mithril.d.ts

js/src/common/utils/extractText.ts
```diff
@@ -1,15 +1,23 @@
...
   if (vdom instanceof Array) {
     return vdom.map((element) => extractText(element)).join('');
   } else if (typeof vdom === 'object' && vdom !== null) {
+    // @ts-ignore
     return vdom.children ? extractText(vdom.children) : vdom.text;
```
askvortsov1, line 12: Would be preferable to avoid ts-ignore, I believe using mithril typing would give us consistency and fix the issue

js/src/common/utils/mapRoutes.ts
```diff
@@ -7,7 +7,7 @@
  * @param {String} [basePath]
  * @return {Object}
  */
-export default function mapRoutes(routes, basePath = '') {
+export default function mapRoutes(routes: Object, basePath: string = ''): Object {
```
askvortsov1, line 10: For basePath, doesn't it automatically detect that it's a string based on the default type?

js/src/common/utils/mapRoutes.ts
```diff
@@ -7,7 +7,7 @@
  * @param {String} [basePath]
  * @return {Object}
  */
-export default function mapRoutes(routes, basePath = '') {
```
askvortsov1, line 10: We'll want to define a custom interface for the routes input. For the output, Mithril might have something, but if not we'll need something custom there too.

askvortsov1 (changes requested):
Thanks for the PR, great work! I left a few comments, mostly about the need to use mithril typing

js/src/common/utils/Drawer.ts
```diff
@@ -21,7 +23,7 @@ export default class Drawer {
    * @return {Boolean}
    * @public
    */
-  isOpen() {
+  isOpen(): Boolean {
```
datitisev, line 26:
```suggestion
  isOpen(): boolean {
```

js/src/common/utils/classList.ts
```diff
@@ -7,10 +7,10 @@
  * // "foo qux"
  *
  * @param {Object} classes
- * @return {String}
+ * @return {string}
  */
-export default function classList(classes) {
-  let classNames;
+export default function classList(classes: Object): string {
```
datitisev, line 12: This can be replaced with the following (not tested, off the top of my head):
```ts
function classList(classes: { [ name: string]: boolean }) {
```

js/src/common/utils/extractText.ts
datitisev: What error exactly is being ignored here with `ts-ignore`?

js/src/common/utils/mapRoutes.ts
datitisev: Routes here can contain typed properties, see `{ path: string, component: Component }` or something similar.

js/src/common/utils/extractText.ts
visibleajay: On using `mithril.VnodeDOM` receiving `Type 'string' is not assignable to type 'VnodeDOM<{}, {}>'.ts(2345)` for the expression, `extractText(vdom.children)`.

js/src/common/utils/mapRoutes.ts
visibleajay: `mithril.Route` as type for routes and `Partial<mithril.Route>` as return type.

js/src/common/utils/mapRoutes.ts
visibleajay: Using `mithril.Route` as type for routes and `Partial<mithril.Route>` as return type.

js/src/common/utils/mapRoutes.ts
datitisev: Does `Mithril.Route` work here? I'm not sure what those typings are exactly 🤔 as we don't follow anything specific to Mithril for `mapRoutes` arguments, I'm unsure here.

visibleajay: No ts error is thrown on using the same type.

js/src/common/utils/classList.ts
```diff
@@ -24,3 +24,7 @@ export default function classList(classes) {
 
   return classNames.join(' ');
 }
+
+export interface IClassParam {
```
askvortsov1, line 28: This might be better plural

js/src/common/utils/extractText.ts
```diff
@@ -1,15 +1,20 @@
...
 /**
  * Extract the text nodes from a virtual element.
  *
- * @param {VirtualElement} vdom
- * @return {String}
+ * @param {mithril.VnodeDOM} vdom
+ * @return {IExtractTextReturn}
  */
-export default function extractText(vdom) {
+export default function extractText(vdom: mithril.VnodeDOM): IExtractTextReturn {
```
askvortsov1, line 9: I feel like `string` makes sense for a return type. Maybe we should cast?

js/src/common/utils/mapRoutes.ts
askvortsov1: The input isn't mithril.Route: mithril's Route maps route to component; our route abstraction maps route names to an object containing route paths and the component (and possibly route resolvers in the future, but under their own keys). A custom interface would be preferable for that input.

Also, I believe Mithril uses mithril.RouteDefs, not mithril.Route, for the actual route definitions

askvortsov1 (changes requested):
Good progress, left a few more comments. Also, I believe we're currently importing Mithril as uppercase (Mithril, not mithril) in existing TS.

PR #2357, opened 2020-10-04
Missing parameter for Laravel's worker in QueueServiceProvider

smartclash:
Laravel's Worker class requires a callback function which returns a boolean. It helps to determine if the application is in maintenance mode or not.

This PR by default passes a falsey callback. But in future, if Flarum implements maintenance mode support, the function can be made to use.


datitisev:
The command code shouldn't be removed in this PR.

smartclash:
> The command code shouldn't be removed in this PR.
@datitisev Oops, fixed that. Thanks :)

luceos:
Same reason as the other one. Thank you for this though.

PR #2359, opened 2020-10-05
Abstract away onNewRoute and routeHasChanged

askvortsov1:
This provides a cleaner interface for dealing with pages that handle multiple routes. We also move `modal` and `drawer` close calls to ensure those happen on route change with the same component.

### DIFF
--- a/js/src/common/components/Page.js
+++ b/js/src/common/components/Page.js
@@ -12,15 +12,18 @@ export default class Page extends Component {
 
     this.onNewRoute();
 
-    app.drawer.hide();
-    app.modal.close();
-
     /**
      * A class name to apply to the body while the route is active.
      *
      * @type {String}
      */
     this.bodyClass = '';
+
+    this.currentPath = m.route.get();
+  }
+
+  routeHasChanged() {
+    return this.currentPath !== m.route.get();
   }
 
   /**
@@ -30,8 +33,21 @@ export default class Page extends Component {
    * adjust the current route name.
    */
   onNewRoute() {
+    this.currentPath = m.route.get();
+
     app.previous = app.current;
     app.current = new PageState(this.constructor, { routeName: this.attrs.routeName });
+
+    app.drawer.hide();
+    app.modal.close();
+  }
+
+  onbeforeupdate(vnode, old) {
+    super.onbeforeupdate(vnode, old);
+
+    if (this.routeHasChanged()) {
+      this.onNewRoute();
+    }
   }
 
   oncreate(vnode) {
--- a/js/src/forum/components/DiscussionPage.js
+++ b/js/src/forum/components/DiscussionPage.js
@@ -32,23 +32,9 @@ export default class DiscussionPage extends Page {
      */
     this.near = m.route.param('near') || 0;
 
-    this.load();
-
-    // If the discussion list has been loaded, then we'll enable the pane (and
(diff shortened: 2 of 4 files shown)


js/src/common/components/Page.js
```diff
@@ -12,15 +12,18 @@ export default class Page extends Component {
...
      * A class name to apply to the body while the route is active.
      *
      * @type {String}
      */
     this.bodyClass = '';
+
+    this.currentPath = m.route.get();
+  }
+
+  routeHasChanged() {
```
clarkwinkelmann, line 25: Do we want this to be public API, or could we just do the test in `onbeforeupdate` ?

clarkwinkelmann (commented):
Seeing this isn't needed for beta 14, I'll do a full review of this later on.

js/src/common/components/Page.js
askvortsov1: I think public API is good: subclasses with more complex use cases (like DiscussionPage) could use it to do more of those checks, and also possibly handle some side effects if the route hasn't changed but we still want to do something (like scrolling to a post).

js/src/common/components/Page.js
```diff
@@ -12,15 +12,18 @@ export default class Page extends Component {
...
-    app.modal.close();
-
     /**
      * A class name to apply to the body while the route is active.
      *
      * @type {String}
      */
     this.bodyClass = '';
+
+    this.currentPath = m.route.get();
```
w-4, line 22: This line can be probably removed I think, since it's called in `onNewRoute` already.

js/src/common/components/Page.js
```diff
@@ -12,15 +12,18 @@ export default class Page extends Component {
 
     this.onNewRoute();
```
w-4, line 13: I'd move that to the end of the function. This way everything in `oninit` would run before `onNewRoute`.

js/src/forum/components/DiscussionPage.js
```diff
@@ -95,30 +81,32 @@ export default class DiscussionPage extends Page {
...
-    if (m.route.get() !== this.prevRoute) {
-      this.onNewRoute();
-      this.prevRoute = m.route.get();
+  onNewRoute() {
+    // If we have routed to the same discussion as we were viewing previously,
+    // prompt the post stream to jump to the new 'near' param.
+    // Otherwise, load in a new discussion. The `else` branch will
+    // be followed when `onNewRoute` is called from `oninit`.
+    const idParam = m.route.param('id');
+    if (this.discussion && idParam && idParam.split('-')[0] === this.discussion.id()) {
```
w-4, line 90: I'd split this out in this PR already, since the diff is very hard to read. Something like this:

```js 

isSameRoute() {
    const idParam = m.route.param('id');
    return this.discussion && idParam && idParam.split('-')[0] === this.discussion.id();
}

onSameRoute() {
    const near = m.route.param('near') || '1';

    if (near !== String(this.near)) {
      this.stream.goToNumber(near);
    }

    this.near = near;
}

```

and:
```js

onNewRoute() {
   if (this.isSameRoute()) return this.onSameRoute();
   
    super.onNewRoute();

    ....

}

```

askvortsov1:
Obsolete with https://github.com/flarum/core/pull/2378

PR #2363, opened 2020-10-05
Fixes the queue for beta 14

luceos:
A few changes were necessary to get the queue working again:

- New argument for maintenance flag in the Worker
- Overriding the WorkerCommand because it has a `Container` instance but assumes that is also a `Application` instance as it tries to call the `isDownForMaintenance` method. I'm resolving our Config.
- Removed the HackishQueueManager class.

---

Someone please test drive this locally as well:

`app/Extend/Queue.php`
```php
<?php

namespace App\Extend;

use Flarum\Extend\ExtenderInterface;
use Flarum\Extension\Extension;
use Illuminate\Contracts\Container\Container;
use Illuminate\Queue\NullQueue;

class Queue implements ExtenderInterface
{
    {
        $container->extend('flarum.queue.connection', function () use ($container) {
            $queue = new NullQueue();
            $queue->setContainer($container);

            return $queue;
        });
    }
}
```
change in your `composer.json`
```json
    "autoload": {
        "psr-4": {
            "App\\": "app/"
        }
    },
```
`composer dumpautoload` to enable the new autoload directive.

This will re-enable the console commands for testing if you use this extender in the `extend.php` 👍 
```php

return [
    new App\Extend\Queue()
];
```


src/Queue/Console/WorkCommand.php
```diff
@@ -0,0 +1,27 @@
...
+class WorkCommand extends \Illuminate\Queue\Console\WorkCommand
+{
+    protected function downForMaintenance()
+    {
+        if ($this->option('force')) {
+            return false;
+        }
+
+        /** @var Config $config */
+        $config = $this->laravel->make(Config::class);
```
askvortsov1, line 23: Can we inject this?

src/Queue/QueueServiceProvider.php
```diff
@@ -61,10 +62,16 @@ public function register()
         });
 
         $this->app->singleton(Worker::class, function ($app) {
+            /** @var Config $config */
+            $config = $app->make(Config::class);
+
             return new Worker(
-                new HackyManagerForWorker($app[Factory::class]),
```
askvortsov1, line 65: Is there a reason why we can use this now but couldn't before?

askvortsov1 (changes requested):

src/Queue/Console/WorkCommand.php
luceos: Yes but this is the least intrusive option available.

src/Queue/QueueServiceProvider.php
luceos: Yes. We weren't able to override something. Not behind a computer but that's the bottom line.

tankerkiller125 (approved):

askvortsov1 (approved):
Alright, makes sense to me! Does this still need to be tested locally?

datitisev:
Did anyone else test this locally? @luceos wanted someone else to test it before merging.

askvortsov1:
yep!

PR #2364, opened 2020-10-05
Refactor PostStream animations

### LINKED ISSUE #2348: Mentioned Post doesn't "flash" while hovering over mention after refresh
## Bug Report

**Current Behavior**
The mentioned post doesn't flash when you hover over the mention after refreshing the page.

**Expected Behavior**
Mentioned post should flash like it does in beta.13

**Environment**
- Flarum version: dev-master (beta.14)
- Browser: [e.g. chrome 67, safari 11] Chromium, Firefox

askvortsov1:
Fixes #2348 

- If the fadeIn animation is specified on the PostStream class itself, any time we add/remove another animation, it will redo fadeIn. To avoid this, we move fadeIn into it's own css class, which is applied, and then immediately removed after the animation is completed to ensure it only runs once.
- The "fix" for flashItem (#2300) was actually broken, as it resulted in '.flash' never being removed, so we never went back to .PostStream's fadeIn. We adjust flashItem to ensure that '.flash' is removed. We also remove 'fadeIn' in case it hasn't yet been removed in oncreate.

### DIFF
--- a/js/src/forum/components/PostStream.js
+++ b/js/src/forum/components/PostStream.js
@@ -42,6 +42,7 @@ export default class PostStream extends Component {
         content = PostComponent ? PostComponent.component({ post }) : '';
 
         attrs.key = 'post' + post.id();
+        attrs.oncreate = (vnode) => setTimeout(() => $(vnode.dom).removeClass('fadeIn'), 500); // We give enough time for the animation to complete
         attrs['data-time'] = time.toISOString();
         attrs['data-number'] = post.number();
         attrs['data-id'] = post.id();
@@ -69,7 +70,7 @@ export default class PostStream extends Component {
       }
 
       return (
-        <div className="PostStream-item" {...attrs}>
+        <div className="PostStream-item fadeIn" {...attrs}>
           {content}
         </div>
       );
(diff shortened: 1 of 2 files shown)


js/src/forum/components/PostStream.js
```diff
@@ -89,7 +90,7 @@ export default class PostStream extends Component {
     // is not already doing so, then show a 'write a reply' placeholder.
     if (viewingEnd && (!app.session.user || this.discussion.canReply())) {
       items.push(
-        <div className="PostStream-item" key="reply">
+        <div className="PostStream-item fadeIn" key="reply">
```
clarkwinkelmann, line 93: This class won't be removed though, no?

js/src/forum/components/PostStream.js
```diff
@@ -42,6 +42,7 @@ export default class PostStream extends Component {
         content = PostComponent ? PostComponent.component({ post }) : '';
 
         attrs.key = 'post' + post.id();
+        attrs.oncreate = (vnode) => setTimeout(() => $(vnode.dom).removeClass('fadeIn'), 500); // We give enough time for the animation to complete
```
clarkwinkelmann, line 45: Mithril-wise wouldn't it be better to insert the `fadeIn` class via jQuery here? Right now I think if an extension triggers a full redraw of the page, the `fadeIn` class would be re-added.

Something like

```
attrs.oncreate = (vnode) => {
    $(vnode.dom).addClass('fadeIn');
    setTimeout(() => $(vnode.dom).removeClass('fadeIn'), 500)
};
```

I'd suggest a comment that references the original duration being CSS. Something like "// Duration of the fadeIn CSS animation + 100ms so it has time to complete"

clarkwinkelmann (changes requested):

clarkwinkelmann (approved):
Looking good code-wise :+1:

datitisev (approved):
Tested locally, works well & I think it improves the animations as well.

PR #2365, opened 2020-10-06
SubtreeRetainer: fix onbeforeupdate needsRebuild

w-4:
Previously the subtree retainer check was done in the beginning of the view function. 
Now being moved to the onbeforeupdate function, it's missing the initial call to set the data and returns `true` on the first call.

### DIFF
--- a/js/src/common/utils/SubtreeRetainer.js
+++ b/js/src/common/utils/SubtreeRetainer.js
@@ -28,6 +28,9 @@ export default class SubtreeRetainer {
   constructor(...callbacks) {
     this.callbacks = callbacks;
     this.data = {};
+    // Build the initial data, so it is available when calling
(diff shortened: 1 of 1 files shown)


datitisev:
🤔 Not sure if it should be called in constructor. Aren't the checks added after the constructor is initialized?

w-4:
Right, sometimes they are, I missed that. We could call `needsRebuild` in the `SubtreeRetainer.check` function aswell. 
Or remove the modification in `SubtreeRetainer`, and call `needsRebuild` from `oncreate` in the component.

askvortsov1 (approved):
Thank you!

datitisev:
I was thinking running it in `oncreate`, but that'd have to be included in each component and this is a simpler solution. I think this is good :+1:

PR #2366, opened 2020-10-07
Fix PostStream Reply Scroll

### LINKED ISSUE #2334: [Frontend Rewrite] PostStream won't scroll to end if last post bigger than page
If the last post in a discussion takes up the whole page, clicking "now" in the scrubber (or otherwise trying to scroll to the end programmatically) won't actually take you to the end.

We should reintroduce a special case for scrolling all the way to the end, instead of relying on the logic we have now, where it assumes that if we're on the last post, we're at the end.

askvortsov1:
Fixes #2334

Currently in master, scrolling to the reply placeholder is considered the same as scrolling to the last post. This relation doesn't actually hold in reality, and as a result, neither is done well. This PR decouples the two actions, and fixes both `PostStreamState`'s `goToLast` method, and actually scrolling to  reply via `goToNumber('reply')`. It also renames references to the "bottom" flag to "reply": we currently only scroll to the bottom of a post when replying; otherwise, it is not done well. This way, the purpose behind that flag is clearer, and more closely matches implementation.

### DIFF
--- a/js/src/forum/components/PostStream.js
+++ b/js/src/forum/components/PostStream.js
@@ -312,12 +312,12 @@ export default class PostStream extends Component {
    * @return {jQuery.Deferred}
    */
   scrollToIndex(index, animate, reply) {
+    const $item = reply ? $('.PostStream-item:last-child') : this.$(`.PostStream-item[data-index=${index}]`);
+
+    this.scrollToItem($item, animate, true, reply);
+
     if (reply) {
-      const $placeholder = this.$('.PostStream-item:last-child');
-      this.scrollToItem($placeholder, animate, true, true);
-      this.flashItem($placeholder);
-    } else {
-      this.scrollToItem(this.$(`.PostStream-item[data-index=${index}]`), animate, true);
+      this.flashItem($item);
     }
   }
 
--- a/js/src/forum/components/PostStream.js
+++ b/js/src/forum/components/PostStream.js
@@ -359,7 +359,7 @@ export default class PostStream extends Component {
       // We manually set the index because we want to display the index of the
       // exact post we've scrolled to, not just that of the first post within viewport.
       this.updateScrubber();
-      this.stream.index = index;
+      this.stream.index = index || this.stream.index;
     };
 
     // If we don't update this before the scroll, the scrubber will start
--- a/js/src/forum/components/PostStream.js
+++ b/js/src/forum/components/PostStream.js
@@ -129,15 +129,15 @@ export default class PostStream extends Component {
    * Start scrolling, if appropriate, to a newly-targeted post.
    */
   triggerScroll() {
-    if (!this.attrs.targetPost || !this.stream.needsScroll) return;
+    if (!this.stream.needsScroll) return;
 
-    const newTarget = this.attrs.targetPost;
+    const target = this.stream.targetPost;
     this.stream.needsScroll = false;
 
-    if ('number' in newTarget) {
-      this.scrollToNumber(newTarget.number, this.stream.animateScroll);
-    } else if ('index' in newTarget) {
-      this.scrollToIndex(newTarget.index, this.stream.animateScroll, newTarget.reply);
+    if ('number' in target) {
+      this.scrollToNumber(target.number, this.stream.animateScroll);
+    } else if ('index' in target) {
+      this.scrollToIndex(target.index, this.stream.animateScroll, target.reply);
     }
   }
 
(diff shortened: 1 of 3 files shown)


js/src/forum/components/PostStream.js
```diff
@@ -296,18 +295,17 @@ export default class PostStream extends Component {
...
-    const $item = this.$(`.PostStream-item[data-index=${index}]`);
-
-    return this.scrollToItem($item, animate, true, bottom).then(() => {
-      if (index == this.stream.count() - 1) {
-        this.flashItem(this.$('.PostStream-item:last-child'));
-      }
-    });
+  scrollToIndex(index, animate, reply) {
+    if (reply) {
+      const $placeholder = this.$('.PostStream-item:last-child');
```
clarkwinkelmann, line 303: Couldn't we call `goToIndex()` with `this.count()` in `goToNumber()`, that way then this bit of code gets executed, `index` already points to the placeholder? We can then remove the `if/else` block and just set the `reply` parameter to `scrollToItem`

clarkwinkelmann (approved):
Reading the code, I'm not seeing anything odd.

I've just one comment regarding a possible optimization

js/src/forum/components/PostStream.js
askvortsov1: The code is nicer, but unfortunately it's a bit glitchy when an event post has been created, but we haven't scrolled to it yet. When ~battling~ working with PostStream over the past few months, I've found that being as explicit as possible is best.

That being said, I rewrote that bit to be a little cleaner.

clarkwinkelmann: Good enough then :+1:

js/src/forum/components/PostStream.js
```diff
@@ -360,18 +357,22 @@ export default class PostStream extends Component {
...
+      // as it will soon be rendered by the "load more" system.
+      let itemOffset;
+      if (reply) {
+        const $placeholder = $('.PostStream-item:last-child');
+        $(window).scrollTop($placeholder.offset().top + $placeholder.height() - $(window).height() + app.composer.computedHeight());
+      } else if (index === 0) {
         $(window).scrollTop(0);
-      } else if (offset) {
-        $(window).scrollTop($(`.PostStream-item[data-index=${index}]`).offset().top - this.getMarginTop());
+      } else if ((itemOffset = $(`.PostStream-item[data-index=${index}]`).offset())) {
```
w-4, line 374: `index` can be undefined. If we passed in an `$item` to the `scrollToItem` function that has length zero `const index = $item.data('index');` would be undefined. In that case it won't scroll to the post after the promise resolves, and it would also in the next step set `this.stream.index = undefined` in `updateScrubberHeight`, which could result in bugs. It's not specifically something this PR has introduced, but might be worth fixing here.

askvortsov1: If `$item` is empty, there isn't much we can do though: we'd need to pass in the post index to `scrollToItem`, which defeats the purpose of having `scrollToItem` be its own function in the first place.

Have you run into cases where `$item` is passed in empty? `triggerScroll` is called in `onupdate`, after DOM modifications have been made, so we should expect the post item container html elements to match the range set in `PostStreamState` when we triggered the scroll.

js/src/forum/components/PostStream.js
```diff
@@ -349,7 +346,7 @@ export default class PostStream extends Component {
       // We manually set the index because we want to display the index of the
       // exact post we've scrolled to, not just that of the first post within viewport.
       this.updateScrubber();
-      this.stream.index = index;
+      this.stream.index = index || this.stream.index;
```
w-4, line 349: Shouldn't it be `index + 1`? 
```suggestion
     if (index !== undefined) this.stream.index = index + 1;
```
We input the post index (which can be 0), but we use this.stream.index as the scrubber index in other parts of the code (which is post index + 1).

js/src/forum/components/PostStream.js
w-4: I think it's fine then. I just tested it, and haven't run into that case.

datitisev:
Not sure if this is supposed to change the behavior when clicking "Now", but clicking it looks to go to the last post and not the reply placeholder still ?

askvortsov1:
Yep! That's the intended behavior, which is also consistent across major forum platforms.

datitisev:
Not sure how exactly I can test this locally, but I didn't encounter any bugs.

PR #2367, opened 2020-10-07
Reduce modal hide timeout

w-4:
**Fixes a minor bug:**
as a result of: https://github.com/flarum/core/pull/2332

When opening a modal quickly after closing one, it shows 2 modals for a moment.

**Screenshot**
image

**Confirmed**

### DIFF
--- a/js/src/common/components/Modal.js
+++ b/js/src/common/components/Modal.js
@@ -35,7 +35,7 @@ export default class Modal extends Component {
       this.attrs.animateHide();
       // Here, we ensure that the animation has time to complete.
       // See https://mithril.js.org/lifecycle-methods.html#onbeforeremove
-      return new Promise((resolve) => setTimeout(resolve, 250));
+      return new Promise((resolve) => setTimeout(resolve, 300));
     }
   }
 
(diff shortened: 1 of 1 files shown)


askvortsov1 (approved):
Thanks!

clarkwinkelmann:
Bootstrap's animation duration appears to be `300` ms. Do we want to use that instead? https://github.com/twbs/bootstrap/blob/v3.4.1/js/modal.js#L39

We could access the constant via `$.fn.modal.Constructor.TRANSITION_DURATION` if we want to avoid a magic number in our code.

askvortsov1:
The constant would probably be preferable, but if we want to use the number directly, we should add a comment explaining the choice.

w-4:
I went with the number, since we're soon replacing bootstrap with other packages.

js/src/common/components/Modal.js
```diff
@@ -35,7 +35,8 @@ export default class Modal extends Component {
       this.attrs.animateHide();
       // Here, we ensure that the animation has time to complete.
       // See https://mithril.js.org/lifecycle-methods.html#onbeforeremove
-      return new Promise((resolve) => setTimeout(resolve, 1000));
+      // Bootstrap's default animation duration is 300 ms.
```
clarkwinkelmann, line 38: One last suggestion (I hope!), I'd suggest the following comment so that one can more easily find where the 300ms is defined in bootstrap's source code, should it be necessary.

```suggestion
      // Bootstrap's Modal.TRANSITION_DURATION is 300 ms.
```

datitisev (approved):

PR #2369, opened 2020-10-07
Make PostStreamScrubber work for Posts that have top margin

### LINKED ISSUE #1897: Page count incorrect when scrolling to bottom
## Bug Report

**Current Behavior**
When scrolling past the last post to the editor, the page count changes to an incorrect value. This has been reproduced across a range of devices and browsers by myself and @Ralkage 

**Steps to Reproduce**
1. Go to any thread. 
2. Browse to the last post and note the post count (eg 399/399)
3. Scroll down to the editor
4. See error - the post count will change (e.g. 349/399) 

**Video of bug**
https://streamable.com/avmrw <--

w-4:
Basically the same as https://github.com/flarum/core/pull/1944. I now know what that PR was fixing. When Posts have a top margin the scrubber will bounce around. 

To replicate go to https://nightly.flarum.site, set `margin-top: 50px` for `.Post` in the dev console. See scrubber bouncing back and forth when scrolling.

Not a priority, but should still be fixed I think.

Edit: New commit fixes #1897

**Confirmed**

### DIFF
--- a/js/src/forum/components/PostStream.js
+++ b/js/src/forum/components/PostStream.js
@@ -225,7 +225,7 @@ export default class PostStream extends Component {
       if (time) period = time;
     });
 
-    let index = indexFromViewPort !== null ? indexFromViewPort : $items.first().data('index') || 0;
+    let index = indexFromViewPort !== null ? indexFromViewPort : 0;
     this.stream.index = index + 1;
     this.stream.visible = visible;
     if (period) this.stream.description = dayjs(period).format('MMMM YYYY');
--- a/js/src/forum/components/PostStream.js
+++ b/js/src/forum/components/PostStream.js
@@ -181,10 +181,9 @@ export default class PostStream extends Component {
     // seen if the browser were scrolled right up to the top of the page,
     // and the viewport had a height of 0.
     const $items = this.$('.PostStream-item[data-index]');
-    let index = $items.first().data('index') || 0;
     let visible = 0;
     let period = '';
-    let indexWasUpdated = false;
+    let indexFromViewPort = null;
 
     // Now loop through each of the items in the discussion. An 'item' is
     // either a single post or a 'gap' of one or more posts that haven't
@@ -212,9 +211,8 @@ export default class PostStream extends Component {
 
       // We take the index of the first item that passed the previous checks.
       // It is the item that is first visible in the viewport.
-      if (!indexWasUpdated) {
-        index = parseFloat($this.data('index')) + visibleTop / height;
-        indexWasUpdated = true;
+      if (indexFromViewPort === null) {
+        indexFromViewPort = parseFloat($this.data('index')) + visibleTop / height;
       }
 
       if (visiblePost > 0) {
@@ -227,6 +225,7 @@ export default class PostStream extends Component {
       if (time) period = time;
     });
 
+    let index = indexFromViewPort !== null ? indexFromViewPort : $items.first().data('index') || 0;
     this.stream.index = index + 1;
     this.stream.visible = visible;
     if (period) this.stream.description = dayjs(period).format('MMMM YYYY');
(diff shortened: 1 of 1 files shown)


askvortsov1:
Hmm, I remember I was looking at this back when doing #2160, but didn't include it for some reason, don't remember why though. I'll test it locally

askvortsov1 (approved):
Tested locally, this fixes the issue and doesn't seem to introduce any others. Thanks!

clarkwinkelmann (commented):
That looks good, but would it maybe be prettier to use two number variables and no boolean?

Pseudo-code:

```
indexFromViewPort = null;
$.each(() => {
    // [...]
    indexFromViewPort = parseFloat(...) + visibleTop/height
    // [...]
})
// [...]
index = indexFromViewPort || $items.first().data() || 0
```

askvortsov1:
> That looks good, but would it maybe be prettier to use two number variables and no boolean?

That looks nice! Could also get us a bit closer to the implementation needed for #1897, so I like it

w-4:
Will do, we need to check against null though both times, since zero can be a valid index.

js/src/forum/components/PostStream.js
```diff
@@ -223,7 +225,7 @@ export default class PostStream extends Component {
       if (time) period = time;
     });
 
-    this.stream.index = index + 1;
+    this.stream.index = indexFromViewPort !== null ? indexFromViewPort + 1 : 1;
```
askvortsov1, line 228: Why remove `$items.first().data('index')`? Although actually, we might be able to solve that other issue in one swoop:

If no posts match, then we are either above or below all posts. Iff we are above all posts, then $items.first().data('index') should be equal to 0. If $items.first().data('index') !== 0, we cannot be above all posts, so we must be below all posts, so we should set index to `this.stream.count()`. I think that makes sense :rofl:?

js/src/forum/components/PostStream.js
```diff
@@ -209,8 +209,10 @@ export default class PostStream extends Component {
       const visibleBottom = Math.min(height, viewportTop + viewportHeight - top);
       const visiblePost = visibleBottom - visibleTop;
 
-      if (top <= viewportTop) {
-        index = parseFloat($this.data('index')) + visibleTop / height;
+      // We take the index of the first item that passed the previous checks.
+      // It is the item that is first visible in the viewport.
+      if (indexFromViewPort === null) {
+        indexFromViewPort = parseFloat($this.data('index')) + visibleTop / height;
```
w-4, line 215: The NaN errors described in https://discuss.flarum.org/d/23667-dev-diary-beta-14/123 and https://discuss.flarum.org/d/23667-dev-diary-beta-14/132, can be probably removed by checking if visibleTop is defined and setting it to 0 if it's not. No idea why it could be undefined though, haven't found a way to reproduce yet. I had this check in my codebase already for beta.13 though, so I believe it's not something new. 

```js
        indexFromViewPort = parseFloat($this.data('index'));
        // Setting visibleTop to zero in rare cases, where it could be undefined.
        let fraction = (visibleTop || 0) / height;
        indexFromViewPort += fraction;
```

js/src/forum/components/PostStream.js
w-4: It does make sense. What would be a scenario that we are above all posts? I simplified it (maybe incorrectly) in the latest commit to just assume we are below the posts.

js/src/forum/components/PostStream.js
askvortsov1: I believe the NaN issue is moreso the result of other issues than its own thing. #2366 and #2377 should eliminate most of those issues.

js/src/forum/components/PostStream.js
askvortsov1: On testing, that was an issue with the previous implementation, but under the new "check for null flag" system it doesn't have an issue.

datitisev (approved):
Looks like it fixes both issues mentioned (only one is an actual issue in the repo). Tested locally :+1:

PR #2371, opened 2020-10-07
Header UI fixes

askvortsov1:
## Issue 1: Missing Header on Refresh

Due to some magic in Mithril 0.1's context:retain flag, some DOM elements were cached across page reloads. Since that has been eliminated, if we refresh the page and we are scrolled down, the "affix" class which makes the header fixed (and as a result, visible) isn't applied until the first scroll. We fix this by running ScrollListener.update() immediately to set initial navbar state.

## Issue 2: Header Overlap with Custom Header

Example

This was accidentially introduced in #2131. In this PR, we revert that, and instead conditionally apply the navbar-fixed-top class only when needed, so that we can take advantage of it without always having the navbar in position:fixed, as was done in the previous solution.

### DIFF
--- a/js/src/common/Application.js
+++ b/js/src/common/Application.js
@@ -204,6 +204,9 @@ export default class Application {
       const offset = $app.offset().top;
 
       $app.toggleClass('affix', top >= offset).toggleClass('scrolled', top > offset);
+      // If we're scrolled down, add the navbar-fixed-top css class, so that Bootstrap's
+      // JS will compensate for shifts when the modal is opened/closed.
+      $('.App-header').toggleClass('navbar-fixed-top', top >= offset);
     }).start();
 
     $(() => {
--- a/js/src/common/Application.js
+++ b/js/src/common/Application.js
@@ -199,7 +199,7 @@ export default class Application {
 
     // Add a class to the body which indicates that the page has been scrolled
     // down.
-    new ScrollListener((top) => {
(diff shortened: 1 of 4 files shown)


clarkwinkelmann (approved):
Having not seen a demo of the original issue I'm unsure what exactly was happening. I see this mostly reverts #2131 so I suppose the regression is fixed.

Does this still fix the originally reported issue from #2131 ?

askvortsov1:
It does; it still adds that navbar-fixed-top class, but does so programatically.

js/src/common/Application.js
```diff
@@ -199,12 +199,18 @@ export default class Application {
 
     // Add a class to the body which indicates that the page has been scrolled
     // down.
-    new ScrollListener((top) => {
+    const scrollListener = new ScrollListener((top) => {
       const $app = $('#app');
       const offset = $app.offset().top;
 
       $app.toggleClass('affix', top >= offset).toggleClass('scrolled', top > offset);
```
w-4, line 206: Let's add a comment explaining that the affix class is used for when custom headers are displayed above Flarum's header.

askvortsov1: Well, affix itself is used so the header is sticky when scrolled down. But yeah I'll expand that comment a bit, thanks for the suggestion!

datitisev (approved):
Code makes sense. Haven't tested locally.

PR #2377, opened 2020-10-08
Don't scroll on m.route.set() to different post on same page.

askvortsov1:
This removes some messy logic, and the potential for glitches. This system worked well with Mithril 0.2 where we could listen in before (and prevent) page unload, but since that's not possible in Mithril 2, the implementation of the replacement was done in `onbeforeupdate`, which might be called while the page route is being updated, glitching out the page. Instead, extensions should check if they are already on the discussion page for the post they are linking to, and if so, use `app.current.get('stream').goToNumber(TARGET)`.

Please note that this does NOT affect going directly to posts from external links (or page reload), OR from other pages via m.route.set.

### DIFF
--- a/js/src/forum/components/DiscussionPage.js
+++ b/js/src/forum/components/DiscussionPage.js
@@ -99,28 +99,12 @@ export default class DiscussionPage extends Page {
   onbeforeupdate(vnode) {
     super.onbeforeupdate(vnode);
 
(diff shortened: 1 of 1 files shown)


askvortsov1:
Obsolete with https://github.com/flarum/core/pull/2378

Issue #2337, opened 2020-10-02
[Frontend Rewrite] Back button showing on IndexPage


w-4:
## Bug Report

See screenshot:
![image](https://user-images.githubusercontent.com/36057469/94893822-41d37900-04b2-11eb-8c20-487b94cb9b02.png)

**Steps to Reproduce**
1. Go to https://nightly.flarum.site
2. Click on 'Tags'
3. Click on 'All Discussions'

w-4:
This is a redraw issue. Adding `m.redraw()` to `app.history.push` would fix the issue (but it isn't pretty).

PR #2379, opened 2020-10-09
Fix search input value disappearing on page reload

w-4:
Similar to https://github.com/flarum/core/pull/2145, appears in beta.14. If you search for something, and then reload the page the input value disappears.

**Confirmed**

### DIFF
--- a/js/src/forum/states/GlobalSearchState.js
+++ b/js/src/forum/states/GlobalSearchState.js
@@ -7,7 +7,7 @@ export default class GlobalSearchState extends SearchState {
   }
 
   getValue() {
-    if (this.value == undefined || !this.value.length) {
+    if (this.value === undefined || this.value === '') {
       this.value = this.getInitialSearch() || '';
     }
 
--- a/js/src/forum/states/GlobalSearchState.js
+++ b/js/src/forum/states/GlobalSearchState.js
@@ -7,7 +7,7 @@ export default class GlobalSearchState extends SearchState {
   }
 
   getValue() {
-    if (this.value === undefined || !this.value.length) {
+    if (this.value == undefined || !this.value.length) {
       this.value = this.getInitialSearch() || '';
     }
 
--- a/js/src/forum/states/GlobalSearchState.js
+++ b/js/src/forum/states/GlobalSearchState.js
@@ -7,7 +7,7 @@ export default class GlobalSearchState extends SearchState {
   }
 
   getValue() {
-    if (this.value === undefined) {
+    if (this.value === undefined || !this.value.length) {
       this.value = this.getInitialSearch() || '';
     }
 


askvortsov1 (approved):
I'm assuming it's currently coming up as `null` instead of `undefined`?

w-4:
@askvortsov1 as an empty string

askvortsov1:
:facepalm: Ah yes of course because `null` would have length zero... Good job me :see_no_evil:!

js/src/forum/states/GlobalSearchState.js
```diff
@@ -7,7 +7,7 @@ export default class GlobalSearchState extends SearchState {
   }
 
   getValue() {
-    if (this.value === undefined || !this.value.length) {
+    if (this.value == undefined || !this.value.length) {
```
datitisev, line 10: Isn't this basically `!this.value` now? If `this.value` is `0` it'll `== undefined`

askvortsov1: We could just return `this.value || this.getInitialSearch() || ''` for simplicity?

datitisev: Yeah I think that's the same. `!!'' === false` so.

askvortsov1 (changes requested):
The syntax discussed in the comment would be simpler

js/src/forum/states/GlobalSearchState.js
askvortsov1: Although actually... What happens if an initial search is present, but then the search is cleared? With `this.value || this.getInitialSearch() || ''`, `getValue()` would still return that initial value. I think that's why it checked for undefined but not null.

w-4: The previous null check I added so it doesn't TypeError in case it actually is null when we check for the length. 
I don't think it can become null. Let's check for the empty string explicitly instead though.

askvortsov1 (changes requested):
With these changes, when you hit (x) to clear a search, the actual text isnt cleared. One way to fix that is to replace `super.clear()` with `this.setValue(null)` in `clear`

w-4:
Couldn't reproduce it locally. When `super.clear` is run, it sets `this.value` to `''`, and in `getvalue` we return `super.getvalue` which just returns `this.value`.

askvortsov1:
In get Value, we return `super.getValue` IF `this.value` is not ''. If `this.value` is `''`, as it would be after clear, we set `this.value = this.getInitialSearch()` before returning. Try searching, refreshing the page, and then hitting clear.

w-4:
Hm, it works for me when I test it. Inside `clear` we also run `if (this.getInitialSearch()) this.clearInitialSearch();`.

I've discovered another bug though: when you search, then open a discussions, then go back, the clear button disappears. (This is unrelated though, and similar to https://github.com/flarum/core/issues/2337, a redraw bug of the header.

askvortsov1:
It doesn't always show up, but I've been able to replicate it (and something else: backspacing very fast results in the input not being deleted)

https://www.youtube.com/watch?v=4ENp6ALBf5c&feature=youtu.be

PR #2382, opened 2020-10-10
Fix routing forced refresh

w-4:
**Fixes the following issue:**

2. Click the home link
3. It doesn't refresh the page, when already on the '/' route.

Depends on https://github.com/flarum/core/pull/2378. Going to rebase after it's merged.

Use the route resolver `onmatch` method to check if links have requested a forced refresh. 

Do we want to make the force refresh the default for links created with the `Link` component?

**Confirmed**

### DIFF
--- a/js/src/common/components/Link.js
+++ b/js/src/common/components/Link.js
@@ -31,10 +31,12 @@ export default class Link extends Component {
       if (!('replace' in options)) options.replace = true;
     }
 
-    // Mithril 2 does not completely rerender the page if a route change leads to the same route
-    // (or the same component handling a different route).
-    // Here, the `force` parameter will use Mithril's key system to force a full rerender
-    // see https://mithril.js.org/route.html#key-parameter
+    // Mithril 2 does not completely rerender the page if a route change leads
+    // to the same route.
+    // Here, the `force` parameter will use Mithril's key system to force a full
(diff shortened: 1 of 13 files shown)


w-4:
Regarding the links here are the options:
1. Use the Link component `force` attribute, so that clicking on a link twice will both times fully re-render the page, otherwise it will just diff the page (when the URL is the same). We can then decide on a case by case basis.
2.  Or make the above `force` behavior the default for all Links. This was the case in beta.13. 

When setting routes programatically, the default behavior will be to just diff, if the URL is identical. One can use the `setRouteWithForcedRefresh` helper instead of `m.route.set` to make the page fully re-render.

askvortsov1:
I prefer option 1. With the changes in #2378, that behavior is recovered where relevant. The only change I'd make is for LinkButtons to have `force` enabled.

Also, when trying out #2378, `setRouteWithForcedRefresh` seemed to work. I'm not sure if the changes are necessary, although I do like that the logic for key generation will be fully contained in the route resolver. I don't like `app.forceRefresh` though (preferable to avoid introducing new globals where possible imo): couldn't we modify the options passed into `m.route.set` like with the options in the `Link` component?

w-4:
Let's close this, it seems to work now more or less. The only thing that's weird is that the site refreshes for me on https://nightly.flarum.site, but not when I run it locally.

PR #2384, opened 2020-10-12
Fix PostStream loadRange doesn't return all posts

w-4:
**Fixes the following issues:**

PostStream `loadRange` is called like this from `loadNearIndex`:
```js
this.loadRange(start, end).then(this.show.bind(this));
```
The `show` function expects to always receive the whole range. 
Whenever posts are partially loaded, it just returned the new posts, which caused bugs when using the scrubber to go to a post index.

**Confirmed**

### DIFF
--- a/js/src/forum/states/PostStreamState.js
+++ b/js/src/forum/states/PostStreamState.js
@@ -282,7 +282,13 @@ class PostStreamState {
         }
       });
 
-    return loadIds.length ? app.store.find('posts', loadIds) : Promise.resolve(loaded);
+    if (loadIds.length) {
+      return app.store.find('posts', loadIds).then((newPosts) => {
+        return loaded.concat(newPosts).sort((a, b) => a.id() - b.id());
+      });
+    }
+
+    return Promise.resolve(loaded);
   }
 
   /**
--- a/js/src/forum/states/PostStreamState.js
+++ b/js/src/forum/states/PostStreamState.js
@@ -284,7 +284,7 @@ class PostStreamState {
 
     if (loadIds.length) {
       return app.store.find('posts', loadIds).then((newPosts) => {
-        return loaded.concat(newPosts).sort((a, b) => a.id() - b.id());
+        return loaded.concat(newPosts).sort((a, b) => a.createdAt() - b.createdAt());
       });
     }
 
--- a/js/src/forum/components/DiscussionPage.js
+++ b/js/src/forum/components/DiscussionPage.js
@@ -155,7 +155,7 @@ export default class DiscussionPage extends Page {
             record.relationships.discussion.data.id === discussionId
         )
         .map((record) => app.store.getById('posts', record.id))
-        .sort((a, b) => a.id() - b.id())
+        .sort((a, b) => a.createdAt() - b.createdAt())
         .slice(0, 20);
     }
 


js/src/forum/states/PostStreamState.js
```diff
@@ -280,7 +280,13 @@ class PostStreamState {
         }
       });
 
-    return loadIds.length ? app.store.find('posts', loadIds) : Promise.resolve(loaded);
+    if (loadIds.length) {
+      return app.store.find('posts', loadIds).then((newPosts) => {
+        return loaded.concat(newPosts).sort((a, b) => a.id() - b.id());
```
datitisev, line 285: Posts should be sorted by number, not ID. An ext like e.g. fof/merge-discussions would break this.

Did the posts previously not get sorted in the JS (before rewrite)? 🤔

datitisev (changes requested):

js/src/forum/states/PostStreamState.js
w-4: Posts are sorted by id in DiscussionPage (although that is probably redundant). I think as long as fof/merge-discussions sorts posts by creation_date and does not alter the creation date, it should be fine. Otherwise the backend APIs would also need to be adjusted to calculate the (`near`) offset based on the post number rather than the date.

datitisev: That doesn't make sense then 🤔 we have a number column, why is it not being used for the sorting?

Also merge-discussions sets the number column based on creation I think, so I guess that's basically the same thing as ID.

Maybe this is not an issue. I think I had just woken up when I commented 🤔 not entirely sure.

askvortsov1: I would prefer to use `created_at` (or `number`) for sorting, as those columns are there to explicitly represent post order. `id` is ordered, but that's not one of it's purposes

w-4: Agree, let's keep consistent with how the backend sorts posts. I updated it to sort by creation date.

askvortsov1 (approved):
Thanks!

js/src/forum/states/PostStreamState.js
datitisev: Hm still don't like that it's creation date, but that's a separate issue.

askvortsov1: Any specific concerns about creation date? Although now that I think about it, doesn't this mean that if 2 posts are created at the same time, their ordering might vary from time to time? From that perspective, `number` might be better, at least to keep things constant.

w-4: I remember, that there is an admin API option to modify the creation date when creating posts. So post numbers can be in a different order than the ids (and creation dates).

datitisev: I'm just saying that sorting in the frontend instead of by the defined numbering system we already have set up is a separate issue that should be addressed separate from this PR.

datitisev: I know. I'm saying that this is a separate issue that we should address at another point.

I'm trying to stop the conversation about it in this PR but you just keep continuing it 🙂

askvortsov1: Alright, sorry for the confusion. If you're good with the PR, could you approve it?

clarkwinkelmann (approved):
I have not tested the code, but the approach looks good.

I was also under the impression that the Post number was the main way we sorted posts, but looking at the source I see `createdAt` is used everywhere in the backend, so I guess that makes sense to continue like this and maybe open an issue about the possible change to `number` for sorting.

PR #2385, opened 2020-10-12
Fix mobile PostStream top scroll adjustment & remove App:before

w-4:
Removing App:before, the behavior I believe doesn't change and is a lot cleaner. This has several advantages:
- the mobile header has now height, which was needed for the PostStream fix to work.
- it's now possible to access the mobile header via element id, and f.e. change the color programmatically on mobile. (I've been using that on my site for quite a while)

### DIFF
--- a/js/src/forum/components/PostStream.js
+++ b/js/src/forum/components/PostStream.js
@@ -275,7 +275,9 @@ export default class PostStream extends Component {
    * @return {Integer}
    */
   getMarginTop() {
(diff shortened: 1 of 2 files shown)


askvortsov1 (approved):
Seems to fix the issue when testing locally, haven't run into side effects. Thanks!

tankerkiller125 (approved):

iPurpl3x:
This broke my CSS when upgrading... Took a while to find out what had changed !

PR #2387, opened 2020-10-13
Passive Scroll Event Listener 

w-4:
see: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener

> A Boolean that, if true, indicates that the function specified by listener will never call preventDefault(). If a passive listener does call preventDefault(), the user agent will do nothing other than generate a console warning. See Improving scrolling performance with passive listeners to learn more.

The `ScrollListener` util is used in PostStream and PostStreamScrubber.

**Confirmed**


askvortsov1 (approved):
Great find!

PR #2388, opened 2020-10-13
PostStream: Fix minor load more issue

w-4:
When we scroll to a post that is already loaded, the load more button might show and not trigger. 
To replicate (in beta.14) go to the start of a long enough discussion and run `app.current.get('stream').goToIndex(19)`.

**Screenshot**
before the fix:
Screenshot_316

**Confirmed**

### DIFF
--- a/js/src/forum/states/PostStreamState.js
+++ b/js/src/forum/states/PostStreamState.js
@@ -172,7 +172,7 @@ class PostStreamState {
    * @return {Promise}
    */
   loadNearIndex(index) {
-    if (index >= this.visibleStart && index < this.visibleEnd) {
+    if (index >= this.visibleStart && index <= this.visibleEnd) {
       return Promise.resolve();
     }
 
--- a/js/src/forum/states/PostStreamState.js
+++ b/js/src/forum/states/PostStreamState.js
@@ -172,7 +172,7 @@ class PostStreamState {
    * @return {Promise}
    */
   loadNearIndex(index) {
-    if (index >= this.visibleStart && index <= this.visibleEnd) {
+    if (index >= this.visibleStart && index < this.visibleEnd) {
       return Promise.resolve();
     }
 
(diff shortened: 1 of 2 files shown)


askvortsov1:
When testing locally, `app.current.get('stream').goToIndex(19)` on a long discussion with these changes results in repeated scrolling to and from the scrolled post.

w-4:
This is might be due to https://github.com/flarum/core/pull/2384. Let's merge that in first and check again.

w-4:
We could later also check by running `app.current.get('stream').goToNumber(20)`.  But I'd also wait for https://github.com/flarum/core/pull/2378 to be merged. It removes the `m.route.set` call in DiscussionPage `postionChanged` that could lead to an additional `goToNumber` call.

askvortsov1 (changes requested):
Tested it again, and it works correctly with `app.current.get('stream').goToIndex(19)` (yay!).

 The only nitpick I have is that I think `loadPostsIfNeeded` would be a better method name than `loadPostsWhenNeeded`

w-4:
Tested it as well, and it works. Also fixed the `app.current.get('stream').goToIndex(20)`.

js/src/forum/states/PostStreamState.js
```diff
@@ -172,7 +172,7 @@ class PostStreamState {
    * @return {Promise}
    */
   loadNearIndex(index) {
-    if (index >= this.visibleStart && index <= this.visibleEnd) {
```
askvortsov1, line 175: This was already done in https://github.com/flarum/core/commit/09e2736cbcc267594b660beabbd001d9030f9880, so a rebase would be better than repeating this change

askvortsov1 (changes requested):
The `goToIndex(20)` issue was already fixed in https://github.com/flarum/core/commit/09e2736cbcc267594b660beabbd001d9030f9880, so I don't think we should duplicate it here (especially since the PR is about another topic), but otherwise LGTM!

askvortsov1 (approved):
Tested locally, works as intended

SychO9 (approved):

PR #2389, opened 2020-10-14
Hotfix the redraw of header components for beta.14

w-4:
**Fixes #2337**

With that additional redraw, header components can update page-specific conditionals like the backbutton, or search. 

**Confirmed**

- [x] Frontend changes: tested on a local Flarum installation.

PR #2390, opened 2020-10-14
Make header and navigation components redraw after page components

### LINKED ISSUE #2337: [Frontend Rewrite] Back button showing on IndexPage

w-4:
**Fixes #2337 and replaces #2389**:

The issue: Page components manipulate the header (state), but the header redraws before the page on route change.

By changing the mount order, we change also the redraw order.

**Confirmed**

### DIFF
--- a/js/src/admin/AdminApplication.js
+++ b/js/src/admin/AdminApplication.js
@@ -27,19 +27,18 @@ export default class AdminApplication extends Application {
    * @inheritdoc
    */
   mount() {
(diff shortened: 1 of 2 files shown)


askvortsov1 (approved):
Makes sense, thanks!

datitisev (approved):
This makes sense. I hadn't thought of the ordering here affecting stuff like state. This shouldn't break anything else, as I don't think the other components access the mounted ones (at least shouldn't).

*Not tested locally by me

PR #2391, opened 2020-10-14
Convert common time helpers to Typescript

lhsazevedo:
See flarum/framework#3533

### DIFF
--- a/js/src/common/helpers/fullTime.tsx
+++ b/js/src/common/helpers/fullTime.tsx
@@ -1,11 +1,11 @@
+import dayjs from 'dayjs';
+import * as Mithrill from 'mithril';
+
 /**
  * The `fullTime` helper displays a formatted time string wrapped in a <time>
  * tag.
- *
- * @param {Date} time
- * @return {Object}
  */
-export default function fullTime(time) {
+export default function fullTime(time: Date): Mithrill.Vnode {
   const d = dayjs(time);
 
   const datetime = d.format();
--- a/js/src/common/helpers/fullTime.tsx
+++ b/js/src/common/helpers/fullTime.tsx
@@ -1,11 +1,11 @@
 import dayjs from 'dayjs';
-import * as Mithrill from 'mithril';
+import * as Mithril from 'mithril';
 
 /**
  * The `fullTime` helper displays a formatted time string wrapped in a <time>
  * tag.
  */
-export default function fullTime(time: Date): Mithrill.Vnode {
+export default function fullTime(time: Date): Mithril.Vnode {
   const d = dayjs(time);
 
   const datetime = d.format();
--- a/js/src/common/helpers/humanTime.tsx
+++ b/js/src/common/helpers/humanTime.tsx
@@ -1,14 +1,13 @@
 import humanTimeUtil from '../utils/humanTime';
+import * as Mithril from 'mithril';
+import * as dayjs from 'dayjs';
 
 /**
  * The `humanTime` helper displays a time in a human-friendly time-ago format
  * (e.g. '12 days ago'), wrapped in a <time> tag with other information about
  * the time.
- *
- * @param {Date} time
- * @return {Object}
  */
-export default function humanTime(time) {
+export default function humanTime(time: Date): Mithril.Vnode {
   const d = dayjs(time);
 
   const datetime = d.format();
--- a/js/src/common/helpers/humanTime.tsx
+++ b/js/src/common/helpers/humanTime.tsx
@@ -1,6 +1,6 @@
 import humanTimeUtil from '../utils/humanTime';
 import * as Mithril from 'mithril';
-import * as dayjs from 'dayjs';
+import dayjs from 'dayjs';
 
 /**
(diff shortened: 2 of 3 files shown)


js/src/common/helpers/fullTime.tsx
```diff
@@ -1,11 +1,11 @@
+import dayjs from 'dayjs';
+import * as Mithrill from 'mithril';
```
askvortsov1, line 2: Typo

askvortsov1 (changes requested):
This mostly makes sense, thanks!

clarkwinkelmann (commented):
Why does one import uses `import * as dayjs from 'dayjs'; ` while the others use `import dayjs from 'dayjs'; ` ?

Do we actually want to import global variables or could we just type-hint them somewhere? Using the global variable instead of an import will save up on the file size, as webpack won't have to add the additional code that resolves each import.

datitisev:
@clarkwinkelmann We can test locally if the bundle size increases if they're just used for typings. Haven't done so myself yet.

js/src/common/helpers/fullTime.tsx
lhsazevedo: Fixed

lhsazevedo:
I removed the `import * as dayjs` variant and now all three changed files use the default export

js/src/common/helpers/fullTime.tsx
askvortsov1: Thanks! :D

askvortsov1 (approved):
Looks good to me! I'm approving this as long as

> Do we actually want to import global variables or could we just type-hint them somewhere? Using the global variable instead of an import will save up on the file size, as webpack won't have to add the additional code that resolves each import.

doesn't prove to be an issue

lhsazevedo:
@askvortsov Thanks, I'll test this as soon as possible.

lhsazevedo:
## Test
So, apparently webpack doesn't inject the module resolving code on each module, but use a global one. [Here's the test i did](https://gist.github.com/lhsazevedo/a49c5962db67251313e0b22ebbed4d37).

## Flarum build
Heres `src/common/utils/humanTime.ts` unminified output:
```js
"use strict";
__webpack_require__.d(__webpack_exports__, "a", function() { return humanTime; });
var dayjs = __webpack_require__(28);
var dayjs_default = /*#__PURE__*/__webpack_require__.n(dayjs);
var dayjs_plugin_relativeTime = __webpack_require__(73);
var dayjs_plugin_relativeTime_default = /*#__PURE__*/__webpack_require__.n(dayjs_plugin_relativeTime);

function humanTime(time) {
  var d = dayjs_default()(time);
  var now = dayjs_default()();

  // ...
}
```

## Dayjs plugins

One thing we could optimize is, even not using directly, we need to import relativeTime dayjs plugin at `js/src/common/utils/humanTime.ts`. Without this import, typescript can't pick up tipes for the [fromNow()](https://github.com/flarum/core/blob/2ff46f5a2ebe2532f780f6e60a921beb41876d69/js/src/common/utils/humanTime.ts#L32) method and complains.
The code still works though, because relativeTime plugin is installed at `js/src/common/index.js`:

https://github.com/flarum/core/blob/2ff46f5a2ebe2532f780f6e60a921beb41876d69/js/src/common/index.js#L12-L16

We can avoid this import if we move this plugin installation process to a separated dayjs library module, something like:
```typescript
// js/src/common/lib/dayjs.ts

import dayjs from 'dayjs'

dayjs.extend(relativeTime)

export default dayjs


// js/src/common/utils/humanTime.ts

// no need import relativeTime from 'dayjs/plugin/relativeTime';
import dayjs from '../../common/lib/dayjs'
```
~Should I open a new issue for this?~ See flarum/framework#2417

askvortsov1:
> So, apparently webpack doesn't inject the module resolving code on each module, but use a global one. [Here's the test i did](https://gist.github.com/lhsazevedo/a49c5962db67251313e0b22ebbed4d37).

Thank you for testing this!



> We can avoid this import if we move this plugin installation process to a separated dayjs library module, something like:

Hmm I really like this idea! @clarkwinkelmann @rob006 could this also work for https://discuss.flarum.org/d/25196-beta-14-dayjs-translations?

datitisev:
@askvortsov1 It could technically work but the config files would need to use `flarum.core.compat['flarum/lib/dayjs']` to access it.

lhsazevedo:
It seems that it would be a more drastic change, and the code already works without it, so I created [a new issue](https://github.com/flarum/core/issues/2417) for this subject so that we can move forward here.

clarkwinkelmann (approved):
I'm triggered by the imports order ahah (`fullTime` imports dayjs first, while `humanTime` imports Mithril first :hear_no_evil: )

Looks good otherwise.

lhsazevedo:
@clarkwinkelmann Fixed following this rule: external imports first, internal last, both alphabetically sorted.

tankerkiller125 (approved):

Issue #1951, opened 2019-12-07
Not scrolling up when a new tag list page is loaded from mobile dropdown


nitaaikumar:
There seems to be quite an important bug in mobiles (both android and ios). I don't know if it has been introduced in Beta 11 or if it was there in Beta 10 also.

If we are scroll down in the All Discussions list index page and then change the tag from the top dropdown on mobile devices then it does not jump to the top of the tag page as it should and it stays at the bottom scrolled position.

This may be quite confusing for the mobile user who may not know what has happened if he/she does not try to scroll up again when the tag page loads, especially if that tag has less number of posts than the All Discussions index page, then all one could see is the footer when the tag is changed in the dropdown.

franzliedke:
I cannot reproduce this with Firefox on Android.

Do you see this issue at https://discuss.flarum.org/ or https://beta.flarum.site/?

askvortsov1:
I can confirm this on discuss.flarum.org with Chrome on Android.

tankerkiller125:
@askvortsov1 how far down did you scroll, I'm on Android 10 with latest Chrome and Chrome Dev and I was unable to replicate the issue scrolling down to hit "load more" twice.

askvortsov1:
When testing yesterday, I scrolled all the way down but didn't hit load more at all. That being said, when I went to test again now, it worked as expected, except for one time. So it's one of _these_ bugs...

askvortsov1:
Closed in https://github.com/flarum/core/commit/1936b9117ded44be291cb002c734081497933e13

PR #2396, opened 2020-10-16
Page Scroll Cleanup

askvortsov1:
This PR makes several modifications to retain (or fix) beta 13 behavior:

- Whenever a new page is rendered, scroll to the top. This was previous behavior. This can also be cancelled in a page by setting `this.dontScrollTopOnCreate` in `oninit`.
- Remove some now-obsolete hacks for browsers, since browsers have gotten better. This also fixes a flash of the top of the page when refreshing.
- Fix broken selector for scrolling to discussion if its out of view
- On mobile, only retain scroll position if coming from a discussion (fixes https://github.com/flarum/core/issues/2284, https://github.com/flarum/core/issues/1951)

### DIFF
--- a/js/src/common/components/Page.js
+++ b/js/src/common/components/Page.js
@@ -30,6 +30,10 @@ export default class Page extends Component {
     if (this.bodyClass) {
       $('#app').addClass(this.bodyClass);
     }
+
+    if (!this.dontScrollTopOnCreate) {
+      window.scrollTo(0, 0);
+    }
   }
 
   onremove() {
--- a/js/src/common/components/Page.js
+++ b/js/src/common/components/Page.js
@@ -22,6 +22,13 @@ export default class Page extends Component {
      * @type {String}
      */
     this.bodyClass = '';
+
+    /**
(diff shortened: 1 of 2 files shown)


clarkwinkelmann (commented):
I'm globally ok with the change and I have tested it working locally, I just have some remarks:

Was there an attribute named `dontScrollTopOnCreate` in beta 13? If not and if that's a new parameter, I'd suggest naming it positively, like `scrollTopOnCreate` and default it to `true`.

Once `window.scrollTo(0, 0);` while the second time `$(window).scrollTop(0);` is used. Should we maybe unify this?

w-4:
Let's also move setting the `app.cache.scrollTop` from `onremove` to `onbeforeremove`.

clarkwinkelmann (approved):

askvortsov1:
> Let's also move setting the `app.cache.scrollTop` from `onremove` to `onbeforeremove`.

Good point! Not sure it practically makes a difference from my testing, but this way we have the assurance that we're calculating height before the DOM is detached.

tankerkiller125 (approved):

PR #2402, opened 2020-10-21
Fix DiscussionListPane jumping around (when going from discussion to discussion)

w-4:
**Fixes the following issue:**

1. Go to https://nightly.flarum.site and open a discussions.
2. On the discussion page scroll the pane a little. 
3. Click on a different discussion and the pane jumps back to the top, not keeping the scrolled position. 

**Confirmed**

### DIFF
--- a/js/src/forum/components/DiscussionListPane.js
+++ b/js/src/forum/components/DiscussionListPane.js
@@ -40,7 +40,7 @@ export default class DiscussionListPane extends Component {
     // When coming from another discussion, scroll to the previous postition
     // to prevent the discussion list jumping around.
     if (app.previous.matches(DiscussionPage)) {
-      const top = app.cache.paneScrollTop || 0;
+      const top = app.cache.discussionListPaneScrollTop || 0;
       $list.scrollTop(top);
     } else {
       // If the discussion we are viewing is listed in the discussion list, then
@@ -61,7 +61,7 @@ export default class DiscussionListPane extends Component {
   }
 
   onremove(vnode) {
-    app.cache.paneScrollTop = $(vnode.dom).scrollTop();
+    app.cache.discussionListPaneScrollTop = $(vnode.dom).scrollTop();
     $(document).off('mousemove', hotEdge);
   }
 
--- a/js/src/forum/components/DiscussionListPane.js
+++ b/js/src/forum/components/DiscussionListPane.js
@@ -40,6 +40,8 @@ export default class DiscussionListPane extends Component {
     // If the discussion we are viewing is listed in the discussion list, then
     // we will make sure it is visible in the viewport – if it is not we will
     // scroll the list down to it.
+    // When coming from another discussion, scroll to the previous postition
+    // to prevent the discussion list jumping around.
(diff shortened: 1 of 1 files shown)


js/src/forum/components/DiscussionListPane.js
```diff
@@ -39,20 +40,28 @@ export default class DiscussionListPane extends Component {
...
     // scroll the list down to it.
-    const $discussion = $list.find('.DiscussionListItem.active');
-    if ($discussion.length) {
-      const listTop = $list.offset().top;
-      const listBottom = listTop + $list.outerHeight();
-      const discussionTop = $discussion.offset().top;
-      const discussionBottom = discussionTop + $discussion.outerHeight();
+    // When coming from another discussion, scroll to the previous postition
+    // to prevent the discussion list jumping around.
+    if (!app.previous.matches(DiscussionPage)) {
```
askvortsov1, line 45: Can we swap this around so that the DiscussionPage special case comes first? That way we avoid negation in the conditional

js/src/forum/components/DiscussionListPane.js
```diff
@@ -39,20 +40,28 @@ export default class DiscussionListPane extends Component {
...
       }
+    } else {
+      const top = app.cache.paneScrollTop || 0;
+      $list.scrollTop(top);
     }
   }
 
-  onremove() {
+  onremove(vnode) {
+    app.cache.paneScrollTop = $(vnode.dom).scrollTop();
```
askvortsov1, line 64: Should this be in `onbeforeremove` to be consistent with IndexPage?

askvortsov1 (changes requested):

js/src/forum/components/DiscussionListPane.js
w-4: I did that initially, but `onbeforeremove` gets only called on the Page component, not in the child components. 
see: https://mithril.js.org/lifecycle-methods.html#onbeforeremove

askvortsov1 (approved):
Thank you!

clarkwinkelmann:
Is that `paneScrollTop` global variable exclusive to this particular feature? If so I think its name should reference the component that makes use of it, so we're certain it does not get used by something unrelated in the future or by extensions.

w-4:
I renamed the variable. Also, after implementing https://github.com/flarum/core/issues/2413 caching of the scroll position won't be necessary anymore.

clarkwinkelmann:
I haven't followed well enough. Will this still be necessary? Should I review again?

tankerkiller125 (approved):

askvortsov1:
@w-4 Just to clarify, you're saying that with flarum/issue-archive#176 the change in this PR won't be needed, right? (since we won't be rerendering the panel completely anyways). If so, maybe we shouldn't be merging this?

w-4:
@askvortsov1 Yes, only the if condition `if (!app.previous.isDiscussion())` would be needed. I'd merge it in, it fixes the bug, and flarum/issue-archive#176 can be done after stable.

PR #2408, opened 2020-10-24
Move locale files from language pack to core

rob006:
See https://github.com/flarum/lang-english/pull/175


locale/validation.yml
```diff
@@ -0,0 +1,108 @@
...
+
+  attributes:
+    username: username
+    password: password
+    email: email
+    title: title
+    content: content
+    name_singular: singular name
+    name_plural: plural name
+    tag_count_primary: number of primary tags
```
askvortsov1, line 107: Is there any way the tag-specific stuff could remain with tags?

askvortsov1 (changes requested):

locale/validation.yml
rob006: I don't think it matters where these strings are located. It may only be a problem if some extensions use these strings and tags extension is not enabled.

askvortsov1: As of right now, the only parts of core that are aware that tags even exists are:

- The installer, which enables bundled extensions
- Some examples in comments

I'd like to avoid expanding that to keep the division between core and extensions as strong as possible. I don't remember off the top of my head, but if we have 2 `validation.yml` files, one in core, the other in tags, would they be merged by the translator? If so, we should do that; if not, we may want to investigate supporting that.

rob006: For extension you need to put these translations in `en.yml` together with the rest of translations for this extension.  They should be merged in the same as any other translation.

BTW: Are your sure that this is still used? I can't find any occurrences of these keys on https://github.com/flarum/tags.

askvortsov1: Yeah that makes sense. Let's go for that if it works!

> BTW: Are you sure that this is still used? I can't find any occurrences of these keys on https://github.com/flarum/tags.

Validation translations aren't referenced directly, but rather used in Laravel validation error messages: If a discussion is submitted with too few primary or secondary tags, these error messages will be used.

rob006: > Validation translations aren't referenced directly, but rather used in Laravel validation error messages: If a discussion is submitted with too few primary or secondary tags, these error messages will be used.

But I don't see `tag_count_primary` attribute anywhere. If I try to start discussion with one primary tag while 4 is required, I get this:

668163b8

askvortsov1: Hmm, looks like we found a bug!

askvortsov1: Written up at https://github.com/flarum/core/issues/2572. Regardless, I think tag's validation translations should live in the tags repo.

rob006: I could remove them here, but IMO this should be done in separate PR.

BTW: I'm not sure if we should use `validation.attributes.name` keys outside of `validation.yml`. It will encourage extensions developers to do the same, and this is straight way to collisions.

askvortsov1: > I could remove them here, but IMO this should be done in separate PR.

When all extensions were in a single folder, it made sense to have a shared file for validation logic separated out, but since it's being split, tag-specific validation errors and attribute names should be located in the tags repo.

> BTW: I'm not sure if we should use validation.attributes.name keys outside of validation.yml. It will encourage extensions developers to do the same, and this is straight way to collisions.

Why would developers need to override that per-extension, and not just reply on the ones from core? I'm envisioning tag's validation translations being just those 2 attribute names.

rob006: > When all extensions were in a single folder, it made sense to have a shared file for validation logic separated out, but since it's being split, tag-specific validation errors and attribute names should be located in the tags repo.

I'm not sure how do you want to proceed with this. As I said, I can remove these keys in this PR. But I don't want to block this PR and wait another X months until you decide where these lines should be and what keys they should use (especially as they don't work anyway). Merging this as is and having 2 unnecessary keys should not be a big deal - it could be fixed later as part of #2572.

> Why would developers need to override that per-extension, and not just reply on the ones from core?

I'm not talking about overwriting core translations, I'm talking about collisions as a result of using the same namespace (`validation.attributes.`) for attributes names of extensions (because official extensions will do this, everyone can copy this pattern) - key `validation.attributes.name` could be used by two separate extensions unaware each other, and they may need to be translated differently depending on context (for example in Polish "name" could be translated either as "imię" or "nazwa"). To avoid this it should be something like `flarum-tags.validation.attributes.name`, but I'm not sure how problematic this would be with Laravel validator (it looks like black magic for me :/).

PR #2409, opened 2020-10-24
AdminUX Overhaul

### LINKED ISSUE #746: Revamp admin interface
**_1 Upvote**_ The Extensions page should be divided into categories so extensions will be easier to locate.

Possible categories to add:
- Authentication & Security
- Formatting (Markdown, BBcode, etc.)
- Language packs
- Themes
- Other

The Extension page will need to use composer.json tag info to determine which category an installed extension should be displayed under.

KyrneDev:
**Closes #746**

Completely redesign the admin dashboard and extension pages.

Extensions will be given a default page that looks similar to all other extension pages, it can be easily overridden by changing the component in the route. The main benefit of this is that currently extensions either have a modal or have a page on the side and it can sometimes be hard to know where to go to find the settings of each extension. This puts them all in a predictable and consistent location. It also helps by showing extension-specific permissions on their own page, removing the need to sift through the often very large permissions grid to find certain permissions.

It also allows extensions to be categorized (potentially adding custom categories option). It also has a backward compatibility layer for extensions that are using the settings modal at the moment.

Please check code quality (specifically the addExtensionPermissions class). There's bound to be something I missed. 
Also, I still need to make a PR to flarum/english for the new translation keys.

**Screenshot**
unknown
unknown-2

**Confirmed**

**Required changes:**

- [ ] Related core extension PRs: (English PR coming soon).


js/src/admin/AdminApplication.js
```diff
@@ -64,4 +70,14 @@ export default class AdminApplication extends Application {
 
     return required;
   }
+
+  setDefaultExtensionPages() {
```
askvortsov1, line 74: I'm still a bit uncertain about registering separate routes for each discussion, but as we discussed yesterday, that might be the best of our options for extensibility.

js/src/admin/components/AppearancePage.js
```diff
@@ -21,6 +22,15 @@ export default class AppearancePage extends Page {
   view() {
     return (
       <div className="AppearancePage">
+        <div className="AppearancePage-header">
```
askvortsov1, line 25: These headers could probably be extracted out into their own components, which the extension headers could maybe extend?

js/src/admin/components/ExtensionPage.js
```diff
@@ -0,0 +1,251 @@
...
+import Switch from '../../common/components/Switch';
+import saveSettings from '../utils/saveSettings';
+
+export default class ExtensionPage extends Page {
+  oninit(vnode) {
+    super.oninit(vnode);
+
+    this.settings = {};
+    this.loading = false;
+    this.extension = app.data.extensions[this.attrs.routeName];
```
askvortsov1, line 19: After thinking about this a bit more, I'm actually fine with something like this: regardless of how we get the extension ID, retrieving the extension object itself will work a bit like this.

js/src/admin/components/ExtensionPage.js
```diff
@@ -0,0 +1,251 @@
...
+    this.changingState = false;
+  }
+
+  className() {
+    return this.extension.id + '-Page';
+  }
+
+  view() {
+    new addExtensionPermission('flarum-flags')
```
askvortsov1, line 28: Is this leftover code?

js/src/admin/components/ExtensionPage.js
```diff
@@ -0,0 +1,251 @@
...
+                )}
+              </div>
+            </div>
+          </div>
+        )}
+      </div>
+    );
+  }
+
+  form() {
```
askvortsov1, line 103: Maybe this should be called content()?

js/src/admin/components/ExtensionPage.js
```diff
@@ -0,0 +1,251 @@
...
+    return '';
+  }
+
+  headerItems() {
+    const items = new ItemList();
+
+    items.add(
```
askvortsov1, line 110: This should only be added if disabled

js/src/admin/components/ExtensionPage.js
```diff
@@ -0,0 +1,251 @@
...
+import icon from '../../common/helpers/icon';
+import ItemList from '../../common/utils/ItemList';
+import listItems from '../../common/helpers/listItems';
+import LoadingModal from './LoadingModal';
+import ExtensionPermissionGrid from './ExtensionPermissionGrid';
+import addExtensionPermission from '../utils/addExtensionPermission';
+import Switch from '../../common/components/Switch';
+import saveSettings from '../utils/saveSettings';
+
+export default class ExtensionPage extends Page {
```
askvortsov1, line 13: How does this look when the extension is disabled? I suppose the form wouldn't show up then (since js isn't registered)?  Might be better to have a special message for that to justify the blank page.

js/src/admin/components/ExtensionPage.js
```diff
@@ -0,0 +1,251 @@
...
+                ) : app.extensionSettings[this.extension.id] ? (
+                  <Button onclick={app.extensionSettings[this.extension.id].bind(this)} className="Button Button--primary">
+                    {app.translator.trans('core.admin.extension.open_modal')}
+                  </Button>
+                ) : (
+                  <h2 className="ExtensionPage-subHeader">{app.translator.trans('core.admin.extension.no_settings')}</h2>
+                )}
+              </div>
+            </div>
+            <div className="ExtensionPage-permissions">
```
askvortsov1, line 85: Should this somehow be abstracted away to allow for new sections to be added?

js/src/admin/components/ExtensionPage.js
```diff
@@ -0,0 +1,251 @@
...
+  infoItems() {
+    const items = new ItemList();
+
+    if (this.extension.authors) {
+      console.log(this.extension);
+      let authors = [];
+      console.log(this.extension.authors);
+
+      Object.keys(this.extension.authors).map((author) => {
+        authors.push(this.extension.authors[author].name);
```
askvortsov1, line 142: Maybe these should be links to author websites / emails?

js/src/admin/components/ExtensionPage.js
```diff
@@ -0,0 +1,251 @@
...
+      });
+
+      items.add('authors', [icon('fas fa-user'), <span>{authors.join(', ')}</span>]);
+    }
+
+    if (this.extension.source || this.extension.support) {
```
askvortsov1, line 148: Feels like these 2 should be separate

js/src/admin/components/ExtensionPage.js
```diff
@@ -0,0 +1,251 @@
...
+      );
+    }
+
+    return items;
+  }
+
+  isEnabled() {
+    const enabled = JSON.parse(app.data.settings.extensions_enabled);
+
+    let isEnabled = enabled.indexOf(this.extension.id) !== -1;
```
askvortsov1, line 167: Lets use `.contains()`

js/src/admin/components/ExtensionPage.js
```diff
@@ -0,0 +1,251 @@
...
+
+    if (this.changingState) {
```
askvortsov1, line 169: I think I'd prefer for `changingState` to be used explicitly in places that refer to it instead of as part of `isEnabled()`.

js/src/admin/components/ExtensionPage.js
```diff
@@ -0,0 +1,251 @@
...
+      return isEnabled;
+    }
+  }
+
+  toggle() {
+    const enabled = this.isEnabled();
+
+    this.changingState = true;
+
+    app
```
askvortsov1, line 181: This should include the dependency error handlers from #2188

js/src/admin/components/ExtensionPage.js
```diff
@@ -0,0 +1,251 @@
...
+
+  submitButton() {
+    return (
+      <Button onclick={this.saveSettings.bind(this)} className="Button Button--primary" loading={this.loading} disabled={!this.changed()}>
+        {app.translator.trans('core.admin.settings.submit_button')}
+      </Button>
+    );
+  }
+
+  // Simple version to allow copy
```
askvortsov1, line 203: Not sure we need this comment

js/src/admin/utils/addExtensionPermission.js
```diff
@@ -0,0 +1,26 @@
+import ItemList from '../../common/utils/ItemList';
+
+export default class addExtensionPermission {
```
askvortsov1, line 3: Do we want this to be a class?

js/src/admin/utils/addExtensionPermission.js
```diff
@@ -0,0 +1,26 @@
...
+  constructor(extension) {
+    this.extension = extension;
+    app.extensionPermissions[extension] = {};
```
askvortsov1, line 6: Not sure if we want to completely clear it

js/src/common/components/SelectDropdown.js
```diff
@@ -12,6 +12,8 @@ import icon from '../helpers/icon';
 function isActive(vnode) {
   const tag = vnode.tag;
 
+  if (typeof tag === 'string') return false;
```
askvortsov1, line 15: Why?

js/src/admin/components/AdminNav.js
```diff
@@ -1,106 +1,197 @@
...
+    $('.SearchBar').on('keyup', () => {
+      const filter = $('.SearchBar').val().toUpperCase();
+      const list = $('.Dropdown-menu');
+      if (!filter) {
+        list.children('li').show();
+      } else {
+        list.children('li').map((key) => {
+          const element = $(list.children('li')[key]);
+          if (
+            (!element
```
askvortsov1, line 104: Perhaps we should also allow searching by descriptions?

askvortsov1 (changes requested):
Another suggestion brought up by David that I really agree with is that it would be nice to keep a page with all extensions (simple tiles?) as an "overview" page. Perhaps we could use David's admin dash design, or revert to the pre-beta7 design for simplicity.

KyrneDev:
@askvortsov1 and I had a call yesterday to discuss this briefly and came up with 3 main options on how to do the extension pages, specifically the routes..

Option 1: Each extension is registered to its own route - easy to override and change - but it is a lot of routes, feels clunky.

Option 2: Route resolver routes by default for extensions that don’t have/need settings
Extensions that need to add basic settings/permission can use a util to register this data, or if they want a entirely custom page they can register a route. Use route resolver, check in route resolver if page is registered -> skip, don’t handle.

Option 3: Route resolver for all exts, ext can register entirely new component (new/extend), or provide array of components (JSX).

We were leaning towards #2 but wanted to discuss it with everyone. For the util I'm working on a concept, it's looking like a frontend extender chain with a construct argument that takes the ext id (flarum-flags).

luceos:
If in #2 you can still add content below the default content (permissions and util-generated simple settings), I'd be fine with it. But I do want to see a concept of such code from the perspective of a dev so that we can review whether this is acceptable enough..

askvortsov1:
As per a recent dev meeting, current thinking is to get this included as early as possible, so we can dogfood it on our local installs during the release. We'd also like to have some permissions overview page, although that might be removed if we find it to be redundant with the sidebar list.

KyrneDev:
@luceos This is what I currently have for registering permissions (globally & for the page) + simple settings

(flarum/flags)

``` js
new extensionData('flarum-flags')
    .registerData('settings', {
        'flarum-flags.guidelines_url': {
            type: 'text',
            label: app.translator.trans('flarum-flags.admin.settings.guidelines_url_label')
        },
        'flarum-flags.can_flag_own': {
            type: 'boolean',
            label: app.translator.trans('flarum-flags.admin.settings.flag_own_posts_label')
        }
    })
    .registerData('permissions', {
        icon: 'fas fa-flag',
        label: app.translator.trans('flarum-flags.admin.permissions.view_flags_label'),
        permission: 'discussion.viewFlags'
    }, 'moderateItems', 65)

    .registerData('permissions', {
        icon: 'fas fa-flag',
        label: app.translator.trans('flarum-flags.admin.permissions.flag_posts_label'),
        permission: 'discussion.flagPosts'
    }, 'replyItems', 65);
```

askvortsov1:
Would these maintain ordering of permissions and settings?

KyrneDev:
> Would these maintain ordering of permissions and settings?

The settings are in the order of the object. The permissions are an item list and have an optional priority - just like current behavior.

KyrneDev:
@askvortsov1 / @luceos It's mostly ready for test installs. It has all backward compatibility layers built-in. First extension has also been converted: https://github.com/flarum/flags/pull/33

less/admin/ExtensionWidget.less
```diff
@@ -0,0 +1,58 @@
...
+  padding: 0;
+  list-style: none;
+  .clearfix();
+
+  > li {
+    float: left;
+    text-align: left;
+    position: relative;
+  }
+}
```
SychO9, line 32: The extension widget gets very messy on different screen sizes. I would really prefer we avoid floats and instead go for flexbox (https://caniuse.com/mdn-css_types_flex)

It makes the list a lot more resilient and dynamic, while keeping it nice looking.

locally this is what I had to edit (it's in css but.. yeah)
```css
.ExtensionsWidget-title {
    color: #667c99;
    margin-top: 0;
}
.ExtensionList {
    margin: 0 -6px;
    padding: 0;
    list-style: none;
    display: flex;
    flex-wrap: wrap;
}
.ExtensionList>li {
    text-align: left;
    position: relative;
    display: block;
}
.ExtensionListItem {
    width: 90px;
    height: 90px;
    margin: 6px;
}
```

PS: also had to drop the ::before and ::after pseudo elements from `.ExtensionList`

less/admin/ExtensionWidget.less
```diff
@@ -0,0 +1,58 @@
...
+.ExtensionIcon {
+  width: 90px;
+  height: 90px;
+  background: @control-bg;
+  color: @control-color;
+  border-radius: 6px;
+  display: inline-block;
+  font-size: 45px;
+  line-height: 90px;
+  text-align: center;
```
SychO9, line 57: Let's also align the icon with flexbox! (it makes maintaining code and making design changes a lot easier in the long run)

```suggestion
  display: inline-flex;
  font-size: 45px;
  text-align: center;
  align-items: center;
  justify-content: center;
```

less/admin/ExtensionPage.less
```diff
@@ -0,0 +1,120 @@
...
+      height: 25px;
+      background: @control-bg;
+      color: @control-color;
+      border-radius: 4px;
+      display: inline-block;
+      font-size: 12.5px;
+      line-height: 25px;
+      text-align: center;
+      position: absolute;
+      margin-left: -35px;
```
SychO9, line 87: Because `.ExtensionIcon` has already been declared and can be used globally, most rules here don't need to be redeclared, the only additions/modifications seem to be with `width` `height` `position` `margin-left` `font-size`

On another note, why do we have to use an absolute position here ? In most cases we can accomplish anything without having to resort to it. (makes maintaining the code and future design changes hard compared to a more straightforward and simple approach)

Also the icon in the extension page is very small, we could probably up that to around 40px.

less/admin/AdminNav.less
```diff
@@ -20,59 +86,88 @@
...
+          height: 25px;
+          background: @control-bg;
+          color: @control-color;
+          border-radius: 4px;
+          display: inline-block;
+          font-size: 15px;
+          line-height: 25px;
+          text-align: center;
+          margin: -2px -29px;
+          position: absolute;
```
SychO9, line 167: same here

less/admin/AdminNav.less
```diff
@@ -1,17 +1,83 @@
...
```
SychO9, line 69: same here, we don't have to redeclare everything, just override and add rules.

less/admin/AdminNav.less
```diff
@@ -86,3 +181,26 @@
     }
   }
 }
+
+.ExtensionListItem-Dot {
```
SychO9, line 185: The menu nicely takes care of (if overflow => ellipsis) cases, but with the nice Dot to visually show if an extension is enabled/disabled, it covers the text. It would be nice to have somewhere something like:

```css
.ExtensionNavButton .Button-label {
  display: inline-block;
  max-width: calc(100% - 18px);
  overflow: hidden;
  text-overflow: ellipsis;
}
```

less/admin/AdminNav.less
```diff
@@ -1,17 +1,83 @@
-@admin-pane-width: 300px;
+@admin-pane-width: 210px;
```
SychO9, line 1: I think this is too tight, 300 seems to actually be the perfect size, but if we could at least go above 250, otherwise 210 is really tight.

SychO9 (commented):
Amazing work! I love it!

When it comes to web design, my spider senses start tingling, so sorry if my comments are a little too much :laughing: (just some suggestions)

less/admin/ExtensionWidget.less
askvortsov1: Even better, could we go for CSS grid?

less/admin/AdminNav.less
askvortsov1: Agreed, some extensions with longer names are cut off.

askvortsov1:
Also, it looks like the logic for handling dependency error messages is missing

less/admin/ExtensionWidget.less
SychO9: oh yes! CSS grid would most likely be much better, something like ?

```css
.ExtensionList {
    display: grid;
    grid-gap: 10px;
    grid-template-columns: repeat(auto-fit, 90px);
}
```

KyrneDev:
> Also, it looks like the logic for handling dependency error messages is missing. Since stuff isn't on the same page anymore, I think the extension IDs of each extension in the alert should link to the relevant extension page.

Right, forgot this was added, I'll put it in

> Another thought: The Flarum native admin sections feel a bit underwhelming in the navbar. Perhaps the text should be a bit bigger? Also, maybe there should be a divider for extensions, or maybe search should be at the top?

Maybe a small font size increase and bold? I like the extensions divider idea.

> Also on my local version, ALL extensions that aren't languages are grouped into the "other" category.

This is because categories are pulled from the composer.json:

```json
"flarum-extension": {
            "title": "Flags",
            "category": "moderation",
            "icon": {
                "name": "fas fa-flag",
                "backgroundColor": "#D659B5",
                "color": "#fff"
            }
        }
```

js/src/admin/components/AdminNav.js
```diff
@@ -1,106 +1,199 @@
...
+      <div className="Search-input">
+        <input className="FormControl SearchBar" type="search" placeholder={app.translator.trans('core.admin.nav.search_placeholder')} />
+      </div>
     );
 
     return items;
   }
+
+  oncreate(vnode) {
+    $('.SearchBar').on('keyup', () => {
```
askvortsov1, line 95: Does this need to be done via jquery? If we could conditionally display things and redraw, that might be preferable.

js/src/admin/components/ExtensionSetting.js
```diff
@@ -0,0 +1,30 @@
+import Component from '../../common/Component';
+import classList from "../../common/utils/classList";
+import Stream from "../../common/utils/Stream";
+import withAttr from "../../common/utils/withAttr";
+
+export default class ExtensionSetting extends Component {
```
askvortsov1, line 6: This doesn't currently seem to be used?

js/src/admin/components/ExtensionsWidget.js
```diff
@@ -0,0 +1,42 @@
...
+        <div className="container">
+          <h2 className="ExtensionsWidget-title">{app.translator.trans('core.admin.dashboard.extensions_title')}</h2>
+          <ul className="ExtensionList">
+            {Object.keys(app.data.extensions).map((id) => {
+              const extension = app.data.extensions[id];
+
+              return (
+                <li className={'ExtensionListItem ' + (!this.isEnabled(extension.id) ? 'disabled' : '')}>
+                  <Link href={app.route('extension', { id: extension.id })}>
+                    <div className="ExtensionListItem-content">
```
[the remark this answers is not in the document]
askvortsov1: The icons themselves look nice, but are a bit confusing without a title

askvortsov1 (changes requested):

less/admin/AppearancePage.less
```diff
@@ -34,3 +35,200 @@
     line-height: 1;
   }
 }
+.StatisticsWidget-table {
```
askvortsov1, line 38: I don't think this all should be here: this seems like statistics extension stuff

less/admin/DashboardPage.less
```diff
@@ -1,18 +1,18 @@
 .DashboardPage {
-  background: @control-bg;
+  background: @body-bg;
   color: @control-color;
   min-height: 100vh;
 
   @media @desktop-up {
-    .container {
+    > .container {
       padding: 30px;
```
askvortsov1, line 8: This padding is now inconsistent with that of other pages

askvortsov1 (changes requested):

less/admin/DashboardPage.less
KyrneDev: I'm experiencing the opposite, with the arrow it is consistent, without, it is extra big.

askvortsov1: It's not the arrow as much as the 30px: that 30px is unnecessarily applied to the top.

KyrneDev: Right, I see now

less/admin/ExtensionWidget.less
```diff
@@ -0,0 +1,86 @@
...
+      h3 {
+        color: @muted-color;
+        text-transform: uppercase;
+        font-size: 12px;
+        margin: 0 0 10px;
+      }
+    }
+
+    .ExtensionList {
+      margin: 0 -6px;
```
SychO9, line 35: Since we decided to go for grid, we can drop these margins that I had suggested with flex

less/admin/ExtensionWidget.less
```diff
@@ -0,0 +1,86 @@
...
+
+    .ExtensionListItem-icon {
+      opacity: 0.5;
+    }
+  }
+
+  .ExtensionListItem {
+    width: 90px;
+    height: 90px;
+    margin: 6px;
```
SychO9, line 64: Same here, we can remove the margins since we're using grid

Also we should remove the height, since we are now showing the extension name

SychO9: Also let's set `text-decoration` to none when the anchor inside the `ExtensionListItem` is hovered, it doesn't look good with underline.

Which makes me wonder if we want to add any kind of hovering effect, not sure yet.

KyrneDev: I dropped the width & margin but left in the height because it tightens it up, reduces the extra spacing. For the hovering effect, I added a small scale increase of the icon and title.

SychO9: the problem with setting the height is the text is hidden though

flarum lan_admin (1)

askvortsov1:
It's looking very nice, except that the titles are overlapping into the images below. https://i.imgur.com/gqg03bk.png

Also, extensions without an author / source have the bottom of the header a bit too close to the switch: https://i.imgur.com/jJvSVNQ.png vs https://i.imgur.com/jJvSVNQ.png
I wonder if it might be better if we bring all that metadata into the same row as the switch, but to the right like in https://github.com/flarum/core/issues/746#issuecomment-267794704.  The version could also be shifted to the right IMO to match that (and reduce the big block of whitespace a bit)

askvortsov1:
Another thought: Perhaps "Other" should be called "Other Extensions"?

askvortsov1:
The author / source buttons / text are a bit cramped: https://i.imgur.com/QXiHZCP.png
Also, perhaps buttons (like Uninstall, although not Source) should use an actual button? And for uninstall, when clicked, I think it should show a JS confirm popup that says that uninstalling will run down all migrations and unpublish assets.

Looks like the sections in the dashboard are in reverse order: https://i.imgur.com/Vij2Aoj.png

Also, should links to other stuff (Support thread on Discuss, optional documentation, etc be added? We could support other optional composer parameters).

Also, I like in https://github.com/flarum/core/issues/746#issuecomment-267794704 the "Get Help" button at the top that links to docs, I think that might be nice.

js/src/admin/components/ExtensionPage.js
```diff
@@ -0,0 +1,347 @@
...
+    );
+  }
+
+  content() {
+    const settings = this.getContent('settings');
+
+    if (settings) {
+      const items = new ItemList();
+
+      Object.keys(settings).map((key) => {
```
askvortsov1, line 56: I think this would be great as a util, so it's easier to use standard controls in custome extension pages

askvortsov1 (changes requested):

js/src/admin/components/ExtensionPage.js
KyrneDev: This problem with that is we need this to be in the context of the extensions page. The can easily use that still by just extending the base extension page.

askvortsov1: Maybe as a method then?

KyrneDev: I think we need to pick whether to combine everything into content or to split the settings section and settings content. We are getting to extendable levels that aren't necessary, adding many lines of code to save developers literally 1 line.

KyrneDev:
> Also, should links to other stuff (Support thread on Discuss, optional documentation, etc be added? We could support other optional composer parameters).

My list:
* Thread on Discuss
* Documentation
* Website
* Support link

Any more ideas?

js/src/admin/components/ExtensionPage.js
askvortsov1: I think the content method should return jsx. That means that if someone is replacing content entirely but still wants to use this útil for converting settings descriptors to actual components they still could.

askvortsov1:
Possibly a donate link?

js/src/admin/components/ExtensionPage.js
KyrneDev: So maybe move the util to a different function in the settings page?

js/src/admin/components/PermissionGrid.js
```diff
@@ -158,12 +163,18 @@ export default class PermissionGrid extends Component {
...
 
+    items.merge(this.getExtensionPermissions('view'));
+
     return items;
   }
 
   startItems() {
     const items = new ItemList();
 
+    if (this.extensionId) {
```
askvortsov1, line 174: Is there anyway all this logic could be moved into ExtensionPermissionGrid? I don't really like PermissionGrid itself being aware of a current extension

js/src/admin/components/PermissionGrid.js
```diff
@@ -366,4 +391,20 @@ export default class PermissionGrid extends Component {
   scopeControlItems() {
     return new ItemList();
   }
+
+  getExtensionPermissions(type) {
```
askvortsov1, line 395: Again, would be nice for this to be part of some ExtensionManager util or something, so it's not duplicated across this and ExtensionPermissionGrid, and so that the PermissionGrids don't need to concern themselves with the implementation of how extensions register / store permissions

js/src/admin/utils/ExtensionData.js
```diff
@@ -0,0 +1,34 @@
+import ItemList from '../../common/utils/ItemList';
+import listItems from '../../common/helpers/listItems';
+
+export default class ExtensionData {
```
askvortsov1, line 4: This needs docblocks.

js/src/admin/utils/ExtensionData.js
```diff
@@ -0,0 +1,34 @@
...
+  constructor(extension) {
+    this.extension = extension;
+    app.extensionData[extension] = app.extensionData[extension] || {};
+  }
+
+  registerData(type, content, permissionType = null, priority = 0) {
```
askvortsov1, line 10: IMO, this would be better if split into a `registerPermission` and `registerSetting` methods.

js/src/admin/utils/ExtensionData.js
```diff
@@ -0,0 +1,34 @@
...
+      }
+      app.extensionData[this.extension][type][permissionType].add(content.permission, content);
+    } else {
+      app.extensionData[this.extension][type] = content;
+    }
+
+    return this;
+  }
+
+  registerPage(component) {
```
askvortsov1, line 29: Hmm, should we store this as a registered route, or perhaps we could just maintain a map of extension IDs to custom page components. That could work just as well for the route resolver, and would minimize any overhead in routes

js/src/admin/utils/ExtensionData.js
askvortsov1, line 4: As per some of my previous comments, I think this should work in tandem with a class that encapsulates storage (and retrieval) of all these pages, permissions, and settings.

js/src/admin/utils/getCategorizedExtensions.js
```diff
@@ -0,0 +1,29 @@
...
+      category = 'other';
+    }
+
+    // Wrap languages packs into new system
+    if (extension.extra['flarum-locale']) {
+      category = 'language';
+    }
+
+    if (category in app.extensionCategories) {
+      extensions[category] = extensions[category] || {};
```
askvortsov1, line 18: Why implement the list of extensions for each category as an object instead of an array?

js/src/admin/utils/getCategorizedExtensions.js
```diff
@@ -0,0 +1,29 @@
+export default function getCategorizedExtensions() {
+  let extensions = {};
+
+  Object.keys(app.data.extensions).map((id) => {
+    const extension = app.data.extensions[id];
+    let category = extension.extra['flarum-extension'].category;
+
+    if (!extension.extra['flarum-extension'].category) {
```
askvortsov1, line 8: Is this needed? If it's null, null wouldn't be in `app.extensionCategories`, so wouldn't it go to `other` regardless?

js/src/admin/utils/getCategorizedExtensions.js
```diff
@@ -0,0 +1,29 @@
...
+      category = 'other';
+    }
+
+    // Wrap languages packs into new system
```
askvortsov1, line 12: This should be clarified to explain what "new system" means. "All language packs should be classified as such, regardless of their direct classification"?

js/src/admin/utils/getCategorizedExtensions.js
```diff
@@ -0,0 +1,29 @@
...
+    if (extension.extra['flarum-locale']) {
+      category = 'language';
+    }
+
+    if (category in app.extensionCategories) {
+      extensions[category] = extensions[category] || {};
+
+      extensions[category][id] = extension;
+    } else {
+      // If the extension doesn't fit
```
askvortsov1, line 22: This comment isn't necessary IMO

js/src/common/components/SelectDropdown.js
askvortsov1, line 15: I maintain my question of why? Not necessarily opposed to the change, but it needs a comment explaining it

js/src/admin/components/AdminHeader.js
```diff
@@ -0,0 +1,21 @@
...
+    const attrs = Object.assign({}, this.attrs);
+
+    return [
+      <div className={classList(['AdminHeader', attrs.className])}>
+        <div className="container">
+          <h2>
+            {icon(attrs.icon)}
+            {vnode.children}
+          </h2>
+          <div className="helpText">{attrs.description}</div>
```
askvortsov1, line 16: Looks like this was mostly used in various labels and such. Maybe use something like AdminHeader-description?

less/admin/AdminHeader.less
```diff
@@ -0,0 +1,19 @@
...
+    margin-top: 0;
+    margin-bottom: 10px;
+    color: @muted-color;
+  }
+
+  .helpText {
+    margin: 0;
+  }
+
+  i {
```
askvortsov1, line 16: Is this for the icon? If so, I'd prefer some semantic selectors based on the icon class

less/admin/AdminNav.less
```diff
@@ -1,17 +1,92 @@
-@admin-pane-width: 300px;
+@admin-pane-width: 250px;
 
 .App {
   padding-bottom: 0;
 }
-.AdminLinkButton-description {
-  display: none;
+.App-content {
```
askvortsov1, line 6: This doesn't seem to have a significant effect. Is it needed? (the addition I mean, maybe remove it entirely?)

less/admin/AdminNav.less
```diff
@@ -1,17 +1,92 @@
...
+      margin-left: 30px;
+    }
+    .ExtensionIcon {
+      margin: 0 0 0 -4px !important;
+    }
+  }
+}
+
+@media @tablet {
+  .ExtensionItem, .item-search {
```
askvortsov1, line 52: This directive seems to appear twice between here and line 22.

less/admin/AdminNav.less
```diff
@@ -1,17 +1,92 @@
...
+    display: none !important;
+  }
+}
+
+@media @phone, @tablet {
+  .App-nav .AdminNav {
+    .Dropdown-menu {
+      .NavDivider {
+        display: none;
```
askvortsov1, line 61: Why?

less/admin/AdminNav.less
```diff
@@ -1,17 +1,92 @@
...
+      }
+      > li {
+        .ExtensionListTitle {
+          color: @muted-color;
+          text-transform: uppercase;
+          margin: 25px 0 10px 15px;
+        }
+
+        .ExtensionListItem-Dot {
+          display: none;
```
askvortsov1, line 71: Again, why? On most phones this would be nice to have IMO

less/admin/AppearancePage.less
```diff
@@ -1,8 +1,9 @@
 .AppearancePage {
+
   @media @desktop-up {
     .container {
       max-width: 600px;
-      padding: 30px;
+      padding: 0 30px;
```
askvortsov1, line 6: Why is this needed here but not on other pages?

less/admin/AdminHeader.less
```diff
@@ -0,0 +1,19 @@
+.AdminHeader {
```
askvortsov1, line 1: The text seems crammed to the left on mobile for some pages

askvortsov1 (changes requested):

js/src/admin/components/ExtensionPage.js
KyrneDev: I'll remove the support part but it isn't possible to move this since the data is in a different location in composer. Unless we want developers to specify the source a second time in composer.json

js/src/admin/utils/getCategorizedExtensions.js
KyrneDev: That is for language packs that don't get updated with the category field. I'll add an extra check.

less/admin.less
```diff
@@ -1,10 +1,12 @@
 @import "common/common";
 
+@import "admin/AdminHeader";
 @import "admin/AdminNav";
 @import "admin/DashboardPage";
 @import "admin/BasicsPage";
 @import "admin/PermissionsPage";
 @import "admin/EditGroupModal";
-@import "admin/ExtensionsPage";
```
SychO9, line 8: does this mean we can safely delete the `ExtensionsPage.less` file ?

less/admin/ExtensionPage.less
```diff
@@ -0,0 +1,143 @@
+@extension-list-column-width: 410px;
```
SychO9, line 1: seems unneeded

less/admin/ExtensionPage.less
```diff
@@ -0,0 +1,143 @@
...
+        color: @muted-color;
+        font-weight: normal;
+      }
+    }
+
+    button {
+      i {
+        display: unset;
+      }
+    }
```
SychO9, line 37: best to use classes where possible, `.Button-icon` would work here

this is a good case of why the primary button weirdly hides the icon lol, something we'll need to fix in the future.

less/admin/ExtensionPage.less
```diff
@@ -0,0 +1,143 @@
...
+
+    .ExtensionInfo {
+      float: right;
+    }
+
+
+      .ExtensionName {
+        display: inline-block;
+        margin-left: 8px;
+      }
```
SychO9, line 86: indentation seems off here

less/admin/AdminNav.less
```diff
@@ -1,17 +1,92 @@
...
+        }
+
+        .ExtensionIcon {
+          margin: -2px -29px;
+          width: 25px !important;
+          height: 25px !important;
+          font-size: 12.5px !important;
```
SychO9, line 78: `!important` seems unnecessary here, best to avoid using it anyway lol

SychO9 (changes requested):

SychO9:
slight suggestion, the information on the extension page seems to be a bit scattered around, I think it'd be nice to just align them all under the extension name, something like this imho would be neat and simple (including the source link and any other links):

flarum lan_admin (2)

askvortsov1:
Personally I prefer the current layout, although I think header size is broken? https://i.imgur.com/kvJqgL2.png

js/src/common/components/SelectDropdown.js
```diff
@@ -12,6 +12,7 @@ import icon from '../helpers/icon';
 function isActive(vnode) {
   const tag = vnode.tag;
 
+  // Allow non-selectable dividers/headers to be added.
   if (typeof tag === 'string') return false;
```
askvortsov1, line 16: Why couldn't a `<a></a>` HTML element be active?

js/src/common/helpers/fullTime.js
```diff
@@ -0,0 +1,19 @@
+/**
```
askvortsov1, line 1: Where is this coming from?

js/src/common/helpers/humanTime.js
```diff
@@ -0,0 +1,23 @@
+import humanTimeUtil from '../utils/humanTime';
```
askvortsov1, line 1: And where is this coming from?

js/src/common/utils/humanTime.ts
```diff
@@ -1,6 +1,3 @@
-import dayjs from 'dayjs';
```
askvortsov1, line 1: And where is this coming from?

js/src/common/compat.js
```diff
@@ -19,7 +19,6 @@ import extract from './utils/extract';
 import ScrollListener from './utils/ScrollListener';
 import stringToColor from './utils/stringToColor';
 import subclassOf from './utils/subclassOf';
-import SuperTextarea from './utils/SuperTextarea';
```
askvortsov1, line 22: Why?

askvortsov1 (changes requested):

js/src/admin/components/ExtensionPermissionGrid.js
```diff
@@ -1,44 +1,38 @@
...
+  }
+
   permissionItems() {
-    const items = super.permissionItems();
+    const permissionCategories = super.permissionItems();
 
-    Object.keys(items.items).map((item) => {
-      if (items.items[item].content.children.length === 0) {
-        items.remove([item]);
+    Object.keys(permissionCategories.items).map((item) => {
```
askvortsov1, line 14: I wonder if this could be done with spread syntax or `.entries()` and `Object.fromEntries()` so that we could use `fiilter`

js/src/admin/utils/ExtensionData.js
```diff
@@ -1,34 +1,159 @@
...
+   * })
+   *
+   *
+   * @param type
+   * @param content
+   * @param permissionType
+   * @param priority
+   * @returns {ExtensionData}
+   */
+  registerSettings(content, permissionType = null, priority = 0) {
```
askvortsov1, line 45: If we're registering settings by key, I think it might be best to do it in individual calls of `registerSetting`. One advantage of keeping it as `registerSettings` is that we could pass in an **ordered** array of settings instead of an unordered object.

js/src/admin/utils/ExtensionData.js
```diff
@@ -1,34 +1,159 @@
...
+   * This function simply takes the extension id
+   *
+   * @example
+   * app.extensionData.load('flarum-tags')
+   *
+   * flarum/flags -> flarum-flags | acme/extension -> acme-extension
+   *
+   * @param extension
+   */
+  load(extension) {
```
askvortsov1, line 19: Maybe it should be called `for` instead of `load`?

`app.extensionData.for('kyrne-shout').addPermission(...)`

js/src/admin/utils/ExtensionData.js
```diff
@@ -1,34 +1,159 @@
...
+      this.data[this.currentExtension].permissions[permissionType] = new ItemList();
     }
 
+    this.data[this.currentExtension].permissions[permissionType].add(content.permission, content);
+
     return this;
   }
 
+  /**
+   * Replace the default extension page with a custom component
```
askvortsov1, line 82: Add something like "this component would typically extend `ExtensionPage`"

js/src/admin/utils/ExtensionData.js
```diff
@@ -1,34 +1,159 @@
...
+
+  /**
+   *
+   * Get an ItemList of all extensions' registered permissions
+   *
+   * @param extension
+   * @param type
+   * @returns {ItemList}
+   */
+  getPermissions(type) {
```
askvortsov1, line 115: `getAllExtensionPermissions`?

js/src/admin/utils/ExtensionData.js
```diff
@@ -1,34 +1,159 @@
...
+  }
+
+  /**
+   * Get a singular extension's registered permissions
+   *
+   * @param extension
+   * @param type
+   * @returns {boolean|*}
+   */
+  getExtensionPermissions(extension, type) {
```
askvortsov1, line 134: Why boolean? What would true/false signify? Shouldn't this also return an ItemList?

js/src/admin/utils/getCategorizedExtensions.js
```diff
@@ -5,23 +5,17 @@ export default function getCategorizedExtensions() {
     const extension = app.data.extensions[id];
     let category = extension.extra['flarum-extension'].category;
 
-    if (!extension.extra['flarum-extension'].category) {
-      category = 'other';
-    }
-
     // Wrap languages packs into new system
-    if (extension.extra['flarum-locale']) {
+    if (extension.extra['flarum-locale'] && !category) {
```
askvortsov1, line 9: Maybe this should be applied to ALL language packs regardless of what they indicate as their category? Not 100% sure though

askvortsov1 (changes requested):

js/src/admin/components/ExtensionPage.js
```diff
@@ -0,0 +1,372 @@
...
+              {this.extension.icon ? icon(this.extension.icon.name) : ''}
+            </span>
+            <div className="ExtensionName">
+              <label>{this.extension.extra['flarum-extension'].title}</label>
+            </div>
+            <div className="ExtensionPage-headerTopItems">
+              <ul>{listItems(this.topItems().toArray())}</ul>
+            </div>
+          </div>
+          <div className="helpText">{this.extension.description}</div>
```
askvortsov1, line 74: Should this also be changed to AdminHeader-helpText?

js/src/admin/components/ExtensionPage.js
askvortsov1: Hmm I'm still not a fan of the actual LinkButton computation being repeated twice. Could we:

- Start by getting an array/object of the infoFields stuff
- Then, add support/source to that array/object
- Then, run the loop at line 196 over that array/object?

js/src/admin/components/ExtensionPage.js
```diff
@@ -0,0 +1,372 @@
...
+   *        'option2': 'Option 2 label',
+   *      },
+   *      default: 'option1',
+   *   }
+   * }
+   *
+   * @param settings
+   * @returns {JSX.Element}
+   */
+  getSettings(settings) {
```
askvortsov1, line 255: I still have mixed feelings about storing settings as an object. See my comment on `ExtensionData` for more info.

askvortsov1 (changes requested):
This is looking really great, I love the new API for ExtensionData!

js/src/admin/utils/getCategorizedExtensions.js
KyrneDev: That would make sense to me, not sure why they would be anything else.

js/src/admin/utils/ExtensionData.js
askvortsov1: To clarify, I think that settings registration would make a bit more sense as an array, since that way extension settings can be added in a certain order. Alternatively, settings could be added individually (instead of as one big object), which would also allow ordered registration of settings

js/src/admin/components/ExtensionPage.js
KyrneDev: No, because it's not apart of the Admin header, this is technically the extension header

js/src/admin/components/ExtensionPage.js
```diff
@@ -178,39 +178,40 @@ export default class ExtensionPage extends Page {
       items.add('authors', [icon('fas fa-user'), <span>{authors.join(', ')}</span>]);
     }
 
+    const infoData = {};
+
+    if (this.extension.extra['flarum-extension'].info) {
+      Object.keys(this.infoFields).map((field) => {
```
askvortsov1, line 184: Maybe something like `Object.keys().filter().forEach` to be a bit more concise?

js/src/admin/components/ExtensionPage.js
```diff
@@ -178,39 +178,40 @@ export default class ExtensionPage extends Page {
       items.add('authors', [icon('fas fa-user'), <span>{authors.join(', ')}</span>]);
     }
 
+    const infoData = {};
+
+    if (this.extension.extra['flarum-extension'].info) {
```
askvortsov1, line 183: Not sure if we need this wrapper: it doesn't save that much performance, and it adds an extra layer of nesting

js/src/admin/components/ExtensionPage.js
```diff
@@ -178,39 +178,40 @@ export default class ExtensionPage extends Page {
...
+    Object.entries(infoData).map(([field, value]) => {
       items.add(
-        'source',
+        field,
         LinkButton.component(
           {
-            href: this.extension.source ? this.extension.source.url : this.extension.support.source,
-            icon: 'fas fa-code',
+            href: value.href,
+            icon: value.href,
```
askvortsov1, line 207: value.icon?

js/src/admin/components/ExtensionPage.js
```diff
@@ -253,28 +254,29 @@ export default class ExtensionPage extends Page {
    * @returns {JSX.Element}
    */
   getSettings(settings) {
```
askvortsov1, line 256: docblock should also be updated

askvortsov1 (changes requested):
I think we're approaching my last few minor nitpicks

js/src/admin/components/ExtensionPage.js
KyrneDev: It would need to be checked each time in the loop. Is that ok?

js/src/admin/utils/getCategorizedExtensions.js
```diff
@@ -0,0 +1,23 @@
...
+    if (extension.extra['flarum-locale']) {
+      category = 'language';
+    }
+
+    if (category in app.extensionCategories) {
+      extensions[category] = extensions[category] || [];
+
+      extensions[category].push(extension);
+    } else {
+      extensions.other.push(extension);
```
askvortsov1, line 18: `extensions.other` needs to be instantiated; I'm getting an error where it's undefined

js/src/admin/components/ExtensionPage.js
askvortsov1: You also need to add `source` to `this.infoItems`. But it's a question of looping over 6 items vs not looping over 6 items. And performance-wise, that's extremely trivial IMO

askvortsov1 (changes requested):

askvortsov1 (approved):
Looks good! Will test out locally one more time before merging (when we get that 2nd approval), but this is perfect. Thank you so much!

Next steps could be:

- Ensuring all bundled extensions with settings are updated
- (possibly) adding a section to the extend documentation on how to set up admin pages / register settings / permissions for extensions

js/src/admin/components/ExtensionPage.js
```diff
@@ -0,0 +1,387 @@
...
+   *        'option2': 'Option 2 label',
+   *      },
+   *      default: 'option1',
+   *   }
+   * ]
+   *
+   * @param settings
+   * @returns {JSX.Element}
+   */
+  getSettings(settings) {
```
askvortsov1, line 269: Ah, one more idea. In `registerSettings`, providing them as an array makes sense. Or actually, maybe that's too much too since a priority can be provided. But here, I think this method would make more sense if called on individual settings (and have the `map` be in the code calling this method).

SychO9 (changes requested):
Few more comments,

Also I don't know if it's just me, but I really can't make the difference between disabled and enabled extensions, I'm wondering if there's anything else we can do so that it's more visually clear.

tankerkiller125:
Given my poor understanding of the JS for the front-end I've removed myself for review on this one.

KyrneDev:
> I'm wondering if there's anything else we can do so that it's more visually clear.

Which part is confusing?

js/src/common/compat.js
```diff
@@ -91,7 +91,7 @@ export default {
   'utils/stringToColor': stringToColor,
   'utils/Stream': Stream,
   'utils/subclassOf': subclassOf,
-  'utils/SuperTextarea': SuperTextarea,
+  'utils/SuperTextArea': SuperTextarea,
```
SychO9, line 94: typo ?

less/admin/AdminNav.less
```diff
@@ -1,17 +1,91 @@
...
+
+.Header-controls {
+  > li {
+    margin-left: 10px;
+  }
+}
+
+.AdminNav-mainLink {
+  > .Button-label {
+    font-size: 14px !important;
```
SychO9, line 22: why `!important` here ? seems to only override a similar definition for desktop screens, we might as well change it there or remove it there and drop the `!important` here.

less/admin/AdminNav.less
```diff
@@ -20,60 +94,84 @@
...
+        > a:hover {
+          background: @control-bg;
+        }
+        &.active > a {
+          background: @primary-color;
+          font-weight: normal;
+
+          .Button-label,
+          .Button-icon {
+            filter: invert(1);
```
SychO9, line 136: I think it'd be better to use `color: @body-bg` on the anchor above it instead of the filter. More consistent with the rest of the codebase and ensures we're getting the correct color.

SychO9 (changes requested):
Few more xD

The appearance and basics pages headings have too much paddings `padding: 30px` we probably want to drop that (but only on the heading element, not content)

SychO9:
> Which part is confusing?

It's not confusing, I was just thinking it might not be very visually clear, but I have nothing to suggest so ignore me :man_shrugging:

js/src/admin/components/ExtensionPage.js
askvortsov1: Is there any advantage to having `getSettings` take an array over calling it individually for each setting?

KyrneDev:
Some updated screenshots

Screen Shot 2020-11-16 at 10 52 18 AM

Screen Shot 2020-11-16 at 10 52 54 AM

js/src/common/utils/humanTime.ts
```diff
@@ -1,6 +1,3 @@
...
-import 'dayjs/plugin/relativeTime';
-
```
SychO9, line 3: do we not need `dayjs` here ? it's used in the code below

jaspervriends:
First of all, the updated design admin dashboard design looks great! 

Based on your screenshots I'm wondering, would it be possible to have multiple setting pages for a single extension? On the screenshot above, I see you've added a `Permissions` section, wouldn't it be an idea there's some submenu availability for extensions? The submenu would only be visible when you hover the extension or just when the extension settings page is active, just like WordPress does for their admin dashboard.

I think it would be better than having one page with all settings sections on it.

Hovering a setting page in WordPress:
image

When the settings page is opened:
image

For example the SEO extension. There are 5 pages available from an dropdown currently on the extension page itself, `Health check`, `SEO settings`, `Sitemap information`, `Search engine information` and `Set up SSL`, I think it would be nice to have a small submenu available which are in sight directly for the end user.

I'm worried that if there are too many different sections for an extension the admin dashboard will get messy. What are your thoughts about this?

KyrneDev:
@jaspervriends the nav buttons extensions use will now have their own component class that you can extend just like anything else. (https://github.com/KyrneDev/core/blob/master/js/src/admin/components/ExtensionLinkButton.js)

I personally believe this is a niche case and would be best added via an extension that needs it. most extensions don't have a massive amount of settings right now.

I obviously don't have the final say in this matters so if a core dev wants to override this feel free and I'll gladly work on it.

SychO9 (approved):
Looks good to me!

This is a big one though, we'd need more reviewing.

askvortsov1 (commented):
Looked over this again, and I'm very happy with it! Outside of any code style suggestions @clarkwinkelmann and @datitisev may have, I would still like to talk a bit about https://github.com/flarum/core/pull/2409/files#r523281272 (and similar passing of an array of settings in `registerSettings`. Is there a benefit to passing these in as arrays over just passing them individually?

js/src/admin/compat.js
```diff
@@ -1,17 +1,21 @@
 import compat from '../common/compat';
 
 import saveSettings from './utils/saveSettings';
+import ExtensionData from './utils/ExtensionData';
+import extensionEnabled from './utils/extensionEnabled';
```
datitisev, line 5: I feel like this file should be called `isExtensionEnabled` - I had to go to the file to see what it did.

js/src/admin/components/AdminNav.js
```diff
@@ -31,76 +31,119 @@ export default class AdminNav extends Component {
...
+  extensionItems() {
+    const items = new ItemList();
+
+    const categorizedExtensions = getCategorizedExtensions();
+    const categories = app.extensionCategories;
+
+    Object.keys(categorizedExtensions).map((category) => {
+      if (!this.query()) {
+        items.add(
+          `${category} NavDivider`,
```
datitisev, line 116: I don't know how I feel about having class names for the parent list item in the item name...

Do we do this anywhere else?

js/src/admin/components/AdminNav.js
```diff
@@ -31,76 +31,119 @@ export default class AdminNav extends Component {
...
+        const query = this.query().toUpperCase();
+
+        if (
+          !query ||
+          extension.extra['flarum-extension'].title.toUpperCase().includes(query) ||
+          extension.description.toUpperCase().includes(query)
+        ) {
+          items.add(
+            `${extension.id} ExtensionItem`,
+            ExtensionLinkButton.component(
```
datitisev, line 132: I would convert this to JSX - we want to avoid the .component() stuff as much as possible, and having JSX makes it more readable.

js/src/admin/components/AdminNav.js
```diff
@@ -31,76 +31,119 @@ export default class AdminNav extends Component {
...
+          categories[category]
+        );
+      }
+
+      categorizedExtensions[category].map((extension) => {
+        const query = this.query().toUpperCase();
+
+        if (
+          !query ||
+          extension.extra['flarum-extension'].title.toUpperCase().includes(query) ||
```
datitisev, line 127: `extension.extra['flarum-extension'].title` should be set to a variable. Should we check that it exists, as well?

js/src/admin/components/ExtensionPage.js
```diff
@@ -0,0 +1,384 @@
...
+
+    if (this.extension.source || this.extension.support) {
+      infoData.source = {
+        icon: 'fas fa-code',
+        href: this.extension.source ? this.extension.source.url : this.extension.support.source,
+      };
+    }
+
+    Object.keys(this.infoFields).map((field) => {
+      if (this.extension.extra['flarum-extension'].info && this.extension.extra['flarum-extension'].info[field]) {
```
datitisev, line 201: Set `this.extension.extra['flarum-extension']` to a temporary variable within the map function for readability.

js/src/admin/components/ExtensionPage.js
```diff
@@ -0,0 +1,384 @@
...
+          icon: this.infoFields[field],
+          href: this.extension.extra['flarum-extension'].info[field],
+        };
+      }
+    });
+
+    Object.entries(infoData).map(([field, value]) => {
+      items.add(
+        field,
+        LinkButton.component(
```
datitisev, line 212: Preferably make this JSX as well.

js/src/admin/components/ExtensionPage.js
```diff
@@ -0,0 +1,384 @@
...
+      })
+      .then(() => {
+        if (!enabled) localStorage.setItem('enabledExtension', this.extension.id);
+        window.location.reload();
+      });
+
+    app.modal.show(LoadingModal);
+  }
+
+  dirty() {
```
datitisev, line 316: We have `isEnabled` below but the `dirty` and `changed` methods aren't prefixed by `is` - we should do one or the other, not both imo.

js/src/admin/components/ExtensionPermissionGrid.js
```diff
@@ -0,0 +1,39 @@
...
+
+    this.extensionId = this.attrs.extensionId;
+  }
+
+  permissionItems() {
+    const permissionCategories = super.permissionItems();
+
+    permissionCategories.items = Object.entries(permissionCategories.items)
+      .filter(([k, v]) => v.content.children.length > 0)
+      .reduce((obj, [key, v]) => {
```
datitisev, line 16: I would change the names of these arguments to be more clear about what this code is doing. `k, v` isn't very specific and doesn't give much context.

js/src/admin/components/HeaderSecondary.js
```diff
@@ -19,6 +20,19 @@ export default class HeaderSecondary extends Component {
   items() {
     const items = new ItemList();
 
+    items.add(
+      'help',
+      LinkButton.component(
```
datitisev, line 25: Use JSX.

js/src/admin/utils/ExtensionData.js
```diff
@@ -0,0 +1,170 @@
+import ItemList from '../../common/utils/ItemList';
+
+export default class ExtensionData {
```
datitisev, line 3: This class looks to me like an extender. Not sure if that's the intention, and I know we don't really use the JS extenders we have, but just an observation - not sure if that was intended.

datitisev (commented):
I've added some comments. Not all are important, some are just some observations or requests to change new .component() code into JSX.

js/src/admin/utils/ExtensionData.js
askvortsov1: I think this is one of those places where having a pseudo-extender is simpler / safer than having users directly monkey-patch stuff on.

datitisev: I'm not saying to change it from that - I like having it like this instead.

js/src/admin/components/AdminNav.js
KyrneDev: I don't think we need to check if it exists, every extension needs to have a title.

js/src/admin/components/ExtensionPage.js
KyrneDev: You're right, `changed` should be `isChanged`, but `dirty` doesn't return a bool so it shouldn't be prefixed with is.

datitisev: Ah, good point.

js/src/admin/components/ExtensionPage.js
KyrneDev: Not really. I'm not sure which way looks better.

js/src/admin/components/ExtensionPage.js
```diff
@@ -126,7 +126,9 @@ export default class ExtensionPage extends Page {
             </Button>
           ) : settings ? (
             <div className="Form">
-              {this.getSettings(settings)}
+              {settings.map((setting) => {
+                return this.getSetting(setting);
```
askvortsov1, line 130: Couldn't we do `settings.map(this.getSetting)`?

askvortsov1 (changes requested):

js/src/admin/components/ExtensionPage.js
KyrneDev: We can do `settings.map(setting => this.getSetting(setting))`

That would mess with `this` context.

js/src/admin/utils/ExtensionData.js
```diff
@@ -26,7 +26,7 @@ export default class ExtensionData {
   /**
    * This function registers your settings with Flarum
    *
-   * @example - settings
+   * @example
    *
    * .registerSettings({
```
askvortsov1, line 31: This should be made singular

js/src/admin/components/ExtensionPage.js
```diff
@@ -126,9 +126,7 @@ export default class ExtensionPage extends Page {
             </Button>
           ) : settings ? (
             <div className="Form">
-              {settings.map((setting) => {
-                return this.getSetting(setting);
-              })}
+              {settings.map((setting) => this.getSetting(setting))}
```
askvortsov1, line 129: Why not `settings.map(this.getSetting)`?

askvortsov1 (changes requested):

askvortsov1 (approved):

js/src/admin/components/ExtensionPage.js
datitisev: Can also do `settings.map(this.getSetting.bind(this))`

KyrneDev: Done

tankerkiller125 (approved):
I've ensured that this PR works locally, everything seems to work well and I was unable to break anything in my testing.

PR #2410, opened 2020-10-24
Access request actor in error handler

### LINKED ISSUE #1847: Unable to access request actor in error handler

SychO9:
**Fixes #1847**

Very well explained here: https://github.com/flarum/core/issues/1847#issuecomment-526529542

With this, the actor can be accessed through `$request->getAttribute('actorReference')->getActor()`
Do we want to deprecate `$request->getAttribute('actor')` ? If so do we want to replace it with the above everywhere as well or leave that for #869 ?

**Confirmed**

### DIFF
--- a/src/Http/RequestUtil.php
+++ b/src/Http/RequestUtil.php
@@ -30,7 +30,7 @@ class RequestUtil
 
         $actorReference->setActor($actor);
 
-        // BC Layer
+        // @deprecated in 1.0
         $request = $request->withAttribute('actor', $actor);
 
         return $request;
--- a/src/Http/RequestUtil.php
+++ b/src/Http/RequestUtil.php
@@ -16,11 +16,23 @@ class RequestUtil
 {
     public static function getActor(Request $request): User
     {
-        return $request->getAttribute('actor');
+        return $request->getAttribute('actorReference')->getActor();
     }
 
     public static function withActor(Request $request, User $actor): Request
     {
-        return $request->withAttribute('actor', $actor);
+        $actorReference = $request->getAttribute('actorReference');
+
+        if (! $actorReference) {
+            $actorReference = new ActorReference;
+            $request = $request->withAttribute('actorReference', $actorReference);
+        }
+
+        $actorReference->setActor($actor);
+
+        // BC Layer
+        $request = $request->withAttribute('actor', $actor);
+
+        return $request;
     }
 }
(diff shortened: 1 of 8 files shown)


askvortsov1:
To be honest, I would prefer to see https://github.com/flarum/core/issues/869 happen first; that way we can start by abstracting away access, then switch up the underlying implementation. For this issue though, I'd lean towards not switching anything up yet, and only deprecating the old way after we abstract it away through RequestUtil.

SychO9:
Sounds good, we can come back to this after #869 is implemented.

askvortsov1:
This is from half a year ago???? How time flies... Anyways, this can be updated now that #2449 has been merged.

src/Http/RequestUtil.php
```diff
@@ -16,11 +16,23 @@ class RequestUtil
...
 
     public static function withActor(Request $request, User $actor): Request
     {
-        return $request->withAttribute('actor', $actor);
+        $actorReference = $request->getAttribute('actorReference');
+
+        if (! $actorReference) {
+            $actorReference = new ActorReference;
+            $request = $request->withAttribute('actorReference', $actorReference);
+        }
```
SychO9, line 29: Not a fan of this at all, but I preferred to go with this instead of explicitly setting the actor reference in `Api\Client`

askvortsov1: Why not? If the whole goal of these methods is to encapsulate the request's user, the actual mechanism used for storing the user should be an implementation detail of the methods.

> explicitly setting the actor reference in Api\Client

We definitely wouldn't want this, it'd break encapsulation.

askvortsov1:
> Do we want to deprecate $request->getAttribute('actor') ?

I would be in favor of deprecating it, and leaving it for now.

askvortsov1 (approved):

tankerkiller125 (approved):

PR #2411, opened 2020-10-25
Drop mixin-like attributes and convert to typescript

clarkwinkelmann:
Remove `Object.assign` from model declarations and add typescript hints. This should make those attributes appear in the documentation.

The difference between `Object.assign`/`mixin` and the proposed solution is that the attributes are no longer part of the model `prototype` and as such cannot be easily overridden by extensions. However I don't think any extension ever attempted that and I'm not sure why the "mixin" style was used at all in the past.

To convert the models to typescript I had to convert a few other classes to typescript as well. Where do we want to put typescript interfaces? The `ModelData` currently in `Model` would probably also be useful for the `Store` class.

There's also the global `ColorThief` package, which is only available in the `forum` frontend, so it probably shouldn't be in the global `shims.d.ts`.

There's also the question on the type-hinting of all `Model.attribute` calls. There's always a risk of the results being `undefined` (if an extension decided to load only part of the attributes or if it's a newly created model client-side) or `null` for attributes with `transformDate`. So do we want to type-hint `|null|undefined` on all of them? That creates warnings in various places of the code where the value is read, assuming to be a string or number.

**Confirmed**

**Required changes:**


clarkwinkelmann:
If anyone wants to give feedback on this. I think this will eventually have to be merged as part of the move to typescript.

There's still the open question of prototype methods vs methods as member variables which impact the ability to use `extend()` on them. If anyone with javascript knowledge wants to pitch in, I'd be happy to hear your thoughts.

The other question was about documentation. Right now the attributes aren't picked up by our automatically generated documentation. I'm not sure how to verify whether the new method solves that or not before we merge.

And lastly there's the typescript question about where to put the new interface that this PR introduces.

datitisev:
> There's still the open question of prototype methods vs methods as member variables 

Can you expand on this? Looking at the diff I don't see a place where .extend() cannot be used...

clarkwinkelmann:
I have not tested this again recently but basically I remember this:

```js
override(Discussion.prototype, 'title', function (original, value) {
    return value ? original(value) : 'All discussions will have this name';
});
```

won't work. Because the method is correctly created on the `prototype`, but the original `Model.attribute` will actually be set as the actual value of `(instanceof Discussion).title` when the object is instantiated, bypassing any `prototype` method.

This fact is already true of the attributes of all models defined by extensions that use this method.

It's just super uncommon to want to `extend` or `override` an attribute or relationship on a javascript model. I've never had to do this in any extension I have worked on so far.

datitisev:
I guess that wasn't the case before because we were assigning the prototype, right? 🤔

datitisev:
I went to check this, and you are correct. The properties aren't exposed in the prototype, and thus would (1) be a breaking change and (2) prevent extensibility.

Hm... I think we'll have to use methods then... we could have some function that merges the prototypes (similar to mixin) but then that'd become messy and probably not work well with typings & we'd be back to using mixin.

Tried looking for a solution that'd make declaring methods look like the fields there so the behavior is kept but its cleaner code, and couldn't really find anything...

clarkwinkelmann:
Should we just use something like

```js
class MyModel extends Model {
    myAttribute(value = undefined) {
        return Model.attribute('myAttribute').call(this, value);
    }
}
```

This could be simplified by introducing a new member method so we don't need to bind `this`:

```js
class MyModel extends Model {
    myAttribute(value = undefined) {
        return this.attributeReadWrite('myAttribute', value)
    }
}
```

For extensions that don't actually use the prop-like writable feature, they could do even simpler, but I'm not sure we want to have multiple ways of doing things:

```js
class MyModel extends Model {
    myAttribute() {
        return this.attribute('myAttribute')
    }
}
```

Personally I very rarely use the prop-like ability of model attributes, because I always have different logic for read and write and generally use `model.pushAttributes()` to apply the changes. Actually most of the time I don't even push the attributes back to the model. I use temporary variables, then the attributes are refreshed in the store when calling `save()` automatically.

I wouldn't be against a new system that's completely getter-based.

PR #2415, opened 2020-10-26
Ensure scripts provided by textformatter are run

### LINKED ISSUE #2412: CommentPost content script tags not executed during oncreate
## Bug Report

**Current Behavior**
When a post is loaded and its template contains javascript, the javascript doesn't run until the post experiences an onupdate cycle.

This can be experienced with syntax-highlighted code blocks, but probably also other extensions.

**Steps to Reproduce**
1. Create post with code block
2. Refresh page
3. See post is not highlighted

This can probably be reproduced with other community extensions. I think the code block is the only one from core with javascript in the template itself.

**Expected Behavior**
Javascript should run during the first render

**Screenshots**
image

**Environment**
- Flarum version: beta 14
- Website URL: tested locally

**Possible Solution**
We need to call the code from `onupdate` in `oncreate`. Previously that code ran in `config` without any regard to `isInitialized`.

Beta 13 code

https://github.com/flarum/core/blob/fd371c1203439810af8f6814cdff402faebf4126/js/src/forum/components/CommentPost.js#L57-L72

Current code

https://github.com/flarum/core/blob/c9a04fe009abf474b39f562e9d4bf13c5ecf25c8/js/src/forum/components/CommentPost.js#L59-L74

While we're at it, I think we should also call `super.onupdate` with a `vnode` parameter. If an extension was to extend `Post.prototype.onupdate`, they wouldn't be able to access `vnode` because `CommentPost` forgets to pass it up.

**Additional Context**
Reported by RebelCoderRU here https://discuss.flarum.org/d/25227-code-highlighting-not-working-properly-on-beta-14

askvortsov1:
Fixes #2412 

It seems that between Mithril 0.2 and Mithril 2.0, scripts contained within m.trust are no longer automatically evaluated. So, now we have to make them run ourselves. To do this, we "clone" the script node (using `node.cloneNode()` didn't work unfortunately), and replace the text formatter provided script node with our cloned one. This works for both external and inline scripts, and also allows us to get rid of eval.

**comments for reviewers**

If I'm understanding this correctly (pun intended), `this` in the `<script>` tag previously pointed to the script tag, now it acts as the default `this` in the global scope. The script tag is still available via `document.currentScript`. However, `this` was never available by the default use of the textformatter extension, so I believe we should be fine. Is this something we are concerned about?

### DIFF
--- a/js/src/forum/components/CommentPost.js
+++ b/js/src/forum/components/CommentPost.js
@@ -65,10 +65,9 @@ export default class CommentPost extends Post {
     if (this.contentHtml !== contentHtml) {
       this.$('.Post-body script').each(function () {
         const script = document.createElement('script');
-        script.innerText = this.innerText;
+        script.textContent = this.textContent;
         Array.from(this.attributes).forEach((attr) => script.setAttribute(attr.name, attr.value));
-        this.parentNode.appendChild(script);
-        this.remove();
+        this.parentNode.replaceChild(script, this);
       });
     }
 
--- a/js/src/forum/components/CommentPost.js
+++ b/js/src/forum/components/CommentPost.js
@@ -56,9 +56,7 @@ export default class CommentPost extends Post {
     ]);
   }
 
-  onupdate(vnode) {
-    super.onupdate();
-
+  refreshContent() {
     const contentHtml = this.isEditing() ? '' : this.attrs.post.contentHtml();
 
     // If the post content has changed since the last render, we'll run through
@@ -71,6 +69,18 @@ export default class CommentPost extends Post {
     }
 
     this.contentHtml = contentHtml;
+  }
+
+  oncreate(vnode) {
+    super.oncreate(vnode);
+
+    this.refreshContent();
+  }
+
+  onupdate(vnode) {
+    super.onupdate(vnode);
+
+    this.refreshContent();
   }
 
   isEditing() {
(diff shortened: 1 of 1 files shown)


tankerkiller125 (approved):
I have not tested this yet, but the code looks good to me.

luceos:
@clarkwinkelmann or @datitisev could either of you check this out locally and review, 🙏

clarkwinkelmann (changes requested):
I tested the PR and it looks good.

However in an effort to use the same strategy everywhere, I checked what TextFormatter does in `preview`. I think they don't actually have the same issue as ourselves, because the javascript is supposed to run fine with the way they insert their HTML. But they do have a snippet almost identical to ours to handle a bug in Chrome.

I actually quite like the differences in TextFormatter. `replaceChild` is used instead of inserting then removing a script tag. And they use `textContent` instead of `innerText`, which is the standard property name and [might be a bit more performant](https://kellegous.com/j/2013/02/27/innertext-vs-textcontent/), although I didn't find any explanation specific to `script` tags.

https://github.com/s9e/TextFormatter/blob/12bdf56b7fd3667290c5336cadb06a0c7720c7d3/src/render.js#L113-L129

So my suggestion would be to change the code as follows, which I have also tested working locally:

```js
      this.$('.Post-body script').each(function () {
        const script = document.createElement('script');
        script.textContent = this.textContent;
        Array.from(this.attributes).forEach((attr) => script.setAttribute(attr.name, attr.value));
        this.parentNode.replaceChild(script, this);
      });
```

Regarding the `this` question, I have no idea. I've never used javascript in a TextFormatter template.

I would actually much prefer if we completely got rid of javascript in templates and force developers to use a frontend extension to add the javascript.

askvortsov1:
Done. I think the `this` issue isn't an issue: I don't know any situation where someone would expect `this` to refer to the script DOM node, and I don't know of a valid reason for supporting it.

clarkwinkelmann (approved):
I tested the last commit locally for good measure :+1:

PR #2423, opened 2020-10-28
Fix Callables for Extenders

### LINKED ISSUE #2401: All callable-based extenders should accept and resolve invoke-able classes
Right now some extenders method only accept `callable` arguments. This complicates development when we want to place a large amount of code in a separate file. It's possible to create an invoke-able class and manually resolve it ourselves before passing it to the extender, but it would be even better if all extenders could do this for us, so that we can only pass the `::class` name to them.

Right now we have:

:heavy_check_mark: = accepts invoke-able classes and resolves them through container
:negative_squared_cross_mark: = only accepts `callable`

- [Compat](https://github.com/flarum/core/blob/master/src/Extend/Compat.php): :heavy_check_mark: (uses container `call`)
- [Event.listen](https://github.com/flarum/core/blob/master/src/Extend/Event.php): :heavy_check_mark: (supported by Laravel's Dispatcher)
- [Formatter.configure](https://github.com/flarum/core/blob/master/src/Extend/Formatter.php): :heavy_check_mark: (manually implemented)
- [Frontend.content](https://github.com/flarum/core/blob/master/src/Extend/Frontend.php): :heavy_check_mark: (manually implemented)
- [Model.default](https://github.com/flarum/core/blob/master/src/Extend/Model.php#L57): :negative_squared_cross_mark: [uses is_callable()](https://github.com/flarum/core/blob/master/src/Database/AbstractModel.php#L85) in the model. Can't really resolve strings through the container since strings are also valid arguments
- [Model.relationship](https://github.com/flarum/core/blob/master/src/Extend/Model.php#L171): :negative_squared_cross_mark: argument restricted to `callable`
- [User.permissionGroups](https://github.com/flarum/core/blob/master/src/Extend/User.php#L47): :negative_squared_cross_mark: argument restricted to `callable`

I think the two that need changing are `Model.relationship` and `User.permissionGroups`. We need to change the method signature to accept string, and manually resolve strings through the container or make everything go through the container `call`.

I'm also not sure whether there's a difference between those different methods of calling the invoke-able classes. We need to make sure it's consistent across the API.

When I started writing this issue I had problems not being able to pass invoke-able classes to `callable` arguments at all. But it seems like I made some kind of error. Invoke-able classes work fine with `callable` arguments, we just need to instanciate the class ourselves beforehand.

askvortsov1:
Fixes #2401, mostly minor cleanup and adding tests to prevent regressions.

### DIFF
--- a/src/Extend/Model.php
+++ b/src/Extend/Model.php
@@ -50,8 +50,9 @@ class Model implements ExtenderInterface
     }
 
     /**
-     * Add a default value for a given attribute, which can be an explicit value, or a closure.
-     * Unlike with some other extenders, it CANNOT be an invokable class.
+     * Add a default value for a given attribute, which can be an explicit value, a closure,
+     * or an instance of an invokable class. Unlike with some other extenders,
+     * it CANNOT be the `::class` attribute of an invokable class.
      *
      * @param string $attribute
      * @param mixed $value
--- a/src/Extend/Model.php
+++ b/src/Extend/Model.php
@@ -11,6 +11,7 @@ namespace Flarum\Extend;
 
 use Flarum\Database\AbstractModel;
 use Flarum\Extension\Extension;
+use Flarum\Foundation\ContainerUtil;
 use Illuminate\Contracts\Container\Container;
 use Illuminate\Support\Arr;
 
@@ -180,11 +181,7 @@ class Model implements ExtenderInterface
     public function extend(Container $container, Extension $extension = null)
     {
         foreach ($this->customRelations as $name => $callback) {
-            if (is_string($callback)) {
-                $callback = $container->make($callback);
-            }
-
-            Arr::set(AbstractModel::$customRelations, "$this->modelClass.$name", $callback);
+            Arr::set(AbstractModel::$customRelations, "$this->modelClass.$name", ContainerUtil::wrapCallback($callback, $container));
         }
     }
 }
--- a/src/Extend/Model.php
+++ b/src/Extend/Model.php
@@ -17,6 +17,7 @@ use Illuminate\Support\Arr;
 class Model implements ExtenderInterface
 {
     private $modelClass;
+    private $customRelations = [];
 
     /**
      * @param string $modelClass The ::class attribute of the model you are modifying.
@@ -171,13 +172,19 @@ class Model implements ExtenderInterface
      */
     public function relationship(string $name, $callback)
     {
-        Arr::set(AbstractModel::$customRelations, "$this->modelClass.$name", $callback);
+        $this->customRelations[$name] = $callback;
 
         return $this;
     }
 
     public function extend(Container $container, Extension $extension = null)
     {
-        // Nothing needed here.
+        foreach ($this->customRelations as $name => $callback) {
+            if (is_string($callback)) {
+                $callback = $container->make($callback);
+            }
+
+            Arr::set(AbstractModel::$customRelations, "$this->modelClass.$name", $callback);
+        }
     }
 }
--- a/src/Extend/Model.php
+++ b/src/Extend/Model.php
@@ -157,7 +157,7 @@ class Model implements ExtenderInterface
      * @param string $name: The name of the relation. This doesn't have to be anything in particular,
      *                      but has to be unique from other relation names for this model, and should
      *                      work as the name of a method.
-     * @param callable $callable
+     * @param callable|string $callback
      *
      * The callable can be a closure or invokable class, and should accept:
      * - $instance: An instance of this model.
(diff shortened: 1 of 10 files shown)


src/Extend/Model.php
```diff
@@ -168,15 +170,21 @@ public function hasMany(string $name, string $related, string $foreignKey = null
...
 
         return $this;
     }
 
     public function extend(Container $container, Extension $extension = null)
     {
-        // Nothing needed here.
+        foreach ($this->customRelations as $name => $callback) {
+            if (is_string($callback)) {
+                $callback = $container->make($callback);
```
clarkwinkelmann, line 184: My concern here is that this gets resolved very early. This could create a NotificationMailer situation.

If we don't want to do the resolving in `AbstractModel`, we could do it with an anonymous callback, something like (untested):

```suggestion
                $callback = function () use ($container) {
                    return $container->call($callback, func_get_args());
                };
```

clarkwinkelmann (commented):
Not tested locally. Looks good, just one comment.

src/Extend/Model.php
askvortsov1: I like this idea, but we might want to extract this out into a util so it can also be used for the `Formatter` and `Frontend` extensions. Maybe an `ExtenderUtil` class with static methods? Not 100% sure.

clarkwinkelmann: Yes for code re-use, but calling the util would take almost as many lines of code as what it's replacing, no? :sweat_smile:

askvortsov1: It turns out to be a lot nicer actually. Only thing I'm unsure of is where to put the util (namespace + class name), I don't think we have too many of these right now?

src/Foundation/ContainerUtil.php
```diff
@@ -0,0 +1,26 @@
...
+class ContainerUtil
+{
+    public static function wrapCallback($callback, $container)
+    {
+        if (is_string($callback)) {
+            $callback = function () use ($container, $callback) {
+                $callback = $container->make($callback);
+
+                return call_user_func_array($callback, func_get_args());
+            };
```
SychO9, line 21: why not just  `$callback = $container->make($callback);` ?

askvortsov1: Clark pointed out in https://github.com/flarum/core/pull/2176#discussion_r517034525 that `$callback = $container->make($callback);` means that the callback will be resolved every time the site boots, even if its not needed for that request.

SychO9: my bad, didn't see that..

src/Extend/Model.php
SychO9: we could potentially put the class in `Flarum\Support` similar to where laravel puts its helper classes, that could be nice especially if we end up with more helper classes in the future.

But I personally don't mind either way.

src/Foundation/ContainerUtil.php
askvortsov1: I do agree it's a bit ugly, if there are simpler ways to do it, I'm all ears. But at least with the util, it's abstracted away so if we need to change it, that should be easy.

SychO9: it looks good to me :shrug:

src/Extend/Model.php
```diff
@@ -49,6 +51,7 @@ public function dateAttribute(string $attribute)
 
     /**
      * Add a default value for a given attribute, which can be an explicit value, or a closure.
+     * Unlike with some other extenders, it CANNOT be an invokable class.
```
clarkwinkelmann, line 54: Technically, it can be an invokable class if you pass an object and not the class name as string.

It's the string-based invokable classes that are specifically not possible.

src/Foundation/ContainerUtil.php
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * For detailed copyright and license information, please view the
+ * LICENSE file that was distributed with this source code.
+ */
+
+namespace Flarum\Foundation;
+
+class ContainerUtil
+{
+    public static function wrapCallback($callback, $container)
```
clarkwinkelmann, line 14: I think we need a comment here explaining the use case for this function.

Something like "Wraps a callback so that string-based invokable classes get resolved only when actually used".

Also, do we want an `@internal` phpdoc to discourage extensions from calling this directly?

clarkwinkelmann (changes requested):
Looking good with the helper! Some more comments.

askvortsov1:
@clarkwinkelmann fixed! Any thoughts on naming for the util?

src/Foundation/ContainerUtil.php
```diff
@@ -0,0 +1,36 @@
...
+namespace Flarum\Foundation;
+
+use Illuminate\Contracts\Container\Container;
+
+class ContainerUtil
+{
+    /**
+     * Wraps a callback so that string-based invokable classes get resolved only when actually used.
+     *
+     * @internal Backwards compatability not guarunteed.
```
SychO9, line 19: guaranteed*

askvortsov1: I love the English language...

SychO9 (approved):
Looks good to me!

luceos (approved):

clarkwinkelmann (approved):

PR #2424, opened 2020-10-29
Add Notification Type Extender

### LINKED ISSUE #2422: Notification Type Extender
Part of #1891

We need an extender and unit tests for registering notification types and subject serializers to replace the `ConfigureNotificationTypes` event.

This ties closely into #2419, especially with the issue about preferences.

SychO9:
**Fixes #2422**
**Part of #1891**

Would it be better to try avoiding duplicating this block of code (by defining a singleton and extending that instead in the extender)
https://github.com/flarum/core/blob/24b7a21507c4bdc5f44d1fed633129915ade0bd7/src/Notification/NotificationServiceProvider.php#L42-L59

**Confirmed**

### DIFF
--- a/src/Api/ApiServiceProvider.php
+++ b/src/Api/ApiServiceProvider.php
@@ -75,6 +75,12 @@ class ApiServiceProvider extends AbstractServiceProvider
 
             return $pipe;
         });
+
+        $this->app->singleton('flarum.api.notification_serializers', function () {
+            return [
+                'discussionRenamed' => BasicDiscussionSerializer::class
+            ];
+        });
     }
 
     /**
@@ -96,12 +102,6 @@ class ApiServiceProvider extends AbstractServiceProvider
      */
     protected function registerNotificationSerializers()
     {
-        $this->app->singleton('flarum.api.notification_serializers', function () {
-            return [
-                'discussionRenamed' => BasicDiscussionSerializer::class
-            ];
-        });
-
         $blueprints = [];
         $serializers = $this->app->make('flarum.api.notification_serializers');
 
--- a/src/Api/ApiServiceProvider.php
+++ b/src/Api/ApiServiceProvider.php
@@ -88,7 +88,7 @@ class ApiServiceProvider extends AbstractServiceProvider
      */
     public function boot()
     {
-        $this->bootNotificationSerializers();
+        $this->setNotificationSerializers();
 
         AbstractSerializeController::setContainer($this->app);
         AbstractSerializeController::setEventDispatcher($events = $this->app->make('events'));
@@ -100,7 +100,7 @@ class ApiServiceProvider extends AbstractServiceProvider
     /**
      * Register notification serializers.
      */
-    protected function bootNotificationSerializers()
+    protected function setNotificationSerializers()
     {
         $blueprints = [];
         $serializers = $this->app->make('flarum.api.notification_serializers');
--- a/src/Api/ApiServiceProvider.php
+++ b/src/Api/ApiServiceProvider.php
@@ -88,7 +88,7 @@ class ApiServiceProvider extends AbstractServiceProvider
      */
     public function boot()
     {
-        $this->registerNotificationSerializers();
+        $this->bootNotificationSerializers();
 
         AbstractSerializeController::setContainer($this->app);
         AbstractSerializeController::setEventDispatcher($events = $this->app->make('events'));
@@ -100,7 +100,7 @@ class ApiServiceProvider extends AbstractServiceProvider
     /**
      * Register notification serializers.
      */
-    protected function registerNotificationSerializers()
+    protected function bootNotificationSerializers()
     {
         $blueprints = [];
         $serializers = $this->app->make('flarum.api.notification_serializers');
--- a/src/Api/ApiServiceProvider.php
+++ b/src/Api/ApiServiceProvider.php
@@ -101,6 +101,7 @@ class ApiServiceProvider extends AbstractServiceProvider
             'discussionRenamed' => BasicDiscussionSerializer::class
         ];
 
+        // Deprecated in beta 15, remove in beta 16
         $this->app->make('events')->dispatch(
             new ConfigureNotificationTypes($blueprints, $serializers)
         );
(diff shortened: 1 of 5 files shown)


src/Api/ApiServiceProvider.php
```diff
@@ -101,6 +101,7 @@ protected function registerNotificationSerializers()
             'discussionRenamed' => BasicDiscussionSerializer::class
         ];
 
+        // Deprecated in beta 15, remove in beta 16
```
askvortsov1, line 104: Another place where the singleton approach might be better: best to avoid having this logic in multiple places if we can aboid it.

tests/integration/extenders/NotificationTest.php
```diff
@@ -0,0 +1,68 @@
...
+     */
+    public function notification_type_doesnt_exist_by_default()
+    {
+        $this->assertArrayNotHasKey('customNotificationType', Notification::getSubjectModels());
+    }
+
+    /**
+     * @test
+     */
+    public function notification_type_exists_if_added()
```
askvortsov1, line 29: More tests that would be nice to have could cover that the serializer is properly registered, and that preferences are properly saved.

askvortsov1 (changes requested):

src/Extend/Notification.php
```diff
@@ -0,0 +1,51 @@
...
+use Flarum\Notification\Notification as NotificationModel;
+use Flarum\User\User;
+use Illuminate\Contracts\Container\Container;
+use ReflectionClass;
+
+class Notification implements ExtenderInterface
+{
+    public function type(string $blueprint, string $serializer, array $enabledByDefault = [])
+    {
+        $type = $blueprint::getType();
```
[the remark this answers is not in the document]
SychO9: do you mean specifically the call to the `getType` method ?

tests/integration/extenders/NotificationTest.php
SychO9: so we just need to do some checks on the `$subjectSerializers` static property of the `NotificationSerializer` class and `$preferences` static property of the `User` model, correct ?

src/Extend/Notification.php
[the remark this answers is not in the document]
askvortsov1: I mean the entire block of code, sorry I should have selected this better.

tests/integration/extenders/NotificationTest.php
askvortsov1: I think so. We could even throw in some reflection for Serializer but I'm not sure that's necessary at the moment

SychO9: well, those two properties are protected, we'd have to create static methods to read them, is that fine when the only use is for tests ?

askvortsov1: Hmm I didn't check for that. In that case, let's leave it as is.

src/Notification/NotificationServiceProvider.php
```diff
@@ -30,34 +30,43 @@ public function boot()
...
-        $blueprints = [
-            DiscussionRenamedBlueprint::class => ['alert']
-        ];
+        $this->app->singleton('flarum.notification.blueprints', function () {
+            return [
+                DiscussionRenamedBlueprint::class => ['alert']
+            ];
+        });
+
+        $blueprints = $this->app->make('flarum.notification.blueprints');
```
askvortsov1, line 39: Feels like this (and the logic that it uses) would make more sense to have in boot, after all extensions have done their thing

src/Notification/NotificationServiceProvider.php
```diff
@@ -30,34 +30,43 @@ public function boot()
...
-            Notification::setSubjectModel(
-                $type = $blueprint::getType(),
-                $blueprint::getSubjectModel()
-            );
+        foreach ($blueprints as $blueprint => $channelsEnabledByDefault) {
+            $this->addType($blueprint, $channelsEnabledByDefault);
+        }
+    }
+
+    public function addType(Blueprint\BlueprintInterface $blueprint, array $channelsEnabledByDefault)
```
askvortsov1: This can probably be protected

src/Api/ApiServiceProvider.php
```diff
@@ -96,10 +96,14 @@ public function boot()
...
+            return [
+                'discussionRenamed' => BasicDiscussionSerializer::class
+            ];
+        });
+
         $blueprints = [];
-        $serializers = [
-            'discussionRenamed' => BasicDiscussionSerializer::class
-        ];
+        $serializers = $this->app->make('flarum.api.notification_serializers');
```
askvortsov1, line 106: Similarly, if we move this to boot it'll only evaluate the singleton after extensions have been processed.

src/Notification/NotificationServiceProvider.php
```diff
@@ -30,34 +30,43 @@ public function boot()
      */
     public function registerNotificationTypes()
     {
-        $blueprints = [
-            DiscussionRenamedBlueprint::class => ['alert']
-        ];
+        $this->app->singleton('flarum.notification.blueprints', function () {
```
askvortsov1, line 33: As a side note, this setup is great, because it will allow us to edit the enabled-by-default settings of some types.

src/Api/ApiServiceProvider.php
askvortsov1: Taking a second look, the logic is already technically in boot, which is why it worked before. We should change it so that the singleton is registered in `register` and evaluated in `boot`.

askvortsov1 (changes requested):

src/Notification/NotificationServiceProvider.php
SychO9: *facepalm*

src/Notification/NotificationServiceProvider.php
```diff
@@ -30,33 +30,43 @@ public function boot()
      */
     public function registerNotificationTypes()
```
SychO9, line 31: this should be `protected` as well I think

askvortsov1: Yeah that makes sense to me

askvortsov1 (approved):
Other than making `registerNotificationTypes` protected (and possibly renaming it to avoid confusion since its run from boot, not from register), looks good to me! Thanks!

tankerkiller125 (approved):

PR #2426, opened 2020-10-30
Updater should show on any subpath, like installer

askvortsov1:
If not yet installed and you go to any subpath, the installer will still be shown.

If an update is not installed and you go to a subpath, it'll crash because it can't find the route.

Now, the updater will work just like the installer.


luceos (approved):

tankerkiller125 (approved):

idk-pixel (commented):
Approve yeah

PR #2427, opened 2020-10-30
Update required PHP Version

tankerkiller125:
Updates the required PHP version to PHP 7.3 as 7.2 will stop receiving security updates on Nov 30th 2020, given that beta 15 will likely be after this date we should drop support for the older version. As we did when PHP 7.1 stopped receiving updates.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,7 +8,7 @@ jobs:
 
     strategy:
       matrix:
-        php: [7.2, 7.3, 7.4]
+        php: [7.3, 7.4]
         service: ['mysql:5.7', mariadb]
         prefix: ['', flarum_]
 
--- a/composer.json
+++ b/composer.json
@@ -39,7 +39,7 @@
         "docs": "https://flarum.org/docs/"
     },
     "require": {
-        "php": ">=7.2",
+        "php": ">=7.3",
         "axy/sourcemap": "^0.1.4",
         "components/font-awesome": "^5.14.0",
         "dflydev/fig-cookies": "^2.0.1",


tankerkiller125:
we dropped support for 7.1 because of it's lack of security updates, not because we actively needed to use PHP 7.2 features (although we now do in some places).

While I understand your sentiment I don't think we should be playing the whole "sure feel free to use PHP 7.2, but we won't support you if something goes wrong or your site gets hacked" part of shipping secure software is using secure and patched dependencies, including the PHP version we require. And to be brutally honest given the modern PHP ecosystem if they really really need 7.2 for an app they have that hasn't updated yet (high unlikely at this rate) then they can run both 7.2 and 7.3 or 7.4 easily.

clarkwinkelmann (approved):

luceos:
I don't agree. We only upped the php requirement because dependencies required that. That 7.1 was also no longer maintained was a welcoming addition, but not the reason.

Especially users on shared hosting don't have a say in which version of php they are running. So ultimately this could cause them these users to be unable to update and/or require them to migrate to another hosting company.

I think we should adhere to the php version requirement that laravel/framework holds as much as possible. 6.x - what we use now has "^7.2.5|^8.0",

https://github.com/laravel/framework/blob/6.x/composer.json#L18

SychO9:
I agree with luceos, we should take our time before dropping support for php 7.2, it's too early at this time.

Also as far as I can tell, this subversion seems to be the most used on the web.

tankerkiller125:
I'll go ahead and close this for now then, I should note that we can't use `|^8.0` yet because of cookie library dependency (last I looked at it about a week ago)

PR #2428, opened 2020-10-30
Remove Deprecated Beta14 Code

### LINKED ISSUE #2278: Remove deprecated "moment" dayjs alias for Beta 15.
See https://github.com/flarum/core/pull/2219#issuecomment-682840911

### LINKED ISSUE #2346: Remove B14 BC layers
Check frontend and backend to remove all deprecated BC layers introduced in beta 14

askvortsov1:
**Fixes #2346 **
**Fixes #2278**

- Remove momentjs alias
- Remove various beta 14 frontend BC layers
- Remove deprecated User events
- Remove path helpers and Application methods
- Remove AssertPermissionTrait

I left in the warnings and errors for invalid attrs in Component and in the various Modal / Alert / etc, because those can be common mistakes. I'd definitely like to keep the Component ones, not 100% sure about modal, alert, and composer.
**Confirmed**


clarkwinkelmann (commented):
Looking good code-wise.

Yes I think it's good to keep some of the warnings for easy errors that aren't specific to the beta 14 update. The `children` warning could probably be removed IMO since that construct will not be found in any example in either Flarum or Mithril anymore. Warnings for parameter orders still make sense, so do the checks for parameters to modals and alerts.

I would still suggest we keep this on hold for a bit (1 or 2 more weeks) until more extensions finished migrating to beta 14. We might still find some surprises related to what we have here, and I'd prefer we not need to revert some of the stuff later on.

askvortsov1:
There isn't really anything in the release that this is blocking, so we could just wait to merge it until feature freeze. Only matter is to decide which warnings to keep and which to remove. My thoughts about `children` are since its one of the most common changes needed, but the initial error message isn't super clear, so if some extensions are migrating late during next release cycle, it might help them. And then all this could be removed before stable.

clarkwinkelmann:
> My thoughts about children are since its one of the most common changes needed, but the initial error message isn't super clear

True, some of those warnings are super helpful when compared to what error message would appear without them. It's worth keeping them for developer sanity even if they rarely happen :innocent:

tankerkiller125 (approved):

clarkwinkelmann (approved):

PR #2431, opened 2020-10-31
ItemList toNodes 

w-4:
Adds a `toNodes` function to `ItemList`. 

### DIFF
--- a/js/src/common/utils/ItemList.ts
+++ b/js/src/common/utils/ItemList.ts
@@ -1,3 +1,6 @@
+import listItems from '../helpers/listItems';
+import Mithril from 'mithril';
+
 class Item {
   content: any;
   priority: number;
@@ -127,5 +130,14 @@ export default class ItemList {
         return 1;
       })
       .map((item) => item.content);
+  }
+
+  /**
+   * Return a list of vnodes.
+   */
+  toNodes(): Mithril.Vnode[] {
+    const items = this.toArray();
+
+    return listItems(items);
   }
 }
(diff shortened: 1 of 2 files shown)


askvortsov1:
I know this is listed in flarum/framework#1317, but I'm not 100% sure about it since the ItemList util is essentially just an ordered map. I don't like adding methods that assume specific types of content, using `listItems` on the output of `toArray` seems cleaner to me. Thoughts @flarum/core @SychO9

clarkwinkelmann:
I'm neither particularly for or against this change.

But if we do it, we should probably deprecate `listItems` as proposed in the issue. It wouldn't make sense to have two different ways of doing the same thing.

`listItems` might have more use outside of `ItemList`s, but I don't remember seeing that in the wild.

If flarum/issue-archive#195 is put into action, then `listItems` will contain even less code, and start making less and less sense as its own util function. Since it seems we agree we want to make this other change, then I think moving everything to `toNodes` makes sense.

askvortsov1:
A few cases where we use listItems outside of ItemLists:

- https://github.com/flarum/core/blob/cee87848fe2c0793c930a1eb9909ceb17524ca62/js/src/common/components/Alert.tsx#L43-L48
- https://github.com/flarum/core/blob/cee87848fe2c0793c930a1eb9909ceb17524ca62/js/src/common/components/Dropdown.js#L36-L40
- https://github.com/flarum/core/blob/cee87848fe2c0793c930a1eb9909ceb17524ca62/js/src/common/components/FieldSet.js#L15-L20

And I maintain my point that the item list is currently content agnostic. If we want to make this change, it should be done on a subclass of ItemList IMO (which I think is unnecessary complexity)

w-4:
How about instead of the `toNodes` method, we make `listItems` accept both an array and the ItemList?
```js
 if (items instanceof ItemList) items = items.toArray()
```

askvortsov1:
Hmm, I have mixed feelings about changing variable type at runtime unnecessarily. I'm still unsure what benefit this change would bring.

askvortsov1:
SInce there hasn't been activity of strong support/argument for this change, I'm going to close this PR for now for the reasons I've listed above.

PR #2432, opened 2020-11-01
Add Notification Channel Extender

### LINKED ISSUE #2419: Notification Channel Extender
Part of #1891 

The following changes are needed IMO:

1. A "NotificationDriverInterface" interface should be created.
2. Implementations of this driver for sending regular notifications and mail notifications should be created
3. In NotificationSyncer, https://github.com/flarum/core/blob/34e98c884fe47537744d0567171dc9fde505fddc/src/Notification/NotificationSyncer.php#L105-L111 should be replaced with code that calls all registered notification drivers. Logic for deciding whether to send a notification and for enqueuing jobs should be part of the drivers
4. Calls for `event(new Sending($this->blueprint, $this->recipients));` should be moved from `SendNotificationsJob` to `NotificationSyncer`, and deprecated (and the event should be deprecated too)
5. An extender to add new notification drivers  should be added `(new Extend\Notification)->driver(driver_name, CustomNotificationsDriver);`, with integration tests

**Up For Discussion**

In `ConfigureNotificationTypes`, we allow the type to specify which drivers/channels it should be enabled for by default. Do we want to do a similar thing here, where channels can provide an array of types that they should be enabled for by default? If so, `NotificationServiceProvider` might need to be refactored. @flarum/core

SychO9:
**Fixes #2419**

* Use drivers for the different channels, each driver takes care of enqueing the notification job and adding the user preference.
* Add an extender for adding new drivers.
* Deprecate `Flarum\Notification\Event\Sending` Event.
* Add some dockblocks for the notification extender.

* Are the drivers being properly registered.
* Class/Interface/Method naming.

**Confirmed**

**Required changes:**
- [x] Related core extension PRs: **flarum/pusher#28**

### DIFF
--- a/src/Extend/Notification.php
+++ b/src/Extend/Notification.php
@@ -16,11 +16,19 @@ class Notification implements ExtenderInterface
 {
     private $blueprints = [];
     private $serializers = [];
+    private $drivers = [];
 
     public function type(string $blueprint, string $serializer, array $channelsEnabledByDefault = [])
     {
         $this->blueprints[$blueprint] = $channelsEnabledByDefault;
         $this->serializers[$blueprint::getType()] = $serializer;
+
+        return $this;
+    }
+
+    public function driver(string $driverName, string $driver)
+    {
+        $this->drivers[$driverName] = $driver;
 
         return $this;
     }
@@ -33,6 +41,10 @@ class Notification implements ExtenderInterface
 
         $container->extend('flarum.api.notification_serializers', function ($existingSerializers) {
             return array_merge($existingSerializers, $this->serializers);
+        });
+
+        $container->extend('flarum.notification.drivers', function ($existingDrivers) {
+            return array_merge($existingDrivers, $this->drivers);
         });
     }
 }
--- a/src/Extend/Notification.php
+++ b/src/Extend/Notification.php
@@ -18,6 +18,14 @@ class Notification implements ExtenderInterface
     private $serializers = [];
     private $drivers = [];
 
+    /**
+     * @param string $blueprint The ::class attribute of the blueprint class.
+     *                          This blueprint should implement \Flarum\Notification\Blueprint\BlueprintInterface.
+     * @param string $serializer The ::class attribute of the serializer class.
+     *                           This serializer should extend from \Flarum\Api\Serializer\AbstractSerializer.
+     * @param array $channelsEnabledByDefault The channels enabled by default for this notification type
+     * @return self
+     */
     public function type(string $blueprint, string $serializer, array $channelsEnabledByDefault = [])
     {
         $this->blueprints[$blueprint] = $channelsEnabledByDefault;
@@ -26,6 +34,12 @@ class Notification implements ExtenderInterface
         return $this;
     }
 
+    /**
+     * @param string $driverName The name of the notification driver.
+     * @param string $driver The ::class attribute of the driver class.
+     *                       This driver should implement \Flarum\Notification\Driver\NotificationDriverInterface.
+     * @return self
+     */
     public function driver(string $driverName, string $driver)
     {
         $this->drivers[$driverName] = $driver;
--- a/src/Extend/Notification.php
+++ b/src/Extend/Notification.php
@@ -23,12 +23,12 @@ class Notification implements ExtenderInterface
      *                          This blueprint should implement \Flarum\Notification\Blueprint\BlueprintInterface.
      * @param string $serializer The ::class attribute of the serializer class.
      *                           This serializer should extend from \Flarum\Api\Serializer\AbstractSerializer.
-     * @param array $channelsEnabledByDefault The channels enabled by default for this notification type
+     * @param array $driversEnabledByDefault The names of the drivers enabled by default for this notification type.
      * @return self
      */
-    public function type(string $blueprint, string $serializer, array $channelsEnabledByDefault = [])
+    public function type(string $blueprint, string $serializer, array $driversEnabledByDefault = [])
     {
-        $this->blueprints[$blueprint] = $channelsEnabledByDefault;
+        $this->blueprints[$blueprint] = $driversEnabledByDefault;
         $this->serializers[$blueprint::getType()] = $serializer;
 
         return $this;
--- a/src/Extend/Notification.php
+++ b/src/Extend/Notification.php
@@ -23,7 +23,8 @@ class Notification implements ExtenderInterface
      *                          This blueprint should implement \Flarum\Notification\Blueprint\BlueprintInterface.
      * @param string $serializer The ::class attribute of the serializer class.
      *                           This serializer should extend from \Flarum\Api\Serializer\AbstractSerializer.
-     * @param array $driversEnabledByDefault The names of the drivers enabled by default for this notification type.
+     * @param string[] $driversEnabledByDefault The names of the drivers enabled by default for this notification type.
+     *                                       (example: alert, email).
      * @return self
      */
     public function type(string $blueprint, string $serializer, array $driversEnabledByDefault = [])
@@ -38,11 +39,12 @@ class Notification implements ExtenderInterface
      * @param string $driverName The name of the notification driver.
      * @param string $driver The ::class attribute of the driver class.
      *                       This driver should implement \Flarum\Notification\Driver\NotificationDriverInterface.
+     * @param string[] $typesEnabledByDefault The names of blueprint classes of types enabled by default for this driver.
      * @return self
      */
-    public function driver(string $driverName, string $driver)
+    public function driver(string $driverName, string $driver, array $typesEnabledByDefault = [])
     {
-        $this->drivers[$driverName] = $driver;
+        $this->drivers[$driverName] = [$driver, $typesEnabledByDefault];
 
         return $this;
     }
@@ -50,7 +52,17 @@ class Notification implements ExtenderInterface
     public function extend(Container $container, Extension $extension = null)
     {
         $container->extend('flarum.notification.blueprints', function ($existingBlueprints) {
-            return array_merge($existingBlueprints, $this->blueprints);
+            $existingBlueprints = array_merge($existingBlueprints, $this->blueprints);
+
+            foreach ($this->drivers as $driverName => [$driver, $typesEnabledByDefault]) {
+                foreach ($typesEnabledByDefault as $blueprintClass) {
+                    if (isset($existingBlueprints[$blueprintClass]) && (! in_array($driverName, $existingBlueprints[$blueprintClass]))) {
+                        $existingBlueprints[$blueprintClass][] = $driverName;
+                    }
+                }
+            }
+
+            return $existingBlueprints;
         });
 
         $container->extend('flarum.api.notification_serializers', function ($existingSerializers) {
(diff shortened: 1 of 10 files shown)


src/Extend/Notification.php
```diff
@@ -16,7 +16,16 @@ class Notification implements ExtenderInterface
...
     private $blueprints = [];
     private $serializers = [];
+    private $drivers = [];
 
+    /**
+     * @param string $blueprint The ::class attribute of the blueprint class.
+     *                          This blueprint should implement \Flarum\Notification\Blueprint\BlueprintInterface.
+     * @param string $serializer The ::class attribute of the serializer class.
+     *                           This serializer should extend from \Flarum\Api\Serializer\AbstractSerializer.
+     * @param array $channelsEnabledByDefault The channels enabled by default for this notification type
```
askvortsov1, line 26: maybe "names of the channels..."

askvortsov1: Also, might be good to standardize wording to "drivers"

src/Extend/Notification.php
```diff
@@ -25,6 +34,19 @@ public function type(string $blueprint, string $serializer, array $channelsEnabl
         return $this;
     }
 
+    /**
+     * @param string $driverName The name of the notification driver.
+     * @param string $driver The ::class attribute of the driver class.
+     *                       This driver should implement \Flarum\Notification\Driver\NotificationDriverInterface.
+     * @return self
+     */
+    public function driver(string $driverName, string $driver)
```
askvortsov1, line 43: Lets add another (optional) argument for "types enabled by default", that takes string names of blueprint classes?

src/Notification/Driver/NotificationDriverInterface.php
```diff
@@ -0,0 +1,34 @@
...
+
+namespace Flarum\Notification\Driver;
+
+use Flarum\Notification\Blueprint\BlueprintInterface;
+use Flarum\User\User;
+
+interface NotificationDriverInterface
+{
+    /**
+     * Enqueues the job for this notification.
```
askvortsov1, line 18: Using the queue is an implementation detail. I think this should be revised to just say "Conditionally sends a notification to users, generally using a queue" or something like that.

src/Notification/Driver/NotificationDriverInterface.php
```diff
@@ -0,0 +1,34 @@
...
+    public function send(BlueprintInterface $blueprint, array $users): void;
+
+    /**
+     * Logic for adding a user preference.
+     *
+     * @param string $blueprintClass
+     * @param bool $default
+     * @return void
+     */
+    public function addUserPreference(string $blueprintClass, bool $default): void;
```
askvortsov1, line 33: I think `registerType` would be better naming, since in theory other stuff could be done too.

src/Notification/Notification.php
```diff
@@ -266,4 +274,23 @@ private static function getBlueprintAttributes(BlueprintInterface $blueprint): a
...
         ];
     }
+
+    /**
+     * Adds a notification driver to the list.
+     *
+     * @param string $driverName
+     * @param NotificationDriverInterface $driver
+     */
+    public static function addNotificationDriver(string $driverName, NotificationDriverInterface $driver)
```
askvortsov1, line 284: I feel like all these might be better stored on the syncer class. It's needed there, but more importantly, `NotificationSyncer` and the drivers concern methods of delivery / distribution for notifications. `Notification` is just the data structure for persisting them, which is currently agnostic of method of delivery.

askvortsov1 (changes requested):

src/Extend/Notification.php
SychO9: ah yes, forgot about this.

src/Notification/Driver/NotificationDriverInterface.php
SychO9: good point.

src/Notification/Notification.php
SychO9: I actually had them in `NotificationSyncer` at first, but then moved them. Me overthinking stuff :D

src/Extend/Notification.php
SychO9: Yes, there are some places where it says channels and others where it says drivers. Would be confusing for those unfamiliar with the code.

askvortsov1:
One interesting point brought up by https://github.com/flarum/pusher/pull/28 is that each driver will end up queing its own, independent jobs. Do we want to split it up like this, or move the logic for running through the drivers to inside the event?

src/Extend/Notification.php
```diff
@@ -38,27 +39,38 @@ public function type(string $blueprint, string $serializer, array $channelsEnabl
...
+            return $existingBlueprints;
         });
 
         $container->extend('flarum.api.notification_serializers', function ($existingSerializers) {
             return array_merge($existingSerializers, $this->serializers);
         });
 
         $container->extend('flarum.notification.drivers', function ($existingDrivers) {
-            return array_merge($existingDrivers, $this->drivers);
+            return array_merge($existingDrivers, $this->drivers[0]);
```
askvortsov1: Isn't `$this->drivers[0]` just the first driver? I think you need `Arr::pluck` here

tests/integration/extenders/NotificationTest.php
```diff
@@ -59,7 +61,23 @@ public function notification_driver_exists_if_added()
...
+    {
+        $this->extend(
+            (new Extend\Notification())
+                ->type(CustomNotificationType::class, 'customSerializer')
+                ->driver('customDriver', CustomNotificationDriver::class, [CustomNotificationType::class])
+        );
+
+        $this->app();
+
+
```
askvortsov1: This feels like its missing an assert

askvortsov1 (changes requested):

src/Extend/Notification.php
SychO9: kinda late there, I already pushed a fix :laughing: but I didn't think of using `Arr::pluck`

src/Extend/Notification.php
```diff
@@ -70,7 +70,13 @@ public function extend(Container $container, Extension $extension = null)
         });
 
         $container->extend('flarum.notification.drivers', function ($existingDrivers) {
-            return array_merge($existingDrivers, $this->drivers[0]);
+            $drivers = [];
+
+            array_walk($this->drivers, function ($driverData, $driverName) use (&$drivers) {
```
askvortsov1, line 75: Yeah let's change this to Arr::pluck for simplicity, If that wasn't an option, I'd prefer to use `array_map` instead of walk to avoid unnecessary mutations: gotta cling to the little bits of functional programming we can! :D

src/Extend/Notification.php
askvortsov1: In that case, lets use array_map to avoid unnecessary mutations

SychO9: `array_map` will not do, since it doesn't pass the array keys, only the values, we need the keys. ($driverName => $driver).

we could use a simple foreach for better readability. Or just add another private array property to save the default enabled types for drivers and use that.

askvortsov1: Urgh I keep forgetting that... Maybe it'd be easier then to just save 2 different arrays, one for the drivers, another for the types.

SychO9: The amount of times I googled if array_map passed keys is insane, I've been doing it constantly for years lmao.

> Maybe it'd be easier then to just save 2 different arrays, one for the drivers, another for the types.

Yea let's do that then.

tests/integration/extenders/NotificationTest.php
```diff
@@ -37,6 +48,21 @@ public function notification_type_exists_if_added()
...
+    {
+        $this->extend((new Extend\Notification())->driver(
+            'customNotificationDriver',
+            CustomNotificationDriver::class
+        ));
+
+        $this->app();
+
+        $this->assertArrayHasKey('customNotificationDriver', NotificationSyncer::getNotificationDrivers());
+    }
```
askvortsov1, line 65: Any way we can test that we can add enabled types via a driver?

askvortsov1 (changes requested):

tests/integration/extenders/NotificationTest.php
SychO9: That's what I was trying to do, but I'm not sure yet, we need a way to access the enabled driver names for specific types, and that's only in the user preferences array which is protected.

askvortsov1: Since the code works by extending the `'flarum.notification.blueprints'` singleton, couldn't we resolve that and check its contents?

SychO9: Ah.. I did not know we had access to the container here.

askvortsov1 (approved):

clarkwinkelmann (commented):
Looking good code-wise.

I see the tests only test against the container bindings, not for the actual sending of notifications.

I'd love to see this tested with an actual use case, like the Telegram notifications of the Telegram login extension before being shipped, so we make sure we ship something that will be practical in the real world. Or was a test already performed on another real world extension?

askvortsov1:
Do you mean in the integration tests? If you mean manual testing, I'm planning to test this with my PWA extension before merging

SychO9:
there's technically also the pucher extension which I tested on (if that counts).

src/Notification/Driver/EmailNotificationDriver.php
```diff
@@ -0,0 +1,69 @@
...
+            }
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function registerType(string $blueprintClass, bool $default): void
+    {
+        if ((new ReflectionClass($blueprintClass))->implementsInterface(MailableInterface::class)) {
```
askvortsov1, line 61: Shouldn't this be \ReflectionClass?

askvortsov1: Nvm you're importing it at the top

src/Notification/NotificationServiceProvider.php
```diff
@@ -49,29 +65,22 @@ protected function setNotificationTypes()
...
-
-        if ((new ReflectionClass($blueprint))->implementsInterface(MailableInterface::class)) {
-            User::addPreference(
-                User::getNotificationPreferenceKey($type, 'email'),
-                'boolval',
-                in_array('email', $channelsEnabledByDefault)
+        foreach (NotificationSyncer::getNotificationDrivers() as $driverName => $driver) {
+            $driver->registerType(
+                $blueprint,
+                in_array($driverName, $driversEnabledByDefault)
```
askvortsov1, line 83: One thing that came up as I was working on my PWA extension: it would be preferable to pass in the array of drivers enabled by default here (either instead of, or in addition to, argument 2). That gives extensions a bit more flexibility (for example, I enable PWA by default if email is enabled by default).

askvortsov1 (changes requested):

src/Notification/NotificationServiceProvider.php
SychO9: Very well

askvortsov1 (approved):
I can confirm that this works with my PWA extension, for both preferences and actually sending notifications.

tankerkiller125 (approved):

PR #2437, opened 2020-11-03
Add Service Provider Extender

### LINKED ISSUE #2421: "Service Provider" extender
Part of #1891 

As the team has been slow to provide official extenders, many community extensions have started using `Flarum\Extend\ExtenderInterface` to register code that runs when flarum boots and has access to the `Container` instance. This isn't really the intended use case of extenders, but it's an important use case nevertheless.

Instead, what if we had a `ServiceProvider` extender, which would essentially just run a function with an instance of `Container` provided?

@clarkwinkelmann @luceos I know you've used this pattern in your extensions; how would you feel about something like this?

SychO9:
**Fixes #2421**

**Confirmed**

### DIFF
--- a/tests/integration/extenders/ServiceProviderTest.php
+++ b/tests/integration/extenders/ServiceProviderTest.php
@@ -30,7 +30,25 @@ class ServiceProviderTest extends TestCase
     /**
      * @test
      */
-    public function providers_order_is_correct()
+    public function providers_first_register_order_is_correct()
+    {
+        $this->extend(
+            (new Extend\ServiceProvider())
+                ->register(CustomServiceProvider::class)
+        );
+
+        $this->app();
+
+        $this->assertEquals(
+            'overriden_by_custom_provider_register',
+            $this->app->getContainer()->make('flarum.forum.middleware')
+        );
+    }
+
+    /**
+     * @test
+     */
+    public function providers_second_register_order_is_correct()
     {
         $this->extend(
             (new Extend\ServiceProvider())
@@ -41,7 +59,27 @@ class ServiceProviderTest extends TestCase
         $this->app();
 
         $this->assertEquals(
-            'overriden_by_provider_test__final',
+            'overriden_by_second_custom_provider_register',
+            $this->app->getContainer()->make('flarum.forum.middleware')
+        );
+    }
+
+    /**
+     * @test
+     */
+    public function providers_boot_order_is_correct()
+    {
+        $this->extend(
+            (new Extend\ServiceProvider())
+                ->register(ThirdCustomProvider::class)
+                ->register(CustomServiceProvider::class)
+                ->register(SecondCustomServiceProvider::class)
+        );
+
+        $this->app();
+
+        $this->assertEquals(
+            'overriden_by_third_custom_provider_boot',
             $this->app->getContainer()->make('flarum.forum.middleware')
         );
     }
@@ -53,15 +91,7 @@ class CustomServiceProvider extends AbstractServiceProvider
     {
         // First we override the singleton here.
         $this->app->extend('flarum.forum.middleware', function () {
-            return 'overriden_by_provider_test';
-        });
-    }
-
-    public function boot()
-    {
-        // Third we override one last time here, to make sure this is the final result.
-        $this->app->extend('flarum.forum.middleware', function ($forumRoutes) {
-            return 'overriden_by_provider_test__final';
+            return 'overriden_by_custom_provider_register';
         });
     }
 }
@@ -72,16 +102,18 @@ class SecondCustomServiceProvider extends AbstractServiceProvider
     {
         // Second we check that the singleton was overriden here.
         $this->app->extend('flarum.forum.middleware', function ($forumRoutes) {
-            if ($forumRoutes !== 'overriden_by_provider_test') {
-                throw new ProviderTestException('CustomServiceProvider did not override the singleton.');
-            }
-
-            return $forumRoutes;
+            return 'overriden_by_second_custom_provider_register';
         });
     }
 }
 
-class ProviderTestException extends \Exception
+class ThirdCustomProvider extends AbstractServiceProvider
 {
-    // ...
+    public function boot()
+    {
+        // Third we override one last time here, to make sure this is the final result.
+        $this->app->extend('flarum.forum.middleware', function ($forumRoutes) {
+            return 'overriden_by_third_custom_provider_boot';
+        });
+    }
 }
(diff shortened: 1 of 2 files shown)


tests/integration/extenders/ServiceProviderTest.php
```diff
@@ -0,0 +1,54 @@
...
+use Flarum\Tests\integration\TestCase;
+use Flarum\Extend;
+use Illuminate\Contracts\Container\BindingResolutionException;
+
+class ServiceProviderTest extends TestCase
+{
+    /**
+     * @test
+     */
+    public function provider_doesnt_work_by_default()
```
askvortsov1, line 22: I think the important thing to check in these tests is the order in which core service providers, service provider register / extenders, and service provider boot resolve. Perhaps:

In one service provider, have `register` override some (existing) singleton from core, and check that that is overriden
In another, have `boot` override it again and check that what `boot` does is the final result

askvortsov1 (changes requested):

clarkwinkelmann (approved):
Looking good code-wise.

Not sure we need a dedicated exception for the tests that doesn't contain any logic though? We could just throw an `\Exception` directly?

tests/integration/extenders/ServiceProviderTest.php
```diff
@@ -0,0 +1,87 @@
...
+    }
+}
+
+class SecondCustomServiceProvider extends AbstractServiceProvider
+{
+    public function register()
+    {
+        // Second we check that the singleton was overriden here.
+        $this->app->extend('flarum.forum.middleware', function ($forumRoutes) {
+            if ($forumRoutes !== 'overriden_by_provider_test') {
```
askvortsov1, line 75: I would prefer to see all assertions within the test code. I get that this assesses stuff to ensure that the exception isn't thrown, but its not a clear use for a service provider, and its not immediately clear what its doing. If you want to test order, more (similar) test cases would be fine.

askvortsov1 (changes requested):

askvortsov1 (approved):
Looks good to me, thank you!

franzliedke:
Sorry I'm a bit late to this game, but I cannot help it...

I believe this extender is not a good idea, for a few reasons:
1. This does not seem use-case driven. Sure, somebody may have said "I wish I could register/replace/manipulate service providers". But that's not a purpose in itself. You always use service providers to do something - that something is interesting. It doesn't necessarily require messing with service providers, and if it does, it might warrant an explicit extender instead.
2. An even bigger problem, IMHO: it significantly increases what can be considered public API - with this extender, basically all of the service providers used by Flarum itself are now part of the public API - their existence, whatever they register and so on. But they are glue code, they can be (re-)structured at will, and one of the main point of the extender API was to prevent such internal restructurings from breaking extensions.

If extension devs want to mess with service providers, they can do so by writing their own extenders. But doing so clearly signals "You're doing this at your own risk." That's good, if you ask me. :slightly_smiling_face:

askvortsov1:
No worries, your feedback is always appreciated and super valued!

I'm a bit torn on this one. On one hand, I definitely see where you're coming from, especially with the expanded public API point. But on the other hand, the purpose of an extender is to be used (and reused) by extensions to modify core. IMO, an extender is something that core (or any extension) provides to expose the public API of it's backend to other extensions. I also believe that when users build hacks or modifications around a product to get it to accomplish something, that's a pretty clear indication that the feature / capability in question is in need. Putting these things together:

- I am concerned that a large number of extensions need to use custom extenders for modifications that aren't really "extender territory"
- I do not think that custom extenders are the right way to provide this advanced access, as that further complicates extensions which genuinely publish extenders to, well, extend the extension (like we'll see when search drivers are implemented). These are 2 very different use cases, and should be provided through separate channels IMO.

To that end, I think that this PR is a necessary evil, but I would be in favor of adding a bunch of warnings to it disclaiming that the only guarunteed public API is the official core extenders, and that this is provided as a courtesy to extension developers for advanced usages done at one's own risk only. I would also like to monitor community usage of this extender (perhaps via `query.flarum.dev`?) to see if any of those usages should receive official public API extender support.

Thoughts @flarum/core @SychO9? Sorry for the excessive tag, but this is an important side of this issue that we should come to a consensus on before b15 is released.

SychO9:
@askvortsov1 Why would an extension developer *need* to create a custom extender (class implementing the `ExtenderIntefrace`) just to have access to the container ? They can just use a closure to directly manipulate the container, or register a service provider from there.

As far as I know there is no difference in the order of execution between using a custom extender and a closure where you either register a service provider or directly do things and a registered service provider. A service provider has the added ability of running code after the app is booted through the boot method.

```php
return [
    function (Application $app, Container $container, ...$otherDependencies) {
        $container->bind(...);

        $app->register(CustomServiceProvider::class);
    },
];
```

askvortsov1:
The closure system is deprecated and removed with stable

SychO9:
I see, I wasn't aware of that. Then I would have to agree with this:

> To that end, I think that this PR is a necessary evil, but I would be in favor of adding a bunch of warnings to it disclaiming that the only guarunteed public API is the official core extenders, and that this is provided as a courtesy to extension developers for advanced usages done at one's own risk only.

We would need to make sure to properly document this.

It would be a mess if we started creating custom extenders, when they are not actually for extending the extension in question. One that comes to mind is the pusher extension.

Also, I'm not entirely sure that leaving extension developers to use custom extenders does clearly signal to them that they're doing it at their own risk.

askvortsov1:
> Also, I'm not entirely sure that leaving extension developers to use custom extenders does clearly signal to them that they're doing it at their own risk.

Either way, documentation that this has NO BC guaruntee is necessary, so we might as well make it easier (and less confusing with no need for custom, single-use extenders).

clarkwinkelmann:
Right now I have two main reasons I use providers/"local extenders":

- Grouping event listeners
- Registering libraries in the container

The first use case gradually fades away as more and more features are exposed through extenders instead of events.

The second use case however won't disappear. Service providers are an important part of the Laravel container. I like to define my own singletons, class aliases, and sometimes I intentionally want to override classes from Flarum or other extensions at the container level.

If callback-based "extenders" go away and "local extenders" are discouraged, there would be no official way to register service providers anymore. So we need this extender.

askvortsov1:
In that case, @franzliedke unless you have further objections I'll proceed with a PR that adds a warning that the service provider level is NOT officially considered public API, and is subject to unannounced BC breaks.

clarkwinkelmann:
We need a separate issue to discuss what we consider public API in the container. We've discussed things like `hasher` are intended to be overridable by extensions, so we probably should have a list of the container bindings we consider public, and have extensions consider anything else as private.

Access to the settings interface, translator interface, validator interface, hasher, should probably be considered for BC if we change them after stable.

PR #2438, opened 2020-11-05
ApiSerializer Extender

### LINKED ISSUE #1624: PHP Extender: Serializing / forum api attributes
Previously in order to append attributes to the forum object in js, one would listen to the Serializing event. This event still exists mind you. 

In flagrow/passport I would use this to add an attribute to the ForumSerializer that held the configured button title.

It would be useful to have an extender that allows setting attributes on returned payload. How about an `Extend\Serializer` class?

```php


    /**
     * @param Serializing $event
     */
    public function prepareApiAttributes(Serializing $event)
    {
        if ($event->isSerializer(ForumSerializer::class)) {
            $event->attributes = array_merge($event->attributes, [
                'flagrow.passport.loginTitle' => $this->settings->get('flagrow.passport.button_title', 'Login')
            ]);
        }
    }
```

SychO9:
**Part A of #1624**
**Part of #1891**

* Deprecate the `Serializing` event.
* Deprecate the `GetApiRelationshipEvent` event.
* Add attributes extenders.
* Add relationships extenders.
* Add Tests for all the new extenders.

* Are the tests on point.

**Confirmed**

**Bundled extensions to update:**
 + [x] **flarum/likes**: https://github.com/flarum/likes/pull/25
 + [x] **flarum/mentions**: https://github.com/flarum/mentions/pull/58
 + [x] **flarum/tags**: https://github.com/flarum/tags/pull/102
 + [x] **flarum/flags**: https://github.com/flarum/flags/pull/34
 + [x] **flarum/suspend**: https://github.com/flarum/suspend/pull/31
 + [x] **flarum/approval**: https://github.com/flarum/approval/pull/23
 + [x] **flarum/sticky**: https://github.com/flarum/sticky/pull/22
 + [x] **flarum/subscriptions**: https://github.com/flarum/subscriptions/pull/36
 + [x] **flarum/lock**: https://github.com/flarum/lock/pull/26
 + [x] **flarum/pusher**: https://github.com/flarum/pusher/pull/30


SychO9:
I think I'll split this into two PRs, one for adding the ApiSerializer Extender (this one) and one for the ApiController Extender.

src/Api/Serializer/AbstractSerializer.php
```diff
@@ -83,6 +94,16 @@ public function getAttributes($model, array $fields = null)
 
         $attributes = $this->getDefaultAttributes($model);
 
+        if (isset(static::$attributeHandlers[static::class])) {
+            foreach (static::$attributeHandlers[static::class] as $handler) {
```
SychO9, line 98: Ah I just noticed I need to check parent classes as well and a test case for that.

src/Api/Event/Serializing.php
```diff
@@ -17,6 +17,8 @@
  *
  * This event is fired when a serializer is constructing an array of resource
  * attributes for API output.
+ *
+ * @deprecated in beta 15, removed in beta 16
```
SychO9, line 21: almost forgot, I need to move the `formatDate` method as well, possibly in AbstractSerializer

askvortsov1 (changes requested):

src/Extend/ApiSerializer.php
```diff
@@ -0,0 +1,156 @@
...
+     * - $model: An instance of the model being serialized.
+     * - $serializer: An instance of this serializer.
+     *
+     * The callable should return:
+     * - An array of additional attributes to merge with the existing array.
+     *   Or a modified $attributes array.
+     *
+     * @return self
+     */
+    public function attributes($callback)
```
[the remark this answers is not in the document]
SychO9: it would have to be `attribute($nameOrCallback, $value = null)` because if you are going to use a callback, there is no point in providing an attribute name.

Or we could provide two methods, but I'm not sure about the naming

src/Extend/ApiSerializer.php
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    /**
+     * Add a setting value to the attributes of this serializer.
+     *
+     * @param string $key: The key of the setting. It should be prefixed with the extension ID. (ex: extension-id.setting_key).
+     * @param mixed $default: The default value of the setting.
+     * @return self
+     */
+    public function setting(string $key, $default = null)
```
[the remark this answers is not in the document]
SychO9: If we're gonna make a Settings extender, we might want to consider how (fof/extend)[https://github.com/FriendsOfFlarum/extend/blob/master/src/Extend/ExtensionSettings.php] does it:
```php
(new Extend\Settings())
    ->setPrefix('sycho-extension-id')
    ->addKey('first_setting_name', 'default')
    ->addKeys([
        'a_setting_name' => 'default',
        'another_setting_name' => true,
        'and_another_setting_name' => 'anotherDefault',
    ])
```

> I don't know that we need to serialize settings on any other serializer?

That's the point I brought up the other time on discord, I don't see why you'd need to attach a setting value on a serialized model.

I think we should remove the `setting` method from this ApiSerializer extender, and move this to a new issue ? Settings Extender.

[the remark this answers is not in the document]
askvortsov1: That is good for defaults, although not a big fan of the syntax. Also, I'd prefer that each default value be added on individually, and that the full setting name be used (not sure that registering the prefix is really necessary here...

Another concern is that the fof settings extender doesn't include anything for serializing settings to the frontend, which is a convenience method that I think should be on one of these 2 extenders.

src/Extend/ApiSerializer.php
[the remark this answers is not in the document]
askvortsov1: The callback would return the value assigned to the attribute. I'm thinking we do this kind of like the Model extender.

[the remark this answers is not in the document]
SychO9: but then wouldn't that mean, that if one chose to use an invokable class, and wanted to add multiple attributes they would have to create multiple classes instead of one ?

src/Extend/ApiSerializer.php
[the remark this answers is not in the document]
SychO9: Yes the convenience method is also good to have, I also noticed while looking at other extensions that some cast the value to a type or do somekind of processing, like json decoding it. So it'd be good to have something for that.

But either way, I think there is enough here to create a new extender/issue/pr :)

src/Extend/ApiSerializer.php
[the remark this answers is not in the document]
askvortsov1: True... I've been thinking that we might want to support laravel's [class, methodName] for callbacks.  @flarum/core any preference?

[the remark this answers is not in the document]
askvortsov1: Idea @SychO9:

- We have an `attribute` method that takes an attribute name and a callback (that takes the serializer instance and model instance), and assigns an attribute with the value returned by that callback. Alternatively, 
- We have a `mutate` method that takes the serializer instance and applies a random callback to it
- We implement `attribute` by assembling all the `attribute` calls within an extender into one mutator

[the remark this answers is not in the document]
SychO9: is this what you mean ?
```php
(new Extend\ApiSerializer())
    ->attribute('attrName', function ($attributes, $serializer, $model) {
        return 'a_value';
    })->mutate(function ($attributes, $serializer, $model) {
        $attributes['attr'] = 'a_value';
        $attributes['attr2'] = 'a_value2';

        // possibly some other logic..

        return $attributes;
    })
```

src/Api/Serializer/AbstractSerializer.php
```diff
@@ -280,4 +313,27 @@ public static function setContainer(Container $container)
...
+     * @param string $serializerClass
+     * @param callable $attributeHandler
+     */
+    public static function addAttributeHandler(string $serializerClass, callable $attributeHandler)
+    {
+        if (! isset(static::$attributeHandlers[$serializerClass])) {
+            static::$attributeHandlers[$serializerClass] = [];
+        }
+
+        static::$attributeHandlers[$serializerClass][] = $attributeHandler;
```
SychO9, line 327: Actually I don't think it's necessary to check ? php doesn't seem to throw an error.

src/Extend/ApiSerializer.php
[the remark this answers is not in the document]
askvortsov1: Yeah something like that!

src/Api/Serializer/AbstractSerializer.php
askvortsov1: Really? That's an interesting quirk. I'd say let's keep it in to be explicit though

tests/integration/extenders/ApiSerializerTest.php
```diff
@@ -0,0 +1,324 @@
...
+
+        $payload = json_decode($response->getBody(), true);
+
+        $this->assertArrayNotHasKey('customAttribute', $payload['data']['attributes']);
+    }
+
+    /**
+     * @test
+     */
+    public function custom_attribute_exists_if_added()
```
askvortsov1, line 78: I'd like to see some more tests for:

- Adding attributes via `attribute()` (with several calls)
- Being able to override attributes via `attribute()`
- Invokable class tests for `attribute()` and `mutate()`
- Prioritization of child `attribute` and `mutate`
- Support for inherited `attribute` and `mutate`

src/Extend/ApiSerializer.php
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    /**
+     * @param string $name: The name of the attribute.
+     * @param callable|string $callback
+     *
+     * The callback can be a closure or an invokable class, and should accept:
+     * - $attributes: An array of existing attributes.
+     * - $model: An instance of the model being serialized.
+     * - $serializer: An instance of this serializer.
```
askvortsov1, line 40: I feel like it would be nice to mention that $serializer has useful methods (like formatDate). Not major though.

askvortsov1 (changes requested):

tests/integration/extenders/ApiSerializerTest.php
SychO9: nice, the priority tests just showed that it was not actually being respected for attributes, had to reverse the classes array.

src/Extend/ApiSerializer.php
SychO9: isn't that a little too much ? IDEs would help with that :shrug:

askvortsov1: Yeah, that's true. We can mention it in the upgrade guide / docs though so people will know where to find the replacements for those.

askvortsov1 (approved):

tankerkiller125 (approved):

clarkwinkelmann (approved):
Just looking at the code looks good.

I just find it annoying sometimes to not have access to the actor in some contexts, but that was already from before this PR.

My only worry is that by using callbacks instead of event class, it makes it harder to add properties like `$actor` or `$request` later on without introducing a bunch of breaking changes...

SychO9:
wouldn't you still be able to do that if you used invokable classes ?

clarkwinkelmann:
> seems like you'd still be able to do $serializer->getActor() just like in the event listeners.

True, those additional parameters could be accessed from objects that are passed to the callback sometimes, so there wouldn't be an issue with adding more parameters to the callback here.

> are there arguments you're expecting to need to pass in, in the near future?

Except from actor/request for some contexts where none of the arguments allow to access them, I'm not seeing anything particular. I'm just seeing this as a potential issue for the future. Adding properties to an event class is easy, removing one only affects those that were using it. Removing properties from a callback will break every implementation that uses one of the arguments that follows.

Luckily PHP allows callback to accept less arguments than passed. Otherwise this would be a guaranteed breaking change anytime we add a new parameter, even if nobody wanted to use it.

SychO9:
Ah I see, that's a good point yeah :thinking:

askvortsov1:
We'll merge this a bit later in case @franzliedke has any comments / suggestions

src/Api/Serializer/AbstractSerializer.php
```diff
@@ -83,6 +94,18 @@ public function getAttributes($model, array $fields = null)
 
         $attributes = $this->getDefaultAttributes($model);
 
+        foreach (array_reverse(array_merge([static::class], class_parents($this))) as $class) {
+            if (isset(static::$mutators[$class])) {
+                foreach (static::$mutators[$class] as $callback) {
+                    $attributes = array_merge(
+                        $attributes,
+                        $callback($attributes, $model, $this)
+                    );
```
SychO9, line 103: would we prefer simply calling the callback instead of doing an array_merge here ? since extension devs can choose to take the $attributes parameter by reference.

src/Extend/ApiSerializer.php
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    public function extend(Container $container, Extension $extension = null)
+    {
+        if (! empty($this->attributes)) {
+            $this->mutators[] = function ($attributes, $serializer, $model) use ($container) {
+                foreach ($this->attributes as $attributeName => $callback) {
+                    $callback = ContainerUtil::wrapCallback($callback, $container);
+
+                    $attributes[$attributeName] = $callback($attributes, $serializer, $model);
```
SychO9, line 141: I noticed that most of the time when adding a single attribute through the `attribute` method of the extender, the first parameter being $attributes is rarely the one needed, so it's useless most of the time, do we want to push it to the end, or would we rather keep it consistent with the callback signature of the `mutate` method ?

SychO9 (commented):
a couple of things I thought about while updating bundled extensions.

src/Api/Serializer/AbstractSerializer.php
askvortsov1: From a functional programming perspective, I prefer return and merge over mutation

src/Extend/ApiSerializer.php
askvortsov1: I think consistency is better here; that being said, we could push it to the end for both `mutate` and `attribute`.

Issue #1675, opened 2018-12-02
Something went wrong: SQLSTATE[HY000] - MySQL: 5.7.23


terabytesoftw:
hi, installed flarum branch dev-master, 

`Something went wrong: SQLSTATE[HY000]: General error: 1215 Cannot add foreign key constraint (SQL: alter table `tag_user` add constraint `tag_user_tag_id_foreign` foreign key (`tag_id`) references `tags` (`id`) on delete cascade)`

The problem tables:

- migrations. --> MyIsam.
- tags --> MyIsam.

The rest od tables if you create in Innodb.

**Environment**
- Flarum version: dev-master
- Webserver: LiteSpeedTech
- Hosting environment: Shared.
- PHP version: 7.1

tobscure:
@franzliedke can you look at this in relation to https://github.com/flarum/core/commit/fb6b51b1cfef0af399607fe038603c8240800b2b?

terabytesoftw:
@tobscure its correct but tags is a separate extension and creates the tables as MyISAM, I proble flarum / core as dev-master and it did not work just those two tables (tags, migrations) all the others are like innodb.

tobscure:
The changes in fb6b51b should theoretically force InnoDB to be used for all tables, including those created by extensions, and the migrations table.

Are you experiencing this on a totally fresh install of dev-master?

terabytesoftw:
If a completely new installation of zero is correct from dev-master.

bobbwal:
I'm having the same issues on a fresh install. Even with the changes from [fb6b51b](https://github.com/flarum/core/commit/fb6b51b1cfef0af399607fe038603c8240800b2b) my migrations and tags are created as MyISAM.

franzliedke:
Ah, I'd figured it would be something like this. However, fb6b51b also sets InnoDB as engine for every table created through `Migration::createTable`. It turns out that the `tags` table is the only one not using that helper, which is why the default was necessary. That said, I just added an explicit InnoDB engine to that table's migration.

franzliedke:
I guess we could also add the explicit schema to the `migrations` repository: https://github.com/flarum/core/blob/796b57753d34d4ea741dbebcbc550b17808f6c94/src/Database/DatabaseMigrationRepository.php#L102-L105

@tobscure @luceos What do you think?

tobscure:
[...]

Issue #1661, opened 2018-11-29
Specify InnoDB as default table type


tobscure:
Since we now use foreign keys, we need to be explicit about wanting tables to be created using the InnoDB engine. MySQL can default to MyISAM in some circumstances, causing [errors like this](https://discuss.flarum.org/d/17736-beta8-install-sqlstate-hy000-general-error-1005).

Probably the best approach is to set `engine` in the database config that's passed to Laravel. We probably wouldn't want this in `config.php` though. Alternatively we could just set the `engine` in the migrations where we create tables?

franzliedke:
> We probably wouldn't want this in config.php though.

With that you mean it shouldn't be configurable, right?

franzliedke:
Setting `engine` in the DB config works for new installations. For old ones, we'll have to add a new migration that converts to InnoDB if necessary. Agreed?

Issue #677, opened 2015-12-16
MyISAM on posts table


Luceos:
I'm not sure, but I noticed that the posts table has MyISAM, the others have correctly InnoDB.

Luceos:
Seems this was a [decision](https://github.com/flarum/core/blob/master/migrations/2015_02_24_000000_create_posts_table.php#L37). Any reason why?

franzliedke:
Probably because of that FULLTEXT index right below that line.

Luceos:
I'm not sure that requires myisam.. [Right?](https://dev.mysql.com/doc/refman/5.6/en/innodb-fulltext-index.html)

franzliedke:
[Not in MySQL 5.5](https://dev.mysql.com/doc/refman/5.5/en/fulltext-search.html), though. That's what we support...

matteocontrini:
Hi everyone, any chance you could reconsider the decision of using MyISAM for the `posts` table? I'm not sure how much this will affect users, but maybe MySQL 5.5 support could be dropped. Since the release of MySQL 5.6, the FULLTEXT index is supported by InnoDB as well.

The fact that there's a MyISAM table among InnoDB table is causing me some headaches here because most backup solutions (including mysqldump) work best with InnoDB, and that MyISAM table is creating some troubles for achieving point-in-time backups.

Thank you.

tobscure:
@matteocontrini Yep in beta 8 the `posts` table will be converted to InnoDB. No more MyISAM 🎉

PR #2442, opened 2020-11-06
Remove MyISAM Requirement

ahosker:
**Fixes**
The requirement for Database to be MyISAM Compatible.

DigitalOcean Managed Databases do not have MyISAM enabled and there is no way of allowing it. The solution offered by the [forum ](https://discuss.flarum.org/d/23108-sql-general-error-3161-storage-engine-myisam-is-disabled/8) was to not elegant.

If this is not have breaking changes, it is an improvement to the installation process for those with Limited Database Engines.

Installation Changes
Edits Create Post, Migration table to remove hardcoded MyISAM requirement.
Edits Installation DatabaseConfig.php to remove hardcoded MyISAM requirement, leaving it up to the database to decide.

1) Can you just edit old Migration Files? It shouldn't affect existing installations and will allow easier to install new installations. That do not have MyISAM.

2) What effect does setting engine to NULL have in real terms. Was MyISAM selected by default for a reason? Would it be better to fallback if MyISAM is not available?

Issue #1661 and #677 and #1675 are loosely relevant

**Screenshot**

MyISAM Disabled

### DIFF
--- a/migrations/2015_02_24_000000_create_posts_table.php
+++ b/migrations/2015_02_24_000000_create_posts_table.php
@@ -30,8 +30,6 @@ return [
             $table->integer('hide_user_id')->unsigned()->nullable();
 
             $table->unique(['discussion_id', 'number']);
-
-
         });
 
         $connection = $schema->getConnection();
--- a/migrations/2015_02_24_000000_create_posts_table.php
+++ b/migrations/2015_02_24_000000_create_posts_table.php
@@ -31,7 +31,7 @@ return [
 
             $table->unique(['discussion_id', 'number']);
 
-            $table->engine = 'MyISAM';
+
         });
 
         $connection = $schema->getConnection();
--- a/src/Install/DatabaseConfig.php
+++ b/src/Install/DatabaseConfig.php
@@ -47,7 +47,7 @@ class DatabaseConfig implements Arrayable
             'collation' => 'utf8mb4_unicode_ci',
             'prefix'    => $this->prefix,
             'strict'    => false,
-            'engine'    => NULL,
+            'engine'    => null,
             'prefix_indexes' => true
         ];
     }
--- a/src/Install/DatabaseConfig.php
+++ b/src/Install/DatabaseConfig.php
@@ -47,7 +47,7 @@ class DatabaseConfig implements Arrayable
             'collation' => 'utf8mb4_unicode_ci',
             'prefix'    => $this->prefix,
             'strict'    => false,
-            'engine'    => 'InnoDB',
+            'engine'    => NULL,
             'prefix_indexes' => true
         ];
     }


tankerkiller125 (approved):
I am fine with making InnoDB the default, overall InnoDB has better performance anyways, I'm not sure if theirs a valid reason it defaults to MyISAM.

src/Install/DatabaseConfig.php
```diff
@@ -47,7 +47,7 @@ public function toArray()
             'collation' => 'utf8mb4_unicode_ci',
             'prefix'    => $this->prefix,
             'strict'    => false,
-            'engine'    => 'InnoDB',
+            'engine'    => NULL,
```
askvortsov1, line 50: Again, why mark this as null and not tell it to use innodb?

askvortsov1 (changes requested):
@tankerkiller125 Just history, we generally don't change old migration (although this seems to be a good reason to do so). 

Aside from the StyleCI complaint, a few minor questions below

tankerkiller125 (changes requested):
After further review (on an actual desktop this time) I think I'd actually rather see a migration or migrations to update existing Flarum installs to InnoDB if we do this at all.

As far as actually doing this though I'm not 100% sure now after further research MyISAM is far more performant when it comes to a lot of read operations. However InnoDB does much better with a mix. Given that Flarum is forum software I'm inclined to believe that InnoDB would be better for it, but I don't have data on SQL reads vs writes on a active Flarum install.

luceos:
For historical reasons, I'm going to pitch in.

The MyISAM requirement on the posts table was added to support Fulltext search. Which hasn't been supported by InnoDB for a very long time. We recently (ahem) updated our requirements to MySQL 5.6+ and Mariadb 10.0.5+, both of these versions now allow using fulltext indices to be created **with InnoDB**. So we can safely start doing this.

Two considerations:

- We had in mind to separate the migrations into two categories: installation migrations and updates migrations. The benefits of that is that we can create new forums faster and prevent an issue where running all migrations on new installs using xhr can kill the process - on short timeout servers - and cause partially installed databases and a ton of fallout.
- Running a migration to update to InnoDB on a multi million posts table is going to be wreak havoc.

tankerkiller125:
The last point from @luceos is a good point that running a migration against millions of post will be very bad. My proposal then would be to change all the current migrations to InnoDB (thus all new installs get that benefit) and then post a set of instructions to migrate existing databases manually (with a backup/restore of tables as part of that instruction set)

This ensures that new installs get the benefits, and if people want to take risk they can get them too. If they don't want to take risk though they can just simply leave things as is.

askvortsov1:
Not stale, but I propose we close this in favor of a discussion to work out exactly how we want to approach this.

migrations/2015_02_24_000000_create_posts_table.php
```diff
@@ -31,7 +31,7 @@
 
             $table->unique(['discussion_id', 'number']);
 
-            $table->engine = 'MyISAM';
```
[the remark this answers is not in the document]
ahosker: The issue I had was the Database did not support **MyISAM**. Wouldn't changing it to **Innodb** just bake in the same issue for a future version of me, with a Database that does not support **Innodb**. As such I determined that NULL (or no default) was the best option. Leaving it up to the Laravel Backend to determine.

src/Install/DatabaseConfig.php
ahosker: The issue I had was the Database did not support **MyISAM**. Wouldn't changing it to **Innodb** just bake in the same issue for a future version of me, with a Database that does not support **Innodb**. As such I determined that NULL (or no default) was the best option. Leaving it up to the Laravel Backend to determine.

ahosker:
This commit does resolve an install breaking issue. That is could have been resolved 3 months ago (how time flys), leading to (presumably) less install issues and greater adoption.

Committing **migrations/2015_02_24_000000_create_posts_table.php** is only removing the default on install. It does not as @tankerkiller125 outlines cause existing databases to be migrated to InnoDB or otherwise. Nor does it prevent a migration file at a later date to do that, if the community chooses.

I'd be repeating myself for Committing **src/Install/DatabaseConfig.php** this is (as far as I know) only used in install. It removes an install braking issue. It does not prevent future modifications as discussions continue.

I can view this as a Critical Commit but not as one that requires a prolonged discussion @askvortsov1 though there is, of course, discussions to be had related to this. Should MySSIAM be the default (instead of none)? Should old tables be migrated?

Thanks for the historical perspective @luceos a lot of thought went into these two lines in the past.

**TL'DR**. The guy who created the pull request obviously wants it committed and leave wider discussions on database engines to those who want to have them.

ahosker (commented):
Replacing Old Default with a New Default could create a future issue. Having no default allows the system to use the available Database Engine. This is a preference thing, and not firmly committed to no default if others believe a default database engine is preferable.

Other systems give the user the preference? Though would require further modificaitons.

luceos:
I'm not really one to look into our open PR's that much, apologies for that. Here's what I think if we consider adopting this right away.

- I see no changes that modify the migration that changes the posts table to MyISAM, so that migration will still be executed without need. This migration can be dropped. Existing installations won't complain and to new ones it doesn't matter. Alternatively modify that migration to check the engine in use and only run it when it's not MyISAM.
- Do we need a test that confirms the engine selected (or rather defaulted) is the one that is being used?


As to not merging this in the past couple of months; we are pushing all effort in reaching stable. These PR's seem minor, but they cannot and should not be merged without proper evaluation. Open source isn't just about accepting code by others, but also about having to deal with every single repercussion way past author involvement.

tankerkiller125:
> This commit does resolve an install breaking issue. That could have been resolved 3 months ago (how time flys), leading to (presumably) less install issues and greater adoption.

If your referring to fixing issues on DigitialOcean managed MySQL installs they have actually changed even more things in the past 3 months that permanently break installing Flarum that are well outside the scope of this projects goals.

askvortsov1:
> TL'DR. The guy who created the pull request obviously wants it committed. It fixed installs and can leave wider discussions on database engines to those who want to have them.

I apologize for my stark wording in the comment above; often PRs are opened, and then activity on them dies. In those cases, discussion can get lost (closed issues are  a better record of decision making closed PRs). I'm also sorry for our inactivity on this: we're in the middle of a large release, and our priority was stabilizing the extension API.

As Daniel noted above, we initially defaulted to MYISAM because InnoDB didn't support fulltext search. Now that it does, I am in favor of dropping the migration and changing the default. That being said, my main concern is that some databases could default to an engine that isn't InnoDB or MyISAM, and someone would end up with a partially broken Flarum install, as opposed to not being able to install due to requirements issues. That being said, I am not that familiar with how Laravel handles database engines (or the engines themselves for that matter), so I might be misinterpreting this.

luceos:
If my recommendation to drop the initial engine change is applied, I would love to see a core dev do a test run with:

- a clean install
- an existing database

Then I'd be okay with merging.

askvortsov1:
> an existing database

What'd be the benefit of that, since the migration is dropped?

luceos:
> What'd be the benefit of that, since the migration is dropped?

Yeah was hoping to confirm nothing breaks, but indeed the entry still exists in the database even though the migration files does not. So .. only a new install test would be required.

tankerkiller125 (changes requested):
I've tested a fresh install using this PR. And while everything works I have concerns over the lack of a proper default engine setting.

My main issue around it is one that's already been stated: If no default is set and then someone sets a very strange engine as default in MySQL it may cause install errors for Flarum. It would be by recommendation that we revert the change in `DatabaseConfig.php` and only change the migration.

As for the concern over a database not having InnoDB, given our current MySQL/MariaDB requirements this is incredibly unlikely unless someone built MySQL without it on-purpose. In which case that user probably has the technical ability required to resolve the issue (not to mention the issues they would get with other PHP projects by not having InnoDB)

luceos:
According to `vendor/doctrine/dbal/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php`:

```php
        // Engine
        if (! isset($options['engine'])) {
            $options['engine'] = 'InnoDB';
        }
```

Flarum will always fallback on InnoDB, unless `database.options.engine` is set in `config.php`.

tankerkiller125:
That works for me then.

tankerkiller125 (approved):

askvortsov1 (approved):
Before merging, let's replace the NULL with `null`, as that's more consistent with what we use.

src/Install/DatabaseConfig.php
luceos, line 50:
```suggestion
            'engine'    => null,
```

Convention to use lower case.

src/Install/DatabaseConfig.php
```diff
@@ -47,7 +47,7 @@ public function toArray()
             'collation' => 'utf8mb4_unicode_ci',
             'prefix'    => $this->prefix,
             'strict'    => false,
-            'engine'    => 'InnoDB',
+            'engine'    => null,
```
askvortsov1, line 50: Should we even specify this key if we're just setting it to null?

askvortsov1 (approved):

src/Install/DatabaseConfig.php
askvortsov1, line 50:
```suggestion
```

luceos: That's the convention in Laravel,  yes.

askvortsov1:
Thank you again for the PR, sorry for the long review process on this one.

PR #2443, opened 2020-11-07
Make checkbox switch component background stand out in modals

### LINKED ISSUE #2119: Switch component background merges with Modal background
The switch background when not turned on is the same as the modal background, making it very hard to see.

image

nina-py:
**Fixes #2119**

As discussed in the linked issue, make checkbox switch background white in modals as otherwise the component is barely discernible. Once I looked at the actual styles used, I decided that a better option would be to use the `body-bg` variable instead of hardcoding white. 

UI changes.

**Screenshot**

- Disabled checkbox switch now has body background (white by default) within modals

switch-off-on-modal

- Still turns green when on

switch-on-on-modal

- Still has control background outside of modals when off

switch-elsewhere

**Confirmed**

### DIFF
--- a/less/common/Checkbox.less
+++ b/less/common/Checkbox.less
@@ -31,6 +31,10 @@
   background: @control-bg;
   .transition(background-color 0.2s);
 
+  .Modal-content & {
+    background: @body-bg;
+  }
+
   .on& {
     background: #58a400;
   }
--- a/less/common/Checkbox.less
+++ b/less/common/Checkbox.less
@@ -31,10 +31,6 @@
   background: @control-bg;
   .transition(background-color 0.2s);
 
-  .Modal-content & {
-    background: @body-bg;
-  }
-
   .on& {
     background: #58a400;
   }
--- a/less/common/Modal.less
+++ b/less/common/Modal.less
@@ -106,12 +106,8 @@
     }
   }
 
-  .Checkbox--switch .Checkbox-display {
+  .off.Checkbox--switch .Checkbox-display {
     background: @body-bg;
-  }
-
-  .on.Checkbox--switch .Checkbox-display {
-    background: #58a400;
   }
 }
 .Modal-footer {
--- a/less/common/Modal.less
+++ b/less/common/Modal.less
@@ -105,6 +105,14 @@
       text-align: left;
     }
   }
+
+  .Checkbox--switch .Checkbox-display {
+    background: @body-bg;
+  }
+
+  .on.Checkbox--switch .Checkbox-display {
+    background: #58a400;
+  }
 }
 .Modal-footer {
   border: 0;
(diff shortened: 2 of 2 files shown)


less/common/Checkbox.less
```diff
@@ -31,6 +31,10 @@
   background: @control-bg;
   .transition(background-color 0.2s);
 
+  .Modal-content & {
+    background: @body-bg;
+  }
+
```
SychO9, line 37: For consistency, this should be added to the Modal.less file.It should also be applied to `.Modal-body` since that's the only part with the `@control-bg` background.

SychO9 (changes requested):
Thank you!

less/common/Checkbox.less
nina-py: Thanks @SychO9, agree on both points. Have made the modifications - please have a look. Note that due to the way LESS appends or prepends parent selectors the added styles in Modal.less have a bit more verbose than essentially a one-liner in Checkbox.less.

less/common/Modal.less
```diff
@@ -105,6 +105,14 @@
       text-align: left;
     }
   }
+
+  .Checkbox--switch .Checkbox-display {
```
SychO9, line 109: I think if we use `.off.Checkbox--switch .Checkbox-display` here instead we could probably avoid having to declaring the block under it :smiley:

less/common/Checkbox.less
SychO9: Thank you, I think we might be able to use a one liner it we used the `off` class, can you please test if that's the case ?

less/common/Modal.less
nina-py: Yep, just gave it a go - it works, thanks for the suggestion!

SychO9 (approved):
Looks good to me!

KyrneDev:
I'm not a huge fan of the white on the switch since the actual switching part is white as well. What about `@muted-more-color` for the background?

nina-py:
@KyrneDev, here's what it looks like with `@muted-more-color` rather than white:

muted-more-colour

askvortsov1:
Oh I like that better! Good idea

nina-py:
Updated the PR with the new background - commit message looks weird because background variable names were in backticks and git ate them :).

SychO9:
Since the checkbox switch doesn't currently have a disabled state CSS style, that looks better indeed.

> commit message looks weird because background variable names were in backticks and git ate them :).

keeps happening to me lol

askvortsov1:
The last thing I'd like to see before approving: how does this look in dark mode?

nina-py:
@askvortsov1, the dark mode looks like this (see screenshot below). Great with the `muted-more-color`, it would have been essentially black-on-black with the body background color.

dark-mode-switch

askvortsov1 (approved):

askvortsov1:
Thank you!

PR #2449, opened 2020-11-11
Introduce RequestUtil

### LINKED ISSUE #869: Switch to a custom subclass of ServerRequest

askvortsov1:
Fixes #869 

Abstracts access to following request attributes:

- actor
- session
- locale
- route name

Are there any attributes on here that we're missing (or that we should exclude)? $actor is the biggest one that needs to be abstracted away IMO, do we even need the others?

### DIFF
--- a/src/Http/RequestUtil.php
+++ b/src/Http/RequestUtil.php
@@ -15,7 +15,8 @@ use Psr\Http\Message\ServerRequestInterface as Request;
 
 class RequestUtil
 {
-    public static function getActor(Request $request): User {
+    public static function getActor(Request $request): User
+    {
         return $request->getAttribute('actor');
     }
 
--- a/src/Http/RequestUtil.php
+++ b/src/Http/RequestUtil.php
@@ -1,0 +1,56 @@
+<?php
+
+/*
+ * This file is part of Flarum.
+ *
+ * For detailed copyright and license information, please view the
+ * LICENSE file that was distributed with this source code.
+ */
+
+namespace Flarum\Http;
+
+use Flarum\User\User;
+use Illuminate\Contracts\Session\Session;
+use Psr\Http\Message\ServerRequestInterface as Request;
+
+class RequestUtil
+{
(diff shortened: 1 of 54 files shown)


SychO9 (commented):
Aren't all methods supposed to be static ?

askvortsov1:
Yeah, also I haven't changed any of the usages in the code yet. Actually I should have put this in the issue as a comment (since I just want to confirm which attributes we want to provide an API for) not as a PR, not sure what I was thinking

src/Http/RequestUtil.php
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+namespace Flarum\Http;
+
+use Flarum\User\User;
+use Psr\Http\Message\ServerRequestInterface as Request;
+
+class RequestUtil
+{
+    public static function getActor(Request $request): User
```
franzliedke, line 17: I have a branch lying around somewhere that basically does the same thing, but more along the lines of `Actor::fromRequest()`. I liked that a lot. Maybe you like the idea, too? 😄

franzliedke: Just saw more discussion in #2703 and related tickets. Have fun. 😀

askvortsov1: Actor is sensible too, but I prefer `RequestUtil` because:

- While getting it is clean, setting it results in clunky naming, as with the user method.
- If we need to encapsulate more stuff, we don't need to add more util classes / force people to import as much stuff. Perhaps separate classes is slightly better with respect to single responsibility, but I think encapsulating request data is still fairly granular. Better than adding it onto `User` anyhow :see_no_evil:

SychO9 (approved):

tankerkiller125 (approved):

PR #2451, opened 2020-11-12
ApiController Extender

SychO9:
**Part B of #1624**
**Part of #1891**

* Deprecate `WillGetData` Event.
* Deprecate `WillSerializeData` Event.
* Add extenders to replace both events above.

* I had to move all the methods in `WillGetData` to the `AbstractSerializeController`, the signature for those methods are somewhat incomplete, some could use proper type hinting while other could explicitly support mixed parameter types of `string|array` do we want to do that ?
* Do we want to add to the extender convenience methods, similar to the ones present in the `WillGetData` event for adding/removing includes ...etc ?

**Confirmed**

### DIFF
--- a/src/Api/Controller/AbstractSerializeController.php
+++ b/src/Api/Controller/AbstractSerializeController.php
@@ -117,7 +117,7 @@ abstract class AbstractSerializeController implements RequestHandlerInterface
         foreach (array_reverse(array_merge([static::class], class_parents($this))) as $class) {
             if (isset(static::$beforeSerializationCallbacks[$class])) {
                 foreach (static::$beforeSerializationCallbacks[$class] as $callback) {
-                    $data = array_merge($data, $callback($this, $data, $request, $document));
+                    $callback($this, $data, $request, $document);
                 }
             }
         }
--- a/src/Api/Controller/AbstractSerializeController.php
+++ b/src/Api/Controller/AbstractSerializeController.php
@@ -14,6 +14,7 @@ use Flarum\Api\Event\WillSerializeData;
 use Flarum\Api\JsonApiResponse;
 use Illuminate\Contracts\Container\Container;
 use Illuminate\Contracts\Events\Dispatcher;
+use Illuminate\Support\Arr;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 use Psr\Http\Server\RequestHandlerInterface;
@@ -83,18 +84,46 @@ abstract class AbstractSerializeController implements RequestHandlerInterface
     protected static $events;
 
     /**
+     * @var array
+     */
+    protected static $beforeDataCallbacks = [];
+
+    /**
+     * @var array
+     */
+    protected static $beforeSerializationCallbacks = [];
+
+    /**
      * {@inheritdoc}
      */
     public function handle(ServerRequestInterface $request): ResponseInterface
     {
         $document = new Document;
 
+        foreach (array_merge([static::class], class_parents($this)) as $class) {
+            if (isset(static::$beforeDataCallbacks[$class])) {
+                foreach (static::$beforeDataCallbacks as $callback) {
+                    $callback($this);
+                }
+            }
+        }
+
+        // Deprected in beta 15, removed in beta 16
         static::$events->dispatch(
             new WillGetData($this)
         );
 
         $data = $this->data($request, $document);
 
+        foreach (array_merge([static::class], class_parents($this)) as $class) {
+            if (isset(static::$beforeSerializationCallbacks[$class])) {
+                foreach (static::$beforeSerializationCallbacks as $callback) {
+                    $data = array_merge($data, $callback($this, $data, $request, $document));
+                }
+            }
+        }
+
+        // Deprecated in beta 15, removed in beta 16
         static::$events->dispatch(
             new WillSerializeData($this, $data, $request, $document)
         );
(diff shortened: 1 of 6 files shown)


src/Api/Controller/AbstractSerializeController.php
```diff
@@ -197,6 +226,106 @@ protected function buildParameters(ServerRequestInterface $request)
         return new Parameters($request->getQueryParams());
     }
 
+    /**
+     * Set the serializer that will serialize data for the endpoint.
+     *
+     * @param string $serializer
+     */
+    public function setSerializer($serializer)
```
askvortsov1, line 234: Perhaps all of these convenience methods should be added as extender methods, with an additional parameter being a callback that is evaluated to check whether the extension in question should be applied?

askvortsov1 (changes requested):

src/Api/Controller/AbstractSerializeController.php
SychO9: Callback would have to be optional I believe
```php
->setSerializer($serializerClass, $optionalCallback)
->addInclude('relation' | [...relations], $optionalCallback)
->removeInclude('relation' | [...relations], $optionalCallback)
->addOptionalInclude('relation' | [...relations], $optionalCallback)
...
```

askvortsov1: What is `'relation' | [...relations]` here? Wouldn't it just be one (string) value like it is in the current event signatures?

Other than that, this makes sense! I think we should also have a `mutate` method here, and package all these up under that, like with ApiSerializer.

SychO9: well, the signature is inconsistent, for instance, addInclude in the event allows either a string or an array of strings, while addOptionalInclude allows only a string, I would rather make them consistent and have them both allow `string|array`, in the controller and the extender.

> Other than that, this makes sense! I think we should also have a mutate method here, and package all these up under that, like with ApiSerializer.

mutate ? we have prepareDataQuery and prepareDataForSerialization, so I'm not sure where `mutate` would come in here?

askvortsov1: > well, the signature is inconsistent, for instance, addInclude in the event allows either a string or an array of strings, while addOptionalInclude allows only a string, I would rather make them consistent and have them both allow `string|array`, in the controller and the extender.

Ah got it, makes sense to me.

> mutate ? we have prepareDataQuery and prepareDataForSerialization, so I'm not sure where `mutate` would come in here?

Sorry I meant the general idea of a callback that's applied, it's covered by those.

SychO9:
Btw, the methods `removeInclude` `removeOptionalInclude` and `removeSortField` which were ported from the WillGetData event, I'm pretty sure they never actually worked, so I changed the code behind them (only the ones imported, the old ones I didn't).

`Arr::forget` removes items using their keys, but in this context we do not use keys, only values.

src/Api/Controller/AbstractSerializeController.php
```diff
@@ -197,6 +225,106 @@ protected function buildParameters(ServerRequestInterface $request)
...
+    }
+
+    /**
+     * Don't include the given relationship by default.
+     *
+     * @param string|array $name
+     */
+    public function removeInclude($name)
+    {
+        $this->include = array_diff($this->include, (array) $name);
```
askvortsov1, line 255: Wouldn't this (and `removeOptionalInclude` only remove includes that have been added in this extender, not includes that were present beforehand?

askvortsov1 (changes requested):

src/Api/Controller/AbstractSerializeController.php
```diff
@@ -197,6 +225,106 @@ protected function buildParameters(ServerRequestInterface $request)
         return new Parameters($request->getQueryParams());
     }
 
+    /**
+     * Set the serializer that will serialize data for the endpoint.
+     *
+     * @param string $serializer
+     */
+    public function setSerializer(string $serializer)
```
askvortsov1, line 233: Maybe we should add @internal to all these?

tests/integration/extenders/ApiControllerTest.php
```diff
@@ -0,0 +1,582 @@
...
+                'authenticatedAs' => 1,
+            ])->withQueryParams([
+                'sort' => 'userId',
+            ])
+        );
+
+        $payload = json_decode($response->getBody(), true);
+
+        $this->assertEquals(200, $response->getStatusCode());
+        $this->assertEquals(3, $payload['data'][0]['id']);
```
askvortsov1, line 468: Hmm can we pluck IDs and compare all IDs here? 

$this->assertEquals(EXPECTED_ORDER, Arr::pluck($payload['data'], 'id'));

tests/integration/extenders/ApiControllerTest.php
```diff
@@ -0,0 +1,582 @@
...
+        $response = $this->send(
+            $this->request('GET', '/api/discussions', [
+                'authenticatedAs' => 1,
+            ])
+        );
+
+        $payload = json_decode($response->getBody(), true);
+
+        $this->assertEquals(200, $response->getStatusCode());
+        $this->assertEquals(2, $payload['data'][0]['id']);
```
askvortsov1, line 534: Same here

askvortsov1 (changes requested):
Looking great! A few minor comments

src/Api/Controller/AbstractSerializeController.php
SychO9: I'm thinking there might be cases where these methods are directly used inside a callback passed to the extender's `prepareDataQuery`, not sure though

askvortsov1: Hmm fair point, someone might want to add variable imports.

askvortsov1 (approved):

SychO9 (commented):
Alright there's an issue I found while updating the bundled extensions to use this extender.

I passed an invokable class to `prepareDataForSerialization` (to replace the event listener) and the problem is that the class cannot receive the `$data` array by reference (which throws a php warning). 

That's because when it comes to invokable classes, they're not directly called, it's the callback that returns the invokable which is called. (ContainerUtil::wrapCallback).

I think we can work around this if we receive the function arguments by reference when wrapping the invokable class object.

I'll make a separate pull request for that.

askvortsov1:
Dumb question, but aren't all arrays received by reference on account of them being mutable objects? I thought PHPs pass by reference was only needed for primitives and strings

askvortsov1:
> Arrays are not considered objects, so just like primitives they are always passed by value and you have to explicitly specify that you would like to receive them by reference to modify them, otherwise you're just modifying a copy.

PHP why must you be like this :grimacing: 


> I passed an invokable class to `prepareDataForSerialization` (to replace the event listener) and the problem is that the class cannot receive the `$data` array by reference (which throws a php warning).

Alright, good catch! Could we add some tests that will ensure that this will work once we fix this in `ContainerUtil`?

SychO9:
> Alright, good catch! Could we add some tests that will ensure that this will work once we fix this in ContainerUtil?

Good idea!

> PHP why must you be like this :grimacing:

<details><summary>CLICK ME</summary>
<p>

www google com_search_q=why+can%27t+you+be+normal+meme+%22php%22 tbm=isch ved=2ahUKEwjSosH1jrXtAhWS7-AKHYRoBDsQ2-cCegQIABAA

</p>

SychO9:
One of the tests naturally fails, until #2485 is merged.

askvortsov1 (approved):

tankerkiller125 (approved):

PR #2452, opened 2020-11-12
Add Settings Extender

SychO9:
**Part of #1891**

Add an extender to expose settings to the frontend through the ForumSerializer.

Naming:
* Class: Settings ? Setting ?
* method: modifier ?

**Confirmed**


src/Extend/Settings.php
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    /**
+     * Set a default value for a setting selected for serialization.
+     *
+     * @param string $key
+     * @param $value
+     * @return $this
+     */
+    public function default(string $key, $value)
```
askvortsov1, line 46: For defaults, I feel like we should somehow modify `DatabaseSettingsRepository` or `MemoryCacheSettingsRepository`, instead of just configuring settings for serialization. Perhaps we could create a `DefaultSettingsManager` class (which could be injected into the settings repos) or even a class wrapping `MemoryCacheSettingsRepository`? That also begs the question: if a setting isn't found in the database and there's an extender-provided default but the user provides a default value in `get`, which should be used? I lean towards treating the extender-provided default as the actual value if there's no actual value regardless of what the user provides.

askvortsov1 (changes requested):

src/Extend/Settings.php
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    /**
+     * Modify a setting's value.
+     *
+     * @param string $key
+     * @param $callback
+     * @return $this
+     */
+    public function modifier(string $key, $callback)
```
askvortsov1, line 60: What's the use case for this?

askvortsov1 (changes requested):

src/Extend/Settings.php
SychO9: I like that, I was thinking that this would be somewhat misleading if all it does is fallback to a default value when serializing only.

a manager class to inject should be fine, not sure about the naming depending about what we decide in the below comment.

> That also begs the question: if a setting isn't found in the database and there's an extender-provided default but the user provides a default value in get, which should be used?

Logically, it should be
-> database value
-> default value provided in the extender
-> default value provided as a parameter

src/Extend/Settings.php
SychO9: There are some extensions that modify the value, for instance json decode it or cast it to a specific type. If we don't provide a way to do that, they'll just end up directly using the ApiSerializer extender and manually add them as attributes.

Would we want to also have this not only done while serializing, but in the repository level ? including it in the manager class we're talking about in the above comment. It does sound like a good idea to me.

askvortsov1: I'm still unsure about this one. For general purpose usage, I think providing this layer could be overengineering settings, at least right now. For serialization, the most I'd do is a optional callback that takes the setting as a third argument to `serializeToForum`, which, if provided, could perform some simple modification like you mentioned

SychO9: Alright let's keep it simple then

src/Extend/Settings.php
```diff
@@ -0,0 +1,87 @@
...
+                ForumSerializer::class,
+                function () use ($container) {
+                    $settings = $container->make(SettingsRepositoryInterface::class);
+                    $attributes = [];
+
+                    foreach ($this->settings as $key => $setting) {
+                        $value = $settings->get($key, null);
+
+                        if (isset($setting['callback'])) {
+                            $setting['callback'] = ContainerUtil::wrapCallback($setting['callback'], $container);
```
askvortsov1, line 75: Could we set this to a new `$callback` variable? I think it's a tiny bit cleaner that way

tests/integration/extenders/SettingsTest.php
```diff
@@ -0,0 +1,226 @@
...
+                ['key' => 'custom-prefix.custom_setting', 'value' => 'customValue'],
+                ['key' => 'custom-prefix.custom_setting2', 'value' => 'customValue']
+            ]
+        ]);
+    }
+
+    /**
+     * @test
+     */
+    public function custom_setting_doesnt_exist_by_default()
```
askvortsov1, line 37: Maybe "isn't serialized" instead of "doesnt exist"?

askvortsov1 (changes requested):
A few more nitpicks, but other than that, I really like this one!

askvortsov1 (approved):

tankerkiller125 (approved):

askvortsov1 (changes requested):
So I've been thinking more about this, and I'm no longer as sure as I was before. Sorry for the confusion and back-and-forth. The thing is, we already have a default settings mechanism: adding those values to the DB via a migration, which feels a bit faster. This new mechanism is more convenient, and makes sense, but I'd like to have a little bit more discussion, if only for the sake of devil's advocacy, before we implement it. How would you feel about splitting the default settings portion of this out into another PR (so this one only adds serialization), and revisiting the default part for beta 16?

SychO9:
:facepalm: I completely forgot about migrations. Also while updating all bundled extensions I didn't use the `default` method, precisely because the values were added through migrations.

> How would you feel about splitting the default settings portion of this out into another PR (so this one only adds serialization)

I don't even think we need to consider it at all unless we really feel the need to do something like this in the future.

The question is, are we fine with having the whole settings extender class with just the `serializeToForum` method ? (as far as I'm concerned, it's best this way than having a method in the ApiSerializer class. It's more verbose and clear as to what it is doing).

> Sorry for the confusion and back-and-forth

Never be sorry when it comes to what's best for the project :D, I can dance.

askvortsov1:
> The question is, are we fine with having the whole settings extender class with just the `serializeToForum` method ? (as far as I'm concerned, it's best this way than having a method in the ApiSerializer class. It's more verbose and clear as to what it is doing).

Yeah I think that's good, seeing as we really only want to serialize them to one serializer. If different use cases emerge, we can always reconsider later, but for now I think this is a good solution.

askvortsov1 (approved):

PR #2453, opened 2020-11-13
WIP: Split Search Filter and Search

askvortsov1:
**Start of flarum/issue-archive#286**

- Split search and filter into separate endpoints
- Add a "filterer" that applies filters and filter mutations

TODO:
- Add filter equivalent of default search gambits
- Convert bundled extensions to use this system
- Clean up code / improve abstractions
- Convert ListPostsController to use this system (in a separate PR)

- If we register things both as filters and as gambits for SimpleFlarumSearch, we're gonna have a lot of duplicated code. That being said, I'm not sure if there's anything we can do here if we want to truly separate the two
- For simplicity and to reduce diff size, I omitted some of the abstractions present in the old search system for filtering (SearchCriteria, etc). I also reused some SimpleFlarumSearch utils / components, like SearchResults, ApplySearchParametersTrait, AbstractSearch. What do we want to do about this? IMO simpler is better, especially seeing that custom filtration drivers aren't going to be a thing, so unnecessary abstraction is unnecessary. Did I cut out too much? Also, I want to eventually move the shared stuff into a Query namespace, but didn't do so here to minimize diff.

What if we used one endpoint, and searched vs filtered based on the presence of the `q` parameter? And what if SimpleFlarumSearch internally used filtration as its gambits (and all gambits were applied as key-values on the `filter` param instead of as part of the `q` string?

### DIFF
--- a/js/src/common/Store.js
+++ b/js/src/common/Store.js
@@ -83,7 +83,7 @@ export default class Store {
    */
   find(type, id, query = {}, options = {}) {
     let params = query;
-    let url = app.forum.attribute('apiUrl') + '/' + type;
+    let url = app.forum.attribute('apiUrl') + (query.search ? '/search/' : '/') + type;
 
     if (id instanceof Array) {
       url += '?filter[id]=' + id.join(',');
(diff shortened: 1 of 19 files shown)


askvortsov1:
Superceded by https://github.com/flarum/core/pull/2454

PR #2456, opened 2020-11-17
Slug Driver Support

tankerkiller125:
**Partially Solves** flarum/issue-archive#203

- Implements drivers for slugging for discussions and users, with the option to update slugs for other models as well.

- Are we happy with this?

**Confirmed**

**Required changes:**
- [ ] Related core extension PRs: https://github.com/flarum/lang-english/pull/179

### DIFF
--- a/js/src/admin/components/BasicsPage.js
+++ b/js/src/admin/components/BasicsPage.js
@@ -25,10 +25,6 @@ export default class BasicsPage extends Page {
       'welcome_message',
       'display_name_driver',
     ];
-    this.values = {};
-
-    const settings = app.data.settings;
-    this.fields.forEach((key) => (this.values[key] = Stream(settings[key])));
 
     this.localeOptions = {};
     const locales = app.data.locales;
@@ -42,7 +38,28 @@ export default class BasicsPage extends Page {
       this.displayNameOptions[identifier] = identifier;
     }, this);
 
+    this.slugDriverOptions = {};
+    Object.keys(app.data.slugDrivers).forEach((model) => {
+      this.fields.push(`slug_driver_${model}`);
+      this.slugDriverOptions[model] = {};
+
+      app.data.slugDrivers[model].forEach((option) => {
+        this.slugDriverOptions[model][option] = option;
+      });
+    });
+
+    this.values = {};
+
+    const settings = app.data.settings;
+    this.fields.forEach((key) => (this.values[key] = Stream(settings[key])));
+
     if (!this.values.display_name_driver() && displayNameDrivers.includes('username')) this.values.display_name_driver('username');
+
+    Object.keys(app.data.slugDrivers).forEach((model) => {
+      if (!this.values[`slug_driver_${model}`]() && 'default' in this.slugDriverOptions[model]) {
+        this.values[`slug_driver_${model}`]('default');
+      }
+    });
 
     if (typeof this.values.show_language_selector() !== 'number') this.values.show_language_selector(1);
   }
@@ -132,20 +149,30 @@ export default class BasicsPage extends Page {
               ]
             )}
 
-            {Object.keys(this.displayNameOptions).length > 1
-              ? FieldSet.component(
-                  {
-                    label: app.translator.trans('core.admin.basics.display_name_heading'),
-                  },
-                  [
-                    <div className="helpText">{app.translator.trans('core.admin.basics.display_name_text')}</div>,
-                    Select.component({
-                      options: this.displayNameOptions,
-                      bidi: this.values.display_name_driver,
-                    }),
-                  ]
-                )
-              : ''}
+            {Object.keys(this.displayNameOptions).length > 1 ? (
+              <FieldSet label={app.translator.trans('core.admin.basics.display_name_heading')}>
+                <div className="helpText">{app.translator.trans('core.admin.basics.display_name_text')}</div>
+                <Select
+                  options={this.displayNameOptions}
+                  value={this.values.display_name_driver()}
+                  onchange={this.values.display_name_driver}
+                ></Select>
+              </FieldSet>
+            ) : (
+              ''
+            )}
+
+            {Object.keys(this.slugDriverOptions).map((model) => {
+              const options = this.slugDriverOptions[model];
+              if (Object.keys(options).length > 1) {
+                return (
+                  <FieldSet label={app.translator.trans('core.admin.basics.slug_driver_heading', { model })}>
+                    <div className="helpText">{app.translator.trans('core.admin.basics.slug_driver_text', { model })}</div>
+                    <Select options={options} value={this.values[`slug_driver_${model}`]()} onchange={this.values[`slug_driver_${model}`]}></Select>
+                  </FieldSet>
+                );
+              }
+            })}
 
             {Button.component(
               {
--- a/js/src/forum/resolvers/DiscussionPageResolver.ts
+++ b/js/src/forum/resolvers/DiscussionPageResolver.ts
@@ -2,15 +2,6 @@ import DefaultResolver from '../../common/resolvers/DefaultResolver';
 import DiscussionPage from '../components/DiscussionPage';
 
 /**
- * This isn't exported as it is a temporary measure.
- * A more robust system will be implemented alongside UTF-8 support in beta 15.
- */
-function getDiscussionIdFromSlug(slug: string | undefined) {
-  if (!slug) return;
-  return slug.split('-')[0];
-}
-
-/**
  * A custom route resolver for DiscussionPage that generates the same key to all posts
  * on the same discussion. It triggers a scroll when going from one post to another
  * in the same discussion.
@@ -18,17 +9,32 @@ function getDiscussionIdFromSlug(slug: string | undefined) {
 export default class DiscussionPageResolver extends DefaultResolver {
   static scrollToPostNumber: string | null = null;
 
+  /**
+   * Remove optional parts of a discussion's slug to keep the substring
+   * that bijectively maps to a discussion object. By default this just
+   * extracts the numerical ID from the slug. If a custom discussion
+   * slugging driver is used, this may need to be overriden.
+   * @param slug
+   */
+  canonicalizeDiscussionSlug(slug: string | undefined) {
+    if (!slug) return;
+    return slug.split('-')[0];
+  }
+
+  /**
+   * @inheritdoc
+   */
   makeKey() {
     const params = { ...m.route.param() };
     if ('near' in params) {
       delete params.near;
     }
-    params.id = getDiscussionIdFromSlug(params.id);
+    params.id = this.canonicalizeDiscussionSlug(params.id);
     return this.routeName.replace('.near', '') + JSON.stringify(params);
   }
 
   onmatch(args, requestedPath, route) {
-    if (app.current.matches(DiscussionPage) && getDiscussionIdFromSlug(args.id) === getDiscussionIdFromSlug(m.route.param('id'))) {
+    if (app.current.matches(DiscussionPage) && this.canonicalizeDiscussionSlug(args.id) === this.canonicalizeDiscussionSlug(m.route.param('id'))) {
       // By default, the first post number of any discussion is 1
       DiscussionPageResolver.scrollToPostNumber = args.near || '1';
     }
--- a/src/Admin/Content/AdminPayload.php
+++ b/src/Admin/Content/AdminPayload.php
@@ -75,7 +75,7 @@ class AdminPayload
         $document->payload['extensions'] = $this->extensions->getExtensions()->toArray();
 
         $document->payload['displayNameDrivers'] = array_keys($this->container->make('flarum.user.display_name.supported_drivers'));
-        $document->payload['slugDrivers'] = array_map(function($resourceDrivers) {
+        $document->payload['slugDrivers'] = array_map(function ($resourceDrivers) {
             return array_keys($resourceDrivers);
         }, $this->container->make('flarum.http.slugDrivers'));
 
--- a/src/Admin/Content/AdminPayload.php
+++ b/src/Admin/Content/AdminPayload.php
@@ -75,6 +75,9 @@ class AdminPayload
         $document->payload['extensions'] = $this->extensions->getExtensions()->toArray();
 
         $document->payload['displayNameDrivers'] = array_keys($this->container->make('flarum.user.display_name.supported_drivers'));
+        $document->payload['slugDrivers'] = array_map(function($resourceDrivers) {
+            return array_keys($resourceDrivers);
+        }, $this->container->make('flarum.http.slugDrivers'));
 
         $document->payload['phpVersion'] = PHP_VERSION;
         $document->payload['mysqlVersion'] = $this->db->selectOne('select version() as version')->version;
--- a/src/Api/Controller/ShowDiscussionController.php
+++ b/src/Api/Controller/ShowDiscussionController.php
@@ -33,14 +33,14 @@ class ShowDiscussionController extends AbstractShowController
     protected $posts;
 
     /**
-     * {@inheritdoc}
-     */
-    public $serializer = DiscussionSerializer::class;
-
-    /**
      * @var SlugManager
      */
     protected $slugManager;
+
+    /**
+     * {@inheritdoc}
+     */
+    public $serializer = DiscussionSerializer::class;
 
     /**
      * {@inheritdoc}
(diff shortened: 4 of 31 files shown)


askvortsov1 (changes requested):
TODO:
- Add an extender for adding slugging drivers to models. Add integration tests for that extender
- Can we encapsulate any of the direct querying done in the drivers into the User / Discussion Repository classes, and inject those via DI?
- @flarum/core Confirm our stance on calling the full slug (whatever it might be) `slug`, and serializing it as such to the frontend. This conflicts with the current naming used by the tags extension (so that's a breaking change if we adopt it there), but it would be nice to use a common convention, and just `slug` seems sensible (I would also be fine with `fullSlug`, but Clark I know you have concerns about that, and I see where you're coming from).
- Decide whether we want to keep the `toResource` URL generator; if so, adopt throughout the core codebase where possible, and add an extender / tests for it
- Figure out what to do with frontend code that obtains discussion ID from the slug. This happens in 2 places:
    - DiscussionListItem's `isActive`, which can be overriden, but could also use `app.current.discussion` (which might make more sense).
    - DiscussionPageResolver's `getDiscussionIdFromSlug` helper. I don't think there's a way to universalize this one, but IMO we should move it onto the `DiscussionPageResolver` class so it can be overriden to generate the key in some different manner if needed by an extension.

tankerkiller125:
Regarding @askvortsov1 points above

- Will be done shortly
- Keeping it slug as `fullSlug` no longer makes sense given that drivers will allow devs to completely replace the slugging logic
- Keeping `toResource()`

tankerkiller125:
@askvortsov1 I'm going to rebase this tonight when I get home so we can fix whatever Admin stuff we need to (Admin UX got merged) and from there this PR is good to go if I'm not mistaken?

askvortsov1:
Yep! It should be good for full review then.

tankerkiller125:
Rebased this onto the master with the Admin UX changes, everything looks great and works correctly in my testing. We still need to create the translations however.

src/Admin/Content/AdminPayload.php
```diff
@@ -75,6 +75,9 @@ public function __invoke(Document $document, Request $request)
         $document->payload['extensions'] = $this->extensions->getExtensions()->toArray();
 
         $document->payload['displayNameDrivers'] = array_keys($this->container->make('flarum.user.display_name.supported_drivers'));
+        $document->payload['slugDrivers'] = array_map(function ($resourceDrivers) {
```
askvortsov1, line 78: Similarly to the internal discussion about moving backend pre-processing of extensions to a method in `AdminPayload`, this chunk of code would also be a good candidate for a `serializeSlugDrivers` method.

src/Http/HttpServiceProvider.php
```diff
@@ -25,5 +32,67 @@ public function register()
         $this->app->bind(Middleware\CheckCsrfToken::class, function ($app) {
             return new Middleware\CheckCsrfToken($app->make('flarum.http.csrfExemptPaths'));
         });
+
+        $this->app->singleton('flarum.http.slugDrivers', function () {
+            return [
+                Discussion::class => [
+                    'default' => IdWithSlugDriver::class
```
askvortsov1, line 39: Maybe we should rename this to `IdWithTransliteratedSlugDriver`?

src/Http/HttpServiceProvider.php
```diff
@@ -25,5 +32,67 @@ public function register()
...
+        });
+
+        $this->app->singleton('flarum.http.selectedSlugDrivers', function () {
+            $settings = $this->app->make(SettingsRepositoryInterface::class);
+
+            $compiledDrivers = [];
+
+            foreach ($this->app->make('flarum.http.slugDrivers') as $resourceClass => $resourceDrivers) {
+                $driverKey = $settings->get("slug_driver_$resourceClass", 'default');
+                if (array_key_exists($driverKey, $resourceDrivers)) {
```
askvortsov1, line 54: Maybe this could be cleaner as a ternary?

src/Http/UrlGenerator.php
```diff
@@ -59,4 +66,19 @@ public function to($collection)
...
+
+    /**
+     * Generate a URL to an instance of a resource.
+     *
+     * @param string $resourceClass
+     * @param AbstractModel $instance
+     * @param $args
+     * @return void
+     */
+    public function toResource(string $resourceClass, AbstractModel $instance, ...$args): string
```
askvortsov1, line 78: Do we want to explicitly pass in the resource type, or extrapolate via calling `get_class()` on the instance? `get_class` would be cleaner on invokation, but then we'd need to check all parent classes

src/User/UsernameSlugDriver.php
```diff
@@ -0,0 +1,26 @@
...
+class UsernameSlugDriver implements SlugDriverInterface
+{
+    public function toSlug(AbstractModel $instance): string
+    {
+        return $instance->username;
+    }
+
+    public function fromSlug(string $slug, User $actor): AbstractModel
+    {
+        return User::where('username', $slug)->whereVisibleTo($actor)->firstOrFail();
```
askvortsov1, line 24: Again, I think this would be better as a method on UserRepository

askvortsov1 (changes requested):

src/Discussion/IdWithSlugDriver.php
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    public function fromSlug(string $slug, User $actor): AbstractModel
+    {
+        if (strpos($slug, '-') == true) {
+            $slug_array = explode('-', $slug);
+            $slug = $slug_array[0];
+        }
+
+        return Discussion::where('id', $slug)->whereVisibleTo($actor)->firstOrFail();
```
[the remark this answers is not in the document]
tankerkiller125: Yeah, we can do that, it would make more sense as well.

src/User/UsernameSlugDriver.php
tankerkiller125: I agree, I'll update this

js/src/forum/resolvers/DiscussionPageResolver.ts
```diff
@@ -18,17 +9,32 @@ function getDiscussionIdFromSlug(slug: string | undefined) {
...
   static scrollToPostNumber: string | null = null;
 
+  /**
+   * Remove optional parts of a discussion's slug to keep the substring
+   * that bijectively maps to a discussion object. By default this just
+   * extracts the numerical ID from the slug. If a custom discussion
+   * slugging driver is used, this may need to be overriden.
+   * @param slug
+   */
+  canonicalizeDiscussionSlug(slug: string | undefined) {
```
tankerkiller125, line 19: @askvortsov1 Why did we need this? I don't remember, and looking at it the backend should handle a full slug including the transliteration correctly.

tankerkiller125:
We also have a frontend bug (I can't request changes on my own PR) 
image

The number is now appearing twice for some reason.

src/Http/UrlGenerator.php
tankerkiller125: Given that passing in the class type we expect causes PHP itself to go through the classes parents and stuff automatically in C, I don't see how using `get_class()` would be beneficial, especially since it would probably require more code than it's worth (slowing things down a bit)

askvortsov1 (approved):

src/Api/Controller/ShowDiscussionController.php
```diff
@@ -62,8 +68,9 @@ class ShowDiscussionController extends AbstractShowController
      * @param \Flarum\Discussion\DiscussionRepository $discussions
      * @param \Flarum\Post\PostRepository $posts
      */
-    public function __construct(DiscussionRepository $discussions, PostRepository $posts)
+    public function __construct(SlugManager $slugManager, DiscussionRepository $discussions, PostRepository $posts)
```
luceos, line 71: New attributes should always go to the end OR be sorted alphabetically:

- to the end because overrides need less changes, more extensibility
- alphabetically  because that seems to be the convention?!

src/Api/Controller/ShowDiscussionController.php
```diff
@@ -77,7 +84,11 @@ protected function data(ServerRequestInterface $request, Document $document)
         $actor = $request->getAttribute('actor');
         $include = $this->extractInclude($request);
 
-        $discussion = $this->discussions->findOrFail($discussionId, $actor);
+        if (Arr::get($request->getQueryParams(), 'bySlug', false)) {
+            $discussion = $this->slugManager->forResource(Discussion::class)->fromSlug($discussionId, $actor);
+        } else {
+            $discussion = $this->discussions->findOrFail($discussionId, $actor);
+        }
```
luceos, line 91: Looking at this I expect that we do this on every controller. I'd rather see (proposal):

- Repository::fromRequest method
- Repository loading the slug manager (this could even just be protected $slug = DiscussionSlug::class in the repository)
- Interface in Repository that implies it has sluggable resources to enforce specific methods?

This would reduce:

- Size of the controller and its bindings
- Increase size of the repository and the (late) bindings required; please note that the arguments in the controller methods are always instantiated soon, delegating this to the repository grants us a minor performance gain as well.
- Cleans up the conditional here, although it is moved elsewhere, we might even abstract the method `slugOrId` to an abstract repository or trait (SluggableRepository)

Let me know what you think.

src/Api/Serializer/BasicDiscussionSerializer.php
```diff
@@ -35,7 +46,7 @@ protected function getDefaultAttributes($discussion)
 
         return [
             'title' => $discussion->title,
-            'slug'  => $discussion->slug,
+            'slug' =>  $this->slugManager->forResource(Discussion::class)->toSlug($discussion),
```
luceos, line 49: Wouldn't it be easier to have a method on the abstract model instead?

```php
$discussion->slugger->slug();
```

So that we resolve the slug manager late here as well, plus we can reduce the overhead on every, single class that tries to resolve the slug manager..

src/Forum/Content/Discussion.php
```diff
@@ -106,6 +104,7 @@ public function __invoke(Document $document, Request $request)
      */
     protected function getApiDocument(User $actor, array $params)
     {
+        $params['bySlug'] = true;
```
luceos, line 107: Ehm... I am not sure this is wise; this looks a lot like a hard limitation on the code? This would practically make any extension have a hard time to do something different..

Thoughts?

luceos (changes requested):
This is looking amazing, I am literally stunned by the implementation. Well done!

I left some considerations, questions; please have a look.

src/Forum/Content/Discussion.php
tankerkiller125: Pinging @askvortsov1 on this one.

src/Api/Controller/ShowDiscussionController.php
askvortsov1: Yeah agreed, good catch

src/Api/Serializer/BasicDiscussionSerializer.php
askvortsov1: I'm a bit confused how that would work tbh. I'm very wary about adding stuff to AbstractModel unless it absolutely affects ALL  models, which isn't the case here.

src/Forum/Content/Discussion.php
askvortsov1: The way I see it, these `Content` classes are where we pre-assemble the page's data in the initial request so we don't need to request that data immediately from the frontend (plus adding SEO tags and all that jazz). So essentially, we're always pinging the API as if we're the frontend, so we should always be requesting stuff by slug (since we don't have any other info than just the URL). What extensions could this limit?

src/Api/Controller/ShowDiscussionController.php
askvortsov1: Mixed feelings about this. Currently, slug drivers use the repository. With this setup, the repository injects a slug manager which injects slug drivers. This feels convoluted to me. More importantly though, I feel like Flarum's internal use of repositories could really use some refinement: it's undocumented, and each repository seems to be assembled very haphazardly. That being said, I don't think this PR is the right place for that.

src/Api/Serializer/BasicDiscussionSerializer.php
luceos: What if it's a trait one can apply to the model, eg Discussion has the Sluggable trait. It adds that trait, sets the property `protected $slugger = DiscussionSlugger::class;` and the trait handles the rest?

js/src/admin/components/BasicsPage.js
```diff
@@ -42,8 +38,29 @@ export default class BasicsPage extends Page {
       this.displayNameOptions[identifier] = identifier;
     }, this);
 
+    this.slugDriverOptions = {};
+    Object.keys(app.data.slugDrivers).forEach((model) => {
+      this.fields.push(`slug_driver_${model}`);
+      this.slugDriverOptions[model] = {};
+
+      app.data.slugDrivers[model].forEach((option) => {
+        this.slugDriverOptions[model][option] = option;
```
SychO9, line 47: Are the keys getting localized somewhere ? I mean `default` and `idOnly` for example (same with display name drivers) They don't seem to be, at least the option's label not the key itself.

src/Http/HttpServiceProvider.php
```diff
@@ -25,5 +30,35 @@ public function register()
...
+
+        $this->app->singleton('flarum.http.selectedSlugDrivers', function () {
+            $settings = $this->app->make(SettingsRepositoryInterface::class);
+
+            $compiledDrivers = [];
+
+            foreach ($this->app->make('flarum.http.slugDrivers') as $resourceClass => $resourceDrivers) {
+                $driverKey = $settings->get("slug_driver_$resourceClass", 'default');
+
+                $driverClass = array_key_exists($driverKey, $resourceDrivers) ? $resourceDrivers[$driverKey] : $resourceDrivers['default'];
```
SychO9, line 53: the ternary looks unnecessary here? `$driverKey` falls back to the provided default `default` anyway.

SychO9 (commented):
Looking good! nice work!

I believe it'd be better to have `IdOnlyUserSlugDriver` here instead of the nicknames extension.

If I make for example another extension that adds a different user display name driver, the admin might want to use this `IdOnlyUserSlugDriver` but then I'd have to also add it to my own extension, because it only resides within the nickname extension.

Let me know what you think.

src/Http/HttpServiceProvider.php
askvortsov1: There's probably a cleaner way to write this. The default fallback in $driverKey assignment is based on the settings: which driver is selected to be active? But the ternary ensures that that driver is registered: if someone enables nicknames, sets the driver to a driver provided by nicknames, but then uninstalls nicknames, we don't want it to load up a nonexistent driver.

The ternary can be replaced with an `Arr::get`

js/src/admin/components/BasicsPage.js
askvortsov1: Nope, and mail drivers aren't either. That's a limitation worth revisiting later IMO

src/Discussion/IdWithTransliteratedSlugDriver.php
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    public function toSlug(AbstractModel $instance): string
+    {
+        return $instance->id.(trim($instance->slug) ? '-'.$instance->slug : '');
+    }
+
+    public function fromSlug(string $slug, User $actor): AbstractModel
+    {
+        if (strpos($slug, '-')) {
```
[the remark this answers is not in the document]
askvortsov1: Under this slug scheme, the ID prefix is always there and always necessary. The rest is optional and essentially just for decoration.

src/Api/Serializer/BasicDiscussionSerializer.php
askvortsov1: I don't like this. It adds unnecessary static stuff to the models, and I'd really prefer to move away from that as much as we possibly can because it's unnecessary leakage of global state. It also makes the extension mechanism a lot more hacky (similar to what I had to do with #2460), which I really didn't like and will revisit again after stable.

src/Http/HttpServiceProvider.php
SychO9: Oh you're right yes, I didn't notice. My bad.

askvortsov1 (changes requested):
> I believe it'd be better to have `IdOnlyUserSlugDriver` here instead of the nicknames extension.

On second thought, I agree with this, and that goes for `IdOnlyDiscussionSlugDriver` as well (and that doesn't break anything with the frontend stuff since the ID is still first).

js/src/admin/components/BasicsPage.js
SychO9: I see now, something like `app.translator.trans('core.admin.basics.slug_drivers.${option}')` wouldn't work, since extensions are adding drivers, and that means they couldn't add locales for the drivers anyway.

The solution would need to be more sophisticated. Yeah worth revisiting it later.

SychO9 (approved):
I have nothing else to add, it all looks good to me.

Just noting that we need of course to remember to add the id only drivers in core and to add the nicknames extension to composer.json

askvortsov1 (approved):

luceos (approved):

PR #2457, opened 2020-11-18
Update tsconfig file to include .tsx files

### LINKED ISSUE #2342: Convert all core JS to Typescript
This should be done in small, maximally isolated PRs: no more than a few files updated per PR so that we can slowly review and phase these in over time.

nina-py:
**Refs flarum/framework#3533**

Doesn't really **fix** the umbrella issue but is related to it.

Change `tsconfig.json` so that project-specific configuration is taken into account not just for `.ts` files but for `.tsx` files as well. Removing the file extension makes TS default to processing just `.ts`, `.tsx`, and `.d.ts`. files.

**Confirmed**

**Required changes:**

### DIFF
--- a/js/tsconfig.json
+++ b/js/tsconfig.json
@@ -1,5 +1,5 @@
 {
-  "include": ["src/**/*"],
+  "include": ["src/**/*.ts", "src/**/*.tsx"],
   "files": ["shims.d.ts"],
   "compilerOptions": {
     "allowUmdGlobalAccess": true,
--- a/js/tsconfig.json
+++ b/js/tsconfig.json
@@ -1,5 +1,5 @@
 {
-  "include": ["src/**/*.ts"],
+  "include": ["src/**/*"],
   "files": ["shims.d.ts"],
   "compilerOptions": {
     "allowUmdGlobalAccess": true,


datitisev:
I edited the description so the issue isn't closed if we merge this PR.

js/tsconfig.json
```diff
@@ -1,5 +1,5 @@
 {
-  "include": ["src/**/*.ts"],
+  "include": ["src/**/*"],
```
askvortsov1, line 2: This feels quite broad, essentially saying any file in src. Couldn't this cause files unrelated to JS/TS/TSX to be recognized incorrectly?

nina-py: @askvortsov1, according to the official documentation, this will actually do what we want it to do - parse only TypeScript files:

> If a glob pattern doesn’t include a file extension, then only files with supported extensions are included (e.g. .ts, .tsx, and .d.ts by default, with .js and .jsx if allowJs is set to true).
https://www.typescriptlang.org/tsconfig#include

Since allowJs is not explicitly set to true, this change will mean that TS will go from parsing just `.ts` files to `.ts`, `.tsx`, and `d.ts`, ignoring JS files and anything else that may be present in those directories.

askvortsov1 (approved):

nina-py:
Thanks for approving this @askvortsov1. I was thinking that perhaps it would be better to set the files to parse explicitly instead of following the TS docs - that way this config value would essentially become self-documenting:

```
"include": ["src/**/*.ts", "src/**/*.tsx"],
```

askvortsov1:
Ah yes that's even better!

nina-py:
@askvortsov1 done!

dsevillamartin (approved):

PR #2461, opened 2020-11-22
Policy Extender and Tests

askvortsov1:
Foundation for fixing #1832
Part of #1891

Please note that most of this diff is moving policy implementation content from the root folder of each service directory to an `Access` subfolder.

- Ask that policies return ->allow, ->deny, ->forceAllow, ->forceDeny instead of true/false to increase flexibility
- Improve evaluation of whether a user has the ability to do something on an instance to avoid order-dependence.
- Rename getPermission (->checkAbility)
- Collect policies via a singleton in user extender
- Move policy stuff into an `Access` sub-namespace
- Add policy extender and integration tests
- Deprecate old policy stuff

**Confirmed**

**Required changes:**


src/User/Access/AbstractPolicy.php
```diff
@@ -0,0 +1,62 @@
...
+    protected function forceDeny()
+    {
+        return static::FORCE_DENY;
+    }
+
+    /**
+     * @param GetPermission $event
+     * @return bool|void
+     */
+    public function checkAbility(User $actor, $ability, $instance)
```
SychO9, line 46: docblock doesn't match

SychO9 (changes requested):
Just some minor comments

~I think we should also mark the old `AbstractPolicy` class deprecated ?~ (nvm, we'd have to do that after deprecating `ScopeModelVisibility` event as well)

src/Extend/Policy.php
```diff
@@ -0,0 +1,54 @@
...
+        return $this;
+    }
+
+    public function extend(Container $container, Extension $extension = null)
+    {
+        $container->extend('flarum.policies', function ($existingPolicies) {
+            if (! array_key_exists($this->modelClass, $existingPolicies)) {
+                $existingPolicies[$this->modelClass] = [];
+            }
+            foreach ($this->addPolicies as $policy) {
```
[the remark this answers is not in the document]
askvortsov1: Ouch...

tests/integration/extenders/PolicyTest.php
```diff
@@ -0,0 +1,174 @@
...
+    public function unrelated_user_cant_hide_discussion_if_force_denied()
+    {
+        // Because the force deny policy is last, this also shows that
+        // order isn't considered, as the last result overrides all.
+        $this->extend(
+            (new Extend\Policy(Discussion::class))
+                ->add(CustomPolicy::class)
+                ->add(DenyHidePolicy::class)
+                ->add(ForceAllowHidePolicy::class)
+                ->add(ForceDenyHidePolicy::class)
```
SychO9, line 131: Should we have one test where they're scrambled? in all tests they seem to be registered in proper order (allow, deny, forceAllow, forceDeny).

tankerkiller125: I agree that we should scramble them on purpose, just to make sure things get registered properly regardless of order, never trust people inputs.

clarkwinkelmann (commented):
Looking good, I just can't figure out where the policy classes are resolved through the container.

Does it happen in `User::setGate($this->app->makeWith(Access\Gate::class, ['policies' => $this->app->make('flarum.policies.compiled')]));` ?

It would be ideal to only instantiate the class when interacting with that model.

askvortsov1:
> Looking good, I just can't figure out where the policy classes are resolved through the container.
> 
> Does it happen in `User::setGate($this->app->makeWith(Access\Gate::class, ['policies' => $this->app->make('flarum.policies.compiled')]));` ?
> 
> It would be ideal to only instantiate the class when interacting with that model.

It happens in `$this->app->make('flarum.policies.compiled')`

SychO9 (approved):
Looks good to me!

SychO9 (approved):

src/User/Access/Gate.php
```diff
@@ -0,0 +1,102 @@
...
+use Flarum\Database\AbstractModel;
+use Flarum\Event\GetPermission;
+use Flarum\User\User;
+use Illuminate\Contracts\Events\Dispatcher;
+use Illuminate\Support\Arr;
+
+class Gate
+{
+    protected const EVALUATION_CRITERIA = [
+        AbstractPolicy::FORCE_DENY => false,
```
clarkwinkelmann, line 21: I think a comment here should say that the order of this array is of the highest importance since it's the order they will be evaluated in.

Maybe the const could even be named `EVALUATION_CRITERIA_PRIORITY` or similar?

src/User/Access/Gate.php
```diff
@@ -0,0 +1,102 @@
...
+
+            foreach ($modelClasses as $class) {
+                $appliedPolicies = array_merge($appliedPolicies, Arr::get($this->policies, $class, []));
+            }
+        } else {
+            $appliedPolicies = Arr::get($this->policies, AbstractPolicy::GLOBAL);
+        }
+
+        foreach ($appliedPolicies as $policy) {
+            $results[] = $policy->checkAbility($actor, $ability, $model);
```
clarkwinkelmann, line 70: I still believe policy classes should only be resolved at this time to save on resources, skip resolving unnecessary constructor dependencies and delay resolves as much as possible.

clarkwinkelmann (changes requested):
Still haven't been able to test locally, but the code looks good. A few things.

src/User/Access/Gate.php
askvortsov1: Hmm, it was moved here to stop resolving them multiple times. Maybe we can resolve them here the first time needed and store that?

askvortsov1:
One other thought: do we want to support returning `true` instead of allow and `false` instead of deny? It would be a bit complex since we'd need to turn the former into the latter, but might make some cases easier.

clarkwinkelmann (changes requested):
Thanks. Your changes solve the problem of delaying resolve, but not of selectively resolving.

Most of the time, one request will only interact with the polices for one model, so it would be best to not resolve all policies for all existing models in the application.

What about calling a new method around line 82 which takes the list of class names, and takes/stores instances from an array which is keyed by the class name and serves as a cache?

Could we maybe even use the container itself to store instances? We could check if `Container::bound`, then `Container::make` and `Container::instance` store the policy.

Or alternatively, in the extender, register every policy as a singleton, then just use `make` in the Gate.

Not sure which of the options is more performant if you end up with a large number of policies.

clarkwinkelmann:
Regarding the `true` and `false` returns, if it makes things more complicated for us, I don't think we need to support it.

I think allowing them might make things ambiguous, because what would they map to?

However it would probably be great to have an error message if you return anything that's not `null` or one of the allowed values from a policy method, otherwise things would just fail silently and it could be a debugging nightmare.

askvortsov1:
Hmm, that makes sense. Making the policies singletons feels inefficient, but I see what you mean about resolving model by model. Will try that.

luceos (approved):

tankerkiller125 (approved):
Testing locally via GUI attempting to break things everything went well. We'll for sure want QA to test the permission stuff more during that process. But overall I'm happy with the results.

clarkwinkelmann (approved):
Code-wise this looks good. Would you like me to test locally with some other extensions?

Should we refactor the other core extensions to use it before merging, as to provide real world testing?

askvortsov1:
I already modified tags (there's an open PR for that). We could prep all the update PRs to include policies (@SychO9 has already set all of them up), and test on that one more time before merging?

PR #2463, opened 2020-11-22
User Preferences Extender and Tests

askvortsov1:
**Part of #1891**
**Replaces #2082**

Adds a simple extender and unit tests without any refactorings to the underlying preferences system. Since the public API of the extender remains unchanged, there isn't a necessity to implement refactorings in the same PR as the extender.

**Confirmed**


SychO9 (approved):

tankerkiller125 (approved):

clarkwinkelmann (commented):
Looking good, but I'm a bit confused with the naming.

"add"
"configure"
"register"

Wouldn't it make more sense to also call the extender method `addPreference` ?

askvortsov1:
I like register more than add, because that makes it clear that no values are going into the preference, and this is just configuration. Perhaps we could add a `registerPreference` static method to User and deprecate `addPreference`?

PR #2467, opened 2020-11-25
fix the ui modal problem in the change email

### LINKED ISSUE #1892: Error alert in modal remains after request succeeds
## Bug Report

**Current Behavior**
Errors handled with the modal handing that shows an Alert of type error remain after the request succeeds. I don't think there's a method to reset that, probably simply setting `this.alert = null` would be fine.

**Steps to Reproduce**
1. Open change email modal
2. Change email
3. Input wrong password
4. Input correct password

**Expected Behavior**
Error message disappears.

**Screenshots**
image
image

**Possible Solution**
Simply set `this.alert = null;` or add a method that does this - easier for extension developers to know it exists and use it.

azibom:
**Fixes #1892**

I just add the "this.alert = null" when i get the success response in the "ChangeEmailModal" component, and now if you write your password correctly, you do not see the last error alert.

### DIFF
--- a/js/src/forum/components/ChangeEmailModal.js
+++ b/js/src/forum/components/ChangeEmailModal.js
@@ -118,10 +118,7 @@ export default class ChangeEmailModal extends Modal {
           meta: { password: this.password() },
         }
       )
-      .then(() => (
-        this.success = true,
-        this.alert = null
-      ))
+      .then(() => ((this.success = true), (this.alert = null)))
       .catch(() => {})
       .then(this.loaded.bind(this));
   }
--- a/js/src/forum/components/ChangeEmailModal.js
+++ b/js/src/forum/components/ChangeEmailModal.js
@@ -118,7 +118,10 @@ export default class ChangeEmailModal extends Modal {
           meta: { password: this.password() },
         }
       )
-      .then(() => (this.success = true))
+      .then(() => (
+        this.success = true,
+        this.alert = null
+      ))
       .catch(() => {})
       .then(this.loaded.bind(this));
   }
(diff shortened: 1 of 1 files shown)


js/src/forum/components/ChangeEmailModal.js
```diff
@@ -118,10 +118,7 @@ export default class ChangeEmailModal extends Modal {
           meta: { password: this.password() },
         }
       )
-      .then(() => (
```
askvortsov1: Hmm, wouldn't it be simpler to do:

```
.then(() => {
    this.success = true;
    this.alert = null;
})
```

askvortsov1 (changes requested):

js/src/forum/components/ChangeEmailModal.js
azibom: yes, you are right, I will fix it

js/src/forum/components/ChangeEmailModal.js
```diff
@@ -118,7 +118,10 @@ export default class ChangeEmailModal extends Modal {
           meta: { password: this.password() },
         }
       )
-      .then(() => ((this.success = true), (this.alert = null)))
+      .then(() => {
+        this.success = true;
+        this.alert = null;
```
askvortsov1, line 123: This should be `this.alertAttrs = null`.

askvortsov1 (changes requested):

askvortsov1 (approved):

tankerkiller125 (approved):

franzliedke:
Random thought: should we remove the "old" error message at the *beginning* of the request already, to make it more obvious that a hypothetical new error is the result of the new request? The request can take any amount of time, so having the error still visible may confuse users. If I am not mistaken, that's a pattern I've often seen with client-side validation: remove errors when submitting, show them anew if they still appear.

:wave:

askvortsov1:
I like that approach! Is that something you'd like to PR, or should I open a new issue for it?

:wave:

PR #2468, opened 2020-11-25
AdminUX Overhaul Small Patch

KyrneDev:
**ref #2409**

2 small issues that were missed are patched in this PR. 

1. Extension list items were not hidden on tablets which made it look pretty bad, tablets are setup well to just use the widget.
2. Use the control color for `AdminHeader` descriptions to fit the color scheme

### DIFF
--- a/js/src/admin/components/AdminNav.js
+++ b/js/src/admin/components/AdminNav.js
@@ -22,12 +22,12 @@ export default class AdminNav extends Component {
   }
 
   onupdate() {
-    if ($('.Dropdown-menu').children('.active').length > 0) {
+    if ($('.Dropdown-menu').children('.active').length > 0 && app.previous.type) {
       $('#admin-navigation').animate(
         {
           scrollTop: $('.Dropdown-menu').children('.active')[0].offsetTop - 25,
         },
-        1000
+        500
       );
     }
   }
--- a/js/src/admin/components/AdminNav.js
+++ b/js/src/admin/components/AdminNav.js
@@ -26,7 +26,7 @@ export default class AdminNav extends Component {
     const nav = $('#admin-navigation');
     const time = app.previous.type ? 250 : 0;
 
-    if (children.length > 0 && (children[0].offsetTop > nav.scrollTop() || children[0].offsetTop + children[0].offsetHeight < nav.scrollTop())) {
+    if (children.length > 0 && (children[0].offsetTop > nav.scrollTop() + nav.outerHeight() || children[0].offsetTop + children[0].offsetHeight < nav.scrollTop())) {
       nav.animate(
         {
           scrollTop: children[0].offsetTop - nav.height() / 2,
--- a/js/src/admin/components/AdminNav.js
+++ b/js/src/admin/components/AdminNav.js
@@ -21,6 +21,14 @@ export default class AdminNav extends Component {
     );
   }
 
+  onupdate() {
+    if ($(".Dropdown-menu").children('.active').length > 0) {
+      $("#admin-navigation").animate({
+        scrollTop: $(".Dropdown-menu").children('.active')[0].offsetTop - 25
+      }, 1000);
+    }
+  }
+
   /**
    * Build an item list of main links to show in the admin navigation.
    *
@@ -28,6 +36,8 @@ export default class AdminNav extends Component {
    */
   items() {
     const items = new ItemList();
+
+    items.add('category-core', <h4 className="ExtensionListTitle">{app.translator.trans('core.admin.nav.categories.core')}</h4>);
 
     items.add(
       'dashboard',
@@ -88,7 +98,7 @@ export default class AdminNav extends Component {
     Object.keys(categorizedExtensions).map((category) => {
       if (!this.query()) {
         items.add(
-          category,
+          `category-${category}`,
           <h4 className="ExtensionListTitle">{app.translator.trans(`core.admin.nav.categories.${category}`)}</h4>,
           categories[category]
         );
--- a/js/src/admin/components/AdminNav.js
+++ b/js/src/admin/components/AdminNav.js
@@ -26,7 +26,10 @@ export default class AdminNav extends Component {
     const nav = $('#admin-navigation');
     const time = app.previous.type ? 250 : 0;
 
-    if (children.length > 0 && (children[0].offsetTop > nav.scrollTop() + nav.outerHeight() || children[0].offsetTop + children[0].offsetHeight < nav.scrollTop())) {
+    if (
+      children.length > 0 &&
+      (children[0].offsetTop > nav.scrollTop() + nav.outerHeight() || children[0].offsetTop + children[0].offsetHeight < nav.scrollTop())
+    ) {
       nav.animate(
         {
           scrollTop: children[0].offsetTop - nav.height() / 2,
--- a/js/src/admin/components/AdminNav.js
+++ b/js/src/admin/components/AdminNav.js
@@ -22,12 +22,16 @@ export default class AdminNav extends Component {
   }
 
   onupdate() {
-    if ($('.Dropdown-menu').children('.active').length > 0 && app.previous.type) {
-      $('#admin-navigation').animate(
+    const children = $('.Dropdown-menu').children('.active');
+    const nav = $('#admin-navigation');
+    const time = app.previous.type ? 250 : 0;
+
+    if (children.length > 0 && (children[0].offsetTop > nav.scrollTop() || children[0].offsetTop + children[0].offsetHeight < nav.scrollTop())) {
+      nav.animate(
         {
-          scrollTop: $('.Dropdown-menu').children('.active')[0].offsetTop - 25,
+          scrollTop: children[0].offsetTop - nav.height() / 2,
         },
-        500
+        time
       );
     }
   }
--- a/js/src/admin/components/AdminNav.js
+++ b/js/src/admin/components/AdminNav.js
@@ -21,7 +21,17 @@ export default class AdminNav extends Component {
     );
   }
 
+  oncreate(vnode) {
+    super.oncreate(vnode);
+
+    this.scrollToActive();
+  }
+
   onupdate() {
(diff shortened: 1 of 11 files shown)


less/admin/AdminNav.less
```diff
@@ -41,16 +41,10 @@
...
-  }
-
-  .ExtensionItem, .item-search {
-    display: none !important;
-  }
-
-  .ExtensionListTitle {
-    display: none !important;
+  .AdminNav {
+    .item-search, .ExtensionListTitle, li[class^="item-extension"] {
```
SychO9, line 45: could we also include `.AdminLinkButton-description` ? for this case scenario:

flarum lan_admin (8)

Also, I think we should probably do the same with `li` items for category names instead of targetting `ExtensionListTitle`, just so that their margins don't add needless width to the nav element :p

SychO9 (changes requested):

SychO9:
Oh and I just noticed something else

flarum lan_admin (9)

Looks like we need to give `.AdminLinkButton-description` `whitespace: normal`

And give the active item anchor a `@bg-color` color, instead of applying it to specifically the child items currently
```less
        &.active > a {
          background: @primary-color;
          font-weight: normal;

          .Button-label,
          .Button-icon {
            color: @body-bg;
            font-weight: bold;
          }
        }
```

Another thing, I noticed this in the code
```less
@media @desktop, @desktop-hd {
```
We can simplify if by using
```less
@media @desktop-up {
```

askvortsov1:
A few more suggestions:

- I think I prefer when the active color for link items in AdminNav is the same as the hover color (that light gray). I think it looks a bit more subtle (https://i.imgur.com/lP66s70.png vs https://i.imgur.com/EvhyCiq.png)
- I prefer `ExtensionListTitle` bottom margin to be 8px instead of 15px (https://i.imgur.com/HjbYcjp.png vs https://i.imgur.com/HIaJmeB.png)
- The core settings options feel a bit cramped towards the top. Perhaps we could add a category header there? (https://i.imgur.com/XygZXeI.png vs https://i.imgur.com/ueDdZKz.png)

Putting it all together: https://i.imgur.com/lv8INGu.png

KyrneDev:
Ideally we should leave this PR open for a week or so and see if we find anything else.

js/src/admin/components/AdminNav.js
```diff
@@ -21,6 +21,17 @@ export default class AdminNav extends Component {
     );
   }
 
+  onupdate() {
+    if ($('.Dropdown-menu').children('.active').length > 0 && app.previous.type) {
```
askvortsov1, line 25: Sorry for the confusion: on page reload, it should still scroll, but just without an animation

askvortsov1 (changes requested):

js/src/admin/components/AdminNav.js
```diff
@@ -21,6 +21,17 @@ export default class AdminNav extends Component {
...
+      $('#admin-navigation').animate(
+        {
+          scrollTop: $('.Dropdown-menu').children('.active')[0].offsetTop - 25,
```
askvortsov1, line 28: Perhaps instead of scrolling so the currently active one is at the top, it should scroll so it's in the middle of the screen? Also, I don't think it should scroll if it's already visible, only if it's not in the current viewport

js/src/admin/components/AdminNav.js
```diff
@@ -21,6 +21,17 @@ export default class AdminNav extends Component {
...
+        },
+        500
```
askvortsov1, line 30: This is still too long IMO

askvortsov1 (changes requested):

matteocontrini:
Wouldn't it be perhaps better to have the "No Settings" text (and similar) aligned to the left, instead of being centred? On large screens, at a first glance, it seems a bit "lost" in the UI. Also, when settings and permissions are actually there, they are aligned to the left.

Before:

image

After:

image

Maybe a more verbose message could also help (e.g. "This extension has no settings").

askvortsov1:
A few more comments:

- The Permissions page feels a little off: it feels like the group items are a bit too high up in the block, and a bit too much to the right. Also, maybe the grid itself should be indented a tad bit?

less/admin/AdminNav.less
```diff
@@ -107,22 +104,31 @@
...
           color: @text-color;
         }
+
         > a:hover {
           background: @control-bg;
         }
+
         &.active > a {
-          background: @primary-color;
+          background: @muted-color;
```
SychO9, line 131: this is technically a nitpick but a good one, can we use `@control-color`, they're the same colors, but best to use `@control-color` because that's the correct purpose here (so that in the future we are not restricted if we edit muted color)

SychO9: also, below this, can we move the `color: @body-bg` up from `.Button-label, .Button-icon` up here to the anchor, it's better and also fixes a problem with the BC extension page item when it's active. (I think I posted an image in a comment above)

less/admin/ExtensionWidget.less
```diff
@@ -3,8 +3,8 @@
   padding: 0;
 }
 
-.ExtensionsWidget-list {
-  > .container {
+.ExtensionsWidget {
+  .ExtensionsWidget-list {
```
SychO9, line 7: wouldn't just `.ExtensionWidget-list` be enough ? it's good practice to avoid over specification, to make customization easier for both admins and theme designers.

less/admin/AdminNav.less
```diff
@@ -180,6 +188,10 @@
 
 }
 
+.AdminLinkButton-description {
+  white-space: normal;
```
SychO9, line 192: I think adding 5px of left padding would be nice as well, to make the description symmetrical with the title.

SychO9 (changes requested):
This is looking so much better!

KyrneDev:
> A few more comments:
> 
> * The Permissions page feels a little off: it feels like the group items are a bit too high up in the block, and a bit too much to the right. Also, maybe the grid itself should be indented a tad bit?

Can you show me what you mean by this? I can't see myself.

askvortsov1:
> > A few more comments:
> > 
> > * The Permissions page feels a little off: it feels like the group items are a bit too high up in the block, and a bit too much to the right. Also, maybe the grid itself should be indented a tad bit?
> 
> Can you show me what you mean by this? I can't see myself.

It looks better if the PermissionsPage Button Group elements have their top padding set to 10px instead of 8px. Also, for `PermissionsPage-groups > container`, the side paddings are too large: it looks better with 10px left and right padding. Also, I think I prefer it when the actual PermissionsGrid element has a left padding of 10-15px.

askvortsov1 (changes requested):
Another suggestion from working on the nicknames extension: there should be an option to pass a callback to `registerSetting`, which has access to the page as the execution context, and could be used to add custom settings inputs or other stuff in between settings. https://github.com/flarum/core/issues/2484#issuecomment-737511267

js/src/admin/components/ExtensionPage.js
```diff
@@ -258,6 +241,10 @@ export default class ExtensionPage extends Page {
    * @returns {JSX.Element}
    */
   buildSettingComponent(entry) {
+    if (typeof entry === 'function') {
+      return entry.bind(this);
```
askvortsov1, line 245: THis just returns a bound function. I think you want `.call` here.

askvortsov1 (changes requested):
The "No Settings" and "No Permissions" text feels a bit too big. Maybe that could be made a bit smaller and/or shifted a tiny bit to the right?

less/admin/DashboardPage.less
```diff
@@ -34,6 +34,10 @@
       &.item-tools {
         float: right;
         margin-right: 0;
+
+        .Button {
+          background: @body-bg;
+        }
```
SychO9, line 40: I think this would be best added inside `.Widget` to make it more generic, not having to do the same thing again in case we or an extension adds a button inside a widget.

Also might be better to use `.Button--color(@control-color, @body-bg)` instead

SychO9 (commented):
It would be great to add a generic `AdminPage` to use with ExtensionPage and other core admin pages. We can then use that to add bottom paddings to the page by default.

Because extension permissions are stuck to the very bottom, and the basics page's save button as well.

askvortsov1:
> It would be great to add a generic `AdminPage` to use with ExtensionPage and other core admin pages. We can then use that to add bottom paddings to the page by default.

That's something I'd like to see too, and furthermore, we could move `buildSettingComponent` there, and streamline all the admin page code. But I think that's better for a separate PR, since this one is designed to fix and tweak the UX redesign

SychO9:
> That's something I'd like to see too, and furthermore, we could move buildSettingComponent there, and streamline all the admin page code. But I think that's better for a separate PR, since this one is designed to fix and tweak the UX redesign

FYI, I only meant a class for now, just so that we can add the bottom padding (we could add the padding differently, but I think this would be the best way)

And yes a component would be awesome, but that's something we can comeback to another time.

js/src/admin/components/AdminNav.js
```diff
@@ -21,6 +21,24 @@ export default class AdminNav extends Component {
     );
   }
 
+  onupdate() {
```
askvortsov1, line 24: Shouldn't this also run oncreate?

js/src/admin/components/AdminNav.js
```diff
@@ -21,6 +21,24 @@ export default class AdminNav extends Component {
...
+    const children = $('.Dropdown-menu').children('.active');
```
askvortsov1, line 25: Can we use `this.$` to restrict jquerying to this component's children?

askvortsov1 (changes requested):

js/src/admin/components/AdminNav.js
askvortsov1: Nvm, this breaks things.

askvortsov1 (approved):

SychO9 (approved):

PR #2472, opened 2020-11-29
update a user's comment count after deleting a discussion

### LINKED ISSUE #2227: comment_count doesn't get updated when deleting a discussion
## Bug Report

**Current Behavior**
When deleting the discussion, the comment count (`users.comment_count` in the database) doesn't get updated. Only after creating a new discussion, this will get a correct value again.

**Steps to Reproduce**
1. Log into an empty instance, as Admin
2. Create a new discussion
3. Check the profile page : should have **Posts 1** and **Discussions 1**.
4. Delete the discussion that was created (delete forever)
5. Check the profile page : should have **Posts 1** and **Discussions 0**.

**Expected Behavior**
At step 5. there should be  **Posts 0**

**Environment**
- Flarum version: 0.1.0-beta.13
- Website URL: localhost
- Webserver: nginx
- Hosting environment: local
- PHP version: 7.3.19
- Browser: Google Chrome Version 83.0.4103.116

_(I Hope this isn't a duplicate, I didn't find anything when searching for this issue...)_

sl-kr:
**Fixes #2227**

Remove type hinting from the `updateCommentsCount()` function in order to allow it to accept a Post or a Discussion as an argument. This allows you to call `updateCommentsCount()` from the `whenDiscussionWasDeleted()` listener.

The Post class is no longer used in this file, so I removed the use directive at the top of this file.

Without union types, there aren't many options for allowing a function to accept two different types of arguments. However, the only other option I could think of other than the one I proposed here is having two arguments with two different types that both default to null. Let me know if there is a better option!

**Confirmed**

Also, I noticed that if you delete the last post in the discussion, it deletes the discussion, but it does not update the user's discussion count. Should I create a separate issue for that?

### DIFF
--- a/src/User/UserMetadataUpdater.php
+++ b/src/User/UserMetadataUpdater.php
@@ -15,6 +15,7 @@ use Flarum\Discussion\Event\Started;
 use Flarum\Post\Event\Deleted as PostDeleted;
 use Flarum\Post\Event\Posted;
 use Illuminate\Contracts\Events\Dispatcher;
+use Flarum\User\User;
 
 class UserMetadataUpdater
 {
@@ -34,7 +35,7 @@ class UserMetadataUpdater
      */
     public function whenPostWasPosted(Posted $event)
     {
-        $this->updateCommentsCount($event->post);
+        $this->updateCommentsCount($event->post->user);
     }
 
     /**
@@ -42,7 +43,7 @@ class UserMetadataUpdater
      */
     public function whenPostWasDeleted(PostDeleted $event)
     {
-        $this->updateCommentsCount($event->post);
+        $this->updateCommentsCount($event->post->user);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/User/UserMetadataUpdater.php
```diff
@@ -60,9 +59,13 @@ public function whenDiscussionWasStarted(Started $event)
...
     {
         $this->updateDiscussionsCount($event->discussion);
+        $this->updateCommentsCount($event->discussion);
     }
 
-    private function updateCommentsCount(Post $post)
+    /**
+     * @param \Flarum\Discussion\Discussion|\Flarum\Post\Post $post
+     */
+    private function updateCommentsCount($post)
```
askvortsov1, line 68: Not sure about calling this argument `$post` if it's not always a post. Maybe we should just pass the user in directly.

askvortsov1 (approved):

src/User/UserMetadataUpdater.php
sl-kr: Good point!

askvortsov1: Apologies for the confusing approval, I clicked the wrong button. That was meant to request changes.

sl-kr: No worries!

askvortsov1 (approved):

tankerkiller125 (approved):

PR #2479, opened 2020-11-30
Use filterer for ListPostsController

askvortsov1:
**Part of flarum/framework#1891**
**Part of flarum/issue-archive#286**
**Part 2 of flarum/framework#2454**

- ListPostsController now uses the same extensible filter system as the other list controllers
- ConfigurePostsQuery has been deprecated in favor of filters and filter mutators
- Added integration tests for ListPostsController

**Confirmed**


SychO9 (commented):
Could you please update this one ?

src/Post/Filter/AuthorFilter.php
```diff
@@ -0,0 +1,45 @@
...
+    {
+        $this->users = $users;
+    }
+
+    public function getFilterKey(): string
+    {
+        return 'author';
+    }
+
+    public function filter(WrappedFilter $wrappedFilter, string $filterValue, bool $negate)
```
SychO9, line 36: `FilterState`* there is no `WrappedFilter` anymore :D

SychO9 (changes requested):

src/Post/Filter/TypeFilter.php
```diff
@@ -0,0 +1,28 @@
...
+use Flarum\Filter\FilterState;
+
+class TypeFilter implements FilterInterface
+{
+    public function getFilterKey(): string
+    {
+        return 'type';
+    }
+
+    public function filter(FilterState $FilterState, string $filterValue, bool $negate)
```
SychO9, line 22: nitpick: can you please change all occurrences of `$FilterState` to `$filterState`

SychO9 (changes requested):

src/Post/Filter/TypeFilter.php
askvortsov1: ... I thought I clicked that button in my find+replace tool... oh well...

src/Post/Filter/IdFilter.php
```diff
@@ -0,0 +1,32 @@
...
+
+    public function filter(FilterState $filterState, string $filterValue, bool $negate)
+    {
+        $idString = trim($filterValue, '"');
+        $ids = explode(',', $idString);
+
+        $filterState
+            ->getQuery()
+            ->whereIn('posts.id', $ids, 'and', $negate)
+            ->orderByRaw("FIELD(id, $idString)");
```
SychO9, line 30: I don't think this is necessary, and it will probably conflict with the order set by the request. Also we should avoid using raws, this for example will lead to `$idString` being evaluated in the sql query (sql injection), so I can do something like `?filter[id]=66,0),(created_at` which will also change the sql code to order by creation date.

SychO9 (changes requested):

src/Post/Filter/IdFilter.php
askvortsov1: Ah yes, the first law of web-based programming: don't insert user-provided content into raw DB queries. How did I not think of this... And you're right: we now sort the returned posts in the frontend anyways.

SychO9 (approved):

tankerkiller125 (approved):

PR #2481, opened 2020-12-01
Fixes issue with the worker defaulting to the illuminate queue manager

luceos:
We are instantiating our own queue handling factory which returns the
flarum.queue.connection binding no matter what. The queue Worker and
other queue related code rely on this manager to get its thing going.
Therefor we need to re-use our own factory everywhere, including in
the worker.


askvortsov1:
Tests are failing?

askvortsov1:
I suspect that the failing tests were fixed in https://github.com/flarum/core/commit/984f751c718c89501cc09857bc271efa2c7eea8c, rebasing might help here.

luceos:
Yeah that fixed it 👍

tankerkiller125 (approved):

askvortsov1 (approved):

PR #2483, opened 2020-12-02
Search Extender and tests

askvortsov1:
**Part of flarum/framework#1891**
**Part of flarum/issue-archive#286**
**Part 3 of flarum/framework#2454**

- All searchers now inherit from AbstractSearcher
- Gambit managers no longer internally use a container
- A `SimpleFlarumSearch` extender has been added
- Events for `Searching` and gambits have been deprecated

**Confirmed**


askvortsov1:
I'm still not fully convinced that we should have so much code duplication between gambits and SFS filters. That being said, using the same class for both is messy, and requires a bit of boilerplate. We'd also need to figure out which namespace to place them into (which could be a breaking change if other extensions are doing something via the current namespace).

src/Discussion/Event/Searching.php
```diff
@@ -12,6 +12,9 @@
 use Flarum\Discussion\Search\DiscussionSearch;
 use Flarum\Search\SearchCriteria;
 
+/**
+ * @deprecated beta 15, remove beta 16
```
tankerkiller125, line 16: Looks like the deprecation messages probably need updating

src/Search/SearchServiceProvider.php
```diff
@@ -40,41 +39,47 @@ public function register()
...
     public function registerUserGambits()
     {
-        $this->app->when(UserSearcher::class)
-            ->needs(GambitManager::class)
-            ->give(function (Container $app) {
-                $gambits = new GambitManager($app);
+        $gambits = AbstractSearcher::gambitManager(UserSearcher::class);
+        $gambits->setFullTextGambit($this->app->make(UserFulltextGambit::class));
+        $gambits->add($this->app->make(EmailGambit::class));
+        $gambits->add($this->app->make(GroupGambit::class));
```
SychO9, line 45: we'd be making the gambits in the registration phase, don't we usually avoid resolving classes too early ?

also why not inject the GambitManager like before ?

askvortsov1: Good point, this should be moved to boot. I'm pretty sure this means that the event BC layer isn't working either.

askvortsov1: I wanted to avoid contextual binding (when needs give), because if extensions want to add their own searchers, they'll also need to include a full service provider.

Although actually now that I think about it, every gambit registered via the `SimpleFlarumSearch` extender is associated with a Searcher class, so we could dynamically construct all the contextual bindings. That's actually quite a bit nicer, and cuts down a little bit on the evil static stuff.

src/Search/SearchServiceProvider.php
```diff
@@ -18,63 +18,91 @@
...
-            ->needs(GambitManager::class)
-            ->give(function (Container $app) {
-                $gambits = new GambitManager($app);
+                    if ($searcher === UserSearcher::class) {
+                        $this->app->make('events')->dispatch(
+                            new ConfigureUserGambits($gambitManager)
+                        );
+                    } elseif ($searcher === DiscussionSearcher::class) {
+                        $this->app->make('events')->dispatch(
+                            new ConfigureDiscussionGambits($gambitManager)
```
SychO9, line 89: The gambits added by old extensions will be classes and not instances, so they won't work. (same with full text gambits)

Let's do the same here with what you did before, add them to a temporary gambit manager, get all gambits and re-add them to the original gambit manager after making them.

SychO9 (changes requested):

src/Search/SearchServiceProvider.php
askvortsov1: Ah, that's why I did that! I was confused in retrospect.

SychO9 (approved):
LGTM!

tankerkiller125 (approved):
I haven't tested this code, but I've been following along with it and I see no glaring issues and in my opinion it's just as easy if not easier to read than the original code.

askvortsov1:
Awesome, thanks guys! We'll merge this after the 2 dependency search refactor PRs

askvortsov1:
Actually looks like this doesn't depend on the search/filter split (or at least I don't believe so), so I'll try merging it tomorrow. That should make review for the search/filter split PR a bit easier.

PR #2485, opened 2020-12-04
Pass callback wrapper parameters by reference

SychO9:
**Fixes https://github.com/flarum/core/pull/2451#pullrequestreview-545124663**

Because invokable class objects are not directly called and instead it's the callback wrapper that calls these objects, it's currently not possible to receive arguments by reference on an invokable class.

To fix this we pass the arguments by reference by default when calling the object in the callback wrapper.

### DIFF
--- a/src/Foundation/ContainerUtil.php
+++ b/src/Foundation/ContainerUtil.php
@@ -24,10 +24,10 @@ class ContainerUtil
     public static function wrapCallback($callback, Container $container)
     {
         if (is_string($callback)) {
-            $callback = function () use ($container, $callback) {
+            $callback = function (&...$args) use ($container, $callback) {
                 $callback = $container->make($callback);
 
-                return call_user_func_array($callback, func_get_args());
+                return $callback(...$args);
             };
         }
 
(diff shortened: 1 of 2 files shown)


askvortsov1 (changes requested):
I think we should add a unit (not integration) test class for ContainerUtil that checks that it works as expected with pass-by-value, pass-by-reference, and a mixture of such.

SychO9:
hmm, the `wrapCallback` method does require the container, I guess creating a custom container class with a dummy implementation for the `make` method should be fine ?

askvortsov1:
Hmm, could we use `mock` to create a container instance?

SychO9:
I'm not familiar with Mockery, as far as I can tell that won't work, `make` always ends up returning null. I do not as of yet know how Mockery works to know if that's even possible.

askvortsov1 (approved):

askvortsov1:
Because invokable class objects are not directly called and instead it's the callback wrapper that calls these objects, it's currently not possible to receive arguments by reference on an invokable class.

To fix this we pass the arguments by reference by default when calling the object in the callback wrapper.

Issue #2085, opened 2020-03-25
Bring back common, forum, and admin namespaces in JS


datitisev:
Related to #1821. 

I think it'd be in our best interest to bring back the `common`, `forum` and `admin` namespaces that were removed at some point (I think beta 8).

Pros:
- we don't have to worry about naming files the same in `common` and `forum`/`admin` causing issues (because of `flarum.core.compat`)
	- also prevents issues when we generate typings - using `flarum/forum/Forum` instead of `flarum/Forum`
	  ![img](https://cdn.discordapp.com/attachments/600769671922319360/691003021060210758/unknown.png)
- makes knowing where files are located in core easy when looking at extensions (no need to guess namespace)
- if looking at core, can easily see what the import should be (no need to remove part of it)
- easier to understand from the perspective of a newly starting extension developer (luceos, https://github.com/flarum/core/issues/2085#issuecomment-604285670)

Cons:
- definitely means ALL JS exts will stop working
- @clarkwinkelmann thinks it'd be more readable, likes bundling <stack> & common

What we could do is allow both ways- keep the current structure without the stack in flarum.core.compat BUT also allow including the stack when typing code (stripped on build). This could cause confusion, however.

If I missed any pro/con, please comment so.

luceos:
It's also easier to understand from the perspective of a newly starting extension developer.

tankerkiller125:
I feel like the majority of JS extensions will probably break in the rewrite anyways. I think the benefits far outweigh the cons here. Just making it easier to understand for new extension developers I think alone outweighs the cons.

clarkwinkelmann:
We currently accept using the `/admin` and `/forum` imports when using `@` imports https://github.com/flarum/flarum-webpack-config/blob/master/index.js#L55

We can add that as well as `/common` support to the "old-style" section so that developers can prepare their extensions for the upcoming change.

datitisev:
@clarkwinkelmann I don't think there's a single extension that uses `@flarum/core` imports - mainly because core exports barely anything and it's not documented in the slightest.

I didn't even know we had JS extenders 🤷‍♂

askvortsov1:
I'm generally in the 'explicit is better than implicit' camp, and I think this makes things clearer. I'm in favor.

luceos:
@franzliedke could you share your thoughts on this one too 🙏

datitisev:
Still relevant. We're probably doing this.

clarkwinkelmann:
[...]

PR #2488, opened 2020-12-06
Replace forum and admin global compat exports with a Proxy

datitisev:
Refs #2085.

Replaces https://github.com/flarum/flarum-webpack-config/pull/7.

Adds ability for import statements of core & core extensions to include their namespaces (`forum`, `admin`, and `common`).

We'll eventually require extensions use the namespaces in their imports. This PR makes them optional for extensions and allows extensions to continue working once we require them.

Original compat objects are left intact & do not contain the namespaces.

- import `admin/utils/extract` --> returns import `utils/extract`
- import `tags/common/utils/sortTags` --> returns import `tags/utils/sortTags`

Technically extensions can also use the incorrect namespace (forum/admin and common) as this code doesn't check whether it's correct. This shouldn't be a problem that we focus on, however.

### DIFF
--- a/js/src/admin/index.js
+++ b/js/src/admin/index.js
@@ -10,8 +10,9 @@ export { app };
 // Export public API
 
 // Export compat API
-import compat from './compat';
+import compatObj from './compat';
+import proxifyCompat from '../common/utils/proxifyCompat';
 
-compat.app = app;
+compatObj.app = app;
 
-export { compat };
+export const compat = proxifyCompat(compatObj, 'admin');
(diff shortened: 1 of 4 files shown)


askvortsov1 (approved):
Seems to work when tested locally, makes sense to me! Do we want to add a deprecation warning for the old system? Or let these exist in parallel before deprecating the old one? Maybe next release we should sweep through and deprecate all the old systems (including the @flarum/namespace/path one)?

datitisev:
> Do we want to add a deprecation warning for the old system?

What would this deprecation warning look like? We'd only want to log one message because if we do every deprecated import, the console would be flooded with these messages.

> Or let these exist in parallel before deprecating the old one?

Yeah I think we should deprecate these in the next release.

> Maybe next release we should sweep through and deprecate all the old systems (including the @flarum/namespace/path one)?

Yes, though I think the `@flarum/core` is for public API so we wouldn't want to do that? Still not entirely sure.

SychO9 (approved):

tankerkiller125 (approved):

PR #2493, opened 2020-12-10
Refactor AccountActivationMailer and SendConfirmationEmailController

### LINKED ISSUE #2193: Confirmation email code duplication
Hello,

I've noticed that the [SendConfirmationEmailController](https://github.com/flarum/core/blob/4da2994d1fab1a167406680876101d19603cc27a/src/Api/Controller/SendConfirmationEmailController.php) and [AccountActivationMailer](https://github.com/flarum/core/blob/4da2994d1fab1a167406680876101d19603cc27a/src/User/AccountActivationMailer.php) classes do almost exactly the same thing, with the only difference that they are triggered in different situations.

I think that there's space for improvement and that the code duplication could be avoided.

sl-kr:
**Fixes #2193**

Moves the common code between [SendConfirmationEmailController](https://github.com/flarum/core/blob/master/src/Api/Controller/SendConfirmationEmailController.php) and [AccountActivationMailer](https://github.com/flarum/core/blob/master/src/User/AccountActivationMailer.php) into a trait.

I wasn't sure what to name the trait or where to put it, but because most of the code for the trait was taken from the AccountActivationMailer class I decided to name it the AccountActivationMailerTrait and keep it in the same directory. Also, I decoupled the generateToken and getEmailData functions because I think it will make more modular and maintainable even through it may make using the trait more verbose.

**Confirmed**

### DIFF
--- a/src/User/AccountActivationMailerTrait.php
+++ b/src/User/AccountActivationMailerTrait.php
@@ -47,7 +47,8 @@ trait AccountActivationMailerTrait
      * @param User $user
      * @param array $data
      */
-    protected function sendConfirmationEmail(User $user, $data) {
+    protected function sendConfirmationEmail(User $user, $data)
+    {
         $body = $this->translator->trans('core.email.activate_account.body', $data);
         $subject = '['.$data['{forum}'].'] '.$this->translator->trans('core.email.activate_account.subject');
 
--- a/src/User/AccountActivationMailerTrait.php
+++ b/src/User/AccountActivationMailerTrait.php
@@ -30,11 +30,10 @@ trait AccountActivationMailerTrait
      * Get the data that should be made available to email templates.
      *
      * @param User $user
-     * @param string $email
      * @param EmailToken $token
      * @return array
      */
-    protected function getEmailData(User $user, $email, EmailToken $token)
+    protected function getEmailData(User $user, EmailToken $token)
     {
         return [
             '{username}' => $user->display_name,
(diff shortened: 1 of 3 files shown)


src/User/AccountActivationMailerTrait.php
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * Get the data that should be made available to email templates.
+     *
+     * @param User $user
+     * @param string $email
+     * @param EmailToken $token
+     * @return array
+     */
+    protected function getEmailData(User $user, $email, EmailToken $token)
```
askvortsov1, line 37: Doesn't look like the email is actually used here

src/User/AccountActivationMailerTrait.php
```diff
@@ -0,0 +1,57 @@
...
+use Flarum\Mail\Job\SendRawEmailJob;
+
+trait AccountActivationMailerTrait
+{
+    /**
+     * @param User $user
+     * @param string $email
+     * @return EmailToken
+     */
+    protected function generateToken(User $user, $email)
```
askvortsov1, line 21: I'm leaning towards keeping it like this, with the email passed in separately to allow future flexibility if we decide to adopt multiple emails per user.

askvortsov1 (changes requested):

askvortsov1 (approved):
Looks good to me!

askvortsov1:
A side effect of this would implement https://github.com/flarum/core/issues/2515, which still needs a bit more discussion before merging this.

SychO9 (approved):

PR #2494, opened 2020-12-11
Minor docblock error

imorland:
Whilst updating an extension to use this new extender, I noticed that the docblock had incorrect namespaces, so here you go.... ;)


askvortsov1:
Thanks! Merged this directly since it's very simple and not part of the code.

PR #2495, opened 2020-12-13
Add default value to Settings extender

SychO9:
**Related to #2452**

So it seems as though adding default values through migrations isn't always the way to go, at least for serializing.

**Confirmed**

### DIFF
--- a/src/Extend/Settings.php
+++ b/src/Extend/Settings.php
@@ -26,11 +26,12 @@ class Settings implements ExtenderInterface
      * @param string $attributeName: The attribute name to be used in the ForumSerializer attributes array.
      * @param string $key: The key of the setting.
      * @param string|callable|null $callback: Optional callback to modify the value before serialization.
+     * @param mixed $default: Optional default serialized value.
      * @return $this
      */
-    public function serializeToForum(string $attributeName, string $key, $callback = null)
+    public function serializeToForum(string $attributeName, string $key, $callback = null, $default = null)
     {
-        $this->settings[$key] = compact('attributeName', 'callback');
+        $this->settings[$key] = compact('attributeName', 'callback', 'default');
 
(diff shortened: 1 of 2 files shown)


src/Extend/Settings.php
```diff
@@ -26,11 +26,12 @@ class Settings implements ExtenderInterface
      * @param string $attributeName: The attribute name to be used in the ForumSerializer attributes array.
      * @param string $key: The key of the setting.
      * @param string|callable|null $callback: Optional callback to modify the value before serialization.
+     * @param mixed $default: Optional default serialized value.
```
askvortsov1, line 29: I think it'd be important to mention (and test) that the default value would still be passed through the callback.

askvortsov1 (changes requested):

src/Extend/Settings.php
SychO9: sure,
the test I added actually already makes sure it passes through the callback, but I guess I'll add another test where there's simply no callback as well.

src/Extend/Settings.php
```diff
@@ -26,7 +26,7 @@ class Settings implements ExtenderInterface
      * @param string $attributeName: The attribute name to be used in the ForumSerializer attributes array.
      * @param string $key: The key of the setting.
      * @param string|callable|null $callback: Optional callback to modify the value before serialization.
-     * @param mixed $default: Optional default serialized value.
+     * @param mixed $default: Optional default serialized value. Is also passed to the optional callback.
```
askvortsov1, line 29: `Will be run through the optional callback`

askvortsov1 (changes requested):

askvortsov1 (approved):

tankerkiller125 (approved):

PR #2497, opened 2020-12-15
Fixes validation failures of avatars that are jpg/jpeg

luceos:
Summary: https://discuss.flarum.org/d/25536-avatar-upload-jpg/47

Due to a commit by @fabpot in october, the mimetypes symfony class
now re-orders the shortened mimetypes that are returned when looking
up based on header mimetype. Our validator uses the first key, pops
the prefix off and then matches against our hardcoded array.

I've added a constraint to symfony/mime ^5.2.0 which ships with this change.
This constraint is fully compatible with our current lineup. In addition
I changed the hardcoded array to use the first entry from symfony mime types
now `jpg` instead of `jpeg`.

This change has been locally tested and confirmed fixed.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -79,6 +79,7 @@
         "symfony/config": "^4.3.4",
         "symfony/console": "^4.3.4",
         "symfony/event-dispatcher": "^4.3.4",
+        "symfony/mime": "^5.2.0",
         "symfony/translation": "^4.3.4",
         "symfony/yaml": "^4.3.4",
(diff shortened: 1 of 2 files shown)


clarkwinkelmann (approved):
Not tested, but looking good :+1:

askvortsov1:
Should we remove `jpeg` entirely? I'm not that familiar with how this system works, but why not have both?

tankerkiller125 (approved):

askvortsov1 (approved):

PR #2507, opened 2020-12-22
PHP 8 support

### LINKED ISSUE #2439: Consider switching cookie library
Our current cookie library is fairly unmaintained and likely won't get PHP8.0 support listed in the composer.json file. This leaves us basically two options to fix the issue:

Option 1: Change the library, https://github.com/delight-im/PHP-Cookie looks pretty good just at a quick glance, this will require swapping out all internals and either making a BC layer or just out right breaking any extensions that set their own cookies.

Options 2: Fork the existing library, this has the benefit of no BC layer or breaking extension, but would require us to maintain the fork.

luceos:
fixes #2439

Attempts to offer support for php 8, so far:

- [x] dropped 7.2 due to [laminas](https://github.com/flarum/core/runs/1594451299?check_suite_focus=true) let's try to add both versions.
- [x] dropped franz/whoops-middleware, needs a new version [compatible with the diactoros lib](https://github.com/flarum/core/runs/1594463301?check_suite_focus=true)
- [x] replaced fig cookies with hansott/psr7-cookies


franzliedke:
`franzl/whoops-middleware` now supports everything from PHP 7.2 through PHP 8.0 with [the new 2.0 release](https://github.com/franzliedke/whoops-middleware/releases/tag/2.0.0). Please use that. :smiley:

tankerkiller125:
A note on 7.2 support. Due to changes in PHPUnit and many of our libraries requiring said updated PHPUnit (without alternative versions) it will be basically impossible to support PHP 7.2

In theory I think we could maybe force it to work by using the `--no-dev` option in composer however I think this might add further confusion and more support issues in the long run if people forget to use it. Further 7.2 support is officially dead from the PHP team.

luceos:
@franzliedke thank you, I was hoping for that 🤗

composer.json
```diff
@@ -61,14 +61,14 @@
         "illuminate/validation": "^6.0",
         "illuminate/view": "^6.0",
         "intervention/image": "^2.5.0",
-        "laminas/laminas-diactoros": "^1.8.4",
-        "laminas/laminas-httphandlerrunner": "^1.0",
-        "laminas/laminas-stratigility": "^3.0",
+        "laminas/laminas-diactoros": "^2.4.1 || ^2.5.0",
```
askvortsov1, line 64: Why can't we just use the former of the conditional requirements? Wouldn't the effect be the same, since caret goes all the way up to the next major version, and all of these are post-1.0 versions?

tankerkiller125 (approved):
Looks like the main showstopper was the cookie library originally, this looks good I see no issues with the cookie lib replacement and CI test are passing.

src/Http/CookieFactory.php
```diff
@@ -72,45 +71,34 @@ public function __construct(Config $config)
      * This method returns a cookie instance for use with the Set-Cookie HTTP header.
      * It will be pre-configured according to Flarum's base URL and protocol.
      *
-     * @param  string  $name
-     * @param  string  $value
-     * @param  int     $maxAge
-     * @return \Dflydev\FigCookies\SetCookie
+     * @param  string       $name
```
askvortsov1, line 74: Why not just use the question mark before the string / int?

src/Http/CookieFactory.php
```diff
@@ -72,45 +71,34 @@ public function __construct(Config $config)
...
+     * @param  string|null  $value
+     * @param  int|null     $maxAge
+     * @return SetCookie
      */
-    public function make($name, $value = null, $maxAge = null)
+    public function make(string $name, ?string $value, ?int $maxAge): SetCookie
```
askvortsov1, line 79: Any particular reasoning for the change from null default to nullable?

src/Http/CookieFactory.php
```diff
@@ -72,45 +71,34 @@ public function __construct(Config $config)
...
      *
      * @param string $name
-     * @return \Dflydev\FigCookies\SetCookie
+     * @return SetCookie
      */
-    public function expire($name)
+    public function expire(string $name): SetCookie
     {
-        return $this->make($name)->expire();
+        return SetCookie::thatExpires($name);
```
askvortsov1, line 101: Looks like we're not providing a value, but that doesn't appear to be a nullable field (https://github.com/hansott/psr7-cookies/blob/master/src/SetCookie.php#L36)? Do we have any automated tests that call this code?

src/Http/CookieFactory.php
askvortsov1: And in the libraries source, it doesn't seem like $value is nullable https://github.com/hansott/psr7-cookies/blob/master/src/SetCookie.php#L36

src/Http/Rememberer.php
```diff
@@ -49,9 +47,7 @@ public function rememberUser(ResponseInterface $response, $userId)
 
     public function forget(ResponseInterface $response)
     {
-        return FigResponseCookies::set(
-            $response,
-            $this->cookie->expire(self::COOKIE_NAME)
-        );
+        return $this->cookie->expire(self::COOKIE_NAME)
```
askvortsov1, line 50: As per the above concerns with nullable value, I'd like to see a unit test for this method

askvortsov1 (changes requested):

src/Http/CookieFactory.php
luceos: phpdoc doesn't allow for conditional type hinting, it is always rendered as constants

composer.json
luceos: Because only those two versions are currently compatible. The old constraint `^1.8.4` would only allow upgrade up to `1.9` and blocks `1.9.0` and up.

The conditional requirement is required because 2.4.1 and up is compatible with php 7.3+; and only 2.5.0 and up is compatible with php 8.

src/Http/CookieFactory.php
luceos: Our code allowed it, I didn't want to introduce even more changes than necessary. The nullable type hint is cleaner imo.

src/Http/CookieFactory.php
luceos: I don't know, I refactored existing code to be exactly as it was before.

composer.json
SychO9: I think what he means is, isn't `^2.4.1` no different than `^2.4.1 || ^2.5.0` ?
https://jubianchi.github.io/semver-check/#/^2.4.1/2.5.0

src/Http/CookieFactory.php
askvortsov1: > It also doesn't make sense for a cookie to not have a value.

But this cookie right here doesn't have a value? Essentially I just want to make sure that doing it this way won't crash the site, and it seems like the new cookie library requires a non-nullable `$value` param

composer.json
luceos: 😱 I mixed up tilde against caret, sorry folks. Change makes sense!
```suggestion
        "laminas/laminas-diactoros": "^2.4.1",
```

src/Http/CookieFactory.php
luceos: Sure, I'll check it out.

tests/unit/Http/CookieFactoryTest.php
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    /** @test */
+    public function can_create_cookies()
+    {
+        $cookie = $this->factory()->make('test', 'australia');
+
+        $this->assertEquals('flarum_test', $cookie->getName());
+        $this->assertEquals('australia', $cookie->getValue());
+        $this->assertEquals(0, $cookie->expiresAt());
```
askvortsov1, line 34: Isn't `expiresAt` an absolute time value, not a relative lifetime? And if so, wouldn't this cause the cookie to expire immediately?

Do we need to do any formatting into the [Date format](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Date), or is that handled by the cookie library?

askvortsov1 (changes requested):
Looks like our current library was updated: https://packagist.org/packages/dflydev/fig-cookies

In the spirit of minimal changes, I'd say we should stick with that, although keep the new tests (and include a test case for calling ->expires())

Sorry for all the back and forth on this one

tests/unit/Http/CookieFactoryTest.php
luceos: It is handled in the lib.

composer.json
```diff
@@ -61,14 +61,14 @@
         "illuminate/validation": "^6.0",
         "illuminate/view": "^6.0",
         "intervention/image": "^2.5.0",
-        "laminas/laminas-diactoros": "^1.8.4",
-        "laminas/laminas-httphandlerrunner": "^1.0",
-        "laminas/laminas-stratigility": "^3.0",
+        "laminas/laminas-diactoros": "^2.4.1",
+        "laminas/laminas-httphandlerrunner": "^1.2.0 || ^1.3.0",
```
askvortsov1, line 65: Didn't we change these sometime previously in the PR's history?

src/Http/CookieFactory.php
```diff
@@ -77,7 +77,7 @@ public function __construct(Config $config)
      * @param  int     $maxAge
      * @return \Dflydev\FigCookies\SetCookie
      */
-    public function make($name, $value = null, $maxAge = null)
+    public function make(string $name, string $value, int $maxAge = null): SetCookie
     {
         $cookie = SetCookie::create($this->getName($name), $value);
```
askvortsov1, line 82: A value is now required here (necessary for the old library). Is that purposeful, or should we make this null default?

tests/integration/BuildsHttpRequests.php
```diff
@@ -34,21 +33,10 @@ protected function requestWithJsonBody(Request $req, array $json): Request
 
     protected function requestAsUser(Request $req, int $userId): Request
     {
-        $token = Str::random(40);
+        $token = AccessToken::generate($userId);
```
askvortsov1: What's the reasoning behind this change? Franz purposefully wanted to avoid this in https://github.com/flarum/core/commit/31765388c1eec0aba090514633111c2a946ac257 (although why I'm not sure)

tests/unit/Foundation/ContainerUtilTest.php
```diff
@@ -17,10 +17,7 @@ class ContainerUtilTest extends TestCase
 {
     private $container;
 
-    /**
-     * @inheritDoc
-     */
-    protected function setUp()
+    protected function setUp(): void
```
askvortsov1, line 20: Why?

tests/unit/Http/CookieFactoryTest.php
```diff
@@ -0,0 +1,56 @@
...
+
+        $this->assertEquals('flarum_test', $cookie->getName());
+        $this->assertEquals('australia', $cookie->getValue());
+        $this->assertEquals(0, $cookie->getExpires());
+        $this->assertFalse($cookie->getSecure());
+        $this->assertEquals('/', $cookie->getPath());
+    }
+
+    /** @test */
+    public function can_override_cookie_settings_from_config()
```
askvortsov1, line 40: I'd like to see one more test case that generates an expired cookie

askvortsov1 (changes requested):

src/Http/CookieFactory.php
luceos: I made it default, because as you said a cookie without value makes no sense. Also the underlying library will simply cast value to string (and then urlencode it)...

tests/unit/Foundation/ContainerUtilTest.php
luceos: I think my test complained, I have no (logged) other reason for this.

composer.json
luceos: Right.

tests/unit/Foundation/ContainerUtilTest.php
luceos: Yeah:

`Return type declaration must be compatible with TestCase->setUp() : void`

src/Http/CookieFactory.php
luceos: I went back with nullable because it made the code cleaner.

tests/unit/Foundation/ContainerUtilTest.php
askvortsov1: My concern here was about removing the docblock, sorry for the confusion

askvortsov1 (changes requested):

askvortsov1 (approved):

PR #2517, opened 2020-12-29
Add forum prefix in re-sent confirmation email

### LINKED ISSUE #2515: Activation email doesn't have forum title prefix when re-sent
## Bug Report

**Current Behavior**
The activation email has the [Forum Name] prefix in the subject. But clicking "send activation email again" sends an identical email, without the prefix.

**Steps to Reproduce**
1. Create new account
2. Ask for new activation email
3. Compare the two emails received

**Expected Behavior**
The two emails should have identical titles. The missing prefix seems to be an oversight.

**Screenshots**
image

**Environment**
- Flarum version: beta 15

**Possible Solution**
Add subject prefix by copy-pasting code

Without prefix

https://github.com/flarum/core/blob/f8edc2d827fee47dc78fbb75ee1d3b28580ed11d/src/Api/Controller/SendConfirmationEmailController.php#L85

With prefix

https://github.com/flarum/core/blob/0a6c5217c11e625aab5deec05bf6a6caf31bfa0d/src/User/AccountActivationMailer.php#L66

Ammarpad:
**Fixes #2515**

In SendConfirmationEmailController::handle(), append forum title to the email subject. The title is provided by the $data array some few lines above.

When a confirmation email is resent, the subject line should be similar to the subject line of activation email. That's should have "[Forum Name]" prefix, followed by the remaining subject text

**Screenshot**

**Confirmed**

### DIFF
--- a/src/Api/Controller/SendConfirmationEmailController.php
+++ b/src/Api/Controller/SendConfirmationEmailController.php
@@ -82,7 +82,7 @@ class SendConfirmationEmailController implements RequestHandlerInterface
         ];
 
         $body = $this->translator->trans('core.email.activate_account.body', $data);
-        $subject = $this->translator->trans('core.email.activate_account.subject');
+        $subject = '['.$data['{forum}'].'] '.$this->translator->trans('core.email.activate_account.subject');
 
         $this->queue->push(new SendRawEmailJob($actor->email, $subject, $body));
 


askvortsov1:
Thank you very much for this PR! Unfortunately it looks like this issue needs a bit more discussion in #2515, and that https://github.com/flarum/core/pull/2493 seems like a more general-case solution. For this reasons, I'm going to close this PR for now. I apologize for the miscommunication on our part. Thanks again!

Issue #2506, opened 2020-12-21
Cannot delete a post or discussion from a deleted user


clarkwinkelmann:
## Bug Report

**Current Behavior**
If a post has no user, an error is likely to happen in `UserMetadataUpdater` when deleting the content.

Errors will look like this:

```
Symfony\Component\Debug\Exception\FatalThrowableError: Argument 1 passed to Flarum\User\UserMetadataUpdater::updateCommentsCount() must be an instance of Flarum\User\User, null given, called in /home/clark/Projects/flarum-beta15/vendor/flarum/core/src/User/UserMetadataUpdater.php on line 45 in /home/clark/Projects/flarum-beta15/vendor/flarum/core/src/User/UserMetadataUpdater.php:68
Stack trace:
#0 /home/clark/Projects/flarum-beta15/vendor/flarum/core/src/User/UserMetadataUpdater.php(45): Flarum\User\UserMetadataUpdater->updateCommentsCount(NULL)
```

**Steps to Reproduce**
1. Create post
2. Delete user, or change author to Guest with "Author Change" extension
3. Soft delete post, then permanently delete post

Also applies to discussions

1. Create discussions
2. Delete user, or change author to Guest with "Author Change" extension
3. Soft delete discussions, then permanently delete discussions

**Expected Behavior**
No error.

**Environment**
- Flarum version: beta 15

**Possible Solution**

This function needs to accept `$user = null`

https://github.com/flarum/core/blob/f8edc2d827fee47dc78fbb75ee1d3b28580ed11d/src/User/UserMetadataUpdater.php#L68

**Additional Context**

Issue introduced in #2472

I'm encountering this particular issue in a closed source extension that creates new posts without any user, which also triggers the same error when `Post\Created` is dispatched.

askvortsov1:
I would prefer to accept a nullable user here instead of a default null user. https://www.php.net/manual/en/migration71.new-features.php

So that would look like:
`private function updateCommentsCount(?User $user)`

PR #2521, opened 2021-01-04
Fix deleting posts/discussions by deleted user

askvortsov1:
Fixes https://github.com/flarum/core/issues/2506

Making the $user argument nullable prevents this unnecessary exception, and doesn't introduce any issues since we check that $user exists as part of the method.


SychO9 (approved):

tankerkiller125 (approved):
Looks good

Issue #1959, opened 2019-12-28
Scroll to top and menu disappears


zerosonesfun:
## Bug Report

**Current Behavior**
On my mobile device when scrolling to the top quickly, sometimes the menu disappears. This happens at the Flarum discuss forum and a forum I have installed as well. Here is a screenshot. The arrow is pointing to a sliver of the new post icon that you can still see. This is to show that the menu text/icons scroll out of view or perhaps don’t get painted back in the right spot in time. 

<img src="https://i.ibb.co/34GrdrS/AF2-D8827-C5-BE-4-FA7-B480-41007-D30-D0-BE.jpg" alt="AF2-D8827-C5-BE-4-FA7-B480-41007-D30-D0-BE" border="0">

**Steps to Reproduce**
1. Ideally use an iPhone and the Chrome iOS browser
2. Scroll down
3. Quickly swipe up to the top
4. Repeat several times until the menu disappears 

**Expected Behavior**
Menu text/icons should not suddenly disappear even if it is rare.

**Environment**
- Flarum version: Beta 11.1
- Website URL: https://discuss.flarum.org
- Webserver: ?
- Hosting environment: ?
- PHP version: ?
- Browser: For sure iOS Chrome. Could be other mobile browsers. 

**Possible Solution**
This fixed it at my forum and I have not seen any new issues created by doing this. I cancelled out this CSS rule:

`transform: translate3d(0,0,0)`

Which I believe is located here: https://github.com/flarum/core/blob/master/less/common/mixins/header-background.less

Why does this fix it? I don’t know for sure but I’ve read that many people add this transform rule to fix paint issues (issues similar to this one where elements are disappearing). However, instead of being the fix, it seems this rule is causing the issue. 

Is this one rule necessary especially if it is periodically making the main nav bar disappear on certain devices?

franzliedke:
Thanks for the report! :raised_hands: 

Has anybody else seen this problem, and - most importantly - on other OS / browser combinations?

@zerosonesfun Could you provide a link to where you read about the problems with this transform rule? Thanks for your research!

Ralkage:
@franzliedke I tried to reproduce this in the latest versions of Chrome and Firefox for iOS and Android as well as Safari on iOS and Samsung's browser for Android. Unfortunately, I was unable to reproduce this issue even with the steps to reproduce provided in this issue.

zerosonesfun:
[...]

PR #2527, opened 2021-01-05
Fix IOS scroll menu bug

askvortsov1:
Fixes https://github.com/flarum/core/issues/1959

These transform lines are known to cause issues on iOS, and were added to hack around chrome issues that have since been fixed upstream.

### DIFF
--- a/less/common/mixins/header-background.less
+++ b/less/common/mixins/header-background.less
@@ -6,7 +6,6 @@
   right: 0;
   z-index: @zindex-header;
   border-bottom: 1px solid @control-bg;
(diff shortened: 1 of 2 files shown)


SychO9 (approved):
Having multiple iOS users confirm this fixes the issues mentioned without any apparent drawbacks or signs of the bug it initially was introduced to fix, and having done some research, I approve!

datitisev (commented):
Not tested locally but we don't want this code here :+1:

datitisev (approved):
I forgot to click approve 🙈

PR #2529, opened 2021-01-07
Update UserValidator.php

dev2lead:
Allow use of dot in username (mostly used in ldap directories when used as authentication method)

**Fixes #0000**

UserValidation.php file will be modified to allow usage of dot in username field

**Screenshot**

**Confirmed**

**Required changes:**

### DIFF
--- a/src/User/UserValidator.php
+++ b/src/User/UserValidator.php
@@ -44,7 +44,7 @@ class UserValidator extends AbstractValidator
         return [
             'username' => [
                 'required',
-                'regex:/^[a-z0-9_-]+$/i',
+                'regex:/^[a-z0-9._-]+$/i',
                 'unique:users,username'.$idSuffix,
                 'min:3',
                 'max:30'


askvortsov1 (changes requested):
Thank you for the PR! This is something I wouldn't mind seeing in theory, but I have a few concerns:

- IIRC, the main reason why we currently don't allow periods in usernames is that it makes mention regex more complicated. If we want to support it, this needs to be accompanied by PRs to mentions to fix that regex. I would also be in favor of only allowing dots as a middle character (username can't start or end with a period)
- Another potential issue is dots in urls past the domain, don't recall if this is allowed
- I would like to see unit/integration tests for this validator. I don't think we have them yet, but why not start now?
- Before proceeding with changes we should confirm with @flarum/core whether this is something we even want (I assume the initial lack of support was purposeful)

Thanks again!

dev2lead:
Thank you for the quick reply, I've modified to regex on the mentions plugin too and it works great (see PR https://github.com/flarum/mentions/pull/60)

As mention in rfc 3986 (cf https://tools.ietf.org/html/rfc3986#section-2.3), dot (.) is not a reserved character : unreserved  = ALPHA / DIGIT / "-" / "." / "_" / "~"

clarkwinkelmann:
Is there any forum/chat software out there that allows dots in usernames?

I'm personally not a big can of allowing more characters than necessary and if nobody else does it, I see no incentive to support it in Flarum.

The problems related to importing/matching usernames from an external system can be handled by other means. Sure, it's nice to have the same usernames across the platforms, but it's far from the only solution.

If we start allowing this, then we might also get requests for odd characters allowed in other (legacy or not) systems, and where do we stop? As said, any new character adds risks of breaking/complicating extensions like Mentions.

Should we move this discussion to an issue?

dev2lead:
Yes, several forum/chat softwares allow dots in username. For example Discourse, Discord or even Synology Chat.

I think that this is really a basic feature (compared to allowing utf8 username for example) that could allow compatibility with many external services (especially authentication services).

askvortsov1:
> If we start allowing this, then we might also get requests for odd characters allowed in other (legacy or not) systems, and where do we stop? As said, any new character adds risks of breaking/complicating extensions like Mentions.

A period is part of ASCII, and is a fairly standard divider (in email addresses, in addition to the ones listed above). As to further requests, adding a period does not obligate us to add anything else.

With mentions, I am still concerned about periods being the first or last character, because that messes with punctuation. For instance, in "Thanks for the help @clarkwinkelmann.", the period after "@clarkwinkelmann" would be considered part of the username by the proposed regex

dev2lead:
> With mentions, I am still concerned about periods being the first or last character, because that messes with punctuation. For instance, in "Thanks for the help @clarkwinkelmann.", the period after "@clarkwinkelmann" would be considered part of the username by the proposed regex

This could be quite easily handled with a regex, you could allow usage of "." "_" "-" but force to begin and end username with an alphanum character with this regex : /^[a-z0-9]+[a-z0-9._-]+[a-z0-9]+$/i

tankerkiller125:
I think I'd be interested to hear exactly why this needs to be in core instead of an extension? This kind of change could be quite easily made via extension except for maybe the mentions part?

While it is a small change I'm not 100% sold on it's use case or reason. As for the use case of making it easier to use other authentication methods 3rd party OAuth providers use a "recommendation" system for usernames.

PR #2532, opened 2021-01-09
Convert highlight helper to Typescript

daniellesniak:
See flarum/framework#3533

### DIFF
--- a/js/src/common/helpers/highlight.ts
+++ b/js/src/common/helpers/highlight.ts
@@ -3,12 +3,6 @@ import { truncate } from '../utils/string';
 /**
  * The `highlight` helper searches for a word phrase in a string, and wraps
  * matches with the <mark> tag.
- *
- * @param {string} string The string to highlight.
- * @param {string|RegExp} phrase The word or words to highlight.
- * @param {number} [length] The number of characters to truncate the string to.
- *     The string will be truncated surrounding the first match.
- * @return {Object}
  */
 export default function highlight(string: string, phrase: string | RegExp, length: number): Object | string {
   if (!phrase && !length) return string;
--- a/js/src/common/helpers/highlight.ts
+++ b/js/src/common/helpers/highlight.ts
@@ -4,21 +4,21 @@ import { truncate } from '../utils/string';
  * The `highlight` helper searches for a word phrase in a string, and wraps
  * matches with the <mark> tag.
  *
- * @param {String} string The string to highlight.
- * @param {String|RegExp} phrase The word or words to highlight.
- * @param {Integer} [length] The number of characters to truncate the string to.
+ * @param {string} string The string to highlight.
+ * @param {string|RegExp} phrase The word or words to highlight.
+ * @param {number} [length] The number of characters to truncate the string to.
  *     The string will be truncated surrounding the first match.
  * @return {Object}
  */
-export default function highlight(string, phrase, length) {
+export default function highlight(string: string, phrase: string | RegExp, length: number): Object | string {
   if (!phrase && !length) return string;
 
   // Convert the word phrase into a global regular expression (if it isn't
   // already) so we can search the string for matched.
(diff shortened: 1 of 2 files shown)


dsevillamartin:
Should the types be removed from the jsdoc comment? 🤔

I feel like there's no reason to have them there twice. The JSDoc is still helpful for the description of the arguments.

daniellesniak:
You're right, I fixed that.

js/src/common/helpers/highlight.ts
```diff
@@ -4,21 +4,20 @@ import { truncate } from '../utils/string';
...
- * @param {String|RegExp} phrase The word or words to highlight.
- * @param {Integer} [length] The number of characters to truncate the string to.
+ * @param string The string to highlight.
+ * @param phrase The word or words to highlight.
+ * @param [length] The number of characters to truncate the string to.
  *     The string will be truncated surrounding the first match.
- * @return {Object}
  */
-export default function highlight(string, phrase, length) {
+export default function highlight(string: string, phrase: string | RegExp, length: number): Object | string {
```
w-4, line 12: The return type should be set to `Mithril.Vnode` here

js/src/common/helpers/highlight.ts
```diff
@@ -4,21 +4,20 @@ import { truncate } from '../utils/string';
...
   if (!phrase && !length) return string;
 
   // Convert the word phrase into a global regular expression (if it isn't
   // already) so we can search the string for matched.
-  const regexp = phrase instanceof RegExp ? phrase : new RegExp(phrase, 'gi');
+  const regexp: RegExp = phrase instanceof RegExp ? phrase : new RegExp(phrase, 'gi');
```
w-4, line 17: Personal preference, I'd remove the type here and the two below, since they are automatically inferred

w-4:
also: `.tsx` file extension

js/src/common/helpers/highlight.ts
daniellesniak: fixed

js/src/common/helpers/highlight.ts
daniellesniak: Sounds good to me, done

js/src/common/helpers/highlight.tsx
```diff
@@ -1,16 +1,16 @@
...
- * @param {String|RegExp} phrase The word or words to highlight.
- * @param {Integer} [length] The number of characters to truncate the string to.
+ * @param string The string to highlight.
+ * @param phrase The word or words to highlight.
+ * @param [length] The number of characters to truncate the string to.
  *     The string will be truncated surrounding the first match.
- * @return {Object}
  */
-export default function highlight(string, phrase, length) {
+export default function highlight(string: string, phrase: string | RegExp, length: number): Mithril.Vnode | string {
```
datitisev, line 13: Length should be `length?: number` as it is an optional parameter.

daniellesniak: In typescript all parameters are required by default so yes - it should be, fixed.

datitisev: True, but we don't have TS errors or warnings enabled. The TS is just converted into plain JS - type hinting is mainly to help during development 🙂

daniellesniak:
I made length parameter optional, everything else look fine to me.

askvortsov1 (approved):
Makes sense to me, thanks!

datitisev:
I wonder - does `m.trust()` only return a `Mithril.Vnode`? I am unable to check right now.

daniellesniak:
It's written in documentation that m.trust() returns Vnode. The same return type is in interface.

Docs:
trust

Interface:
hyperscript interface

datitisev:
> Should I change return type to Mithril.Vnode<any, any> as in interface instead of Mithril.Vnode?

I think so, yeah.

dsevillamartin (approved):
LGTM!

PR #2533, opened 2021-01-09
Add Notification extender beforeSending method

SychO9:
**Part of #1891**

This is needed for extensions to be able to prevent certain notifications from being sent.

Just need to figure out what to do for integration tests for this.

**Confirmed**

### DIFF
--- a/src/Notification/NotificationSyncer.php
+++ b/src/Notification/NotificationSyncer.php
@@ -44,6 +44,11 @@ class NotificationSyncer
     protected static $notificationDrivers = [];
 
     /**
+     * @var array
+     */
+     protected static $beforeSendingCallbacks = [];
+
+    /**
      * Sync a notification so that it is visible to the specified users, and not
      * visible to anyone else. If it is being made visible for the first time,
      * attempt to send the user an email.
@@ -92,6 +97,10 @@ class NotificationSyncer
 
         if (count($toUndelete)) {
             $this->setDeleted($toUndelete, false);
+        }
+
+        foreach (static::$beforeSendingCallbacks as $callback) {
+            $callback($blueprint, $newRecipients);
         }
 
         // Create a notification record, and send an email, for all users
@@ -175,5 +184,13 @@ class NotificationSyncer
     public static function getNotificationDrivers(): array
     {
         return static::$notificationDrivers;
+    }
+
+    /**
+     * @param callable|string $callback
+     */
+    public static function beforeSending($callback): void
+    {
+        static::$beforeSendingCallbacks[] = $callback;
     }
 }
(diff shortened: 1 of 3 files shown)


src/Notification/NotificationSyncer.php
```diff
@@ -94,6 +99,10 @@ public function sync(Blueprint\BlueprintInterface $blueprint, array $users)
             $this->setDeleted($toUndelete, false);
         }
 
+        foreach (static::$beforeSendingCallbacks as $callback) {
+            $callback($blueprint, $newRecipients);
```
askvortsov1, line 103: Should we offer a way for recipients to be modified other than via pass-by-reference? Maybe have the callbacks return a new list of recipients, since that would be the main purpose of inserting logic here?

askvortsov1 (changes requested):
Would also be nice to see a test for this. Here's what I'm thinking:

- Have a custom driver that modifies some attr on each user object.
- Have the beforeSending callback restrict the list of recipients
- Test that only the desired recipients have been modified

src/Notification/NotificationSyncer.php
SychO9: I would be in favor since that is indeed the main purpose here, also I think it's always better to avoid pass by reference where possible.

SychO9:
I like the idea!

tests/integration/extenders/NotificationTest.php
```diff
@@ -119,23 +125,64 @@ public function notification_driver_enabled_types_exist_if_added()
...
+                        unset($users[1]);
+                    }
+
+                    return $users;
+                })
+        );
+
+        $this->prepareDatabase([
+            'discussions' => [
+                ['id' => 1, 'title' => __CLASS__, 'created_at' => Carbon::now()->toDateTimeString(), 'user_id' => 3, 'first_post_id' => 1, 'comment_count' => 1],
```
askvortsov1, line 149: Is this used somewhere in the test case?

tests/integration/extenders/NotificationTest.php
```diff
@@ -119,23 +125,64 @@ public function notification_driver_enabled_types_exist_if_added()
...
     public function getFromUser()
     {
-        // ...
+        return User::find(3);
     }
 
     public function getSubject()
     {
-        // ...
+        return Discussion::find(1);
```
askvortsov1, line 180: Ah, I see. Could this just return null, or some other object?

tests/integration/extenders/NotificationTest.php
askvortsov1: Nvm, found it below. I'm still not sure it's necessary though, see the commend below.

askvortsov1 (changes requested):

askvortsov1 (approved):

tankerkiller125 (approved):

PR #2534, opened 2021-01-09
Policies: treat `true` as `allow`, and `false` as `deny`

### LINKED ISSUE #2525: Policies: treat `true` as `allow`, and `false` as `deny`
## Feature Request

I think the new tiered levels of response in the policies system is well worth it, but is also somewhat inconvenient: I have already forgotten to return one of those 2 variables multiple times. Returning a boolean expression is much more than returning `boolean expression ? $this->allow() : $this->deny()`. Perhaps `AbstractPolicy` should treat `false` as `$this->deny()` and `true` as `$this->allow()`?

askvortsov1:
Fixes #2525

This change allows devs to return boolean expressions directly, which is more intuitive and simpler.

```
// Without This Change
return SOME BOOLEAN LOGIC HERE ? $this->allow() : $this->deny();

// With This Change
return SOME BOOLEAN LOGIC HERE;
```

### DIFF
--- a/src/User/Access/AbstractPolicy.php
+++ b/src/User/Access/AbstractPolicy.php
@@ -70,7 +70,7 @@ abstract class AbstractPolicy
     {
         if ($result === true) {
             $result = $this->allow();
-        } else if ($result === false) {
+        } elseif ($result === false) {
             $result = $this->deny();
         }
 
--- a/src/User/Access/AbstractPolicy.php
+++ b/src/User/Access/AbstractPolicy.php
@@ -43,13 +43,13 @@ abstract class AbstractPolicy
      * @param User $user
      * @param string $ability
      * @param $instance
-     * @return bool|void
+     * @return string|void
      */
     public function checkAbility(User $actor, string $ability, $instance)
     { // If a specific method for this ability is defined,
         // call that and return any non-null results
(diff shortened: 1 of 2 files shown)


tankerkiller125 (approved):

src/User/Access/AbstractPolicy.php
```diff
@@ -58,7 +58,22 @@ public function checkAbility(User $actor, string $ability, $instance)
...
     }
+
+    /**
+     * @param mixed $result
+     * @return string|void
+     */
+    public function sanitizeResult($result)
+    {
+        if ($result === true) {
+            $result = $this->allow();
```
clarkwinkelmann, line 72: You could directly `return` on lines 72 and 74.

Not sure if there's any performance improvement to that but the assignment seems unnecessary :sweat_smile:

clarkwinkelmann (commented):
Looking good! Not tested locally. Just a nitpick below.

clarkwinkelmann (approved):
Looks good!

Something else I notice now: would a comment be useful? Either in the phpdoc or above the first `if`, to say something like "Allow returning true and false in addition to $this->allow() and $this->deny()". To someone not in the know, this code might look like some odd unneeded redundancy

PR #2535, opened 2021-01-10
Add `subscribe` method to event extender

askvortsov1:
Historically, extensions using subscribers has caused problems because subscribers were constructed/applied at extension boot. This caused some classes (e.g. UrlGenerator) to be resolved early, breaking parts of Flarum. For this reason, subscriber support wasn't included in the initial version of the Event extender.

However, updating extensions has shown that there is a legitimate use case for subscribers in organizing clean code; for instance, core's own `DiscussionMetadataUpdater`.

This commit introduces support for subscribers, but only applies them after the app has booted, which avoids the early resolution issues. Since event listeners/subscribers are only intended to be used with domain events, which would never be dispatched during app boot, the late activation of subscribers should not cause issue.

### DIFF
--- a/src/Extend/Event.php
+++ b/src/Extend/Event.php
@@ -22,8 +22,10 @@ class Event implements ExtenderInterface
      * Add a listener to a domain event dispatched by flarum or a flarum extension.
      *
      * The listener can either be:
-     *  - a callback function or
+     *  - a callback function
      *  - the class attribute of a class with a public `handle` method, which accepts an instance of the event as a parameter
+     *  - an array, where the first argument is an object or class name, and the second argument is the method on the
+     *    first argument that should be executed as the listener
      *
      * @param string $event
      * @param callable|string $listener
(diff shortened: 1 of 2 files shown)


SychO9 (approved):

src/Extend/Event.php
```diff
@@ -34,12 +35,32 @@ public function listen(string $event, $listener)
         return $this;
     }
 
+    /**
+     * Add a subscriber for a set of domain events dispatched by flarum or a flarum extension.
+     *
+     * @param string $subscriber
```
clarkwinkelmann, line 41: Should this parameter be a bit more documented, like `listen` ?

I see Laravel's own phpdoc for `subscribe` is also lacking a proper description. But they do allow `object|string`.

Since there's no interface for listeners I'm not exactly sure what we can say, but we should at least specify it's a `::class` parameter.

clarkwinkelmann (changes requested):

clarkwinkelmann (approved):

PR #2541, opened 2021-01-13
Convert common helpers to Typescript

daniellesniak:
See flarum/framework#3533

### DIFF
--- a/js/src/common/helpers/avatar.tsx
+++ b/js/src/common/helpers/avatar.tsx
@@ -1,26 +1,29 @@
+import * as Mithril from "mithril";
+import User from "../models/User";
+
 /**
  * The `avatar` helper displays a user's avatar.
  *
- * @param {User} user
- * @param {Object} attrs Attributes to apply to the avatar element
- * @return {Object}
+ * @param user
+ * @param attrs Attributes to apply to the avatar element
  */
-export default function avatar(user, attrs = {}) {
+
+export default function avatar(user: User, attrs: Object = {}): Mithril.Vnode {
   attrs.className = 'Avatar ' + (attrs.className || '');
-  let content = '';
+  let content: string = '';
 
   // If the `title` attribute is set to null or false, we don't want to give the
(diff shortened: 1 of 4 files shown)


js/src/common/helpers/avatar.tsx
```diff
@@ -1,26 +1,28 @@
+import * as Mithril from 'mithril';
+import User from '../models/User';
+
 /**
  * The `avatar` helper displays a user's avatar.
  *
- * @param {User} user
- * @param {Object} attrs Attributes to apply to the avatar element
- * @return {Object}
+ * @param attrs Attributes to apply to the avatar element
```
askvortsov1, line 7: I wouldn't want to have one param documented but the other not

js/src/common/helpers/avatar.tsx
```diff
@@ -1,26 +1,28 @@
...
  */
-export default function avatar(user, attrs = {}) {
+
```
askvortsov1, line 9: Extra space?

askvortsov1 (changes requested):
Looking good, just a few nitpicks

js/src/common/helpers/avatar.tsx
datitisev: The other is still documented in the typehinting. Attrs only remains for the description in the JSDoc. I don't know if we want to do that or keep tham all, howeverr 🤷

daniellesniak: I think we should leave all parameters in JSDoc if there's any with description and get rid of all parameters if none of them have any description. By the way this should resolve PHPStorm's warning "Parameter is not described in JSDoc".

js/src/common/helpers/avatar.tsx
daniellesniak: Yes, fixed in avatar.tsx and username.tsx.

askvortsov1 (approved):
Thank you!

dsevillamartin (approved):
:+1:

PR #2548, opened 2021-01-16
Various iOS scroll improvements

askvortsov1:
Addresses https://discuss.flarum.org/d/25643-up-scroll-skips-posts/10

Changes made are:
- throttling the loading of additional pages
- only updating the scrubber once a page has loaded to mitigate jumping scrubbers

I was unable to fix an issue where it skips posts when a finger is held on the screen while at the top of the page, but that's present in previous releases of Flarum too (and I've managed to replicate that in Discourse as well): seems like an iOS quirk. However, fast/momentum based scrolls (vast majority), or scrolls when the user lets go when they get to the top, should be almost completely fixed.


datitisev:
Code looks fine. How much does the import increase bundle size by?

askvortsov1:
355.3 kB without, 357.1kB with

SychO9 (approved):

askvortsov1:
> 355.3 kB without, 357.1kB with

@datitisev is that alright with you? I got my hands on an ipad to test the way it is now, and it's really inconvenient for iOS users, so it'd be good to get this fixed.

datitisev:
@askvortsov1 Yeah fine with me.

PR #2555, opened 2021-01-19
Add user badges to post preview #1765

### LINKED ISSUE #248: [Feature] See user badge when reply

### LINKED ISSUE #1765: [Feature] See user badge when reply
Feature approved by Toby

I would suggest to add the badge when user post or reply also where the red row show. This alow who are posting to remember he his posting from a special profile... just remember is not a reply from a "normal" profile as actually the forum show.

This is useful also when a new post is created. Will be nice because who are writting a post can see they are writting as badged user.

Consider who this change will be useful to – most Flarum forums, or just a few?

This is part of the community why should be an extension?

imorland:
**Fixes flarum/issue-archive#248 **

For consistency, the user badges are added to the post preview, so that it cosmetically fits with the rest of the post stream

**Screenshot**
image

**Confirmed**

### DIFF
--- a/js/src/forum/components/ReplyPlaceholder.js
+++ b/js/src/forum/components/ReplyPlaceholder.js
@@ -5,6 +5,7 @@ import avatar from '../../common/helpers/avatar';
 import username from '../../common/helpers/username';
 import DiscussionControls from '../utils/DiscussionControls';
 import ComposerPostPreview from './ComposerPostPreview';
+import listItems from '../../common/helpers/listItems';
 
 /**
  * The `ReplyPlaceholder` component displays a placeholder for a reply, which,
@@ -25,6 +26,7 @@ export default class ReplyPlaceholder extends Component {
                 {avatar(app.session.user, { className: 'PostUser-avatar' })}
                 {username(app.session.user)}
               </h3>
+              <ul className="PostUser-badges badges">{listItems(app.session.user.badges().toArray())}</ul>
             </div>
           </header>
           <ComposerPostPreview className="Post-body" composer={app.composer} surround={this.anchorPreview.bind(this)} />


askvortsov1:
Rebasing on origin/master should fix the failing tests

imorland:
Thanks! Done

askvortsov1:
Looks like some unrelated commits got in, perhaps you might have pushed instead of force pushed?

imorland:
Whoops

askvortsov1:
Would we want to include the full PostUser composer, so the user card shows up on hover?

imorland:
I did look at that. As things stand, the full `PostUser` also requires a `post` attr, which at this composing/preview state does not actually exist

askvortsov1:
Fair, another argument against PostUser would be that there's not that big a need for the current user to see more information about themselves.

js/src/forum/components/ReplyPlaceholder.js
```diff
@@ -25,6 +26,7 @@ export default class ReplyPlaceholder extends Component {
                 {avatar(app.session.user, { className: 'PostUser-avatar' })}
                 {username(app.session.user)}
               </h3>
+              <ul className="PostUser-badges badges">{listItems(app.session.user.badges().toArray())}</ul>
```
askvortsov1, line 29: Another concern: we're reusing CSS from an unrelated component here. This is fairly common in official extensions, but I'd like to keep core's CSS as organized as we can here.

askvortsov1 (changes requested):

js/src/forum/components/ReplyPlaceholder.js
imorland: As this change is to retain consistency with the rest of the post stream, I'd put forward that re-using the CSS is valid in this scenario, otherwise there is a (small) overhead of maintaining two identical CSS classes?

askvortsov1: Looking at this again, I see that other parts of this component are already using the `PostUser` css class so my concern is invalid

askvortsov1 (approved):

SychO9 (approved):

PR #2557, opened 2021-01-20
Remove deprecated PHP events, bootstrap.php fallback

### LINKED ISSUE #1557: Extensions: Drop support for bootstrap.php
Kill the fallback introduced in #1556.

askvortsov1:
Fixes #1557 

I left in the BC layer for old-style extension modals, as that's not that urgent a deprecation, and some extensions may be behind. I think we should leave it in for one more release, but if we decide otherwise, we can remove it in another PR.


SychO9 (changes requested):
Tests are failing because you missed the dispatching of `WillGetData` event in `AbstractSerializeController`

KyrneDev (approved):
Straightforward changes. Tests are green and so is my review.

src/Api/Serializer/AbstractSerializer.php
```diff
@@ -153,27 +140,21 @@ public function getRelationship($model, $name)
...
-        if ($relationship && ! ($relationship instanceof Relationship)) {
-            throw new LogicException(
-                'GetApiRelationship handler must return an instance of '.Relationship::class
-            );
-        }
+                if (! ($relationship instanceof Relationship)) {
+                    throw new LogicException(
+                        'GetApiRelationship handler must return an instance of '.Relationship::class
+                    );
+                }
```
SychO9, line 153: hmm, I'm not sure here.

The method used to return `Relationship|null`, but with this change, it's guaranteed to always return Relationship or error out, meaning callbacks which used to be able to return null and work fine, will error out.

I can't think of any real examples why a relationship callback would return null, not have I looked into it, but this is definitely a breaking change.

SychO9 (changes requested):

src/Api/Serializer/AbstractSerializer.php
SychO9: Ah nevermind, it wouldn't get passed the is_callable test

askvortsov1: Actually, you bring up a really good point. That IS actually a valid use case, and in one of my own extensions no less (https://github.com/askvortsov1/flarum-moderator-warnings/blob/master/extend.php#L65-L72)! My goodness is this PR a mess...

src/Foundation/AbstractValidator.php
```diff
@@ -38,25 +36,18 @@ public function addConfiguration($callable)
...
      */
     protected $translator;
 
     /**
      * @param Factory $validator
-     * @param Dispatcher $events
      * @param TranslatorInterface $translator
      */
-    public function __construct(Factory $validator, Dispatcher $events, TranslatorInterface $translator)
+    public function __construct(Factory $validator, TranslatorInterface $translator)
```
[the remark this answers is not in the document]
askvortsov1: That's a tough one... It would be a breaking change and there's not really a graceful way around it, but I don't really see a reason to keep an extra param around, especially such an unnecessary one.

I'd lean towards removing this now, while we're in beta.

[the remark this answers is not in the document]
SychO9: Yea, there is no way to have some sort of BC layer, we can only directly remove it.

We should probably document it when the time comes, it's a good thing there appears to be no public extensions that would break, but we can't know for closed source code.

SychO9 (approved):

PR #2558, opened 2021-01-20
Set this.changingState back to false in ExtensionPage if an error occurs

### LINKED ISSUE #2536: When an extension dependency is not met, the extension page still switches to enabled
## Bug Report

**Current Behavior**
When you try enabling an extension that doesn't have its extension dependency met, the error message is correctly shown and the extension is not enabled. However the page shows the extension as enabled until you change page.

**Steps to Reproduce**
1. Go to admin panel
2. Disable Approval & Flags
3. Try to enable Approval
4. Observe error message correctly shown
5. Observe Switch stays in enabled state and page says "No settings" and "No permissions"

**Expected Behavior**
The page should switch back to disabled state upon error

**Screenshots**
Capture d’écran 2021-01-12 à 02 41 18

**Environment**
- Flarum version: beta 15
- Website URL: https://beta.flarum.site/

**Additional Context**
Observed locally. Reproduced on the demo website.

datitisev:
**Fixes #2536**

- Set `this.changingState = false` in `onerror` handler.
- Only use `this.changingState` for Switch status & its loading attribute
  - do not take it into account in whether to show the ext settings as that makes it look bad (see reviewer focus below)

<details>
<summary><b>Reviewers should focus on:</b></summary>

I feel like we should instead get rid of `changingState` completely. It causes the UI to glitch out and there's no reason for the text & info shown to update while the "Please wait" modal is active and then be wrong.

https://user-images.githubusercontent.com/6401250/105232641-acc6bf80-5b36-11eb-8235-cf97a436af02.mp4

</details>

**Video:**

https://user-images.githubusercontent.com/6401250/105638030-8a86b780-5e3e-11eb-922d-6ce8850eae7d.mp4

**Confirmed**

### DIFF
--- a/js/src/admin/components/ExtensionPage.js
+++ b/js/src/admin/components/ExtensionPage.js
@@ -346,6 +346,8 @@ export default class ExtensionPage extends Page {
       app.modal.close();
     }, 300); // Bootstrap's Modal.TRANSITION_DURATION is 300 ms.
 
+    this.changingState = false;
+
     if (e.status !== 409) {
       throw e;
     }
--- a/js/src/admin/components/ExtensionPage.js
+++ b/js/src/admin/components/ExtensionPage.js
@@ -59,6 +59,8 @@ export default class ExtensionPage extends Page {
   }
 
   header() {
+    const isEnabled = this.isEnabled();
+
     return [
       <div className="ExtensionPage-header">
         <div className="container">
@@ -75,10 +77,12 @@ export default class ExtensionPage extends Page {
           </div>
           <div className="helpText">{this.extension.description}</div>
           <div className="ExtensionPage-headerItems">
-            <Switch state={this.isEnabled()} onchange={this.toggle.bind(this, this.extension.id)}>
-              {this.isEnabled(this.extension.id)
-                ? app.translator.trans('core.admin.extension.enabled')
-                : app.translator.trans('core.admin.extension.disabled')}
+            <Switch
+              state={this.changingState ? !isEnabled : isEnabled}
+              loading={this.changingState}
+              onchange={this.toggle.bind(this, this.extension.id)}
+            >
+              {isEnabled ? app.translator.trans('core.admin.extension.enabled') : app.translator.trans('core.admin.extension.disabled')}
             </Switch>
             <aside className="ExtensionInfo">
               <ul>{listItems(this.infoItems().toArray())}</ul>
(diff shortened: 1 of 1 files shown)


askvortsov1:
> I feel like we should instead get rid of `changingState` completely. It causes the UI to glitch out and there's no reason for the text & info shown to update while the "Please wait" modal is active and then be wrong.

I think I would prefer this too: we have a continuous stream of animation via the loading modal, so it's not like users wouldn't get visual confirmation that they're trying to activate/deactivate an extension.

datitisev:
> That being said, it would be nice to keep the header switch itself enabling; perhaps `changingState` could be moved out of isEnabled, and directly into the "checked" criteria for the switch?

I like this. I also added set it to the `loading` attr because as we don't want the Enabled/Disabled text changing, and in case there's an issue with the modal not appearing or something, the user knows that it's possibly not fully enabled/disabled and it was in progress.

datitisev:
This currently has an issue - the switch does not go back to the original state after the error.

js/src/admin/components/ExtensionPage.js
```diff
@@ -75,10 +77,12 @@ export default class ExtensionPage extends Page {
...
-            <Switch state={this.isEnabled()} onchange={this.toggle.bind(this, this.extension.id)}>
-              {this.isEnabled(this.extension.id)
-                ? app.translator.trans('core.admin.extension.enabled')
-                : app.translator.trans('core.admin.extension.disabled')}
+            <Switch
+              state={this.changingState ? !isEnabled : isEnabled}
+              loading={this.changingState}
+              onchange={this.toggle.bind(this, this.extension.id)}
+            >
+              {isEnabled ? app.translator.trans('core.admin.extension.enabled') : app.translator.trans('core.admin.extension.disabled')}
```
askvortsov1, line 85: Shouldn't the switch and the label display the same value?

askvortsov1 (changes requested):

js/src/admin/components/ExtensionPage.js
datitisev: I don't think so. The text shows whether the extension is currently enabled or disabled, while the Switch can show that the extension is being enabled/disabled through it's loading spinner. The switch is visible behind the modal, while the text not so much. 

I don't think the text should change at all.

askvortsov1: Ah got it, I see your point.

askvortsov1 (approved):

SychO9 (approved):

Issue #1913, opened 2019-10-16
Touch devices of certain widths forced to double tap to open discussion


zerosonesfun:
## Bug Report

**Current Behavior**
Touch devices of certain widths are forced to double tap to open discussions. This is because the CSS media rules only work up to a certain number of pixels. Which works fine for most mobile devices in Portrait mode. However, on an iPhone for example, if you rotate to Landscape, or if you use an iPad in Portrait or Landscape, then the "mobile" version of the site is not used and it takes two taps to open a post. This is because you are served the regular desktop version of the site which includes the three dot toggle menu to the right of each post in the discussions list and the way things are currently coded, the first tap on a touch device does not go into the discussion probably because it is wondering if you are trying to instead open the drop down toggle menu. 

This toggle menu is not needed on touch devices no matter the width because the slide to toggle feature works on touch devices even if the normal desktop CSS is being served. 

See end of post for possible solution.

**Steps to Reproduce**
1. Using an iPhone in landscape mode or an iPad go to any Flarum forum.
2. Tap a discussion in the discussions list.

**Expected Behavior**
One tap should open the discussion.

**Screenshots**
I don't think screenshots are truly needed for this report.

**Environment**
- Any iOS environment and most likely any other touch environment. Browser doesn't matter. This happens in Safari iOS, Chrome iOS, and Firefox iOS.

**Possible Solution**
The exact solution is up to core devs, but a temporary fix for current Flarum webmasters is to copy and paste the following CSS into their custom CSS:
~~~
.DiscussionListItem .Dropdown-toggle.Button {
display: none;
}
.no-touch .DiscussionListItem .Dropdown-toggle.Button {
display: block;
}
~~~

What is this doing? If a "no-touch" device (a regular desktop computer) is detected display the drop down toggle as normal (display:block). This is possible because Flarum already conveniently adds "no-touch" to the body tag with non-touch devices. And then the display:none part of the above CSS ensures the toggle button is not displayed with touch devices. You can still slide to toggle on touch devices, but the 3 dot toggle is hidden which fixes the double tap issue.

OrdinaryJellyfish:
I remember reading about this somewhere. It's intended behavior in iOS, though it can get annoying depending on the use case. Android doesn't do this, as it fires the click event on the first tap as well as hover. Further reading: https://css-tricks.com/annoying-mobile-double-tap-link-issue/

zerosonesfun:
[...]

PR #2561, opened 2021-01-23
Update DiscussionListItem.less to fix touch double tap

zerosonesfun:
Adds rule for devices without hover capabilities which hides discussion item toggle button. This prevents having to tap twice to open the discussion. Also, the toggle button is not used on mobile devices and so it should be completely hidden. Additional nested rules could be added here in the future if needed. 

**Fixes https://github.com/flarum/core/issues/1913**

Adds rule for devices without hover capabilities which hides discussion item toggle button. This prevents having to tap twice to open the discussion. Also, the toggle button is not used on mobile devices and so it should be completely hidden. Additional nested rules could be added here in the future if needed. See issue: https://github.com/flarum/core/issues/1913

Fix uses a relatively new hover CSS media feature which can be used to test whether the user’s primary input mechanism can hover over elements. It is compatible will all major/modern browsers. No Internet Explorer support but on August 17th, 2021, Internet Explorer 11 will no longer be supported by Microsoft. All IE versions are on their way out. Also, if someone visits a forum with this one browser, the worst that would happen is they would have to double tap like before because the browser wouldn’t recognize the rule. 

If reviewer does not have an iPhone or iPad, reviewer could at least confirm that this fix does not affect other devices. Toggle button should still work on desktop. If reviewer has an iPhone or iPad they will be able to open a discussion with one tap. 

**Confirmed**

### DIFF
--- a/less/forum/NotificationList.less
+++ b/less/forum/NotificationList.less
@@ -103,13 +103,6 @@
     font-weight: bold;
     text-transform: uppercase;
   }
-
-  @media (any-hover: none) { 
-  .Notification-action.Button { 
-    display: none; } 
-    }
-  }
-
   .Notification-action {
     float: right;
     display: none;
--- a/less/forum/NotificationList.less
+++ b/less/forum/NotificationList.less
@@ -104,6 +104,12 @@
     text-transform: uppercase;
   }
 
+  @media (any-hover: none) { 
+  .Notification-action.Button { 
+    display: none; } 
+    }
+  }
+
   .Notification-action {
     float: right;
     display: none;
--- a/less/forum/DiscussionListItem.less
+++ b/less/forum/DiscussionListItem.less
@@ -110,7 +110,7 @@
   }
 }
 
-@media (hover: none) { 
+@media (hover-any: none) { 
   .DiscussionListItem .Dropdown-toggle.Button { 
     display: none; } 
 }
--- a/less/forum/DiscussionListItem.less
+++ b/less/forum/DiscussionListItem.less
@@ -110,6 +110,10 @@
   }
 }
 
+@media (hover: none) { 
+  .DiscussionListItem .Dropdown-toggle.Button { 
+    display: none; } 
+}
 
 @media @phone {
   .DiscussionListItem-controls {
--- a/less/forum/DiscussionListItem.less
+++ b/less/forum/DiscussionListItem.less
@@ -110,7 +110,7 @@
   }
 }
 
-@media (hover-any: none) { 
+@media (any-hover: none) { 
   .DiscussionListItem .Dropdown-toggle.Button { 
     display: none; } 
 }


askvortsov1:
When I tested this locally, it removed it on desktop as well for me. Same with pointer: none. Not sure why though?

zerosonesfun:
> When I tested this locally, it removed it on desktop as well for me. Same with pointer: none. Not sure why though?

Does your Desktop have any touch or touch-like features? It looks like I should have used “any-hover” instead of “hover.” Any hover would say, “if this device has any normal hover abilities at all, don’t hide the toggle button. 

https://developer.mozilla.org/en-US/docs/Web/CSS/@media/any-hover

askvortsov1:
Maybe it's just my computer being weird? Either way, it still considers my computer an `any-hover: none` device. Very odd...

zerosonesfun:
> Maybe it's just my computer being weird? Either way, it still considers my computer an `any-hover: none` device. Very odd...

Weird. But you can see that the any-hover rule is being applied? Your computer wishes it was a touch screen. It’s got touch screen envy. 

I was trying to avoid this fix relying on screen width. Because there’s so many different sizes of tablets.

zerosonesfun:
There are other options to try as well:

https://css-tricks.com/touch-devices-not-judged-size/

askvortsov1 (approved):
I have confirmed that the issues I experienced are due to an issue with my Wayland setup (and even then, only occured on chrome). I'm in favor of merging this.

@zerosonesfun The issue also mentioned some problems with notifications; would you be able to include a fix for that here as well?

zerosonesfun:
> I have confirmed that the issues I experienced are due to an issue with my Wayland setup (and even then, only occured on chrome). I'm in favor of merging this.
> 
> @zerosonesfun The issue also mentioned some problems with notifications; would you be able to include a fix for that here as well?

I started to add a change for the notifications but then I removed the change and put the file back to the way it was because I realized that notifications probably should stay as-is for now. The reason is, the first time makes a little mark as read check mark appear. Then you can tap the check mark to mark the notification as read without having to visit the discussion if you don’t want to. So, although a little unintuitive, I don’t think the notification double tap on mobile is an “issue” per se. It’s kind of a feature. You can tap once, then you have the option of marking as read, or going to the discussion. 

What should be done down the road (I don’t know how to do it) is the notifications list should have a swipe to mark as read feature just like you can swipe discussion list items on mobile... on mobile it should be, swipe to mark as read, tap to go to the discussion.

askvortsov1 (approved):

less/forum/NotificationList.less
```diff
@@ -103,7 +103,6 @@
     font-weight: bold;
     text-transform: uppercase;
   }
-
```
SychO9, line 106: you removed a line by mistake

SychO9 (changes requested):

zerosonesfun:
Easier to re-do.

PR #2563, opened 2021-01-23
Use extension names instead of IDs when erroring on enable/disable reqs

### LINKED ISSUE #2341: Show extension titles instead of IDs for dependency error messages
Right now, if you enable an extension that's missing dependencies, or disable an extension with other extensions depending on it, you'll get an error message that lists extension IDs of all extensions involved. It might be more readable to users to show extension titles instead of extension IDs.

Screenshots:
- https://imgur.com/X2npKFt
- https://imgur.com/rioQMXc

datitisev:
**Fixes #2341**

- Use extension name where available
- Create `Extension@getTitle` method
- Create `ExtensionManager::pluckTitles` method

image

**Confirmed**

---

**_ORIGINAL DESCRIPTION_**

**Changes needed:**

Is there a better way to obtain the extension name? I assume we also want to change the extension to use its name (apart from dependencies/dependents) in error message, but this is a lot of duplicated code.
Not sure if we want to have a fallback to ID in case title doesn't exist.
I'm surprised the `Extension` class doesn't have a method to obtain the extension title.

- Should the method name be renamed? Could technically be a breaking change
- Should these 2 methods be condensed into 1 in a separate class? The logic is the exact same.

### DIFF
--- a/src/Extension/Exception/DependentExtensionsException.php
+++ b/src/Extension/Exception/DependentExtensionsException.php
@@ -11,6 +11,7 @@ namespace Flarum\Extension\Exception;
 
 use Exception;
 use Flarum\Extension\Extension;
+use Flarum\Extension\ExtensionManager;
 
 /**
  * This exception is thrown when someone attempts to disable an extension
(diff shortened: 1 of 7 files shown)


askvortsov1:
Is there a use case for the old method outside of this? If no, I say it's fine.

Maybe we could move them to a static util method on ExtensionManager to avoid duplication?

SychO9:
> I'm surprised the Extension class doesn't have a method to obtain the extension title.

I would be in favour of introducing that method here, and using it in the codebase (there should be 4 spots in total with this PR)

> Should these 2 methods be condensed into 1 in a separate class? The logic is the exact same.

I think it's fine to leave it as is, I feel like it'd be overdoing it.

datitisev:
@askvortsov1
>Is there a use case for the old method outside of this? If no, I say it's fine.
>
> Maybe we could move them to a static util method on ExtensionManager to avoid duplication?

Not sure what you mean exactly with the second.

---

@SychO9 
> I would be in favour of introducing that method here, and using it in the codebase (there should be 4 spots in total with this PR)

I assume this would look something like `getTitle()` and return the title OR extension ID if the title doesn't exist, as a fallback?

datitisev:
How do the added `Extension@getTitle` and `ExtensionManager::pluckTitles` methods look? I also replaced other instances of the composer attribute with the `getTitle()` method - but in separate commit so it's easier to review (not that there's much).

askvortsov1:
Looks good to me!

askvortsov1 (approved):

SychO9 (approved):

Issue #2345, opened 2020-10-02
Remove and replace jQuery and Bootstrap JS


askvortsov1:
Followup to https://github.com/flarum/core/issues/1821

Right now, we're considering:

* [ ]  jquery => cash (Typescript, better maintained, smaller) + velocity (separate animation library)
* [ ]  bootstrap => micromodal, tooltipjs, custom solutions

askvortsov1:
Relevant: https://github.com/flarum/core/issues/2303#issuecomment-751206229 (replacing bootstrap modals with something better, or a vanilla solution).

askvortsov1:
> Relevant: [#2303 (comment)](https://github.com/flarum/core/issues/2303#issuecomment-751206229) (replacing bootstrap modals with something better, or a vanilla solution would fix that issue).

askvortsov1:
Should also strive to use vanilla JS wherever possible to make the eventual transition simpler.

davwheat:
Have you considered using [`clsx`](https://www.npmjs.com/package/clsx) as opposed to `classnames`? It's a drop-in replacement and far, far smaller.

askvortsov1:
If it works, I'm all for it!

askvortsov1:
@datitisev I think we should also include your idea of tooltips as a composable component as part of this. Do we have a separate issue for that?

datitisev:
@askvortsov1 I don't think we have a separate issue; I don't remember making one & wasn't able to find one.

davwheat:
I'm going to try and push for modals to use the native `<dialog>` element as opposed to any custom JS-based solutions.

[It's supported in Firefox and Chrome for Android](https://caniuse.com/dialog), and [Google even made their own polyfill](https://github.com/GoogleChrome/dialog-polyfill) which works with their desktop version of Chrome.

Check out the demos to see what's possible: https://googlechrome.github.io/dialog-polyfill/

SychO9:
I love that CSS is powerful enough to do things like tooltips, but the problem with this solution is that it uses psuedo elements to achieve that, quoting:

> Drawbacks
Balloon.css make use of pseudo-elements thus self-closing elements such as `<img>`, `<input>` and `<hr>` will not render tooltips.
>
> Also keep in mind that if pseudo elements are already in use on an element, the tooltip will conflict with them resulting in potential bugs.

davwheat:
Any issue arising from the use of pseudoelements could be resolved immediately by wrapping the node in a plain div or span.

[...]

PR #2570, opened 2021-01-26
Add new modal options to require confirmation before closing

davwheat:
> For clarity, this PR was originally for adding attrs which disabled modal dismissing via backdrop clicks and Esc keypresses, but was changed after discussion.

I'm aware that Bootstrap Modals is set to get replaced by Micromodal (or another library) in the near future, but implementing this functionality now should (hopefully) mean it gets carried across to the new library. 🤞 

This provides more customisability for modals, allowing extensions (and core) to determine if confirmation should be required before closing the modal.

The current field `isDismissible` was previously incorrectly labelled as only determining if an X button is visible, not stating that it also controlled if a backdrop click or Esc key press would dismiss the modal. (To be fair, I can't think of any situation where you want to remove a close button and keep Esc and backdrop dismissal.)

This PR adds two new static fields: `requireImplicitCloseConfirmation` and `requireExplicitCloseConfirmation`. This allows extensions to set whether confirmation should be needed for implicit actions, such as backdrop clicks and Esc keypresses.

I've been in the middle of completing several modals with forms in some extensions where I begin to click the button, then decide to edit one of the form entries, so drag my cursor off the submit button onto the backdrop, which then closes the modal. I then either have to re-complete the entire modal again, or just don't bother.

Another situation where backdrop dismissals frustrate me is if I'm copying text from the modal. I tend to click and drag off the end of the component (bottom or sides) to quickly copy all the text I need quickly. Releasing the click causes the modal to dismiss.

While extensions should probably find a way around this, or Bootstrap itself should be patched, in the long run, more customisability is better (in my mind), so exposing these options to extensions is a good thing!

---

This also addresses an issue with Esc-based closure. BS Modals currently only detect Esc presses when performed within their content, which requires that they have focused content. Modals without any autofocused content (no inputs, selects, or textareas) cannot be closed via Esc by default.

This PR disables the default detection techniques used by BS Modal and instead implements custom handlers to match behaviour between the confirmation-based and standard closures, and to fix this issue.

https://user-images.githubusercontent.com/7406822/111080808-0d9bc400-84f8-11eb-8c0e-d0e7fac5e2cf.mp4

Where should `requireImplicitCloseConfirmation` be overridden to `true`? I've thought of the tag selection modal off the top of my head, but please suggest any other locations.

**Confirmed**

### DIFF
--- a/js/src/common/components/Modal.js
+++ b/js/src/common/components/Modal.js
@@ -10,7 +10,9 @@ import Button from './Button';
  */
 export default class Modal extends Component {
   /**
-   * Determine whether or not the modal should be dismissible via an 'x' button.
+   * Determine whether or not the modal should be dismissible.
+   *
+   * This overrides `doesBackdropClickDismiss` and `doesEscKeyDismiss`.
    */
   static isDismissible = true;
 
@@ -21,9 +23,9 @@ export default class Modal extends Component {
    * inadvertantly trigger this.
    *
    * It's recommended to disable this for modals which contain forms that may
-   * take a user a longer amount of time to complete.
+   * take a user a longer amount of time to complete to prevent frustration.
    */
-  static doesBackdropDismiss = false;
+  static doesBackdropClickDismiss = false;
 
   /**
    * Determine whether pressing the Escape key should dismiss the modal.
--- a/js/src/common/components/Modal.js
+++ b/js/src/common/components/Modal.js
@@ -12,7 +12,8 @@ export default class Modal extends Component {
   /**
    * Determine whether or not the modal should be dismissible.
    *
-   * This overrides `doesBackdropClickDismiss` and `doesEscKeyDismiss`.
+   * Setting to `false` will override `doesBackdropClickDismiss` and
+   * `doesEscKeyDismiss`.
    */
   static isDismissible = true;
 
@@ -25,12 +26,12 @@ export default class Modal extends Component {
    * It's recommended to disable this for modals which contain forms that may
    * take a user a longer amount of time to complete to prevent frustration.
    */
-  static doesBackdropClickDismiss = false;
+  static dismissOnBackdropClick = false;
 
   /**
    * Determine whether pressing the Escape key should dismiss the modal.
    */
-  static doesEscKeyDismiss = true;
+  static dismissOnEscapeKeyPress = true;
 
   /**
    * Attributes for an alert component to show below the header.
--- a/js/src/common/components/Modal.js
+++ b/js/src/common/components/Modal.js
@@ -11,27 +11,28 @@ import Button from './Button';
 export default class Modal extends Component {
   /**
    * Determine whether or not the modal should be dismissible.
-   *
-   * Setting to `false` will override `doesBackdropClickDismiss` and
-   * `doesEscKeyDismiss`.
    */
   static isDismissible = true;
 
   /**
-   * Determine whether a click on the modal backdrop should dismiss the modal.
+   * Require the user to confirm they wish to close the modal **after clicking
+   * on the backdrop or pressing Escape**.
    *
-   * Clicking and dragging from certain elements onto the backdrop can sometimes
-   * inadvertantly trigger this.
+   * It's recommended to keep this enabled for modals which contain forms that
+   * may take a user a longer amount of time to complete to prevent frustration.
    *
-   * It's recommended to disable this for modals which contain forms that may
-   * take a user a longer amount of time to complete to prevent frustration.
+   * This does not affect closure via the close button.
    */
-  static dismissOnBackdropClick = false;
+  static requireImplicitCloseConfirmation = true;
 
   /**
-   * Determine whether pressing the Escape key should dismiss the modal.
+   * Require the user to confirm they wish to close the modal **after clicking
+   * the top close button, or when calling `hide()`**.
+   *
+   * This does not affect closure via an Esc key press, or a click on the
+   * backdrop.
    */
-  static dismissOnEscapeKeyPress = true;
+  static requireExplicitCloseConfirmation = false;
 
   /**
    * Attributes for an alert component to show below the header.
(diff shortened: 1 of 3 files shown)


askvortsov1 (changes requested):
Looks like some dist changes were accidentially included, could you remove those before we get into review?

js/src/common/components/ModalManager.js
```diff
@@ -33,7 +33,11 @@ export default class ModalManager extends Component {
   }
 
   animateShow(readyCallback) {
-    const dismissible = !!this.attrs.state.modal.componentClass.isDismissible;
+    const isDismissible = !!this.attrs.state.modal.componentClass.isDismissible;
+
+    // If the modal isn't dismissible, set these options to false.
+    const backdropDismissible = isDismissible ? !!this.attrs.state.modal.componentClass.doesBackdropClickDismiss : false;
```
askvortsov1, line 39: Couldn't these be `isDismissible && !!theAttr`?

js/src/common/components/Modal.js
```diff
@@ -10,10 +10,28 @@ import Button from './Button';
...
+  /**
+   * Determine whether a click on the modal backdrop should dismiss the modal.
+   *
+   * Clicking and dragging from certain elements onto the backdrop can sometimes
+   * inadvertantly trigger this.
+   *
+   * It's recommended to disable this for modals which contain forms that may
+   * take a user a longer amount of time to complete to prevent frustration.
+   */
+  static doesBackdropClickDismiss = false;
```
askvortsov1, line 28: How do we feel about this wording vs `dismissOnBackdropClick`?

js/src/common/components/ModalManager.js
```diff
@@ -46,10 +50,23 @@ export default class ModalManager extends Component {
     this.$()
       .one('shown.bs.modal', readyCallback)
       .modal({
-        backdrop: dismissible || 'static',
-        keyboard: dismissible,
+        // 'static' means that a backdrop click doesn't dismiss the modal
+        backdrop: backdropDismissible ? true : 'static',
```
askvortsov1, line 54: Why is this change necessary?

js/src/common/components/ModalManager.js
```diff
@@ -46,10 +50,23 @@ export default class ModalManager extends Component {
...
       .modal('show');
+
+    // Disabling backdrop dismissal also disables keyboard dismissibility
+    // but we want to be able to provide key-based dismissal while having
+    // backdrop-based disabled. To do this, we set up our own event
+    // handler to hide the modal if we press Escape.
+    if (!backdropDismissible && keyboardDismissible) {
+      this.$().on('keypress', function (e) {
+        if (e.key === 'Escape') {
+          this.animateHide();
```
askvortsov1, line 66: Not a change request but a comment for reviewers: we'll need to test that dismissing this way doesn't result in weird UI glitches.

askvortsov1 (changes requested):

davwheat:
Dist changes totally my fault... I think I got too greedy with `git commit -a`! On that now.

js/src/common/components/ModalManager.js
davwheat: Yes, it could and it definitely should!

js/src/common/components/Modal.js
davwheat: That name sounds way better in my opinion. I think we should go for it.

js/src/common/components/ModalManager.js
davwheat: Oops, it isn't. I think this slipped through the net while I was trying other approaches.

I'll revert back to the `||` method, but keep the comment as it makes the reasoning behind `"static"` clearer.

js/src/common/components/ModalManager.js
davwheat: I just realised that when `isDismissible` is true and backdrop dismissal is false, the modal will still close on backdrop click if using the `||` method as `isDismissible` will override the custom properties.

My intention was to only override if the dialog was not dismissible, otherwise we're in the same situation as before this PR.

askvortsov1: But wouldn't it work with `&&`?

davwheat: Yes... if only I didn't misread what you wrote 😅

askvortsov1 (changes requested):
Had a chance to test this out. I get where this is coming from with a desire to prevent accidential closings, but not being able to click outside of the modal to close it is problematic IMO. It definitely shouldn't be the default behavior, but I'm not even sure whether we want to make it an option, since it might be confusing to users.

davwheat:
Maybe instead of fully disabling closing via backdrop click, it should enable a prompt, similar to discarding a post draft?

askvortsov1:
> Maybe instead of fully disabling closing via backdrop click, it should enable a prompt, similar to discarding a post draft?

I think a "confirm modal close" system would be great!

davwheat:
Whoops... accidentally closed when I was cleaning up my fork 🙈

askvortsov1:
How would this look if we proceeded with https://github.com/flarum/core/issues/2345#issuecomment-798904086? Would it still be necessary / maybe it might make sense to implement it after the modal rewrite, if we can get it into the release?

davwheat:
It's an interesting point. My initial idea was to implement this first just so that, when I get round to implementing native `<dialog>` elements, that this gets carried over to maintain the same API, but if you'd rather put it all in one, that's fine with me.

askvortsov1:
Separate PRs is preferably, but you raise a good point about API: even if the new approach fixes closing if clicking on the modal and then dragging mouse outside, the API would be consistent.

The biggest reason I'm wary about this is that it feels `added  on`. What if instead we had a similar system to composers, where based on the content, the component could either require or not require close confirmation? That feels more light weight and practical.

One more thing: do you know if stacking modals is possible with `detail`? If so, I'd really like to consider having that: it's incredibly annoying when I have a modal, get an error, close debug, and then lose my original modal.

davwheat:
How would we decide what "content" is worthy of requiring close confirmation?

I think that no matter what we should always have the option to override whatever logic is used to decide this.

Native dialogs have support for stacking, backdrops, and more! 

Take a look at the Chrome team's demos: https://googlechrome.github.io/dialog-polyfill/

The only downside is that we need a polyfill for full support in Chrome, but the Chrome team has made one of them anyway.

askvortsov1:
> Native dialogs have support for stacking, backdrops, and more!

Beautiful!

> The only downside is that we need a polyfill for full support in Chrome, but the Chrome team has made one of them anyway. (For some reason Chrome Android supports it natively though...)

If it works, I'm fine with it :grin: Bundle size looks to be ~11kb, and I believe that should pay for itself when we scrap bootstrap?

> How would we decide what "content" is worthy of requiring close confirmation?

Look at how the composer does it: each composer body type can provide a callback that checks whether closing should be prevented: https://github.com/flarum/core/blob/a00c0ac6aa66e218b4f961fcd1d8c58fd6535126/js/src/forum/components/ComposerBody.js#L42-L42

davwheat:
This PR.is deprecated in favour of using HTML5 dialog elements. Tracked in issue #<I can't find this easily on mobile>

Issue #2560, opened 2021-01-22
Restrict `ShowUserController` by ID to users who can `viewUserList


askvortsov1:
## Bug Report

**Current Behavior**
`ShowUserController currently only uses `viewDiscusions` (via `ScopeUserVisibility`) to restrict who can access user profiles (changed from `viewUserList` in #2305). This makes sense when accessing the profile by slug, but we should restrict it to `viewUserList` when accessing by ID to prevent enumeration.

**Environment**
- Flarum version: beta 15

**Possible Solution**
When not accessing by slug, a check for $user->can('viewUserList') should be done.

PR #2573, opened 2021-01-26
Require viewUserList to get user by ID

### LINKED ISSUE #157: Restrict `ShowUserController` by ID to users who can `viewUserList

askvortsov1:
Fixes https://github.com/flarum/core/issues/2560

This makes it more difficult to scape the API for users.

Is there another place where this check would be more appropriate? I'm unsure about putting it in the repository since that method is used by other classes.

### DIFF
--- a/src/Api/Controller/ShowUserController.php
+++ b/src/Api/Controller/ShowUserController.php
@@ -61,6 +61,7 @@ class ShowUserController extends AbstractShowController
         if (Arr::get($request->getQueryParams(), 'bySlug', false)) {
             $user = $this->slugManager->forResource(User::class)->fromSlug($id, $actor);
         } else {
+            $actor->assertCan('viewUserList');
             $user = $this->users->findOrFail($id, $actor);
         }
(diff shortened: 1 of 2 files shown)


clarkwinkelmann (commented):
Couldn't this cause issues with extensions that use GET /api/users/id to refresh a user profile? That's a hypothetical, but just like clicking on homepage refreshes the current user, an extension might generate a request to it when clicking on the same profile you are on to refresh. Would the extension have access to the required information to make the request using the correct slugging?

Actually speaking of current user, doesn't this break the exact feature I describe? Doesn't clicking on homepage or notifications refreshes the user by making a request to their own profile by ID? We would have to allow making a request by ID to the current actor at the very least.

Other thing, can't you obtain the same information by hitting PATCH /api/users/id without any `attributes` ?

askvortsov1:
These are good points. Assuming the correct serializer is always used, do we even want to prevent ID-based enumeration in core?

PR #2576, opened 2021-01-27
Laravel components v8

### LINKED ISSUE #2538: Upgrade Illuminate components to 8.x
Let's upgrade Laravel to 8.x. Just sitting on 6.x will eventually cause Flarum to lag behind, possibly even outside of the LTS scope which is harmful to the product especially now that we close in on stable.

Laravel isn't really moving that fast anymore in terms of changes (yes they do tag more versions, but the actual changes are rather limited compared to diffs between 4 - 5 - 6), so the higher we can bump our dependency the more likely it is that we can keep up and reduce maintenance overhead on extension developers when we do.

Will look into this as soon as I can and once #2507 is merged.

luceos:
**Fixes #2538 **

Still needs to be tested inside the skeleton.

**Confirmed**

**Required changes:**

**Companion PRs**

- [ ] https://github.com/flarum/tags/pull/121


composer.json
```diff
@@ -77,7 +80,6 @@
         "psr/http-server-middleware": "^1.0",
         "s9e/text-formatter": "^2.3.6",
         "symfony/config": "^4.3.4",
-        "symfony/console": "^4.3.4",
```
askvortsov1, line 80: Where is this coming from if it's not explicitly required? We still use the Command class

askvortsov1 (changes requested):
On the Laravel upgrade guide, a dependence on Symfony 5 is listed as one of the high impact changes, but we don't seem to have changed our symfony version at all: https://laravel.com/docs/7.x/upgrade#symfony-5-related-upgrades. Symfony 5 seems to be implicitly required for console though. Are there downsides to having separate versions of symfony for different symfony components?

It seems like the most impactful change would be the deprecation of the Translator interface that's used commonly in extensions.

composer.json
luceos:
```
illuminate/console   v8.25.0  requires  symfony/console (^5.1.4)
illuminate/database  v8.25.0  requires  symfony/console (^5.1.4)
```

luceos:
> 
> 
> On the Laravel upgrade guide, a dependence on Symfony 5 is listed as one of the high impact changes, but we don't seem to have changed our symfony version at all: https://laravel.com/docs/7.x/upgrade#symfony-5-related-upgrades. Symfony 5 seems to be implicitly required for console though. Are there downsides to having separate versions of symfony for different symfony components?
> 
> It seems like the most impactful change would be the deprecation of the Translator interface that's used commonly in extensions.

I still need to go over testing inside the skeleton as mentioned in OP. I bet there are a ton of other issues to fix before this all works ;)

composer.json
askvortsov1: If we're using it explicitly, I'd prefer to have the dependency listed explicitly as well.

askvortsov1:
@luceos I tested this with a skeleton install, and made the above 2 commits, which mitigate most issues with extensions that use the translator.

Aside from that, there are a few issues we won't be able to compensate for:

- Extensions requiring fof/console will break, as that depends on illuminate/console v6, which depends on symfony console v4, which conflicts with core's use of symfony console v5.
- The signature of `setKeysForSaveQuery` on line 65 of TagState in tags needs the typehint removed to comply with Laravel's interface / abstract class.

Other than that, I haven't found issues in the things I tried.

composer.json
```diff
@@ -76,25 +79,26 @@
         "psr/http-server-handler": "^1.0",
         "psr/http-server-middleware": "^1.0",
         "s9e/text-formatter": "^2.3.6",
-        "symfony/config": "^4.3.4",
-        "symfony/console": "^4.3.4",
-        "symfony/event-dispatcher": "^4.3.4",
+        "symfony/config": "^v5.2.2",
+        "symfony/console": "^5.2.2",
```
askvortsov1, line 83: This should probably also have a v for consistency?

composer.json
```diff
@@ -29,8 +32,8 @@
             "email": "matthew@kilgore.dev"
         },
         {
-          "name": "Alexander (Sasha) Skvortsov",
-          "email": "askvortsov@flarum.org"
+            "name": "Alexander (Sasha) Skvortsov",
```
askvortsov1, line 35: These changes should be reverted, since individual names were removed from this file recently.

askvortsov1 (changes requested):
Found one very minor formatting issue in composer.json

composer.json
luceos: Removed all v's because they add no value and no other dep has it.

composer.json
luceos: rebased

askvortsov1 (approved):

askvortsov1 (changes requested):
Looks like there's a few more phpunit related changes needed to the tests before this can be merged

askvortsov1 (approved):

tankerkiller125 (approved):
Now that the test are good I'm happy with this

PR #2577, opened 2021-01-27
Allow overriding routes

### LINKED ISSUE #199: Flarum\Forum\Content\Discussion overriding extension settings
## Bug Report

**Current Behavior**
Whenever I try to extend _content_ behavior using `(new Extend\Frontend('forum'))->content(MyDispatcher::class)`, my callback is added before `Flarum\Forum\Content\Discussion`. So I can't override settings like `$canonicalUrl`, because they always get overwritten by the core component.

**Steps to Reproduce**
1. Create an extension and add a callback to `(new Extend\Frontend('forum'))->content(MyDispatcher::class)`
2. On the `__invoke` function, try to set `$flarumDocument->canonicalUrl` value
3. Discussion pages load with a different canonical url

**Expected Behavior**
It would be very cool if we could prioritize the order of how extensions are executed. But for this specific issue, `MyComponent@__invoke` should be called after `Flarum\Forum\Content\Discussion@__invoke` in order to proper extend the desired functionality.

**Environment**
- Flarum version: Flarum 0.1.0-beta.13
- Webserver: php built-in server
- Hosting environment: localhost
- PHP version: 7.4.3
- Browser: Chrome 84 on Windows 10

### LINKED ISSUE #2239: Flarum\Forum\Content\Discussion overriding extension settings
## Bug Report

**Current Behavior**
Whenever I try to extend _content_ behavior using `(new Extend\Frontend('forum'))->content(MyDispatcher::class)`, my callback is added before `Flarum\Forum\Content\Discussion`. So I can't override settings like `$canonicalUrl`, because they always get overwritten by the core component.

**Steps to Reproduce**
1. Create an extension and add a callback to `(new Extend\Frontend('forum'))->content(MyDispatcher::class)`
2. On the `__invoke` function, try to set `$flarumDocument->canonicalUrl` value
3. Discussion pages load with a different canonical url

**Expected Behavior**
It would be very cool if we could prioritize the order of how extensions are executed. But for this specific issue, `MyComponent@__invoke` should be called after `Flarum\Forum\Content\Discussion@__invoke` in order to proper extend the desired functionality.

**Environment**
- Flarum version: Flarum 0.1.0-beta.13
- Webserver: php built-in server
- Hosting environment: localhost
- PHP version: 7.4.3
- Browser: Chrome 84 on Windows 10

### LINKED ISSUE #2553: Allow overriding routes
## Feature Request

Refs #870 

Once a route is added to FastRoute's RouteCollector, it can't really be overriden or removed. As discussed in https://github.com/nikic/FastRoute/issues/111, that is not something that is planned to be changed. With #2425, a middleware could override the handler for a route, but this is hacky and not immediately intuitive.

Extensions should be able to override or remove routes. While we can't do this on the FastRoute level, we actually already wrap FastRoute's route registration system in our `RouteCollection` class (https://github.com/flarum/core/blob/master/src/Http/RouteCollection.php). In order to allow overriding routes, we could adjust `RouteCollection` to only pass routes into FastRoute once all extensions have booted. `addRoute` would just add routes to an internal data structure, and a new method (`commit()`? `apply()`?) would actually add them to the route collector. In the meantime, if extensions register a new route with the same name as an existing one, the existing one would be overriden. 

Alternatively, we could explicitly require that a route be removed, and then the replacing route be added, to avoid extensions accidentially overriding routes if they copy-paste or otherwise reuse route names.

That being said, the recommended way to insert custom logic into the execution of a controller should be listening to events, and the recommended way to redirect / wrap routes should be middleware. Replacing routes completely has its use cases, but the aforementioned approaches shouldn't be ignored in favor of just swapping the route out completely, which is more difficult for extensions to maintain.

See #2133 for some other ideas

SychO9:
**Fixes flarum/framework#2553**

Store registered routes in an array before applying them to FastRoute, thus allowing extensions to replace them (as described in the issue).

The default route `/` is a special route, it is dynamically set in `ForumServiceProvider` after `flarum.forum.routes` has been resolved, at which point, extensions have already registered their routes, which means it's the one route they cannot override.
That said, I don't think there's a problem with that, they should not be directly overriding this route, instead they should be adding an option for a homepage route like tags.

**TODO**
* [x] Check that this fixes flarum/issue-archive#199

**Confirmed**

### DIFF
--- a/src/Extend/Frontend.php
+++ b/src/Extend/Frontend.php
@@ -31,6 +31,7 @@ class Frontend implements ExtenderInterface
     private $css = [];
     private $js;
     private $routes = [];
+    private $removedRoutes = [];
     private $content = [];
 
     public function __construct(string $frontend)
@@ -55,6 +56,13 @@ class Frontend implements ExtenderInterface
     public function route(string $path, string $name, $content = null)
     {
         $this->routes[] = compact('path', 'name', 'content');
+
+        return $this;
+    }
+
+    public function removeRoute(string $name)
+    {
+        $this->removedRoutes[] = compact('name');
 
         return $this;
     }
@@ -141,7 +149,7 @@ class Frontend implements ExtenderInterface
 
     private function registerRoutes(Container $container)
     {
-        if (empty($this->routes)) {
+        if (empty($this->routes) && empty($this->removedRoutes)) {
             return;
         }
 
@@ -150,6 +158,10 @@ class Frontend implements ExtenderInterface
             function (RouteCollection $collection, Container $container) {
                 /** @var RouteHandlerFactory $factory */
                 $factory = $container->make(RouteHandlerFactory::class);
+
+                foreach ($this->removedRoutes as $route) {
+                    $collection->removeRoute('GET', $route['name']);
+                }
 
                 foreach ($this->routes as $route) {
                     $collection->get(
--- a/src/Http/RouteCollection.php
+++ b/src/Http/RouteCollection.php
@@ -68,7 +68,18 @@ class RouteCollection
 
     public function addRoute($method, $path, $name, $handler)
     {
+        if (isset($this->routes[$method][$name])) {
+            throw new \RuntimeException("Route $name already exists");
+        }
+
         $this->routes[$method][$name] = compact('path', 'handler');
+
+        return $this;
+    }
+
+    public function removeRoute(string $method, string $name): self
+    {
+        unset($this->routes[$method][$name]);
 
         return $this;
     }
--- a/src/Http/RouteCollection.php
+++ b/src/Http/RouteCollection.php
@@ -30,6 +30,11 @@ class RouteCollection
      */
     protected $routeParser;
 
+    /**
+     * @var array
+     */
+    protected $routes = [];
+
     public function __construct()
     {
         $this->dataGenerator = new DataGenerator\GroupCountBased;
@@ -63,19 +68,37 @@ class RouteCollection
 
     public function addRoute($method, $path, $name, $handler)
     {
-        $routeDatas = $this->routeParser->parse($path);
-
-        foreach ($routeDatas as $routeData) {
-            $this->dataGenerator->addRoute($method, $routeData, ['name' => $name, 'handler' => $handler]);
-        }
-
-        $this->reverse[$name] = $routeDatas;
+        $this->routes[$method][$name] = compact('path', 'handler');
 
         return $this;
     }
 
+    protected function applyRoutes(): void
+    {
+        foreach ($this->routes as $method => $routes) {
+            foreach ($routes as $name => $route) {
+                $routeDatas = $this->routeParser->parse($route['path']);
+
+                foreach ($routeDatas as $routeData) {
+                    $this->dataGenerator->addRoute($method, $routeData, ['name' => $name, 'handler' => $route['handler']]);
+                }
+
+                $this->reverse[$name] = $routeDatas;
+            }
+        }
+    }
+
+    public function getRoutes(): array
+    {
+        return $this->routes;
+    }
+
     public function getRouteData()
     {
+        if (empty($this->reverse)) {
+            $this->applyRoutes();
+        }
+
         return $this->dataGenerator->getData();
     }
 
@@ -88,6 +111,10 @@ class RouteCollection
 
     public function getPath($name, array $parameters = [])
     {
+        if (empty($this->reverse)) {
+            $this->applyRoutes();
+        }
+
         if (isset($this->reverse[$name])) {
             $maxMatches = 0;
             $matchingParts = $this->reverse[$name][0];
--- a/src/Http/RouteCollection.php
+++ b/src/Http/RouteCollection.php
@@ -35,6 +35,11 @@ class RouteCollection
      */
     protected $routes = [];
 
+    /**
+     * @var array
+     */
+    protected $pendingRoutes = [];
+
     public function __construct()
     {
         $this->dataGenerator = new DataGenerator\GroupCountBased;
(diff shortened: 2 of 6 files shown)


askvortsov1 (changes requested):
Should we allow directly overriding, or might it be clearer to require that a route be somehow unset? This would prevent accidentially overriding routes by copying route names.

We should also test that this fixes https://github.com/flarum/core/issues/2239

SychO9:
> Should we allow directly overriding, or might it be clearer to require that a route be somehow unset? This would prevent accidentially overriding routes by copying route names.

Yup, that's why it's a draft, I have yet to add the ability to remove a route, which means we'll have to throw an exception if a route is overridden but already exists.

Wasn't sure at first, but I do favour the more explicit way of overriding a route. *(by having to remove it first)*

src/Http/RouteCollection.php
```diff
@@ -63,19 +68,37 @@ public function delete($path, $name, $handler)
...
+    }
+
+    public function getRoutes(): array
+    {
+        return $this->routes;
     }
 
     public function getRouteData()
     {
+        if (empty($this->reverse)) {
```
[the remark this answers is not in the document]
SychO9: so, we would essentially apply the routes on the boot phase by manually resolving the route collection, instead of on first demand. That would technically work, I'm just unsure if it's a good idea to take early resolution into account and explicitly throw an exception just to discourage extension devs.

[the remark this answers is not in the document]
askvortsov1: That's a good point... Plus, now that:

- Classes registered via extenders will be resolved when needed instead of at subscribe time
- Subscribers will be resolved after boot

early resolution should be much, much less frequent. Yay extenders!!!

SychO9:
one thing that's bothering me, is that adding an integration test for not being able to add an already existing route didn't work, there seems to be an issue with expecting the exception.

src/Extend/Frontend.php
```diff
@@ -151,6 +159,10 @@ function (RouteCollection $collection, Container $container) {
                 /** @var RouteHandlerFactory $factory */
                 $factory = $container->make(RouteHandlerFactory::class);
 
+                foreach ($this->removedRoutes as $route) {
+                    $collection->removeRoute('GET', $route['name']);
```
askvortsov1, line 163: Why GET?

src/Extend/Frontend.php
```diff
@@ -59,6 +60,13 @@ public function route(string $path, string $name, $content = null)
         return $this;
     }
 
+    public function removeRoute(string $name)
```
askvortsov1, line 63: I don't like that this is in both extenders, but I don't suppose we have a choice?

src/Http/RouteCollection.php
```diff
@@ -63,19 +68,48 @@ public function delete($path, $name, $handler)
 
     public function addRoute($method, $path, $name, $handler)
     {
-        $routeDatas = $this->routeParser->parse($path);
-
-        foreach ($routeDatas as $routeData) {
-            $this->dataGenerator->addRoute($method, $routeData, ['name' => $name, 'handler' => $handler]);
+        if (isset($this->routes[$method][$name])) {
```
askvortsov1, line 71: Shouldn't we have unique names regardless of method?

askvortsov1 (changes requested):

src/Extend/Frontend.php
SychO9: because the routes are only added by `GET`, look right below this code

askvortsov1: Ah, gotcha

src/Extend/Frontend.php
SychO9: well, Frontend adds this type of content routes, I thought it'd makes sense to be able to remove them here as well, especially since the method is always `GET` here, so it needn't be a parameter.

But technically, using the `Routes` extender to remove these routes will still work I believe.

src/Http/RouteCollection.php
SychO9: While it makes sense for the names to be unique, regardless of the method, and for extension developers to always use unique route names as it is good practice. FastRoute doesn't error out when same route names are used, or even same method and route names, it only errors out when the same path is provided.

Which means, that any extensions that have used similar route names, but with different methods would break if we don't take the method into account.
Here's an example: https://github.com/v17development/flarum-seo/blob/master/extend.php
Search: https://github.com/search?l=PHP&p=3&q=Flarum+Extend+Routes+get+-repo%3Aflarum%2Fapi-docs&type=Code

Although, if there are any extensions that used the same route name and method with a different path, they would break with this implementation as well, I just find it less likely for an extension to have done that, but I haven't it it up (I'm not how I'd look that up either)

askvortsov1: Since the URL Generator is based on unique route names per-app, I would prefer that we enforced unique route names in each app. Any extensions that use the same route name are improperly implemented, and I'd prefer that we break this now than later.

Alternatively, we could do this in phases: method + name unique for this release, just name unique for next release?

SychO9: > Alternatively, we could do this in phases: method + name unique for this release, just name unique for next release?

I would be in favour of this, we would warn extension devs that non unique route names will no longer work in the next release.

tests/integration/extenders/RoutesTest.php
```diff
@@ -47,6 +47,29 @@ public function custom_route_can_be_added_by_extender()
...
+
+    /**
+     * @test
+     */
+    public function custom_route_can_override_existing_route_if_removed()
+    {
+        $this->extend(
+            (new Extend\Routes('api'))
+                ->remove('GET', 'forum.show')
+                ->get('/overridenRoute', 'forum.show', CustomRoute::class)
```
askvortsov1, line 59: Why not have the overriden route at the same path as the old route?

src/Http/RouteCollection.php
```diff
@@ -63,19 +68,48 @@ public function delete($path, $name, $handler)
 
     public function addRoute($method, $path, $name, $handler)
```
askvortsov1, line 69: If routes have already been applied, should the new one be applied too? A few options of doing this:

- Refactor `$this->applyRoutes()` to process routes individually. I don't really like that
- Treat $this->routes as `$this->pendingRoutes`, and clear it at the end of `$this->applyRoutes()`. Then we can just call `$this->applyRoutes()` again.

askvortsov1 (changes requested):

tests/integration/extenders/RoutesTest.php
SychO9: Don't remember, I've tweaked the tests now.

src/Http/RouteCollection.php
SychO9: Second option sounds good

askvortsov1 (approved):

tankerkiller125 (approved):

PR #2578, opened 2021-01-27
Rename ApiSerializer's mutate to attributes

SychO9:
**Part of #1891**

**Confirmed**

**Required changes:**

### DIFF
--- a/src/Api/Serializer/AbstractSerializer.php
+++ b/src/Api/Serializer/AbstractSerializer.php
@@ -43,7 +43,7 @@ abstract class AbstractSerializer extends BaseAbstractSerializer
     /**
      * @var callable[]
      */
-    protected static $mutators = [];
+    protected static $attributeMutators = [];
 
     /**
      * @var array
@@ -87,8 +87,8 @@ abstract class AbstractSerializer extends BaseAbstractSerializer
         $attributes = $this->getDefaultAttributes($model);
 
         foreach (array_reverse(array_merge([static::class], class_parents($this))) as $class) {
-            if (isset(static::$mutators[$class])) {
-                foreach (static::$mutators[$class] as $callback) {
+            if (isset(static::$attributeMutators[$class])) {
+                foreach (static::$attributeMutators[$class] as $callback) {
                     $attributes = array_merge(
                         $attributes,
                         $callback($this, $model, $attributes)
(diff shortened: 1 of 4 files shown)


src/Extend/ApiSerializer.php
```diff
@@ -67,13 +67,35 @@ public function attribute(string $name, $callback)
      *
      * @return self
      */
-    public function mutate($callback)
+    public function attributes($callback)
     {
         $this->mutators[] = $callback;
```
askvortsov1, line 72: Lets rename this property too

askvortsov1 (changes requested):

src/Extend/ApiSerializer.php
```diff
@@ -145,10 +167,10 @@ public function extend(Container $container, Extension $extension = null)
             };
         }
 
-        foreach ($this->mutators as $mutator) {
-            $mutator = ContainerUtil::wrapCallback($mutator, $container);
+        foreach ($this->attributes as $callback) {
+            $callback = ContainerUtil::wrapCallback($callback, $container);
 
-            AbstractSerializer::addMutator($this->serializerClass, $mutator);
+            AbstractSerializer::addMutator($this->serializerClass, $callback);
```
SychO9, line 173: do we want to rename the method as well as the property in `AbstractSerializer` while we're at it as well ?

askvortsov1: Yep! Especially since that's not really public API.

src/Api/Serializer/AbstractSerializer.php
```diff
@@ -43,7 +43,7 @@ abstract class AbstractSerializer extends BaseAbstractSerializer
     /**
      * @var callable[]
      */
-    protected static $mutators = [];
+    protected static $customAttributes = [];
```
askvortsov1, line 46: `customAttributes` doesn't feel plural enough IMO. Maybe something like `attributeMutator`?

askvortsov1 (changes requested):

src/Api/Serializer/AbstractSerializer.php
SychO9: wouldn't it be `attributeMutators` then? since it's many

askvortsov1: Yeah that sounds good!

askvortsov1 (approved):

tankerkiller125 (approved):

src/Extend/ApiSerializer.php
```diff
@@ -67,13 +67,35 @@ public function attribute(string $name, $callback)
      *
      * @return self
      */
-    public function mutate($callback)
+    public function attributes($callback)
```
askvortsov1, line 70: I just realized, we need to deprecate the `mutate` method, since that was released in the previous version

askvortsov1 (changes requested):

src/Extend/ApiSerializer.php
SychO9: it is deprecated, look below this code

askvortsov1: I.... don't know how I missed that... What.... :see_no_evil:

askvortsov1 (approved):

PR #2317, opened 2020-09-26
WIP: Extension Order Resolution

askvortsov1:
WIP, DO NOT REVIEW. This is here so it doesn't get lost.

PR #2579, opened 2021-01-28
Optional Dependencies

askvortsov1:
**Refs https://github.com/flarum/core/issues/1318**

- Add "optional dependencies" as per #1318
- Reintroduce extension dependency resolution from https://github.com/flarum/core/pull/2317
- Sort extensions every time an extension is enabled/disabled

- I have mixed feelings about the name "optional dependencies". Maybe something like "bootBeforeTheseExtensions" would be better? But that's very verbose and would make for a clunky getter method name.

**Confirmed**

**Required changes:**

### DIFF
--- a/src/Extension/Extension.php
+++ b/src/Extension/Extension.php
@@ -218,7 +218,6 @@ class Extension implements Arrayable
     {
         $this->extensionDependencyIds = (new Collection(Arr::get($this->composerJson, 'require', [])))
             ->keys()
-            ->merge(Arr::get($this->composerJson, 'extra.flarum-extension.optional-dependencies', []))
             ->filter(function ($key) use ($extensionSet) {
                 return array_key_exists($key, $extensionSet);
             })
--- a/src/Extension/Extension.php
+++ b/src/Extension/Extension.php
@@ -64,7 +64,7 @@ class Extension implements Arrayable
      * Unique Id of the extension.
      *
      * @info    Identical to the directory in the extensions directory.
-     * @example flarum_suspend
+     * @example flarum-suspend
      *
      * @var string
      */
@@ -90,6 +90,14 @@ class Extension implements Arrayable
      * @var string[]
      */
     protected $extensionDependencyIds;
+
+    /**
+     * The IDs of all Flarum extensions that this extension should be booted after
+     * if enabled.
+     *
+     * @var string[]
+     */
+    protected $optionalDependencyIds;
 
     /**
      * Whether the extension is installed.
@@ -203,16 +211,30 @@ class Extension implements Arrayable
      * @param array $extensionSet: An associative array where keys are the composer package names
      *                             of installed extensions. Used to figure out which dependencies
      *                             are flarum extensions.
+     * @param array $enabledIds:   An associative array where keys are the composer package names
+     *                             of enabled extensions. Used to figure out optional dependencies.
      */
-    public function calculateDependencies($extensionSet)
+    public function calculateDependencies($extensionSet, $enabledIds)
     {
         $this->extensionDependencyIds = (new Collection(Arr::get($this->composerJson, 'require', [])))
             ->keys()
+            ->merge(Arr::get($this->composerJson, 'extra.flarum-extension.optional-dependencies', []))
             ->filter(function ($key) use ($extensionSet) {
                 return array_key_exists($key, $extensionSet);
-            })->map(function ($key) {
+            })
+            ->map(function ($key) {
                 return static::nameToId($key);
-            })->toArray();
+            })
+            ->toArray();
+
+        $this->optionalDependencyIds = (new Collection(Arr::get($this->composerJson, 'extra.flarum-extension.optional-dependencies', [])))
+            ->map(function ($key) {
+                return static::nameToId($key);
+            })
+            ->filter(function ($key) use ($enabledIds) {
+                return array_key_exists($key, $enabledIds);
+            })
+            ->toArray();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/Extension/Extension.php
```diff
@@ -203,16 +211,30 @@ public function setVersion($version)
...
      *                             are flarum extensions.
+     * @param array $enabledIds:   An associative array where keys are the composer package names
+     *                             of enabled extensions. Used to figure out optional dependencies.
      */
-    public function calculateDependencies($extensionSet)
+    public function calculateDependencies($extensionSet, $enabledIds)
     {
         $this->extensionDependencyIds = (new Collection(Arr::get($this->composerJson, 'require', [])))
             ->keys()
+            ->merge(Arr::get($this->composerJson, 'extra.flarum-extension.optional-dependencies', []))
```
KyrneDev, line 221: What's the purpose of including the optional dependencies with the "hard" dependencies? If we already get the optional ones later, then merge the two in the order function, wouldn't it make sense to keep them separate?

askvortsov1: Ah, yeah that was a mistake

src/Extension/Extension.php
```diff
@@ -203,16 +211,30 @@ public function setVersion($version)
...
-            })->map(function ($key) {
+            })
+            ->map(function ($key) {
+                return static::nameToId($key);
+            })
+            ->toArray();
+
+        $this->optionalDependencyIds = (new Collection(Arr::get($this->composerJson, 'extra.flarum-extension.optional-dependencies', [])))
+            ->map(function ($key) {
                 return static::nameToId($key);
```
askvortsov1, line 231: Should optional dependencies be provided as composer package names or Flarum extension IDs? I'm leaning towards switching it to the latter since this is Flarum-specific config

datitisev: Hm... multiple extensions can technically resolve to the same ID, so I feel like we'd want dependencies to use the package name. Specially as that's used for required dependencies (ignoring the fact that uses composer require in composer.json)

askvortsov1: Thing is, the composer package name isn't actually used anywhere in calculations; in the lines directly below here, it turns the composer package names into flarum extension IDs anyway.

Although on the other hand, if we ever decide to switch back to using pure composer package names, having that be the provided format now will reduce breaks then.

SychO9: I would lean towards using package names, just so that we're using one single format when referring to extensions in composer.json (consistency).

askvortsov1: In that case, let's keep as is

src/Extension/ExtensionManager.php
```diff
@@ -86,7 +86,9 @@ public function getExtensions()
             // We calculate and store a set of composer package names for all installed Flarum extensions,
             // so we know what is and isn't a flarum extension in `calculateDependencies`.
             // Using keys of an associative array allows us to do these checks in constant time.
+            // We do the same for enabled extensions, for optional dependencies.
             $installedSet = [];
+            $enabledIds = array_flip($this->getEnabled());
```
SychO9, line 91: This is probably just a nitpick, but leaving it unflipped and then using `in_array` instead of `array_key_exists` seems more simple

SychO9 (commented):
Code looks good!

src/Extension/ExtensionManager.php
askvortsov1: The rationale here was to use constant time key-based lookup instead of linear time value, since multiple extensions will perform multiple lookups.

SychO9: I see, I hadn't noticed or thought about it before, that's better performance wise, and since we're certain that the values are unique than it works just fine, nice.

askvortsov1: Any reason for not explicitly approving the PR?

SychO9: running some tests locally right now

src/Extension/ExtensionManager.php
```diff
@@ -382,4 +392,86 @@ public static function pluckTitles(array $exts)
...
+        $extensionGraph = [];
+        $output = [];
+        $missingDependencies = []; // Extensions are invalid if they are missing dependencies, or have circular dependencies.
+        $circularDependencies = [];
+        $pendingQueue = [];
+        $inDegreeCount = []; // How many extensions are dependent on a given extension?
+
+        foreach ($extensionList as $extension) {
+            $extensionIdMapping[$extension->getId()] = $extension;
+            $extensionGraph[$extension->getId()] = array_merge($extension->getExtensionDependencyIds(), $extension->getOptionalDependencyIds());
```
SychO9, line 422: could it be because we're merging the required extensions array with the optional extensions array ?

SychO9 (changes requested):
So I noticed an odd behaviour,

I created two dummy extensions that have `flarum/tags` as an optional dependency, I enabled both of them and observed the enabled extensions list order, looked fine and logical. I then proceeded to disable the tags extension, being an optional dependency it should not affect the other two extensions, however everytime I tried doing so one of the extensions (not always the same one) gets disabled.

Tracked it down to an issue in the extension order resolution algorithm.

src/Extension/ExtensionManager.php
askvortsov1: Ah, good thing you tested! That's an edge case I didn't think of (and that needs tests). Merging is correct, but we need to filter the optional extensions array to only have installed extensions.

askvortsov1: Added a test, made sure that test failed with the current code, implemented the solution. Test now passes. Thanks for finding this!

SychO9 (approved):

KyrneDev (approved):
Can confirm the bug reported by @SychO9 is fixed, works well locally!

Issue #2539, opened 2021-01-13
Cannot use Tags extension if Flags is uninstalled, inexisting model cannot be visibility scoped


clarkwinkelmann:
## Bug Report

**Current Behavior**
Trying to use the Tags extension while not having Flags installed causes a boot error that prevents access to the forum.

**Steps to Reproduce**
1. Install fresh Flarum, or just enable any extensions including Tags
2. `composer remove flarum/approval flarum/flags` (Approval must be removed as well because it requires Flags)
3. Try to access forum. Observe boot error

```
Flarum encountered a boot error (Exception)
Model Flarum\Flags\Flag cannot be visibility scoped as it does not use Flarum\Database\ScopeVisibilityTrait.
thrown in /home/clark/Projects/flarum-beta15/vendor/flarum/core/src/Extend/ModelVisibility.php on line 48

Exception: Model Flarum\Flags\Flag cannot be visibility scoped as it does not use Flarum\Database\ScopeVisibilityTrait. in /home/clark/Projects/flarum-beta15/vendor/flarum/core/src/Extend/ModelVisibility.php:48
Stack trace:
#0 /home/clark/Projects/flarum-beta15/vendor/flarum/tags/extend.php(86): Flarum\Extend\ModelVisibility->__construct('Flarum\\Flags\\Fl...')
#1 /home/clark/Projects/flarum-beta15/vendor/flarum/core/src/Extension/Extension.php(307): require('/home/clark/Pro...')
#2 /home/clark/Projects/flarum-beta15/vendor/flarum/core/src/Extension/Extension.php(129): Flarum\Extension\Extension->getExtenders()
#3 /home/clark/Projects/flarum-beta15/vendor/flarum/core/src/Extension/ExtensionManager.php(334): Flarum\Extension\Extension->extend(Object(Illuminate\Container\Container))
#4 /home/clark/Projects/flarum-beta15/vendor/flarum/core/src/Extension/ExtensionServiceProvider.php(30): Flarum\Extension\ExtensionManager->extend(Object(Illuminate\Container\Container))
#5 [internal function]: Flarum\Extension\ExtensionServiceProvider->Flarum\Extension\{closure}(Object(Flarum\Foundation\Application))
#6 /home/clark/Projects/flarum-beta15/vendor/flarum/core/src/Foundation/Application.php(322): call_user_func(Object(Closure), Object(Flarum\Foundation\Application))
[...]

PR #2580, opened 2021-01-28
Fixes model visibility

luceos:
Model Visibility extender does not take into consideration missing
dependencies. For instance flarum/tags adds a policy on the Flag model
from flarum/flags. But because flarum/flags might as well not be
installed we need to check for the existence of that model. Otherwise
the exception is thrown or flarum fails to boot.

### DIFF
--- a/src/Extend/ModelVisibility.php
+++ b/src/Extend/ModelVisibility.php
@@ -44,7 +44,7 @@ class ModelVisibility implements ExtenderInterface
     {
         $this->modelClass = $modelClass;
 
-        if (! method_exists($modelClass, 'registerVisibilityScoper')) {
+        if (class_exists($this->modelClass) && ! is_callable([$modelClass, 'registerVisibilityScoper'])) {
             throw new Exception("Model $modelClass cannot be visibility scoped as it does not use Flarum\Database\ScopeVisibilityTrait.");
         }
     }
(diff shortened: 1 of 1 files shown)


askvortsov1:
Issue: https://github.com/flarum/core/issues/2539

Alternate PR: https://github.com/flarum/tags/pull/109

I like this one better though since it's more generic. Maybe there are other issues we need to fix this for?

src/Extend/ModelVisibility.php
```diff
@@ -44,7 +44,7 @@ public function __construct(string $modelClass)
     {
         $this->modelClass = $modelClass;
 
-        if (! method_exists($modelClass, 'registerVisibilityScoper')) {
+        if (class_exists($this->modelClass) && ! is_callable([$modelClass, 'registerVisibilityScoper'])) {
```
askvortsov1, line 47: Should this fail silently if the model doesn't exist? Although I suppose that would be inline with behavior of current extensions.

askvortsov1 (changes requested):
Could we have a test case extending a non-existent class?

src/Extend/ModelVisibility.php
luceos: I think the original idea behind the policy was to apply only when flags was installed. I don't think we need to log anything here.

askvortsov1 (approved):

SychO9 (approved):

PR #2589, opened 2021-02-03
Return promises in onsubmit

askvortsov1:
This allows extensions using the `extend` util to react to the server's response with `then`s. There might be other places we want to do this.

One example of a use case is https://github.com/FriendsOfFlarum/drafts/issues/18

### DIFF
--- a/js/src/forum/components/SignUpModal.js
+++ b/js/src/forum/components/SignUpModal.js
@@ -161,7 +161,7 @@ export default class SignUpModal extends Modal {
 
     const body = this.submitData();
 
-    return app
+    app
       .request({
         url: app.forum.attribute('baseUrl') + '/register',
         method: 'POST',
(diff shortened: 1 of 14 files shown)


js/src/forum/components/SignUpModal.js
```diff
@@ -161,7 +161,7 @@ export default class SignUpModal extends Modal {
 
     const body = this.submitData();
 
-    app
+    return app
```
KyrneDev, line 164: Wouldn’t this be pointless/need more changes since the window is reloaded? Any then statement after wouldn’t be able to run.

KyrneDev:
This really isn’t the solution I would hope for. The reason extensions override the sign up and login modal are because it reloads at the end. It’s great that the mail page can now do this, but it isn’t all that useful compared to these 2.

askvortsov1:
I think I'm going to close this for now, it's not particularly universal as a solution.

PR #2593, opened 2021-02-04
AdminUX Patch and Admin Page

### LINKED ISSUE #2498: Refactor Admin Dashboard (next steps)
It would be very nice if we could move the `buildSettingComponent` logic (and other underlying logic) from ExtensionPage to a new base AdminPage, and try to standardize how each admin page works as much as possible: that would make it easier to maintain and add new admin pages.

Also, `buildSettingComponent` currently doesn't support placeholder, min, max, etc. Perhaps all non-reserved keys in the setting object should be applied as attrs to the input?

### LINKED ISSUE #2503: [Tags] Extension permission grids can be confusing
One can see all the tag-specific permissions in an extension's permission grid even if there are no tag-specific permissions specific to that extension.

The problem with this is that the tag-specific feature is global and not per-extension. This is made confusing by the fact that you have an X to delete tag-specific permissions for a specific tag

_Permission grid of fof/drafts on Flarum Discuss_
image

Users may think this is specific to the extension they're currently in the page of and may delete ALL their tag-specific perms for that tag without meaning to.

### LINKED ISSUE #2504: Admin UX - cannot scroll horizontally on permission grids in extension pages
Mainly an issue because of tag-specific permissions with flarum/tags.

### LINKED ISSUE #2520: [Accessibility] Enabled/Disabled extension dot indicators not accessibility-friendly
Apart from the color difference (red/green) the shapes/forms of the dots indicating if an extension is enabled or disabled should also be different enough to be able to tell without the colors. And they should also have tooltips.

**References:**
https://discuss.flarum.org/d/7585-free-flarum-hosting-on-an-expert-platform-by-freeflarum-com/1878
https://discuss.flarum.org/d/25844-old-style-admin-panel/21

Some propositions:
. | .
-- | --
flarum lan_admin (17) | flarum lan_admin (15)

KyrneDev:
**Fixes #2498**
**Fixes #2503**
**Fixes #2504**
**Fixes #2520**

* Add an AdminPage component that all admin pages extend. It contains the `buildSettingsComponent` helper and a few other helpful save functions.
* `buildSettingsComponent` now assigns attrs to the components it is building (min/max/classname etc). Also allows built in helptext.
* Add padding to the bottom of extension page for permissions
* Remove `AdminLinkButton` (breaking/deprecated), `AddExtensionModal`, and `app.extensionSettings`. Drop support for `SettingsModals` for extensions.
* Prevent scope changes and allow horizontal scrolling from the extension permissions grid.
* Split `ExtensionWidget` functions of JSX.
* LESS despecifications.

AdminPage, specifically the info section. Is this how we want to do it or do you have any better ideas?

**Confirmed**

**Required changes:**

- [ ] Related documentation PR: Will put docs PR when I make it
- [x] Related core extension PRs: Tags: https://github.com/flarum/tags/pull/116 and English: https://github.com/flarum/lang-english/pull/183


datitisev (commented):
This PR is a bit hard, there's many changes that address different things.
I do have a few things to point out and/or ask about.

I think we want methods for the options in `info()` instead of returning an object.

Why were `fields` removed in `BasicsPage` and `MailPage`? I don't see a reason given in the PR description.

js/src/admin/components/BasicsPage.js
```diff
@@ -40,157 +20,109 @@ export default class BasicsPage extends Page {
...
+              }),
+              this.buildSettingComponent({
+                type: 'switch',
+                setting: 'show_language_selector',
+                label: app.translator.trans('core.admin.basics.show_language_selector_label'),
+              }),
+            ]
+          : ''}
+
+        {FieldSet.component(
```
askvortsov1, line 71: Maybe this one should be JSX?

askvortsov1 (changes requested):
Overall makes sense, left some code nitpicks. Will test locally later.

Going forward, it would be preferable to split big changes like this up into commits (or even better, separate PRs when possible) to make review easier.

KyrneDev:
> I think we want methods for the options in `info()` instead of returning an object.

Like `icon()` and `description()`?

> Why were `fields` removed in `BasicsPage` and `MailPage`? I don't see a reason given in the PR description.

There's no need for them as there is a settings helper already apart of AdminPage now.

datitisev:
> Like icon() and description()?

Yes, that's what other components do. This was just a thought by the way, but I feel like it'd be more intuitive than returning an object. If we wanted a simple `info()` method we'd probably create a separate class that encapsulates it, I think (or maybe I'm not making sense).

js/src/admin/components/BasicsPage.js
```diff
@@ -40,157 +20,109 @@ export default class BasicsPage extends Page {
...
+
+        <div className="BasicsPage-welcomeBanner-input">
+          {this.buildSettingComponent({
+            type: 'text',
+            setting: 'welcome_title',
+            label: app.translator.trans('core.admin.basics.welcome_banner_heading'),
+            className: 'BasicsPage-welcomeBanner',
+            help: app.translator.trans('core.admin.basics.welcome_banner_text'),
+          })}
+          <textarea className="FormControl" bidi={this.setting('welcome_message')} />
```
SychO9, line 97: I would prefer to directly use JSX for this one since it's a custom setting. The advantage to directly using JSX is that the HTML structure remains consistent in that `.FormGroup` will always be the direct child of `.Form` and will contain the setting inputs, whereas currently it doesn't make much sense.

```jsx
<div className="Form-group">
  <label>{app.translator.trans('core.admin.basics.welcome_banner_heading')}</label>
  <div className="helpText">{app.translator.trans('core.admin.basics.welcome_banner_text')}</div>
  <input type="text" bidi={this.setting('welcome_title')} />
  <textarea className="FormControl" bidi={this.setting('welcome_message')} />
</div>
```

SychO9 (changes requested):
I like this, it makes the interface so much more customizable!

KyrneDev:
@datitisev I'm mostly in favor of keeping it inside one function and return an object. I think having functions for all four info items is a bit redundant and although there will have to be a certain level of redundancy this takes up the smallest footprint. It's just as easy to extend as well.

Let's keep discussing! For now, I'm going to just change it to `headerInfo`

less/admin/ExtensionPage.less
```diff
@@ -142,11 +142,13 @@
...
-        overflow-x: scroll;
-        padding-bottom: 20px;
-      }
+    .PermissionGrid-removeScope {
+      display: none;
+    }
+
+    > .container {
+      overflow-x: scroll;
+      padding-bottom: 20px;
```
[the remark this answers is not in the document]
KyrneDev: I solved it by selecting `PermissionGrid` instead of `container`. Let me know if this works for you.

js/src/admin/components/BasicsPage.js
KyrneDev: Done, also did the same in the mail page.

js/src/admin/components/BasicsPage.js
```diff
@@ -68,23 +68,17 @@ export default class BasicsPage extends AdminPage {
...
-                  <input type="radio" name="homePage" value={path} bidi={this.setting('default_route')} />
-                  {label}
-                </label>
-              )),
-          ]
-        )}
+          <FieldSet className='BasicsPage-homePage Form-group' label={app.translator.trans('core.admin.basics.home_page_heading')}>
+            <div className="helpText">{app.translator.trans('core.admin.basics.home_page_text')}</div>
+            {this.homePageItems()
+            .toArray()
```
askvortsov1, line 74: Shouldn't this be indented?

askvortsov1 (approved):

less/admin/ExtensionPage.less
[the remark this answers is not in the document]
SychO9: It's still the same unfortunately 😅, essentially anywhere you scroll the horizontal overflowing, it will automatically also affect the vertical one, from local tests, I believe what I proposed above might be the best way to approach this.

flarum lan_admin (20)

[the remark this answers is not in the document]
KyrneDev: Check now

askvortsov1 (commented):
Could any of the CSS common across pages be factored out into an `AdminPage.less` file?

SychO9 (changes requested):
@askvortsov1 there are some pages like Basics, Appearence and Email, that have similar bottom padding and container CSS code, but it's also different in Dashboard, ExtensionPage and PermissionsPage, so I don't think we could

less/admin/PermissionsPage.less
```diff
@@ -3,13 +3,8 @@
   border-radius: @border-radius;
   max-width: calc(~'100% - 60px');
```
SychO9, line 4: we need to drop this now that we have the common AdminPage container.

also there is a `padding-left: 0 12px;` somewhere in this file, could you remove that please ?

SychO9 (changes requested):

less/admin/PermissionsPage.less
```diff
@@ -1,15 +1,9 @@
...
   border-radius: @border-radius;
-  max-width: calc(~'100% - 60px');
   display: block;
-  margin-left: 30px;
   overflow-x: auto;
   padding: 10px 0 8px;
-
-  > .container {
-    padding: 0 10px;
-  }
```
SychO9, line 12: sorry, another thing I noticed is since we removed the container here we need to fix the parent element's paddings, `padding: 10px;` would work just fine

SychO9 (approved):
Looks good to me!

Issue #2566, opened 2021-01-25
Editor Driver Support


askvortsov1:
A WYSIWYG editor is probably among the most requested Flarum features of all time (https://discuss.flarum.org/d/1537-pick-three-features-you-d-love-to-see-in-flarum), with numerous extension attempts (https://discuss.flarum.org/t/extensions?q=editor). However, this has been challenging for several reasons:

- Any editor must be compatible with s9e/textformatter (so no Quill or HTML-based, must output plain markdown)
- A good editor should look modern and have small bundle size (so no tinyMCE, ckEditor, or anything of that sort)
- We should be able to add support for custom elements/customizations
- A succesful editor would need to integrate with mentions and emoji's autocomplete dropdowns
- Inserting an editor into the existing textarea tends to class with mdarea, used in Markdown
- Some users might prefer a raw markdown experience
- Use of `execCommand` is a no-go (https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand)

So I set out to experiment and explore potential improvements that would enable smooth integration of a WYSIWYG editor with Flarum. For this experiment I used [proseMirror](https://prosemirror.net/), a library for building editors. Other options:
- simpleMDE, which is based on codeMirror, which is proseMirror's sister product
- https://github.com/benweet/stackedit
- slate

### Try 1

I attempted to do what most rich text extensions so far have done, and extend/override various methods of `TextEditor`. This works on a basic level. https://github.com/askvortsov1/flarum-rich-text

Observations:
- Switching between mdarea and a replacement is messy while the editor is open, simpler to go with one or the other, and control which one in user settings
- Mentions and emoji autocomplete dropdowns work by wrapping `TextEditor`'s `textarea` html element in a div wrapper, then manipulating that. We can't assume that `textarea` will exist
- On a similar note, direct interactions by extensions with whichever editor is used should be avoided at all costs: we should expand `SuperTextarea` to act as a full abstraction for the text editor

### Try 2

I tried going in a significantly different direction, and including a basic proseMirror setup directly in core. I added styles and customizations so that it exactly resembled the html textarea-based solution. Then, `flarum/markdown` could extend that basic proseMirror config to add markdown support. I was also hoping to get mentions working with this solution, but wasn't able to get that done. https://github.com/flarum/core/pull/2567, https://github.com/flarum/markdown/pull/23

Observations:
- This caused minified forum size to skyrocket from 350kb to 550kb. Ouch...
[...]

PR #2594, opened 2021-02-05
Editor Drivers Abstraction

askvortsov1:
This will allow drop-in replacements of the editor with a more advanced WYSIWYG solution such as ProseMirror

See #2566


js/src/common/compat.js
```diff
@@ -92,7 +91,6 @@ export default {
   'utils/stringToColor': stringToColor,
   'utils/Stream': Stream,
   'utils/subclassOf': subclassOf,
-  'utils/SuperTextarea': SuperTextarea,
```
askvortsov1, line 95: Should we provide BasicEditorDriver as a (deprecated) alias for this? All the logic from there is still there, but we'd need to include something from forum in common. Although in this case I don't think it's too bad since BasicEditorDriver doesn't import anything from the rest of forum.

js/src/forum/components/TextEditor.js
```diff
@@ -22,25 +23,22 @@ export default class TextEditor extends Component {
...
+     * The value of the editor.
      *
      * @type {String}
      */
     this.value = this.attrs.value || '';
+
+    /**
+     * Whether the editor is disabled.
+     */
+    this.disabled = !!this.attrs.disabled || true;
```
datitisev, line 35: What is this? Won't `this.disabled` always equal true here? If `this.attrs.disabled` is falsy, you will still be setting it to true.

js/src/forum/components/Composer.js
```diff
@@ -157,7 +157,7 @@ export default class Composer extends Component {
    * Draw focus to the first focusable content element (the text editor).
    */
   focus() {
-    this.$('.Composer-content :input:enabled:visible:first').focus();
+    this.state.editor.focus();
```
datitisev, line 160: This changes the behavior from focusing on the first available field (e.g. title) to always focusing on the post content.

datitisev (commented):
Not looked very deep into this, just some stuff that stood out initially.

js/src/common/compat.js
SychO9: Since the logic is the same, and there's always a chance an extension could've used it, I don't see why not.

Interesting that it was in common though since it was only used in forum.

SychO9 (commented):
Code looks good. Tested locally.

There's one reference to `SuperTextArea` left in `ComposerState` though.

js/src/common/compat.js
askvortsov1: This actually brings up an interesting point: it was in common, not forum, because it didn't depend on any stuff in forum. And that might be something we want to keep: TextEditor might be eventually useful for editing stuff on the admin side that uses formatting (e.g. if tags descriptions adopt formatting).

For now why don't we just add it to forum, but we can keep an eye out for this in the future?

SychO9: Agreed.

SychO9 (approved):

SychO9 (changes requested):
emoji and mentions aren't working properly, the dropdown isn't showing at the proper position we're typing in, and it errors out here with, I haven't investigated the issue

```
addComposerAutocomplete.js:148 Uncaught TypeError: Cannot read property 'top' of undefined
    at addComposerAutocomplete.js:148
    at addComposerAutocomplete.js:154
    at BasicEditorDriver.ts:22
    at Array.forEach (<anonymous>)
```

and 

```
Uncaught TypeError: Cannot read property 'scrollTop' of undefined
    at g (addComposerAutocomplete.js:163)
    at addComposerAutocomplete.js:179
    at BasicEditorDriver.ts:22
    at Array.forEach (<anonymous>)
    at HTMLTextAreaElement.i (BasicEditorDriver.ts:21)
```

askvortsov1:
You're on the editor abstraction branches of both those extensions, right? Also, could you share the test post that caused it to break?

SychO9:
> You're on the editor abstraction branches of both those extensions, right? Also, could you share the test post that caused it to break?

Yea, I'm on all the relevant dev branches, as for the test post, just triggering the dropdowns with either `@` or `:` or directly clicking the buttons triggers the errors.

askvortsov1:
> > You're on the editor abstraction branches of both those extensions, right? Also, could you share the test post that caused it to break?
> 
> Yea, I'm on all the relevant dev branches, as for the test post, just triggering the dropdowns with either `@` or `:` or directly clicking the buttons triggers the errors.

Ah.... When I converted it to use input listeners, we lost the ability to get the event target value, which broke this. I'll change the caret position method to return position relative to viewport, since the editor driver has access to the dom/underlying mechanism (and so can figure this out), whereas the input listeners really have no clue, and so shouldn't be responsible for locating it. This will DRY the dropdown positioning code between emoji and mentions.

On that topic, the input listeners part of this is cool, but it's probably the weakest part of the new system in terms of design. Ideally it'd be nice to pass `e`, but that's not really possible because alternative implementations might not have an `e` (although currently my wysiwyg approach implements input listeners by just listening to the vanilla js/dom events. Constructing a new abstraction for an "input event" is challenging and needs a fair bit more research. The one thing I think we could do so far is provide the type of event (click, input, keypress), but even that I'm not 100% sure about. I left this out of the initial implementation because we could always add arguments later, but if there's strong sentiment towards having them now, we can try and do so.

askvortsov1:
Actually never mind, I am no longer convinced fully DRYing away the mentions/emoji dropdown positioning code is a good idea, as it depends on the dropdown, and I don't think that should be an argument.

SychO9 (approved):

KyrneDev (approved):
Played around with this for a little while from the perspective of an extension dev and I really like it. It's super easy to work with now!

PR #2596, opened 2021-02-07
Support global functions in ContainerUtil

askvortsov1:
It can be very annoying if we want to use something like `boolval`, but have to define an entire anonymous function to pass it in. This PR adds support for tpassing it in directly as a string, like is posible with User::registerPreference.

### DIFF
--- a/src/Foundation/ContainerUtil.php
+++ b/src/Foundation/ContainerUtil.php
@@ -23,7 +23,7 @@ class ContainerUtil
      */
     public static function wrapCallback($callback, Container $container)
     {
-        if (is_string($callback)) {
+        if (is_string($callback) && !function_exists($callback)) {
             $callback = function (&...$args) use ($container, $callback) {
                 $callback = $container->make($callback);
 
--- a/src/Foundation/ContainerUtil.php
+++ b/src/Foundation/ContainerUtil.php
@@ -18,7 +18,7 @@ class ContainerUtil
      *
      * @internal Backwards compatability not guaranteed.
      *
-     * @param callable|string $callback: A callable, or a ::class attribute of an invokable class
+     * @param callable|string $callback: A callable, global function, or a ::class attribute of an invokable class
      * @param Container $container
      */
     public static function wrapCallback($callback, Container $container)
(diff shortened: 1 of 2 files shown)


src/Foundation/ContainerUtil.php
```diff
@@ -18,12 +18,12 @@ class ContainerUtil
...
      * @internal Backwards compatability not guaranteed.
      *
-     * @param callable|string $callback: A callable, or a ::class attribute of an invokable class
+     * @param callable|string $callback: A callable, global function, or a ::class attribute of an invokable class
      * @param Container $container
      */
     public static function wrapCallback($callback, Container $container)
     {
-        if (is_string($callback)) {
+        if (is_string($callback) && ! function_exists($callback)) {
```
SychO9, line 26: perhaps we should use `! is_callable($callback)` instead, at least then we are certain that the string is neither a function nor a static method ?

askvortsov1: Would we want to support static methods being passed in like this? This opens up an interesting question, as we could also theoretically support calling methods on non-static classes (https://stackoverflow.com/a/3499756/9716252)

SychO9: well, technically, we allow `string` or `callable` to be passed right ? so it only makes sense to check with `! is_callable($callback)` since even those types of arrays pass the callable type (from local testing).

So, yeah, it makes sense to support callables in general, since that's a type we allow passing.

askvortsov1: Sorry, I think I worded that poorly. In the scenario that `$callable` is an array, where the first argument is the name of a class, and the second argument is the name of a non-static method, should we construct the class indicated by the first argument?

SychO9: Oh I see what you mean. Hmm, not sure to be honest, I guess that could be very useful instead of having multiple invokable classes, and I'm assuming it should be fairly simple code-wise. I don't see why not, but it might be better to create a new issue I think and tackle it later on.

SychO9 (approved):

tankerkiller125 (approved):

PR #2602, opened 2021-02-16
Don't push bidi function to DOM

KyrneDev:
As of now, the bidi function is viewable in the DOM via inspect element. Although this doesn't cause any issues it doesn't look super clean and isn't necessary. This removes the attr before it is pushed into the view and after it sets the `onchange` and `value` attrs.

**Confirmed**


askvortsov1 (approved):

askvortsov1:
Confirmed to work locally

SychO9 (approved):

PR #2606, opened 2021-02-17
WIP Floating Discussion hero

### LINKED ISSUE #157: I quickly forget which topic I am reading

### LINKED ISSUE #433: Floating discussion title

KyrneDev:
**Closes flarum/issue-archive#433**

Float the discussion header on scroll so it's visible farther down. I am also going to submit a PR to tags to hide secondary tags from the floated discussion as too many tags makes it very cluttered and cramped.

Please check the CSS. There may be better ways of doing what I am doing.

**Screenshot**
Screen Shot 2021-02-16 at 5 30 14 PM

**Confirmed**

**Required changes:**

- [x] Related core extension PRs: Tags: https://github.com/flarum/tags/pull/119

### DIFF
--- a/js/src/forum/components/DiscussionPage.js
+++ b/js/src/forum/components/DiscussionPage.js
@@ -58,10 +58,13 @@ export default class DiscussionPage extends Page {
     const scrollListener = new ScrollListener((top) => {
       const $hero = $('.DiscussionHero');
       if ($hero.offset()) {
-        $hero.toggleClass('DiscussionHero--floating', top > 92);
+        const container = $('.DiscussionHero').children('.container');
+        const containerPadding = parseInt(container.css('padding-top')) + parseInt(container.css('padding-bottom'));
+
+        $hero.toggleClass('DiscussionHero--floating', top > 22 + containerPadding);
         $('.DiscussionPage-discussion')
           .children('.container')
-          .toggleClass('scrolled', top > 92);
+          .toggleClass('scrolled', top > 22 + containerPadding);
       }
     });
 
--- a/js/src/forum/components/DiscussionPage.js
+++ b/js/src/forum/components/DiscussionPage.js
@@ -9,6 +9,7 @@ import SplitDropdown from '../../common/components/SplitDropdown';
 import listItems from '../../common/helpers/listItems';
 import DiscussionControls from '../utils/DiscussionControls';
 import PostStreamState from '../states/PostStreamState';
+import ScrollListener from "../../common/utils/ScrollListener";
 
 /**
  * The `DiscussionPage` component displays a whole discussion page, including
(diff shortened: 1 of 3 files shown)


ornanovitch:
> I am also going to submit a PR to tags to hide secondary tags from the floated discussion as too many tags makes it very cluttered and cramped.

Hey Kyrne, I was thinking about this: do you think tags >=3 could be partially hidden with a `transition` and a `mask`, like in `DiscussionListItem` ? Hiding them could lead some users to misunderstand what's going on.

Another alternative : hide them but put a `(+[number of hidden tags])` just to the right of the last displayed tag. E.g. `| Support | Upgrading | (+1)`

In any case, you'r PR could represent a huge step forward regarding DiscussionPage UX :partying_face:

askvortsov1 (commented):
A few UX thoughts:

- The badges feel a little bit squashed together, could we add some margin between them?
- The transition from full hero to floating hero feels a bit abrupt. Is there any way we could smooth it out with animation / trigger it a bit earlier?
- https://i.imgur.com/RwpGAcS.png Can we disable shadow from the app header while this is open? The double shadow feels like a bit much, and I prefer Toby's original layout: https://github.com/flarum/core/issues/157#issuecomment-116172854. 
- I like the idea of https://github.com/flarum/core/issues/157#issuecomment-261511600, I think that might be less intrusive

SychO9 (commented):
imho, if we do decide to go ahead with this, we better show it when we scroll up and hide it when we scroll down, to not always take up space on the screen (not a fan of fixing multiple elements on the screen)

Ref: https://github.com/flarum/core/issues/157#issuecomment-550339023 (but not just for mobile)

kecrily:
图片
Hey, I have a idea. Why don't we use the blank area located on the right side?
But I don't know how we can implement the display in the mobile version.

davwheat:
@scvoet It's an interesting thought, but the narrowness would result in a lot of text wrapping.

```
This is a
long
discussion
title so is
shown
across
multiple
lines
```

is quite hard to read quickly.

tolgaaaltas:
> * The badges feel a little bit squashed together, could we add some margin between them?

I feel it too. Maybe the team should try to put the badges over the labels. Their height and width are 30px. 

image

davwheat:
That looks a bit weird to me -- we don't do this anywhere else (we put them on top of the avatar, but they remain full size), and it causes mismatches in styling.

I'm for adding 2px margin between the badges instead, personally.

tolgaaaltas:
> we don't do this anywhere else (we put them on top of the avatar, but they remain full size), 

Yes, you're right. But when you use these element in full size, floating hero looks too much full. This is looking more fresh imo. Of course, the decision is yours.

davwheat:
> Of course, the decision is yours.

Everyone can input, and my opinion is definitely not the deciding factor, and for all I know the core team may end up going for it in the end. :)

ekumanov:
@SychO9 Why was this closed? It seemed like a nice dev to me.

SychO9:
@ekumanov it's an old PR which was not finished and is no longer up to date. The author wasn't available to finish it.

PR #2607, opened 2021-02-17
Update DiscussionListItem.less to fix double tap on mobile

### LINKED ISSUE #1913: Touch devices of certain widths forced to double tap to open discussion

zerosonesfun:
Fix double tap to open discussion on mobile.

**Fixes #1913**

Adds a rule to the discussion list less file which targets touch devices whose primary way of interacting does not include a mouse / ability to hover. For those devices the toggle button is hidden which fixes the double tap issue. 

Test on small screens. You should no longer have to double tap to open a discussion.

**Confirmed**

### DIFF
--- a/less/forum/DiscussionListItem.less
+++ b/less/forum/DiscussionListItem.less
@@ -110,6 +110,11 @@
   }
 }
 
+@media (any-hover: none) { 
+  .DiscussionListItem-controls > .Dropdown-toggle { 
+    display: none;
+  } 
+}
 
 @media @phone {
   .DiscussionListItem-controls {
@@ -265,10 +270,4 @@
       }
     }
   }
-}
-
-@media (any-hover: none) { 
-  .DiscussionListItem-controls > .Dropdown-toggle { 
-    display: none;
-  } 
 }
--- a/less/forum/DiscussionListItem.less
+++ b/less/forum/DiscussionListItem.less
@@ -266,3 +266,9 @@
     }
   }
 }
+
+@media (any-hover: none) { 
+  .DiscussionListItem-controls > .Dropdown-toggle { 
+    display: none;
+  } 
+}


less/forum/DiscussionListItem.less
```diff
@@ -266,3 +266,9 @@
     }
   }
 }
+
+@media (any-hover: none) {
```
askvortsov1, line 270: Could we move this back up to lines 113?

askvortsov1 (changes requested):

askvortsov1 (approved):

SychO9 (approved):

Issue #1796, opened 2019-06-16
Review/deprecate global helpers functions


luceos:
We need to prevent the use of helpers. The core still has a helpers set at `src\helpers.php` we should consider which of these should remain or whether we should deprecate them all.

clarkwinkelmann:
Just to be clear, this issue is about PHP helpers ? Not javascript ?

franzliedke:
Yep, only about PHP.

askvortsov1:
I would say the event one should go, since we want events as private API (if at all). I'd also say app gives extensions waaayyyyy too much access into the container, and should be removed. I think the path ones are alright (unless there's a better place to get that information, in which case I'd say we remove them too?

datitisev:
@askvortsov1 Extensions must sometimes be able to fire their own events & Flarum events to work properly though

Also - this isn't the place to discuss private vs public API, though I think in the past we've said that private API would be discouraged to use but still available for very specific things an extension may not be able to do with the public one (public API will never account for all the functionality extensions will need).

luceos:
If we keep the app Helper that should open up our code sufficiently for developers to hook into the deeper workings. Everything else, could or should, potentially be droppable.

clarkwinkelmann:
I'd say it's all a question of convenience. Just like in Laravel, most of the stuff is just there to make the developer life easier, and sometimes make the core more readable.

I'm actually using every helper function we have left in one extension or another. I know I could also use `app()` and go from there, but those functions are quicker to write and more readable.

The `_path` methods seems like the less useful, because it doesn't add much to the `app()->somethingPath()` that already exists.

`app()` and `event()` I would keep. It's also probably more convenient to users that come from Laravel and expect those methods to exist.

askvortsov1:
Couldnt app be injected into any part of any extension? At least thats the approach I've been taking.

clarkwinkelmann:
Constructor injection is not always the best option. You might want to resolve a class as late as possible, either for performance or simply to let other extensions act before yours.

For example, I only resolve the SettingsRepositoryInterface after I checked the user is allowed to see the settings. Or I only resolve the logger once an exception has been triggered. The `app()` function is extremely useful for that.

I don't see any disadvantage of keeping those functions. As long as we keep a single one, the `helpers.php` file will be loaded anyway, so removing half of the functions will not reduce the number of files loaded by the autoload either.

franzliedke:
[...]

PR #2609, opened 2021-02-17
Rename `app` to container

askvortsov1:
**Refs #1796 **

In #2142, the container was detached from the application in preparation for the upcoming Laravel upgrade. However, the terminology of `app` is still used everywhere which is fairly misleading. This PR renamed containers called $app to $container in core, and deprecates `AbstractServiceProvider::container` and the `app` helper in favor of `container`.

### DIFF
--- a/src/Admin/AdminServiceProvider.php
+++ b/src/Admin/AdminServiceProvider.php
@@ -36,18 +36,18 @@ class AdminServiceProvider extends AbstractServiceProvider
      */
     public function register()
     {
-        $this->app->extend(UrlGenerator::class, function (UrlGenerator $url) {
-            return $url->addCollection('admin', $this->app->make('flarum.admin.routes'), 'admin');
+        $this->container->extend(UrlGenerator::class, function (UrlGenerator $url) {
+            return $url->addCollection('admin', $this->container->make('flarum.admin.routes'), 'admin');
         });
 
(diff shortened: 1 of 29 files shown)


tankerkiller125 (approved):
I think this looks good, I haven't tested locally yet.

clarkwinkelmann:
I'm a bit worried about the change to `AbstractServiceProvider` :grimacing: That's a lot of diff

What happens if we use a service provider that extends Illuminate's service provider instead of Flarum's service provider? Does `app` still works on those?

askvortsov1:
> What happens if we use a service provider that extends Illuminate's service provider instead of Flarum's service provider? Does app still works on those?

What do you mean? The change is to Flarum's AbstractServiceProvider, and all classes extending from it. If an extension is using Laravel's provider, it shouldn't be affected: we haven't changed the type of argument being passed into the ServiceProvider in this PR.

clarkwinkelmann (commented):
Ok then that looks fine.

I guess my other PR will require changes once this is merged :sweat_smile: 

I see there are still some `$app` variables lying around in anonymous functions. Should those be renamed to `$container` as well?

clarkwinkelmann (approved):
If tests are green and it works on your setup, I don't think it's worth I try locally. If anything is broken it'll probably be some very unusual use case anyway :sweat_smile: 

All good for me!

PR #2611, opened 2021-02-18
Add Back Button to Post Scrubber

KyrneDev:
Adding a Back button to the post stream scrubber when moving around. This makes it super convenient to jump back and forth when checking out different locations in the discussion - Discourse has something similar. 

**Screenshot**
Screen Shot 2021-02-17 at 12 50 25 PM

**Confirmed**

**Required changes:**

- [ ] Related English PR: Will do this when/if merged.

### DIFF
--- a/js/src/forum/components/PostStreamScrubber.js
+++ b/js/src/forum/components/PostStreamScrubber.js
@@ -31,8 +31,15 @@ export default class PostStreamScrubber extends Component {
       count: <span className="Scrubber-count">{formatNumber(count)}</span>,
     });
 
+    const index = this.stream.index;
+
+    const showBackButton =
+      this.lastPosition > count / 20 &&
+      this.lastPosition < count - count / 20 &&
+      100 * Math.abs((index - this.lastPosition) / count) > 25;
+
     const unreadCount = this.stream.discussion.unreadCount();
-    const unreadPercent = count ? Math.min(count - this.stream.index, unreadCount) / count : 0;
+    const unreadPercent = count ? Math.min(count - index, unreadCount) / count : 0;
 
     function styleUnread(vnode) {
       const $element = $(vnode.dom);
@@ -65,6 +72,7 @@ export default class PostStreamScrubber extends Component {
             </a>
 
             <div className="Scrubber-scrollbar">
+              {showBackButton ? this.backButton() : ''}
               <div className="Scrubber-before" />
               <div className="Scrubber-handle">
                 <div className="Scrubber-bar" />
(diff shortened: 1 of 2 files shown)


js/src/forum/components/PostStreamScrubber.js
```diff
@@ -199,6 +226,7 @@ export default class PostStreamScrubber extends Component {
    * Go to the first post in the discussion.
    */
   goToFirst() {
+    this.lastPosition = this.stream.index;
```
askvortsov1, line 229: Why not have these on `PostStreamState` under `goToNumber` and `goToIndex`?

askvortsov1 (changes requested):
Should the "back" button be symmetrical? A few potential ideas that could be good or... not good:

- Some form of tapered edges/arrow on the left hand side pointing to the scrubber / the bar
- A back icon?

Might be a bit much though.

From a review perspective, I feel like some of this would be better as part of `PostStreamState`.

KyrneDev:
Now uses `PostStreamState` to store most of the data, accessible from extensions.

Also converted the button to a `chevron-left` icon, much more minimal and sleek:

Screen Shot 2021-02-19 at 1 06 52 PM

js/src/forum/components/PostStreamScrubber.js
```diff
@@ -31,8 +34,13 @@ export default class PostStreamScrubber extends Component {
       count: <span className="Scrubber-count">{formatNumber(count)}</span>,
     });
 
+    const index = this.stream.index;
+    const previousIndex = this.stream.previousIndex;
+
+    const showBackButton = previousIndex > count / 20 && previousIndex < count - count / 20 && 100 * Math.abs((index - previousIndex) / count) > 25;
```
matteocontrini, line 40: Maybe a comment about what's going on in this condition could be added? It doesn't seem that straightforward to understand 😅

js/src/forum/components/PostStreamScrubber.js
```diff
@@ -31,8 +34,17 @@ export default class PostStreamScrubber extends Component {
       count: <span className="Scrubber-count">{formatNumber(count)}</span>,
     });
 
+    const index = this.stream.index;
+    const previousIndex = this.stream.previousIndex;
+
+    // We want to make sure the back button isn't crammed in.
+    // If the previous post index is less than 5% from the last/first post,
+    // or if the previous post index is less than 25% from the current post, we will
```
askvortsov1, line 42: I think that 25% threshold might be a bit much; it's confusing when the button shows up some times but not others.

js/src/forum/components/PostStreamScrubber.js
```diff
@@ -65,6 +77,18 @@ export default class PostStreamScrubber extends Component {
             </a>
 
             <div className="Scrubber-scrollbar">
+              {showBackButton ? (
+                <a
+                  style={'top: ' + this.percentPerPost().index * this.stream.previousIndex + '%'}
+                  className="Scrubber-back"
+                  onclick={this.returnToLastPosition.bind(this)}
+                  title={app.translator.trans('core.forum.post_scrubber.back_title')}
```
askvortsov1, line 85: The translation for this should be added to this repo, as there are no more translations in the lang english repo. Also, a jquery tooltip might look better, although I'm not 100% sure that it's worth the pain of getting it to work.

js/src/forum/states/PostStreamState.js
```diff
@@ -133,6 +134,7 @@ class PostStreamState {
     this.needsScroll = true;
     this.targetPost = { index };
     this.animateScroll = !noAnimation;
+    this.previousIndex = index === this.previousIndex ? index : this.index;
```
askvortsov1, line 137: I wonder if we might want to also support this on goToNumber, so that back position is retained when clicking on a mention?

askvortsov1 (changes requested):
Left some code nitpicks. Thinking over this again, I'm not 100% sure that this needs to be core and not an extension; my biggest argument towards having it in core is that getting it to work as an extension (and injecting that back button into the component view function) would be very challenging and messy.

PR #2618, opened 2021-02-20
added powered-by-header

luceos:
Adds a `X-Powered-By` header. It can be disabled using:

`Flarum\Http\Middleware\FlarumPromotionHeader::disable()` in extend.php.


KyrneDev (approved):
Something I did know we needed but I love it! Everything works locally from testing.

src/Http/Middleware/FlarumPromotionHeader.php
```diff
@@ -0,0 +1,36 @@
...
+namespace Flarum\Http\Middleware;
+
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\MiddlewareInterface as Middleware;
+use Psr\Http\Server\RequestHandlerInterface;
+
+class FlarumPromotionHeader implements Middleware
+{
+    public static $enabled = true;
```
askvortsov1, line 19: I would prefer for this to inject `Config`, and read from that instead.

askvortsov1 (changes requested):
> It can be disabled using:
> 
> `Flarum\Http\Middleware\FlarumPromotionHeader::disable()` in extend.php.

The `Compat` extender is going away with stable, so no more callbacks in extend.php. I suppose it could also be done outside a callback, but I would prefer for `extend.php` to remain just extenders. Instead, let's use the config.php.

src/Http/Middleware/FlarumPromotionHeader.php
tankerkiller125: I agree, a config option would be much better in my opinion as well.

askvortsov1: I don't think an extension should be needed to toggle this, hence my preference for config.php

luceos: I like the idea of using config.php for this. Will refactor.

luceos:
I've refactored based on the review. You can now disable the promotion header in the `config.php` like so:

```php
<?php

return [
    'url' => 'http://toby.local',
    'debug' => true,
    'promote' => false,
    // ..
];

File: src/Http/Middleware/FlarumPromotionHeader.php [hunk 2]
```diff
@@ -9,25 +9,26 @@
...
 class FlarumPromotionHeader implements Middleware
 {
-    public static $enabled = true;
+    protected $enabled = true;
 
-    public static function disable()
+    public function __construct(Config $config)
     {
-        static::$enabled = false;
+        $this->enabled = $config['promote'] ?? true;
```
askvortsov1, line 24:
```suggestion
        $this->enabled = $config['flarumPromotionalHeader'] ?? true;
```

askvortsov1 (changes requested):

SychO9:
can we also add the key to StoreConfig.php so that it's there by default ?

askvortsov1:
> can we also add the key to StoreConfig.php so that it's there by default ?

Good point. It'll also be in the release announcement and config.php docs

askvortsov1 (changes requested):
I think "promote" is too broad, "promoteFlarumHeader" or "promoteFlarum" might be better.

src/Http/Middleware/FlarumPromotionHeader.php
tankerkiller125: I feel like that suggested name is a little long? I'm perfectly fine with simple "promote" if we did want to explain it more maybe "promotionHeader"

askvortsov1: I like it!

KyrneDev:
> I think "promote" is too broad, "promoteFlarumHeader" or "promoteFlarum" might be better.

What about "promoHeader"

askvortsov1 (approved):

tankerkiller125 (approved):

PR #2620, opened 2021-02-21
Initial user permission split

### LINKED ISSUE #1965: Users with edit user permission can make themselves admins

tankerkiller125:
**Fixes #1965 **

- Currently only prevents non-admins from editing admin
- Splits the edit.user permission into three permissions (username, credentials, groups)

- Is this the best way of doing this?

**Screenshot**

image

**Confirmed**

### DIFF
--- a/js/src/admin/components/PermissionGrid.js
+++ b/js/src/admin/components/PermissionGrid.js
@@ -326,12 +326,28 @@ export default class PermissionGrid extends Component {
       60
     );
 
+    items.add('userEditCredentials', {
+      icon: 'fas fa-user-cog',
+      label: app.translator.trans('core.admin.permissions.edit_users_credentials_label'),
+      permission: 'user.editCredentials',
+    });
+
     items.add(
-      'userEdit',
+      'userEditGroups',
       {
-        icon: 'fas fa-user-cog',
-        label: app.translator.trans('core.admin.permissions.edit_users_label'),
-        permission: 'user.edit',
+        icon: 'fas fa-users-cog',
+        label: app.translator.trans('core.admin.permissions.edit_users_groups_label'),
+        permission: 'user.editGroups',
+      },
+      60
+    );
+
+    items.add(
+      'userEditUsername',
+      {
+        icon: 'fas fa-address-card',
+        label: app.translator.trans('core.admin.permissions.edit_users_username_label'),
+        permission: 'user.editUsername',
       },
       60
     );
--- a/js/src/common/models/User.js
+++ b/js/src/common/models/User.js
@@ -29,8 +29,14 @@ Object.assign(User.prototype, {
   discussionCount: Model.attribute('discussionCount'),
   commentCount: Model.attribute('commentCount'),
 
-  canEdit: Model.attribute('canEdit'),
+  canEditCredentials: Model.attribute('canEditCredentials'),
+  canEditUsername: Model.attribute('canEditUsername'),
+  canEditGroups: Model.attribute('canEditGroups'),
   canDelete: Model.attribute('canDelete'),
+
+  canEdit: computed('canEditCredentials', 'canEditGroups', 'canEditUsername', function (canEditCredentials, canEditGroups, canEditUsername) {
+    return canEditCredentials || canEditGroups || canEditUsername;
+  }),
 
   avatarColor: null,
   color: computed('username', 'avatarUrl', 'avatarColor', function (username, avatarUrl, avatarColor) {
--- a/locale/core.yml
+++ b/locale/core.yml
@@ -176,7 +176,9 @@ core:
       delete_posts_label: Delete posts
       description: Configure who can see and do what.
       edit_posts_label: Edit posts
-      edit_users_label: Edit users
+      edit_users_credentials_label: Edit users credentials
+      edit_users_groups_label: Edit users groups
+      edit_users_username_label: Edit users username
       global_heading: Global
       moderate_heading: Moderate
       new_group_button: New Group
--- a/src/Api/Serializer/UserSerializer.php
+++ b/src/Api/Serializer/UserSerializer.php
@@ -19,14 +19,14 @@ class UserSerializer extends BasicUserSerializer
     {
         $attributes = parent::getDefaultAttributes($user);
 
-        $canEdit = $this->actor->can('edit', $user);
-
         $attributes += [
-            'joinTime'         => $this->formatDate($user->joined_at),
-            'discussionCount'  => (int) $user->discussion_count,
-            'commentCount'     => (int) $user->comment_count,
-            'canEdit'          => $canEdit,
-            'canDelete'        => $this->actor->can('delete', $user),
+            'joinTime'           => $this->formatDate($user->joined_at),
+            'discussionCount'    => (int) $user->discussion_count,
+            'commentCount'       => (int) $user->comment_count,
+            'canEditCredentials' => $this->actor->can('user.editCredentials'),
+            'canEditGroups'      => $this->actor->can('user.editGroups'),
+            'canEditUsername'    => $this->actor->can('user.editUsername'),
+            'canDelete'          => $this->actor->can('delete', $user),
         ];
 
         if ($user->getPreference('discloseOnline') || $this->actor->can('viewLastSeenAt', $user)) {
@@ -35,7 +35,7 @@ class UserSerializer extends BasicUserSerializer
             ];
         }
 
-        if ($canEdit || $this->actor->id === $user->id) {
+        if ($attributes['canEditCredentials'] || $this->actor->id === $user->id) {
             $attributes += [
                 'isEmailConfirmed' => (bool) $user->is_email_confirmed,
                 'email'            => $user->email
--- a/src/User/Command/EditUserHandler.php
+++ b/src/User/Command/EditUserHandler.php
@@ -58,7 +58,6 @@ class EditUserHandler
 
         $user = $this->users->findOrFail($command->userId, $actor);
 
-        $canEdit = $actor->can('edit', $user);
         $isSelf = $actor->id === $user->id;
 
         $attributes = Arr::get($data, 'attributes', []);
@@ -66,7 +65,7 @@ class EditUserHandler
         $validate = [];
 
         if (isset($attributes['username'])) {
-            $actor->assertPermission($canEdit);
+            $actor->assertPermission($actor->can('editUsername'));
             $user->rename($attributes['username']);
         }
 
@@ -78,7 +77,7 @@ class EditUserHandler
                     $validate['email'] = $attributes['email'];
                 }
             } else {
-                $actor->assertPermission($canEdit);
+                $actor->assertPermission($actor->can('editCredentials'));
                 $user->changeEmail($attributes['email']);
             }
         }
@@ -88,7 +87,7 @@ class EditUserHandler
         }
 
         if (isset($attributes['password'])) {
-            $actor->assertPermission($canEdit);
+            $actor->assertPermission($actor->can('editCredentials'));
             $user->changePassword($attributes['password']);
 
             $validate['password'] = $attributes['password'];
@@ -108,7 +107,7 @@ class EditUserHandler
         }
 
         if (isset($relationships['groups']['data']) && is_array($relationships['groups']['data'])) {
-            $actor->assertPermission($canEdit);
+            $actor->assertPermission($actor->can('editGroups'));
 
             $newGroupIds = [];
             foreach ($relationships['groups']['data'] as $group) {
--- a/js/src/forum/components/EditUserModal.js
+++ b/js/src/forum/components/EditUserModal.js
@@ -47,96 +47,106 @@ export default class EditUserModal extends Modal {
   fields() {
     const items = new ItemList();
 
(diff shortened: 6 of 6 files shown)


src/Api/Serializer/UserSerializer.php
```diff
@@ -19,14 +19,14 @@ protected function getDefaultAttributes($user)
...
-            'joinTime'         => $this->formatDate($user->joined_at),
-            'discussionCount'  => (int) $user->discussion_count,
-            'commentCount'     => (int) $user->comment_count,
-            'canEdit'          => $canEdit,
-            'canDelete'        => $this->actor->can('delete', $user),
+            'joinTime'           => $this->formatDate($user->joined_at),
+            'discussionCount'    => (int) $user->discussion_count,
+            'commentCount'       => (int) $user->comment_count,
+            'canEditCredentials' => $this->actor->can('user.editCredentials'),
+            'canEditGroups'      => $this->actor->can('user.editGroups'),
```
askvortsov1, line 27: These should be applied on the $user object so policies can work

askvortsov1: Also, I don't think we need the `user` prefix

src/User/Command/EditUserHandler.php
```diff
@@ -58,15 +58,14 @@ public function handle(EditUser $command)
...
-        $canEdit = $actor->can('edit', $user);
         $isSelf = $actor->id === $user->id;
 
         $attributes = Arr::get($data, 'attributes', []);
         $relationships = Arr::get($data, 'relationships', []);
         $validate = [];
 
         if (isset($attributes['username'])) {
-            $actor->assertPermission($canEdit);
+            $actor->assertPermission($actor->can('user.editUsername'));
```
askvortsov1, line 68: Same applies here; this should apply on the $user object, and the `'user'` prefix is unnecessary

askvortsov1 (changes requested):
This also needs some integration tests for EditUserController

src/User/Command/EditUserHandler.php
tankerkiller125: The user prefix is required, in my testing without the user. prefix it did not work correctly and always returned false. This is one of those cases of the magic not working correctly.

src/Api/Serializer/UserSerializer.php
tankerkiller125: The user prefix is required, in my testing without the user. prefix it did not work correctly and always returned false. This is one of those cases of the magic not working correctly.

js/src/admin/components/PermissionGrid.js
```diff
@@ -326,12 +326,28 @@ export default class PermissionGrid extends Component {
...
     items.add(
-      'userEdit',
+      'userEditUsername',
       {
-        icon: 'fas fa-user-cog',
-        label: app.translator.trans('core.admin.permissions.edit_users_label'),
-        permission: 'user.edit',
+        icon: 'fas fa-address-card',
+        label: app.translator.trans('core.admin.permissions.edit_users_username_label'),
+        permission: 'user.editUsername',
```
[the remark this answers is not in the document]
tankerkiller125: The user prefix is required, in my testing without the user. prefix it did not work correctly and always returned false. This is one of those cases of the magic not working correctly.

As for the edit permission I guess it could do the username. TBH I split these permissions based on the previous PR with the only change being the removal of dot notated permissions and upgrading to beta.15.

src/Api/Serializer/UserSerializer.php
askvortsov1: The "magic" only applies when it's run through User Policy, which happens when the subject argument is a user. The prefix needs to be removed at the same time that the second user argument is added.

js/src/admin/components/PermissionGrid.js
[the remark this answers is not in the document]
askvortsov1: Alternatively, username could also be credentials, and regular edit should be "Other attributes", which would be used by stuff like extensions. Not sure which is better, I'll think about it.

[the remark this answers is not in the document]
tankerkiller125: I think for right now we might want to make the username a credential. But I think we should have a conversation about using the username for login in the future as I've voiced my opinion on that matter before that it shouldn't be permitted for many reasons. I'll probably create an issue for that later.

js/src/common/models/User.js
```diff
@@ -29,9 +29,15 @@ Object.assign(User.prototype, {
   discussionCount: Model.attribute('discussionCount'),
   commentCount: Model.attribute('commentCount'),
 
-  canEdit: Model.attribute('canEdit'),
+  canEditCredentials: Model.attribute('canEditCredentials'),
```
askvortsov1, line 32: Leftover?

js/src/admin/components/PermissionGrid.js
```diff
@@ -329,13 +329,29 @@ export default class PermissionGrid extends Component {
     items.add(
```
askvortsov1, line 329: I'd move this after the rest if it represents "other attributes"

locale/core.yml
```diff
@@ -177,6 +177,8 @@ core:
       description: Configure who can see and do what.
       edit_posts_label: Edit posts
       edit_users_label: Edit users
```
askvortsov1, line 179: Should this be changed to "Edit user attributes"?

locale/core.yml
```diff
@@ -177,6 +177,8 @@ core:
...
+      edit_users_credentials_label: Edit users credentials
```
askvortsov1, line 180: User should be singular here, in the translated text

askvortsov1 (changes requested):

js/src/common/models/User.js
```diff
@@ -29,9 +29,14 @@ Object.assign(User.prototype, {
...
+  canEditGroups: Model.attribute('canEditGroups'),
   canDelete: Model.attribute('canDelete'),
 
+  canEdit: computed('canEditCredentials', 'canEditGroups', 'canEdit', function (canEdit, canEditCredentials, canEditGroups) {
```
askvortsov1, line 36: `computer` will try to call its dependencies as methods, so it'll try to call `canEdit` and end up in a recursive loop. Maybe we can set it to `canEditAttributes` on the js user model?

askvortsov1 (changes requested):

askvortsov1 (changes requested):
Code looks fine, but since it's dealing with API stuff (and quite important API stuff at that) it still needs integration tests

src/User/Command/EditUserHandler.php
```diff
@@ -109,15 +110,21 @@ public function handle(EditUser $command)
...
+            $oldGroupIds = Arr::pluck($oldGroups, 'id');
+
             $newGroupIds = [];
             foreach ($relationships['groups']['data'] as $group) {
                 if ($id = Arr::get($group, 'id')) {
                     $newGroupIds[] = $id;
                 }
             }
 
+            $adminChanged = in_array('1', array_diff($oldGroupIds, $newGroupIds)) || in_array('1', array_diff($newGroupIds, $oldGroupIds));
```
askvortsov1, line 123: Should this be moved out to a guard (like SelfDemotionGuard)? We decided to leave it here since that felt simpler. Alternatively, we could move SelfDemotionGuard here for simplicity?

tests/integration/api/users/UpdateTest.php
```diff
@@ -26,7 +27,24 @@ protected function setUp(): void
         $this->prepareDatabase([
             'users' => [
                 $this->normalUser(),
-            ]
+                [
```
askvortsov1, line 30: We decided to put this here and not in `RetrievesAuthorizedUsers` since it's only used in one test case

askvortsov1 (approved):
Note for reviewers: the vast majority of changes is tests

askvortsov1:
We also added restrictions that stop non-admins from editing credentials of admins, and stop non-admins from promoting/demoting admins

askvortsov1 (approved):
On a side note, when this is merged it should be squashed into 3 commits: one for the split, the other 2 for the other restrictions introduced.

src/Api/Serializer/UserSerializer.php
```diff
@@ -19,14 +19,14 @@ protected function getDefaultAttributes($user)
...
         $attributes += [
-            'joinTime'         => $this->formatDate($user->joined_at),
-            'discussionCount'  => (int) $user->discussion_count,
-            'commentCount'     => (int) $user->comment_count,
-            'canEdit'          => $canEdit,
-            'canDelete'        => $this->actor->can('delete', $user),
+            'joinTime'           => $this->formatDate($user->joined_at),
+            'discussionCount'    => (int) $user->discussion_count,
+            'commentCount'       => (int) $user->comment_count,
+            'canEdit'            => $this->actor->can('edit', $user),
```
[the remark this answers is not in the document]
tankerkiller125: I think I didn't do that to prevent breaking extensions that might use canEdit.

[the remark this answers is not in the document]
askvortsov1: We're not losing the method on the frontend user model, so we'd be preserving this for API consumers. But the meaning of the value has changed, so I'm not sure it's worth it to preserve; renaming would force consumers to adapt to the new system.

askvortsov1 (changes requested):
Temporarily removing approval as per https://github.com/flarum/core/pull/2620#discussion_r583663634, we should either rename the value returned by the API to `canEditAttributes`, or rename the frontend methods to `canEdit` (and use something else for the combined method name).

The second option might actually be better; that way, we are consistent across permission name, frontend, and backend.

src/Api/Serializer/UserSerializer.php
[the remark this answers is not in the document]
SychO9: we should rename this, it's quiet confusing right now what the difference between `canEdit` and `canEditAttributes` is, on the frontend `canEdit` is true if the user has the ability to edit any of the three new permissions, on the backend (serialization) it just refers to the `canEditAttributes` ability.

Are we gonna keep `canEdit` the way it is on the frontend ? (ie any of the three abilities), I'm assuming we're deprecating it, if so, can we comment on the frontend model that it's deprecated please ?

SychO9 (changes requested):
The backend code looks solid, but the frontend not so much, here are some issues I noticed while testing the permissions:

* The EditUserModal is empty if the only permission given to the user is edit attributes.
flarum lan_u_SychO (1)

* By giving the user only permission to edit credentials, they can open the EditUserModal on normal members, but can't actually change credentials. (haven't looked too deep into it, but the frontend seems to still send an array of groups relationship which naturally fails).
flarum lan_u_simonisbailey

* Giving a user edit credentials permission does not allow to edit an admin's credentials (as it should) but once the user is given edit attributes permission, they can open the admin's EditUserModal (it fails on edit of course, because the backend seems solid in the behaviour).

* If a user have permission to edit groups only, and they try to edit groups (anything but the admin group ofc) they're not allowed to (because the username is included in the payload).

askvortsov1:
> * The EditUserModal is empty if the only permission given to the user is edit attributes.

Edited to show a message

> Are we gonna keep canEdit the way it is on the frontend ? (ie any of the three abilities), I'm assuming we're deprecating it, if so, can we comment on the frontend model that it's deprecated please ?

Renamed everything to `editUser` for consistency.

All the other stuff has been fixed, at least based on the tests I ran locally.

askvortsov1 (approved):

src/User/Command/EditUserHandler.php
```diff
@@ -78,17 +77,18 @@ public function handle(EditUser $command)
...
-        if ($actor->isAdmin() && ! empty($attributes['isEmailConfirmed'])) {
+        if (! empty($attributes['isEmailConfirmed'])) {
+            $actor->assertAdmin();
             $user->activate();
         }
 
         if (isset($attributes['password'])) {
-            $actor->assertPermission($canEdit);
+            $actor->assertPermission($actor->can('editCredentials', $user));
             $user->changePassword($attributes['password']);
```
SychO9, line 92: off topic observation: we force users to use the change password button (reset password email) on the frontend but not on the backend, so they could just send a payload and directly edit it.

I don't know if this was done on purpose.

src/User/Command/EditUserHandler.php
SychO9: I don't feel strongly about this, but I wouldn't remove the SelfDemotionGuard either, maybe just leave things as is ?

SychO9 (approved):
Code looks good, did some tests locally as well.

src/User/Command/EditUserHandler.php
askvortsov1: I don't disagree, but this PR already contains a lot so I'd leave this alone for now

SychO9: Yea, I wouldn't want that changed here anyway :P was an observation as it looked odd.

Issue #2416, opened 2020-10-27
New, fixed version of reverse proxy PR


askvortsov1:
Even if it's extremely simple, it'd be good to move this logic into a middleware.

PR #2624, opened 2021-02-22
Centralized IP Handler

tankerkiller125:

Centralizes the IP processing/handler to a middleware that's loaded on all request (can not be removed or replaced)

- Did I find and replace all the IP insert locations
- Is InstalledApp the best place to do this (was originally recommended by Franz)
- 
**Confirmed**

### DIFF
--- a/src/Http/Middleware/IpHandler.php
+++ b/src/Http/Middleware/IpHandler.php
@@ -1,5 +1,12 @@
 <?php
 
+/*
+ * This file is part of Flarum.
+ *
+ * For detailed copyright and license information, please view the
+ * LICENSE file that was distributed with this source code.
+ */
+
 namespace Flarum\Http\Middleware;
 
 use Illuminate\Support\Arr;
--- a/src/Http/Middleware/IpHandler.php
+++ b/src/Http/Middleware/IpHandler.php
@@ -1,0 +1,18 @@
+<?php
+
+namespace Flarum\Http\Middleware;
+
+use Illuminate\Support\Arr;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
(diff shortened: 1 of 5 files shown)


askvortsov1 (changes requested):
I think I'd prefer `ProcessIp`, since it's not an event handler, but other than that, looks good to me!

We could add tests now, or when the full #2416 is implemented later down the line

src/Http/Middleware/IpHandler.php
```diff
@@ -0,0 +1,26 @@
...
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\MiddlewareInterface as Middleware;
+use Psr\Http\Server\RequestHandlerInterface;
+
+class IpHandler implements Middleware
+{
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
+    {
+        $ipAddress = Arr::get($request->getServerParams(), 'REMOTE_ADDR', '127.0.0.1');
```
KyrneDev, line 22: Won't this break with cloudflare? They use a different header to pass the real IP

askvortsov1: The point of this PR is to centralize the logic for getting the PR in a Middleware, and set it as a request attribute. That way we don't have duplicate code for calculating it in multiple controllers. In the future this will be augmented to deal with reverse proxies and other stuff but for now it's just a refactor of where the logic is to prevent future BC breaks.

KyrneDev: Sounds good

askvortsov1 (approved):

KyrneDev (approved):

PR #2626, opened 2021-02-22
Add users list to admin dashboard

### LINKED ISSUE #283: User list
We want to offer an extensible user list which displays users sorted by metrics (no. of posts/discussions in core, and could be extended to show no. of likes received, etc.)

What needs to be done:
- Decide if this should be part of core, or should be a default extension. (i.e. should we allow it to be disabled?)
- Design it.
- Implementation needs to be nutted out

davwheat:
**Fixes #283**

Adds a list of all forum users to the Admin dashboard

**~~Screenshot~~ GIF!**

**Confirmed**

**Required changes:**

## To do

- [x] Add username heading
- [x] Add join date heading
- [x] Add email heading
- [x] Add Edit User button
- [x] Add link to user profile
- [x] Add group badges
- [ ] Add "other data" modal and button (will show all attributes of the User model, including those of extensions)

### DIFF
--- a/js/src/admin/components/AdminNav.js
+++ b/js/src/admin/components/AdminNav.js
@@ -96,7 +96,7 @@ export default class AdminNav extends Component {
 
     items.add(
       'userList',
-      <LinkButton href={app.route('userlist')} icon="fas fa-users" title={app.translator.trans('core.admin.nav.userlist_title')}>
+      <LinkButton href={app.route('users')} icon="fas fa-users" title={app.translator.trans('core.admin.nav.userlist_title')}>
         {app.translator.trans('core.admin.nav.userlist_button')}
       </LinkButton>
     );
--- a/js/src/admin/components/UserListPage.tsx
+++ b/js/src/admin/components/UserListPage.tsx
@@ -36,7 +36,7 @@ type ApiPayload = {
   };
 };
 
-type UsersApiResponse = User[] & { payload: ApiPayload; };
+type UsersApiResponse = User[] & { payload: ApiPayload };
 
 /**
  * Admin page which displays a paginated list of all users on the forum.
--- a/js/src/admin/components/UserListPage.tsx
+++ b/js/src/admin/components/UserListPage.tsx
@@ -243,42 +243,50 @@ export default class UserListPage extends AdminPage {
       {
         name: app.translator.trans('core.admin.users.grid.columns.email.title'),
         content: (user: User) => {
-          function toggleEmailVisibility() {
+          function setEmailVisibility(visible: boolean) {
             // Get needed jQuery element refs
             const emailContainer = $(`[data-column-name=emailAddress][data-user-id=${user.id()}] .UserList-email`);
             const emailAddress = emailContainer.find('.UserList-emailAddress');
             const emailToggleButton = emailContainer.find('.UserList-emailIconBtn');
             const emailToggleButtonIcon = emailToggleButton.find('.icon');
 
-            const emailShown = emailContainer.attr('data-email-shown') === 'true';
+            emailToggleButton.attr(
+              'title',
+              extractText(
+                visible
+                  ? app.translator.trans('core.admin.users.grid.columns.email.visibility_hide')
+                  : app.translator.trans('core.admin.users.grid.columns.email.visibility_show')
+              )
+            );
 
-            if (emailShown) {
-              // Email currently shown, switching to hidden
+            emailAddress.attr('aria-hidden', visible ? 'false' : 'true');
 
-              emailToggleButton.attr('title', extractText(app.translator.trans('core.admin.users.grid.columns.email.visibility_show')));
-
-              emailAddress.attr('aria-hidden', 'true');
-
-              emailToggleButtonIcon.removeClass('fa-eye');
-              emailToggleButtonIcon.addClass('fa-eye-slash');
-            } else {
-              // Email currently hidden, switching to shown
-
-              emailToggleButton.attr('title', extractText(app.translator.trans('core.admin.users.grid.columns.email.visibility_hide')));
-
-              emailAddress.removeAttr('aria-hidden');
-
+            if (visible) {
               emailToggleButtonIcon.addClass('fa-eye');
               emailToggleButtonIcon.removeClass('fa-eye-slash');
+            } else {
+              emailToggleButtonIcon.removeClass('fa-eye');
+              emailToggleButtonIcon.addClass('fa-eye-slash');
             }
 
             // Need the string interpolation to prevent TS error.
-            emailContainer.attr('data-email-shown', `${!emailShown}`);
+            emailContainer.attr('data-email-shown', `${visible}`);
+          }
+
+          function toggleEmailVisibility() {
+            const emailContainer = $(`[data-column-name=emailAddress][data-user-id=${user.id()}] .UserList-email`);
+            const emailShown = emailContainer.attr('data-email-shown') === 'true';
+
+            if (emailShown) {
+              setEmailVisibility(false);
+            } else {
+              setEmailVisibility(true);
+            }
           }
 
           return (
             <div class="UserList-email" key={user.id()} data-email-shown="false">
-              <span class="UserList-emailAddress" aria-hidden>
+              <span class="UserList-emailAddress" aria-hidden onclick={() => setEmailVisibility(true)}>
                 {user.email()}
               </span>
               <button
--- a/js/src/admin/components/UserListPage.tsx
+++ b/js/src/admin/components/UserListPage.tsx
@@ -166,7 +166,7 @@ export default class UserListPage extends AdminPage {
           const profileUrl = `${app.forum.attribute('baseUrl')}/u/${user.slug()}`;
 
           return (
-            <a target="_blank" href={profileUrl} title={app.translator.trans('core.admin.user_list.grid.default_columns.profile_link')}>
+            <a target="_blank" href={profileUrl} title={app.translator.trans('core.admin.user_list.grid.default_columns.profile_link_tooltip')}>
               {icon('fas fa-link')}
             </a>
           );
--- a/js/src/admin/components/UserListPage.tsx
+++ b/js/src/admin/components/UserListPage.tsx
@@ -1,5 +1,5 @@
 import GroupBadge from '../../common/components/GroupBadge';
-import EditGroupModal from './EditGroupModal';
+import EditUserModal from '../../common/components/EditUserModal';
 import Group from '../../common/models/Group';
 import icon from '../../common/helpers/icon';
 import type User from '../../common/models/User';
@@ -140,6 +140,23 @@ export default class UserListPage extends AdminPage {
         </span>
       ),
     });
+
+    columns.add(
+      'editUser',
+      {
+        name: app.translator.trans('core.admin.user_list.grid.default_columns.edit_user'),
+        content: (user: User) => (
+          <Button
+            className="Button UserList-editModalBtn"
+            title={app.translator.trans('core.admin.user_list.grid.default_columns.edit_user_tooltip', { username: user.username() })}
+            onclick={() => app.modal.show(EditUserModal, { user })}
+          >
+            {app.translator.trans('core.admin.user_list.grid.default_columns.edit_user_button')}
+          </Button>
+        ),
+      },
+      -90
+    );
 
     columns.add(
       'profileLink',
--- a/js/src/admin/components/UserListPage.tsx
+++ b/js/src/admin/components/UserListPage.tsx
@@ -105,7 +105,7 @@ export default class UserListPage extends AdminPage {
         role="table"
         // +1 to account for header
         aria-rowcount={this.pageData.length + 1}
-        aria-colcount={this.columns.length}
+        aria-colcount={columns.length}
         aria-live="polite"
         aria-busy={this.isLoadingPage ? 'true' : 'false'}
       >
--- a/js/src/admin/components/UserListPage.tsx
+++ b/js/src/admin/components/UserListPage.tsx
@@ -25,11 +25,14 @@ type ColumnData = {
 };
 
 type ApiPayload = {
-  data: [];
-  included: [];
+  data: Record<string, unknown>[];
+  included: Record<string, unknown>[];
   links: {
     first: string;
     next?: string;
+  };
+  meta: {
+    total: number;
   };
 };
 
--- a/js/src/admin/components/UserListPage.tsx
+++ b/js/src/admin/components/UserListPage.tsx
@@ -196,7 +196,13 @@ export default class UserListPage extends AdminPage {
           const profileUrl = `${app.forum.attribute('baseUrl')}/u/${user.slug()}`;
 
           return (
-            <a target="_blank" href={profileUrl} title={app.translator.trans('core.admin.user_list.grid.default_columns.profile_link_tooltip')}>
+            <a
+              target="_blank"
+              href={profileUrl}
+              title={extractText(
+                app.translator.trans('core.admin.user_list.grid.default_columns.profile_link_tooltip', { username: user.username() })
+              )}
+            >
               {icon('fas fa-link')}
             </a>
           );
--- a/js/src/admin/components/UserListPage.tsx
+++ b/js/src/admin/components/UserListPage.tsx
@@ -257,7 +257,6 @@ export default class UserListPage extends AdminPage {
 
               emailToggleButton.attr('title', extractText(app.translator.trans('core.admin.users.grid.columns.email.visibility_show')));
 
-              emailAddress.text(app.translator.trans('core.admin.users.grid.columns.email.hidden_placeholder'));
               emailAddress.attr('aria-hidden', 'true');
 
               emailToggleButtonIcon.removeClass('fa-eye');
@@ -267,7 +266,6 @@ export default class UserListPage extends AdminPage {
 
               emailToggleButton.attr('title', extractText(app.translator.trans('core.admin.users.grid.columns.email.visibility_hide')));
 
-              emailAddress.text(user.email());
               emailAddress.removeAttr('aria-hidden');
 
               emailToggleButtonIcon.addClass('fa-eye');
@@ -281,7 +279,7 @@ export default class UserListPage extends AdminPage {
           return (
             <div class="UserList-email" key={user.id()} data-email-shown="false">
               <span class="UserList-emailAddress" aria-hidden>
-                {app.translator.trans('core.admin.users.grid.columns.email.hidden_placeholder')}
+                {user.email()}
               </span>
               <button
                 onclick={toggleEmailVisibility}
--- a/js/src/admin/components/UserListPage.tsx
+++ b/js/src/admin/components/UserListPage.tsx
@@ -90,7 +90,7 @@ export default class UserListPage extends AdminPage {
 
       return [
         <section class="UserListPage-grid UserListPage-grid--loading">
-          <LoadingIndicator />
+          <LoadingIndicator containerClassName="LoadingIndicator--block" size="large" />
         </section>,
       ];
     }
@@ -138,7 +138,7 @@ export default class UserListPage extends AdminPage {
         )}
 
         {/* Loading spinner that shows when a new page is being loaded */}
-        {this.isLoadingPage && <LoadingIndicator />}
+        {this.isLoadingPage && <LoadingIndicator size="large" />}
       </section>,
       <nav class="UserListPage-gridPagination">
         <Button
--- a/js/src/admin/components/UserListPage.tsx
+++ b/js/src/admin/components/UserListPage.tsx
@@ -198,7 +198,7 @@ export default class UserListPage extends AdminPage {
         content: (user: User) => {
           return (
             <div class="UserList-email" data-user-id={user.id()} data-email-shown="false">
-              <span class="UserList-emailAddress">{app.translator.trans('core.admin.user_list.grid.default_columns.email_hidden')}</span>
+              <span class="UserList-emailAddress">{app.translator.trans('core.admin.user_list.grid.default_columns.email_visibility_show')}</span>
               <button
                 onclick={() => {
                   // Get needed jQuery element refs
@@ -215,7 +215,7 @@ export default class UserListPage extends AdminPage {
                     // Update tooltip
                     emailToggleButton.attr(
                       'title',
-                      extractText(app.translator.trans('core.admin.user_list.grid.default_columns.email_visibility_hide'))
+                      extractText(app.translator.trans('core.admin.user_list.grid.default_columns.email_visibility_show'))
                     );
 
                     // Replace real email with placeholder email
@@ -230,7 +230,7 @@ export default class UserListPage extends AdminPage {
                     // Update tooltip
                     emailToggleButton.attr(
                       'title',
-                      extractText(app.translator.trans('core.admin.user_list.grid.default_columns.email_visibility_show'))
+                      extractText(app.translator.trans('core.admin.user_list.grid.default_columns.email_visibility_hide'))
                     );
 
                     // Replace placeholder email with real email
@@ -246,7 +246,7 @@ export default class UserListPage extends AdminPage {
                 class="UserList-emailIconBtn"
                 title={app.translator.trans('core.admin.user_list.grid.default_columns.email_show')}
               >
-                {icon('far fa-eye', { className: 'icon' })}
+                {icon('far fa-eye fa-fw', { className: 'icon' })}
               </button>
             </div>
           );
--- a/js/src/admin/components/UserListPage.tsx
+++ b/js/src/admin/components/UserListPage.tsx
@@ -253,34 +253,29 @@ export default class UserListPage extends AdminPage {
             const emailShown = emailContainer.attr('data-email-shown') === 'true';
 
             if (emailShown) {
-              //! Email currently shown, switching to hidden
+              // Email currently shown, switching to hidden
 
-              // Update tooltip
               emailToggleButton.attr('title', extractText(app.translator.trans('core.admin.users.grid.default_columns.email_visibility_show')));
 
-              // Replace real email with placeholder email
               emailAddress.text(app.translator.trans('core.admin.users.grid.default_columns.email_hidden'));
               emailAddress.attr('aria-hidden', 'true');
 
-              // Change button icons
               emailToggleButtonIcon.removeClass('fa-eye');
               emailToggleButtonIcon.addClass('fa-eye-slash');
             } else {
-              //! Email currently hidden, switching to shown
+              // Email currently hidden, switching to shown
 
-              // Update tooltip
               emailToggleButton.attr('title', extractText(app.translator.trans('core.admin.users.grid.default_columns.email_visibility_hide')));
 
-              // Replace placeholder email with real email
               emailAddress.text(user.email());
               emailAddress.removeAttr('aria-hidden');
 
-              // Change button icons
               emailToggleButtonIcon.addClass('fa-eye');
               emailToggleButtonIcon.removeClass('fa-eye-slash');
             }
 
-            emailContainer.attr('data-email-shown', !emailShown);
+            // Need the string interpolation to prevent TS error.
+            emailContainer.attr('data-email-shown', `${!emailShown}`);
           }
 
           return (
--- a/js/src/admin/components/UserListPage.tsx
+++ b/js/src/admin/components/UserListPage.tsx
@@ -173,7 +173,7 @@ export default class UserListPage extends AdminPage {
         },
       },
       // This should probably come last unless an ext really wants to be last
-      1000
+      -100
     );
 
     return columns;
--- a/js/src/admin/components/UserListPage.tsx
+++ b/js/src/admin/components/UserListPage.tsx
@@ -31,9 +31,6 @@ type ApiPayload = {
     first: string;
     next?: string;
   };
-  meta: {
-    totalCount: number;
-  };
 };
 
 type UsersApiResponse = User[] & { payload: ApiPayload };
@@ -54,8 +51,12 @@ export default class UserListPage extends AdminPage {
 
   /**
    * Total number of forum users.
+   *
+   * Fetched from the active `AdminApplication` (`app`), with
+   * data provided by `AdminPayload.php`, or `flarum/statistics`
+   * if installed.
    */
-  protected userCount: number = -1;
+  readonly userCount: number = app.data.statistics.users.total;
 
   /**
    * Get total number of user pages.
@@ -112,7 +113,6 @@ export default class UserListPage extends AdminPage {
       .then((apiData: UsersApiResponse) => {
         // Next link won't be present if there's no more data
         this.moreData = !!apiData.payload.links.next;
-        this.userCount = (apiData.payload.meta && apiData.payload.meta.totalCount) || -1;
 
         let data = apiData;
 
--- a/js/src/admin/components/UserListPage.tsx
+++ b/js/src/admin/components/UserListPage.tsx
@@ -32,11 +32,11 @@ type ApiPayload = {
     next?: string;
   };
   meta: {
-    total: number;
+    totalCount: number;
   };
 };
 
-type UsersApiResponse = User[] & { payload: ApiPayload; meta: { totalCount: number } };
+type UsersApiResponse = User[] & { payload: ApiPayload; };
 
 /**
  * Admin page which displays a paginated list of all users on the forum.
@@ -45,7 +45,7 @@ export default class UserListPage extends AdminPage {
   /**
    * Number of users to load per page.
    */
-  private numPerPage: number = 3;
+  private numPerPage: number = 50;
 
   /**
    * Current page number. Zero-indexed.
@@ -55,7 +55,7 @@ export default class UserListPage extends AdminPage {
   /**
    * Total number of forum users.
    */
-  userCount: number = -1;
+  protected userCount: number = -1;
 
   /**
    * Get total number of user pages.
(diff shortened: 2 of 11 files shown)


js/src/admin/components/AdminNav.js
```diff
@@ -94,6 +94,13 @@ export default class AdminNav extends Component {
       </LinkButton>
     );
 
+    items.add(
+      'userslist',
```
askvortsov1, line 98: I think `userList` is more correct

less/admin/UsersListPage.less
```diff
@@ -0,0 +1,34 @@
...
+    width: 100%;
+
+    &--loaded {
+      display: grid;
+      overflow-x: auto;
+    }
+
+    &--header {
+      font-weight: bold;
+      border-bottom: 1px solid #999;
```
askvortsov1, line 12: These colors should use less variables for css / theming compatibility

less/admin/UsersListPage.less
```diff
@@ -0,0 +1,34 @@
+.UsersListPage {
```
askvortsov1, line 1: I'm not sure how well this will scale on mobile if new columns are added

askvortsov1 (commented):
Thank you for putting this together! It's certainly a much needed feature. A few initial comments (I know this is WIP) to guide development:

- `EditUserModal` doesn't rely on any forum components. Perhaps that could be added here as a button?
- A list of group badges might be appropriate here
- Would we want to include any kind of search? I'd lean towards doing that in a separate PR since all the search stuff is currently in the forum component
- Could a table component be a candidate for its own component? It would also be useful for https://github.com/flarum/flags/pull/21

davwheat:
Thanks for the comment! I'll address those a bit later on.

`EditUserModal` is under forum at the moment. Should I move it to common or just import it from forum?

askvortsov1:
> EditUserModal is under forum at the moment. Should I move it to common or just import it from forum?

Stuff used by both common and forum should be under common.

I'm not 100% sure about some of the stuff I listed (especially search, as all components for that are on the forum side). I think that it might be preferable to keep this minimal at first. We'll certainly want to eventually add more powerful management (for example, searching users, selecting checkboxes on the side, batch group assignment, etc), but that might be better as an extension.

less/admin/UsersListPage.less
davwheat: Now it should scroll horizontally when the viewport is shrinked.

js/src/admin/components/AdminNav.js
davwheat: Updated :)

less/admin/UsersListPage.less
davwheat: Done!

davwheat:
> * `EditUserModal` doesn't rely on any forum components. Perhaps that could be added here as a button?

Added a button to open it up.

> * A list of group badges might be appropriate here

Also implemented!

> * Would we want to include any kind of search? I'd lean towards doing that in a separate PR since all the search stuff is currently in the forum component

I think it'd be a great addition but, as you said, probably best in a separate PR.

> * Could a table component be a candidate for its own component? It would also be useful for [flarum/flags#21](https://github.com/flarum/flags/pull/21)

If you'd like me to make it a separate component, we should probably discuss a better API for it. I personally like the dynamic `columns` idea I "created", but we might want to separate these into `data` and `actions`, for example, as actions would almost always go at the end of a table row.

askvortsov1:
> I personally like the dynamic columns idea I "created" where each cell is generated by a function with the data passed to it, but we might want to separate these into data and actions, for example, as actions would almost always go at the end of a table row.

Yeah I think that's quite sensible. PermissionGrid uses something similar I believe, but I don't know if we'd want to refactor the two under a common use case, that might be a bit much (definitely not in this PR). 

> But if it needs back-end changes, I will run away faster than you can say "PHP"

It wouldn't not require PHP :stuck_out_tongue:

davwheat:
That sounds like a good idea to me!

> It wouldn't not require PHP 😛

🏃

src/Api/Controller/ListUsersController.php
```diff
@@ -78,6 +79,8 @@ protected function data(ServerRequestInterface $request, Document $document)
         $offset = $this->extractOffset($request);
         $load = $this->extractInclude($request);
 
+        $document->addMeta('totalCount', User::count());
```
askvortsov1, line 82: I see you've come running back! :smile:

davwheat: Shhh! I was trying to hide it, then write a miraculous victory speech! You've exposed me now!

davwheat:
Ok, so I've implemented `User::count()` as part of the `ListUsersController`. This works perfectly and allows me to show the correct number of "total pages".

I've no clue about what sort of impact this has on the API response times, especially for very big forums. It's probably something that should be looked into, just in case.

I'm just adding emails to the grid -- I'm thinking they should be blurred by default, just an extra thing to prevent "over shoulder data leakage".

src/Api/Controller/ListUsersController.php
askvortsov1: According to all known laws
of computer science,

  
there is no way a DavWheat
should be able to PHP.

  
Its Javascript focus is too intense to get
its logic to the backend.

  
The DavWheat, of course, PHPs anyway

  
because DavWheats don't care
what computer scientists think is impossible.



... 
I'll see myself out.

davwheat: That was scarily accurate! 😂

Are you *sure* you don't want to become a ~~bee movie plagiarist~~ writer in the future? :P

askvortsov1, line 82: I'll open this as a separate comment. On a more serious note, I'm a bit wary about making this modification. However, we're currently improperly using the 5th argument of `addPaginationLinks`: we should be providing a total, which would put together a "last" item in the JSON:API links. That change should be done in a separate issue/PR though.

For now, maybe the API Payload would be a better place to pass this?

js/src/forum/compat.js
```diff
@@ -72,6 +71,7 @@ import Search from './components/Search';
 import DiscussionListItem from './components/DiscussionListItem';
 import LoadingPost from './components/LoadingPost';
 import PostsUserPage from './components/PostsUserPage';
+import EditUserModal from '../common/components/EditUserModal';
```
askvortsov1, line 74: This should be labelled as deprecated

askvortsov1 (changes requested):
Overall this is looking very, very nice! Left a few more comments

KyrneDev:
> PS: Moving the EditUserModal to common is breaking -- should it be exported from forum as well to allow backwards-compatibility?

It actually shouldn't be breaking. Right now we don't use the forum/common/admin namespaces for imports on the extension side - so as long as it is re-exported in the common compat.js file compatibility should be maintained. 

Also, is it only 3 users per page? That could be quite cumbersome for a large forum, maybe expand to 10 users per page? I would also like to see a search/filter function on the grid. We already have a user search API so it's only a matter of connecting the two.

davwheat:
> @KyrneDev It actually shouldn't be breaking. Right now we don't use the forum/common/admin namespaces for imports on the extension side - so as long as it is re-exported in the common compat.js file compatibility should be maintained.

Oh, I learn something new every day! That's great :P

> Also, is it only 3 users per page? That could be quite cumbersome for a large forum, maybe expand to 10 users per page?

It is 10 and can be changed easily. I had it set to 3 for demonstration purposes because I didn't feel like creating 100s of fake users :P

> I would also like to see a search/filter function on the grid. We already have a user search API so it's only a matter of connecting the two.

I would too. @askvortsov1 suggested that this should be done in a separate PR because stuffing all of the changes needed for it would bulk up the PR a bunch.

To be honest, a live search feature wouldn't be too tricky and could just use the endpoint without any of the forum search code/components. What do you think of that?

KyrneDev:
> To be honest, a live search feature wouldn't be too tricky and could just use the endpoint without any of the forum search code/components. What do you think of that? 

That's what I was thinking. We don't need any special search components, it should be a *filter* rather than a *search.* That's the key distinction. @askvortsov1, thoughts?

askvortsov1:
> Right now we don't use the forum/common/admin namespaces for imports on the extension side - so as long as it is re-exported in the common compat.js file compatibility should be maintained.

Starting in beta 15, the `forum/common/admin` are the officially recommended ways to import, although I don't think we announced that. See #2085.

> I had it set to 3 for demonstration purposes because I didn't feel like creating 100s of fake users :P

https://discuss.flarum.org/d/21160-fake-data has been a godsend for testing

> maybe expand to 10 users per page

I'd say significantly more, maybe 25-50? 10 is not that many with properly sized table rows.

> To be honest, a live search feature wouldn't be too tricky and could just use the endpoint without any of the forum search code/components. What do you think of that?

If there's something we can remove from this PR and move into a different one, we should do so because it's already quite large.

davwheat:
> https://discuss.flarum.org/d/21160-fake-data has been a godsend for testing

I'll use that, thanks! Riiight after I update it to work with Beta 16+

src/Api/Controller/ListUsersController.php
davwheat: Even with the `last` item, surely that won't provide the total number of users directly, right?

We could extract it from the URL, but imo that's extra processing that should be done server-side instead of client-side, unless I'm misunderstanding what you meant.

js/src/forum/compat.js
davwheat: I've moved this to the common `compat.js`. This should be combined with the other `compat`s and mean that people importing `flarum/components/EditUserModal` should be fine, right?

Will having this be exported from common and forum's `compat.js` cause issues?

askvortsov1: If someone is importing via the new import scheme, this is a breaking change.

src/Api/Controller/ListUsersController.php
askvortsov1: Yeah, fair. My concern is inconsistency with other endpoints, and that the REST response is public API. That's why I prefer to have it in AdminPayload (https://github.com/flarum/core/blob/ea291508ab954a04682927ebe3bfafbfb584e752/src/Admin/Content/AdminPayload.php#L22-L22), as that's not considered public.

js/src/forum/compat.js
davwheat: The new scheme that wasn't announced? 😝 

You make a good point, I'll re-add it!

askvortsov1: Public relations fails aside, public API is public API :shrug: :hear_no_evil:

src/Api/Controller/ListUsersController.php
davwheat: Done. I opted to use the same API that `flarum/statistics` uses (`statistics.users.total`), because it's a core ext and will override the data set in `AdminPayload.php`, and provides one single source of truth, rather than duplicating data.

js/src/admin/components/UserListPage.tsx
```diff
@@ -0,0 +1,392 @@
...
+          } else {
+            return app.translator.trans('core.admin.user_list.grid.default_columns.group_badges_none');
+          }
+        },
+      },
+      70
+    );
+
+    columns.add(
+      'emailAddress',
```
davwheat, line 195: Toggling the visibility of an email, then switching page, leaves the original email in view rather than redrawing this component. I'll look for a fix for this.

davwheat:
I think this is ready for review, minus the issue I identified [above](https://github.com/flarum/core/pull/2626#discussion_r584001867).

tankerkiller125:
I forced the checks to re-run as they failed during the PHP selection part originally. It seems that their passing now.

askvortsov1:
> I forced the checks to re-run as they failed during the PHP selection part originally. It seems that their passing now.

Unfortunately they aren't

KyrneDev:
I don't think it's his fault. Tests on master are failing as well.

https://github.com/flarum/core/pull/2626/checks?check_run_id=2015700780

PHP issue?

js/src/admin/components/UserListPage.tsx
askvortsov1: Add a key based on user ID (https://mithril.js.org/keys.html), this will force a redraw when the key changes.

js/src/admin/components/UserListPage.tsx
```diff
@@ -0,0 +1,393 @@
...
+      {
+        name: app.translator.trans('core.admin.user_list.grid.default_columns.email'),
+        content: (user: User) => {
+          return (
+            <div class="UserList-email" data-user-id={user.id()} data-email-shown="false">
+              <span class="UserList-emailAddress">{app.translator.trans('core.admin.user_list.grid.default_columns.email_visibility_show')}</span>
+              <button
+                onclick={() => {
```
askvortsov1, line 204: This is a lot of logic. Can we move it out into a method?

js/src/admin/components/UserListPage.tsx
```diff
@@ -0,0 +1,393 @@
...
+      -100
+    );
+
+    return columns;
+  }
+
+  nextPage() {
+    this.isLoadingPage = true;
+    // Make sure the render doesn't end up happening after the API refresh
+    m.redraw.sync();
```
askvortsov1, line 307: Are these necessary based on testing? It's generally preferable to avoid this when we can.

js/src/admin/components/UserListPage.tsx
```diff
@@ -0,0 +1,393 @@
...
+
+    const newPageNum = this.pageNumber - 1;
+    this.loadNextUserSet(newPageNum * this.numPerPage, newPageNum);
+  }
+
+  /**
+   * Component to render.
+   */
+  content() {
```
askvortsov1, line 325: I think I'd prefer for this to be the very first method; we don't follow this perfectly, but so far there's a trend of:

- Mithril lifecycle hooks
- `view` (or equivalent)
- ItemList methods
- Functional methods

js/src/admin/components/UserListPage.tsx
```diff
@@ -0,0 +1,393 @@
...
+  /**
+   * Asynchronously fetch the next set of users to be rendered.
+   *
+   * Returns an array of Users, plus the raw API payload.
+   *
+   * Uses the `this.numPerPage` as the response limit.
+   *
+   * @param offset Offset to pass to API
+   */
+  async loadNextUserSet(offset: number = 0, pageNumber) {
```
askvortsov1, line 102: Couldn't we infer offset from pageNumber?

Also, `loadPage` feels like a better name

askvortsov1 (changes requested):
Unless a user is deleted (that would throw off pages), I think it might make sense to cache page datas so if someone is switching back and forth, we might not need to reload it.

Also, we might want to support sorting users by username.

I've also been thinking more about extraction, and I feel like having an "AbstractModelTable" would be very powerful: most of the code here (other than the columns, but those would be in each subclass) is applicable to any model.

js/src/admin/components/UserListPage.tsx
```diff
@@ -0,0 +1,393 @@
...
+          columns.map((col) => {
+            const columnContent = col.content && col.content(user);
+
+            return (
+              <div
+                class={classList(['UserListPage-grid--rowItem', rowIndex % 2 > 0 && 'UserListPage-grid--shadedRow'])}
+                data-user-id={user.id()}
+                data-column-name={col.itemName}
+              >
+                {columnContent || 'Invalid'}
```
askvortsov1, line 360: Invalid should be a translation

src/Admin/Content/AdminPayload.php
```diff
@@ -81,5 +82,18 @@ public function __invoke(Document $document, Request $request)
...
         $document->payload['mysqlVersion'] = $this->db->selectOne('select version() as version')->version;
+
+        /**
+         * Used in the admin user list. Implemented as this as it matches the API in flarum/statistics.
+         * If flarum/statistics ext is enabled, it will override this data with its own stats.
+         *
+         * This allows the front-end code to be simpler and use one single source of truth to pull the
+         * total user count from.
+         */
+        $document->payload['statistics'] = [
```
askvortsov1, line 93: modelStatistics might be better. This is private API so we can always change it later.

askvortsov1 (changes requested):

src/Admin/Content/AdminPayload.php
davwheat: I chose this because it matches with what's currently used in `flarum/statistics`. Should we change it in that extension too?

Ralkage:
Awesome work on this addition!

A little tidbit regarding the `core.admin.nav.userlist_button` translation; could we perhaps rename the translation value to just "Users" instead of "Users list"? Currently, the admin nav buttons have single-word names, and I believe "Users" would fight right in place with the other nav items in that aspect. The same should apply to the page header/title.

davwheat:
@Ralkage

Implemented your suggestions. I think it fits in to the nav a bit better now! :)

Do you think we should keep internal references as `user_list` to match with the existing `View user list` permission, or just swap all of the internal references (from this PR) to `users` as well?

Ralkage:
> Do you think we should keep internal references as `user_list` to match with the existing `View user list` permission, or just swap all of the internal references (from this PR) to `users` as well?

That's a good question 😅 both are good options, but I would seek input from the other core devs on which would be more practical. I like `users` as it would also match the current theme of the Admin area like how it is for the rest of the Admin nav buttons and their associated routes.

`permissions`  button -> `admin#/permissions`
`appearance` button -> `admin#/appearance`
`basics` button -> `admin#/basics`

askvortsov1:
> with the existing `View user list` permission

This permission badly needs to be renamed to `search users` (or something like that). I would lean towards renaming things to `users`.

Ralkage:
Could we also have the profile link icon/button left-aligned like the rest of the rows in the table? I also think that can be simplified further by allowing the usernames that are returned to be clickable and take said Admin to the user's profile that way.

Now for the "Edit User" column, I think renaming it to "Actions" would leave room for additional operations later down the road such as adding a "Delete" button. We could probably take it one step further to match the front end and simply name it "Controls".

image

I don't want us to add too much to this PR as @askvortsov1 already noted it's size as it is, but more so food for thought ;)

Ralkage:
There are so many ways to approach the way the table is presented, such as only adding icons if we want to keep it minimalistic (bless thee mobile users who have to scroll) 🙏

<details>
<summary>For example [click me pls]</summary>

 image

</details>

askvortsov1:
Looks like this needs a rebase; also, formatting is broken.

This was mentioned before, but I would like for the table component behind this to become some kind of `AbstractModelTable` component; it could then be reused for everything from API keys to dismissed flags to audit logging to GDPR erasure request records. The split could either be in this PR or as a follow-up PR, but I think that should be our plan.

js/src/admin/components/UserListPage.tsx
```diff
@@ -0,0 +1,378 @@
...
+    }
+
+    const columns: (ColumnData & { itemName: string })[] = this.columns().toArray();
+
+    return [
+      <p class="UserListPage-totalUsers">{app.translator.trans('core.admin.users.total_users', { count: this.userCount })}</p>,
+      <section
+        class={classList(['UserListPage-grid', this.isLoadingPage ? 'UserListPage-grid--loadingPage' : 'UserListPage-grid--loaded'])}
+        style={`grid-template-columns: repeat(${columns.length}, minmax(max-content, 300px))`}
+      >
```
SychO9, line 105: Right now the cells are all siblings under the same element, I would much prefer if we could wrap every row in its own element and apply grid on that, and also apply grid on the element containing the rows.

It would also be better to avoid inline styling and instead use a CSS variable, I know they are currently unused in our codebase, but I feel like that's not reason enough for us not to.

Basically having: `style={"--columns": columns.length}` and have the actual CSS in the proper files.

less/admin/UsersListPage.less
```diff
@@ -0,0 +1,118 @@
+.UserListPage {
+  // Pad bottom of page to make nav area look less squashed
+  margin-bottom: 24px;
```
SychO9, line 3: In most other pages where we need bottom spacing, we use `padding-bottom: 30px;` could we use that here for consistency ?

less/admin/UsersListPage.less
```diff
@@ -0,0 +1,118 @@
...
+
+  &-gridPagination {
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+    margin-top: 16px;
+
+    button .Button-icon {
+      margin-right: 0;
+    }
```
SychO9, line 70: let's remove this and use the `Button--icon` class on those arrow buttons instead.

js/src/admin/components/UserListPage.tsx
```diff
@@ -0,0 +1,378 @@
...
+      'id',
+      {
+        name: app.translator.trans('core.admin.users.grid.default_columns.id'),
+        content: (user: User) => user.id(),
+      },
+      100
+    );
+
+    columns.add(
+      'username',
```
SychO9, line 179: What about avatars 🤔 or is there a reason we wouldn't want them here

SychO9 (commented):
Great work! I like this a lot!

I left a few comments below about the code, but I also have a suggestion about the general design of the table, though only a suggestion because we all have different preferences design wise :P 

The only instance we've ever had a table UI component, is the notification preferences table, this is will be the second (unless I missed one), and imo I really think we should stick to the same design (which I do like), for the sake of consistency and so that we can at some point abstract it away. I think having buttons take up the whole cell (like the checkboxes in the notif prefs table) is also really neat, would work nicely in the three button we have here.

flarum lan_admin (23)

js/src/admin/components/UserListPage.tsx
davwheat: I just thought it'd make the rows too tall to easily go through data

less/admin/UsersListPage.less
davwheat: Sure, but I'd rather use a multiple of 8px to match the [widespread standard](https://spec.fm/specifics/8-pt-grid) used across the web, and [specifically recommended by Apple](https://developer.apple.com/design/human-interface-guidelines/ios/icons-and-images/image-size-and-resolution/#designing-high-resolution-artwork) to prevent odd blurring caused by scaling.

js/src/admin/components/UserListPage.tsx
davwheat: The issue is that then columns don't scale correctly as each row element is unaware of the next, and could result in mismatched column sizes.

I'd love to be able to do this, but until [CSS subgrid is supported in browsers](https://caniuse.com/css-subgrid), it's basically impossible.

RE CSS custom properties: I was under the impression that CSS custom properties had far worse support than they actually do (~95%). Let's use them!

js/src/admin/components/UserListPage.tsx
```diff
@@ -0,0 +1,378 @@
...
+              // Change button icons
+              emailToggleButtonIcon.addClass('fa-eye');
+              emailToggleButtonIcon.removeClass('fa-eye-slash');
+            }
+
+            emailContainer.attr('data-email-shown', !emailShown);
+          }
+
+          return (
+            <div class="UserList-email" key={user.id()} data-user-id={user.id()} data-email-shown="false">
```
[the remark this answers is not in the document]
davwheat: Changed! :)

js/src/admin/components/UserListPage.tsx
davwheat: Here's an example of what it looks like if we have cells in row elements.

image

SychO9: Yeah, I didn't think of that, we would be obliged to use fixed widths to make it happen, which isn't flexible, it's a shame subgrid would take years before it can be applied. Let's keep it as is for now then.

Btw what about accessibility ? At a first glance it seems suggested to use actual tables for presenting data, but I'm not well versed in this to know. I'm gonna research this later.

less/admin/UsersListPage.less
SychO9: That's so interesting! I've learned something new. Fine by me, though most of our codebase doesn't follow that standard, so I'm not sure if it would make a difference.

davwheat:
- Combined profile link and username columns
- Let the grid columns not always be 300px wide
- Rename `app.data.statistics` to `app.data.modelStatistics` to prevent possible incompatibilities with flarum/statistics.

https://user-images.githubusercontent.com/7406822/114958652-5f1ed200-9e5b-11eb-94b4-33ed6a54320d.mp4

locale/core.yml
```diff
@@ -217,6 +219,41 @@ core:
...
+
+          username: => core.ref.username
+          profile_link_tooltip: Visit {username}'s profile
+
+          join_time: Joined
+
+          group_badges: Groups
+          group_badges_none: None
+
+          email: Email
```
SychO9, line 242: `=> core.ref.email`

locale/core.yml
```diff
@@ -217,6 +219,41 @@ core:
...
+          email_visibility_show: Show email address
+          email_visibility_hide: Hide email address
+          # Fake email address which is blurred before being replaced by the real email
+          email_hidden: "abcdefg12345@gmail.com"
+
+          edit_user: => core.lib.edit_user.title
```
SychO9, line 248: If referencing another key, we should use `ref` instead, lib is for string that are called directly from either forum or admin frontends.

So let's create a `core.ref.edit_user.title` and change the rest accordingly.

js/src/admin/components/UserListPage.tsx
```diff
@@ -0,0 +1,383 @@
...
+              emailToggleButton.attr('title', extractText(app.translator.trans('core.admin.users.grid.default_columns.email_visibility_hide')));
+
+              // Replace placeholder email with real email
+              emailAddress.text(user.email());
+              emailAddress.removeAttr('aria-hidden');
+
+              // Change button icons
+              emailToggleButtonIcon.addClass('fa-eye');
+              emailToggleButtonIcon.removeClass('fa-eye-slash');
+            }
```
SychO9, line 281: The code is auto descriptive and well organised, I don't think these comments are necessary.

SychO9 (changes requested):
Any thoughts on whether we should consider staying consistent with the one existing table design 
 that I mentioned above ?

locale/core.yml
```diff
@@ -217,6 +219,41 @@ core:
...
+          join_time: Joined
+
+          group_badges: Groups
+          group_badges_none: None
+
+          email: Email
+          email_visibility_show: Show email address
+          email_visibility_hide: Hide email address
+          # Fake email address which is blurred before being replaced by the real email
+          email_hidden: "abcdefg12345@gmail.com"
```
askvortsov1, line 246: This is out of order, should be alphabetic. Also applies to other translations in this file.

locale/core.yml
```diff
@@ -217,6 +219,41 @@ core:
...
+    users:
+      title: => core.ref.users
+      description: A paginated list of all users on your forum.
+
+      total_users: "Total users: {count}"
+
+      grid:
+        invalid_column_content: Invalid
+
+        default_columns:
```
askvortsov1, line 231: Some of these are columns, others are default column values. Not sure how I feel about the organization.

askvortsov1 (changes requested):

locale/core.yml
Ralkage: Can we also replace `gmail.com` with `example.com`? I think It would be best to use something more generic instead of being brand specific to avoid any future headaches :)

locale/core.yml
```diff
@@ -220,40 +220,46 @@ core:
       upload_button: Choose an Image...
 
     users:
```
askvortsov1, line 222: This needs a comment on the line above for consistency

locale/core.yml
```diff
@@ -220,40 +220,46 @@ core:
...
-      title: => core.ref.users
       description: A paginated list of all users on your forum.
 
-      total_users: "Total users: {count}"
-
       grid:
         invalid_column_content: Invalid
```
askvortsov1, line 226: The alphabetization requirement is recursive all the way down, so a bit more rearrangement is needed.

askvortsov1 (changes requested):

locale/core.yml
davwheat: That line totally flew over my head 🙈

locale/core.yml
```diff
@@ -217,6 +219,48 @@ core:
...
+            title: => core.ref.username
+
+          join_time:
+            title: Joined
+
+          group_badges:
+            no_badges: None
+            title: Groups
+
+          email:
```
askvortsov1, line 242: `email` < `group_badges`, and `join_time` < `user_id`.

locale/core.yml
```diff
@@ -498,6 +527,20 @@ core:
     username:
       deleted_text: "[deleted]"
 
+    # These translations are used in the Edit User modal dialog (admin function).
```
askvortsov1, line 530: `edit_user < username`.

askvortsov1 (changes requested):

locale/core.yml
```diff
@@ -498,6 +527,20 @@ core:
...
+    edit_user:
```
askvortsov1, line 531: edit_user still < username

askvortsov1 (changes requested):

askvortsov1:
Also,

- Let's drop the placeholder value for emails, it's already blurred. Placeholder makes things messy because if the actual value is longer, the whole table shifts.
- Can we just unblur when clicking on the email? Then we can drop the button entirely.

davwheat:
> Can we just unblur when clicking on the email?

We discussed this, and I think we should keep the button. I can still make it unblur on click, if you'd like though.

askvortsov1 (approved):

less/admin/UsersListPage.less
```diff
@@ -0,0 +1,127 @@
...
+
+    &--header {
+      font-weight: bold;
+      border-bottom: 1px solid @muted-more-color;
+      padding: 8px 16px;
+      background: @control-bg;
+    }
+
+    &--rowItem {
+      padding: 4px 16px;
```
SychO9, line 52: The elements inside aren't well centered, let's add `display:flex; align-items:center;`

less/admin/UsersListPage.less
```diff
@@ -0,0 +1,127 @@
...
+
+    &IconBtn {
+      appearance: none;
+      border: none;
+      padding: 2px;
+      margin: 0;
+      margin-left: 12px;
+      background: transparent;
+      outline: none;
+    }
```
SychO9, line 125: I prefer using `Button Button--text` and dropping what can be dropped from these declarations.

js/src/admin/components/UserListPage.tsx
```diff
@@ -0,0 +1,384 @@
...
+
+    return [
+      <p class="UserListPage-totalUsers">{app.translator.trans('core.admin.users.total_users', { count: this.userCount })}</p>,
+      <section
+        class={classList(['UserListPage-grid', this.isLoadingPage ? 'UserListPage-grid--loadingPage' : 'UserListPage-grid--loaded'])}
+        style={{ '--columns': columns.length }}
+        role="table"
+        // +1 to account for header
+        aria-rowcount={this.pageData.length + 1}
+        aria-colcount={this.columns.length}
```
SychO9, line 108: `columns.length`*

SychO9 (changes requested):
So I've been thinking about this, and I think we should be using a `table` instead. While I do like `grid` it doesn't make much sense to use it here since we want a straightforward table, and it would be more semantically correct to use that, right now we're just trying to recreate a table with `grid` and the fact that the HTML code has little to no separation between the key elements of a tabular layout which is can also end up limiting customization ability.

We could leave it as is for now and change it when we tackle the Table component abstraction.

js/src/admin/components/UserListPage.tsx
```diff
@@ -0,0 +1,384 @@
...
+        aria-live="polite"
+        aria-busy={this.isLoadingPage ? 'true' : 'false'}
+      >
+        {/* Render columns */}
+        {columns.map((column, colIndex) => (
+          <div class="UserListPage-grid--header" role="columnheader" aria-colindex={colIndex + 1} aria-rowindex={1}>
```
SychO9, line 114: also, this should be `UserListPage-grid-header` since it's not a modifier but a descendant.

less/admin/UsersListPage.less
```diff
@@ -0,0 +1,127 @@
...
+    text-decoration-style: dotted;
+  }
+
+  &-editModalBtn {
+    position: absolute;
+    top: 0;
+    bottom: 0;
+    left: 0;
+    right: 0;
+    width: 100%;
```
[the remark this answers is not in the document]
davwheat: I was just looking in devtools, and it looks like we don't even need that!

```css
width: 100%;
border-radius: 0;
```

will do it all for us :tada:

image

davwheat:
> So I've been thinking about this, and I think we should be using a table instead.

I completely agree.

I brought this up when discussing it with Sasha on Discord and I suggested to leave it until we refactor the table into a separate component. For now this is only on the Admin page, so it's not the biggest deal, but we should look into it.

When subgrid comes, I propose we switch to that with table `role`s instead, but we need to use a table for now.

less/admin/UsersListPage.less
davwheat: I find this easier because I'd need to override the background and everything else from the default button, and it's not actually a textual button either (hence the modifier doesn't really make any sense). I initially tried with the background there, but it felt far too clunky and didn't look nice.

SychO9: I wasn't suggesting to change the design (I [realize](https://github.com/flarum/core/blob/43d6b3104d8ca7d4343aa864dd898e2af38f955a/less/common/Button.less#L153-L167) not everyone feels the same about it)

> because I'd need to override the background and everything else from the default button

That's the thing, you wouldn't, `Button--text` is a class for a style stripped button, you're already overriding things here, it would more consistent to use the built in generic classes we have for situations like these, and not have to override anything a second time, `Button Button-text` would preserve the same current look, while staying consistent, which is important for theming purposes 🤗

davwheat: Ohhh! I read it as `.Button-text` instead so assumed it was something else -- that's super helpful, actually!

SychO9 (approved):

davwheat:
ITS MERGE TIME! 🔥🔥

PR #2629, opened 2021-02-23
Recalculate enabled extensions and their dependencies if some listed in settings aren't installed

### LINKED ISSUE #1318: Allow optional dependency / one-of-several dependency requirements to control extension boot order

### LINKED ISSUE #2627: Enabling/disabling extensions sometimes causes crash
## Bug Report

**Current Behavior**
Introduced by #2579.
Whenever we enable/disable an extension, we now re-run dependency resolution. We do this by pulling in all enabled extension IDs (from settings), mapping that to the extension objects (via `getExtension()`), and passing that to the algorithm.

However, that enabled ID list is just a list stored in settings, and does not update when extensions are removed, so some entries passed to the algorithm might be null, which causes a crash.

**Steps to Reproduce**
1. Uninstall an enabled extension via composer
2. Try to enable/disable any other extension
3. It crashes.

**Possible Solution**
The way I see it, we have 2 options:

1. Add an extra `array_filter` in the `setEnabled` method so that extensions that don't exist aren't set to enabled. This prevents the bug, but it won't remove extensions from the enabled list on composer remove until some other extension is toggled
2. In `getExtensions`, check if any keys under `$enabledIds` don't correspond to an extension. If so, filter out the nonexistent ones and trigger `setEnabled`. This would close https://github.com/flarum/core/issues/1318 without the need for any composer scripts (a big positive) but would add extremely minimal overhead to boot. I think this is the better approach.

askvortsov1:
**Fixes #2627**
**Fixes #1318** 

On boot, if some extensions listed under "enabled" aren't installed, recalculate dependency resolution. This:

- Ensures issues don't occur if extensions have been removed via composer (or FTP!)
- Allows uninstalling an extension that breaks admin; otherwise, it would be enabled again when installed again, which is problematic

### DIFF
--- a/src/Extension/ExtensionManager.php
+++ b/src/Extension/ExtensionManager.php
@@ -115,6 +115,21 @@ class ExtensionManager
                 $extension->calculateDependencies($installedSet, $enabledIds);
             }
 
+            $needsReset = false;
+            $enabledExtensions = [];
+            foreach ($this->getEnabled() as $enabledKey) {
+                $extension = $extensions->get($enabledKey);
+                if (is_null($extension)) {
+                    $needsReset = true;
+                } else {
+                    $enabledExtensions[] = $extension;
+                }
+            }
+
+            if ($needsReset) {
+                $this->setEnabledExtensions($enabledExtensions);
+            }
+
             $this->extensions = $extensions->sortBy(function ($extension, $name) {
                 return $extension->getTitle();
             });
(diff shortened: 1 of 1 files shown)


SychO9 (approved):
Code looks good, tested locally.

tankerkiller125 (approved):

PR #2631, opened 2021-02-23
Fix keyboard on small mobile screens hiding composer

### LINKED ISSUE #2628: Composer is partially hidden on small mobile screens when keyboard is open
## Bug Report

**Current Behavior**

After #1670 was fixed, the composer is now fixed at the top of the screen on mobile. On small screens, this means that the keyboard could cover the composer and what you're trying to write.

This was also discussed in the PR (#2279), but I think the consequences were underestimated, because the new behaviour is a serious regression in the usability of the composer. Watch the video below to feel the frustration :)

**Steps to Reproduce**
1. Go to nightly.flarum.site with a "small" screen (16:9 is probably enough)
2. Attempt to write a long post
3. Observe that you can't see what you're typing

**Expected Behavior**

There should definitely be a way to see what you're typing. Previously, it was possible to scroll to move the composer.

**Screenshots**

This video was provided by a user of my forum, where I cherry-picked and deployed the commit 9b27b0d9d737931b8b7d93769c1e69bcc1ef6b6a.

https://user-images.githubusercontent.com/2164763/108773766-22bbad80-755f-11eb-92d3-b82a2c718ed1.mp4

**Environment**
- Flarum version: any version after 9b27b0d9d737931b8b7d93769c1e69bcc1ef6b6a

SychO9:
**Fixes #2628**

You can test this by going to nightly and adding/removing this custom CSS
```CSS
@media (max-width: 767px) {
  .Composer:not(.minimized) {
    height: 550px !important; /* This makes the composer long enough to emulate your mobile keyboard hiding a part of it (in case your screen is large enough for that not to happen by default) */
    max-height: 100%; /* The actual fix, try with and without */
  }
}
```

**Confirmed**

### DIFF
--- a/less/forum/Composer.less
+++ b/less/forum/Composer.less
@@ -117,6 +117,7 @@
       position: fixed;
       top: 0;
       height: 350px !important;
+      max-height: 100%;
       padding-top: @header-height-phone;
 
       &:before {


askvortsov1 (approved):

datitisev (approved):
It does look like this fixes the issue :+1:

askvortsov1:
@SychO9 I'm guessing https://github.com/flarum/core/issues/2628#issuecomment-786771268 would be a separate PR (different bug), so this is probably good to merge?

SychO9:
Yes, separate bug, we can fix that in a separate pull request.

PR #2633, opened 2021-02-23
Hide boot error

### LINKED ISSUE #2290: Boot exception exposes database username and file system path
When Flarum cannot connect to the database it shows the following error message:

```
Flarum encountered a boot error (Doctrine\DBAL\Driver\PDOException)
SQLSTATE[HY000] [1045] Access denied for user 'flarum'@'localhost' (using password: YES)
thrown in /[REDACTED]/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOConnection.php on line 31
```

Therefore exposing the database username and the file system path to Flarum.

An improvement could be to print the exception only in debug mode, or even better to log the exception with the logger so that users never see it. Both cases require to split the boot phase in multiple calls.

Another proposal (by me) is to simply rethrow the exception after printing a generic error message, so that PHP takes care of it.

This was discussed more in detail with @clarkwinkelmann [on Discuss](https://discuss.flarum.org/d/24851-boot-exception-exposes-db-username-and-file-system-path).

clarkwinkelmann:
**Fixes #2290**

 Completely redact boot error unless debug mode or display_errors is enabled. Attempt to use Flarum log file when possible.

Use HTTP error code 500 whenever a boot error occurs, whatever the handling method is.

**Screenshot**

If debug mode is enabled, most errors will look exactly as before:
image

If debug mode is disabled, most errors including database and extension related will end up showing this:
image

If an error happened even earlier, which is unlikely, the error will appear in one of two ways (example: the file `vendor/flarum/core/src/Foundation/Application.php` is missing).

If `display_errors` is `Off`:
image

If `display_errors` is `On`:
image

**Confirmed**

### DIFF
--- a/src/Http/Server.php
+++ b/src/Http/Server.php
@@ -85,7 +85,8 @@ class Server
             $line = $error->getLine();
             $type = get_class($error);
 
-            $this->printMessageAndExit(<<<ERROR
+            $this->printMessageAndExit(
+                <<<ERROR
             Flarum encountered a boot error ($type)<br />
             <b>$message</b><br />
             thrown in <b>$file</b> on line <b>$line</b>
--- a/src/Http/Server.php
+++ b/src/Http/Server.php
@@ -57,6 +57,9 @@ class Server
         try {
             return $this->site->bootApp()->getRequestHandler();
         } catch (Throwable $e) {
+            // Apply response code first so whatever happens, it's set before anything is printed
+            http_response_code(500);
+
             try {
                 $this->cleanBootExceptionLog($e);
             } catch (Throwable $e) {
@@ -85,15 +88,14 @@ class Server
             $line = $error->getLine();
             $type = get_class($error);
 
-            $this->printMessageAndExit(
-                <<<ERROR
+            echo <<<ERROR
             Flarum encountered a boot error ($type)<br />
             <b>$message</b><br />
             thrown in <b>$file</b> on line <b>$line</b>
 
 <pre>$error</pre>
-ERROR
-            );
+ERROR;
+            exit(1);
         } elseif (app()->has(LoggerInterface::class)) {
             // If the application booted far enough for the logger to be available, we will log the error there
             // Considering most boot errors are related to database or extensions, the logger should already be loaded
@@ -101,19 +103,9 @@ ERROR
             // then instantiate LogReporter through the container for automatic dependency injection
             app(LogReporter::class)->report($error);
 
-            $this->printMessageAndExit('Flarum encountered a boot error. Details have been logged to the Flarum log file.');
+            echo 'Flarum encountered a boot error. Details have been logged to the Flarum log file.';
+            exit(1);
         }
-    }
-
-    /**
-     * A simple and reliable way to stop the script and make sure the headers are set.
-     * @param string $message
-     */
-    private function printMessageAndExit(string $message)
-    {
-        http_response_code(500);
-        echo $message;
-        exit(1);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


clarkwinkelmann:
There's one thing you could give me input on: do you think the two special cases (debug mode and logger available) should be wrapped in a `try`/`catch` themselves? If the container is broken, `app()` could fail. There's also the possibility of an extension registering a binding to either of the abstracts resolved in this code, in which case the code could be faulty. This is very unlikely, but if this happens, there will be no way of accessing the original error.

`flarum.config` is probably safe, and I don't see why someone would bind an alias to `LogReporter`. I was considering manually instantiating `LogReporter` by passing the `LoggerInterface` so we know for sure our own code always runs here.

The remaining extensible aspect is if an extension binds to the `LoggerInterface`. But even if that happened, it would be very unlikely to have a boot error after that binding and also having the logger fail. If a custom logger fails, then the extension likely is very broken and wouldn't be fit for production.

askvortsov1:
> There's one thing you could give me input on: do you think the two special cases (debug mode and logger available) should be wrapped in a try/catch themselves? If the container is broken, app() could fail. There's also the possibility of an extension registering a binding to either of the abstracts resolved in this code, in which case the code could be faulty. This is very unlikely, but if this happens, there will be no way of accessing the original error.

As incredibly niche as this edge case is, I'm leaning towards handling it too, just so we don't need to worry about it ever again.

src/Http/Server.php
```diff
@@ -55,26 +57,70 @@ private function safelyBootAndGetHandler()
...
+        } elseif (app()->has(LoggerInterface::class)) {
+            // If the application booted far enough for the logger to be available, we will log the error there
+            // Considering most boot errors are related to database or extensions, the logger should already be loaded
+            // We check for LoggerInterface binding because it's a constructor dependency of LogReporter,
+            // then instantiate LogReporter through the container for automatic dependency injection
+            app(LogReporter::class)->report($error);
+
+            echo 'Flarum encountered a boot error. Details have been logged to the Flarum log file.';
+            exit(1);
+        }
```
askvortsov1, line 108: If none of these hold, should an exception be thrown so it goes into `fallbackBootExceptionLog`?

clarkwinkelmann:
It does seem a bit overkill now, but at least everything should be covered.

I also moved the code into methods that more accurately describe what is ideal and what is fallback.

src/Http/Server.php
```diff
@@ -55,26 +57,70 @@ private function safelyBootAndGetHandler()
...
+     * @param Throwable $error
+     * @throws Throwable
+     */
+    private function fallbackBootExceptionLog(Throwable $error)
+    {
+        echo 'Flarum encountered a boot error. Details have been logged to the system PHP log file.<br />';
+
+        // Throwing the exception ensures it will be visible with PHP display_errors=On
+        // but invisible if that feature is turned off
+        // PHP will also automatically choose a valid place to log it based on the system settings
```
askvortsov1, line 123: This last line of the comment feels redundant given the contents of the echo above

src/Http/Server.php
clarkwinkelmann: The idea was that no exit() would simply continue to the fallback. I hesitated adding a comment to explain it.

An additional thrown error seems excessive.

src/Http/Server.php
clarkwinkelmann: Sure, I could remove it

askvortsov1 (approved):

tankerkiller125 (approved):

PR #2634, opened 2021-02-24
Restrict who can use the lastSeenAt user sort

### LINKED ISSUE #2519: List user endpoint allows guessing user last time even when not disclosed
## Bug Report

**Current Behavior**
The list user API endpoint sorts by `lastSeenAt`, bypassing the `discloseOnline` user setting and ignoring `viewLastSeenAt` permission.
This doesn't expose the value, but with enough users on the forum with public online status, the index in the list leaks whether the user is online or roughly when they were last online.

**Steps to Reproduce**
1. Create 2 users.
2. Allow Guest to "View user list" (it could also be a user)
3. Disable `discloseOnline` on user 1
4. Connect to user 2 account
5. Connect to user 1 account
6. From Guest, fetch `/api/users?sort=-lastSeenAt`
7. Observe user 2 shows recently online, and user 1 above it, leaking that user 1 is online

**Expected Behavior**
The users with `discloseOnline` disabled should be placed at the bottom of the results, in no particular order.

If you have `viewLastSeenAt` permission, then you should be able to sort everyone.

**Screenshots**
Observed from User Directory extension:
image

**Environment**
- Flarum version: beta 15, but existed in previous versions as well

**Possible Solution**
This presents some technical challenges. Since the permission is inside a JSON payload, it will be resource-intensive to retrieve the value of the settings in the database request. Also default setting values not in the table need to be taken into account.

I don't think we have any sort attribute that has SQL wheres linked to it at this time. This will probably require changes to the way we define sortable properties. Probably something to link to the search refactor #1355

Another, maybe less resource intensive solution could be to use a second database column to store the "public" online status. That value can only be updated for users disclosing their status, and set to null when the setting is set to not disclose. (we can't use a single column, because users with `viewLastSeenAt` must still have access to everyone's values, ref #1797)

**Additional Context**

Possibly relevant code:

https://github.com/flarum/core/blob/f8edc2d827fee47dc78fbb75ee1d3b28580ed11d/src/Api/Controller/ListUsersController.php#L39

https://github.com/flarum/core/blob/0a6c5217c11e625aab5deec05bf6a6caf31bfa0d/src/Api/Serializer/UserSerializer.php#L32-L36

https://github.com/flarum/core/blob/0a6c5217c11e625aab5deec05bf6a6caf31bfa0d/src/Http/Middleware/AuthenticateWithSession.php#L39-L41

Reported by @DursunCanPoyraz in https://github.com/FriendsOfFlarum/user-directory/issues/55

clarkwinkelmann:
**Fixes #2519**

Now only users with `user.viewLastSeenAt` permission can use the `lastSeenAt` sort on `GET /api/users` endpoint.

Trying to use the sort when unauthorized will result in the standard error `{"errors":[{"status":"400","code":"invalid_parameter"}]}`

This is based on what was discussed in the issue. Even though we agreed on removing the sort completely, I noticed it wasn't too complicated to keep it only for users with a given permission. So I implemented the original suggestion from the first post of the issue.

There's just one question with this approach... Do we want the sort field present by default or absent by default? Since we alter it in `data`, it means `Extend\ApiController::addSortField`/`Extend\ApiController::removeSortField` has limited control over it. Question is, would an extension be more likely to want to disable it or enable it? Since it's already a mod-only feature I don't see much interest in disabling it, and since enabling would expose data I don't see that happening either :hankey: 

**Confirmed**

**Required changes:**

We need to warn about the "removed" sort in the upgrade guide.


askvortsov1 (commented):
I like the approach taken here. That being said, I'd like to see 2 integration tests showing that requesting this sort succeeds/fails depending on user permission.

src/Api/Controller/ListUsersController.php
```diff
@@ -69,6 +68,12 @@ protected function data(ServerRequestInterface $request, Document $document)
 
         $actor->assertCan('viewUserList');
 
+        if ($actor->hasPermission('user.viewLastSeenAt')) {
+            // At the moment, only people able to see everyone's last online date can sort by it. Ref #2519
+            // Otherwise this sort field would defeat the privacy setting discloseOnline
+            $this->addSortField('lastSeenAt');
+        }
```
clarkwinkelmann, line 75: This was my alternate implementation (together with re-adding `lastSeenAt` to `public $sortFields` above)

```suggestion
        if (!$actor->hasPermission('user.viewLastSeenAt')) {
            // If a user cannot see everyone's last online date, we prevent them from sorting by it
            // Ideally we want to use the sort, but it needs to take discloseOnline preference into account
            // We use remove instead of add so that extensions can still disable the sort using the extender
            $this->removeSortField('lastSeenAt');
        }
```

askvortsov1: Ah now I see what you mean. If we're shipping something in core, I lean towards having it on by default more than having it off, so I this method (in the list, alternatively removable) makes sense to me.

src/Api/Controller/ListUsersController.php
```diff
@@ -76,6 +76,15 @@ protected function data(ServerRequestInterface $request, Document $document)
...
         $actor->assertCan('viewUserList');
 
+        if (! $actor->hasPermission('user.viewLastSeenAt')) {
+            // If a user cannot see everyone's last online date, we prevent them from sorting by it
+            // Otherwise this sort field would defeat the privacy setting discloseOnline
+            // We use remove instead of add so that extensions can still completely disable the sort using the extender
+            $this->removeSortField('lastSeenAt');
+        }
+
+        $query = Arr::get($this->extractFilter($request), 'q');
```
clarkwinkelmann, line 86: You might have messed up the merge. I think this line has disappeared from master

askvortsov1 (approved):

SychO9 (approved):

Issue #2637, opened 2021-02-25
N+1 Queries


SychO9:
Somewhat related to #2177 

We currently have a problem with N+1 queries, some endpoints have certain relationships eager loaded, but not others, because they are either not needed by the frontend yet still called in certain serializers, or needed but just not eager loaded.

I have found these endpoints so far with this issue.

* [x] /api - Tags (flarum/tags#120)
* [x] /api/discussions?filter[q]=lorem
  + [x] `mostRelevantPost` (#2639)
  + [x] `tags` (https://github.com/flarum/tags/pull/125)
* [ ] /api/notifications
* [x] /api/posts?include=user
  + [x] `discussion` (https://github.com/flarum/core/pull/2717)
  + [x] `discussion.tags` (https://github.com/flarum/tags/pull/125)
  + [x] `user.groups` (https://github.com/flarum/core/pull/2717)
  + [x] `mentionedBy` (https://github.com/flarum/mentions/pull/64)

* [x] /api/flags
  + [x] `post.discussion.tags` (https://github.com/flarum/tags/pull/125)
  + [x] `post.user.groups` (https://github.com/flarum/flags/pull/38)

**Tags - ShowForumController (flarum/tags#120)**
Relations: `lastPostedDiscussion` (`lastPostedDiscussion.state`, `lastPostedDiscussion.tags`)

In https://github.com/flarum/tags/pull/120 I decided to eager load the two relations used to fix the problem, but looking at it again, we really shouldn't have to do that, the TagsPage component only needs [three things](https://github.com/flarum/tags/blob/046913b5841dd14d1f3eb9651fabc7ea7429365a/js/src/forum/components/TagsPage.js#L57-L68) from the lastPostedDiscussion: title, lastPostNumber, lastPostedAt. All of which are basic information.

Could we either add that data to BasicDiscussionSerializer and use that instead of DiscussionSerializer, or create a custom DiscussionSerializer in the tags extension, exposing only the information we need, that would be cleaner than eager loading relations just because the default serializer assumes they're available.

**Search - ListDiscussionsController (#2639)**
Relations: `mostRelevantPost` (`mostRelevantPost.discussion`, `mostRelevantPost.discussion.tags`, `mostRelevantPost.user.groups`)

This one should be rather simple, this is what we currently do for the `firstPost` and `lastPost` relations, we can just apply the same logic to the `mostRelevantPost` relation.
[...]

PR #2639, opened 2021-02-26
Eagerload some needed relations in ListDiscussionsController

SychO9:
**Part Of #2637**

Ensures some needed relations are present/eagerloded.
Can be tested by going to `/api/discussions?filter[q]=ipsum`

**Screenshot**
Screenshot from 2021-02-26 20-44-16
    ⬇️ 
Screenshot from 2021-02-26 20-47-54

**Confirmed**

### DIFF
--- a/src/Api/Controller/ListDiscussionsController.php
+++ b/src/Api/Controller/ListDiscussionsController.php
@@ -110,9 +110,13 @@ class ListDiscussionsController extends AbstractListController
 
         Discussion::setStateUser($actor);
 
+        if (in_array('mostRelevantPost.user', $include)) {
+            $include[] = 'mostRelevantPost.user.groups';
+        }
+
         $results = $results->getResults()->load($include);
 
-        if ($relations = array_intersect($include, ['firstPost', 'lastPost'])) {
+        if ($relations = array_intersect($include, ['firstPost', 'lastPost', 'mostRelevantPost'])) {
             foreach ($results as $discussion) {
                 foreach ($relations as $relation) {
                     if ($discussion->$relation) {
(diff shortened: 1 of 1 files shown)


src/Api/Controller/ListDiscussionsController.php
```diff
@@ -110,9 +110,13 @@ protected function data(ServerRequestInterface $request, Document $document)
 
         Discussion::setStateUser($actor);
 
+        if (in_array('mostRelevantPost.user', $include)) {
```
askvortsov1, line 113: I'd add a comment saying this is done for eager loading

askvortsov1 (changes requested):

src/Api/Controller/ListDiscussionsController.php
SychO9: Do we want to explain why we need to eager load it or just state that we're eager loading it ?

Observation: this is currently only necessary because the suspend extension's UserPolicy calls `isAdmin` on the `$user` object which loads all groups, in which case we should be eager loading the relation in the extension and not here, but I'm seeing in https://github.com/flarum/core/pull/2620/files#diff-5f5aff0db193a1b7b5bdff796982aefe84572973059c31c65f33233e7fafd597R34-R36 that we might end up doing it in core as well, so it will belong here after all (if that pr isn't changed).

askvortsov1 (approved):

src/Api/Controller/ListDiscussionsController.php
clarkwinkelmann: Should the comment reflect that? Suggestion:

    // Eager load groups for use in the policies (isAdmin check)

clarkwinkelmann (commented):
I'm going to try it locally. But here's a reply on the existing comments

clarkwinkelmann (approved):
Tested the earlier version locally. Final version looks good code-wise

PR #2645, opened 2021-03-01
Query Namespace

askvortsov1:
**Part of flarum/issue-archive#286 **

Move search results, search criteria, and ApplySearchParametersTrait to Query namespace.

They still extend (empty) search equivalents as a BC layer for extensions that expect those types.

A few changes I'm considering, but am unsure about:

- Moving FilterState to query. It's shared, but it's also an implementation detail. 
- Adding a QueryableInterface with a `query` method, and having both searchers and filterers implement it instead of `filter` and `search` methods. Leaning against this though because it isn't really necessary, and I don't see the benefits.

An actual extender for search drivers will have to wait until next release (or later), although that shouldn't be too difficult. Not a priority right now though.


SychO9:
> A few changes I'm considering, but am unsure about:
> Moving FilterState to query. It's shared, but it's also an implementation detail.

I agree, what about having it and `SearchState` extend an `AbstractQueryState` ? might be better organized, `FilterState` would end up an empty class, but I think that'd be fine.

> Adding a QueryableInterface with a query method, and having both searchers and filterers implement it instead of filter and search methods. Leaning against this though because it isn't really necessary, and I don't see the benefits.

Yea, We'd be overdoing it imo.

src/User/Search/UserSearcher.php
```diff
@@ -20,8 +20,8 @@
 use Illuminate\Database\Eloquent\Builder;
 
 /**
- * Takes a UserSearchCriteria object, performs a search using gambits,
- * and spits out a UserSearchResults object.
+ * Takes a UserQueryCriteria object, performs a search using gambits,
+ * and spits out a UserQueryResults object.
```
SychO9, line 24: Let's fix these

SychO9 (approved):

tankerkiller125 (approved):

Issue #2621, opened 2021-02-21
Search results cannot be scrolled fully on small mobile screens


matteocontrini:
## Bug Report

**Current Behavior**

When using the global search feature, it might happen that results cannot be scrolled fully if the screen is not tall enough.

**Steps to Reproduce**
1. Go to https://forum.fibra.click
2. Open the Dev Tools, enable the responsive mode, set a resolution like 360x400
3. Type "matteo" in the search box
4. Notice that you cannot scroll to the end of the results

**Expected Behavior**

The search results dropdown should not go beyond the border of the screen.

**Screenshots**

Provided by a user of my forum with a 16:9 screen:

![image](https://user-images.githubusercontent.com/2164763/108632647-58359d80-7470-11eb-8c32-1b50e82249be.png)

**Environment**
- Flarum version: 0.1.0-beta.15
- Website URL: https://forum.fibra.click
- Browser: Chrome, Firefox

askvortsov1:
Ah, I've experienced this before, and it was extremely irritating, but I forgot to file an issue. Yeah this is one of the UI bugs that needs to go

SychO9:
This is better fixed with flarum/issue-archive#348 imho

askvortsov1:
Fair, but that one's a bit controversial, since we need to settle on a design. I know @KyrneDev had some concerns about it, plus the placement of the nav bar at the top or bottom wasn't too clearly agreed upon. Either way, good that we have this in the system!

SychO9:
Well, I haven't fully taken a look at the conversation there, but it dates back to 2015-2016, we'll take a second look at it when we get to it and see.

askvortsov1:
Did some more investigation, the issue seems to be caused by links pushing down search. Its max height is set to 70vh which works fine it it's positioned where it would be by default, but not if there's extra content above (such as with fof links). Opening a PR to set max height dynamically, until we switch to a better solution with flarum/issue-archive#348

PR #2650, opened 2021-03-02
Fix search box out of screen

askvortsov1:
**Fixes https://github.com/flarum/core/issues/2621**

Set search results max height programatically instead of relying on 70vh, as elements like fof links might mess this up.

### DIFF
--- a/js/src/forum/components/Search.js
+++ b/js/src/forum/components/Search.js
@@ -116,7 +116,8 @@ export default class Search extends Component {
 
     if (this.state.getValue() && this.hasFocus) {
       const resultsElementMargin = 14;
-      const maxHeight = window.innerHeight - this.element.querySelector('.Search-input>.FormControl').getBoundingClientRect().bottom - resultsElementMargin;
+      const maxHeight =
+        window.innerHeight - this.element.querySelector('.Search-input>.FormControl').getBoundingClientRect().bottom - resultsElementMargin;
       this.element.querySelector('.Search-results').style['max-height'] = `${maxHeight}px`;
     }
   }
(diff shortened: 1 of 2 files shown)


js/src/forum/components/Search.js
```diff
@@ -113,6 +113,13 @@ export default class Search extends Component {
   onupdate() {
     // Highlight the item that is currently selected.
     this.setIndex(this.getCurrentNumericIndex());
+
+    if (this.state.getValue() && this.hasFocus) {
```
SychO9, line 117: can we add a comment about what it is for please

js/src/forum/components/Search.js
```diff
@@ -177,6 +177,13 @@ export default class Search extends Component {
           .one('mouseup', (e) => e.preventDefault())
           .select();
       });
+
+    // Since extensions might add elements above the search box on mobile,
+    // we need to calculate and set the max height dynamically.
+    const resultsElementMargin = 14;
+    const maxHeight =
+      window.innerHeight - this.element.querySelector('.Search-input>.FormControl').getBoundingClientRect().bottom - resultsElementMargin;
+    this.element.querySelector('.Search-results').style['max-height'] = `${maxHeight}px`;
```
SychO9, line 186: since we're adding it in `oncreate`, the max height is set early, but what happens when the keyboard shows up ? I'm assuming the value won't match then and so the issue will still persist, I can't test this but onupdate seems a better place.

SychO9 (approved):

clarkwinkelmann (approved):
Tested locally (with desktop tools) and seems to work well.

It does not react to `window.onresize` though, so being in `oncreate` or `onupdate` probably won't change anything if the browser resizes due to keyboard. It's however a good thing in case it needs updating while you type additional things.

PR #2657, opened 2021-03-03
Updated GroupFilterGambit to prevent hidden groups being visible wher…

### LINKED ISSUE #2559: Hidden Groups Visible in GroupGambit
## Bug Report

**Current Behavior**
When users searching by group name, GroupGambit will work even if the group in question is a hidden group.

**Steps to Reproduce**
1. Go to 'https://discuss.flarum.org'
2. Search `group:admin`
3. See admins

**Expected Behavior**
Hidden groups shouldn't be searchable

**Screenshots**

**Environment**
- Flarum version: beta 15

**Possible Solution**
`$groupQuery` in https://github.com/flarum/core/blob/4561f56fb9a7dfe8c6fd48e396e133f3aee7d91a/src/User/Search/Gambit/GroupGambit.php#L50-L58 shouldn't be using `orWhere` to search by name, as that voids the effects of `whereVisibleTo`.

ajaypayne:
…e they shouldn't be and to ensure that only the selected groups are returned on a search. #2559

**Fixes #2559 **

modify db query to use `and` instead of `or`

**Screenshot**
n/a

**Confirmed**


tests/integration/api/users/GroupSearchTest.php
```diff
@@ -0,0 +1,331 @@
...
+use Flarum\Tests\integration\RetrievesAuthorizedUsers;
+use Flarum\Tests\integration\TestCase;
+
+class GroupSearchTest extends TestCase
+{
+    use RetrievesAuthorizedUsers;
+
+    public function setUp(): void
+    {
+        parent::setUp(); // TODO: Change the autogenerated stub
```
askvortsov1, line 21: Why the todo here?

tests/integration/api/users/GroupSearchTest.php
```diff
@@ -0,0 +1,331 @@
...
+        $this->assertEquals(403, $response->getStatusCode());
+    }
+
+    /**
+     * @test
+     */
+    public function allows_group_filter_for_admin()
+    {
+        $response = $this->send(
+            $this->request('GET', '/api/users?filter%5Bq%5D=group%3Aadmin', [
```
askvortsov1, line 49: The get params should be removed from the URL

tests/integration/api/users/GroupSearchTest.php
```diff
@@ -0,0 +1,331 @@
...
+            'group_permission' => [
+                ['permission' => 'viewUserList', 'group_id' => 2],
+            ],
+        ]);
+
+        $response = $this->send(
+            $this->request('GET', '/api/users')
+                ->withQueryParams(['filter' => ['q' => 'group:admin']])
+        );
+        $responseBodyContents = json_decode($response->getBody()->getContents());
```
askvortsov1, line 90: Why are we decoding this, then immediately re-encoding it?

tests/integration/api/users/GroupSearchTest.php
```diff
@@ -0,0 +1,331 @@
...
+
+        $this->assertCount(0, $responseBodyContents->data, json_encode($responseBodyContents));
+        $this->assertObjectNotHasAttribute('included', $responseBodyContents, json_encode($responseBodyContents));
+    }
+
+    /**
+     * @test
+     */
+    public function admin_gets_correct_results_group_name_singular()
```
askvortsov1, line 212: These cases and the "non admin" cases could probably be condensed into "authorized_user_gets_correct_results". The hidden group test case is an important distinction and should be kept.

askvortsov1 (changes requested):
Thank you very much for this! The code looks good, I just have a few comments on test cases. It might also be good to test that multiple comma-separated groups can be included in the search.

tests/integration/api/users/GroupSearchTest.php
ajaypayne: Auto-generated from parent, but I will remove

tests/integration/api/users/GroupSearchTest.php
ajaypayne: It's decoded so that we can access the object properties, and then encoded for the message part of the assert so that if the test fails, you can see what was returned in the output.
I am happy to remove it if it isn't required though.

askvortsov1: Ah I see, I like that! We should do that more actually...

askvortsov1 (commented):
Any chance we could include a test case for multiple groups in the same query? IE:

'q' => 'group:admin,mod'

or even
'q' => 'group:admin,4'

ajaypayne:
Sorry, I missed that comment the first time around. Adding now.

askvortsov1 (approved):
Excellent, thank you so much!

tankerkiller125 (approved):

PR #2660, opened 2021-03-05
Use absolute positioning for the Composer on Safari

### LINKED ISSUE #2652: Composer goes off screen on iOS when starting to type
## Bug Report

**Current Behavior**

After #2279 and #2631 were applied, when replying to a discussion on iOS by tapping the "reply" area at the bottom of a discussion, it happens that when the user starts typing the composer goes off the screen, even if its position is fixed, and the user needs to scroll up.

**Steps to Reproduce**
1. Go to a forum on nightly with Safari on iOS
2. Open a discussion
3. Tap the "reply" area at the bottom
4. Tap the composer textbox and start writing
5. Sometimes: the page will scroll to the bottom and the composer will go off screen, forcing you to scroll up manually

**Expected Behavior**

The composer should be fixed at the top.

**Screenshots**

ezgif-com-video-to-gif

**Environment**
- Flarum version: beta.15 with changes from #2279 and #2631 applied
- Website URL: https://forum.fibra.click
- Browser: Safari on iOS. I don't have the versions.

I'll ask the iOS user reporting this to test on nightly.flarum.site to see if it might be an extension or something else on my forum causing this.

SychO9:
**Fixes #2652**

Use absolute positioning on Safari mobile, we can't use absolute positioning in all browsers because that causes other issues, however Safari mobile doesn't handle fixed positioning and inputs well either.

I don't really like this tbh, there is no reliable way to tell what browser is used, the user agent string might change in the future, and feature inference (testing for specific features unique to the browser) is also not reliable as features might also be removed in the future.

I would have preferred trying to fix this on the JS side by maybe checking the composer position isn't off the viewport or perhaps trying somethings with a timeout before focusing on the text editor, but not being able to reproduce and test the issue doesn't help.

If anyone else has access to an iOS device and able to reproduce this wants to try and find a better fix please feel free.

**Confirmed**

### DIFF
--- a/js/src/forum/components/Composer.js
+++ b/js/src/forum/components/Composer.js
@@ -265,7 +265,10 @@ export default class Composer extends Component {
     this.animateHeightChange().then(() => this.focus());
 
     if (app.screen() === 'phone') {
-      this.$().css('top', 0);
+      // On safari fixed position doesn't properly work on mobile,
+      // So we use absolute and set the top value.
+      // https://github.com/flarum/core/issues/2652
+      this.$().css('top', $('.App.mobile-safari').length ? $(window).scrollTop() : 0);
       this.showBackdrop();
     }
   }
--- a/js/src/forum/components/Composer.js
+++ b/js/src/forum/components/Composer.js
@@ -268,7 +268,7 @@ export default class Composer extends Component {
       // On safari fixed position doesn't properly work on mobile,
       // So we use absolute and set the top value.
       // https://github.com/flarum/core/issues/2652
-      this.$().css('top', $('.App.mobile-safari').length ? $(window).scrollTop() : 0);
+      this.$().css('top', $('.App').is('.mobile-safari') ? $(window).scrollTop() : 0);
       this.showBackdrop();
     }
   }
--- a/js/src/forum/utils/isSafariMobile.ts
+++ b/js/src/forum/utils/isSafariMobile.ts
@@ -5,9 +5,9 @@ export default function isSafariMobile(): boolean {
   return (
     'ontouchstart' in window &&
     navigator.vendor &&
-    navigator.vendor.indexOf('Apple') > -1 &&
+    navigator.vendor.includes('Apple') &&
     navigator.userAgent &&
-    navigator.userAgent.indexOf('CriOS') == -1 &&
-    navigator.userAgent.indexOf('FxiOS') == -1
+    !navigator.userAgent.includes('CriOS') &&
+    !navigator.userAgent.includes('FxiOS')
   );
 }
--- a/js/src/forum/utils/isSafariMobile.ts
+++ b/js/src/forum/utils/isSafariMobile.ts
@@ -1,0 +1,13 @@
+/**
+ * @see https://stackoverflow.com/a/31732310
+ */
+export default function isSafariMobile(): boolean {
+  return (
+    'ontouchstart' in window &&
+    navigator.vendor &&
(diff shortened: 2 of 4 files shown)


js/src/forum/utils/isSafariMobile.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * @see https://stackoverflow.com/a/31732310
+ */
+export default function isSafariMobile(): boolean {
+  return (
+    'ontouchstart' in window &&
+    navigator.vendor &&
+    navigator.vendor.indexOf('Apple') > -1 &&
```
askvortsov1, line 8: Can we use `includes` since we don't care about IE anymore?

less/forum/Composer.less
```diff
@@ -120,6 +120,12 @@
       max-height: 100%;
       padding-top: @header-height-phone;
 
+      // Fixes a bug where fixed position doesn't properly work in Safari mobile
+      // https://github.com/flarum/core/issues/2652
+      .mobile-safari & {
```
askvortsov1, line 125: Isn't this class on App, not Composer?

askvortsov1 (commented):
Interesting that this introduces a system for introducing workarounds to safari mobile. Hopefully we won't have to use it too much :fearful: 

I've gotten my hands on an old iphone, I'll test later today.

js/src/forum/utils/isSafariMobile.ts
SychO9: I need to start making it a habit of using includes instead.

less/forum/Composer.less
SychO9: yes and it is evaluated as such `.mobile-safari .Composer:not(.minimized)`

askvortsov1: Ah I got my ampesand positions confused

askvortsov1 (approved):
Confirmed that this fixes the issue. Code looks good to me :+1:

js/src/forum/components/Composer.js
```diff
@@ -265,7 +265,10 @@ export default class Composer extends Component {
     this.animateHeightChange().then(() => this.focus());
 
     if (app.screen() === 'phone') {
-      this.$().css('top', 0);
+      // On safari fixed position doesn't properly work on mobile,
+      // So we use absolute and set the top value.
+      // https://github.com/flarum/core/issues/2652
+      this.$().css('top', $('.App.mobile-safari').length ? $(window).scrollTop() : 0);
```
clarkwinkelmann, line 271: Would `$('.App').is('.mobile-safari')` be better syntactic sugar?

clarkwinkelmann (commented):
I don't have a Safari Mobile device to test.

I don't like the idea of having Safari-specific code in there because that means changing it will require having someone with Safari to check if it's still needed, but if there's no other solution so be it. Looks ok.

js/src/forum/components/Composer.js
SychO9: Yea that'd be better actually

SychO9:
I agree, I really don't like having browser specific code either, but there doesn't seem to be a solution for both safari & non-safari, seems like it's either this, or choosing one bug over the other. If we do go ahead with this solution, It'd be best not to use this util anywhere else unless really forced to such as this case.

clarkwinkelmann (approved):
I'm going to approve even though I have no way of testing :innocent:

matteocontrini:
I can confirm that this fixes the issue after having a friend test on nightly.

PR #2666, opened 2021-03-07
[A11Y] Fix nav drawer being focusable when off-screen on small viewports

### LINKED ISSUE #2565: [A11Y] App drawer on mobile is still part of accessibility tree when hidden
## Bug Report

**Current Behavior**
The app drawer on mobile is still part of the accessibility tree and can be tabbed into even when it's not shown.

**Steps to Reproduce**
1. Open Flarum and shrink browser viewport
2. Press tab to cycle through elements
3. Elements in the drawer (e.g. home link, user dropdown, etc) will be focused

**Expected Behavior**
Elements that are off-screen shouldn't be part of the accessibility tree, and shouldn't be able to be tabbed into.

**Environment**
- Flarum version: 0.1.0-beta.15
- Webserver: Apache
- Browser: Chrome 87

**Possible Solution**
`visibility: hidden` should be added to the `.App-drawer` when it's off-screen. This can be added to the styles as `.App:not(.drawerOpen) .App-drawer`.

The issue with this is that it breaks the drawer close animation. The way around this is to set `visibility: visible` when the user begins to close the drawer, then use an event handler in JS (`transitionend`, supported IE10+) to remove this inline style.

This also handles removing it from the accessibility tree as screen-readers ignore elements with `visibility: hidden`.

If you have any better ideas, please let me know.

davwheat:
Fixes flarum/framework#2565
Progresses towards completion of flarum/framework#3360 

- Implements `visibility: hidden` on mobile when the drawer is not open.
- Adds inline `visibility: visible` as drawer exits.
- Removes inline style to re-apply `visibility: hidden` once transition is complete.

In the GIFs below, keep an eye on the watched variables in the dev console to see the effect of the change.

### Before
LKOf0u

### After
TpPNeQ

**Confirmed**

### DIFF
--- a/js/src/common/utils/Drawer.js
+++ b/js/src/common/utils/Drawer.js
@@ -31,7 +31,18 @@ export default class Drawer {
    * @public
    */
   hide() {
-    $('#app').removeClass('drawerOpen');
+    const $app = $('#app');
+
+    // If the drawer isn't open, stop execution
+    if (!$app.hasClass('drawerOpen')) return;
+
+    const $drawer = $('#drawer');
+
+    // Used to prevent `visibility: hidden` from breaking the exit animation
+    $drawer.one('transitionend', () => $drawer.css('visibility', ''));
+    $drawer.css('visibility', 'visible');
+
+    $app.removeClass('drawerOpen');
 
     if (this.$backdrop) this.$backdrop.remove();
   }
--- a/js/src/common/utils/Drawer.js
+++ b/js/src/common/utils/Drawer.js
@@ -31,16 +31,22 @@ export default class Drawer {
    * @public
    */
   hide() {
+    /**
+     * As part of hiding the drawer, this function also ensures that the drawer
+     * correctly animates out, while ensuring it is not part of the navigation
+     * tree while off-screen.
+     * 
+     * More info: https://github.com/flarum/core/pull/2666#discussion_r595381014
+     */
+
     const $app = $('#app');
 
-    // If the drawer isn't open, stop execution
     if (!$app.hasClass('drawerOpen')) return;
 
     const $drawer = $('#drawer');
 
     // Used to prevent `visibility: hidden` from breaking the exit animation
-    $drawer.one('transitionend', () => $drawer.css('visibility', ''));
-    $drawer.css('visibility', 'visible');
+    $drawer.css('visibility', 'visible').one('transitionend', () => $drawer.css('visibility', ''));
 
     $app.removeClass('drawerOpen');
 
--- a/js/src/common/utils/Drawer.js
+++ b/js/src/common/utils/Drawer.js
@@ -35,7 +35,7 @@ export default class Drawer {
      * As part of hiding the drawer, this function also ensures that the drawer
      * correctly animates out, while ensuring it is not part of the navigation
      * tree while off-screen.
-     * 
+     *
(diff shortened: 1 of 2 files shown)


less/common/App.less
```diff
@@ -27,13 +27,16 @@
       .box-shadow(0 2px 6px @shadow-color);
     }
   }
-  .App-primaryControl, .App-titleControl, .App-backControl {
+  .App-primaryControl,
```
askvortsov1, line 30: Can we revert the style changes? On a related note, we should look into a formatter for less.

askvortsov1 (changes requested):

less/common/App.less
davwheat: Whoops! Fixed that. Also, yes we should :P

js/src/common/utils/Drawer.js
```diff
@@ -31,7 +31,18 @@ export default class Drawer {
...
+    const $app = $('#app');
+
+    // If the drawer isn't open, stop execution
+    if (!$app.hasClass('drawerOpen')) return;
+
+    const $drawer = $('#drawer');
+
+    // Used to prevent `visibility: hidden` from breaking the exit animation
+    $drawer.one('transitionend', () => $drawer.css('visibility', ''));
+    $drawer.css('visibility', 'visible');
```
askvortsov1, line 43: I see that this works, but why are we setting the visibility to visible?  If that's what the above comment refers to, maybe likes 42 and 43 should be swapped?

js/src/common/utils/Drawer.js
```diff
@@ -31,7 +31,18 @@ export default class Drawer {
    * @public
    */
   hide() {
-    $('#app').removeClass('drawerOpen');
+    const $app = $('#app');
+
+    // If the drawer isn't open, stop execution
```
askvortsov1, line 36: This comment is probably unnecessary.

Also, I think it'd be nice to have a comment at the top explaining the "why": This ensures that drawer elements aren't keyboard selectable while off screen.

askvortsov1 (commented):
Works well, left a few minor code comments

js/src/common/utils/Drawer.js
davwheat: As soon as the drawer starts leaving the screen, the `drawerOpen` class is removed, which causes the `visibility: hidden` to get applied from here:

https://github.com/flarum/core/pull/2666/files#diff-cb687979202880be4ab817b086176117563c352bf18c053b6c9e6370ede80a0eR126-R128

`visibility: hidden` means that the close animation will stop as it hides the element. So, before we remove the class, we set `visibility: visible` to override that style, then we use `ontransitionend` to remove this when the drawer is hidden.

askvortsov1: I understand what you mean here, but I think it would be more readable if it was set to visible first, and then the animation one callback applied. Or even better, if we chained them (which I believe is possible)

askvortsov1 (approved):
Excellent!

dsevillamartin:
I'm not comfortable enough with combining JS & CSS code for animations (I struggled a lot with Composer open & close animation in my mithril 2 branch) 🙈 

If it works and doesn't break other things (which it doesn't look to me like it would), I'm fine with it.

SychO9 (approved):
Looks good to me!

PR #2668, opened 2021-03-07
[A11Y] Add aria-label to dropdown toggles

### LINKED ISSUE #2581: [A11Y] Dropdown toggle buttons missing accessibility label
## Bug Report

**Current Behavior**
All dropdown toggle buttons (the ones with only carets) are missing accessible labels.

image

**Expected Behavior**
Accessibility labels should be provided on all interactive elements with purely visual explanations, like these. Preferably something similar to `aria-label="Show more options"`, with the text being part of language packs.

Without accessible labels, screen-readers will only hear "Button".

**Environment**
- Flarum version: 0.1.0-beta.15

davwheat:
**Fixes #2581**

- Adds a new `accessibleToggleLabel` attr for Dropdown and SplitDropdown components
- Implements this new attr on dropdowns used in core on the forum-side

We should integrate this new attr in core extensions, too.

**Screenshot**
d1QfBD

**Confirmed**

### DIFF
--- a/locale/core.yml
+++ b/locale/core.yml
@@ -269,7 +269,6 @@ core:
 
     # These translations are used by the discussion control buttons.
     discussion_controls:
-      toggle_dropdown_accessible_label: Toggle discussion actions dropdown menu
       cannot_reply_button: Can't Reply
       cannot_reply_text: You don't have permission to reply to this discussion.
       delete_button: => core.ref.delete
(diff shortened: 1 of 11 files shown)


locale/core.yml
```diff
@@ -269,6 +269,7 @@ core:
 
     # These translations are used by the discussion control buttons.
     discussion_controls:
```
askvortsov1, line 271: Translations should be ordered alphabetically by key

askvortsov1 (changes requested):

askvortsov1 (approved):

SychO9 (approved):

PR #2669, opened 2021-03-07
[A11Y] Add aria-label and landmark role to search input

### LINKED ISSUE #2667: [A11Y] Missing accessible label and landmark role on search input
## Bug Report

**Current Behavior**
There is no accessible label on the Search input, and the landmark `search` role is missing.

Placeholders should **never** be used in place of real landmarked elements, such as a `<label>`. This is because they prevent users with assistive technologies from selecting the input by voice control.

> While placeholder text provides valuable guidance for many users, **placeholder text is not a replacement for labels**. Assistive technologies, such as screen readers, do not treat placeholder text as labels.

**Possible Solution**
The W3 consortium recommend using hidden `<label>`s for inputs where we don't want to visibly display the label to the user, as these allow assistive tech to still use them.

> A label for a form control helps everyone better understand its purpose. In some cases, the purpose may be clear enough from the context when the content is rendered visually. The label can be hidden visually, though it still needs to be provided within the code to support other forms of presentation and interaction, such as for screen reader and speech input users.

**However**, in this scenario, using `aria-label` on the `<input>` is far easier, and also [suggested as a possible approach for hiding label text](https://www.w3.org/WAI/tutorials/forms/labels/#using-aria-label)

Source: https://www.w3.org/WAI/tutorials/forms/labels/#hidden

davwheat:
**Fixes #2667**

Adds the `aria-label` attribute to the search input.

Also adds `role="search"` to the containing div. See [this page](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Search_role) for more info.

For info about why this is required, see the related issue (#2667).

**Screenshot**
jAoNe3

**Confirmed**

### DIFF
--- a/js/src/forum/components/Search.js
+++ b/js/src/forum/components/Search.js
@@ -88,6 +88,7 @@ export default class Search extends Component {
             className="FormControl"
             type="search"
             placeholder={extractText(app.translator.trans('core.forum.header.search_placeholder'))}
+            aria-label={extractText(app.translator.trans('core.forum.header.search_placeholder'))}
             value={this.state.getValue()}
             oninput={(e) => this.state.setValue(e.target.value)}
             onfocus={() => (this.hasFocus = true)}
--- a/js/src/forum/components/Search.js
+++ b/js/src/forum/components/Search.js
@@ -73,6 +73,7 @@ export default class Search extends Component {
 
     return (
       <div
+        role="search"
         className={
           'Search ' +
           classList({
--- a/js/src/forum/components/Search.js
+++ b/js/src/forum/components/Search.js
@@ -71,6 +71,8 @@ export default class Search extends Component {
     // Hide the search view if no sources were loaded
     if (!this.sources.length) return <div></div>;
 
+    const searchCtaText = extractText(app.translator.trans('core.forum.header.search_placeholder'));
+
     return (
       <div
         role="search"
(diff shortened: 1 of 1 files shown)


js/src/forum/components/Search.js
```diff
@@ -88,6 +89,7 @@ export default class Search extends Component {
             className="FormControl"
             type="search"
             placeholder={extractText(app.translator.trans('core.forum.header.search_placeholder'))}
+            aria-label={extractText(app.translator.trans('core.forum.header.search_placeholder'))}
```
askvortsov1, line 92: Should we break this out into a variable? Also, should we consider putting all the aria stuff at the top?

davwheat: Good idea for both

js/src/forum/components/Search.js
```diff
@@ -71,8 +71,11 @@ export default class Search extends Component {
     // Hide the search view if no sources were loaded
     if (!this.sources.length) return <div></div>;
 
+    const searchCtaText = extractText(app.translator.trans('core.forum.header.search_placeholder'));
```
askvortsov1, line 74: What is cta? Can we just say label or placeholder?

askvortsov1 (changes requested):

js/src/forum/components/Search.js
davwheat: CTA means call to action.

I would have done that but it's a label and a placeholder, rather than one or the other.

askvortsov1: In other parts of the codebase, we use `label` to describe placeholders. Let's use label: even if not as descriptive, it's clearer.

davwheat: Done! :)

askvortsov1 (approved):

davwheat:
Fingers crossed the bundle size status works! 🤞

SychO9 (approved):

PR #2695, opened 2021-03-15
Add bundlewatch to track bundle size changes in PRs

davwheat:
Add PR status to show the change in bundle size as a result of it. This CI workflow must run on master before it begins working.

It's also possible to define a max size for each of the JS files which will cause a failing status if the bundle is over that size, but I decided not to implement this.

For more configuration options, see: https://bundlewatch.io/#/reference/configuration

Example PR: https://github.com/davwheat/flarum-core/pull/8

**Setup steps:**
1. Head here to get auth ID: https://service.bundlewatch.io/setup-github
2. Create repo secret called `BUNDLEWATCH_GITHUB_TOKEN` with the token inside

**Screenshot**
image

### DIFF
--- a/.github/workflows/pr_size_change.yml
+++ b/.github/workflows/pr_size_change.yml
@@ -1,0 +1,40 @@
+name: Bundle size checker
+
+on:
+  workflow_dispatch:
+
+  push:
+    paths:
+      - "js/**"
+  pull_request:
+    paths:
+      - "js/**"
+
+jobs:
+  bundlewatch:
+    runs-on: ubuntu-latest
+
+    name: Bundlewatch
+
+    steps:
+      - uses: actions/checkout@master
+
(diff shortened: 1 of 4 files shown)


askvortsov1 (approved):

datitisev:
What if we merged this action with the JS build? That way we could have the action be called `JavaScript / Bundlewatch` (unsure if you can have different actions with the same main name 🤔)

davwheat:
@datitisev The main reason I opted not to is that we only run that workflow on pushes to `master`, and not for PRs and more. We don't want the bot committing dist JS onto PRs as that could create all sorts of merge conflicts.

tankerkiller125 (approved):

davwheat:
@datitisev Any objections before I merge?

PR #2697, opened 2021-03-16
WIP: Use CSS tooltips instead of Bootstrap

### LINKED ISSUE #131: Replace tooltip library with CSS tooltips and helper component

### LINKED ISSUE #144: max-width of 200px on tooltip element

### LINKED ISSUE #2696: Replace tooltip library with CSS tooltips and helper component
Progress towards #2345.

Some discussion has already taken place about this in #2345.

Using JS tooltips where not needed isn't a good idea as DOM manipulation takes far more time for devices to handle and paint than just a CSS transform. It also means that we don't need to worry about "creating" the tooltips.

Let's replace JS-based tooltips with better CSS-based tooltips! :tada:

My API suggestion for this would be something like this:

```html
<button aria-label="This will delete your forum!" data-tooltip>
  My cool button
</button>
```

Using [CSS's `attr()`](https://caniuse.com/mdn-css_types_attr), we can use a pseudoelement to show a tooltip:

```less
// This isn't the full set of styles needed, obviously!
[data-tooltip][aria-label]:not([aria-label=""]) {
  &::before {
    content: attr(aria-label);
  }
}
```

We should also create a Tooltip helper component which we can wrap content in to provide tooltips when pseudoelements are already used on the element itself.

This is a great starting point for our CSS: https://kazzkiq.github.io/balloon.css/

We currently use tooltips throughout core.

There are MANY JS-based tooltip libraries available, but when things can be done with CSS alone, it's always a better option. CSS parses and paints much faster than JS, and has much better and easier support for animations.

davwheat:
**Fixes flarum/issue-archive#131. Progresses flarum/issue-archive#179. Fixes flarum/issue-archive#144**

This PR replaces Bootstrap tooltips with pure CSS tooltips. 🎉 

Despite this not implementing any extension-breaking API changes, some exts will need to modify their styling to remove uses of `overflow: hidden` to correctly show these CSS-only tooltips.

Are there any issues with tooltip usage in core? **Issues would specifically relate to the use of `overflow: hidden` with tooltips-wielding nodes inside.**

**Known issues:**
- [x] Composer toolbar suffers from ["the problematic overflow spec"](https://css-tricks.com/popping-hidden-overflow/#the-problem)

**Screenshot**
image
image
image

**Confirmed**


less/common/Tooltip.less
```diff
@@ -1,72 +1,209 @@
...
-  opacity: 0;
-  .transition(0.15s opacity linear);
-
-  &.in     { opacity: 1; }
-  &.top    { margin-top:  -3px; padding: @tooltip-arrow-width 0; }
-  &.right  { margin-left:  3px; padding: 0 @tooltip-arrow-width; }
-  &.bottom { margin-top:   3px; padding: @tooltip-arrow-width 0; }
-  &.left   { margin-left: -3px; padding: 0 @tooltip-arrow-width; }
+// This fixes an issue in iOS where `:hover` is not handled
+[data-device-os="iOS"] {
```
askvortsov1, line 6: We already have a tool for checking mobile safari, I don't see a reason to include the generic getOperatingSystem script beyond that. Plus a more generic tool opens the possibility of special cases, which would be preferable to avoid when possible

askvortsov1 (changes requested):

less/common/Tooltip.less
davwheat: Except the mobile Safari tool only checks for mobile Safari, not other "browsers" which just embed Safari within their own UI.

If I only used the mobile Safari check, then Chrome for iOS, Firefox for iOS, etc would break.

askvortsov1: Then we need to modify the mobile safari check, as issues with mobile safari will also occur on those other browsers. That being said, have we confirmed it doesn't hold for chrome ios?

SychO9: When I researched about that specific iOS issue with absolute positioning, I found out it was mobile safari only, but I obviously haven't confirmed that.

That said, seeing how we only apply a `cursor: pointer` here, and it's the only place we use an iOS check, couldn't we just drop the iOS check and apply it for all devices, since I don't think it'll affect other devices.

davwheat: But then we'd be left with weird pointers all over the place, which would look weird, like this: 

image

SychO9: Ah I see, this seems related to https://github.com/flarum/core/pull/2607/files I think we might be able to directly use the same media query ? (can't test..)

davwheat: Hmmm maybe that would work, yeah. I'll give it a go, but I have no iOS devices to test with.

SychO9: I'll try with https://app.lambdatest.com/ when possible and confirm, I've used it before, it's not bad

davwheat:
I've not found any good solution around CSS tooltips in the composer toolbar, sadly.

Setting `overflow-y: visible; overflow-x: auto` isn't possible because the CSS spec is a pain.

The one thing I can think of is setting a taller height, so that the tooltips don't overflow, and hiding this extra height with CSS.

askvortsov1:
> I've not found any good solution around CSS tooltips in the composer toolbar, sadly.

What's the issue with them now?

davwheat:
> What's the issue with them now?

https://u.davwheat.dev/DOB9fw.gif

askvortsov1:
> > What's the issue with them now?
> 
> 

Interestingly enough, it works flawlessly when I have my rich text editor extension enabled, whether or not the editor itself is used; I think I fixed this via custom less. Applicable block: https://github.com/askvortsov1/flarum-rich-text/blob/e7c77b535f82ffcba2798bfc5d03d2920cb7cbf6/resources/less/forum.less#L1-L1 (this has some extra stuff though)

davwheat:
> EDIT: setting overflow-x: initial on TextEditor-controls fixes it

Does that still allow horizontal scrolling with loaddds of items in the toolbar (can't check rn, but will later if you can't)

askvortsov1:
> > EDIT: setting overflow-x: initial on TextEditor-controls fixes it
> 
> ~Does that still allow horizontal scrolling with loaddds of items in the toolbar (can't check rn, but will later if you can't)~
> 
> This brings back another issue that breaks the toolbar if it has too many items (or you shrink the viewport), it seems:
> 
> image

Unfortunately that's true. That being said, I'd argue that many items is very poor UX in the first place (simplicity is one of Flarum's core values). Collapsing additional ones in a dropdown (like rich text does) works well.

davwheat:
Even a forum that uses fof upload and another ext would need the scrolling.



This is from the giffgaff preview community and it requires scrolling on mobile to see the spoiler button.

I don't think we should remove scrolling to add this feature, unfortunately. I'll take a look at some other solutions tomorrow.

davwheat:
I fixed the Composer toolbar issue using some hacky CSS. For more info, see the commit message: https://github.com/flarum/core/pull/2697/commits/5fed429c2ea4f7c13d9b990701ef5afe503c7645.

From the end of the message, that needs discussion:

Unfortunately this hack can't fix the tooltips on elements close to the edges of the screen -- there's no way to set a minimum-left in CSS. Should we maybe consider *some* JS to fix this, or live with it instead? See: https://medium.com/carwow-product-engineering/building-a-simple-tooltip-component-that-never-goes-off-screen-c7039dcab5f9

What's your opinions on this @askvortsov1 @SychO9?

image

less/forum/Composer.less
```diff
@@ -10,94 +10,107 @@
...
-  right: 0;
-  bottom: 0;
-  background: fade(@body-bg, 90%);
-  display: none;
-  border-radius: @border-radius @border-radius 0 0;
-
-  &.active {
-    display: block;
+
+  &-loading {
```
askvortsov1, line 95: Could we revert some of the stylistic changes? It's making it a bit difficult to see what has and hasn't changed

askvortsov1 (changes requested):
I don't mind the hack that much, but going off-screen is problematic IMO. Would JS even be able to fix this if we're using pseudoelements?

If we need a placement library, https://github.com/tobyzerner/placement.js could be a candidate

askvortsov1:
> > @askvortsov1 Would JS even be able to fix this if we're using pseudoelements?
> 
> My plan was to create an off-screen element with the tooltip text, measure the width, then adjust the `transform()` as needed, but I might explore other possibilities too.

Doesn't this mean we've just ended up with regular tooltips as a fallback to css tooltips? Maybe it might be easier to just stick with traditional tooltips for now, and try again with CSS sometime later when browser support is better?

davwheat:
My idea was to set up a Mutation Observer to watch for tooltips to be created (or modified), then perform the calculations in advance.

---

There actually might be no need for all that extra element rubbish! JS has the ability to get the size of pseudoelements natively: https://stackoverflow.com/a/44912365/11091039

Cross-browser support is unknown, but it's in IE9, so it must be in other browsers, surely: https://caniuse.com/mdn-api_window_getcomputedstyle_pseudo-element_support

image

askvortsov1:
> My idea was to set up a Mutation Observer to watch for tooltips to be created (or modified), then perform the calculations in advance.
> 
> There actually might be no need for all that extra element rubbish! JS has the ability to get the size of pseudoelements natively: https://stackoverflow.com/a/44912365/11091039
> 
> Cross-browser support is unknown, but it's in IE9, so it must be in other browsers, surely: https://caniuse.com/mdn-api_window_getcomputedstyle_pseudo-element_support
> 
> image

That makes sense to me!

davwheat:
Just going to chuck this in here so I don't lose it: https://stackoverflow.com/a/39332340

SychO9:
Whoa there's a lot going on here, thanks for the hard work 🤗 

I like replacing JS with CSS where possible but I have mixed feelings about this change, having the tooltip be a direct descendant of the element we're tooltiping is something I am *uncomfortable with*/*uncertain about* since the only scenarios we're faced with are the default ones. I see we've added a hack to fix the overflow issue where it occurs in core, but an extension facing the same situation will have to do something similar, in addition to potential scenarios where z-indexes might cause trouble.

I don't have any concrete examples though which is not helpful to anyone but I wanted to voice my concern nonetheless.
Because I am uncertain about this 😕 I will refrain from reviewing.

askvortsov1:
> Because I am uncertain about this 😕 I will refrain from reviewing.

On the contrary, I think this is exactly a reason to review. I also share some of these concerns. This is all quite cool, but it's also a huge PR with a lot of edge cases and workarounds. Having CSS only tool tips would be ideal, but if it's not possible without this much modification and hacks, I would prefer to wait. Our goal currently is to:

- get rid of bootstrap stuff (possibly replaced with a smaller library)
- provide an abstract API so we can change how tool tips are implemented in flarum without breaking use in extensions.

I don't think I'm comfortable with introducing so many workarounds, especially directly in flarum source code. Also, if we need so many workarounds, is it really CSS-only tool tips? I know you've put a huge amount of time and work into this, but I'm not sure we should proceed with this solution right now (maybe in some time browser support will be better?).

davwheat:
The issue is that there are no issues with browser support at all.

There's only one workaround in the code, and that's for the composer.

The MutationObserver is purely to prevent them overflowing. From preliminary tests, it _seems_ to still be faster than Bootstrap tooltips to run because there's no reacting to events (hover) needed.

JS and checking rects, etc, seems to be the "standard way" that people work around this problem. We only need to use a MutationObserver because Flarum is very interactive and is an SPA.

> "Prevent CSS tooltip from going out of page/window - Stack Overflow" https://stackoverflow.com/questions/50970336/prevent-css-tooltip-from-going-out-of-page-window/51066294#51066294

askvortsov1:
How much of this could we put in a separate library? Since this is all quite non-flarum-specific, I feel like that would be a cleaner way to organize this.

We're saving 2kB so size-wise it's still a win. I'm just not sure we need this directly in the Flarum codebase.

PR #2698, opened 2021-03-16
Fix searching titles in discussions

askvortsov1:
This was originally attempted in https://github.com/flarum/core/pull/1741, but reverted in https://github.com/flarum/core/commit/9f6ec804324d0d7dfa5f378482f855a7bbdc3151 due to performance issues. From pure speculation, this should not cause such issues. From local dev environment testing (on a small dataset), that claim holds.

**Implementation:**
Instead of using a left join to include discussions, we add an additional subquery where we search discussions by title, union that into the posts subquery, and only then join. 

- This MUST be tested for performance on a large site before we consider merging it.

**Confirmed**

### DIFF
--- a/src/Discussion/Search/Gambit/FulltextGambit.php
+++ b/src/Discussion/Search/Gambit/FulltextGambit.php
@@ -9,6 +9,7 @@
 
 namespace Flarum\Discussion\Search\Gambit;
 
+use Flarum\Discussion\Discussion;
 use Flarum\Post\Post;
 use Flarum\Search\GambitInterface;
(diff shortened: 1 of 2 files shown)


tankerkiller125 (approved):
This has been running on https://discuss.flarum.org for about a week now with excellent results.

There has been no noticeable changes to resources used on the VPS and the search seems to be just as fast as it was without title searching.

SychO9 (approved):
Tested with a 700K discussions database, the search time was the same before and after.

Based on that, I'm approving this.

PR #2703, opened 2021-03-17
User::fromRequest method

### LINKED ISSUE #869: Switch to a custom subclass of ServerRequest

askvortsov1:
**Fixes #869**

Add a `fromRequest()` static method to `User`, use this instead of `$request->getAttribute('actor')`

Do we want to do this for other Flarum-specific stuff? I don't think it's worth it (at least for now), the only thing we really need this for is encapsulating the current user in the request.

**Confirmed**

### DIFF
--- a/src/User/User.php
+++ b/src/User/User.php
@@ -880,9 +880,9 @@ class User extends AbstractModel
 
     /**
      * Set a user as the actor on a request instance.
-     * 
+     *
      * @param Request $request
-     * @return 
+     * @return
      */
     public static function setRequestActor(self $actor, Request $request): Request
     {
--- a/src/User/User.php
+++ b/src/User/User.php
@@ -879,6 +879,17 @@ class User extends AbstractModel
     }
 
     /**
+     * Set a user as the actor on a request instance.
+     * 
+     * @param Request $request
+     * @return 
+     */
+    public static function setRequestActor(self $actor, Request $request): Request
+    {
+        return $request->withAttribute('actor', $actor);
+    }
+
+    /**
      * Get the current user from a request instance.
      *
      * @param Request $request
--- a/src/User/User.php
+++ b/src/User/User.php
@@ -882,9 +882,9 @@ class User extends AbstractModel
      * Get the current user from a request instance.
      *
      * @param Request $request
-     * @return static
+     * @return self
      */
-    public static function fromRequest(Request $request)
+    public static function fromRequest(Request $request): self
     {
         return $request->getAttribute('actor');
     }
--- a/src/User/User.php
+++ b/src/User/User.php
@@ -897,6 +897,6 @@ class User extends AbstractModel
      */
     public static function fromRequest(Request $request): self
     {
-        return $request->getAttribute('actor');
+        return $request->getAttribute('actor', new Guest());
     }
 }
--- a/src/User/User.php
+++ b/src/User/User.php
@@ -35,6 +35,7 @@ use Flarum\User\Exception\PermissionDeniedException;
 use Illuminate\Contracts\Hashing\Hasher;
 use Illuminate\Contracts\Session\Session;
 use Illuminate\Support\Arr;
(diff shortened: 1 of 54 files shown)


src/User/User.php
```diff
@@ -877,4 +878,15 @@ public function refreshDiscussionCount()
...
         return $this;
     }
+
+    /**
+     * Get the current user from a request instance.
+     *
+     * @param Request $request
+     * @return static
+     */
+    public static function fromRequest(Request $request)
```
SychO9, line 888: Let's type hint the return value

SychO9 (changes requested):

SychO9:
is it normal for the test changes to appear here 🤔 ?

askvortsov1:
> is it normal for the test changes to appear here  ?

Fixed, not sure how that happened.

SychO9 (approved):

askvortsov1:
Should we also include a static method for setting the actor on a request?

SychO9:
> Should we also include a static method for setting the actor on a request?

Oh yeah, we should also use it in the codebase.

askvortsov1:
Done. It feels a bit weird to use a static User method when we have a user instance, but I don't think making it non-static would be cleaner. 

```
$request = User::setRequestActor($actor, $request); // Should we flip argument order?
```
vs
```
$request = $actor->setRequestActor($request);
```

I feel like the latter is "more correct", but the naming of `setRequestActor` feels like it should be setting something on the $actor object. Maybe we should go with the second, but call the method `requestWithActor`? `requestWithSelfAsActor` is too verbose IMO.

src/User/User.php
```diff
@@ -876,4 +877,26 @@ public function refreshDiscussionCount()
...
         return $this;
     }
+
+    /**
+     * Set a user as the actor on a request instance.
+     *
+     * @param Request $request
+     * @return
+     */
+    public static function setRequestActor(self $actor, Request $request): Request
```
SychO9, line 887: nitpick: dockblock isn't accurate

askvortsov1: I'll fix it when we settle on naming / structure; any thoughts on that?

SychO9: Oh I didn't see that, my bad

SychO9:
> Done. It feels a bit weird to use a static User method when we have a user instance, but I don't think making it non-static would be cleaner.
> 
> ...
> 
> I feel like the latter is "more correct", but the naming of `setRequestActor` feels like it should be setting something on the $actor object. Maybe we should go with the second, but call the method `requestWithActor`? `requestWithSelfAsActor` is too verbose IMO.

`$actor->requestWithActor($request)` looks good, though I'm reconsidering if having a `RequestUtil` would be better and more straightforward, since we're doing these operations on a request instance, and now that we're adding this second method, it feels off.

Even if it only has `getActor` and `withActor`, what do you think if we went back to that instead ?

askvortsov1:
> $actor->requestWithActor($request) looks good, though I'm reconsidering if having a RequestUtil would be better and more straightforward, since we're doing these operations on a request instance, and now that we're adding this second method, it feels off.

Even if it only has getActor and withActor, what do you think if we went back to that instead ?

Agreed, even though I don't like introducing a new "util" class, I don't think adding itself to a request should be a responsibility of the User model (which already does a LOT). Any thoughts @flarum/core? I'd prefer to agree on an approach before changing ~100 instances of $request->getActor()`.

tankerkiller125:
What's the exact reason for making this change again? I don't know if maybe I was zoned out when it was talked about or maybe I was talking to a co-worker at the time.

luceos:
Now that I see this code I am not sure I like it. This approach also goes against our use other conventions. I'm increasingly leaning towards creating an Actor interface (applied to User) and then binding the authenticated actor into ioc from the middleware. Thoughts?

PR #2704, opened 2021-03-17
Include task scheduling in core

### LINKED ISSUE #2281: Integrate task scheduling into core
## Feature Request

Scheduling console commands currently depends on fof\console. Since that extension is essentially just a util library for Flarum extensions, it might make sense to make it a core feature (with a dedicated extender). Then, we could also provide fuller documentation for it in our documentation.

An extender for scheduling commands.

This is a feature whose only purpose is enabling extensions to do more.

Not doing this and leaving it in `fof/console`.

askvortsov1:
**Fixes #2281**

Add a `schedule` method to the `Console` extender.

- Should we add methods for each way to schedule? There's a LOT of options, so I think what I have here is simpler. On the other hand it's slightly less BC compatible, in case laravel decides to break something on its side. But I still think its way too many methods, and figuring out how to chain them is complicated.
- Turns out we need the helpers we literally just removed for this.... Given this it might make sense to add them back, but still discourage use just like we do now. Should we do this in the same PR?

**Confirmed**


composer.json
```diff
@@ -25,10 +25,13 @@
         "components/font-awesome": "^5.14.0",
         "dflydev/fig-cookies": "^3.0.0",
         "doctrine/dbal": "^2.7",
+        "dragonmantank/cron-expression": "^3.1.0",
+        "flarum/laravel-helpers": "^0.1.0",
```
tankerkiller125, line 29: If we're including this in core why wouldn't we just reverse the change that made this package needed? Am I missing something about that whole split?

askvortsov1: > Given this it might make sense to add them back, but still discourage use just like we do now. Should we do this in the same PR?

That's what I was thinking here. I'd vote for bringing them all back, not including the package, and "perpetually deprecating" them, as mentioned in https://github.com/flarum/core/pull/2705#issue-594956355

tankerkiller125: Yeah I'd also vote to do that then. If we need it for our own features then it makes sense to just bring them back to core and deprecate them forever.

askvortsov1:
When merging, if we decide to keep the helper changes done here, I'll split it into 2 commits:

1. the helper stuff
2. everything else

tankerkiller125 (approved):

imorland (approved):
Checks out fine locally

composer.json
```diff
@@ -25,10 +25,12 @@
         "components/font-awesome": "^5.14.0",
         "dflydev/fig-cookies": "^3.0.0",
         "doctrine/dbal": "^2.7",
+        "dragonmantank/cron-expression": "^3.1.0",
         "franzl/whoops-middleware": "^2.0.0",
         "illuminate/bus": "^8.0",
         "illuminate/cache": "^8.0",
         "illuminate/config": "^8.0",
+        "illuminate/console": "^8.0",
```
SychO9, line 33: so the helpers were added back because illuminate/console makes use of them right ? a bit annoying after having created the `flarum/laravel-helpers`, I was under the impression that laravel's internals didn't use the helpers.

src/Console/Schedule.php
```diff
@@ -0,0 +1,39 @@
...
+namespace Flarum\Console;
+
+use Flarum\Foundation\Config;
+use Illuminate\Console\Scheduling\Schedule as LaravelSchedule;
+
+class Schedule extends LaravelSchedule
+{
+    public function dueEvents($container)
+    {
+        return collect($this->events)->filter->isDue(new FakeApp($container));
```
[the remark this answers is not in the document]
askvortsov1: Yeah let's keep this consistent. Great catch!

composer.json
askvortsov1: Correct, that was an oversight: we missed that they are needed for fof console, and that we were planning to integrate fof console back. And given how much time we spent discussing, it is annoying that we have to revert that decision so quickly. Since they **are** needed though. we might as well simplify and get back of the extra package though, I suppose: I'm not really seeing an alternative (IMO having the `schedule:list` command is critical)

SychO9 (commented):
> Should we add methods for each way to schedule? There's a LOT of options, so I think what I have here is simpler. On the other hand it's slightly less BC compatible, in case laravel decides to break something on its side. But I still think its way too many methods, and figuring out how to chain them is complicated.

We could make an adapter that reroutes method calls to the actual Event class, that way if laravel breaks something we'll have a BC layer, that said typehinting would be a problem, we could use a phpDoc block for that, but maintaining that can easily become a pain.

askvortsov1:
> > Should we add methods for each way to schedule? There's a LOT of options, so I think what I have here is simpler. On the other hand it's slightly less BC compatible, in case laravel decides to break something on its side. But I still think its way too many methods, and figuring out how to chain them is complicated.
> 
> We could make an adapter that reroutes method calls to the actual Event class, that way if laravel breaks something we'll have a BC layer, that said typehinting would be a problem, we could use a phpDoc block for that, but maintaining that can easily become a pain.

I think it's safe to assume that if Laravel breaks this, they've probably broken other things, meaning that we'd only be including that upgrade in a major release.

SychO9 (approved):

src/Console/Schedule.php
```diff
@@ -0,0 +1,40 @@
...
+
+class Schedule extends LaravelSchedule
+{
+    public function dueEvents($container)
+    {
+        return (new Collection($this->events))->filter->isDue(new FakeApp($container));
+    }
+}
+
+class FakeApp
```
luceos, line 24: Multiple class declarations in one file are really not done. It makes discoverability extremely hard. Either:

- move that class into its own file
- or use an anonymous class

askvortsov1: I keep forgetting anonymous classes exist. This is the perfect use case for one. I'll open a PR to clean this up, thanks!

PR #2708, opened 2021-03-18
Add option to build with Webpack Bundle Analyzer

davwheat:
Adds an **optional** build script to build core's JS with the Webpack Bundle Analyzer plugin installed and enabled. This allows us to get a better view into what takes up the most space within core's JS bundle.

While concatenated module size is not wholly accurate, it still provides a good idea about what takes up so much damn space!

Normal builds still function exactly how they did before, and the CI bot shouldn't be affected.

```sh
# Build with webpack bundle analyzer
npm run analyze
```

**Screenshot**
image

### DIFF
--- a/.github/workflows/pr_size_change.yml
+++ b/.github/workflows/pr_size_change.yml
@@ -29,7 +29,8 @@ jobs:
         run: sudo npm install -g npm@7.x.x
 
       - name: Install JS dependencies
-        run: npm ci
+        # We need to use `npm install` here. If we don't, the workflow will fail.
+        run: npm install
         working-directory: ./js
 
       - name: Build production assets
--- a/.github/workflows/pr_size_change.yml
+++ b/.github/workflows/pr_size_change.yml
@@ -25,6 +25,9 @@ jobs:
         with:
           node-version: "14"
 
+      - name: Use npm v7
+        run: sudo npm install -g npm@7.x.x
+
       - name: Install JS dependencies
         run: npm ci
         working-directory: ./js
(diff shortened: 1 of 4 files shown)


datitisev:
I actually did something similar in my mithril 2 branch and forgot about it 😂 

https://github.com/flarum/core/blob/4c8af1cdf876b9bd3770f76c33eb6c9ac80608f0/js/webpack-flarum-config.js#L32-L37

davwheat:
I don't know why the JS workflows are failing when trying to install deps 🤔

askvortsov1:
The tasks are still failing

davwheat:
Oooops lemme see if I can fix that 🙈

davwheat:
Well I fixed the issue, and I'm not totally sure what's causing it.

It seems like it could be related to npm/cli#558, but that's no guarantee. For now, for the workflow to pass, we need to use `npm install` instead of `npm ci`.

We might also need to do the same for the build workflow, but the dep installation for that happens outside the workflow inside flarum/action-build.

askvortsov1:
Ah, and now you have a conflict! Let's merge #2753 first though, to avoid having to rebase twice

askvortsov1 (approved):

datitisev (approved):
A lot of the changes here I feel like should've been done separately, things like the package.json deps updates as I think we also have another PR that modified those. Same with the GitHub action changes. 
 
Aside from that, LGTM.

davwheat:
Rebased and removed the deps update commit.

PR #2709, opened 2021-03-18
Speed up JS linting

davwheat:
Currently, the linting workflow installs all of core's deps before linting with Prettier.

We can save a bit of time by only installing Prettier, as no other deps are required to format our source code.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,6 +1,7 @@
 name: Lint
 
 on:
+  workflow_dispatch:
   push:
     paths:
       - 'js/src/**'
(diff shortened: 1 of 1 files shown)


datitisev:
How about we replace it with `npx prettier`? That way we combine 2 steps into one.

datitisev:
Why change the name? I'd rather have the action names be short and simple.
I prefer "Lint / JS / Prettier" over "Lint / Validate JS formatting"

davwheat:
@datitisev I'd just copied that from one of my other workflows -- my bad. Changed it back now.

askvortsov1 (approved):

datitisev (approved):

PR #2710, opened 2021-03-18
Optimise JS build workflow

davwheat:
Also supports testing JS builds on all branches, not just master. Changes will only be committed following commit pushes to master..

This PR also adds caching of `node_modules` between CI runs, reducing Actions time.

Adapted from my own workflows: https://github.com/davwheat/personal-portfolio/actions/runs/664540482


datitisev:
The action didn't run because JS files weren't modified 😂 so we can't test if it works until JS is modified lol

.github/workflows/build.yml
```diff
@@ -11,6 +11,16 @@ jobs:
...
       - uses: actions/checkout@master
+
+      - name: Cache node_modules
+        uses: actions/cache@v2
+        id: module-cache
+        with:
+          path: ./js/node_modules
+          key: ${{ runner.os }}-node-modules-${{ github.sha }}-${{ hashFiles('**/package-lock.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-node-modules-${{ github.sha }}-
```
tankerkiller125, line 22: Shouldn't this also be `${{ runner.os }}-node-modules-${{ github.sha }}-${{ hashFiles('**/package-lock.lock') }}`?

davwheat:
Build hasn't got that condition, I don't think, does it?

.github/workflows/build.yml
davwheat: This is a set of keys to match as a fallback if the normal key doesn't exist.

This lets us also restore the previous node modules if we add new deps/update deps, so we only install deps that changed, just like locally.

Took me a few hours of research to understand this... It's not very clear.

tankerkiller125:
@davwheat It does, JS is only built on JS changes. That's why only the PHP test are running on your commits.

davwheat:
Turns out JS is built no matter what, but only on pushes to master (no changes are made because the JS compiles to the exact same output though).

Prettier is only run on changes to the JS source.

davwheat:
Just discovered that caching `mode_modules` has some incompatibilities with `npm ci`, so I'm gonna be changing the caching strategy to [what's recommended by the cache action repo](https://github.com/actions/cache/blob/main/examples.md#macos-and-ubuntu).

.github/workflows/build.yml
```diff
@@ -2,15 +2,43 @@ name: JavaScript
...
 jobs:
   build:
+    name: Build JS
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@master
-      - uses: flarum/action-build@master
+      - name: Check out code
+        uses: actions/checkout@master
```
askvortsov1, line 14: Should this be tagged

davwheat: Not now I removed it ;)

But the flarum action should 100% be tagged.

I want to add some `with` options to it to allow custom build scripts and toggling between npm and yarn, and those changes should be tagged.

davwheat:
I removed some steps from the workflow after realising they were a part of `flarum/action-build` 😅 

I've fixed the cache issue now, too :)

askvortsov1 (approved):

tankerkiller125 (approved):

PR #2717, opened 2021-03-20
Eager load ListPostsController needed relations

SychO9:
**Part of #2637**

Eager loads the `discussion` and `user.groups` relations, leaving `discussion.tags` and `mentionedBy` relations until there is a mechanism for extensions to specify needed relations to be eager loaded (which I'll push after this).

You can test this by hitting the `/api/posts?page[offset]=0&page[limit]=50&include=user` endpoint.

**Screenshot**
flarum lan___clockwork_app (1)
flarum lan___clockwork_app (2)

**Confirmed**

### DIFF
--- a/src/Api/Controller/ListPostsController.php
+++ b/src/Api/Controller/ListPostsController.php
@@ -93,6 +93,14 @@ class ListPostsController extends AbstractListController
             $results->areMoreResults() ? null : 0
         );
 
+        if (! in_array('discussion', $include)) {
+            $include[] = 'discussion';
+        }
+
+        if (in_array('user', $include)) {
+            $include[] = 'user.groups';
+        }
+
         return $results->getResults()->load($include);
     }
 
--- a/src/Api/Controller/ListPostsController.php
+++ b/src/Api/Controller/ListPostsController.php
@@ -93,6 +93,9 @@ class ListPostsController extends AbstractListController
             $results->areMoreResults() ? null : 0
         );
 
+        // Eager load discussion for use in the policies,
+        // eager loading does not affect the JSON response,
+        // the response only includes relations included in the request.
         if (! in_array('discussion', $include)) {
             $include[] = 'discussion';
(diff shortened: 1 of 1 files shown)


tankerkiller125 (approved):

SychO9:
actually, don't merge this yet, I'd like to confirm that these relations are not only needed by specific extensions.

src/Api/Controller/ListPostsController.php
```diff
@@ -93,6 +93,14 @@ protected function data(ServerRequestInterface $request, Document $document)
             $results->areMoreResults() ? null : 0
         );
 
+        if (! in_array('discussion', $include)) {
```
askvortsov1, line 96: This is concerning to me, doesn't this mean that we'd always be loading the discussion, even not when requested? 

Shouldn't we be checking that `'discussion.*'` is in the array? We might want to add a util for `AbstractSerializeController` that allow wildcard checks for inclusion in `$include`. Or maybe there's some PHP functio n already out there?

askvortsov1 (changes requested):

src/Api/Controller/ListPostsController.php
SychO9: that's the point, we need to always eager load it, because the PostSerializer which calls the PostPolicy uses the `$post->discussion` relationship, so you end up with either tens of database queries for each discussion, or one that loads them all.

The key difference to note here between including a relationship in a request and eager loading, is that whether you eager load the relationship or not, it will only be added to the JSON response if it was included in the request.

askvortsov1: Ah, got it. Can you add a comment explaining this? It's a bit counterintuitive from just looking at the code

SychO9: Sure

askvortsov1 (approved):

PR #2721, opened 2021-03-20
Harden Headers

### LINKED ISSUE #353: Hardening HTTP Headers

tankerkiller125:
**Partial Fix #353 **

- Adds the Referer Policy header (with config option)
- Adds Xss Protection header
- Adds Content Type header

- Does the config option for the Referer policy make sense? Or should we use some sort of constants?
- We probably need to add some test for this but I'm unsure

**Confirmed**

**Additional PRs Required**
- #2777


src/Http/Middleware/ReferrerPolicyHeader.php
```diff
@@ -0,0 +1,35 @@
...
+use Psr\Http\Server\MiddlewareInterface as Middleware;
+use Psr\Http\Server\RequestHandlerInterface;
+
+class ReferrerPolicyHeader implements Middleware
+{
+    protected $policy = '';
+
+    public function __construct(Config $config)
+    {
+        $this->policy = $config['headers']['referrer-policy'] ?? 'same-origin';
```
askvortsov1, line 24: Do we use `config['headers']` for anything else? If we choose to use it here, we should move the promotional header key to use that as well.

src/Http/Middleware/XssProtectionHeader.php
```diff
@@ -0,0 +1,25 @@
...
+use Psr\Http\Server\MiddlewareInterface as Middleware;
+use Psr\Http\Server\RequestHandlerInterface;
+
+class XssProtectionHeader implements Middleware
+{
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
+    {
+        $response = $handler->handle($request);
+
+        return $response->withAddedHeader('X-XSS-Protection', '1; mode=block');
```
askvortsov1, line 23: How does this affect custom JS added in the head element, or via custom header / footer? Or the custom JS used in TextFormatter for code highlighting and media embed, for that matter?

src/Admin/AdminServiceProvider.php
```diff
@@ -57,7 +57,10 @@ public function register()
                 HttpMiddleware\SetLocale::class,
                 'flarum.admin.route_resolver',
                 HttpMiddleware\CheckCsrfToken::class,
-                Middleware\RequireAdministrateAbility::class
+                Middleware\RequireAdministrateAbility::class,
```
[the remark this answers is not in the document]
askvortsov1: Leaning towards no

askvortsov1 (changes requested):
This definitely should have tests. We could add a `header` subfolder to the integration test directory?

Or alternatively, construct a fake middleware pipe with a fake handler, and use unit tests.

src/Http/Middleware/XssProtectionHeader.php
davwheat: This header is deprecated and only supported in IE.

Chrome disabled all effects of this header from 77, and FF never implemented it in the first place.

Source: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection

tankerkiller125: It won't block any of the JS added via the header or via embeds as far as I've tested. To be honest I'm not 100% honest how the X-XSS-Protection header works but I do know that it doesn't seem to affect anything on my own forum where I've added it via my webserver.

askvortsov1: Maybe we shouldn't include it, and focus efforts on supporting CSP instead? That seems to be the more modernly accepted option for doing what we want to do here.

src/Http/Middleware/ReferrerPolicyHeader.php
tankerkiller125: We currently don't but I think we might want too?

src/Admin/AdminServiceProvider.php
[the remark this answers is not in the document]
tankerkiller125: I'm personally thinking no, my reason being that we allow extension devs to remove and replace middleware. So for extensibility I think we should keep them seperate.

src/Http/Middleware/ReferrerPolicyHeader.php
askvortsov1: I'm in favor, but let's add a PR to change the promotional header to use that too

src/Http/Middleware/ReferrerPolicyHeader.php
```diff
@@ -0,0 +1,35 @@
...
+    public function __construct(Config $config)
+    {
+        $this->policy = $config['headers']['referrerPolicy'] ?? 'same-origin';
+    }
+
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
+    {
+        $response = $handler->handle($request);
+
+        $response = $response->withAddedHeader('Referrer-Policy', $this->policy);
```
askvortsov1, line 31: For consistency with the one above, let's return `$response->withAddedHeader(...)` directly?

askvortsov1 (changes requested):
Stylistic comment. Also would be good to see some basic integration tests (ping `/`, check for headers)

tankerkiller125:
> This definitely should have tests. We could add a header subfolder to the integration test directory?

I'm thinking maybe a `middleware` subfolder, that way in the future we can test other middleware?

tankerkiller125:
As a note I'm not sure why the test are failing, it seems to be that it's hitting rate limits or something? Everything is working locally 100% just fine and all test are passing on my machine.

src/Install/Steps/StoreConfig.php
```diff
@@ -55,10 +55,13 @@ private function buildConfig()
     {
         return [
             'debug'    => $this->debugMode,
-            'poweredByHeader'  => true,
             'database' => $this->dbConfig->toArray(),
             'url'      => (string) $this->baseUrl,
             'paths'    => $this->getPathsConfig(),
+            'headers'  => [
+                'poweredByHeader'  => true,
```
davwheat, line 62: Noticed this as I scrolled past and it couldn't stop bugging me 🙈 

```suggestion
                'poweredByHeader' => true,
```

src/Install/Steps/StoreConfig.php
```diff
@@ -55,10 +55,13 @@ private function buildConfig()
...
+                'referrerPolicy'
```
davwheat, line 63: Not sure if I misunderstand what this is, but should we maybe add the default value of this from `ReferrerPolicyHeader.php`?

davwheat (commented):
All looks good to me from a headers point of view!

askvortsov1:
> As a note I'm not sure why the test are failing, it seems to be that it's hitting rate limits or something? 

FIgured it out. The `setting` method of test case runs BEFORE transactions are started, so those changes don't go into transactions. Opened a PR: https://github.com/flarum/testing/pull/11

luceos (approved):

davwheat (approved):

composer.json
```diff
@@ -69,7 +69,7 @@
         "wikimedia/less.php": "^3.0"
     },
     "require-dev": {
-        "flarum/testing": "^0.1.0-beta.16"
+        "flarum/testing": "dev-main#718cd86e5d4d247d63cb1034741d7b26a16e3332"
```
luceos, line 72: This should not be part of this PR.

askvortsov1:
> This should not be part of this PR.
Unfortunately it probably should, as tests break without it. There's a bunch of stuff that shouldn't though. From chat:

I think it might be easiest here to:

- start a new branch from master
- cherry-pick over relevant commits
- run tests
- force update current branch to this new branch

tankerkiller125:
I've re-based correctly this time, but it appears that test unrelated to this PR are failing again. I'm not entirely sure what's going on here. Maybe @askvortsov1 can take a look when he gets a chance?

askvortsov1 (approved):
Fixed up tests, moved those commits out to master to keep this PR clean. @davwheat did you self-request review because you had some concerns?

davwheat:
@tankerkiller125 No, sorry, just wanted to clear my review as I hadn't tested it locally after the rebase and stuff.

No real changes have taken place, and Sasha has no issues with it, so it's good with me.

PR #2724, opened 2021-03-21
Eager loading extender

SychO9:
**Part of #2637**

Adds a `load()` method on the `ApiController` extender that allows extensions to specify relations that need to be loaded in listing controllers. This works when controllers use the appropriate `loadRelations($collections, $relations)` method.

A first level relationship will always be loaded, whereas a sublevel relationship will only be loaded if the parent level is/will be loaded.

As for the noticiations endpoint that has the `subject` polymorphic relationship, I would like to add in another pull request a `loadMorph` method to this same extender that would solve the problem.

**Confirmed**

- [X] Related core extension PRs: (Remove if irrelevant).
  + [X] https://github.com/flarum/tags/pull/125
  + [X] https://github.com/flarum/mentions/pull/64
  + [X] https://github.com/flarum/flags/pull/38


src/Api/Controller/AbstractSerializeController.php
```diff
@@ -139,6 +146,47 @@ abstract protected function data(ServerRequestInterface $request, Document $docu
...
     abstract protected function createElement($data, SerializerInterface $serializer);
 
+    /**
+     * Eager loads the required relationships.
+     *
+     * @param Collection $models
+     * @param array $relations
+     * @return void
+     */
+    protected function loadRelations(Collection $models, array $relations): void
```
askvortsov1, line 156: Could this replace the special case on line 129 of ListDiscussionsController?

src/Extend/ApiController.php
```diff
@@ -216,6 +217,27 @@ public function setSort(array $sort, $callback = null)
         return $this;
     }
 
+    /**
+     * Eager loads relationships.
+     *
+     * First level relationships will be loaded regardless of them being included in the request.
```
askvortsov1, line 223:
```suggestion
     * First level relationships will be loaded regardless of whether they are included in the response.
```

askvortsov1, line 223: This is because they're needed for serializer logic, right? If so, can we mention that?

askvortsov1 (commented):
This will need a bit in the extension development docs, so that we can point extension devs to an explanation of how to use it.

src/Api/Controller/AbstractSerializeController.php
SychO9: You are referring to this ? https://github.com/flarum/core/blob/2009f790d9112a3a8705e02569ff407c953cc7db/src/Api/Controller/ListDiscussionsController.php#L128-L136

I suppose we could, but then it would mean more queries to re-load the discussions and their required relations (tags for instance) that'd be at least 6+ queries, even though they've already been loaded. Unless you meant having the method handle that situation, but that also sounds like more complexity.

askvortsov1: Let's keep it as an edge case for now then :+1:

askvortsov1 (approved):

tankerkiller125 (approved):

PR #2731, opened 2021-03-23
Asset Publish Command

### LINKED ISSUE #1644: New command to publish assets
From https://github.com/flarum/core/commit/b26eb8e609006f7199ffe07e6a36e5503165dc43:

Extract publishing assets from the `migrate` command into a new command. Possible names:

* `publish`
* `vendor:publish` (this is what Laravel uses right?)
* `asset(s):publish`
* ...

Also will need to consolidate the publishing logic with the installer while we're at it.

askvortsov1:
**Followup to flarum/framework#2729**
**Fixes flarum/framework#1644**

- Add a new `assets:publish` command that includes:
  - Publishing fonts (previously in `migrate`, for some reason...)
- Use the filesystem disks, as started in flarum/framework#2729

- Already uploaded logos and favicons won't be updated. However, I think that's the correct way to approach this: this command should publish assets from source code and extension code; if a forum is migrating from one filesystem driver to another, that should be an entirely different system IMO (and probably not even in core).

**Confirmed**


luceos (approved):
Not tested but looking excellent ;)

tankerkiller125 (approved):
I haven't tested locally, but code wise this looks good

askvortsov1:
I'm going to wait until the other 2 PRs are approved, then merge everything at once in case something changes.

luceos:
I'm missing commits to fix the installation procedure.

askvortsov1:
> I'm missing commits to fix the installation procedure.

Pardon? Modifying the installation procedure isn't part of this PR, and I'm not sure it can be: wouldn't we need InstalledSite in place before we can run commands?

luceos:
The installer publishes the fonts. That's what I meant. You can't split it out and then not fix the installer.

askvortsov1:
> The installer publishes the fonts. That's what I meant. You can't split it out and then not fix the installer.

Yes, but it doesn't use the command, it directly copies it over within the filesystem: https://github.com/flarum/core/blob/eac7d1b7ade79f0d4ee3b44b210bc09005491316/src/Install/Steps/PublishAssets.php#L39-L45

luceos:
@askvortsov1 that, I missed. Disregard and continue ;)

PR #2732, opened 2021-03-23
Filesystem Extender and Tests

askvortsov1:
**Followup to flarum/framework#2729**

- Use our own subclass of Laravel's filesystem manager, which gives us control over disk resolution, and most important, creating drivers for disks.
- Add a filesystem extender that allows:
  - Declaring a new disk (see docblock for explanation)
  - Registering a filesystem driver

Currently, drivers for disks are determined via the following process:

1. If a `disk_driver.${DISK_NAME}` setting key exists, proceed to step 2. Else return `'local'`
2. If a driver with the name returned by step 1 exists, use that. Else, use `'local'`

This gives us safety in case the extension providing the filesystem driver being used is removed.

This PR does not attempt to introduce a UI for:

- Selecting drivers for disks
- Selecting settings for disk drivers

We might want to eventually add that though.

Do we want to use `config.php` or settings? Right now, this PR selects the driver for each disk from settings. It passes both settings and config to drivers, so they can choose which they prefer. Should we also support setting drivers for disks in config?

I also put together a proof of concept extension for Azure at https://github.com/askvortsov1/flarum-azure-poc.

**Confirmed**


tankerkiller125:
I think I'd prefer `config.php`, however I may have a bias towards that since Laravel uses a filesystem.php config file. However if a proper UI can be introduced I think that could work just as well, especially given that Flarum can't operate without a database anyways.

askvortsov1:
> I think I'd prefer `config.php`, however I may have a bias towards that since Laravel uses a filesystem.php config file. However if a proper UI can be introduced I think that could work just as well, especially given that Flarum can't operate without a database anyways.

For mapping disks to drivers, we could support both: first check settings, then config.php, and if none have a value, default to `local`?

tankerkiller125:
That works too

src/Filesystem/FilesystemManager.php
```diff
@@ -0,0 +1,82 @@
...
+
+        return new FilesystemAdapter($driver->build($name, $settings, $config, $localConfig));
+    }
+
+    protected function getDriver($name)
+    {
+        $config = $this->app->make(Config::class);
+        $settings = $this->app->make(SettingsRepositoryInterface::class);
+
+        $key = "disk_driver.$name";
```
luceos, line 62: I dislike `disk_driver` it strokes against the Laravel naming convention and the manager elsewhere uses `filesystems` too. Can we use `filesystem` or the plural version?

luceos (changes requested):
I don't understand why we are extending the Laravel filesystem manager only to return a Flysystem Filesystem afterwards (see DriverInterface).

Other than that experience will tell whether this helps and callables might be hard to use as they don't have type hinting, but that was pointed out by @clarkwinkelmann in the past as well.

src/Filesystem/FilesystemManager.php
askvortsov1: In the Laravel config approach, there is an array of "disks", with each "disk" having a "driver" field. In this particular place, that is not inaccurate.

We could replace the `'flarum.filesystem.drivers'` binding key with `'flarum.filesystem.filesystems'`?

askvortsov1:
> I don't understand why we are extending the Laravel filesystem manager only to return a Flysystem Filesystem afterwards (see DriverInterface).

We are extending (and overriding) the filesystem manager so that we have control over:

- How new disks are declared
- How disks are mapped to drivers
- How available drivers are registered
- How config for drivers is obtained

That being said, I think you have a good point that it's a bit odd that the output is a Flysystem interface. I did this because the vast majority of drivers will probably use flysystem so we might as well wrap in `FilesystemAdapter` for them, but that's not too clean. We could probably require the output to match the laravel Cloud interface.

src/Filesystem/FilesystemManager.php
```diff
@@ -0,0 +1,81 @@
...
+use Illuminate\Filesystem\FilesystemManager as LaravelFilesystemManager;
+use Illuminate\Support\Arr;
+use InvalidArgumentException;
+
+class FilesystemManager extends LaravelFilesystemManager
+{
+    protected $diskLocalConfig = [];
+    protected $drivers = [];
+
+    public function __construct($app, $diskLocalConfig, $drivers)
```
SychO9, line 25: `array $diskLocalConfig, array $drivers`

src/Filesystem/FilesystemManager.php
```diff
@@ -0,0 +1,81 @@
...
+            return $this->createLocalDriver($localConfig);
+        }
+
+        $settings = $this->app->make(SettingsRepositoryInterface::class);
+        $config = $this->app->make(Config::class);
+
+        return $driver->build($name, $settings, $config, $localConfig);
+    }
+
+    protected function getDriver($name)
```
SychO9, line 56: `string $name): DriverInterface`

src/Filesystem/DriverInterface.php
```diff
@@ -0,0 +1,37 @@
...
+namespace Flarum\Filesystem;
+
+use Flarum\Foundation\Config;
+use Flarum\Settings\SettingsRepositoryInterface;
+use Illuminate\Contracts\Filesystem\Cloud;
+
+interface DriverInterface
+{
+    /**
+     * Construct a Flysystem filesystem adapter for this filesystem driver.
```
SychO9, line 19: sorry I'm a bit confused here, doesn't it construct an adapter for the disk ?

SychO9 (commented):
Code looks good!

Just have a few minor nitpicks.

src/Filesystem/FilesystemManager.php
SychO9: I do agree that `disk_driver` is more appropriate here.

src/Filesystem/FilesystemManager.php
askvortsov1: Could also be a string, so I'm putting this in a docblock. Can't wait for union types!

src/Extend/Filesystem.php
```diff
@@ -0,0 +1,80 @@
...
+    {
+        $this->drivers[$name] = $driverClass;
+
+        return $this;
+    }
+
+    public function extend(Container $container, Extension $extension = null)
+    {
+        $container->extend('flarum.filesystem.disks', function ($existingDisks) {
+            return array_merge($existingDisks, $this->disks);
```
SychO9, line 73: disks is an array of callables, I think we should use `ContainerUtil::wrapCallback` on them ?

src/Filesystem/FilesystemManager.php
```diff
@@ -0,0 +1,83 @@
...
+    protected function getLocalConfig(string $name): array
+    {
+        if (! array_key_exists($name, $this->diskLocalConfig)) {
+            return [];
+        }
+
+        return $this->diskLocalConfig[$name](
+            $this->app->make(Paths::class),
+            $this->app->make(UrlGenerator::class)
+        );
```
SychO9, line 81: Tests are failing because of this https://github.com/flarum/core/pull/2485/files.

we have to declare two variables and pass them, annoying but well..

SychO9 (approved):

luceos (approved):

tankerkiller125 (approved):

Issue #2727, opened 2021-03-22
Scalability meta issue


luceos:
This issue is meant to provide actionable tasks to improve scalability in specialised, highly volatile hosting environments.

- The renderer is stored and cached to local disk. #2588
- the `users.avatar_url` string length is too short disallowing the use of CDN's
- `migrations` table has no auto increment primary key ID on some DB's this causes errors and fails migration completely, I also remember faintly some db tools dislike this and remove checkboxes (phpmyadmin?)
- The DatabaseSessionHandler requires knowledge of the actor. As user nor request isn't available from the container one cannot override the `session.handler` easily; right now fixed by binding the actor into the container.
- The js (forum, admin and locale)/css are stored using `something-<hash>.filetype`, when invalidating cache on one node and that node stores a new version into a CDN, users on other nodes will require to reload the page to retrieve the latest rev-manifest to load the right files. In the meantime their pages might break. A solution is to use the revision/hash as query param thus auto invalidating client-side cached files, but also providing a fallback in case the revision is incorrect.
- Overriding any of the compilers (because you want to override the revision generation), requires you to override the Assets class which you cannot override easily. It is possible using `extend` but then you also need to hit the original `FrontendServiceProvider` to retrieve `addBaseCss` twice (css, localeCss).
- `logoUrl` and `favIconUrl` both need to start using a similar implementation like the avatar_url of User. Right now the only way to override these is by mutating the attributes in the ForumSerializer.
- It would be helpful to be able to add migrations from the local `extend.php`. Right now my default go to is to create a custom command that replicates our own migration command, but with a different path.

askvortsov1:
> the users.avatar_url string length is too short disallowing the use of CDN's

Wouldn't we want this to be the path relative to the CDN's root? That would give more flexibility in changing avatar filesystems.

> migrations table has no auto increment primary key ID on some DB's this causes errors and fails migration completely, I also remember faintly some db tools dislike this and remove checkboxes (phpmyadmin?)

Yeah, we should add this. I suppose we'll want to do something like what Clark did in #2651. Luckily I don't think we'll ever have tens of thousands of migrations, so that should make life quite a bit easier.

> The DatabaseSessionHandler requires knowledge of the actor. As user nor request isn't available from the container one cannot override the session.handler easily; right now fixed by binding the actor into the container.

[...]

PR #2738, opened 2021-03-24
Removes duplication of cache clearing

luceos:

This PR removes the duplicated cache clearing of the locales. The locales are cleared using the cache:clear command which is centrally dispatched from the CacheClear command. Using an listener on the ClearingCache event causes extensions to fail to do stuff with locales.

Eg:

- extension listens on ClearingCache to run cache:assets to create locales ahead of next visit
- LocaleServiceProvider, executed last, removes those files immediately

Ref: https://github.com/flarum/core/blob/d642fb531c1335c33ff7515da367f6815ea93b12/src/Foundation/Console/CacheClearCommand.php#L69

**Confirmed**

### DIFF
--- a/src/Locale/LocaleManager.php
+++ b/src/Locale/LocaleManager.php
@@ -112,4 +112,11 @@ class LocaleManager
     {
         $this->translator = $translator;
     }
+
+    public function clearCache()
+    {
+        if ($this->cacheDir) {
+            array_map('unlink', glob($this->cacheDir.'/*'));
+        }
+    }
 }
--- a/src/Locale/LocaleManager.php
+++ b/src/Locale/LocaleManager.php
@@ -112,11 +112,4 @@ class LocaleManager
     {
         $this->translator = $translator;
     }
-
-    public function clearCache()
-    {
-        if ($this->cacheDir) {
-            array_map('unlink', glob($this->cacheDir.'/*'));
-        }
-    }
 }
--- a/src/Locale/LocaleServiceProvider.php
+++ b/src/Locale/LocaleServiceProvider.php
@@ -10,25 +10,13 @@
 namespace Flarum\Locale;
 
 use Flarum\Foundation\AbstractServiceProvider;
-use Flarum\Foundation\Event\ClearingCache;
 use Flarum\Foundation\Paths;
 use Flarum\Settings\SettingsRepositoryInterface;
-use Illuminate\Contracts\Events\Dispatcher;
 use Illuminate\Contracts\Translation\Translator as TranslatorContract;
 use Symfony\Contracts\Translation\TranslatorInterface;
 
 class LocaleServiceProvider extends AbstractServiceProvider
 {
-    /**
-     * {@inheritdoc}
-     */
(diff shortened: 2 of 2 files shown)


askvortsov1 (approved):
Makes sense to me! I would eventually like to see caches move to a unified system ala https://github.com/flarum/core/issues/2727#issuecomment-805774878, but that's a matter for another release.

src/Locale/LocaleManager.php
```diff
@@ -112,11 +112,4 @@ public function setTranslator(Translator $translator)
     {
         $this->translator = $translator;
     }
-
-    public function clearCache()
```
SychO9, line 116: We can't delete this method, it is used elsewhere, `LanguagePack` extender lifecycle hooks for example.

SychO9 (changes requested):

src/Locale/LocaleManager.php
luceos: Wow! I'm so curious about the sensibility of that. Enabling any extension flushes the whole cache anyway right? Wait I can't find where we do that. So if we don't clear the full cache, wouldn't it make sense to do that because the JS is still cache without clearing that?

SychO9: Enabling or disabling an extension does indirectly clear the full cache, most of that is done through the extenders lifecycle hooks (`LanguagePack`, `Locales`, `Formatter`, `View`) with the exception of frontend assets (js, css) those are cleared/recompiled by listening to the (`Enabled`, `Disabled`, `ClearingCache`) events.

luceos: So:

ForumServiceProvider and AdminServiceProvider:

```php
        $events->listen(
            [Enabled::class, Disabled::class, ClearingCache::class],
            function () {
                $recompile = new RecompileFrontendAssets(
                    $this->container->make('flarum.assets.forum'),
                    $this->container->make(LocaleManager::class)
                );
                $recompile->flush();
            }
        );
```

Seems to flush only css and js. Not the locale cache.

The Frontend Extender does something similar but for different frontends if they haven't been registered yet (ps this is code duplication, maybe all three locations need a trait?)

```php
            $events->listen(
                [Enabled::class, Disabled::class, ClearingCache::class],
                function () use ($container, $abstract) {
                    $recompile = new RecompileFrontendAssets(
                        $container->make($abstract),
                        $container->make(LocaleManager::class)
                    );
                    $recompile->flush();
                }
            );
```
Same logic, only js, css, not locale cache.

SychO9 (approved):

src/Locale/LocaleManager.php
SychO9: > (ps this is code duplication, maybe all three locations need a trait?)

Yea I think a trait'd be nice for this, we should remember to do that

> I think keeping the clearCache method then makes sense, but only to be used by the LanguagePack.

and Locales extender

askvortsov1 (approved):

davwheat:
2 approvals. Anything I'm missing or is this good to merge?

PR #2739, opened 2021-03-25
Small Admin Patches

### LINKED ISSUE #2728: Admin setting help text disappears after redraw
## Bug Report

**Current Behavior**
The help text defined Using `ExtensionData.registerSetting()` disappears after the first Mithril redraw of the settings page. 

**Steps to Reproduce**

Use following javascript code:

```js
app.initializers.add('my-extension', () => {
    app.extensionData.for('my-extension')
        .registerSetting({
            type: 'switch',
            setting: 'my-setting',
            label: 'A label',
            help: 'A help text',
        });
});
```
image

Click switch:

image

*Poof* help text is gone and won't come back until next time the page is opened.

**Expected Behavior**
Help text shouldn't disappear.

**Environment**
- Flarum version: beta 16, likely earlier

**Possible Solution**

The problem is here https://github.com/flarum/core/blob/706eaeda4153de130b8530713f00772df7146458/js/src/admin/components/AdminPage.js#L103

The `help` attribute is `delete`ed from the object, but since the object is modified by reference, it's gone forever.

The object needs to be cloned before the `help` attribute is removed.

That `delete` statement seems very out of place. Why do we remove `help` and not `label` ?

Also there's likely a similar issue with `className` lower in the file where an additional `FormControl` class will be added on each redraw https://github.com/flarum/core/blob/706eaeda4153de130b8530713f00772df7146458/js/src/admin/components/AdminPage.js#L124

### LINKED ISSUE #2736: Extensions with no title or description breaks the search input in admin
## Bug Report

**Current Behavior**
If an extension with empty title or description is installed, the search input in the admin panel will broke.

**Steps to Reproduce**
1. Install an extension without title or description
2. Go to 'admin panel'
3. Click on 'Search Extensions'
4. Type something
5. See error

**Expected Behavior**
Shouldn't happen.

**Environment**
- Flarum version: 0.1.0-beta.16

**Additional Context**

```js
TypeError: Cannot read property 'toUpperCase' of undefined
    at AdminNav.js:131
    at Array.map (<anonymous>)
    at AdminNav.js:127
    at Array.map (<anonymous>)
    at e.n.extensionItems (AdminNav.js:118)
    at e.n.view (AdminNav.js:19)
    at Function.s (render.js:30)
    at render.js:482
    at h (render.js:419)
    at f (render.js:302)
```

The problem is here:

https://github.com/flarum/core/blob/d642fb531c1335c33ff7515da367f6815ea93b12/js/src/admin/components/AdminNav.js#L131

KyrneDev:
Probably best to not merge this yet as we wait for any further issues to crop up and be reported.

* Allow extensions to have an empty title or description without breaking the search. **Fixes #2736**
* Copy the settings entry data so it can be mutated without affecting future redraws. **Fixes #2728**

**Confirmed**

### DIFF
--- a/js/src/admin/components/AdminNav.js
+++ b/js/src/admin/components/AdminNav.js
@@ -126,16 +126,17 @@ export default class AdminNav extends Component {
 
       categorizedExtensions[category].map((extension) => {
         const query = this.query().toUpperCase();
-        const title = extension.extra['flarum-extension'].title;
+        const title = extension.extra['flarum-extension'].title || '';
+        const description = extension.description || '';
 
-        if (!query || title.toUpperCase().includes(query) || extension.description.toUpperCase().includes(query)) {
+        if (!query || title.toUpperCase().includes(query) || description.toUpperCase().includes(query)) {
           items.add(
             `extension-${extension.id}`,
             <ExtensionLinkButton
               href={app.route('extension', { id: extension.id })}
               extensionId={extension.id}
               className="ExtensionNavButton"
-              title={extension.description}
+              title={description}
             >
               {title}
             </ExtensionLinkButton>,
--- a/js/src/admin/components/AdminPage.js
+++ b/js/src/admin/components/AdminPage.js
@@ -98,35 +98,41 @@ export default class AdminPage extends Page {
       return entry.call(this);
     }
 
-    const setting = entry.setting;
-    const help = entry.help;
-    delete entry.help;
+    const { setting, help, ...componentAttrs } = entry;
+
+    delete componentAttrs.help;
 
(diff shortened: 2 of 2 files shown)


askvortsov1 (approved):

davwheat (approved):

davwheat:
@KyrneDev Is it worth merging now, or just wait longer?

askvortsov1:
> @KyrneDev Is it worth merging now, or just wait longer?

I think we should merge now; if other issues pop up, they would be better in a separate PR. I see this is nicely broken into separate commits, so let's merge via merge commit (so PR is still references)

PR #2740, opened 2021-03-26
error handling when extending flarum from extensions fails

luceos:

Captures any errors throw when trying to extend flarum from an extension. Right now when an extender misbehaves we get errors like these:

```
PHP Fatal error:  Uncaught Error: Call to undefined method Closure::extend() in /home/forge/nightly.flarum.site/nightly/vendor/flarum/core/src/Extension/Extension.php:137
Stack trace:
```

Which isn't really useful, the trace doesn't show what extension nor what extender is causing the issue.

Please do a local test I haven't been able to do so.

**Confirmed**


src/Extension/Exception/ExtendingFailedException.php
```diff
@@ -0,0 +1,28 @@
...
+{
+    public $extension;
+    public $extender;
+
+    public function __construct(Extension $extension, $extender, Throwable $previous = null)
+    {
+        $this->extension = $extension;
+        $this->extender = $extender;
+
+        parent::__construct($extension->getTitle().' has an invalid extend.php: '.get_class($extender), null, $previous);
```
askvortsov1, line 26: This message is a bit confusing. Maybe:

` has an invalid extend.php: error experienced when applying extender of type: `?

src/Extension/Exception/ExtendingFailedExceptionHandler.php
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * For detailed copyright and license information, please view the
+ * LICENSE file that was distributed with this source code.
+ */
+
+namespace Flarum\Extension\Exception;
+
+use Flarum\Foundation\ErrorHandling\HandledError;
+
+class ExtendingFailedExceptionHandler
```
askvortsov1, line 14: I am not sure this is necessary; when we enable an extension, it won't actually be applied until the next refresh. At that point, the error will happen as a boot error, so our error handler won't have an opportunity to intercept. And this isn't used by the logging system. I don't think we need to do anything beyond just throwing the extension.

src/Extension/Exception/ExtendingFailedException.php
askvortsov1: Actually we might want to go a little further:

```
Experienced an error when booting extension: EXTENSION_TITLE
Error occured when applying an extender of type: EXTENDER_CLASS
```

askvortsov1 (changes requested):

askvortsov1 (approved):

SychO9 (approved):

PR #2753, opened 2021-04-04
Bump dependencies, add missing typing libraries

davwheat:
- Bumps dependencies to their latest versions in `package.json`.
  - ~~Mostly of interest is `spin.js` which has had a [few breaking changes](https://github.com/fgnass/spin.js/blob/master/CHANGELOG.md#400---2018-05-06), but they shouldn't affect us. They actually improve performance :)~~
    - It might be worth dropping this in favour of a native CSS-only solution as part of flarum/issue-archive#179, saving on [about 1.5 kB of JS](https://bundlephobia.com/result?p=spin.js@4.1.0)
  - I've also bumped Prettier and made sure it's now auto-updating to `^2.2.1` as before it was fixed to `2.0.2` (was this intentional?)
  - `expose-loader` has also been updated and had breaking changes, but these have been fixed as part of the upgrade process
- Adds missing typing libraries
  - `@types/jquery`
    - Remove export from our typing lib as `@types/jquery` should be used by extensions (and because re-exporting it causes issues)
  - `@types/lodash-es`
  - `@types/punycode`
  - `@types/textarea-caret`
- Moves `devdeps` that were incorrectly listed under `deps`
  - `webpack` (and related)
  - `@types/mithril`
  - `@babel/preset-typescript`
- Use lockfile version 2 syntax (npm 7+)

**Confirmed**


davwheat:
Woop woop -147 bytes of JS 😂

js/src/common/index.js
```diff
@@ -1,6 +1,8 @@
-import 'expose-loader?$!expose-loader?jQuery!jquery';
-import 'expose-loader?m!mithril';
-import 'expose-loader?dayjs!dayjs';
+// Expose jQuery, mithril and dayjs to the window browser object
+import 'expose-loader?exposes[]=$&exposes[]=jQuery!jquery';
+import 'expose-loader?exposes=m!mithril';
```
askvortsov1, line 3: This is a nicer API, I like this change

less/common/LoadingIndicator.less
```diff
@@ -12,3 +12,50 @@
 .LoadingIndicator--block {
   height: 100px;
 }
+
```
askvortsov1, line 15: Is this the CSS-only solution? If so, it should be a separate PR.

askvortsov1 (changes requested):

js/package.json
```diff
@@ -2,29 +2,33 @@
   "private": true,
   "name": "@flarum/core",
   "dependencies": {
-    "@babel/preset-typescript": "^7.10.1",
-    "@types/mithril": "^2.0.3",
     "bootstrap": "^3.4.1",
-    "classnames": "^2.2.5",
+    "classnames": "^2.3.1",
```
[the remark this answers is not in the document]
davwheat: My plan was that, while it's in, we might as well update it. I was actually planning to submit the `clsx` PR tonight :P

less/common/LoadingIndicator.less
davwheat: No, these are the styles from `spin.js`. Before, they were dynamically injected when the JS was loaded (or called?), but now they're split.

They either need to be included with `import "node_modules/spin.js/style.css"`, or this methods. I don't think our webpack setup allows for important stylesheets, so I just added it here, as it was the most logical place.

js/src/common/index.js
davwheat: Me too! v2 has an even nicer API too!

```js
import 'expose-loader?exposes=$,jQuery!jquery';
```

askvortsov1: > Me too! v2 has an even nicer API too, but that's Webpack 5 only :(

We'll need to update to Webpack 5 for the automated export stuff. Supporting it isn't difficult at all either, see https://github.com/flarum/core/compare/as/support-webpack-5

less/common/LoadingIndicator.less
askvortsov1: Can we skip updating spin.js then? I'd rather revisit it as a whole in #2345

askvortsov1 (approved):

davwheat:
Fixed the conflicts from `clsx` PR

tankerkiller125 (approved):

davwheat:
I just suddenly noticed that I managed to accidentally keep spin.js updated! Must have forgotten to push the changes. I'll quickly push that and merge.

PR #2755, opened 2021-04-04
Fix registering custom searchers, allow searchers without fulltext 

### LINKED ISSUE #2712: Extensions cannot register custom Searcher classes
## Bug Report

**Current Behavior**
Beta 16 introduced the new Search API, including the `AbstractSearcher` class and container-bound gambits.

It's not made super clear in the documentation, but just like for custom frontends, it's assumed that the developer needs to register at least one gambit via `Extend\SimpleFlarumSearch` for the container bindings to be set up for a new searcher class.

It's also not super clear in the documentation but clear from the source code that a full text gambit must be registered as it's the basis for the `SearchServiceProvider` to register the bindings.

However registering a full text gambit does not sets up the bindings, and as a result the Searcher class cannot be resolved through the container.

The issue is actually that the full text gambits for new searcher are not saved in the container due to `resolving` used instead of `extend`.

**Steps to Reproduce**
1. Create custom `Searcher` class extending `AbstractSearcher`
2. Register class with full text gambit `(new Extend\SimpleFlarumSearch(MyCustomSearcher::class))->setFullTextGambit(AFullTextGambit::class)`
3. Try resolving `MyCustomSearcher`
4. Observe error `Illuminate\Contracts\Container\BindingResolutionException: Unresolvable dependency resolving [Parameter #1 [ <required> array $searchMutators ]] in class Flarum\Search\AbstractSearcher in /home/clark/Projects/flarum-beta16/vendor/illuminate/container/Container.php:1057`

A dump at line 67 of `SearchServiceProvider` confirms there is no entry for the new searcher.

**Environment**
- Flarum version: beta 16

**Possible Solution**

`resolving` needs to be replaced with `extend`. Tests need to be added for custom searcher classes.

askvortsov1:
**Fixes #2712**

- Fix registering custom searchers
- Add integration tests
- Require FulltextGambit as constructor arg for GambitManager

**Confirmed**

### DIFF
--- a/src/Search/SearchServiceProvider.php
+++ b/src/Search/SearchServiceProvider.php
@@ -74,7 +74,9 @@ class SearchServiceProvider extends AbstractServiceProvider
                 }
             }
 
-            throw new \RuntimeException('You cannot add gambits to searchers that do not have fulltext gambits. The following searchers have this issue: '.implode(', ', $affectedGambits));
+            if (count($affectedGambits)) {
+                throw new \RuntimeException('You cannot add gambits to searchers that do not have fulltext gambits. The following searchers have this issue: '.implode(', ', $affectedGambits));
+            }
         }
 
         foreach ($fullTextGambits as $searcher => $fullTextGambitClass) {
--- a/src/Search/SearchServiceProvider.php
+++ b/src/Search/SearchServiceProvider.php
@@ -68,8 +68,7 @@ class SearchServiceProvider extends AbstractServiceProvider
                 ->when($searcher)
                 ->needs(GambitManager::class)
                 ->give(function () use ($searcher, $fullTextGambitClass) {
-                    $gambitManager = new GambitManager();
-                    $gambitManager->setFulltextGambit($this->container->make($fullTextGambitClass));
+                    $gambitManager = new GambitManager($this->container->make($fullTextGambitClass));
                     foreach (Arr::get($this->container->make('flarum.simple_search.gambits'), $searcher, []) as $gambit) {
                         $gambitManager->add($this->container->make($gambit));
                     }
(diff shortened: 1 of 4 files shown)


SychO9 (commented):
> if there doesn't exist a fulltext gambit for a model, then it's not really "searchable", and as such, a filterer should be used.

I agree, extension developers might try to use gambits when what they really need is filters instead. Throwing an exception somewhere -if we can- that makes it clear would be great as well.

src/Search/SearchServiceProvider.php
```diff
@@ -58,19 +58,27 @@ public function register()
...
+
+        $searchersWithGambitsWithoutFulltext = array_diff(array_keys($gambits), array_keys($fullTextGambits));
+
+        if (count($searchersWithGambitsWithoutFulltext)) {
+            $affectedGambits = [];
+            foreach ($searchersWithGambitsWithoutFulltext as $searcher) {
+                $affectedGambits += $gambits[$searcher];
+            }
+
+            throw new \RuntimeException('You cannot add gambits to searchers that do not have fulltext gambits. The following searchers have this issue: '.implode(', ', $affectedGambits));
```
clarkwinkelmann, line 72: I'm worried about this for "no-op" extender calls.

Suppose I have an extension that registers a new searcher.

Another extension has that extension as an optional dependency and inject an additional gambit to it.

Now if the original extension is not present, an error will be thrown instead of the extender just having no effect, all for a searcher that just won't be used anyway.

Maybe add a check for `class_exists($searcherClass)`, or do this test inside of the `give()` binding? It will just require merging the two searcher keys from the two array so that all are looped and bound.

clarkwinkelmann (commented):
I have not tested locally, but this looks good to me.

I was worried that gambits might be resolved too early, but that doesn't seem to be an issue. The array is resolved in `boot`, but the actual classes are only `container->make()` inside of the container binding so it's fine.

My only issue is the comment below.

askvortsov1:
Reverted the exception as I was unable to catch it.

SychO9 (approved):
Had a similar problem with trying to catch an exception from tests, it just wasn't possible.

tankerkiller125 (approved):

Issue #2719, opened 2021-03-20
[Testing] TestCase::extension() causes core extension extenders to run twice, fails tests


clarkwinkelmann:
## Bug Report

**Current Behavior**
Calling `$this->extension('kilowhat-formulaire')` in my integration tests causes the error `RuntimeException: Route tags.index on method GET already exists`. This is because when extensions are registered through `$this->extension()`, Flarum runs all extenders for all extensions again. Since core extensions that might be installed are already marked enabled in the database as part of the `setup.php` script, they run both in `ExtensionServiceProvider` and again in `OverrideExtensionManagerForTests`

**Steps to Reproduce**
1. Setup community extension integration tests according to https://docs.flarum.org/extend/testing.html
2. Install `flarum/tags` as a dependency of the community extension
3. Create integration test with `$this->extension('community-extension')` in the `setUp`
4. Run `composer setup`
5. Run `composer test`

Observe tests erroring with error `RuntimeException: Route tags.index on method GET already exists`.

**Expected Behavior**
See possible solutions.

**Environment**
- Flarum version: beta 16

**Possible Solution**
Solution 1: we don't enable any core extension in the database level, so that `ExtensionServiceProvider` don't try running any extender, and require the developer to specify all extensions including core extensions they need in `TestCase::extension()`

Solution 2: in `OverrideExtensionManagerForTests`, don't run the extenders for extensions already enabled in the database. It's not easy because that class actually cannot control that, it just calls `ExtensionManager::extend()`

A alternative to solution 2 would be to somehow make the change to enabled extensions before `ExtensionServiceProvider` runs, so that we don't need to run and extender inside `OverrideExtensionManagerForTests`

**Additional Context**
Encountered with my Formulaire extension that depends on flarum/tags.

askvortsov1:
I like Solution 1. We should be able to set that to an empty array as an additional setup step, maybe?

clarkwinkelmann:
The only challenge with solution 1 is that the code responsible for white-listing default extensions is part of our installation script for Flarum itself, separate from the tests setup. It's just called by the tests setup.

I've always thought we needed to make the whitelist of extensions configurable somehow, that would be useful for custom Flarum skeletons that could include other bundled extensions. The same configurable option in the installer could be used by the tests.

askvortsov1:
> The only challenge with solution 1 is that the code responsible for white-listing default extensions is part of our installation script for Flarum itself, separate from the tests setup. It's just called by the tests setup.
> 
[...]

PR #2757, opened 2021-04-04
Allow configuring default enabled extensions as part of installation

askvortsov1:
This is needed for the testing library

**Part of #2719**

Allow configuring default enabled extensions as part of installation. Tested locally.

### DIFF
--- a/src/Install/Steps/EnableBundledExtensions.php
+++ b/src/Install/Steps/EnableBundledExtensions.php
@@ -22,6 +22,23 @@ use League\Flysystem\Filesystem;
 
 class EnableBundledExtensions implements Step
 {
+    const EXTENSION_WHITELIST = [
+        'flarum-approval',
+        'flarum-bbcode',
+        'flarum-emoji',
+        'flarum-lang-english',
+        'flarum-flags',
+        'flarum-likes',
+        'flarum-lock',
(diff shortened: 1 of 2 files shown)


src/Install/Steps/EnableBundledExtensions.php
```diff
@@ -37,11 +37,17 @@ class EnableBundledExtensions implements Step
      */
     private $assetPath;
 
-    public function __construct(ConnectionInterface $database, $vendorPath, $assetPath)
+    /**
+     * @var string[]
```
SychO9, line 41: `string[]|null`

src/Install/Steps/EnableBundledExtensions.php
```diff
@@ -109,7 +115,7 @@ private function loadExtensions()
 
                 return $extension;
             })->filter(function (Extension $extension) {
-                return in_array($extension->getId(), self::EXTENSION_WHITELIST);
+                return in_array($extension->getId(), $this->enabledExtensions);
```
SychO9, line 118: Since we're now using the constant in the constructor, could we please move the declaration of the constant to the top of the class ?

SychO9 (approved):

tankerkiller125 (approved):

PR #2759, opened 2021-04-05
Switch to ICU MessageFormat

### LINKED ISSUE #2062: Javascript Translator interval/explicit syntax is broken
## Bug Report

**Current Behavior**
The javascript Translator implements the "interval" syntax of the Symfony translator (called "explicitRules" internally). However the implementation is broken and calling `transChoice` fails with a `t is undefined` error.

**Steps to Reproduce**
1. In an extension, register a translation with value `{0} There are no apples|{1} There is one apple|]1,Inf] There are {count} apples`
2. Try showing that translation via `app.translator.transChoice(theKey, count, {count})
3. Interface breaks and console shows `t is undefined`

**Expected Behavior**
A translation should be chosen based on the count passed.

Additionally a clear error should be thrown if no translation can be found.

**Environment**
- Flarum version: beta 12, applies to all betas since the Translator was introduced

**Possible Solution**
The cause of the issue is that an object is incorrectly looped through as an array in https://github.com/flarum/core/blame/master/js/src/common/Translator.js#L112 , which means none of the options are evaluated. `Translator.pluralize()` ends up returning `undefined`, which breaks the `Translator.apply()` method.

Additionally the translation is `return`ed inside the `forEach()` loop, but not from the function. So even if `explicitRules` was correctly used as an array the `pluralize` function would still return `undefined`.

I have implemented a patch in https://github.com/clarkwinkelmann/flarum-ext-who-read/blob/master/js/src/forum/patchTranslatorPlural.js while solving https://github.com/clarkwinkelmann/flarum-ext-who-read/issues/7

I will prepare a patch for core with some additional changes, in particular:

- Proper error messages are needed
- The regexes must be commented and the groups should be named to make the code readable

**Additional Context**
As indicated by the comment at the top of `Translator.js`, the code is based on https://github.com/willdurand/BazingaJsTranslationBundle/blame/master/Resources/js/translator.js as was likely mostly copy-pasted by Toby 5 years ago.

However `BazingaJsTranslationBundle` never used an array, meaning the error was likely introduced by Tobdy when adapting the code for Flarum.

The reason no one ever noticed is because neither core neither any third-party extension ever attempted to use interval-based translations. Core only uses index-based plural and [few extensions](https://query.flarum.dev/q/110a1e09-021f-4dbd-b96e-37bccdf0c40c) use plural at all.

### LINKED ISSUE #2745: transChoice broken in PHP translator
## Bug Report

The `transChoice` method was removed from the Symfony translator in v5, breaking usage in all backend code.

See https://github.com/flarum/core/issues/528#issuecomment-809923310 for discussion on fully switching to the new ICU MessageFormat as an alternative.

askvortsov1:
**Foundation for flarum/issue-archive#385**
**Closes flarum/framework#2745**
**Closes flarum/framework#2062 as moot**

Done:
- Support `+intl-icu` files in locales and language packs for backend translator
- Add integration tests for locale extender (can't for language pack extender, as that can only run in an extension)

TODO:
- Wait for https://github.com/ultraq/icu-message-formatter/pull/10, https://github.com/ultraq/icu-message-formatter/pull/9 to be merged (if they will be, but feels likely)
- If https://github.com/ultraq/icu-message-formatter/pull/11 isn't merged, we'll need to include it ourselves. I would prefer to put it in a separate package for now, as I'm not comfortable including it without the tests. Parsers tend to be ugly :laughing:

- Should `+intl-icu` be the default for ALL files? This might have slight performance ramifications, but it would also be more consistent (see expected results of `custom_translation_exists_if_added` vs `custom_translation_does_not_exist_by_default` in the backend tests). This would be a further breaking change though. But it would be more consistent with what's done on the frontend. Currently in the PR.
- We still need a translation args pre-processing system, but I think that should go in a second PR.

**Confirmed**


davwheat:
I'm for the first option too. Smaller bundle! 👍

tankerkiller125:
Maybe we should include `+intl-icu` as a prefix by default? That way language packs don't have too? But I'm not 100% certain on that one.

askvortsov1:
> Maybe we should include +intl-icu as a prefix by default? That way language packs don't have too? But I'm not 100% certain on that one.

This might hurt performance, as MessageFormat is more complex than non-MessageFormat. However, consistency would be a benefit: we would no longer need to include curly braces in variable keys. Plus, we're using MessageFormat code in the frontend anyway.

askvortsov1:
> I'm for the first option too. Smaller bundle!

I got it working for 95% of what we need via https://github.com/ultraq/icu-message-formatter/issues/7, other needed changes are outlined in the same issue. If that isn't merged, we can just include the code from the PR in our repo; this library is very extensible!

askvortsov1:
After some more thought, why don't we ise `intl+icu` for everything, and transform backend keys to remove wrapping curly braces if present as a BC layer?

askvortsov1:
This is ready for review, but we can't merge until the items in TODO are implemented over in https://github.com/ultraq/icu-message-formatter

askvortsov1:
No longer WIP, this is now ready for review.

luceos (approved):
This looks awesome, but please involve our translators if this hasn't happened yet (a discussion under the i18n - https://discuss.flarum.org/t/i18n - tag would be smart. At least they need to know in advance because this will be NBC for all translations.

askvortsov1:
https://discuss.flarum.org/d/27030-upcoming-translation-changes

tankerkiller125 (approved):
Looks good

PR #2766, opened 2021-04-05
Remove unneeded vendor prefixes

### LINKED ISSUE #2762: Remove unneeded vendor Less mixins
We have a lot of mixins in our less that we don't really have a need for anymore, specifically the vendor ones.

I propose that we scrap what we can, and keep the mixins for backwards compatibility purposes.

In my testing...

**With prefixes:** 29937 bytes
**Removed:** 29462 bytes

...totalling 475 bytes of savings within core. That's 0.5 GB for every million page views.

**Here's what we can remove:**

- animations
  - Supported without prefixes from Safari 9+
- backface visibility
  - Need to keep `webkit` prefix for Safari, but `moz` can go
- box shadow
- box sizing
- content columns
  - Safari needs it below iOS 10, so we should probably keep the `webkit` prefix. I don't think we even use this anyway.
- hyphens
  - Only keep `webkit`
- placeholder
  - Add new CSS pseudo-element (`::placeholder`)
  - Remove all but `webkit` prefix for Safari below iOS 10
- transforms
- perspectives
- transitions
- user-select
  - Keep `webkit` for Safari and new Edge

davwheat:
**Fixes #2762**

- Removes pointless vendor prefixes that do nothing as we don't support the browsers that they provide support for.
- I've also bundled in a fix for the post scrubber where it wasn't using the `-webkit-` prefix for `user-select`

**Confirmed**

### DIFF
--- a/js/src/forum/components/PostStreamScrubber.js
+++ b/js/src/forum/components/PostStreamScrubber.js
@@ -112,7 +112,6 @@ export default class PostStreamScrubber extends Component {
 
       // Now we want to make the scrollbar handle draggable. Let's start by
       // preventing default browser events from messing things up.
-      .css({ cursor: 'pointer', 'user-select': 'none' })
       .bind('dragstart mousedown touchstart', (e) => e.preventDefault());
 
     // When the mouse is pressed on the scrollbar handle, we capture some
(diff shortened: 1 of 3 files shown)


askvortsov1 (approved):
Haven't tested, but changes make sense. When merged, this should be merged via merge commit, since changes are granular. :+1:

PR #2771, opened 2021-04-07
Require unique route names

### LINKED ISSUE #2707: Require unique name + method for routes per-app
Followup to https://github.com/flarum/core/pull/2577#discussion_r572228376

This is a prerequisite for https://github.com/flarum/core/issues/2474, as we need to be able to uniquely obtain the controller with just the route name.

SychO9:
**Fixes #2707**

Require route names to be unique instead of a unique couple [name, method].

I don't think we remembered to warn extension devs about this change, but I didn't add a BC layer, I think it's fine to go through with this.

**Confirmed**

### DIFF
--- a/src/Http/RouteCollection.php
+++ b/src/Http/RouteCollection.php
@@ -73,34 +73,32 @@ class RouteCollection
 
     public function addRoute($method, $path, $name, $handler)
     {
-        if (isset($this->routes[$method][$name])) {
-            throw new \RuntimeException("Route $name on method $method already exists");
+        if (isset($this->routes[$name])) {
+            throw new \RuntimeException("Route $name already exists");
         }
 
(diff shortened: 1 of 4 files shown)


src/Http/RouteCollection.php
```diff
@@ -73,34 +73,32 @@ public function delete($path, $name, $handler)
 
     public function addRoute($method, $path, $name, $handler)
```
askvortsov1, line 74: Perhaps we should mark all these methods as internal? Or even the entire class?

askvortsov1 (approved):

tankerkiller125 (approved):

src/Http/RouteCollection.php
SychO9: Yeah, the whole class shouldn't be considered public API. We have extenders for this.

PR #2773, opened 2021-04-07
Fix relevance sort

### LINKED ISSUE #2772: Discussion relevance ordering is broken
## Bug Report

**Current Behavior**
The beta 16 search system seems to have broken searching by relevance. Since we moved `$defaultSort` out of the search state and into the controller, a `$sort` value is always passed for discussion queries. But there are several places that used the nullity of `$sort` to check whether the user has specified a sort in the query:

- https://github.com/flarum/core/blob/master/src/Query/ApplyQueryParametersTrait.php#L24
- https://github.com/flarum/sticky/blob/master/src/PinStickiedDiscussionsToTop.php#L34

The latter of these is a hacky workaround; the former is broken and means that relevance sorts won't work, and a simple "latest" sort will always be used instead.

**Steps to Reproduce**
1. Go to https://discuss.flarum.org
2. Search for anything without specifying sort
3. See that results are always ordered by "latest"

**Possible Solution**
The old system was confusing and IMO, had a poor separation of responsibilities. Eliminating per-resource search state classes was a solid step, but we need to clean it up a bit. I would argue that `QueryCriteria` should take a 4th argument `$sortInRequest`, which would default to `false`. If a sort is specified in the query, that would be set to `true`. That bit of data:

- Would be accessible to the sticky logic, since it's part of `QueryCriteria`
- Could be passed to `applySort`, via a 3rd parameter defaulted to `false`.

While we're at it, I think we should also mark the entire `ApplyQueryParametersTrait` as `@internal`.

askvortsov1:
**Fixes #2772**

- Adds a field to `QueryCriteria` that determines whether the sort provided is the controller's default sort
- Set this field to `true` iff `sort` not in query params. Default it to false
- Override $sort if a new default sort has been set on search state, and the param is true.
- Add tests!

**Confirmed**

### DIFF
--- a/src/Api/Controller/AbstractSerializeController.php
+++ b/src/Api/Controller/AbstractSerializeController.php
@@ -12,6 +12,7 @@ namespace Flarum\Api\Controller;
 use Flarum\Api\JsonApiResponse;
 use Illuminate\Contracts\Container\Container;
 use Illuminate\Database\Eloquent\Collection;
+use Illuminate\Support\Arr;
 use Illuminate\Support\Str;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
@@ -261,7 +262,7 @@ abstract class AbstractSerializeController implements RequestHandlerInterface
      */
     protected function sortIsDefault(ServerRequestInterface $request)
     {
-        return ! array_key_exists('sort', $request->getQueryParams());
+        return !Arr::get($request->getQueryParams(), 'sort');
     }
 
     /**
--- a/src/Api/Controller/AbstractSerializeController.php
+++ b/src/Api/Controller/AbstractSerializeController.php
@@ -256,6 +256,15 @@ abstract class AbstractSerializeController implements RequestHandlerInterface
     }
 
     /**
+     * @param ServerRequestInterface $request
+     * @return bool
+     */
+    protected function sortIsDefault(ServerRequestInterface $request)
+    {
+        return !array_key_exists('sort', $request->getQueryParams());
+    }
+
+    /**
      * Set the serializer that will serialize data for the endpoint.
      *
      * @param string $serializer
--- a/src/Api/Controller/AbstractSerializeController.php
+++ b/src/Api/Controller/AbstractSerializeController.php
@@ -256,11 +256,7 @@ abstract class AbstractSerializeController implements RequestHandlerInterface
         return new Parameters($request->getQueryParams());
     }
 
-    /**
-     * @param ServerRequestInterface $request
-     * @return bool
-     */
-    protected function sortIsDefault(ServerRequestInterface $request)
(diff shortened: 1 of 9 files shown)


src/Api/Controller/AbstractSerializeController.php
```diff
@@ -255,6 +256,15 @@ protected function buildParameters(ServerRequestInterface $request)
         return new Parameters($request->getQueryParams());
     }
 
+    /**
+     * @param ServerRequestInterface $request
+     * @return bool
+     */
+    protected function sortIsDefault(ServerRequestInterface $request)
```
SychO9, line 263: we can explicitly type-hint the return value.

src/Api/Controller/ListPostsController.php
```diff
@@ -78,12 +78,13 @@ protected function data(ServerRequestInterface $request, Document $document)
...
         $filters = $this->extractFilter($request);
         $sort = $this->extractSort($request);
+        $sortIsDefault = $this->sortIsDefault($request);
 
         $limit = $this->extractLimit($request);
         $offset = $this->extractOffset($request);
         $include = $this->extractInclude($request);
 
-        $results = $this->filterer->filter(new QueryCriteria($actor, $filters, $sort), $limit, $offset);
+        $results = $this->filterer->filter(new QueryCriteria($actor, $filters, $sort, $sortIsDefault), $limit, $offset, );
```
SychO9, line 87: missed unneeded comma at the end.

src/Api/Controller/ListUsersController.php
```diff
@@ -85,12 +85,13 @@ protected function data(ServerRequestInterface $request, Document $document)
 
         $filters = $this->extractFilter($request);
         $sort = $this->extractSort($request);
+        $sortIsDefault = $this->sortIsDefault($request);
```
SychO9, line 88: can we rename `sortIsDefault` to `isDefaultSorting` or maybe `isSortingByDefault`.

SychO9 (changes requested):

src/Api/Controller/ListUsersController.php
askvortsov1: Mixed feelings about this one. Any use of sort**ing** describes whether or not **any** sort is being applied. But this variable should describe whether the sort currently being used (whatever that is, could be no sort at all) is the controller's default sort.

Alternatively, we could use `userRequestedSort` or `userSpecifiedSort` and invert everything, but that sounds very verbose.

SychO9: True, we only want to know if we're falling back to the default sort.

Though I was more looking for just a `isSomething` naming, since that's the usual convention for boolean variables/methods where possible. But `isDefaultSort` might be weird.

Let's keep it as is then!

SychO9 (approved):

tankerkiller125 (approved):

PR #2775, opened 2021-04-09
Adjust search height on resize

askvortsov1:
Identified as a potential issue in https://github.com/flarum/core/pull/2650

When typing, the keyboard generally obstructs half the screen. However, when the keyboard is closed, search results don't expand to take up full space.


davwheat (approved):
Looks good to me!

tankerkiller125 (approved):

PR #2777, opened 2021-04-10
Move Powered By Header to headers config

tankerkiller125:
Moves the Powered By Header to to the headers config to match the new security headers PR #2721

### DIFF
--- a/src/Http/Middleware/FlarumPromotionHeader.php
+++ b/src/Http/Middleware/FlarumPromotionHeader.php
@@ -22,7 +22,7 @@ class FlarumPromotionHeader implements Middleware
 
     public function __construct(Config $config)
     {
-        $this->enabled = Arr::get($config,'headers.poweredByHeader') ?? true;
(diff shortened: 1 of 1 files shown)


src/Http/Middleware/FlarumPromotionHeader.php
```diff
@@ -21,7 +21,7 @@ class FlarumPromotionHeader implements Middleware
 
     public function __construct(Config $config)
     {
-        $this->enabled = $config['poweredByHeader'] ?? true;
+        $this->enabled = $config['headers']['poweredByHeader'] ?? true;
```
askvortsov1, line 24: What if there is no headers key? Should we use `Arr::get`?

tankerkiller125: That's a good point, I'll fix that here and the security header PR

askvortsov1 (approved):

luceos (approved):

PR #2778, opened 2021-04-10
[WIP] Reverse Proxy Handling

### LINKED ISSUE #174: New, fixed version of reverse proxy PR
Let's try to restore https://github.com/flarum/core/pull/2236/files. For the initial implementation, we can keep the `Arr::get($request->getServerParams(), 'REMOTE_ADDR', '127.0.0.1');` implementation currently present in post / discussion code; we can add a more rigorous system in the future. We can also inject config into the middleware constructor directly, so no need for a singleton definition

This also centralizes IP processing, so that we don't need to retrieve it in 20 different places. It would also then become available to throttling middleware.

### LINKED ISSUE #2416: New, fixed version of reverse proxy PR

tankerkiller125:
**Fixes flarum/issue-archive#174**

**Screenshot**

**Confirmed**

**Required changes:**

### DIFF
--- a/src/Http/Middleware/ProcessIp.php
+++ b/src/Http/Middleware/ProcessIp.php
@@ -23,8 +23,6 @@ class ProcessIp implements Middleware
         // If no header set via the proxy headers, use the IP given by the web server
         $ipAddress ?? Arr::get($request->getServerParams(), 'REMOTE_ADDR', '127.0.0.1');
 
-        var_dump($ipAddress);
-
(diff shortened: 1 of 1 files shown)


askvortsov1 (changes requested):
Thinking back to Franz's original review, we should probably have tests for this. Also, are there no libraries that we could depend on to implement this?

Issue #2397, opened 2020-10-16
Improve discussion page near canonical url


w-4:
## Bug Report

**Current Behavior**
In beta.13 and beta.14, there is a canonical url included in the hmtl document. 
For example for  https://discuss.flarum.org/d/187-word-association-game/1000 
the canonical url is https://discuss.flarum.org/d/187-word-association-game?near=1000

As per https://github.com/flarum/core/issues/189#issuecomment-709429091 this causes search engines to crawl that URL.

Edit: the redirect works, my link was just broken.

w-4:
We could do it the way the Discourse does. They also use a postnumber for the URL, but use a `page` parameter in the canonical url. 
As an example:
For the first 20 posts the canonical would be  https://discuss.flarum.org/d/187-word-association-game
for the next 20, it would be https://discuss.flarum.org/d/187-word-association-game?page=2

That is probably better for SEO, because of less links and no overlap.

davwheat:
I'm really confused by this. What's the issue here?

Using the `/d/1233456/100` syntax could make search engines think it's a totally separate page from `/d/123456` causing page duplication.

We should keep that canonical to prevent that, but we should also add `rel=next` to link to the page before that.

It might be work replacing this canonical with a `?page=x`-based one, though, as the `?near=1000` seems to take me to post 950...

enricoeur:
This is big mistake. Really i dont understand why was not fixed...

askvortsov1:
> I'm really confused by this. What's the issue here?
> 
> Using the `/d/1233456/100` syntax could make search engines think it's a totally separate page from `/d/123456` causing page duplication.
> 
> We should keep that canonical to prevent that, but we should also add `rel=next` to link to the page before that.
> 
> It might be work replacing this canonical with a `?page=x`-based one, though, as the `?near=1000` seems to take me to post 950...

I like using `page=x` as the canonical URL. Post numbers can be messy.

I'll also note that the current no-js approach is missing next page / prev page buttons. THAT is quite problematic, and something that definitely needs to be fixed before stable.

askvortsov1:
When redesigning, we should also make sure to fix https://github.com/flarum/core/issues/2130

askvortsov1:
https://github.com/flarum/core/blob/f2168f939db15f58c1511b56b4d95c737b62ce1a/views/frontend/content/discussion.blade.php#L18-L24 is incorrect, because there won't be prev/next links on the API document for a single item. That's what's causing the missing pagination links.

PR #2781, opened 2021-04-11
Create abstract PaginatedListState for DiscussionList and others

dsevillamartin:
**Refs flarum/issue-archive#162**

Look at `// TODO` comments in code 

- What does the *current page* track?
- Do we want the abstract state tied to store models?
- How do we want to treat `this.pages`?
- Logic to set when loading/done loading needs to be revised
- Calling `refresh` sets page location to pg 1, this needs to be changed (refresh is called on page load, will replace params or whatever)
- The `params` implementation in general needs clarification - do we want like DiscussionListState or something else?

**Confirmed**

**Required changes:**


js/src/common/states/PaginatedListState.ts
```diff
@@ -0,0 +1,171 @@
...
+  startIndex?: number;
+  endIndex?: number;
+}
+
+export default abstract class PaginatedListState<T extends Model> {
+  // TODO:
+  //  - what does this page track? last page request? last page that is loaded
+  //      (e.g. pg 7 if u have 3-7 loaded)
+  //  - When do we set it?
+  protected location!: PaginationLocation;
```
askvortsov1, line 27: `pages` is a contiguous structure that expands from both ends. So it could either be an array or a doubly linked list. If it's a doubly linked list, `location`'s `page` prop should point to the active element. If it's an array, `location`'s `page` prop should be the index of the active page (although we'd need to remember to increment by 1 when adding new pages.

js/src/common/states/PaginatedListState.ts
```diff
@@ -0,0 +1,171 @@
...
+
+  // TODO: might want interface so pagination can work w/o model-specific stuff
+  protected type: string;
```
askvortsov1, line 30: Well, I suppose it could be any JSON:API resource. Frontend models represent those (not necessarily a db model), so requiring a model should be fine.

js/src/common/states/PaginatedListState.ts
```diff
@@ -0,0 +1,171 @@
...
+  protected initialLoading: boolean = false;
+  protected loadingPrev: boolean = false;
+  protected loadingNext: boolean = false;
+
+  hasPrev: boolean = false;
+  hasNext: boolean = false;
+
+  // TODO: this constructor is for class that implements
+  // what data do we want passed?
+  protected constructor(type: string, offset: number = 20) {
```
askvortsov1, line 45: If this is an abstract class, shouldn't type be static?

We could also do an initial page of data, so that if there's stuff in the preloaded API doc we can reuse that without immediately triggering an API call.

js/src/common/states/PaginatedListState.ts
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  protected parseResults(pageNum, results: T[]) {
+    const page = {
+      number: pageNum,
+      items: results,
+    };
+
+    // TODO: is this how we were thinking of treating this.pages ?
+    if (this.isEmpty() || pageNum > this.getNextPageNumber() - 1) {
```
askvortsov1, line 57: Makes sense to me!

js/src/common/states/PaginatedListState.ts
```diff
@@ -0,0 +1,171 @@
...
+      this.pages.push(page);
+    } else {
+      this.pages.unshift(page);
+    }
+
+    this.hasNext = !!results.payload?.links?.next;
+    this.hasPrev = !!results.payload?.links?.prev;
+
+    // TODO move loading logic to better place
+    this.initialLoading = false;
```
askvortsov1, line 67: This isn't that bad a place tbh

js/src/common/states/PaginatedListState.ts
```diff
@@ -0,0 +1,171 @@
...
+    const page = {
+      number: pageNum,
+      items: results,
+    };
+
+    // TODO: is this how we were thinking of treating this.pages ?
+    if (this.isEmpty() || pageNum > this.getNextPageNumber() - 1) {
+      this.pages.push(page);
+    } else {
+      this.pages.unshift(page);
```
askvortsov1, line 60: If location.page is a number it'd need to be incremented by 1 when unshifting, unless the page being added is the first page.

askvortsov1 (commented):
> The params implementation in general needs clarification - do we want like DiscussionListState or something else?

I think `DiscussionListState`'s implementation makes sense, with one exception: in https://github.com/FriendsOfFlarum/user-directory/commit/bc4ffe801c1d4feac5f01aa4b05fbcd787a65e91, Ian and I added a "qBuilder", which makes it easier to assemble the `q` query string. I think that's something we might want to move over into the official implementation.

On a side note, to keep things simple, we might want to do this in 2 phases:

1. Swap out `DiscussionListState` for `PaginatedListState`, implement the url pagination
2. Possibly in a separate release (since #2397 is still on the milestone), swap out `DiscussionList` for an abstract `PaginatedList`

js/src/common/states/PaginatedListState.ts
datitisev: What are we defining with "active" page though? If the user loads the previous page, does the active page still remain the last number in ascending order (e.g. loaded pg 2 but pg 5 is still active) or does it change to page 2?

js/src/common/states/PaginatedListState.ts
datitisev: > shouldn't type be static?

That makes more sense, yes. 

> We could also do an initial page of data, so that if there's stuff in the preloaded API doc we can reuse that without immediately triggering an API call.

Hm this might work better as another method that is called from the subclass constructor 🤔

js/src/common/states/PaginatedListState.ts
askvortsov1: Would the previous page be the page that the user is currently on?

datitisev: What are you referring to with previous page? I meant as load next page, load prev page, for loading a page that hasn't been loaded.

askvortsov1: Active page, sorry.

js/src/common/states/PaginatedListState.ts
datitisev: What would we need to do this for? Sorry, can't follow the logic right now, could you explain it?

js/src/common/states/PaginatedListState.ts
datitisev: Static properties that are overriden in subclasses is a bit more scuffed in JS than PHP.

Do we *need* a static class or do we want to use an abstract getter, for example (wouldn't be static)
image

askvortsov1: Dont think it matters that much, abstract method or getter is probs fine.

js/src/common/states/PaginatedListState.ts
askvortsov1: Just saying we need to update the location page number when pages are added to the start.

js/src/common/states/PaginatedListState.ts
datitisev: Currently got is as static. Will probs change to getter then (will also make TS error disappear)

js/src/common/states/PaginatedListState.ts
datitisev: Ah I see, makes sense. We'd be updating the location page number to the new pages that have been added at the beginning, right?

datitisev:
Regarding the handling of query parameters, some thoughts from Sasha:

- A default params method is good 
- As is a method to assemble params before sending (since we'd like to add in a qBuilder field)
 - e.g. `requestParams` in `DiscussionListState.js`
- So the params object will be in a "construction" form, not fully in the format it'd be sent in, for some of the methods and for internal storage

askvortsov1:
Looked at this again. `DiscussionListState` only has 2 methods: one for getting the underlying params object, the other for converting that into the format expected by JSON:API. I think having these 2 methods is sensible for `PaginatedList`. It shouldn't need to worry about global / URL query params IMO.

Have you tried plugging this into the `NotificationList`? Does it work as well as with `DiscussionList`?

datitisev:
> Have you tried plugging this into the `NotificationList`? 

I have not. I will try to do so today.

datitisev:
- How do we feel about the `getAllItems` method in https://github.com/flarum/core/pull/2781/commits/c7d129bbccaf556430d6cb0388a6f61c73763483?
- How do we like the `DiscussionListState#removeDiscussion` implementation in e01ab57?
- How do we want to handle `DiscussionListState#addDiscussion`
  - Do we want an extra field for the added discussions that is cleared on refresh?
    That way we don't do weird stuff with pagination.
    We can have an extra method that adds the extra field to the array of discussions returned.
- Do we want getter methods for `hasPrev`, `hasNext`? (I assume yes)

js/src/common/states/PaginatedListState.ts
askvortsov1: Yep!

js/src/common/states/PaginatedListState.ts
```diff
@@ -0,0 +1,196 @@
...
+    params.page = { offset: this.offset * (page - 1) };
+
+    if (Array.isArray(params.include)) {
+      params.include = params.include.join(',');
+    }
+
+    return app.store.find(this.type, params);
+  }
+
+  public hasItems(): boolean {
```
askvortsov1, line 66: Can we add comments explaining the difference between this and `isEmpty`?

js/src/common/states/PaginatedListState.ts
```diff
@@ -0,0 +1,196 @@
...
+
+  /**
+   * Get the parameters that should be passed in the API request.
+   * Do not include page offset unless subclass overrides loadPage.
+   *
+   * @abstract
+   * @see loadPage
+   */
+  protected requestParams(): any {
+    return {};
```
askvortsov1, line 100: This should probably use `params`.

Also, would like to see the qBuilder included into this system

js/src/common/states/PaginatedListState.ts
```diff
@@ -0,0 +1,196 @@
...
+
+  /**
+   * Update the `this.params` object, calling `refresh` if they have changed.
+   * Use `requestParams` for converting `this.params` into API parameters
+   *
+   * @param newParams
+   * @see requestParams
+   */
+  public refreshParams(newParams) {
+    if (this.isEmpty() || Object.keys(newParams).some((key) => this.requestParams()[key] !== newParams[key])) {
```
askvortsov1, line 118: Should we make this a protected `paramsChanged` method so that it's easier to add/override logic?

js/src/common/states/PaginatedListState.ts
```diff
@@ -0,0 +1,196 @@
...
+      return this.refresh();
+    }
+  }
+
+  public refresh(page: number = 1) {
+    this.initialLoading = true;
+    this.loadingPrev = false;
+    this.loadingNext = false;
+
+    m.redraw();
```
askvortsov1, line 130: Shouldn't we clear it here?

js/src/common/states/PaginatedListState.ts
```diff
@@ -0,0 +1,196 @@
...
+      })
+      .finally(() => (this.initialLoading = false));
+  }
+
+  public clear() {
+    this.pages = [];
+    m.redraw();
+  }
+
+  public loadMore() {
```
askvortsov1, line 147: `loadNext`?

askvortsov1, line 147: Let's annotate these as async

js/src/forum/components/IndexPage.js
```diff
@@ -37,7 +37,7 @@ export default class IndexPage extends Page {
       app.discussions.clear();
     }
 
-    app.discussions.refreshParams(app.search.params());
+    app.discussions.refreshParams(app.search.params(), m.route.param('page'));
```
askvortsov1, line 40: This method only takes one argument

js/src/forum/states/DiscussionListState.ts
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  /**
+   * In the last request, has the user searched for a discussion?
+   */
+  isSearchResults() {
+    return !!this.params.q;
+  }
+
+  removeDiscussion(discussion: Discussion) {
```
askvortsov1, line 67: Do we want to include the add/remove methods in the main class? I think this is a pattern we are likely to see relatively often.

js/src/forum/states/NotificationListState.ts
```diff
@@ -0,0 +1,53 @@
...
+  public isLoading(): boolean {
+    return this.isInitialLoading() || this.isLoadingNext();
+  }
+
+  /**
+   * Load the next page of notification results.
+   */
+  load(): Promise<void> {
+    if (app.session.user.newNotificationCount()) {
+      this.pages = [];
```
askvortsov1, line 24: Shouldn't this reset location too?

js/src/common/states/PaginatedListState.ts
```diff
@@ -0,0 +1,196 @@
...
+  protected params: any = {};
+
+  protected initialLoading: boolean = false;
+  protected loadingPrev: boolean = false;
+  protected loadingNext: boolean = false;
+
+  hasPrev: boolean = false;
+  hasNext: boolean = false;
+
+  protected constructor(offset: number = 20) {
```
askvortsov1, line 28: Maybe we should try to somewhat mirror https://github.com/flarum/core/blob/c119731e652fcfb96e1bb410e0550c6fbb3dc445/src/Filter/AbstractFilterer.php#L48-L48 (accept params, page size, and initial offset)?

askvortsov1 (changes requested):
> Do we want getter methods for hasPrev, hasNext? (I assume yes)

I think so, if only for the consistency with other methods

> Do we want to keep methods like hasDiscussions and empty that call to the actual methods for BC?

Yes, but let's deprecate them. Before the release, we can do a sweep and decide which deprecated stuff to drop (since the next time we'll get the chance to do so will be v2)

> Do we want an extra field for the added discussions that is cleared on refresh?

That way we don't do weird stuff with pagination.
We can have an extra method that adds the extra field to the array of discussions returned.

So when the next page comes in, we'd check if the added discussion is present, and if so, override it in the returned results? I like this idea.

> How do we feel about the getAllItems method in c7d129b?

Makes sense to me. Let's keep it protected for now though.

> How do we like the DiscussionListState#removeDiscussion implementation in e01ab57?

Also looks good to me!

I'm not seeing any code for navigating to the previous page in the component. Will this be part of the followup component PR?

js/src/forum/components/IndexPage.js
```diff
@@ -37,7 +37,7 @@ export default class IndexPage extends Page {
       app.discussions.clear();
```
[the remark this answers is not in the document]
datitisev: Not sure if we wanted to do that in this PR or not.

js/src/common/states/PaginatedListState.ts
```diff
@@ -0,0 +1,196 @@
...
+    this.offset = offset;
```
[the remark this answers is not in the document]
datitisev: Yes 🙃

js/src/common/states/PaginatedListState.ts
datitisev: You mean return `this.params` by default?

js/src/common/states/PaginatedListState.ts
datitisev: Sounds good to me.

js/src/common/states/PaginatedListState.ts
datitisev: That doesn't really make any difference though? I thought `async` was to use `await` in the body, not do denote any method returning a Promise... not sure if the build dist would change after adding that.

js/src/forum/states/DiscussionListState.ts
datitisev: Hm, we might want a separate abstract class that implements them instead. I don't like having so many methods in the main class tbh.

js/src/common/states/PaginatedListState.ts
datitisev: Would we use page # instead of initial offset?

js/src/common/states/PaginatedListState.ts
askvortsov1: Implemented by Ian and myself in https://github.com/FriendsOfFlarum/user-directory/commit/bc4ffe801c1d4feac5f01aa4b05fbcd787a65e91

Essentially, if we have 2 different places in extensions that are trying to modify the `q` parameter, it gets messy, fast. Instead, we use a qBuilder object. Keys are just unique identifiers, and the values are space-joined to obtain the actual `q` value.

js/src/common/states/PaginatedListState.ts
askvortsov1: I think page number and page size would make sense here, since that's what's in the URL. We can add an additional offset later I think

js/src/forum/states/DiscussionListState.ts
askvortsov1: Fair enough. We can tackle this one when it's actually needed.

js/src/forum/components/IndexPage.js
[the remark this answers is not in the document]
askvortsov1: Yeah probably better to put that in the abstract component PR, since that would depend on being able to load previous pages.

js/src/common/states/PaginatedListState.ts
datitisev: I'd say no, in case there's an error when fetching results?

js/src/forum/states/NotificationListState.ts
datitisev: Do we want a method to reset location perhaps? Not sure. Don't like having `this.location = { page: 1 }` a lot tbh.

js/src/common/states/PaginatedListState.ts
datitisev: Hm, I feel like we probably want a separate PR for that functionality. We probs want to discuss it a bit more outside this PR.

js/src/common/states/PaginatedListState.ts
```diff
@@ -0,0 +1,211 @@
...
+
+  public getPages() {
+    return this.pages;
+  }
+  public getLocation(): PaginationLocation {
+    return this.location;
+  }
+
+  public isLoading(): boolean {
+    return this.initialLoading || this.loadingNext;
```
askvortsov1, line 150: Why not prev?

js/src/common/states/PaginatedListState.ts
```diff
@@ -0,0 +1,211 @@
...
+      // If the calculated page number is less than 1,
+      // return 1 as the prev page (first possible page number)
+      return Math.max(pg - 1, 1);
+    } else {
+      return this.location.page;
+    }
+  }
+
+  protected paramsChanged(newParams): boolean {
+    return Object.keys(newParams).some((key) => this.requestParams()[key] !== newParams[key]);
```
askvortsov1, line 205: We should also check params.filter and params.qBuilder

askvortsov1 (changes requested):
I've given it some more thought, and I'm no longer sure that qBuilder should go here. I think it might make sense to put it in GlobalSearchState instead (which is a refactor for another day).

js/src/common/states/PaginatedListState.ts
datitisev: The way the method was being used made it be this.

If the UI shows multiple loading spinners, most likely the implementation will include one for *both* initial & prev, and then another for after (or vice versa). I guess it's confusing what the method does and this logic should probably go in the component itself.

js/src/common/states/PaginatedListState.ts
askvortsov1: Scap qBuilder as per the comment above, but `filter` should be checked.

datitisev: Wouldn't we want to use e.g. lodash's `isEqual` function for this, and not just check specific arguments? 

Disregard the fact that we removed lodash from core recently 😂

askvortsov1: Not sure. The way it is right now, IIRC filter is assembled in `requestParams`, other than that it's a shallow object. Question is whether we want to keep it as shallow (in which case this comment is unnecessary).

js/src/common/states/PaginatedListState.ts
datitisev: We only do this because of our logic with whether the previous page matches `IndexPage`, I think... to force the refresh when calling `refreshParams`.

Perhaps we should just return a loading spinner if `isInitialLoading()` is true, that'd avoid any issues I feel like...

js/src/common/states/PaginatedListState.ts
datitisev: Hm this code also assumes `this.params` is in the format of `this.requestParams()`... wouldn't that most likely not be the case? As we have the params method + `requestParams` for the actual request? 🤔

askvortsov1: `requestParams` MUST be JSON API format, but `params` can be a weird intermediate format. We could go either way on this, but I think an intermediate format might make sense, as extensions can add data to it, and then modify how that data is serialized through `requestParams`.

datitisev: Don't we have an intermediate format now? With `params` and `getParams` - tho perhaps we should rename `getParams` as that doesn't sound like a method you'd want to extend in a subclass. 

And we can't really set `this.params` to the output of the intermediate format cause it could end up in things added over and over...

Unless I'm overthinking this.

js/src/common/states/PaginatedListState.ts
datitisev: I changed it so that `isLoading` checks all 3 load states, and the components themselves use the individual ones depending on how they implement initial loading vs loading prev & next.

js/src/forum/states/DiscussionListState.ts
```diff
@@ -0,0 +1,137 @@
...
+
+    if (this.extraDiscussions.length) {
+      return [
+        {
+          number: -1,
+          items: this.extraDiscussions,
+        },
+        ...pages,
+      ];
+    }
```
datitisev, line 115: I don't love this, but I don't know how else we could do this.

This is so that `DiscussionList` can do `.getPages()` and add a page separator (as an example) that forum owners can then style.

Not sure whether the state or the component should check whether the page is empty, it's currently on the state (checks that extraDiscussions isn't empty before adding the page).

datitisev:
I found a logic error - `hasNext` and `hasPrev` are calculated using the current page, and not the earliest or latest pages loaded.

So, if there are 3 pages of discussions, and we begin by loading the 3rd, then load the 2nd, the "load more" button appears because there _are_ more after the 2nd page, even though the 3rd is already loaded.

I think the fix for this would be to move the `hasPrev` and `hasNext` values to the `Page` interface, and then use the methods to check the first & last items in `this.pages`. Thoughts?

askvortsov1:
> I found a logic error - `hasNext` and `hasPrev` are calculated using the current page, and not the earliest or latest pages loaded.
> 
> So, if there are 3 pages of discussions, and we begin by loading the 3rd, then load the 2nd, the "load more" button appears because there _are_ more after the 2nd page, even though the 3rd is already loaded.
> 
> I think the fix for this would be to move the `hasPrev` and `hasNext` values to the `Page` interface, and then use the methods to check the first & last items in `this.pages`. Thoughts?

That makes sense to me!

datitisev:
This is ready to review now. I'll try moving the unnecessary UI changes for DiscussionList to another PR so that can be discussed there (e.g. the <hr> marking page & load prev button)

js/src/common/states/PaginatedListState.ts
askvortsov1: We need to keep it, as without it, clicking the refresh button shows no UI confirnation.

askvortsov1 (approved):

SychO9 (approved):

PR #2783, opened 2021-04-12
Run API Client requests through middleware

### LINKED ISSUE #2474: Run API Throttler in API Client
## Feature Request

As discussed in https://github.com/flarum/core/pull/2170#issuecomment-733527432, it would be extremely useful for requests sent to the API Client to be run through the ThrottleApi middleware. The hard part here is obtaining the route name, which is used by throttlers to determine whether they should apply. 

I propose that instead of providing a controller class as the first argument to `Client::send`, we provide a route name. From there, the applicable controller (or other handler) can be determined. We can support passing in controllers as a BC layer through reflection (if it's a class name, we assume it's a controller, otherwise we look for a named route). This would also decouple routes from their default controllers, in case we allow routes to be overriden in the future.

askvortsov1:
**Fixes #2474**

- Add integration tests for login and registration; realized that this change might break them and they are currently very missing.
- API client now has a fluent API, and sends via request methods using a path.

**Confirmed**


src/Api/Client.php
```diff
@@ -39,46 +41,69 @@ class Client
...
+            public function __construct(RouteCollection $routeCollection)
+            {
+                $this->routeCollection = $routeCollection;
+            }
+
+            public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
+            {
+                $route = $this->routeCollection->getRoutes()[$request->getAttribute('routeName')];
+
+                $request = $request->withMethod($route['method']);
```
askvortsov1, line 71: We could also set the URI path from `$route['path']`

askvortsov1:
Moved the pipe construction into the service provider, this feels a lot cleaner to me.

askvortsov1:
@clarkwinkelmann I'm a bit conflicted on how to proceed here with regards to https://github.com/flarum/core/issues/2800.

Just adding the "Start Session" middleware won't fix it, because the session that would get started is different from the one from the original request.

We could add `session` as an argument, but that would require a lot of code changes, and as you mentioned, wouldn't scale well to other changes.

One idea I had was allowing the entire "parent" request to be passed in as an optional argument after $actor, and then ApiClient could transfer over certain attributes (like session). Actor could still be provided, but if it wasn't, maybe we could read that from the parent request too?

askvortsov1:
Tests are failing due to the default route setting in `DefaultRouteTest`

SychO9 (approved):

src/Api/Client.php
```diff
@@ -10,75 +10,56 @@
...
     /**
      * Execute the given API action class, pass the input and return its response.
      *
-     * @param string|RequestHandlerInterface $controller
+     * @param string $routeName
+     * @param Request|null $parent
      * @param User|null $actor
      * @param array $queryParams
      * @param array $body
      * @return ResponseInterface
```
clarkwinkelmann, line 44: The `use` alias hasn't been updated here.

Is it really necessary to alias request and response to shorter names though?

src/Api/Middleware/ResolveRouteFromName.php
```diff
@@ -0,0 +1,38 @@
...
+    public function __construct(RouteCollection $routeCollection)
+    {
+        $this->routeCollection = $routeCollection;
+    }
+
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
+    {
+        $route = $this->routeCollection->getRoutes()[$request->getAttribute('routeName')];
+
+        $request = $request->withMethod($route['method']);
```
clarkwinkelmann, line 29: How will this work with routes that might be available via multiple methods? Can we predict which one will be chosen? Should there be a way to specify?

I wouldn't mind having `$client->post()`, `$client->get()` and such and pass the method as an additional request attribute.

clarkwinkelmann (commented):
I haven't tested the changes, but here are a few comments based on the code.

I'm not particularly a fan of `private` statements as I have said before. I'm sure this change will break multiple of my extensions where I will have to copy the whole file instead of just re-implementing `getApiDocument` (when it's private, not only it cannot be called, but re-implementing it doesn't work without also re-implementing the method that calls it)

Couldn't a method chain maybe make for a cleaner client API? Like `$client->withParentRequest($request)->withParams($params)->get('discussions.show')` to avoid all the `null` parameters?

src/Api/Middleware/ResolveRouteFromName.php
askvortsov1: Converted to use paths, as per our chat discussion

askvortsov1:
> I'm not particularly a fan of `private` statements as I have said before. I'm sure this change will break multiple of my extensions where I will have to copy the whole file instead of just re-implementing `getApiDocument` (when it's private, not only it cannot be called, but re-implementing it doesn't work without also re-implementing the method that calls it)

Reverted, not sure why I changed that. Although tbh, the entire Content system needs an extensibility review.

> Couldn't a method chain maybe make for a cleaner client API? Like $client->withParentRequest($request)->withParams($params)->get('discussions.show') to avoid all the null parameters?

Agreed, changed. Much nicer, and clearer to read.

clarkwinkelmann (commented):
Still not tested locally, but it looks really good like this now!

Just one suggestion below.

src/Api/Client.php
```diff
@@ -9,76 +9,129 @@
...
-     * @param string|RequestHandlerInterface $controller
-     * @param User|null $actor
-     * @param array $queryParams
-     * @param array $body
+     * @param string $routeName
      * @return ResponseInterface
-     * @throws Exception
      */
-    public function send($controller, User $actor = null, array $queryParams = [], array $body = []): ResponseInterface
+    protected function send(string $method, string $path): ResponseInterface
```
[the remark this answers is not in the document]
askvortsov1: How about public, but with an `@internal` warning?

src/Api/Client.php
```diff
@@ -115,15 +120,19 @@ public function delete(string $path): ResponseInterface
...
      * @param string $routeName
      * @return ResponseInterface
+     *
+     * @internal
      */
-    protected function send(string $method, string $path): ResponseInterface
+    public function send(string $method, string $path): ResponseInterface
     {
         $request = ServerRequestFactory::fromGlobals(null, $this->queryParams, $this->body)
+            ->withParsedBody($this->body)
```
clarkwinkelmann, line 129: Is `$this->body` really the same variable for both the original body and the parsed body? Wouldn't those two values be different?

askvortsov1: https://github.com/flarum/core/blob/c1992b45debaacae6024d36d1b32a32fb129b312/src/Http/Middleware/ParseJsonBody.php#L18-L18 converts body to json, but the provided body must already be an array. With this change, any code expecting an array as parsedBody will see exactly that.

Perhaps we should json encode the original body?

clarkwinkelmann: I wasn't sure when I wrote the comment but now I see that the ParseJsonBody middleware is part of the excluded middlewares.

Can't we copy both the parsed and non-parsed body from the parent request?

JSON encoding sounds like a good alternative, but it requires checking if the body has actually been parsed, because if it's not JSON in the first place this will mess things up.

I suppose it's rare for an extension to read the original raw body WHILE it's a valid JSON, but I feel like this should be consistent to avoid any issue.

askvortsov1: > Can't we copy both the parsed and non-parsed body from the parent request?

Almost always, the body being sent in an API client request will not be the same as the body from the original request. As opposed to the actor and session, which should be kept.

askvortsov1: Actually, this turned out to be unnecessary, as ::fromGlobals sets the provided body as parsed body, not as raw body.

src/Api/Client.php
```diff
@@ -9,76 +9,137 @@
...
+
+    public function withQueryParams(array $queryParams): Client
+    {
+        $new = clone $this;
+        $new->queryParams = $queryParams;
+
+        return $new;
+    }
+
+    public function withBody(array $body): Client
```
clarkwinkelmann, line 85: Shouldn't we allow for text body as well with this method?

askvortsov1: Why?

clarkwinkelmann: To cover more use cases. Though admittedly I can't think of any reason for string right now.

There's also the concern of 500 errors due to custom data... If a client sends `null` (as JSON) in the body, it should return a 400 error or just do nothing depending on the context, but here it would result in a 500 error because we try to pass down the valid JSON down to the API client which will reject it due to type-hint.

clarkwinkelmann (approved):
Looks good.

Just one more thing below.

Issue #2784, opened 2021-04-12
Session bound to User object creates Queueing issues


luceos:
## Bug Report

**Current Behavior**

We are setting the session of the user on the User object. When serializing happens - for instance the Notification dispatches - the User object is serialized. Now whenever you are using a Session driver that is not serializable (like Redis) queue handling will fail.

**Possible Solution**

- Bind the solution to ioc (isn't that the case) and remove from the User object.
- Override the `__serialize` method on the User class to drop the session property.

askvortsov1:
Do we need the session to be serialized with the user? Any use cases for accessing it from within the queue?

> Bind the solution to ioc (isn't that the case)

No per-request state in IoC!!! :angry:

PR #2790, opened 2021-04-16
Drop session from user class

askvortsov1:
This was originally introduced inhttps://github.com/flarum/core/commit/3612ca7aca90b5d45310da1602256e247451cec3, but has not seen usage, since usually when the session needs to be modified, the request is available.

It causes issues with certain queue drivers, as it can't be serialized.

It's also not entirely accurate, as a user can have multiple sessions at once. Therefore, a given session is a property of the request, not of the user.

### DIFF
--- a/src/Http/Middleware/AuthenticateWithSession.php
+++ b/src/Http/Middleware/AuthenticateWithSession.php
@@ -26,8 +26,6 @@ class AuthenticateWithSession implements Middleware
 
         $actor = $this->getActor($session, $request);
 
-        $actor->setSession($session);
-
         $request = RequestUtil::withActor($request, $actor);
 
         return $handler->handle($request);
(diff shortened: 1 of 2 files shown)


luceos (approved):
The reason this causes issues in the Queue is that when a Job has payload that consists User(s), the Queue will try to serialize that. Serializing the User object will require serializing the `session` too; this causes a `Serialization of Closure is not allowed` error, see image.

One can circumvent that in many ways, the most obvious one is adding a `__sleep` and `__wakeup` implementation in the User class (or the session handler). But as we aren't really using the session on the User model anywhere in core, bundled or most community extensions it is best to simply detach this from the user.

image

tankerkiller125 (approved):

askvortsov1:
Fixes https://github.com/flarum/core/issues/2784

PR #2796, opened 2021-04-19
Remove unnecessary laravel config

askvortsov1:
These are probably left over from when we relied more on Laravel's mail config; we configure this entirely ourselves in https://github.com/flarum/core/blob/c9d4a492d664f745c7b288d083c26f4d283f1d83/src/Mail/MailServiceProvider.php#L19-L19


tankerkiller125 (approved):

luceos (approved):
Great find :)

PR #2797, opened 2021-04-20
Fix uses of loading spinner

### LINKED ISSUE #2786: Fix new CSS loading spinner issues
I broke master! Well... one small part of it. Am I a real dev now?

- [ ] Composer
- [ ] Loading a new discussion
- [ ] Buttons

davwheat:
**Fixes #2786**

I've fixed what I broke!

I've got a drafted post on discuss about the new spinner ready-to-go, if you think I should post that and then we can link to it in release notes to help extension authors.

I don't think anything is left broken now, but please have a look :P

### DIFF
--- a/js/src/common/components/LoadingIndicator.tsx
+++ b/js/src/common/components/LoadingIndicator.tsx
@@ -31,7 +31,7 @@ export interface LoadingIndicatorAttrs extends ComponentAttrs {
  * spinner's **container**. Applying the `block` attribute handles this for you.
  *
  * To apply a custom size to the loading indicator, set the `--size` and
- * `--thickness` custom properties on the loading indicator itself.
+ * `--thickness` custom properties on the loading indicator container.
  *
  * If you really want to change how this looks as part of your custom theme,
  * you can override the `border-radius` and `border` then set either a
--- a/js/src/common/components/LoadingIndicator.tsx
+++ b/js/src/common/components/LoadingIndicator.tsx
@@ -51,7 +51,7 @@ export interface LoadingIndicatorAttrs extends ComponentAttrs {
  */
 export default class LoadingIndicator extends Component<LoadingIndicatorAttrs> {
   view() {
-    const { inline, block, size, containerClassName, className, ...attrs } = this.attrs;
+    const { inline = false, block = false, size = 'medium', containerClassName, className, ...attrs } = this.attrs;
 
     if (inline && block) throw new Error('LoadingIndicator cannot have both `inline` and `block` attributes.');
 
--- a/js/src/common/components/LoadingIndicator.tsx
+++ b/js/src/common/components/LoadingIndicator.tsx
@@ -18,8 +18,10 @@ export interface LoadingIndicatorAttrs extends ComponentAttrs {
    * Optional attributes to apply to the loading indicator's container.
    */
   containerAttrs?: Partial<ComponentAttrs>;
-  inline?: boolean;
-  block?: boolean;
+  /**
+   * Display type of the spinner.
+   */
+  display?: 'block' | 'inline' | undefined;
 }
 
 /**
@@ -42,24 +44,21 @@ export interface LoadingIndicatorAttrs extends ComponentAttrs {
  *
  * - `containerClassName` Class name(s) to apply to the indicator's parent
  * - `className` Class name(s) to apply to the indicator itself
- * - `size` Size of the loading indicator (small, medium or large)
+ * - `size` Size of the loading indicator (`small`, `medium` or `large`)
  * - `containerAttrs` Optional attrs to be applied to the container DOM element
- * - `block` Displays the spinner as a block element with a height of 100px
- * - `inline` Displays the spinner as an inline element, centrally aligned vertically.
+ * - `display` Determines how the spinner should be displayed (`inline` or `block`)
  *
  * All other attrs will be assigned as attributes on the DOM element.
  */
 export default class LoadingIndicator extends Component<LoadingIndicatorAttrs> {
   view() {
-    const { inline = false, block = false, size = 'medium', containerClassName, className, ...attrs } = this.attrs;
-
-    if (inline && block) throw new Error('LoadingIndicator cannot have both `inline` and `block` attributes.');
+    const { display = undefined, size = 'medium', containerClassName, className, ...attrs } = this.attrs;
 
     const completeClassName = classList('LoadingIndicator', className);
     const completeContainerClassName = classList(
       'LoadingIndicator-container',
-      block && 'LoadingIndicator--block',
-      inline && 'LoadingIndicator--inline',
+      display === 'block' && 'LoadingIndicator--block',
+      display === 'inline' && 'LoadingIndicator--inline',
       containerClassName
     );
 
--- a/js/src/common/components/LoadingIndicator.tsx
+++ b/js/src/common/components/LoadingIndicator.tsx
@@ -18,6 +18,8 @@ export interface LoadingIndicatorAttrs extends ComponentAttrs {
    * Optional attributes to apply to the loading indicator's container.
    */
   containerAttrs?: Partial<ComponentAttrs>;
+  inline?: boolean;
+  block?: boolean;
 }
 
 /**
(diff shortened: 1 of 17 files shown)


less/common/Search.less
```diff
@@ -61,11 +62,21 @@
...
     float: left;
     margin-left: -36px;
-    width: 36px !important;
     outline: none;
+    width: 36px !important;
+
+    &.LoadingIndicator {
+      width: var(--size) !important;
+      padding: 0;
+    }
```
SychO9, line 79: I don't think this is used since LoadingIndicator is a child here.

js/src/common/components/LoadingIndicator.tsx
```diff
@@ -40,21 +42,30 @@ export interface LoadingIndicatorAttrs extends ComponentAttrs {
  *
  * - `containerClassName` Class name(s) to apply to the indicator's parent
  * - `className` Class name(s) to apply to the indicator itself
- * - `size` Size of the loading indicator
+ * - `size` Size of the loading indicator (small, medium or large)
```
SychO9, line 45: You changed `tiny` to `small` but the CSS still targets `tiny`, any reason for the change ?

js/src/common/components/LoadingIndicator.tsx
```diff
@@ -40,21 +42,30 @@ export interface LoadingIndicatorAttrs extends ComponentAttrs {
...
+      'LoadingIndicator-container',
+      block && 'LoadingIndicator--block',
+      inline && 'LoadingIndicator--inline',
+      containerClassName
+    );
 
     return (
-      <div {...attrs.containerAttrs} data-size={size} className={attrs.containerClassName}>
-        <div {...attrs}></div>
+      <div {...attrs.containerAttrs} data-size={size} className={completeContainerClassName}>
```
SychO9, line 67: a bit out of scope of this PR, but any reason for using `data-*` instead of a modifier class which is more in line with convention ?

SychO9 (changes requested):

js/src/common/components/LoadingIndicator.tsx
davwheat: Data attributes are intended as a private storage area for DOM elements, and I prefer using them wherever possible for things like this.

They're slightly less performant than classes (9000 operations/sec vs 13000 ops/s), but I find `data-size="large"` more verbose and clear than `class="LoadingIndicator LoadingIndicator--large"`. The change in performance is basically negligible -- it's still 9000 operations per second!

less/common/Search.less
davwheat: It's not. We make the loading indicator take the Button class and a few more classes for... questionable reasons. This replaces the normal "clear field" button while sources are loading.

davwheat (commented):
~~I've noticed there's still a few places I need to fix which I've missed in here too.~~

I just accidentally switched branch and forgot!

js/src/common/components/LoadingIndicator.tsx
SychO9: > Data attributes are intended as a private storage area for DOM elements, and I prefer using them wherever possible for things like this.

Hmm, I still think a modifier class is more appropriate and correct, as per https://johnresig.com/blog/html-5-data-attributes/ data attributes aren't meant to affect layout or presentation, we're only using `size` here to change the way the loading indicator is presented, so I think we're misusing it.

davwheat: That blog post talks about the fact that data attributes aren't used by the browser to adjust layout and presentation, like other attributes can be (e.g. deprecated `align`, `hidden`, etc). Maybe we should get some more opinions on this?

SychO9: > Places such as CSS Tricks seem to show them being used for things like this

I'm not completely opposed to using it, there are valid use cases, I just don't see the benefit of using it over a class when the sole purpose is styling, especially as the rest of the code base already uses a convention in which modifier classes are used.

Additionally there is a reason for the current convention we follow, one that I find very valuable with a customizable software. Specificity, a `LoadingIndicator--small` is less specific than `LoadingIndicator[data-size="small"]`, and having tried multiple times customizing the software, specificity was always the most tedious thing to deal with, as you have to overspecify even more or even use `!important`.

It's best if we figure this out now 🤔

js/src/common/components/LoadingIndicator.tsx
davwheat: To me, small medium and large sounds more natural than tiny medium and large.

js/src/common/components/LoadingIndicator.tsx
davwheat: You're right -- it has higher specificity, but that's not inherently wrong.

The main reason I like data attributes is because it makes it clear what data can be used to modify the element and what that actually means. Most of the time, we won't need to customise the size, so developers may only see the `LoadingIndicator` class and never the size classes, meaning that they may not be aware that they exist.

If we have `LoadingIndicator--small` (and other sizes), then `LoadingIndicator--inline` (and block), it starts becoming unclear what the modifiers are actually representing. It's not a massive issue in this component, but could be in others.

askvortsov1: I think I prefer using classes. I strongly doubt that this is the bottleneck for performance, but better customizability is very important. Also, I don't see why block/inline should be classes, but not small/medium/large.

Alternatively, maybe we could set it as a css variable?

js/src/common/components/LoadingIndicator.tsx
```diff
@@ -40,21 +42,30 @@ export interface LoadingIndicatorAttrs extends ComponentAttrs {
...
  * - `containerAttrs` Optional attrs to be applied to the container DOM element
+ * - `block` Displays the spinner as a block element with a height of 100px
+ * - `inline` Displays the spinner as an inline element, centrally aligned vertically.
  *
  * All other attrs will be assigned as attributes on the DOM element.
  */
 export default class LoadingIndicator extends Component<LoadingIndicatorAttrs> {
   view() {
-    const { size, ...attrs } = this.attrs;
+    const { inline = false, block = false, size = 'medium', containerClassName, className, ...attrs } = this.attrs;
```
askvortsov1, line 54: Why not default to either block or inline? Also, shouldn't these be mutually exclusive?

js/src/common/components/LoadingIndicator.tsx
tankerkiller125: In this particular use case I believe that using `LoadingIndicator--{size}` would be correct to meet our existing standards and uses.

davwheat: We can't really set it as a CSS variable, but we do have support for manually specifying size as variables using `--size: 36px` (for example).

I would have personally chosen to use `data-display` with `block` and `inline`, but I don't see any other display options that we might need.

I see the size as more of a variable than anything, which splitting into an attribute makes it more similar to normal HTML elements, such as `<input type="number">`, for example.

js/src/common/components/LoadingIndicator.tsx
davwheat: I didn't want to default to one or the other as there may be some cases where extensions wish to do their own custom styling.

Maybe this would be better as a `display` attribute on the component, such as `display="block"` and `display="inline"`?

js/src/common/components/LoadingIndicator.tsx
davwheat: Matt's comment disappeared into thin air when I rebased onto master, so I copy-pasted it from the email notification:

> In this particular use case I believe that using `LoadingIndicator--{size}` would be correct to meet our existing standards and uses.

js/src/common/components/LoadingIndicator.tsx
```diff
@@ -40,27 +45,33 @@ export interface LoadingIndicatorAttrs extends ComponentAttrs {
...
+ * - `display` Determines how the spinner should be displayed (`inline` or `block`)
+ * - `size` Size of the loading indicator (`small`, `medium` or `large`)
  * - `containerAttrs` Optional attrs to be applied to the container DOM element
  *
  * All other attrs will be assigned as attributes on the DOM element.
  */
 export default class LoadingIndicator extends Component<LoadingIndicatorAttrs> {
   view() {
-    const { size, ...attrs } = this.attrs;
+    const { display = undefined, size = 'medium', containerClassName, className, ...attrs } = this.attrs;
```
askvortsov1, line 56: Can we default display to block? That's more consistent with how it was before, and will require less changes on the part of extension devs.

askvortsov1 (changes requested):

askvortsov1 (approved):

SychO9 (approved):

PR #2805, opened 2021-04-22
Revision compiler revised

luceos:

Moving from `<asset>-<revision>.<js|css>` to `<asset>.<js|css>?v=<revision>`. As a consequence we no longer need to delete the old file, browsers will automatically pick up that a new file was generated. 

Additional changes

- clean up the revision classes, adding more type hinting (compatible with php 7.3), removing unneeded phpdoc
- changed the cache differentiator for Less; previously it would use `time()` causing a recompile every time assets are compiled; but as all sources (string, file) offer their own differentiator and the less compiler only adds import dirs, I've used that as another differentiator. Let me clarify the impact:

For production sites there won't be any change, except that css is compiled less often.
For development environments where import dirs see changes this might be problematic, however this PR also takes this into consideration by allowing the `commit()` method of the recompiler to get an argument `$force (false)`  to force recompilation when in debug mode.

Whether we want to modify the differentator for Less like this.

**Screenshot**

image

**Confirmed**

**Required changes:**

### DIFF
--- a/src/Frontend/Content/Assets.php
+++ b/src/Frontend/Content/Assets.php
@@ -48,6 +48,7 @@ class Assets
             'css' => [$this->assets->makeCss(), $this->assets->makeLocaleCss($locale)]
         ];
 
+        // In debug we force a recompilation of assets.
         if ($this->config->inDebugMode()) {
             $this->commit(Arr::flatten($compilers));
         }
@@ -58,8 +59,9 @@ class Assets
 
     private function commit(array $compilers)
     {
+        /** @var CompilerInterface $compiler */
         foreach ($compilers as $compiler) {
-            $compiler->commit();
+            $compiler->commit(true);
         }
     }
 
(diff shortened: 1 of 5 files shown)


davwheat:
I love this! :)

src/Frontend/Content/Assets.php
```diff
@@ -58,8 +59,9 @@ public function __invoke(Document $document, Request $request)
 
     private function commit(array $compilers)
```
askvortsov1, line 60: Can we add a comment saying "This method only runs when in debug mode, so we force a recompilation"?

askvortsov1 (changes requested):
Changes look good to me, including the differentiator changes. Local testing confirms that less is recompiled when changed as expected. Just a slight nitpick, `commit` assumes debug mode (because it's only called during debug mode). Can we explicitly state this expectation either in a comment, or by renaming to something? `forceCommit` maybe?

luceos:
If the method is named commit, it feels superfluous to name the variable forceCommit. That's a waste of screen width and only makes it unclear as I'd expect something else to be forceable too 🙈

askvortsov1:
> If the method is named commit

I meant renaming the method in the `Assets` class to `forceCommit`

askvortsov1 (approved):

SychO9 (approved):

PR #2806, opened 2021-04-22
Harden formatter

luceos:
**See flarum/issue-archive#121**

This is not excellent code, but it takes into consideration performance and handles the incomplete class issue that some people experienced. You can reproduce this error by deleting `storage/formatter/*` and not cache clearing. My changes test whether the file exists and stores it on disk for autoload inclusion.

- prevents __PHP_Incomplete_Class
- stores a revision

I dislike extending the RevisionCompiler as many methods go unused and it could open errors if used incorrectly. Any suggestions on how to improve this welcome.

**Confirmed**

**Required changes:**

### DIFF
--- a/src/Formatter/Formatter.php
+++ b/src/Formatter/Formatter.php
@@ -9,12 +9,17 @@
 
 namespace Flarum\Formatter;
 
+use __PHP_Incomplete_Class;
+use ArrayObject;
+use Flarum\Frontend\Compiler\RevisionCompiler;
 use Illuminate\Contracts\Cache\Repository;
+use Illuminate\Contracts\Filesystem\Filesystem;
+use Illuminate\Support\Arr;
 use Psr\Http\Message\ServerRequestInterface;
 use s9e\TextFormatter\Configurator;
 use s9e\TextFormatter\Unparser;
 
-class Formatter
+class Formatter extends RevisionCompiler
 {
     protected $configurationCallbacks = [];
 
@@ -35,13 +40,17 @@ class Formatter
     protected $cacheDir;
 
     /**
-     * @param Repository $cache
-     * @param string $cacheDir
+     * @var array|null
      */
-    public function __construct(Repository $cache, $cacheDir)
+    protected static $formatter;
+
+    public function __construct(Repository $cache, string $cacheDir, Filesystem $assetsDir)
     {
         $this->cache = $cache;
         $this->cacheDir = $cacheDir;
+        $this->assetsDir = $assetsDir;
+
+        $this->filename = 'formatter';
     }
 
     public function addConfigurationCallback($callback)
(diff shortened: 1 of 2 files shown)


src/Formatter/Formatter.php
```diff
@@ -227,4 +250,69 @@ public function getJs()
...
+        } else {
+            $this->finalize();
+        }
+
+        // Reload because finalizing might have generated a new one.
+        $renderer = Arr::get(static::$formatter, 'renderer');
+
+        if ($renderer && static::$formatter['renderer'] instanceof __PHP_Incomplete_Class) {
+            // Autoload the file from disk using a simple include, while suppressing errors.
+            @include $this->cacheDir."/Renderer_$revision.php";
```
askvortsov1, line 279: If the renderer is an incomplete class, why would loading it in again help? I'm not sure what the goal of this whole `if` statement is.

src/Formatter/Formatter.php
```diff
@@ -227,4 +250,69 @@ public function getJs()
...
+            // Reload the formatter again from cache to resolve the __PHP_Incomplete_Class
+            static::$formatter = $this->cache->get('flarum.formatter');
+        }
+    }
+
+    protected function finalize()
+    {
+        $formatter = $this->getConfigurator()->finalize();
+
+        preg_match('~^Renderer\_(?<revision>[^\.]+)$~', get_class($formatter['renderer']), $m);
```
askvortsov1, line 290: Can we make this regex a private method? It's repeated several times now. IMO it should just return the revision so we don't have to carry around the `$m` object

src/Formatter/Formatter.php
```diff
@@ -227,4 +250,69 @@ public function getJs()
     {
         return $this->getComponent('js');
     }
+
+    protected function ensureRendererExists()
+    {
+        if (! static::$formatter) {
+            return;
```
askvortsov1, line 257: When would this happen?

askvortsov1 (changes requested):
> I dislike extending the RevisionCompiler as many methods go unused and it could open errors if used incorrectly. Any suggestions on how to improve this welcome.

I don't think we should do this. `RevisionCompiler` and the whole `rev-manifest` system is for frontend assets, using it for these internals is not good IMO. Can we store the revision ID in the laravel cache instead? All we're using `RevisionCompiler` for is getting/setting the revision id from/to a persistent store.

On a more general note, the logic here is a bit confusing.

To clarify the process:

For the formatter, we cache 2 things:

- A serialized version of the entire formatter (cached by us via the laravel cache)
- A renderer file (cached by TextFormatter)

Sometimes, the latter might get deleted/renamed, and so the former will look for it and not find it. This logic recreates it if it is not found.

This feels like it could occur on a non-cached TextFormatter environment too: maybe it might be a decent candidate for inclusion in TextFormatter itself (pinging @JoshyPHP)? I'm assuming TextFormatter internals are somehow aware of the renderer revision number, so that would reduce the need to cache it separately.

JoshyPHP:
> I'm assuming TextFormatter internals are somehow aware of the renderer revision number, so that would reduce the need to cache it separately.

No, when a renderer is created (by `finalize()`) the PHP source is generated and saved to the disk. If the class doesn't already exist, the file is loaded before the renderer instance is created. https://github.com/s9e/TextFormatter/blob/fa6ce8f1ab17b51da4d96af05ddfa73782510cab/src/Configurator/RendererGenerators/PHP.php#L113-L138

Apart from that, there's no mechanism for autoloading the required files. I don't know how you currently handle it in Flarum but what I'd do is I'd cache the path to the renderer's source. If it's missing then I'd regenerate everything.

PR #2814, opened 2021-04-29
[A11Y] Add focus ring mixin to restore ring to elements which no longer have it

davwheat:

These mixins allow us to restore default browser focus rings on elements which no longer have them.

I chose to make the styles mixin "private" as it shouldn't be used manually elsewhere in core -- we should instead use the other available mixins in the file. Is this the right choice?

**Screenshot**
image

This screenshot uses:

```less
.add-keyboard-focus-ring();
.add-keyboard-focus-ring-offset(4px);
```

**Confirmed**

### DIFF
--- a/less/common/mixins.less
+++ b/less/common/mixins.less
@@ -1,3 +1,4 @@
+@import "mixins/accessibility.less";
 @import "mixins/border-radius.less";
 @import "mixins/clearfix.less";
 @import "mixins/light-contents.less";
--- a/less/common/mixins/accessibility.less
+++ b/less/common/mixins/accessibility.less
@@ -1,0 +1,96 @@
+// This mixin should **only** be used in this file.
+.__priv-focus-ring-styles() {
+  // This uses the browser's default outline styles, rather than
+  // using custom ones, which could introduce more issues
+  //
+  // Source: https://css-tricks.com/copy-the-browsers-native-focus-styles
+  outline: 5px auto Highlight;
+  outline: 5px auto -webkit-focus-ring-color;
+}
+
+/**
+ * Adds a focus ring to an element.
+ *
+ * This is only shown when focus is provided via keyboard, using the
+ * `:focus-visible` selector, and `:-moz-focusring` for older Firefox.
+ */
+.add-keyboard-focus-ring() {
+  // We need to declare these separately, otherwise
+  // browsers will ignore `:focus-visible` as they
+  // don't understand `:-moz-focusring`
+
+  // These are the keyboard-only versions of :focus
+  &:-moz-focusring {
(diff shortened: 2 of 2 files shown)


askvortsov1:
It doesn't seem like these are actually being used right now, unless I'm misunderstanding something?

davwheat:
No, but I'm writing some improvements for the Notifications styles which will require this.

(Grid and flex!)

askvortsov1 (approved):

SychO9:
wouldn't it be easier to remove any `outline: none` on element that don't have it and let the browser do it ? or is it better for us to specify the outline ourselves ?

davwheat:
Removing `outline: none` is also good, but there are weird nuances and situations that result in outlines not existing anyway.

This is intended to be a temporary fix anyway -- I hope to get as many of our mixins removed as possible so we don't have border radius repeated 5 times when only one of them does anything :P

SychO9:
> but there are weird nuances and situations that result in outlines not existing anyway.

that shouldn't be the case by default, what you're seeing locally (in the image) is most likely because you also technically have the button `:active` or something that sets off the outline as well (we probably have `outline: none` all over the place 🙈)

However you make a good point about customization! how does that work here, by overriding mixins ? I don't think that's possible in less 🤔, we might have to move the outline value to `variables.less`

davwheat:
In Less, mixins get merged together, with the later mixins taking precedence.

Though, saying that, moving it to a variable might be a good idea.

Should the variable be the entire value? I'm leaning towards that instead of splitting it into 3 separate ones.

davwheat:
Actually, we can't move it to a variable as we have two values to make it work. Leaving it as a mixin might be our only choice.

SychO9 (approved):
> In Less, mixins get merged together, with the later mixins taking precedence.

Aha, interesting that's the case indeed!

Let's leave it as is then.

davwheat:
Just added a quick comment about how to override the outline styles, and switched to using a `#private` namespace instead of cluttering up the root namespace even more.

I'll give both of you a few mins to note any objections to that. I think core should have probably used a namespace for its own mixins from the start, and we should probably recommend extensions do the same.

SychO9:
neat, no objections!

> I think core should have probably used a namespace for its own mixins from the start, and we should probably recommend extensions do the same.

I like the idea, I think it'd be worth implementing that when we get to refactoring more of the frontend code. I think it'd be good to have an issue about it.

davwheat:
I've created a docs issue -- hopefully I'll get the chance to write a small bit about it before stable releases, and we can note that we strongly recommend switching to that: https://github.com/flarum/docs/issues/322

PR #2824, opened 2021-05-02
Move user badges to bottom of avatar

zerosonesfun:
Move user badges to bottom of avatars.

Changes .PostUser-badges `top` to 15px for phone and `margin-top` to 50px for desktop which moves the badges from the top of avatars to the bottom. Also, compared to how much they overlapped the avatar before, this puts the badges right on the edge of the avatar.

The reason I am proposing this change is because currently the badges cover the user's face. Other proposed changes have been made in the past such as moving the badges to a completely different spot, off of the avatar completely. However, it is my opinion that one of the things that makes Flarum look like Flarum is the badges on the avatar. This keeps the badges tied to the avatar while uncovering faces.

Does it look nice on both mobile and desktop?

**Screenshot**

**Confirmed**

### DIFF
--- a/less/forum/Post.less
+++ b/less/forum/Post.less
@@ -354,7 +354,7 @@
   }
   .PostUser-badges {
     position: absolute;
-    top: -12px;
+    top: 15px;
     left: 6px;
     width: 32px;
 
@@ -391,7 +391,7 @@
     float: left;
     position: relative;
     margin-left: -@avatar-column-width + 5px;
-    margin-top: -3px;
+    margin-top: 50px;
     width: 64px;
   }
   .EventPost-icon {


luceos (approved):
Love it.

askvortsov1:
I have mixed feelings about this one, I'd like to see it in action first before changing the default.

I've enabled this change on Discuss for now. Let's see how it feels.

justoverclockl:
love this definitely

tankerkiller125:
After seeing it on discuss with single badges I'm not sold on it. It just feels.... Off. I think part of that is that with a single group the icon is not aligned to anything else.



I feel like the center of the group circle should align with the edged of the avatar circle (almost as if it where in a corner of a box)

davwheat:
I'm with Matt on this one. 👍

The badges/username are already misaligned on mobile by error, so we should probably consider fixing that too.

justoverclockl:
I don't know if this bug depend from this changement, but on mobile I see this on my post that is pending approval

image

jordanjay29:
> After seeing it on discuss with single badges I'm not sold on it. It just feels.... Off. I think part of that is that with a single group the icon is not aligned to anything else.

I'd agree with @tankerkiller125 here, maybe for different reasons. Bottom-right feels wrong, especially with multiples making it flow to the left. Why does it not start at bottom-left and flow right? Or center and expand?

zerosonesfun:
I’ve seen a couple comments mentioning something about alignment or size. As a reminder, the only change has to do with the “top” CSS and “margin-top.” The left/right alignment and the size of the badges are the same. 

This option at least keeps the badges on the avatar. The inline option I’ve seen many times is nice. But, that’s even less “like Flarum.” I mention this because I saw comments about this not feeling like Flarum.

The main goal though is to stop hiding people’s faces. So, if folks don’t like this, my next vote would be to make that inline version the UI in core.  

Or, the badges could easily be moved so they start on the left. I didn’t do that though because they don’t start on the left when at the top. This change was purposely done so that hardly nothing is changed, except two CSS rules.

zerosonesfun:
Here’s my final opinion. 

- They can’t start on the left. Because then you have badges creeping into the far left margin. I bet badges were originally positioned where they are today (to the right) due to things like that. 

- I love the “inline” option people have shared but that’s an even bigger change IMO. 

- Whether they move to the bottom or stay where they are, on mobile, maybe we can all agree to make the badges just a tiny bit smaller...

Both of these screenshots reduce the badge size on mobile. 

Bottom:


Top:

zerosonesfun:
> After seeing it on discuss with single badges I'm not sold on it. It just feels.... Off. I think part of that is that with a single group the icon is not aligned to anything else.
> 
> I'd agree with @tankerkiller125 here, maybe for different reasons. Bottom-right feels wrong, especially with multiples making it flow to the left. Why does it not start at bottom-left and flow right? Or center and expand?

They flow left currently. Just looks different because it’s on the bottom. 🙂

zerosonesfun:
😁 @luceos I have a feeling based on this early feedback that it’ll be best to leave avatar badges as-is. **Maybe** reduce the size of the badge just slightly on mobile so a little more of the avatar shows. 

Then, anyone who wants to move the badges can do so with custom CSS.

Good call @askvortsov1 posting at discuss for feedback!

davwheat:
Post needs a CSS overhaul as well as part of flarum/issue-archive#355. This might open the doors to other methods, such as using grid for the badge stack.

zerosonesfun:
Another idea that I've tested locally: https://discuss.flarum.org/d/27058-avatar-badges-ui-suggestions/9

I'll hold off doing any other pull requests though around badge placement. If enough people like this new idea I'd be willing to take the time to make the necessary updates in core.

Why this new idea?

- First and foremost, the goal is to uncover avatars at least a little
- Keeps badges at top of avatar - more like original Flarum
- Making them vertical saves space, and unhides avatars more
- Someone else mentioned the thin white border. That's optional. But, I really like it.
- On mobile only, in the post stream, group badges move to being inline just before the "days ago" - due to the smaller avatar this looks better on mobile on this particular page

zerosonesfun:
How does everyone feel if the badges (minus the discussion topics list) moved to that inline layout either before, or after the date? I might test it out regardless at some point.

What I would do is change the templates/HTML slightly first. Move the badges div and list after the username or after the date. Then once the HTML is moved, it’ll be much easier to style. With less CSS.

KyrneDev:
I'm personally not a fan of them at the bottom, it feels unnatural to me. I like the fanned out version shown here: 
https://discuss.flarum.org/d/27058-avatar-badges-ui-suggestions/19

I think it comes down to having extra padding above currently so it fits better with post content. With the badge at the bottom with no padding it's weirdly close to the closet content.

zerosonesfun:
> I'm personally not a fan of them at the bottom, it feels unnatural to me. I like the fanned out version shown here:
> https://discuss.flarum.org/d/27058-avatar-badges-ui-suggestions/19
> 
> I think it comes down to having extra padding above currently so it fits better with post content. With the badge at the bottom with no padding it's weirdly close to the closet content.

All kinds of small adjustments could be made and maybe on the bottom would look better. For example, here is on the bottom but moved up and smaller icons (smaller icons optional). And padding added to separate the content better.

luceos:
I second @KyrneDev; I dislike them at the bottom and start to appreciate the version on the right (or possibly left) side; as long as the badges follow the circle of the avatar.

zerosonesfun:
> I second @KyrneDev; I dislike them at the bottom and start to appreciate the version on the right (or possibly left) side; as long as the badges follow the circle of the avatar.

It's a small detail, but, by sticking with the right side, if a user only have 1 badge, they won't notice a different. The first badge will start where it starts today. The other badges will curve around.

Closing this and will submit a new pull request once this curved/circular idea is perfect.

PR #2827, opened 2021-05-02
Remove `lodash` from core

davwheat:
**Progresses flarum/issue-archive#179**

Removed `lodash-es` from our dependencies, and have replaced them with a custom util, a polyfill for a native browser method, and the `throttle-debounce` library.

These replacements don't function identically to the original library's, but should suit our use cases.

**Confirmed**

### DIFF
--- a/js/src/common/utils/arrayFlatPolyfill.ts
+++ b/js/src/common/utils/arrayFlatPolyfill.ts
@@ -3,7 +3,7 @@
 // Needed to provide support for Safari on iOS < 12
 
 if (!Array.prototype['flat']) {
-  Array.prototype['flat'] = function flat(depth?: number) {
+  Array.prototype['flat'] = function flat(this: any, depth?: number) {
     const realDepth = isNaN(depth as any) ? 1 : depth;
 
     return realDepth
@@ -11,15 +11,17 @@ if (!Array.prototype['flat']) {
           this,
           function (acc, cur) {
             if (Array.isArray(cur)) {
-              acc.push.apply(acc, flat.call(cur, realDepth - 1));
+              (acc as Array<any>).push.apply(acc, flat.call(cur, realDepth - 1));
             } else {
-              acc.push(cur);
+              (acc as Array<any>).push(cur);
             }
 
             return acc;
           },
-          []
+          [] as Array<any>
         )
-      : Array.prototype.slice.call(this);
+      : // If no depth is provided, or depth is 0, just return a copy of
+        // the array. Spread is supported in all major browsers (iOS 8+)
+        [...this];
   };
 }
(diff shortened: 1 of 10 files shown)


davwheat:
-3.4 kB from bundle (-1.65 kB gzipped)

datitisev:
Could you not include imports reordering in this PR? That, if at all, should be done outside it.

I'm not in love with copying lodash methods simply to save a few kilobytes, as that means we then are responsible for maintaining those functions, but I guess the ones chosen shouldn't be an issue...

davwheat:
I've switched out the flatten util for the native browser `Array.flat(depth)` call and a polyfill to support older browsers (read: Safari).

This does make the bundle slightly bigger, but I think it's worth it for the likely better-optimised native browser code and the simplicity of just calling `.flat(Infinity)`.

askvortsov1 (approved):

davwheat:
Just going to see what we gain/lose by using `[...this]` instead of `Array.prototype.slice(this)`. Unsure what Webpack will compile it to.

davwheat:
So it's 366 073 bytes with `Array.prototype.slice.call` and 366 056 bytes with `...`. Every little helps! 😂

js/src/common/utils/arrayFlatPolyfill.ts
```diff
@@ -3,23 +3,25 @@
...
+  Array.prototype['flat'] = function flat(this: any, depth?: number) {
     const realDepth = isNaN(depth as any) ? 1 : depth;
 
     return realDepth
       ? Array.prototype.reduce.call(
           this,
           function (acc, cur) {
             if (Array.isArray(cur)) {
-              acc.push.apply(acc, flat.call(cur, realDepth - 1));
+              (acc as Array<any>).push.apply(acc, flat.call(cur, realDepth - 1));
```
datitisev, line 14: Can this not be done in the `function (acc: )` argument type hinting?

davwheat: I would love to if I could figure out what Typescript was continuously complaining about...

I think `Array.prototype.reduce.call` expects the input to be anything, as it's a generic function, so it just uses `any` as the types for the accumulator and current value, which plays havoc when we want to use `.push`.

datitisev: Yeah I can't easily read that xD

davwheat: Y e a h...

I might just type the function args then do `as any` on the callback to prevent any complaints. That makes it a *bit* cleaner.

askvortsov1 (approved):

PR #2837, opened 2021-05-05
Fix unavailable translator catalogue for cli use

### LINKED ISSUE #100: Translations unavailable in cli after cache clearing
## Bug Report

**Current Behavior**

On current dev-master (post beta.16) when you `php flarum cache:clear` the `storage/locale` directory is cleared. When you now use the Translator by resolving it from ioc it can no longer use translations. 

The reason for this is that:

1. the service providers for Admin and Forum bind the Assets into ioc and there call AddTranslations for their specific frontend.
2. through the AddTranslations the Locale Manager is resolved (causing extensions to load their files) and the catalogue to be cached to disk
3. the console service provider has no logic at all related to this and as such is unable to restore the translations.

Understand that this bugs affects anything executed in cli, including scheduled tasks, cli command but also every single queue job that runs a non-default `sync` driver. Now usually this does not have a huge impact, because most forums regenerate their assets whenever the forum has visitors, but this will result in edge cases that are very hard to explain.

**Steps to Reproduce**
1. run `php flarum cache:clear`
2. create a command that uses the translator to parse a locale string
4. See raw locale string instead of translation

### LINKED ISSUE #2836: Locales not read in cli without cache
## Bug Report

**Current Behavior**

On current dev-master (post beta.16) when you `php flarum cache:clear` the `storage/locale` directory is cleared. When you now use the Translator by resolving it from ioc it can no longer use translations. 

The reason for this is that:

1. the service providers for Admin and Forum bind the Assets into ioc and there call AddTranslations for their specific frontend.
2. through the AddTranslations the Locale Manager is resolved (causing extensions to load their files) and the catalogue to be cached to disk
3. the console service provider has no logic at all related to this and as such is unable to restore the translations.

Understand that this bugs affects anything executed in cli, including scheduled tasks, cli command but also every single queue job that runs a non-default `sync` driver. Now usually this does not have a huge impact, because most forums regenerate their assets whenever the forum has visitors, but this will result in edge cases that are very hard to explain.

**Steps to Reproduce**
1. run `php flarum cache:clear`
2. create a command that uses the translator to parse a locale string
4. See raw locale string instead of translation

luceos:
Fixes flarum/issue-archive#100

When there's no locale cache, cli cannot translate. This re-instantiates
that cache when needed.

**Confirmed**

### DIFF
--- a/src/Console/ConsoleServiceProvider.php
+++ b/src/Console/ConsoleServiceProvider.php
@@ -68,7 +68,7 @@ class ConsoleServiceProvider extends AbstractServiceProvider
         }
 
         $this->container->make('flarum.locales')->getTranslator()->getCatalogue(
-            $this->container->make(SettingsRepositoryInterface::class)->get('default_locale', 'en')
+            $container->make(SettingsRepositoryInterface::class)->get('default_locale', 'en')
         );
     }
 }
--- a/src/Console/ConsoleServiceProvider.php
+++ b/src/Console/ConsoleServiceProvider.php
@@ -67,7 +67,7 @@ class ConsoleServiceProvider extends AbstractServiceProvider
             $scheduled['callback']($event);
         }
 
-        $this->container->make('flarum.locales')->getTranslator()->getCatalogue(
+        $container->make('flarum.locales')->getTranslator()->getCatalogue(
             $container->make(SettingsRepositoryInterface::class)->get('default_locale', 'en')
         );
     }
--- a/src/Console/ConsoleServiceProvider.php
+++ b/src/Console/ConsoleServiceProvider.php
@@ -15,6 +15,7 @@ use Flarum\Foundation\AbstractServiceProvider;
 use Flarum\Foundation\Console\AssetsPublishCommand;
 use Flarum\Foundation\Console\CacheClearCommand;
 use Flarum\Foundation\Console\InfoCommand;
+use Flarum\Settings\SettingsRepositoryInterface;
 use Illuminate\Console\Scheduling\Schedule as LaravelSchedule;
 use Illuminate\Console\Scheduling\ScheduleListCommand;
 use Illuminate\Console\Scheduling\ScheduleRunCommand;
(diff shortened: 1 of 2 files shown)


src/Console/ConsoleServiceProvider.php
```diff
@@ -65,5 +66,9 @@ public function boot(Container $container)
             $event = $schedule->command($scheduled['command'], $scheduled['args']);
             $scheduled['callback']($event);
         }
+
+        $this->container->make('flarum.locales')->getTranslator()->getCatalogue(
+            $this->container->make(SettingsRepositoryInterface::class)->get('default_locale', 'en')
```
askvortsov1, line 71: Should use $container, not $this->container. Also, why do we need to call `getCatalogue`?

askvortsov1 (changes requested):

src/Console/ConsoleServiceProvider.php
luceos: the `getCatalogue` method is needed for the catalogue to be written to cache/disk

luceos:
```suggestion
            $container->make(SettingsRepositoryInterface::class)->get('default_locale', 'en')
```

src/Console/ConsoleServiceProvider.php
```diff
@@ -65,5 +66,9 @@ public function boot(Container $container)
             $event = $schedule->command($scheduled['command'], $scheduled['args']);
             $scheduled['callback']($event);
         }
+
+        $this->container->make('flarum.locales')->getTranslator()->getCatalogue(
```
luceos, line 70:
```suggestion
        $container->make('flarum.locales')->getTranslator()->getCatalogue(
```

askvortsov1 (approved):

askvortsov1:
Actually, I would like to see a quick test for this. I can add it to the branch

askvortsov1 (changes requested):
The test I added is failing. Also, this would result in translator ALWAYS being resolved, not just when the console is actually used. I think this would make more sense in https://github.com/flarum/core/blob/ce01822ff6310dd5151a450bb61b12fa77bc5e9d/src/Console/Server.php#L21-L21.

PR #2842, opened 2021-05-08
Squash core migrations

### LINKED ISSUE #2258: Web timeouts when installing through web installer
## Bug Report

**Current Behavior**
From @luceos : 
Seen an increasingly number of support issues relating to installing through the web interface. The database installs incorrectly probably due to execution or proxy timeout.

askvortsov1:
**Fixes #2258**

- Squashes core's migrations into a schema dump. On my local machine, this shaves 7-8 seconds off of install time (likely to be more on slower machines/dbs). 
- Remove some now unnecessary parts of the migrator and migration repository public API

Unfortunately we can't do extension migrations here, as we have no way of filtering out data migrations other than hardcoding in a list. Their impact is relatively small though, so it shouldn't be that bad.

**Confirmed**

### DIFF
--- a/migrations/install.dump
+++ b/migrations/install.dump
@@ -1,0 +1,367 @@
+/*!40103 SET @OLD_TIME_ZONE=@@TIME_ZONE */;
+/*!40103 SET TIME_ZONE='+00:00' */;
+/*!40014 SET @OLD_UNIQUE_CHECKS=@@UNIQUE_CHECKS, UNIQUE_CHECKS=0 */;
+/*!40014 SET @OLD_FOREIGN_KEY_CHECKS=@@FOREIGN_KEY_CHECKS, FOREIGN_KEY_CHECKS=0 */;
+/*!40101 SET @OLD_SQL_MODE=@@SQL_MODE, SQL_MODE='NO_AUTO_VALUE_ON_ZERO' */;
+/*!40111 SET @OLD_SQL_NOTES=@@SQL_NOTES, SQL_NOTES=0 */;
+DROP TABLE IF EXISTS `db_prefix_access_tokens`;
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+/*!40101 SET character_set_client = utf8 */;
+CREATE TABLE `db_prefix_access_tokens` (
+  `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
+  `token` varchar(40) COLLATE utf8mb4_unicode_ci NOT NULL,
+  `user_id` int(10) unsigned NOT NULL,
+  `last_activity_at` datetime NOT NULL,
+  `created_at` datetime NOT NULL,
+  `type` varchar(100) COLLATE utf8mb4_unicode_ci NOT NULL,
+  `title` varchar(150) COLLATE utf8mb4_unicode_ci DEFAULT NULL,
+  `last_ip_address` varchar(45) COLLATE utf8mb4_unicode_ci DEFAULT NULL,
+  `last_user_agent` varchar(255) COLLATE utf8mb4_unicode_ci DEFAULT NULL,
+  PRIMARY KEY (`id`),
+  UNIQUE KEY `db_prefix_access_tokens_token_unique` (`token`),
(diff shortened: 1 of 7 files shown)


src/Database/Migrator.php
```diff
@@ -242,6 +263,32 @@ public function resolve($path, $file)
...
+     *
+     * @param string $path to the directory containing the dump.
+     */
+    public function installFromSchema(string $path)
+    {
+        $schemaPath = "$path/install.dump";
+
+        $currDumpFile = file_get_contents($schemaPath);
+
+        file_put_contents($schemaPath, str_replace('db_prefix_', $this->tablePrefix, $currDumpFile));
```
clarkwinkelmann, line 277: Is this writing over the original file?

Won't this fail if the user delete their `config.php`, the database and try to install again with a different prefix?

I see later in the code this might be reverted (line 289?) but what if the installation errors in between?

Wouldn't it be safer to create a second temporary file?

src/Database/Migrator.php
```diff
@@ -271,22 +271,23 @@ public function resolve($path, $file)
     public function installFromSchema(string $path)
     {
         $schemaPath = "$path/install.dump";
+        $prefixedSchemaPath = "$path/install_prefixed.dump.tmp";
```
clarkwinkelmann, line 274: One more question :sweat_smile: 

Isn't this inside of the `vendor` folder? Do we expect that folder to be writable by the user the installer runs as? Do we have access to the storage folder path at that point? Or should we use PHP's `tmpfile` feature?

migrations/install.dump
```diff
@@ -0,0 +1,367 @@
...
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;
+/*!40101 SET character_set_client = @saved_cs_client */;
+/*!40103 SET TIME_ZONE=@OLD_TIME_ZONE */;
+
+/*!40101 SET SQL_MODE=@OLD_SQL_MODE */;
+/*!40014 SET FOREIGN_KEY_CHECKS=@OLD_FOREIGN_KEY_CHECKS */;
+/*!40014 SET UNIQUE_CHECKS=@OLD_UNIQUE_CHECKS */;
+/*!40111 SET SQL_NOTES=@OLD_SQL_NOTES */;
+
+INSERT INTO `db_prefix_migrations` VALUES (1,'2015_02_24_000000_create_access_tokens_table',NULL);
```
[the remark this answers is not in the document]
askvortsov1: The migratons still exist: we're not removing their files.

src/Database/Console/GenerateDumpCommand.php
```diff
@@ -0,0 +1,87 @@
...
+            ->setDescription('Dump DB schema');
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function fire()
+    {
+        $dumpPath = __DIR__.'/../../../migrations/install.dump';
+        /** @var Connection */
```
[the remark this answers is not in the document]
askvortsov1: Works for me locally.

src/Console/ConsoleServiceProvider.php
```diff
@@ -46,6 +46,8 @@ public function register()
                 ResetCommand::class,
                 ScheduleListCommand::class,
                 ScheduleRunCommand::class
+                // Used internally to create DB dumps before major releases.
+                // \Flarum\Database\Console\GenerateDumpCommand::class
```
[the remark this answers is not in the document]
askvortsov1: I wanted to put it in the Flarum CLI, but didn't have time to rewrite it so that it worked for this.

src/Database/Migrator.php
askvortsov1: I'm not sure what a better solution would be. Storage folder maybe?

askvortsov1: Actually we can't do storage, since the paths object might not exist yet.

clarkwinkelmann (approved):
That's probably good like is is now. If there's any issue with the approach QA should hopefully find it.

tankerkiller125 (approved):
I looked at this previously and it works well locally.

PR #2843, opened 2021-05-08
Add tooltip component

davwheat:
**Prevents breaking changes as a result of flarum/framework#2697. Progresses flarum/issue-archive#179.**

This is still WIP. Created as a draft now to allow for any comments to be made while the component is still being made.

This adds a new Tooltip component to core. This should be used by extensions instead of `$.tooltip` to aid in the shift to other tooltip libraries (hopefully CSS tooltips!) in the future.

I've tried to include the main tooltip options that are used by extensions, such as `delay` and `html`, but these won't be easy or possible to keep if we shift to CSS tooltips, so they've been marked as deprecated and only exist to allow extensions to more easily shift to this component.

Will the deprecation of `html` and `delay` cause an issue with flarum/framework#2697 as they aren't available in that? Do we consider removing deprecated features within a stable release ok, bearing in mind there is a fairly bold warning in the docblock, or should we just not implement them as part of the component in the first place.

I think implementing them as deprecated features will help ext devs migrate to this component faster and more easily now, bearing in mind the large number of other changes they need to perform.

Should we also add a deprecation warning to the tooltip helper when it's called from outside the `Tooltip` component to make this more obvious to devs?

**Screenshot**

**Confirmed**

**Required changes:**

### DIFF
--- a/js/src/common/components/Tooltip.tsx
+++ b/js/src/common/components/Tooltip.tsx
@@ -141,6 +141,12 @@ export default class Tooltip extends Component<TooltipAttrs> {
       this.$().tooltip('destroy');
       this.createTooltip();
     }
+
+    if (this.attrs.tooltipVisible === true) {
+      this.$().tooltip('show');
+    } else if (this.attrs.tooltipVisible === false) {
+      this.$().tooltip('hide');
+    }
   }
 
   private createTooltip() {
@@ -150,7 +156,12 @@ export default class Tooltip extends Component<TooltipAttrs> {
       delay,
       // This will have no effect when switching to CSS tooltips
       html = false,
+      tooltipVisible,
     } = this.attrs;
+
+    const trigger = (typeof tooltipVisible === 'boolean'
+      ? 'manual'
+      : classList('hover', [showOnFocus && 'focus'])) as TooltipCreationOptions['trigger'];
 
     // https://getbootstrap.com/docs/3.3/javascript/#tooltips-options
     this.$().tooltip(
@@ -159,7 +170,7 @@ export default class Tooltip extends Component<TooltipAttrs> {
         delay,
         placement: position,
         // Fancy "hack" to assemble the trigger string
-        trigger: classList('hover', [showOnFocus && 'focus']) as TooltipCreationOptions['trigger'],
+        trigger,
       },
       // @ts-expect-error We don't want this arg to be part of the public API. It only exists to prevent deprecation warnings when using `$.tooltip` in this component.
       'DANGEROUS_tooltip_jquery_fn_deprecation_exempt'
--- a/js/src/common/components/Tooltip.tsx
+++ b/js/src/common/components/Tooltip.tsx
@@ -86,7 +86,10 @@ export interface TooltipAttrs extends ComponentAttrs {
  */
 export default class Tooltip extends Component<TooltipAttrs> {
   private oldText: string = '';
+  private oldVisibility: boolean | undefined;
+
   private shouldRecreateTooltip: boolean = false;
+  private shouldChangeTooltipVisibility: boolean = false;
 
   view(vnode) {
     const { children } = vnode;
@@ -119,6 +122,11 @@ export default class Tooltip extends Component<TooltipAttrs> {
       this.shouldRecreateTooltip = true;
     }
 
+    if (tooltipVisible !== this.oldVisibility) {
+      this.oldVisibility = this.attrs.tooltipVisible;
+      this.shouldChangeTooltipVisibility = true;
+    }
+
     return (
       <div title={realText} className={classList('tooltip-container', `tooltip-container--${containerType}`, className, classes)} {...attrs}>
         {children}
@@ -138,14 +146,35 @@ export default class Tooltip extends Component<TooltipAttrs> {
 
   private recreateTooltip() {
     if (this.shouldRecreateTooltip) {
-      this.$().tooltip('destroy');
+      this.$().tooltip(
+        'destroy',
+        // @ts-expect-error We don't want this arg to be part of the public API. It only exists to prevent deprecation warnings when using `$.tooltip` in this component.
+        'DANGEROUS_tooltip_jquery_fn_deprecation_exempt'
+      );
       this.createTooltip();
+      this.shouldRecreateTooltip = false;
     }
 
+    if (this.shouldChangeTooltipVisibility) {
+      this.shouldChangeTooltipVisibility = false;
+      this.updateVisibility();
+
+    }
+  }
+
+  private updateVisibility() {
     if (this.attrs.tooltipVisible === true) {
-      this.$().tooltip('show');
+      this.$().tooltip(
+        'show',
+        // @ts-expect-error We don't want this arg to be part of the public API. It only exists to prevent deprecation warnings when using `$.tooltip` in this component.
+        'DANGEROUS_tooltip_jquery_fn_deprecation_exempt'
+      );
     } else if (this.attrs.tooltipVisible === false) {
-      this.$().tooltip('hide');
+      this.$().tooltip(
+        'hide',
+        // @ts-expect-error We don't want this arg to be part of the public API. It only exists to prevent deprecation warnings when using `$.tooltip` in this component.
+        'DANGEROUS_tooltip_jquery_fn_deprecation_exempt'
+      );
     }
   }
 
--- a/js/src/common/components/Tooltip.tsx
+++ b/js/src/common/components/Tooltip.tsx
@@ -1,10 +1,10 @@
-import Component, { ComponentAttrs } from '../Component';
+import Component from '../Component';
 import type Mithril from 'mithril';
 import classList from '../utils/classList';
 import { TooltipCreationOptions } from '../../../@types/tooltips';
 import extractText from '../utils/extractText';
 
-export interface TooltipAttrs extends ComponentAttrs {
+export interface TooltipAttrs extends Mithril.CommonAttributes<TooltipAttrs, Tooltip> {
   /**
    * Tooltip textual content.
    *
@@ -12,13 +12,6 @@ export interface TooltipAttrs extends ComponentAttrs {
    * into strings.
    */
   text: string | string[];
-  /**
-   * Defines the type of container to use. Chosen option defines the `display`
-   * property of the container element in CSS.
-   *
-   * Default: `'block'`.
-   */
-  containerType?: 'block' | 'inline' | 'inline-block';
   /**
    * Manually show tooltip. `false` will show based on cursor events.
    *
@@ -41,8 +34,8 @@ export interface TooltipAttrs extends ComponentAttrs {
    * Whether HTML content is allowed in the tooltip.
    *
    * **Warning:** this is a possible XSS attack vector. This option shouldn't
-   * be used wherever possible, and will not work when we migrate to CSS-only
-   * tooltips.
+   * be used wherever possible, and may not work when we migrate to another
+   * tooltip library. Be prepared for this to break in Flarum stable.
    *
    * Default: `false`.
    *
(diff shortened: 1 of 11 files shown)


js/src/common/components/Tooltip.tsx
```diff
@@ -0,0 +1,125 @@
...
+ *          </Tooltip>
+ */
+export default class Tooltip extends Component<TooltipAttrs> {
+  view(vnode) {
+    const { children } = vnode;
+
+    // We remove these to get the remaining attrs to pass to the DOM element
+    const { text, inline, tooltipVisible, showOnFocus, position, html, delay, ...attrs } = this.attrs;
+
+    if (inline) {
```
askvortsov1, line 83: Could we return the children directly instead of wrapping it in an additional level of div/span?

askvortsov1 (changes requested):

js/src/common/components/Tooltip.tsx
askvortsov1: What challenges could that pose? I'm not familiar with this particular technology

davwheat: If we have:
```html
<div data-tooltip aria-label="Very cool">
  Cool beans!
</div>
```

Where that `<div>` already has `::before` or `::after` styles, that would interfere with with the pseudoelement styles defined by the use of `data-tooltip` on it. This could result in very odd tooltip styling due to conflicts.

I initially made this component to work around this issue by wrapping the element in a container.

Maybe a `wrap: 'block' | 'inline'` attr would be better?

askvortsov1: Since you've been working a lot more with this than I have, is there a danger to inserting unexpected wrapping elements other than breakage of `>` css specifiers?

askvortsov1:
No need for both the issue and the PR on the milestone

js/src/common/components/Tooltip.tsx
davwheat: Not that I can think of off of the top of my head, no.

askvortsov1: > Maybe we could allow for the ability to specify a custom container element, then create that using m()?

This seems overly hacky IMO. Let's try the `div` and `span` change.

davwheat: Just had to swap it out for using `<div>` with `inline-block` due to an issue with padding on `display: inline`.

`inline-block` seems to work in all areas of core so far.

davwheat:
I'd love to hear opinions on the deprecation warning added in https://github.com/flarum/core/pull/2843/commits/356df4d3d76a31a3909d61766074bfe8fc958952.

I don't know how I feel about the way this is implemented, but it was the best I could think of. Maybe my brain at a time other than 3am might be able to do a better job!

It only serves to inform extension devs that their current tooltip method is deprecated and that they should migrate to the new version.

davwheat:
I've forgotten to implement `tooltipVisible` 🙈 

I'll do that tomorrow.

js/src/common/components/Tooltip.tsx
```diff
@@ -0,0 +1,179 @@
...
+    // We need to recreate the tooltip if the text has changed
+    if (realText !== this.oldText) {
+      this.oldText = realText;
+      this.shouldRecreateTooltip = true;
+    }
+
+    return (
+      <div title={realText} className={classList('tooltip-container', `tooltip-container--${containerType}`, className, classes)} {...attrs}>
+        {children}
+      </div>
```
SychO9, line 125: is the HTML container necessary ? can't we do whatever we're doing directly on the actual node ?

SychO9: actually nevermind

js/src/common/index.js
```diff
@@ -25,3 +25,18 @@ import * as Extend from './extend/index';
...
+
+const tooltipGen = $.fn.tooltip;
+
+// Remove in a future version of Flarum.
+$.fn.tooltip = function (options, caller) {
+  // Show a warning when `$.tooltip` is used outside of the Tooltip component.
+  // This functionality is deprecated and should not be used.
+  if (!['DANGEROUS_tooltip_jquery_fn_deprecation_exempt'].includes(caller)) {
+    console.warn(
+      "Calling `$.tooltip` is now deprecated. Please use the `<Tooltip>` component exposed by flarum/core instead. `$.tooltip` may be removed in a future version of Flarum.\n\nIf this component doesn't meet your requirements, please open an issue: https://github.com/flarum/core/issues/new?assignees=davwheat&labels=type/bug,needs-verification&template=bug-report.md&title=Tooltip%20component%20unsuitable%20for%20use%20case"
```
askvortsov1, line 37: Shouldnt this be negated?

askvortsov1 (commented):
The more I look at this, the less I like that the tooltip is created as an additional HTML node. Adding a tooltip is a **modification** to an existing element, not the creation of a new element. 

Would pseudoelement conflicts arise when a JS tooltip is used? Conceptually a CSS-only tooltip is cool, but the more I see workarounds needed to implement it, the less I'm convinced that CSS is ready.

js/src/common/index.js
davwheat: It is ;)

I noticed that after.

js/src/common/components/Tooltip.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+    if (tooltipVisible !== this.oldVisibility) {
+      this.oldVisibility = this.attrs.tooltipVisible;
+      this.shouldChangeTooltipVisibility = true;
+    }
+
+    return (
+      <div title={realText} className={classList('tooltip-container', `tooltip-container--${containerType}`, className, classes)} {...attrs}>
+        {children}
+      </div>
```
SychO9, line 133: would it be possible to either directly apply the tooltip logic on the node, or add an `as` prop that allows specifying the tag of the tooltip container, so that in situations such as tooltiping a link, we are not forced to have an html container and therefore not need the `inline-block` and `block` class modifiers, as I can see this as a potential theming issue

SychO9 (commented):
I'm bad at typescript so I'd rather not review this, but I do have a comment

js/src/common/components/Tooltip.tsx
askvortsov1: https://github.com/flarum/core/pull/2843#discussion_r628808845 lol

SychO9: ah.. well, I'd honestly still like to be able to specify the container's HTML tag through a prop, or have the option to not add an HTML container and have the tooltipping instead directly on the node.

As long as that container is not forced and there is a way to not have it and still be able to use tooltips.

askvortsov1: Sorry, that was a very ambiguous lol. I agree with you on this one. Before other components were changed, it kinda made sense, but looking at it again (especially with class names going on the tooltip :see_no_evil:), I've reverted to https://github.com/flarum/core/pull/2843#pullrequestreview-655105253,

davwheat:
Ok, I've changed the component to apply the tooltip directly to the first child of the component.

Along comes a lot of error detection measures to ensure devs find out if they're using this component incorrectly.

davwheat:
This has made the bundle a lot bigger, but a chunk of this can be shaved off if you don't think that manually showing/hiding tooltips is important. I personally don't at all, as tooltips shouldn't be used for that.

js/src/common/components/Tooltip.tsx
```diff
@@ -0,0 +1,275 @@
...
+ *
+ * @example <caption>Incorrect usage</caption>
+ *          // This is wrong! Surround the children with a <span> or similar.
+ *          <Tooltip text="This won't work">
+ *            Click
+ *            <a href="/">here</a>
+ *          </Tooltip>
+ */
+export default class Tooltip extends Component<TooltipAttrs> {
+  private firstChild: Mithril.Vnode<any, any> | null = null;
```
askvortsov1, line 101: Can't we just `private firstChild!: Mithril.Vnode<any, any>;`

js/src/common/components/Tooltip.tsx
```diff
@@ -0,0 +1,275 @@
...
+
+    return children;
+  }
+
+  oncreate(vnode: Mithril.VnodeDOM<TooltipAttrs, this>) {
+    super.oncreate(vnode);
+
+    const domNode = (this.firstChild as Mithril.VnodeDOM<any, any>).dom as HTMLElement;
+
+    if (!domNode.isSameNode(this.childDomNode)) {
```
askvortsov1, line 179: Are these operations costly?

js/src/common/components/Tooltip.tsx
```diff
@@ -0,0 +1,275 @@
...
+      this.shouldRecreateTooltip = true;
+    }
+
+    this.recreateTooltip();
+  }
+
+  onupdate(vnode: Mithril.VnodeDOM<TooltipAttrs, this>) {
+    super.onupdate(vnode);
+
+    const domNode = (this.firstChild as Mithril.VnodeDOM<any, any>).dom as HTMLElement;
```
askvortsov1, line 190: Can we factor out the repeated logic?

askvortsov1 (changes requested):
Overall, I like this very much! Left some relatively minor style comments.

js/src/common/components/Tooltip.tsx
```diff
@@ -0,0 +1,275 @@
...
+    if (this.shouldChangeTooltipVisibility) {
+      this.shouldChangeTooltipVisibility = false;
+      this.updateVisibility();
+    }
+  }
+
+  private updateVisibility() {
+    if (this.childDomNode === null) return;
+
+    if (this.attrs.tooltipVisible === true) {
```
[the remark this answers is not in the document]
davwheat: It's bool or undefined. We need the strong checks to ensure we don't attempt to hide or show the tooltip when it's undefined.

We don't need the true one, but I thought matching them looked nicer and less like a mistake.

js/src/common/components/Tooltip.tsx
davwheat: It's identical to calling `===`, but with better browser support.

https://dom.spec.whatwg.org/#dom-node-issamenode

js/src/common/components/Tooltip.tsx
```diff
@@ -0,0 +1,275 @@
...
+
+    if (!domNode.isSameNode(this.childDomNode)) {
+      this.childDomNode = domNode;
+      this.shouldRecreateTooltip = true;
+    }
+
+    this.recreateTooltip();
+  }
+
+  private recreateTooltip() {
```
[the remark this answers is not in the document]
davwheat: As far as I'm aware, there's no reason to create another class that inherits from this one, which is the only reason to use `protected` over `private`. Do you want me to switch it just in case?

js/src/common/components/Tooltip.tsx
davwheat: I don't like this as it hides the fact that `firstChild` might not be set for some reason by removing undefined/null from its typings.

js/src/common/components/Tooltip.tsx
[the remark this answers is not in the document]
askvortsov1: True, let's keep it private.

askvortsov1:
Other than these, code works locally.

js/src/common/components/Tooltip.tsx
davwheat: Yes, but that's only done after we know it won't be null.

Your suggestion would hide the original value (actually undefined, I'd guess) in the original typings, meaning that if we edit the file later and attempt to use the variable before we know it's not null, we could unwittingly introduce issues.

askvortsov1 (approved):

tankerkiller125 (approved):

PR #2845, opened 2021-05-09
Export registry and Webpack 5

askvortsov1:
**Fixes #0000**

- Add support for webpack 5 (by separating extension JS with semicolons)
- Don't export objects directly in core, this makes it simpler to auto-export in the loader
- Use webpack 5 in core
- Implement the Export Registry as discussed in https://discuss.flarum.org/d/26942-code-splitting-lazy-load/2

We need to make sure the bidi call runs after the export registry has been instantiated. I had to move the patchMithril call to application boot for this. I don't think this should cause isssues.

**Confirmed**

**Required changes:**

### DIFF
--- a/js/src/common/FlarumRegistry.ts
+++ b/js/src/common/FlarumRegistry.ts
@@ -1,0 +1,60 @@
+interface ExportRegistry {
+  moduleExports: object;
+
+  onLoads: object;
+
+  /**
+   * Add an instance to the registry.
+   * This serves as the equivalent of `flarum.core.compat[id] = object`
+   */
+  add(namespace: string, id: string, object: any): void;
+
+  /**
+   * Add a function to run when object of id "id" is added (or overriden).
+   * If such an object is already registered, the handler will be applied immediately.
+   */
+  onLoad(namespace: string, id: string, handler: Function): void;
+
+  /**
+   * Retrieve an object of type `id` from the registry.
+   */
+  get(namespace: string, id: string): any;
+}
+
+export default class FlarumRegistry implements ExportRegistry {
+  moduleExports = new Map<string, any>();
+  onLoads = new Map<string, Function[]>();
+
+  protected genKey(namespace: string, id: string): string {
+    return `${namespace};${id}`;
+  }
+
+  add(namespace: string, id: string, object: any) {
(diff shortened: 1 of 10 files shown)


js/src/common/FlarumRegistry.ts
```diff
@@ -0,0 +1,60 @@
+interface ExportRegistry {
+  moduleExports: object;
+
+  onLoads: object;
+
+  /**
+   * Add an instance to the registry.
+   * This serves as the equivalent of `flarum.core.compat[id] = object`
+   */
+  add(namespace: string, id: string, object: any): void;
```
davwheat, line 10:
```suggestion
  add(namespace: string, id: string, object: unknown): void;
```

js/src/common/FlarumRegistry.ts
```diff
@@ -0,0 +1,60 @@
+interface ExportRegistry {
```
davwheat, line 1: Should we export this, just in case?

js/src/common/FlarumRegistry.ts
```diff
@@ -0,0 +1,60 @@
...
+   */
+  onLoad(namespace: string, id: string, handler: Function): void;
+
+  /**
+   * Retrieve an object of type `id` from the registry.
+   */
+  get(namespace: string, id: string): any;
+}
+
+export default class FlarumRegistry implements ExportRegistry {
```
davwheat, line 24: If we use `implements`, I don't think we need to define the types inside the class body.

js/src/common/FlarumRegistry.ts
```diff
@@ -0,0 +1,60 @@
...
+  /**
+   * Add a function to run when object of id "id" is added (or overriden).
+   * If such an object is already registered, the handler will be applied immediately.
+   */
+  onLoad(namespace: string, id: string, handler: Function): void;
+
+  /**
+   * Retrieve an object of type `id` from the registry.
+   */
+  get(namespace: string, id: string): any;
```
davwheat, line 21: We use the result of this with a spread operator in an array. Do we know that this should return an Array? If so, we should probably use something like `unknown[]` as the return type instead.

PR #2846, opened 2021-05-09
Some minor JS cleanup

askvortsov1:
See commits for details


js/src/common/Translator.ts
```diff
@@ -4,34 +4,44 @@ import username from './helpers/username';
...
+   * @ts-ignore
+   */
+  protected formatter = new RichMessageFormatter(null, this.formatterTypeHandlers(), mithrilRichHandler);
+
+  setLocale(locale: string) {
+    this.formatter.locale = locale;
+  }
 
-    this.formatter = new RichMessageFormatter(null, this.formatterTypeHandlers(), mithrilRichHandler);
+  addTranslations(translations: { [key: string]: string }) {
```
davwheat, line 23: We should use the `Record` type for this.

```suggestion
  addTranslations(translations: Record<string, string>) {
```

js/src/common/Translator.ts
```diff
@@ -45,7 +55,7 @@ export default class Translator {
     return parameters;
   }
 
-  trans(id, parameters) {
+  trans(id: string, parameters: any) {
```
davwheat, line 58: Could we replace `any` with something else? Should `parameters` also be marked as optional using `?:`?

Maybe `parameters?: Record<string, unknown>` would work?

js/src/common/Translator.ts
```diff
@@ -4,34 +4,44 @@ import username from './helpers/username';
...
-
-  preprocessParameters(parameters) {
+  /**
+   * A temporary system to preprocess parameters.
+   * Should not be used by extensions.
+   * TODO: An extender will be added in v1.x.
+   *
+   * @internal
+   */
+  protected preprocessParameters(parameters) {
```
davwheat, line 44: See below comment about type -- same type should be used up here.

`parameters: Record<string, unknown>`?

js/src/common/states/ModalManagerState.js
```diff
@@ -11,22 +11,11 @@ export default class ModalManagerState {
    * @public
    */
   show(componentClass, attrs) {
-    // Breaking Change Compliance Warning, Remove in Beta 15.
     if (!(componentClass.prototype instanceof Modal)) {
       // This is duplicated so that if the error is caught, an error message still shows up in the debug console.
       console.error('The ModalManager can only show Modals');
       throw new Error('The ModalManager can only show Modals');
```
davwheat, line 17: Saves bytes and we're already editing this area anyway :)

```suggestion
      const invalidModalWarning = 'The ModalManager can only show Modals.';
      console.error(invalidModalWarning);
      throw new Error(invalidModalWarning);
```

js/src/common/Translator.ts
```diff
@@ -4,34 +4,44 @@ import username from './helpers/username';
...
-    this.translations = {};
+  /**
+   * A map of translation keys to their translated values.
+   */
+  translations = {};
+
+  /**
+   * The underlying ICU MessageFormatter util.
+   *
+   * @ts-ignore
```
davwheat, line 15: What's this ignoring?

davwheat (changes requested):

js/src/common/Translator.ts
askvortsov1: That makes sense to me. Although then we should default to an empty object.

js/src/common/Translator.ts
askvortsov1: Hmm, it's not erroring anymore. Weird...

davwheat (approved):

PR #2851, opened 2021-05-10
Add page separators and load previous button to DiscussionList

dsevillamartin:
**Refs #2781**

https://user-images.githubusercontent.com/6401250/117724452-c2c5c200-b1b1-11eb-9bb6-4c49cd28849c.mp4

**Confirmed**

**Required changes:**

### DIFF
--- a/js/src/forum/components/DiscussionList.js
+++ b/js/src/forum/components/DiscussionList.js
@@ -43,11 +43,14 @@ export default class DiscussionList extends Component {
       <div className={'DiscussionList' + (state.isSearchResults() ? ' DiscussionList--searchResults' : '')}>
         <ul className="DiscussionList-discussions">
           {state.getPages().map((pg) => {
-            return pg.items.map((discussion) => (
-              <li key={discussion.id()} data-id={discussion.id()}>
-                {DiscussionListItem.component({ discussion, params })}
-              </li>
-            ));
+            return [
+              <hr key={`page-${pg.number}`} data-page={pg.number} />,
+              ...pg.items.map((discussion) => (
+                <li key={discussion.id()} data-id={discussion.id()}>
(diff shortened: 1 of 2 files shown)


askvortsov1 (changes requested):
Could we use anchorScroll so it doesn't jump through the prev loaded page?

datitisev:
> Could we use anchorScroll so it doesn't jump through the prev loaded page?

I just tried to do so, but it doesn't work as `anchorScroll` simply runs the `callback()` expecting functionality to be synchronous. As `loadPrev` is asynchronous (Promise), it doesn't work...

I tried adding it manually after the promise remained, but wasn't able to get it quite right - I'm not very good at this scroll stuff. I also don't think it's necessarily a good idea either, because as `m.redraw()` is called within `PaginatedListState`, the user could've scrolled away from between before calling the Promise & after it is resolved.

davwheat:
Can we use the `display=block` on the spinner instead of manually setting the height?

datitisev:
> Can we use the display=block on the spinner instead of manually setting the height?

This wouldn't change the height as far as I'm aware? The reason that CSS rule was added was to occupy around the same amount of space as the button iirc.

davwheat:
[It would](https://github.com/flarum/core/blob/d1e38558c5a0bab54babd3c62a85a93d1a63e0ad/less/common/LoadingIndicator.less#L47), but now I look at it, it might be too tall.

datitisev:
@davwheat Oh, I thought you meant simply `display: block;` CSS, not an attribute on the `<LoadingIndicator />` component.

davwheat:
I wish I picked a better name for that attr now 😅

askvortsov1:

PR #2856, opened 2021-05-12
Update core to use new Webpack config, `flarum-tsconfig`, and build action

davwheat:
**Before merging this, we need to tag and publish `flarum-webpack-config` and update the version here.**


js/tsconfig.json
```diff
@@ -1,21 +1,11 @@
...
-    "emitDecoratorMetadata": true,
-    "experimentalDecorators": true,
-    "esModuleInterop": true,
-    "moduleResolution": "node",
-    "target": "es6",
-    "jsx": "preserve",
-    "lib": ["es2015", "es2017", "dom"],
-    "allowSyntheticDefaultImports": true
+    // This will output typings to `dist-typings`
+    "declarationDir": "./dist-typings"
```
askvortsov1, line 9: Do we still want this even though we have the typings repo?

askvortsov1 (changes requested):

js/tsconfig.json
davwheat: It's mainly for extensions to have their typings in a standardised place.

We discussed moving typings to the core repo -- do you want to do that?

askvortsov1: If possible, I think we should try. Not having to download typings as a separate package, and us not having to update / publish it separately should make things a lot simpler and easier. And that's always good.

davwheat: Totally! It might be worth keeping both, though. That way people can dev outside of a local Flarum install. I've had to do that a few times every now and again.

davwheat: Sticking this here: https://github.com/marketplace/actions/push-directory-to-another-repository

askvortsov1 (commented):
Should BundleAnalyzer also be moved to webpack config?

Also, when we commit this, it should compile typings directly in this repo, right?

davwheat:
That might be a good idea, actually!

Yes, typings will be added to this repo under `js/dist-typings`.

askvortsov1:
> That might be a good idea, actually!
> 
> Yes, typings will be added to this repo under `js/dist-typings`.

Should we move the tooltip ones there?

davwheat:
No, they should be automatically added as part of the typings compilation process. (About to double check that...)

davwheat:
Looks like `tsc` doesn't copy across `.d.ts` files from source "by design" :roll_eyes: 

We'll need to add it the `build-typings` script via a `cp` command, or change them to `.ts` files.

askvortsov1:
> via a cp command

I like this one.

davwheat:
I was thinking the exact opposite! :joy: 

Changing to `.ts` requires no syntax modification, and automates the process compared to a manual cp that might need to be updated in the future.

askvortsov1:
> I was thinking the exact opposite! joy

I don't have a strong preference. As long as renaming them to `.ts` doesn't break anything I'm good with it.

davwheat:
I kept with `.d.ts` as it keeps the meaning (that they're type declarations) clear.

askvortsov1:
> Should BundleAnalyzer also be moved to webpack config?
> 
> Also, when we commit this, it should compile typings directly in this repo, right?

Are we doing this here, or in a separate PR?

davwheat:
> Are we doing this here, or in a separate PR?

Might as well do it here while we're doing all the rest too

davwheat:
PR opened: https://github.com/flarum/flarum-webpack-config/pull/12

askvortsov1 (approved):

Issue #2700, opened 2021-03-17
General pre-stable cleanup tasks


askvortsov1:
- [ ] Clean up and standardize backend docblocks
- [ ] Clean up and standardize frontend docblocks
- [ ] Clean up private vs protected usage
- [ ] Add (or remove everywhere) file top comment blocks in JS

davwheat:
Would it be worth adding "update JS imports in bundled exts" to the list?

All the bundled exts still use `flarum/components` (or similar) imports rather than the forum/admin/common namespaced imports.

askvortsov1:
If there's time, we should also add prettier to all exts and run their JS through that.

luceos:
As nothing else was tackled, let's "clean up" for the upcoming 1.1 release.

PR #2859, opened 2021-05-12
Backend cleanup

SychO9:
**Part of #2700**

* Extender docblocks cleanup
* Explicit type hinting in extenders
* Remove beta references
* Remove BC layers

Any other classes/methods we should be marking as internal ?
Any that we actually should not be ?

### DIFF
--- a/src/Extend/ApiController.php
+++ b/src/Extend/ApiController.php
@@ -48,7 +48,7 @@ class ApiController implements ExtenderInterface
      *
      * @return self
      */
-    public function prepareDataQuery($callback)
+    public function prepareDataQuery($callback): self
     {
         $this->beforeDataCallbacks[] = $callback;
 
@@ -70,7 +70,7 @@ class ApiController implements ExtenderInterface
      *
      * @return self
      */
-    public function prepareDataForSerialization($callback)
+    public function prepareDataForSerialization($callback): self
     {
         $this->beforeSerializationCallbacks[] = $callback;
 
@@ -91,7 +91,7 @@ class ApiController implements ExtenderInterface
      *
      * @return self
      */
-    public function setSerializer(string $serializerClass, $callback = null)
+    public function setSerializer(string $serializerClass, $callback = null): self
     {
         $this->serializer = [$serializerClass, $callback];
 
@@ -112,7 +112,7 @@ class ApiController implements ExtenderInterface
      *
      * @return self
      */
-    public function addInclude($name, $callback = null)
+    public function addInclude($name, $callback = null): self
     {
         $this->addIncludes[] = [$name, $callback];
 
@@ -133,7 +133,7 @@ class ApiController implements ExtenderInterface
      *
      * @return self
      */
-    public function removeInclude($name, $callback = null)
+    public function removeInclude($name, $callback = null): self
     {
         $this->removeIncludes[] = [$name, $callback];
 
@@ -154,7 +154,7 @@ class ApiController implements ExtenderInterface
      *
      * @return self
      */
-    public function addOptionalInclude($name, $callback = null)
+    public function addOptionalInclude($name, $callback = null): self
     {
         $this->addOptionalIncludes[] = [$name, $callback];
 
@@ -175,7 +175,7 @@ class ApiController implements ExtenderInterface
      *
      * @return self
      */
-    public function removeOptionalInclude($name, $callback = null)
+    public function removeOptionalInclude($name, $callback = null): self
     {
         $this->removeOptionalIncludes[] = [$name, $callback];
 
@@ -196,7 +196,7 @@ class ApiController implements ExtenderInterface
      *
      * @return self
      */
-    public function setLimit(int $limit, $callback = null)
+    public function setLimit(int $limit, $callback = null): self
     {
         $this->limit = [$limit, $callback];
 
@@ -217,7 +217,7 @@ class ApiController implements ExtenderInterface
      *
      * @return self
      */
-    public function setMaxLimit(int $max, $callback = null)
+    public function setMaxLimit(int $max, $callback = null): self
     {
         $this->maxLimit = [$max, $callback];
 
@@ -238,7 +238,7 @@ class ApiController implements ExtenderInterface
      *
      * @return self
      */
-    public function addSortField($field, $callback = null)
+    public function addSortField($field, $callback = null): self
     {
         $this->addSortFields[] = [$field, $callback];
 
@@ -259,7 +259,7 @@ class ApiController implements ExtenderInterface
      *
      * @return self
      */
-    public function removeSortField($field, $callback = null)
+    public function removeSortField($field, $callback = null): self
     {
         $this->removeSortFields[] = [$field, $callback];
 
@@ -280,7 +280,7 @@ class ApiController implements ExtenderInterface
      *
      * @return self
      */
-    public function setSort(array $sort, $callback = null)
+    public function setSort(array $sort, $callback = null): self
     {
         $this->sort = [$sort, $callback];
 
@@ -298,10 +298,10 @@ class ApiController implements ExtenderInterface
      * To force load the relationship, both levels have to be specified,
      * example: ['relation', 'relation.subRelation'].
      *
-     * @param string|array
+     * @param string[]
      * @return self
      */
-    public function load($relations)
+    public function load($relations): self
     {
         $this->load = array_merge($this->load, (array) $relations);
 
--- a/src/Extend/ApiController.php
+++ b/src/Extend/ApiController.php
@@ -32,7 +32,7 @@ class ApiController implements ExtenderInterface
     private $load = [];
 
     /**
-     * @param string $controllerClass The ::class attribute of the controller you are modifying.
+     * @param string $controllerClass: The ::class attribute of the controller you are modifying.
      *                                This controller should extend from \Flarum\Api\Controller\AbstractSerializeController.
      */
     public function __construct(string $controllerClass)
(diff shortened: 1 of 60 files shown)


src/Extend/ApiController.php
```diff
@@ -228,10 +298,10 @@ public function setSort(array $sort, $callback = null)
      * To force load the relationship, both levels have to be specified,
      * example: ['relation', 'relation.subRelation'].
      *
-     * @param string|array
+     * @param string[]
```
clarkwinkelmann, line 301: the name of the parameter is missing, is this intentional? (`$relations`)

The implementation below seems to be designed for string parameter as well. Otherwise if it's just array, the type-hint can be added to the method and the cast to array can be removed.

src/Extend/Event.php
```diff
@@ -21,16 +21,18 @@ class Event implements ExtenderInterface
...
+     * @param string $event: Name of the event, can be the ::class attribute of the event class.
+     * @param callable|string $listener
+     *
      * The listener can either be:
-     *  - a callback function
-     *  - the class attribute of a class with a public `handle` method, which accepts an instance of the event as a parameter
-     *  - an array, where the first argument is an object or class name, and the second argument is the method on the
-     *    first argument that should be executed as the listener
+     *  - A callback function that accepts an instance of the event as a parameter.
+     *  - The class attribute of a class with a public `handle` method, which accepts an instance of the event as a parameter.
```
clarkwinkelmann, line 29:
```suggestion
     *  - The ::class attribute of a class with a public `handle` method, which accepts an instance of the event as a parameter.
```

src/Extend/Event.php
```diff
@@ -44,9 +46,10 @@ public function listen(string $event, $listener)
      *
      * @see https://laravel.com/docs/8.x/events#writing-event-subscribers
      *
-     * @param string $subscriber: The class attribute of the subscriber class
+     * @param string $subscriber: The class attribute of the subscriber class.
```
clarkwinkelmann, line 49:
```suggestion
     * @param string $subscriber: The ::class attribute of the subscriber class.
```

src/Extend/Middleware.php
```diff
@@ -21,40 +21,80 @@ class Middleware implements ExtenderInterface
...
         $this->addMiddlewares[] = $middleware;
 
         return $this;
     }
 
-    public function replace($originalMiddleware, $newMiddleware)
+    /**
+     * Replaces an existing middleware of the frontend.
+     *
+     * @param string $originalMiddleware: ::class attribute of the original middleware class.
```
clarkwinkelmann, line 49: Technically, it can also be a container binding name. That's probably true for multiple ::class parameters across all extenders, but even more so here because some middlewares are registered in the stack with their container name (like `flarum.api.route_resolver`)

I'm not sure if we want to mention this at all in the phpdoc here.

src/Extend/Routes.php
```diff
@@ -21,37 +21,160 @@ class Routes implements ExtenderInterface
     private $routes = [];
     private $removedRoutes = [];
 
+    /**
+     * @param string $appName: Name of the app (api, forum, admin).
+     */
     public function __construct($appName)
```
clarkwinkelmann, line 27: we can type-hint `string` in the constructor as well perhaps?

src/Extend/Routes.php
```diff
@@ -21,37 +21,160 @@ class Routes implements ExtenderInterface
...
-    public function get($path, $name, $handler)
+    /**
+     * Add a GET route.
+     *
+     * @param string $path: The path of the route
+     * @param string $name: The name of the route, must be unique.
+     * @param callable|string $handler: ::class attribute of the controller class, or a closure.
+     *
+     * If the handler is a controller class, it should implement \Psr\Http\Server\RequestHandlerInterface,
+     * or extend one of the Flarum Api controllers:
```
clarkwinkelmann, line 40: Is it really necessary to specify the list of all abstract controllers since they implement the required interface already?

That list might also get outdated in the future. It might be better if that's just explained in the documentation somewhere if it's not already the case?

src/Extend/ServiceProvider.php
```diff
@@ -18,11 +18,14 @@ class ServiceProvider implements ExtenderInterface
 
     /**
      * Register a service provider.
+     * There are no backwards compatibility guarantees for service providers.
```
clarkwinkelmann, line 21: I find our naming around service providers quite confusing.

It should be clear that the extender is fully part of the public API, it's what you might do with the service provider that might be outside of the public API. It's also my understanding that the `AbstractServiceProvider` class we provide with Flarum *is* part of the public API and won't break during the 1.x release cycle.

There are many legitimate use cases for service providers even for non-hacks, like registering singletons or custom container resolving for the extension's own class, or registering an external package into the container.

Maybe a warning in the form "Service providers are an advanced feature and might give access to Flarum internals that do not come with backward compatibility. Please read our documentation about service providers for recommendations" ?

src/Extend/View.php
```diff
@@ -30,11 +30,11 @@ class View implements ExtenderInterface, LifecycleInterface
      * You can also pass variables into a view: for more information, see https://laravel.com/api/8.x/Illuminate/View/Factory.html#method_make
      *
      * @param  string  $namespace: The name of the namespace.
-     * @param  string|array  $hints: This is a path (or an array of paths) to the folder(s)
+     * @param  string[]  $hints: This is a path (or an array of paths) to the folder(s)
```
clarkwinkelmann, line 33: the typing no longer matches the description of the parameter.

If it really has to be an array, then the text can be updated to remove the "or an array" portion.

Likewise, the `array` type-hint could be added to the method. Unless this is for undocumented backward-compatibility?

Just like my other comment above, it seems like this was intended to be a simple string as well. Maybe `string|string[]` is the expected type-hint?

clarkwinkelmann (commented):
Some things I noticed.

I also see that there are inconsistencies within the phpdoc. Sometimes there are columns before the parameter definition, sometimes there aren't (`string $example: Description` vs `string $example Description`).

I didn't even know having a column between the type and definition was allowed, but I see we already have that thorough the source code.

I don't know if you want to update the new doc blocks to all have the column?

src/Extend/ApiController.php
SychO9: ah yeah 🤦🏼‍♂️ not sure how I ended up doing that.

src/Extend/Middleware.php
SychO9: Good point, I think it's worth mentioning it for original middleware, but not for new, because technically extensions shouldn't be binding middleware classes, at least not in normal circumstances.

src/Extend/Routes.php
SychO9: It is too much, I think I'll only mention the `Flarum\Api\Controller` namespace, that's enough.

src/Extend/View.php
SychO9: Yup, I clearly was confused there.

src/Extend/Middleware.php
askvortsov1: > technically extensions shouldn't be binding middleware classes

If people are advanced enough to use custom bindings, they should understand the IoC container well enough that they realize binding strings can be used instead of class attrs.

src/Extend/ServiceProvider.php
askvortsov1: I like that wording!

src/Install/Steps/WriteSettings.php
```diff
@@ -71,7 +71,7 @@ private function getDefaults()
             'theme_dark_mode' => '0',
             'theme_primary_color' => '#4D698E',
             'theme_secondary_color' => '#4D698E',
-            'welcome_message' => 'This is beta software and you should not use it in production.',
+            'welcome_message' => 'Have a look around, and get used to your new forum!',
```
askvortsov1, line 74: I think we should link to Discuss here. 

Enjoy your new forum! Hop over to `https://discuss.flarum.org` if you have any questions, or to join our community!

Idk, not sure how to word this.

askvortsov1 (changes requested):

src/Extend/Auth.php
```diff
@@ -23,17 +23,22 @@ class Auth implements ExtenderInterface
...
-     *                                             Generally, `null` should be returned instead of `false` so that other
-     *                                             password checkers can run.
-     *                                   - `false` if the given password is invalid, and no other checkers should be considered.
-     *                                              Evaluation will be immediately halted if any checkers return `false`.
+     *
+     * The callable should accept:
+     * - $user: An instance of the User model.
+     * - $password: A string.
+     *
+     * The callable should return:
```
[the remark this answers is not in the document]
SychO9: Yea no I definitely want to scroll to see what's written and yeah I definitely have a large screen.

src/Install/Steps/WriteSettings.php
SychO9: I like that, I think that's simple enough and nice.

src/Extend/Auth.php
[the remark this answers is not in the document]
askvortsov1: Glad we agree!

askvortsov1 (approved):

src/Install/Steps/WriteSettings.php
clarkwinkelmann: But the link won't autolink, no?

clarkwinkelmann (approved):
Looks good to me.

I just added a comment in the discussion above about the link in the welcome message. I suppose it doesn't autolink, so I'm not sure if that's a good user experience? Maybe reference the community by name and not by link if the link cannot be clicked anyway.

src/Install/Steps/WriteSettings.php
SychO9: Ah, perhaps just `discuss.flarum.org` or `flarum.org` ?

Issue #2230, opened 2020-07-10
Future UTF-8 Support


tankerkiller125:
The following kinds of bugs regarding UTF-8 are known and require the following kinds of changes/steps:

#### Problems:
- UTF-8 Usernames are unsupported (https://github.com/flarum/core/issues/1949 and https://github.com/flarum/core/issues/1990)
    - Mentions currently can't match UTF-8
    - User slugs are currently hard coded to be the username

- Discussion slugs can not be UTF-8 and transliteration does not work for some languages/cultures (CJK) (https://github.com/flarum/core/issues/194)

- UTF-8 search is awful (https://github.com/flarum/core/issues/2003)

#### Possible solutions
- UTF-8 mentions: Convert current `<@{username}>` mention to `<@U:{UserID}>` similar to discord, also allows group mention in the future. Ensure that regex for detection properly supports UTF-8 characters ([special regex syntax is available for this](http://www.regular-expressions.info/unicode.html#category), also of note is that the `alpha_dash` validator supports UTF-8). Further we should always use the `display_name` column for rendering the mentions.

- UTF-8 user slugs: This one is harder but we should allow extensions/site owners to customize slugs and match to a database entry called `slugs` instead of usernames or display names slug column should never be null. This gives extensions complete control over slugs.

Once the username slugs and the mentions issue is resolved we can change the regex for the user validation logic to support UTF-8 in usernames/display names.

- UTF-8 discussion slugs: This one is also harder possibly, but it should probably follow the same exact proposal as the usernames with the exception that the discussion ID should always be prepended to whatever slug solution the extension dev or admin chooses (so slugs will always be `/{id}` at a minimum)

- UTF-8 search: Unfortunately this one is a limitation of the database engine itself often. The best solution around this is to support different search drivers that extension developers can build/tie into.

askvortsov1:
> UTF-8 mentions: Convert current <@{username}> mention to <@U:{UserID}> similar to discord, also allows group mention in the future

I really like this, but my only concern is that it wouldn't be understandable while in the text editor. Perhaps we could save a mapping of usernames <=> ids somewhere, still show the usernames in the text editor, and swap it out on save? Essentially, just cuz we store it that way shouldn't mean that we display it in the text editor that way.

> Unfortunately this one is a limitation of the database engine itself often. The best solution around this is to support different search drivers that extension developers can build/tie into.

I agree, this should resolve itself naturally with #1355 

[...]

PR #2870, opened 2021-05-16
Unicode Username Pull

SeRazon:
**Fixes #0000**

**Screenshot**

**Confirmed**


SeRazon:
文件命名错误

askvortsov1:
Hi! Thank you for the PR, but this is an issue where the decision has already been made. See https://github.com/flarum/core/issues/557 and linked issues for a lot of discussion. The tl;dr is that unicode usernames open a lot of rabbit holes leading to matters out of our control (like URL encoding) that could cause unexpected issues and user-side confusion. 

The solution introduced in https://github.com/flarum/core/issues/2230 is using https://github.com/flarum/nicknames, possibly in conjunction with an ID or nickname based model slug driver. Unfortunately, this means that we won't be merging a change like this at this time (although we might be open to smaller changes with discussion, like allowing periods ala flarum/framework#2529),

Going forward, if there isn't already an issue discussing an issue with proposed implementation, opening an issue to discussed things like this before proceeding to PR could help avoid confusion. Thank you again for the PR, sorry that we won't be able to make this change.

PR #2873, opened 2021-05-17
use display name for avatar color gen

### LINKED ISSUE #2690: Generate user colour from configured display driver
## Feature Request

User avatar colours are generated using the [username string](https://github.com/flarum/core/blob/a6dd545dbc5c173bf470fdd9ce5c89fa5603d51c/js/src/common/models/User.js#L50). On forums that use single sign on and do not rely on user input for the display name (e.g. may use numerical username user1, user2, user3) the colours are almost all the same for every user.

The function for generating the colour should instead use the nickname so that whichever display driver is set will be used to generate the colours.

I assume we could just substitute display name into the existing color function.

```javascript
  color: computed('displayName', 'avatarUrl', 'avatarColor', function (displayName, avatarUrl, avatarColor) {
    // If we've already calculated and cached the dominant color of the user's
    // avatar, then we can return that in RGB format. If we haven't, we'll want
    // to calculate it. Unless the user doesn't have an avatar, in which case
    // we generate a color from their display name.
    if (avatarColor) {
      return 'rgb(' + avatarColor.join(', ') + ')';
    } else if (avatarUrl) {
      this.calculateAvatarColor();
      return '';
    }

    return '#' + stringToColor(displayName);
  }),
```

The display driver is a core feature.

Workarounds are to get users to set an avatar so the colour gets generated from it but it cannot be forced.

ctml91:
**Fixes #2690**

Generate avatar colours based on display name, rather than username

**Confirmed**

**Required changes:**

None

### DIFF
--- a/js/src/common/models/User.js
+++ b/js/src/common/models/User.js
@@ -35,11 +35,11 @@ Object.assign(User.prototype, {
   canDelete: Model.attribute('canDelete'),
 
   avatarColor: null,
-  color: computed('username', 'avatarUrl', 'avatarColor', function (username, avatarUrl, avatarColor) {
+  color: computed('displayName', 'avatarUrl', 'avatarColor', function (displayName, avatarUrl, avatarColor) {
     // If we've already calculated and cached the dominant color of the user's
     // avatar, then we can return that in RGB format. If we haven't, we'll want
     // to calculate it. Unless the user doesn't have an avatar, in which case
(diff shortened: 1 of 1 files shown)


askvortsov1 (approved):
Great, thank you! We'll need to wait until the next release to merge this though, as we're currently in feature freeze.

tankerkiller125 (approved):

PR #2886, opened 2021-05-25
Mark typings as generated code

davwheat:

Marks the `dist-typings` folder as generated code.

This excludes it from diffs, and *might* exclude it from search, but I'm not too sure -- sources conflict on their info on this.

Nothing, really. Matches the existing syntax for the `.gitattributes` folder.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -12,5 +12,6 @@ tests export-ignore
 
 js/dist/* -diff
 js/dist/* linguist-generated
+js/dist-typings/* linguist-generated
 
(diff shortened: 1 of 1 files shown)


tankerkiller125 (approved):

datitisev:
`dist` doesn't have subfolders - will the same line work for `dist-typings`?

davwheat:
@datitisev The docs don't explicitly state either way, but based on the examples put forward in them, I would think so, yes.

datitisev (approved):

PR #2906, opened 2021-06-07
Fix image avatar alignment in notifications

### LINKED ISSUE #2905: Notification avatar alignment issue
https://discuss.flarum.org/d/27601-flarum-v10-or-above-notification-area-aliment-issues

davwheat:
**Fixes #2905**

Fixes issue introduced with #2822 where image-based avatars would display misaligned in the notifications list.

Are there any other issues with Notification avatars? Can anyone think of a less hacky way to achieve this?

**Screenshot**

### Before
image

### After
image

**Confirmed**

### DIFF
--- a/less/forum/NotificationList.less
+++ b/less/forum/NotificationList.less
@@ -136,6 +136,14 @@
     .Avatar--size(24px);
     grid-area: avatar;
   }
+  
+  // Since images don't have baselines, aligning against the baseline won't work.
+  // Instead we need to do some manual hackery to fix then, otherwise they won't
+  // be correctly vertically aligned.
+  img.Avatar {
+    align-self: flex-start;
+    margin-top: -2px;
+  }
 
   &-icon {
     font-size: 14px;


tankerkiller125 (approved):

SychO9 (commented):
Any reason for doing this instead of changing `  align-items: baseline;` to `  align-items: center;` in `.Notification` ?

davwheat:
Doing center will make the avatar centrally aligned on multi-line notifications, which doesn't match the style in previous versions, or the style of the text-based avatars.

SychO9:
> Doing center will make the avatar centrally aligned on multi-line notifications, which doesn't match the style in previous versions, or the style of the text-based avatars.

It doesn't though because of how the notification grid is structured.

davwheat:
If the title is multi-lined, it does.

image

SychO9:
> If the title is multi-lined, it does.
> 
> image

Ah, I see what you mean now, it's because the image itself is a cell of the grid that we can't just change its vertical alignment to middle. Another way of fixing this would be to wrap the notification avatar img with a div or span, then setting the vertical alignment of the `.Avatar` to middle.

davwheat:
I'd agree with something similar to that, but it could break 3rd party styles, annoyingly :(

SychO9 (approved):
True, we could do that for 2.0, but better avoid for 1.x

PR #2912, opened 2021-06-07
prevents missing route params

### LINKED ISSUE #2513: RouteCollection::getPath() throws Array to string conversion instead of proper error when parameter is missing
## Bug Report

**Current Behavior**
Calling `RouteCollection::getPath()` or `UrlGenerator::route()` with a parameterized route name but without specifying all parameters throws a *Array to string conversion* PHP notice and `Array` is inserted in the URL.

**Steps to Reproduce**

Use following `extend.php` code:

```php
<?php

use Flarum\Extend;

return [
    new Extend\Compat(function (\Illuminate\Contracts\Container\Container $container) {
        $container->afterResolving('flarum.forum.routes', function () {
            /**
             * @var $url UrlGenerator
             */
            $url = app(\Flarum\Http\UrlGenerator::class);

            var_dump($url->to('forum')->route('user', [ /* no parameter mapping */ ]));
            die('End of test case');
        });
    }),
];
```

And visit any page of the forum. See:

```
Notice: Array to string conversion in /home/clark/Projects/flarum-master/workbench/core/src/Http/RouteCollection.php on line 110
string(38) "http://master.flarum.localhost/u/Array" End of test case
```

**Environment**
- Flarum version: beta 15, dev-master, probably earlier

**Possible Solution**

We should either accept missing parameters and replace them with an empty string, or throw a proper error.

**Additional Context**
While it might not make much sense to call the method without specifying all parameters, there might still be situations where passing an arbitrary array makes sense.

My actual use case was different. I was calling `RouteCollection::getPath()` to find whether a particular route name is defined, since that seems to be the only way to find if a name has been registered in the router. So I was calling the method without any argument on purpose.

This is the `var_dump` of `$matchingParts` on line 110 when the parameter is not specified:

    array(2) { [0]=> string(3) "/u/" [1]=> array(2) { [0]=> string(8) "username" [1]=> string(5) "[^/]+" } }

luceos:
fixes #2513

**Confirmed**

### DIFF
--- a/tests/unit/Http/RouteCollectionTest.php
+++ b/tests/unit/Http/RouteCollectionTest.php
@@ -49,7 +49,7 @@ class RouteCollectionTest extends TestCase
     public function can_add_empty_parameters()
     {
         $this->expectException(\InvalidArgumentException::class);
-        $this->expectExceptionMessage("Route is missing argument for part user.");
+        $this->expectExceptionMessage('Route is missing argument for part user.');
 
         $routeCollection = (new RouteCollection)->addRoute('GET', '/user/{user}', 'user', function () {
             echo 'user';
--- a/tests/unit/Http/RouteCollectionTest.php
+++ b/tests/unit/Http/RouteCollectionTest.php
@@ -44,4 +44,17 @@ class RouteCollectionTest extends TestCase
 
         $this->assertEquals('/posts', $routeCollection->getPath('forum.posts.delete'));
     }
+
+    /** @test */
+    public function can_add_empty_parameters()
+    {
+        $this->expectException(\InvalidArgumentException::class);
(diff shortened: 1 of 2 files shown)


tests/unit/Http/RouteCollectionTest.php
```diff
@@ -44,4 +44,17 @@ public function can_add_routes_late()
 
         $this->assertEquals('/posts', $routeCollection->getPath('forum.posts.delete'));
     }
+
+    /** @test */
+    public function can_add_empty_parameters()
```
SychO9, line 49: I think it's more correct to say can't, since it properly errors out now for required parameters.

```suggestion
    public function must_provide_required_parameters()
```

SychO9 (commented):
Looks good

PR #2923, opened 2021-06-14
Fix path traversal vulnerability in registration avatar image URL downloader

luceos:
no description

### DIFF
--- a/src/User/Command/RegisterUserHandler.php
+++ b/src/User/Command/RegisterUserHandler.php
@@ -149,10 +149,10 @@ class RegisterUserHandler
             throw new InvalidArgumentException('Provided avatar URL must be a valid URI.', 503);
         };
 
-        $parsed_url = parse_url($url);
+        $scheme = parse_url($url, PHP_URL_SCHEME);
 
-        if (!($parsed_url['scheme'] === 'http' || $parsed_url['scheme'] === 'https')) {
-            throw new InvalidArgumentException('Provided avatar URL must have scheme http or https. Scheme provided was "' . $parsed_url['scheme'] . '".', 503);
+        if (! in_array($scheme, ['http', 'https'])) {
+            throw new InvalidArgumentException("Provided avatar URL must have scheme http or https. Scheme provided was $scheme.", 503);
         }
 
         $image = (new ImageManager)->make($url);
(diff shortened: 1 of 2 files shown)


src/User/Command/RegisterUserHandler.php
```diff
@@ -134,8 +136,25 @@ private function applyToken(User $user, RegistrationToken $token)
...
+
+        if ($urlValidator->fails()) {
+            throw new InvalidArgumentException('Provided avatar URL must be a valid URI.', 503);
+        };
+
+        $parsed_url = parse_url($url);
+
+        if (!($parsed_url['scheme'] === 'http' || $parsed_url['scheme'] === 'https')) {
+            throw new InvalidArgumentException('Provided avatar URL must have scheme http or https. Scheme provided was "' . $parsed_url['scheme'] . '".', 503);
+        }
```
luceos, line 156: What about this to make things slightly simpler?
```suggestion
        $scheme = parse_url($url, PHP_URL_SCHEME);

        if (! in_array($scheme, ['http', 'https'])) {
            throw new InvalidArgumentException("Provided avatar URL must have scheme http or https. Scheme provided was $scheme.", 503);
        }
```
Also I think this needs to throw a ValidationException with an array `[$attribute=>$reason]` so that the frontend can properly render it? @datitisev or @SychO9 please confirm?

luceos (commented):
I can't demand changes on my own PR, stupid GH :P

src/User/Command/RegisterUserHandler.php
davwheat: We discussed that but thought that displaying that would make users confused as it would seem as if they've done something wrong, when they have no control over that input.

luceos: Cool makes sense!

luceos:
Last thing to add is a test.

davwheat:
@luceos Tests have been added for each potential decision path, including multiple cases of test data which should all fail for different reasons.

Also included tests for _valid_ avatar URLs via the free [placeholder.com](https://placeholder.com/) API (PNG, JPG, and GIF).

tankerkiller125 (approved):
Everything looks good to me

luceos:
Merged in https://github.com/flarum/core/pull/2931

PR #2925, opened 2021-06-16
Lazy draw dropdowns to improve performance

### LINKED ISSUE #876: Lazy-draw some dropdown menus to improve performance
On pages with a lot of dropdown menus, like the Admin > Permissions page, redraws can get a bit sluggish because there are so many virtual DOM nodes to diff. We can improve this situation by only building the dropdown menu contents when a dropdown menu is open.

Note that we can't generalize this optimization to *all* dropdown menus because some of our responsive design relies on the dropdown menu contents being populated. But there are probably other places where we *can* do this too.

datitisev:
**Fixes #876**

- Add `lazyDraw` attribute to `Dropdown` component
- Make `PermissionGrid` dropdowns use lazy draw

- Do we like this implementation? Did we have something else in mind?
- Could this potentially be breaking if an extension has logic that occurs on the `items()` call of a Dropdown?

**Confirmed**

### DIFF
--- a/js/src/common/components/Dropdown.js
+++ b/js/src/common/components/Dropdown.js
@@ -37,11 +37,12 @@ export default class Dropdown extends Component {
 
   view(vnode) {
     const items = vnode.children ? listItems(vnode.children) : [];
+    const renderItems = this.attrs.lazyDraw ? this.showing : true;
 
     return (
       <div className={'ButtonGroup Dropdown dropdown ' + this.attrs.className + ' itemCount' + items.length + (this.showing ? ' open' : '')}>
         {this.getButton(vnode.children)}
-        {this.getMenu(items)}
+        {renderItems && this.getMenu(items)}
       </div>
     );
   }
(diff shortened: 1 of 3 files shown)


davwheat:
I assume the comment about breaking exts was meant for `getMenu`?

I think we should take that risk. `getMenu` was called whenever `view`  was called, and I can't see it being done anywhere else. If for some crazy reason someone extended `getMenu` with anything not related to just getting the menu, they should be able to move that logic to `view` instead.

davwheat (approved):

js/src/common/components/Dropdown.js
```diff
@@ -53,13 +54,25 @@ export default class Dropdown extends Component {
...
+      const { lazyDraw, onshow } = this.attrs;
+
       this.showing = true;
 
-      if (this.attrs.onshow) {
-        this.attrs.onshow();
+      // If using lazy drawing, redraw before calling `onshow` function
+      // to make sure the menu DOM exists in case the callback tries to use it.
+      if (lazyDraw) {
+        m.redraw.sync();
```
askvortsov1, line 64: Could we get around this complexity by attaching the onshow to the dropdown content vnode as an oncreate attr?

datitisev: Are you suggesting keeping the `if (!lazyDraw)` bellow, and simply move this one to the oncreate? 

The oncreate would then need to be conditional for `lazyDraw`, else it would be called on Dropdown render for non-lazy drawn dropdowns.

Am I making sense 🤔

askvortsov1: Not quite. What if we wrap `this.getMenu(items)` in a fragment or div/span, and attached the body of this callback to that fragment/div/span as an `oncreate`? Then, the onshow and redraw logic would only run whenever the element is opened, without relying on bootstrap, or having to account separately for lazy vs non-lazy (which is a potential point of failure)

askvortsov1 (changes requested):

js/src/common/components/Dropdown.js
datitisev: Unless I'm misunderstanding, wouldn't that change the behavior? With non-lazy, the component *will* be rendered, therefore the `oncreate` would run when the entire Dropdown is created.

datitisev: We're still technically changing behavior of a public component 🤷‍♂️

And we haven't defined public/private API anywhere for developers to know what might break and what shouldn't...

askvortsov1 (approved):

PR #2926, opened 2021-06-16
Hide post footer when empty

davwheat:

This PR changes the post footer to hide itself when it contains no items.

Currently, even if the footer has no items, it still exists and has margin, resulting in unexpected and weird white space at the bottom of posts.

In the screenshots below, notice the non-even spacing on the EventPost and on the whitespace below the post actions on the CommentPost.

I believe we previously had a more "hacky" way of doing this, by setting the footer bottom margin to 0 for Event Posts, but this fix will affect all posts instead of just Event Posts.

This **shouldn't** be breaking (unless extensions are adding content to footers in hacky ways which we should probably discourage anyway).

As long as extensions are using the ItemList to add items to the footer, this class shouldn't end up being added.

### Screenshot

**Before:**

image

**After:**

image

**Confirmed**

### DIFF
--- a/js/src/forum/components/Post.js
+++ b/js/src/forum/components/Post.js
@@ -72,7 +72,7 @@ export default class Post extends Component {
               )}
             </ul>
           </aside>
-          <footer className="Post-footer">{(footerItems.length && <ul>{listItems(footerItems)}</ul>) || null}</footer>
+          <footer className="Post-footer">{(footerItems.length ? <ul>{listItems(footerItems)}</ul>) : null}</footer>
         </div>
       </article>
     );
--- a/js/src/forum/components/Post.js
+++ b/js/src/forum/components/Post.js
@@ -44,6 +44,7 @@ export default class Post extends Component {
     attrs.className = this.classes(attrs.className).join(' ');
 
     const controls = PostControls.controls(this.attrs.post, this).toArray();
+    const footerItems = this.footerItems().toArray();
 
     return (
       <article {...attrs}>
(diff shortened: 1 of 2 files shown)


datitisev:
Why not just not draw the footer in the DOM, instead of hiding it with CSS?

davwheat:
I thought about that but assumed it'd be more breaking than just hiding.

If someone was doing weird stuff, they could override the style if needed to force it to show.

js/src/forum/components/Post.js
```diff
@@ -71,8 +72,8 @@ export default class Post extends Component {
               )}
             </ul>
           </aside>
-          <footer className="Post-footer">
-            <ul>{listItems(this.footerItems().toArray())}</ul>
+          <footer className={classList('Post-footer', !footerItems.length && 'Post-footer--empty')}>
+            <ul>{listItems(footerItems)}</ul>
```
SychO9, line 76: How about
```jsx
<footer className="Post-footer">
  {footerItems.length && (
    <ul>{listItems(footerItems)}</ul>
  ) || ''}
```

and then use the fancy `:empty` in CSS, so 
```CSS
&:empty {
  display: none;
}
```

:P too much nitpicking ? sorry..

davwheat: Oh I didn't think about that! That's a better idea! :P

Not sure how it'll handle a blank text node, if Mithril inserts one, as the empty pseudoselector doesn't match elements with any content (whitespace, even).

SychO9: >Not sure how it'll handle a blank text node, if Mithril inserts one, as the empty pseudoselector doesn't match elements with any content (whitespace, even).

true 🤔

davwheat:
@SychO9 Seems like your suggestion worked fine! Using `:empty` pseudoclass but rendering `null` if no items are in the footer (null renders as nothing, whereas `''` renders as a text node it seems).

SychO9 (approved):
nice!

tankerkiller125 (approved):

PR #2928, opened 2021-06-20
Convert extend util to TypeScript

datitisev:
- Use `.ts` extension when referring to `extend` util in core code
- Allow extensions to use `.ts` in import path through compat for typings to work properly
- Convert `extend` util to TypeScript

- Do we want to create a bunch more exported types/interfaces to make reading the code easier
- I don't know how to make it so the typings accept a nonexistent key for `extend`
- Are we fine with the new RegExp for the `compat` proxy to accept `.js`, `.ts` and `.tsx`

**Screenshot**
image

**Confirmed**

### DIFF
--- a/js/src/common/extend.ts
+++ b/js/src/common/extend.ts
@@ -1,10 +1,3 @@
-export type KeyOfType<T extends object, Type> = Exclude<
-  {
-    [K in keyof T]: T[K] extends Type ? K : never;
-  }[keyof T],
-  undefined
->;
-
 /**
  * Extend an object's method by running its output through a mutating callback
  * every time it is called.
--- a/js/src/common/extend.ts
+++ b/js/src/common/extend.ts
@@ -40,13 +40,13 @@ export function extend<T extends object, K extends KeyOfType<T, Function>>(
   allMethods.forEach((method: K) => {
     const original: Function | undefined = object[method];
 
-    object[method] = <T[K]>function (this: T, ...args: any[]) {
+    object[method] = function (this: T, ...args: Parameters<T[K]>) {
       const value = original ? original.apply(this, args) : undefined;
 
       callback.apply(this, [value, ...args]);
 
       return value;
-    };
+    } as T[K];
 
     Object.assign(object[method], original);
   });
@@ -89,9 +89,9 @@ export function override<T extends object, K extends KeyOfType<T, Function>>(
   allMethods.forEach((method) => {
     const original: Function = object[method];
 
-    object[method] = <T[K]>function (this: T, ...args: any[]) {
+    object[method] = function (this: T, ...args: Parameters<T[K]>) {
       return newMethod.apply(this, [original.bind(this), ...args]);
-    };
+    } as T[K];
 
     Object.assign(object[method], original);
   });
--- a/js/src/common/extend.ts
+++ b/js/src/common/extend.ts
@@ -36,9 +36,9 @@ export function extend<T extends object, K extends KeyOfType<T, Function>>(objec
   const allMethods = Array.isArray(methods) ? methods : [methods];
 
   allMethods.forEach((method: K) => {
-    const original: Function|undefined = object[method];
+    const original: Function | undefined = object[method];
 
-    object[method] = <T[K]> function (this: T, ...args: any[]) {
+    object[method] = <T[K]>function (this: T, ...args: any[]) {
       const value = original ? original.apply(this, args) : undefined;
 
       callback.apply(this, [value, ...args]);
@@ -77,13 +77,17 @@ export function extend<T extends object, K extends KeyOfType<T, Function>>(objec
  * @param methods The name or names of the method(s) to override
  * @param newMethod The method to replace it with
  */
-export function override<T extends object, K extends KeyOfType<T, Function>> (object: T, methods: K|K[], newMethod: (orig: T[K], ...args: any) => void) {
+export function override<T extends object, K extends KeyOfType<T, Function>>(
+  object: T,
+  methods: K | K[],
+  newMethod: (orig: T[K], ...args: any) => void
+) {
   const allMethods = Array.isArray(methods) ? methods : [methods];
 
   allMethods.forEach((method) => {
     const original: Function = object[method];
 
-    object[method] = <T[K]> function (this: T, ...args: any[]) {
+    object[method] = <T[K]>function (this: T, ...args: any[]) {
       return newMethod.apply(this, [original.bind(this), ...args]);
     };
 
--- a/js/src/common/extend.ts
+++ b/js/src/common/extend.ts
@@ -5,8 +5,6 @@ export type KeyOfType<T extends object, Type> = Exclude<
   undefined
 >;
 
-export type ExtendCallback<T extends (...args: any) => any> = (val: ReturnType<T>, ...args: any[]) => void;
-
 /**
  * Extend an object's method by running its output through a mutating callback
  * every time it is called.
@@ -32,7 +30,11 @@ export type ExtendCallback<T extends (...args: any) => any> = (val: ReturnType<T
  * @param methods The name or names of the method(s) to extend
  * @param callback A callback which mutates the method's output
  */
-export function extend<T extends object, K extends KeyOfType<T, Function>>(object: T, methods: K | K[], callback: ExtendCallback<T[K]>) {
+export function extend<T extends object, K extends KeyOfType<T, Function>>(
+  object: T,
+  methods: K | K[],
+  callback: (val: ReturnType<T[K]>, ...args: any[]) => void
+) {
   const allMethods = Array.isArray(methods) ? methods : [methods];
 
   allMethods.forEach((method: K) => {
--- a/js/src/common/extend.ts
+++ b/js/src/common/extend.ts
@@ -33,7 +33,7 @@ export type KeyOfType<T extends object, Type> = Exclude<
 export function extend<T extends object, K extends KeyOfType<T, Function>>(
   object: T,
   methods: K | K[],
-  callback: (val: ReturnType<T[K]>, ...args: Parameters<T[K]>) => void
+  callback: (this: T, val: ReturnType<T[K]>, ...args: Parameters<T[K]>) => void
 ) {
   const allMethods = Array.isArray(methods) ? methods : [methods];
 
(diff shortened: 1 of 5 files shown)


davwheat:
Out of curiosity, what's the reasoning behind stating the file extension? Is it just to fix typings so that they work properly?

datitisev:
@davwheat Yes. We currently have an `extend.js` and an `extend/index.js`, and so no file extension resolves to `extend/index.js`.

datitisev:
Currently fixing issues with my proxifyCompat changes, as they break imports of e.g. tags, and others

js/src/common/extend.ts
```diff
@@ -19,23 +26,27 @@
...
+  object: T,
+  methods: K | K[],
+  callback: (val: ReturnType<T[K]>, ...args: Parameters<T[K]>) => void
+) {
   const allMethods = Array.isArray(methods) ? methods : [methods];
 
-  allMethods.forEach((method) => {
-    const original = object[method];
+  allMethods.forEach((method: K) => {
+    const original: Function | undefined = object[method];
```
davwheat, line 41: Can we not replace `Function | undefined` with `T[K]`? `Function` is like `any`, so it's best not to use it if we can.

Same for `override`.

datitisev: `.apply` doesn't exist on `T[K]`, for whatever reason

datitisev:
what happened here

davwheat:
Rebased only master

davwheat (approved):

js/src/common/extend.ts
```diff
@@ -1,3 +1,10 @@
+export type KeyOfType<T extends object, Type> = Exclude<
```
askvortsov1, line 1: This is a bit hard to follow. Am I correct in understanding that:

1. We take all keys of T (each locally K)
2. We map each K => K (if T[K] = T)
3. We take those values with `[keyof T]`
4. We filter out everything that doesn't with Exclude

Could we make this a bit easier to digest by assembling this type operator step by step? The variable names for each intermediate step would then serve as documentation.

js/src/common/extend.ts
```diff
@@ -1,3 +1,10 @@
...
+  {
+    [K in keyof T]: T[K] extends Type ? K : never;
+  }[keyof T],
+  undefined
```
askvortsov1, line 5: Why `undefined` and not `never`? Aren't we already filtering for nonexistent keys through the type map?

js/src/common/extend.ts
```diff
@@ -19,23 +26,27 @@
...
- * @param {function} callback A callback which mutates the method's output
+ * @param object The object that owns the method
+ * @param methods The name or names of the method(s) to extend
+ * @param callback A callback which mutates the method's output
  */
-export function extend(object, methods, callback) {
+export function extend<T extends object, K extends KeyOfType<T, Function>>(
+  object: T,
+  methods: K | K[],
+  callback: (this: T, val: ReturnType<T[K]>, ...args: Parameters<T[K]>) => void
```
askvortsov1, line 36: Typescript recognizes that `this` is `this`? Cool!

On a side note, I really wish PHP had callback typing...

js/src/common/extend.ts
askvortsov1: > `.apply` doesn't exist on `T[K]`, for whatever reason

Aren't we enforcing that `T[K] extends Function` though? Not a blocker, just curious :)

js/src/common/utils/proxifyCompat.ts
```diff
@@ -1,10 +1,12 @@
 export default (compat: { [key: string]: any }, namespace: string) => {
   // regex to replace common/ and NAMESPACE/ for core & core extensions
+  // and remove .js, .ts and .tsx extensions
   // e.g. admin/utils/extract --> utils/extract
   // e.g. tags/common/utils/sortTags --> tags/utils/sortTags
-  const regex = new RegExp(`(\\w+\\/)?(${namespace}|common)\\/`);
+  const regex = new RegExp(`^(.+\/)?(${namespace}|common)\/`);
```
askvortsov1, line 6: Why change the part of the regex before the namespace?

askvortsov1 (changes requested):

js/src/common/Application.js
```diff
@@ -12,7 +12,7 @@ import mapRoutes from './utils/mapRoutes';
 import RequestError from './utils/RequestError';
 import ScrollListener from './utils/ScrollListener';
 import liveHumanTimes from './utils/liveHumanTimes';
-import { extend } from './extend';
+import { extend } from './extend.ts';
```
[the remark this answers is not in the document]
davwheat: We have a folder called `extend` and apparently this can mess with typings in some cases.

I desperately think we should rename one of them in 2.0.

js/src/common/extend.ts
```diff
@@ -1,3 +1,10 @@
+export type KeyOfType<T extends object, Type> = Exclude<
+  {
+    [K in keyof T]: T[K] extends Type ? K : never;
+  }[keyof T],
+  undefined
+>;
```
davwheat, line 6: @askvortsov1 I think this is clearer.

```suggestion
type KeyOfType<Type extends object, Match> = { [Key in keyof Type]-?: Type[Key] extends Match ? Key : never }[keyof Type];
```

1. `= { ... }` -  Create a new object type...
2. `[K in keyof Type]` - ...with all keys of `Type`
3. `-?:` -  ...and make them all required (convert `string | undefined` to `string`)
4. `Type[Key] extends Match ?` ...and set each key's type conditionally based on whether the type of the key is a subtype of the `Match` type
5. `Key : never` - ...so that the key's type becomes the key itself if its one we care about (`Key`), or it gets dropped if we don't care about it (`never`)
6. `[keyof Type]` - Then we convert the map of key to key into an array of the keys

js/src/common/extend.ts
davwheat: I think this is the old equivalent of the `-?:` in my type.

js/src/common/extend.ts
davwheat: You'd think so, but `T[K] extends Function` actually ends up being true when `T[K]` is a function, **and** when it is a subtype of `Function`.

This is something I ran into when playing with deep cloning objects: https://github.com/microsoft/TypeScript/issues/45492

davwheat (changes requested):

js/src/common/extend.ts
askvortsov1: Dropping the `Exclude` feels cleaner, but I still think we should break this up into steps so it's more parseable.

js/src/common/extend.ts
datitisev: Yep! You can type this

js/src/common/extend.ts
datitisev: i have no idea how that code works

js/src/common/utils/proxifyCompat.ts
datitisev: i don't remember

davwheat:
> Do we want to create a bunch more exported types

I'd say that for something like this (utility types), we should declare them in the `@types` folder we now have. That lets us use them across core without imports, and extensions can use them too (from 1.0.5 now we have the fix for global typings).

js/src/common/extend.ts
davwheat: How, though? The only way we can split it is into KeysOfType and KeyOfType. Is that fine?

js/src/common/utils/proxifyCompat.ts
davwheat: I think we should revert that. If we don't remember what it does, and the old code works fine, there's no need to change it.

askvortsov1 (approved):
LGTM, let's run prettier and revert the regex change, then good to go!

SychO9:
>  (from 1.0.5 now we have the fix for global typings).

1.1 actually, since it was merged into master and not `1.0.5` unless you want to cherry pick it.

I am not well informed to know if it's something that *needs* to be in 1.0.5 and not wait 1.1

davwheat:
Oops. I'd say we should stick it in 1.0.5 if we can. It's just a developer improvement and waiting a whole release cycle to provide it seems pointless.

askvortsov1:
> Oops. I'd say we should stick it in 1.0.5 if we can. It's just a developer improvement and waiting a whole release cycle to provide it seems pointless.

I would prefer to keep patch releases focused on bugfixes tbh.

davwheat (commented):
I'm going to dismiss my review as I've contributed a fair bit to this PR.

askvortsov1:
> I'm going to dismiss my review as I've contributed a fair bit to this PR.

I don't believe that's necessary. That being said, @datitisev could you confirm that you approve of the changes?

datitisev:
@askvortsov1 I... think? I mean now I understand the `KeyOfType` types even less but I didn't understand them before so that doesn't matter. The issue of using `override` with non-declared properties still exists, not sure if we wanted to fix that or not. Doesn't really matter as we allow building with TS errors.

image

askvortsov1:
> The issue of using `override` with non-declared properties still exists

In those cases, wouldn't we want to modify the prototype directly?

datitisev:
@askvortsov1 Not if it has the potential to exist but its optional? Cause then you can take advantage of the override... I think my main example was `extend(options, 'config')` or something in `app.request`.

askvortsov1:
> @askvortsov1 Not if it has the potential to exist but its optional? Cause then you can take advantage of the override... I think my main example was `extend(options, 'config')` or something in `app.request`.

Wdym by "take advantage of the override"? Are there any advantages compared to directly editing the prototype?

datitisev:
@askvortsov1 Oh I'm stupid, I'm thinking about extend() and not override()

In `app.request`
```js
    extend(options, 'config', (result, xhr) => xhr.setRequestHeader('X-CSRF-Token', this.session.csrfToken));
```

Here it would be useful because we account for it not existing. My apologies. 

image

davwheat:
I think if we allow ` | string`, we still benefit from the Intellisense of the `KeysOfType<>` type? Don't quote me on that tho...

davwheat:
Honestly, I didn't even know that was a feature.

In the past I've used:

```ts
function myMethod() { }

if (Class.prototype.method) {
  extend(Class.prototype, 'method', myMethod);
} else {
  Class.prototype.method = myMethod;
}
```

I know it's not the most beautiful syntax, but I don't realistically see any issue with it.

datitisev:
> Maybe we should add a separate util for that then? A guaruntee that the extended/overriden method exists is pretty useful generally.

This'd be breaking then tho. And I honestly don't see a problem with the current implementation 🤷‍♂️

askvortsov1:
> > Maybe we should add a separate util for that then? A guaruntee that the extended/overriden method exists is pretty useful generally.
> 
> This'd be breaking then tho. And I honestly don't see a problem with the current implementation 🤷‍♂️ 

Not really, as we'd be changing the type interface, not the implementation.

Its not that there's a problem with the current non-restrictive approach as much as we're losing out on the ability to enforce something statically through the type system: in particular, because defining a function as a composition of other functions isn't quite the same as extending an existing function.

PR #2929, opened 2021-06-20
Convert some common classes/utils to TS

datitisev:
- Convert `Application`, `Session`, `utils/extractText`
- Update `states/AlertManagerState` ~~and `utils/RequestError`~~

- Properly typing `extract` is close to impossible because we override it with a different type. We use it and just pass `xhr.responseText`, but Mithril's `extract` option takes `xhr` and `options`.
    <details>
    <summary>PhpStorm screenshots</summary>

    image
    image

    </details>
- Do we want the Application types to be exported at the top or a different file?

**Screenshot**

**Confirmed**


davwheat:
Rebased onto master. I've dropped Application's conversion in favour of more recent progress on #3006. I didn't realise you did it in here, but the new PR is more likely to be up-to-date with new changes, and has support for new ItemList typings from #3005.

js/src/common/Session.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  constructor(user: User | null, csrfToken: string | null) {
+    this.user = user;
+    this.csrfToken = csrfToken;
+  }
+
+  /**
+   * Attempt to log in a user.
+   */
+  public login(body: LoginParams, options: Omit<RequestOptions, 'url'> = {}) {
```
davwheat, line 39: These options are provided as part of the function body and other params, so we should omit them too.

```suggestion
  public login(body: LoginParams, options: Omit<RequestOptions, 'url' | 'method' | 'body'> = {}) {
```

js/src/common/Session.ts
```diff
@@ -0,0 +1,58 @@
...
+    return app.request(
+      Object.assign(
+        {
+          method: 'POST',
+          url: `${app.forum.attribute('baseUrl')}/login`,
+          body,
+        },
+        options
+      )
```
davwheat, line 48: This is more in-line with our new TS, I think

```suggestion
      {
        method: 'POST',
        url: `${app.forum.attribute('baseUrl')}/login`,
        body,
        ...options,
      },
```

js/src/common/Session.ts
```diff
@@ -0,0 +1,58 @@
...
+   * Password
+   */
+  password: string;
+};
+
+export default class Session {
+  /**
+   * The current authenticated user.
+   */
+  public user: User | null;
```
davwheat, line 24: We don't seem to explicitly state `public` in most of our other TS files, but I'm not inherently opposed to it.

I'd like to hear others' opinions on it.

js/src/common/utils/extractText.ts
```diff
@@ -1,15 +1,14 @@
+import Mithril from 'mithril';
```
davwheat, line 1:
```suggestion
import type Mithril from 'mithril';
```

davwheat (changes requested):

js/src/common/states/AlertManagerState.ts
```diff
@@ -55,7 +55,7 @@ export default class AlertManagerState {
   /**
    * Dismiss an alert.
    */
-  dismiss(key: AlertIdentifier): void {
+  dismiss(key: AlertIdentifier | null): void {
```
askvortsov1, line 58: What's the motivation for this change? I see that we check for a falsy key below (maybe that should also be changed), but is there a use case for `null` to be provided as a key to dismiss?

askvortsov1 (changes requested):

js/src/common/utils/extractText.ts
```diff
@@ -1,15 +1,14 @@
...
- * @param {VirtualElement} vdom
- * @return {String}
  */
-export default function extractText(vdom) {
+export default function extractText(vdom: Mithril.Children): string {
   if (vdom instanceof Array) {
     return vdom.map((element) => extractText(element)).join('');
   } else if (typeof vdom === 'object' && vdom !== null) {
-    return vdom.children ? extractText(vdom.children) : vdom.text;
+    return vdom.children ? extractText(vdom.children) : (vdom.text as string);
```
askvortsov1, line 10: Since text is `string | number | boolean`, maybe we should explicitly cast instead of asserting? If we're telling our type system that something is a string but it actually isn't, that could lead to trouble.

js/src/common/utils/extractText.ts
```diff
@@ -1,15 +1,14 @@
...
   } else {
-    return vdom;
+    return vdom as string;
```
askvortsov1, line 12: Same here. At this point, `vdom` is `string | number | boolean | null | undefined`. An explicit match would be safer, otherwise there's not much benefit from this typing.

js/src/common/states/AlertManagerState.ts
datitisev: Not necessarily a use case, but if we check for null/falsy then we should probably show that we are handling said type... then exts don't need to do the check themselves in the code (even though it's barely any extra code). 🤷‍♂️

js/src/common/Session.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  /**
+   * The CSRF token.
```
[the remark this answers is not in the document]
datitisev: I didn't add it. This was in the original session file.

js/src/common/utils/extractText.ts
datitisev: Do we want to cast it or have `extractText` return `string | number | boolean | null | undefined`, then?

Casting would technically be change in behavior, but I don't know what the use case of that behavior would be.

js/src/common/Session.ts
```diff
@@ -0,0 +1,58 @@
...
+   */
+  public csrfToken: string | null;
+
+  constructor(user: User | null, csrfToken: string | null) {
```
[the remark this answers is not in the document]
datitisev: I... don't know. Do you want me to remove the nullable type?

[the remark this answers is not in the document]
askvortsov1: I think we should set it to sensible values. Unless there's a valid use case for null token, I don't think we should allow nullables.

js/src/common/states/AlertManagerState.ts
askvortsov1: Fair enough

js/src/common/utils/RequestError.ts
```diff
@@ -1,21 +1,21 @@
 export default class RequestError {
-  status: string;
+  status: string | number;
```
askvortsov1, line 2: Shouldn't these always be `number`? I know we currently use strings, but I'm not sure that's actually correct.

js/src/common/utils/extractText.ts
askvortsov1: `extractText` implies that it returns text. I think casting it would be best.

askvortsov1 (changes requested):

askvortsov1:
Before review, this should be rebased, and tested against the current TypeScript environment to ensure that it doesn't introduce any new bugs.

datitisev:
@askvortsov1 I rebased this PR a week ago, that's why I re-requested the reviews.

askvortsov1:
> @askvortsov1 I rebased this PR a week ago, that's why I re-requested the reviews.

Right, but the Typecheck and type coverage GH actions aren't showing up, so those must have been merged after you rebased.

askvortsov1 (approved):

js/src/common/utils/extractText.ts
```diff
@@ -1,15 +1,14 @@
...
- * @param {VirtualElement} vdom
- * @return {String}
  */
-export default function extractText(vdom) {
+export default function extractText(vdom: Mithril.Children): string {
   if (vdom instanceof Array) {
     return vdom.map((element) => extractText(element)).join('');
   } else if (typeof vdom === 'object' && vdom !== null) {
-    return vdom.children ? extractText(vdom.children) : vdom.text;
+    return vdom.children ? extractText(vdom.children) : String(vdom.text);
```
davwheat, line 10:
```suggestion
    return vdom.children ? extractText(vdom.children) : `${vdom.text}`;
```

js/src/common/utils/extractText.ts
```diff
@@ -1,15 +1,14 @@
...
   } else {
-    return vdom;
+    return String(vdom);
```
davwheat, line 12:
```suggestion
    return `${vdom}`;
```

davwheat (approved):
Looks good to me! 

Not sure if these suggestions might be better as Babel compiles them to a smaller bundle size. Up to you.

datitisev:
@davwheat I'd rather keep it as is, I feel like `String()` is clearer.

PR #2946, opened 2021-06-27
Use laravel validator to replace avatar validation error params

### LINKED ISSUE #2933: Alert variables not being replaced
## Bug Report

**Current Behavior**
When attempting to add an invalid file as the user's avatar, the variables in the alert box don't seem to be getting replaced.

**Steps to Reproduce**
1. Go to your **Profile**
2. Click on the **avatar** (and click **Upload** if one already exists).
3. Select a file that is either **too large** or the **wrong file type**, such as a PSD.
4. Note text that appears in the alert box: 
    > The :attribute must not be greater than :max kilobytes.

**Expected Behavior**
The alert box should say something like:
> The avatar must not be greater than 1024 kilobytes.

**Screenshots**
Screenshot 2021-06-22 at 13 14 29@2x

**Environment**
- Flarum version: 1.0.2
- Website URL: https://ancient-badlands-jogrg5kqhk.ploi.link/
- Webserver: NGINX
- Hosting environment: VPS
- PHP version: 7.4.20
- Browser: Chrome 91

```
Flarum core 1.0.2
PHP version: 7.4.20
Loaded extensions: Core, date, libxml, openssl, pcre, zlib, filter, hash, pcntl, Reflection, SPL, session, standard, sodium, mysqlnd, PDO, xml, bcmath, bz2, calendar, ctype, curl, dom, mbstring, FFI, fileinfo, ftp, gd, gettext, gmp, iconv, igbinary, imagick, imap, intl, json, exif, msgpack, mysqli, pdo_mysql, pdo_pgsql, pdo_sqlite, pgsql, Phar, posix, readline, redis, shmop, SimpleXML, soap, sockets, sqlite3, sysvmsg, sysvsem, sysvshm, tokenizer, xmlreader, xmlwriter, xsl, zip, memcached, Zend OPcache
| flarum-flags         | v1.0.0  |        |
| flarum-nicknames     | v1.0.0  |        |
| flarum-approval      | v1.0.0  |        |
| flarum-bbcode        | v1.0.0  |        |
| flarum-emoji         | v1.0.0  |        |
| flarum-lang-english  | v1.0.0  |        |
| flarum-likes         | v1.0.0  |        |
| flarum-lock          | v1.0.0  |        |
| flarum-markdown      | v1.0.1  |        |
| flarum-mentions      | v1.0.0  |        |
| flarum-statistics    | v1.0.0  |        |
| flarum-sticky        | v1.0.0  |        |
| flarum-subscriptions | v1.0.0  |        |
| flarum-suspend       | v1.0.0  |        |
Base URL: https://ancient-badlands-jogrg5kqhk.ploi.link
Installation path: /home/ploi/flarum.pswd.biz
Debug mode: off
```

This is a fresh Flarum install.

SychO9:
**Fixes #2933**

When we switched to intl icu message format, the translation parameters 
have become required to be in the format `{param}`, so we can no longer 
use the translator here.

This PR switches to consistently rely on the laravel validator to make 
the replacements, just like with all other validation errors in the rest 
of the application.

Is this the best way to go about this ?

**Confirmed**

### DIFF
--- a/src/User/AvatarValidator.php
+++ b/src/User/AvatarValidator.php
@@ -73,7 +73,10 @@ class AvatarValidator extends AbstractValidator
     protected function raise($error, array $parameters = [], $rule = null)
     {
         $message = $this->laravelValidator->makeReplacements(
-            $this->translator->trans("validation.$error"), 'avatar', $rule ?? $error, array_values($parameters)
+            $this->translator->trans("validation.$error"),
+            'avatar',
+            $rule ?? $error,
+            array_values($parameters)
         );
 
         throw new ValidationException(['avatar' => $message]);
--- a/src/User/AvatarValidator.php
+++ b/src/User/AvatarValidator.php
@@ -17,12 +17,19 @@ use Symfony\Component\Mime\MimeTypes;
 class AvatarValidator extends AbstractValidator
 {
     /**
+     * @var \Illuminate\Validation\Validator
+     */
+    protected $laravelValidator;
+
+    /**
      * Throw an exception if a model is not valid.
      *
      * @param array $attributes
      */
     public function assertValid(array $attributes)
     {
+        $this->laravelValidator = $this->makeValidator($attributes);
+
         $this->assertFileRequired($attributes['avatar']);
         $this->assertFileMimes($attributes['avatar']);
         $this->assertFileSize($attributes['avatar']);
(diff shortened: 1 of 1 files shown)


luceos:
Why don't we change the format on the validation.yml strings instead?

SychO9:
The strings in `validation.yml` are not specific to the avatar validation which uses the translator, those same validation strings are/will be used by extension directly without custom handling like in the avatar validator, in which case the laravel validator will be responsible for rendering the validation error messages, and it requires the laravel format of `:param`.

Hence why I opted to just rely on the laravel validator as well, just like we do by default, because we're now in a situation where the translator requires `{param}` and the laravel validator `:param`.

luceos:
I don't know, it still feels like a lot of fuss to get something working. Can't we override the translator instance in the validator to use the ICU one? It's a long shot, but it would result in much cleaner translation files. Now we have to explain why one file behaves differently than the rest.

SychO9:
> I don't know, it still feels like a lot of fuss to get something working. Can't we override the translator instance in the validator to use the ICU one? It's a long shot, but it would result in much cleaner translation files. Now we have to explain why one file behaves differently than the rest.

The laravel validator already uses the flarum translator
https://github.com/flarum/core/blob/e92c267cdec46266b633f71c2f41040731cdaf39/src/Locale/LocaleServiceProvider.php#L56

But it doesn't use the translator to replace the parameters, it uses it just to get the raw message string, then it makes the replacements on its own, hence the `makeReplacements` method on the validator.

We could also just do manual replacement here instead of relying on the laravel validator for that, I dislike recreating existing logic, but since we're only using three validation strings here, they would all be straightforward replacements. That would be more simple and less of a hustle, and fine for this particular spot, the avatar validator is the only one that does this custom validation anyway.

PR #2947, opened 2021-06-28
Allow easier extensibility of page document

luceos:
This allows extensions to:

- mutate the json api document sent with php documents
- override/interact with the Document

Right now extensions need to replace the complete Frontend class
in order to interact with the document. Let's abstract into ioc
so that advanced devs can interact with it.

**Confirmed**

**Required changes:**

### DIFF
--- a/src/Frontend/Frontend.php
+++ b/src/Frontend/Frontend.php
@@ -49,7 +49,9 @@ class Frontend
     {
         $forumDocument = $this->getForumDocument($request);
 
-        $document = new Document($this->view, $forumDocument, $request);
+        $responseDocument = resolve('flarum.frontend.document');
+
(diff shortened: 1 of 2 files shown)


SychO9:
Isn't everything in the document easily editable through the Frontend content extender except for the title ? and in which case wouldn't it be easier and more efficient to pre-make the title so that it can be edited in the aforementioned extender ?

luceos:
@SychO9 that's a fair point, I don't remember why exactly I made this PR and binding all of it. But to me this offers far more flexibility than just binding a title.

SychO9:
That's true, but I think that making changes based on use cases will allow us to provide better solutions, also to avoid binding to the container when possibly not necessary.

Though we can comeback to this for 1.1

PR #2952, opened 2021-07-03
Sorted

Justman10000:
Is that clearer

### DIFF
--- a/locale/core.yml
+++ b/locale/core.yml
@@ -221,6 +221,7 @@ core:
 
     # These translations are used for the users list on the admin dashboard.
     users:
+      title: => core.ref.users    
       description: A paginated list of all users on your forum.
 
       grid:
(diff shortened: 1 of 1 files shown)


davwheat:
We always keep our translations sorted alphabetically to make finding them easier. What's this attempting to fix? Could you provide a description?

Justman10000:
> 
> 
> We always keep our translations sorted alphabetically to make finding them easier. What's this attempting to fix? Could you provide a description?

Oh well ... Well, it belongs together!

SychO9:
They are together but alphabetically sorted.

tankerkiller125 (changes requested):
We sort translations by alphabetical order. I don't think a change is required here.

SychO9:
Thank you for the PR @Justman10000! but we alphabetically sort translation keys so this change is unnecessary. We do appreciate contributions however so don't be discouraged 👍🏼

PR #2954, opened 2021-07-04
Move Day.js plugin types import to global typings

### LINKED ISSUE #2417: Day.js plugins and Typescript
_Originally posted at #2391_

Even not using directly, we need to import relativeTime dayjs plugin at `js/src/common/utils/humanTime.ts`. Without this import, typescript can't pick up tipes for the [fromNow()](https://github.com/flarum/core/blob/2ff46f5a2ebe2532f780f6e60a921beb41876d69/js/src/common/utils/humanTime.ts#L32) method and complains.
The code still works though, because relativeTime plugin is installed at `js/src/common/index.js`:

https://github.com/flarum/core/blob/2ff46f5a2ebe2532f780f6e60a921beb41876d69/js/src/common/index.js#L12-L16

We can avoid this import if we move this plugin installation process to a separated dayjs library module, something like:
```typescript
// js/src/common/lib/dayjs.ts

import dayjs from 'dayjs'

dayjs.extend(relativeTime)

export default dayjs


// js/src/common/utils/humanTime.ts

// no need import relativeTime from 'dayjs/plugin/relativeTime';
import dayjs from '../../common/lib/dayjs'
```

lhsazevedo:
**Fixes #2417**

Remove Day.js relativeTime plugin types import from humanTime util and add it to the global typings file.
See @davwheat's [proposal](https://github.com/flarum/core/issues/2417#issuecomment-866151040) on #2417.

Ensuring that no error occurs when compiling the humanTime util.
```
$ ./node_modules/.bin/tsc --esModuleInterop ./@types/global/index.d.ts ./src/common/utils/humanTime.ts
```
Note that you'll need to pass the `--esModuleInterop` to avoid the following error (unrelated to this PR):
```
node_modules/dayjs/index.d.ts:3:1
    3 export = dayjs;
      ~~~~~~~~~~~~~~~
    This module is declared with using 'export =', and can only be used with a default import when using the 'esModuleInterop' flag.
```
This flag is [highly recommended](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html#support-for-import-d-from-cjs-from-commonjs-modules-with---esmoduleinterop) by the TypeScript team for both new and existing projects. I didn't add it to `tsconfig.json` because I don't understand ts enough right now to guarantee that it won't break anything :)

### DIFF
--- a/js/@types/global/index.d.ts
+++ b/js/@types/global/index.d.ts
@@ -3,6 +3,7 @@ import Mithril from 'mithril';
 
 // Other third-party libs
 import * as _dayjs from 'dayjs';
+import 'dayjs/plugin/relativeTime';
 import * as _$ from 'jquery';
 
 // Globals from flarum/core
--- a/js/src/common/utils/humanTime.ts
+++ b/js/src/common/utils/humanTime.ts
@@ -1,5 +1,4 @@
 import dayjs from 'dayjs';
-import 'dayjs/plugin/relativeTime';
 
 /**
  * The `humanTime` utility converts a date to a localized, human-readable time-


davwheat:
> This flag is [highly recommended](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html#support-for-import-d-from-cjs-from-commonjs-modules-with---esmoduleinterop) by the TypeScript team for both new and existing projects. I didn't add it to `tsconfig.json` because I don't understand ts enough right now to guarantee that it won't break anything :)

Great spot. I'll update the tsconfig repo now.

davwheat (approved):

davwheat:
Looks good to me.

Running `./node_modules/.bin/tsc.cmd --esModuleInterop ./src/common/utils/humanTime.ts` seems to work. (Specifying the typings file in `global` isn't needed because it's included by the tsconfig. No errors when doing so.

lhsazevedo:
Great, thanks for reviewing!

datitisev (approved):
Not tested locally, changes make sense though.

PR #2957, opened 2021-07-09
Allow document title to be edited through extender

SychO9:
**Related to #2947**

Of all the document fields that can be manipulated through the Frontend\content extender, the title is the only one that can't be changed. This PR fixes that by premaking it and allowing content callbacks to then manipulate that value.

Code, any better way to go about doing this ?

**Confirmed**

### DIFF
--- a/src/Frontend/Document.php
+++ b/src/Frontend/Document.php
@@ -123,6 +123,11 @@ class Document implements Renderable
     public $css = [];
 
     /**
+     * @var array
+     */
+    public $premadeData = [];
+
+    /**
      * @var Factory
      */
     protected $view;
@@ -159,6 +164,16 @@ class Document implements Renderable
     }
 
     /**
+     * Prepares the document to be populated.
+     *
+     * @return void
+     */
+    public function prepare()
+    {
+        $this->premadeData['title'] = $this->makeTitle();
+    }
+
+    /**
      * @return View
      */
     protected function makeView(): View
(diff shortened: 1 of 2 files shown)


src/Frontend/Document.php
```diff
@@ -180,6 +195,10 @@ protected function makeView(): View
      */
     protected function makeTitle(): string
     {
+        if (isset($this->premadeData['title'])) {
+            return $this->premadeData['title'];
+        }
+
         $onHomePage = rtrim($this->request->getUri()->getPath(), '/') === '';
 
         return ($this->title && ! $onHomePage ? $this->title.' - ' : '').Arr::get($this->forumApiDocument, 'data.attributes.title');
```
luceos, line 204: Where did this code block go?

SychO9: it still executes a first time when called from the `prepare` method

luceos: On mobile those lines were marked as removed ..

src/Frontend/Document.php
```diff
@@ -158,6 +163,16 @@ public function render(): string
...
     }
 
+    /**
+     * Prepares the document to be populated.
+     *
+     * @return void
+     */
+    public function prepare()
+    {
+        $this->premadeData['title'] = $this->makeTitle();
```
luceos, line 173: I don't like this. 

```php
        $this->premadeData['title'] = $this->makeTitle();
```

`makeTitle` assigns the value of the array itself. Why can't we do something like:

```suggestion
        $this->premadeData['title'] = $this->premadeData['title'] ?? $this->makeTitle();
```

And then remove this in makeTitle:

```
```php
        if (isset($this->premadeData['title'])) {
            return $this->premadeData['title'];
        }

```

luceos (changes requested):

PR #2959, opened 2021-07-12
Allow extra attrs provided to `<Select>` to be passed through to the DOM element

davwheat:
- Extra attrspassed to the Select component are given straight to the DOM element
  - This is **critical** for accessibility. Currently, there is no way to provide a11y attrs (such as `for`, `id`, `aria-described-by`, etc.) to the DOM element.
- ~~Custom class names can be provided to the outer wrapping element for easier customisability~~
- Custom attrs can also be passed to the wrapper via the `wrapperAttrs` attr

~~Should we allow a way to pass any attrs wanted to the outer element, too? Something like  a `wrapperAttrs` attribute?~~ Done!

**Confirmed**

### DIFF
--- a/js/src/common/components/Select.js
+++ b/js/src/common/components/Select.js
@@ -11,16 +11,30 @@ import classList from '../utils/classList';
  * - `onchange` A callback to run when the selected value is changed.
  * - `value` The value of the selected option.
  * - `disabled` Disabled state for the input.
- * - `outerClassName` class given to the containing Select element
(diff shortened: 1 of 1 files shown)


datitisev:
> Should we allow a way to pass any attrs wanted to the outer element, too? Something like a wrapperAttrs attribute?

This sounds like a good idea.

davwheat:
> This sounds like a good idea.

@datitisev Implemented!

davwheat:
I'll fix formatting in a second.

tankerkiller125 (approved):

PR #2961, opened 2021-07-13
Improve selected text stylization

the-turk:

1. It was using inline affix if you have block prefix but don't have block suffix.
This makes impossible to use suffix-less block syntaxes (e.g. spoiler).

2. It makes sense to use block affix (if there are any) for multiline selection.

3. It makes sense to surround multiline selection with new lines if you have block prefix but don't have block suffix. 
I needed to set `surroundWithNewlines` argument to `true` to achieve this behaviour on the block spoiler. But this also affected the inline spoiler and that's undesired. So I assumed that If you don't have any suffix and try to style multiple lines at once, then you probably will want to surround selection with new lines.

**Confirmed**

**Required for:**

- [x] Related core extension PRs: (https://github.com/flarum/markdown/pull/30)


askvortsov1 (approved):
Tested locally, LGTM

tankerkiller125 (approved):

Issue #2942, opened 2021-06-24
Replying / Linking Event Post leads to JS Error


GreXXL:
## Bug Report

**Current Behavior**
When mentioning an event post on preview there is a JS-Error in console

**Steps to Reproduce**
1. Go to https://discuss.flarum.org/d/27858-affiliate-marketing-flarum-extension or https://discuss.flarum.org/d/27878-test
2. Hover over the post mention
3. See error

**Expected Behavior**
No JS Error

**Screenshots**
https://user-images.githubusercontent.com/1630413/123234820-9359e300-d4db-11eb-9123-683229a5abe0.mov

**Environment**
- Flarum version: 1.0.3
- Website URL: https://discuss.flarum.org

**Additional Context**
```
Uncaught (in promise) TypeError: Cannot read property 'substring' of undefined
    at i (forum.js?v=c14f5145:1)
    at r (forum.js?v=c14f5145:1)
    at e.view (forum.js?v=c14f5145:45)
    at Function.a (forum.js?v=c14f5145:22)
    at forum.js?v=c14f5145:22
    at forum.js?v=c14f5145:22
    at l (forum.js?v=c14f5145:22)
    at u (forum.js?v=c14f5145:22)
    at forum.js?v=c14f5145:22
    at l (forum.js?v=c14f5145:22)
```

SychO9:
Here's the error when in dev mode:
```
string.ts:5 Uncaught (in promise) TypeError: Cannot read property 'substring' of undefined
    at i (string.ts:5)
    at r (highlight.tsx:27)
    at e.view (PostPreview.js:19)
    at Function.a (render.js:30)
    at render.js:155
    at render.js:160
    at l (render.js:66)
    at u (render.js:50)
    at render.js:134
    at l (render.js:63)
```

Besides fixing the JS error, it would be a good idea to improve the preview for an event post mention, even though people aren't supposed to reply to event posts, they still can by manually writing the post mention, as long as they have the post's ID.

PR #2964, opened 2021-07-14
Don't show excerpt in PostPreview component if there are no plain content

the-turk:
**Fixes https://github.com/flarum/core/issues/2942**

`span.PostPreview-excerpt` is still necessary for proper view.

Quoting from @SychO9:
> Besides fixing the JS error, it would be a good idea to improve the preview for an event post mention, even though people aren't supposed to reply to event posts, they still can by manually writing the post mention, as long as they have the post's ID.

Since we don't store any information about event posts in db, I think this kind of enhancement will need some breaking changes.

**Confirmed**

### DIFF
--- a/js/src/forum/components/PostPreview.js
+++ b/js/src/forum/components/PostPreview.js
@@ -16,7 +16,8 @@ export default class PostPreview extends Component {
   view() {
     const post = this.attrs.post;
     const user = post.user();
-    const excerpt = highlight(post.contentPlain(), this.attrs.highlight, 300);
+    const content = post.contentType() === 'comment' && post.contentPlain();
+    const excerpt = content ? highlight(content, this.attrs.highlight, 300) : '';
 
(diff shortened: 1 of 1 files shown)


SychO9 (approved):
Thank you for the PR!

This is enough, I don't think there is any value in it containing info about the event post, since such post mentions shouldn't happen anyway.

> we don't store any information about event posts in db

We do btw, the posts themselves have info on the event, thus event posts.

the-turk:
> We do btw, the posts themselves have info on the event, thus event posts.

I mean, yeah, but we don't have the information on how to turn that into readable text like " x locked the discussion". All info we have is it's an event post which has 'locked' type and nothing more.

davwheat (approved):
Coincidence! I was just taking a look at the issue for this PR :P

Looks good to me and appears to work locally 👍

PR #2970, opened 2021-07-20
Fix Select-based setting breaking admin pages

davwheat:
**Fixes bug caused by #2959**

Prevents some attrs being passed to the `<Select>` component when they should not be.

For example, `type: "select"` creates a Select component, but this `type` attr was also passed to the component, resulting in Mithril attempting to perform weird changes on elements which are then disallowed by the DOM engine, causing the page to error.

Are there other areas broken by the Select change? Extension setting pages seem to be fine when using these changes.

**Screenshot**

image

**Confirmed**

### DIFF
--- a/js/src/admin/components/AdminPage.js
+++ b/js/src/admin/components/AdminPage.js
@@ -98,39 +98,37 @@ export default class AdminPage extends Page {
       return entry.call(this);
     }
 
-    const { setting, help, ...componentAttrs } = entry;
+    const { setting, help, type, label, ...componentAttrs } = entry;
 
-    const value = this.setting([setting])();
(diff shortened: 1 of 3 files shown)


SychO9 (commented):
Looks good, though there is still another issue with the Select component changes, the select in Mail page for example is lacking the default className,

flarum lan_admin (27)

This is the relevant code, it seems like the provided additional className overrides the default one, while before the changes to Select component, it ignored it, idk why that class is provided, it doesn't make sense to have there, I'd remove it entirely (from both inputs).

https://github.com/flarum/core/blob/4bcfc5078c37e1adc9c9b62ce9791f0733de515c/js/src/admin/components/MailPage.js#L60-L66

SychO9 (approved):

PR #2971, opened 2021-07-20
Adds missing settings defaults for display name driver and User slug driver

davwheat:

Adds missing default setting values for Display Name Driver and User Slug Driver. This will prevent empty select boxes for these settings on new Flarum installs.

Should we add a migration to input these defaults on existing forums where they haven't already been set?

**Confirmed**


davwheat:
Rebased

SychO9 (approved):

askvortsov1 (approved):
Fine for now, but in the long term, I would prefer to get rid of DB backed defsult settings entirely in favor of an extender.

PR #2984, opened 2021-07-21
Rewrite Button to Typescript

davwheat:
**Progresses flarum/framework#3533, flarum/framework#3364**

### Changes proposed in this pull request

**Button:**
- Prefer `aria-label` over `title`
- Don't duplicate button content to `title` attribute
  - This has no effect on a11y. Content of nodes is preferred over `title` attributes by screen readers.
- Warn in debug mode if button has no accessible content
- Use modern JS/TS syntax (`||=`, spread, etc)

**Others:**
- Add debug warning helper: fires `console.warn`, but only when the forum is in debug mode. Can help to inform extension developers of possible issues with their JS code.
- Update TextEditorButton to work with changes that prefer `aria-label`

### Reviewers should focus on

Is anything else affected by the preferring of `aria-label` as opposed to `title`? Most likely components that deeply interact with Button and those which extend it.

### Confirmed

### Required changes


js/src/common/components/Button.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+    className = classList(_class, className, {
+      hasIcon: iconName,
+      disabled: disabled || loading,
+      loading: loading,
+    });
+
+    if (!ariaLabel && !extractText(vnode.children) && !this.element?.getAttribute?.('aria-label')) {
+      fireDebugWarning(
+        '[Flarum Accessibility Warning] This button has no content but does not have any accessible label. This means that screen-readers will not be able to interpret its meaning. Consider providing accessible text via the `aria-label` attribute.\n\nLearn more: https://web.dev/button-name',
```
tankerkiller125, line 96:
```suggestion
        '[Flarum Accessibility Warning] This button has no content and does not have any accessible label. This means that screen-readers will not be able to interpret its meaning. Consider providing accessible text via the `aria-label` attribute.\n\nLearn more: https://web.dev/button-name',
```

tankerkiller125 (changes requested):
I think the warning message requires a very slight re-wording so that it isn't confusing, otherwise it looks good to me.

datitisev:
The warning will be logged every time the button redraws... I think it'd be good to only do it once per element, but that'd probably be outside the scope of this PR.

davwheat:
I couldn't think of a better way to handle that, to be honest, without risking the checks not working correctly. Things like `this.element` aren't defined in `oninit` as far as I'm aware, meaning we can't check if the Tooltip component has set an `aria-label` attribute.

tankerkiller125 (approved):

datitisev:
> I couldn't think of a better way to handle that, to be honest, without risking the checks not working correctly. Things like this.element aren't defined in oninit as far as I'm aware, meaning we can't check if the Tooltip component has set an aria-label attribute.

It's defined in Component.oncreate(), I believe. Warning in `oncreate` should do the trick.

js/src/common/components/Button.tsx
```diff
@@ -0,0 +1,124 @@
...
+      delete attrs.onclick;
+    }
+
+    className = classList(_class, className, {
+      hasIcon: iconName,
+      disabled: disabled || loading,
+      loading: loading,
+    });
+
+    if (!ariaLabel && !extractText(vnode.children) && !this.element?.getAttribute?.('aria-label')) {
```
datitisev, line 94: Shouldn't the `aria-label` attribute of element be the same as `ariaLabel`? Why do you need to check both?

davwheat: ariaLabel is provided directly to the Button component, but we also need to see if `aria-label` is present on the DOM element as that's how the Tooltip component provides accessible labels.

datitisev: shouldn't it be present because aria label is passed to the <button> attributes?

davwheat: It's assigned to the actual DOM node's attributes, but the relationship between the DOM and VDOM in Mithril is one-way. If we update the actual DOM, the component's attributes don't change, as far as I'm aware.

https://github.com/flarum/core/blob/ce824b0ccff6a6570e2254e207e0d17fbbdd1b1c/js/src/common/components/Tooltip.tsx#L241

js/src/common/components/Button.tsx
```diff
@@ -0,0 +1,130 @@
...
+ * The `Button` component defines an element which, when clicked, performs an
+ * action.
+ *
+ * Other attrs will be assigned as attributes on the `<button>` element.
+ *
+ * Note that a Button has no default class names. This is because a Button can
+ * be used to represent any generic clickable control, like a menu item. Common
+ * styles can be applied by providing `className="Button"` to the Button component.
+ */
+export default class Button extends Component<IButtonAttrs> {
```
askvortsov1, line 69: Shouldn't Button's attrs be extensible?

js/src/common/helpers/fireDebugWarning.ts
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * This function is intended to provide warnings to extension developers about issues with
+ * their extensions that may not be easily noticed when testing, such as accessibility
+ * issues.
+ *
+ * These warnings should be hidden on production forums to ensure webmasters are not
+ * inundated with do-gooders telling them they have an issue when it isn't something they
+ * can fix.
+ */
+export default function fireDebugWarning(...args: any[]): void {
```
askvortsov1, line 12: Excellent, this is a great idea!

askvortsov1 (approved):

js/src/common/components/Button.tsx
```diff
@@ -66,7 +66,7 @@ export interface IButtonAttrs extends ComponentAttrs {
  * be used to represent any generic clickable control, like a menu item. Common
  * styles can be applied by providing `className="Button"` to the Button component.
  */
-export default class Button extends Component<IButtonAttrs> {
+export default class Button<CustomAttrs extends Record<string, unknown> = {}> extends Component<IButtonAttrs & CustomAttrs> {
```
askvortsov1, line 69: Why can't we extend ComponentAttrs

davwheat: We can, there's just no point in doing so. Including `ComponentAttrs` is done with the `IButtonAttrs` interface.

datitisev: Can't the same thing be achieved by doing `CustomAttrs extends IButtonAttrs`? I think that's what is being done in other components?

js/src/common/components/Button.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+    return <button {...buttonAttrs}>{this.getButtonContent(vnode.children)}</button>;
+  }
+
+  oncreate(vnode: Mithril.VnodeDOM<IButtonAttrs, this>) {
+    super.oncreate(vnode);
+
+    const { 'aria-label': ariaLabel } = this.attrs;
+
+    if (!ariaLabel && !extractText(vnode.children) && !this.element?.getAttribute?.('aria-label')) {
```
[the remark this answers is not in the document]
davwheat: It should cover that because of the `aria-label` attribute check on the element itself

js/src/common/components/Button.tsx
davwheat: It can be done that way, but that runs the risk of depending on the component extending it to correctly extend the `IButtonAttrs` interface. I'd prefer to enforce IButtonAttrs as being present.

davwheat: Nope. Screw that. Changed my mind. Totally agree with you.
```suggestion
export default class Button<CustomAttrs extends IButtonAttrs = IButtonAttrs> extends Component<CustomAttrs> {
```

js/src/common/components/Button.tsx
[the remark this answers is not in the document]
datitisev: Your comment just gave me deja vu, my bad.

PR #2986, opened 2021-07-22
Revert "Fix intervention/image version to 2.5.0"

SychO9:
intervention/image have pushed a fix and release 2.6.1, so this is no longer necessary.

https://github.com/Intervention/image/commit/e2b8aebe3e5baf1e3db61a0f5227502313d74b29

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -59,7 +59,7 @@
         "illuminate/support": "^8.0",
         "illuminate/validation": "^8.0",
         "illuminate/view": "^8.0",
-        "intervention/image": "^2.6.1 || 2.5.*",
+        "intervention/image": "2.5.* || ^2.6.1",
         "laminas/laminas-diactoros": "^2.4.1",
         "laminas/laminas-httphandlerrunner": "^1.2.0",
         "laminas/laminas-stratigility": "^3.2.2",
--- a/composer.json
+++ b/composer.json
@@ -59,7 +59,7 @@
         "illuminate/support": "^8.0",
         "illuminate/validation": "^8.0",
         "illuminate/view": "^8.0",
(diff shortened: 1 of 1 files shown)


davwheat (approved):

davwheat (changes requested):
Shouldn't we force 2.6.1 or 2.5.x in our composer JSON then with something like...


```
"intervention/image": "^2.6.1 || 2.5.*",
```

SychO9:
> Shouldn't we force 2.6.1 or 2.5.x in our composer JSON then with something like...

Not really, if composer picks up `2.6.x` as the wanted version, it'll directly pick `2.6.1`, latest patch

davwheat:
> Not really, if composer picks up 2.6.x as the wanted version, it'll directly pick 2.6.1, latest patch

I think it's better safe than sorry. The suggested version range allows for a broken version of `intervention/image`, which we want to avoid.

If another package required `2.6.0` for some reason, this would break the entire forum unnecessarily instead of composer telling us that it won't work.

If we can tell composer to not pick 2.6.0 under any circumstances, like we can, I think that's the better option.

composer.json
```diff
@@ -59,7 +59,7 @@
         "illuminate/support": "^8.0",
         "illuminate/validation": "^8.0",
         "illuminate/view": "^8.0",
-        "intervention/image": "2.5.*",
+        "intervention/image": "^2.5.0",
```
SychO9, line 62:
```suggestion
        "intervention/image": "^2.6.1 || 2.5.*",
```

SychO9:
> If another package required it at 2.6.0 or 2.5.* || 2.6.0 for some reason 

That would be a pretty bad package :p, sure why not.

davwheat (approved):

PR #2991, opened 2021-07-30
flarum info command additions

luceos:
This PR adds a couple of new details to the info command:

- MySQL version, so that we can spot mysql issues faster
- Queue driver, so that we can understand issues where email isn't sent when the queue worker isn't running
- Mail driver, so that we can understand issues where email isn't sent when set to log, null or something else

I've also added a newline above the debug warning and I've made debug string "ON" red.

image

### DIFF
--- a/src/Foundation/Console/InfoCommand.php
+++ b/src/Foundation/Console/InfoCommand.php
@@ -16,8 +16,8 @@ use Flarum\Foundation\Config;
 use Flarum\Settings\SettingsRepositoryInterface;
 use Illuminate\Contracts\Queue\Queue;
 use Illuminate\Database\ConnectionInterface;
-use Illuminate\Database\MySqlConnection;
 use Illuminate\Support\Str;
+use PDO;
 use Symfony\Component\Console\Helper\Table;
 use Symfony\Component\Console\Helper\TableStyle;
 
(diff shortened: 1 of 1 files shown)


tankerkiller125 (approved):

src/Foundation/Console/InfoCommand.php
```diff
@@ -126,4 +137,28 @@ private function findPackageVersion($path, $fallback = null)
...
+        // Drop everything like queue SyncQueue, RedisQueue
+        $queue = str_replace('queue', null, $queue);
+
+        return $queue;
+    }
+
+    private function identifyDatabaseVersion(): string
+    {
+        /** @var MySqlConnection $connection */
+        $connection = resolve(ConnectionInterface::class);
```
SychO9, line 160: why aren't we injecting this instead just like the settings repository ?

luceos: Yeah I wasn't sure about that either. I think either we inject all of them or none. The drawback of injecting them in the construct is that they will be resolved when the command is registered, not just when executed. So best would be to delay resolving as long as possible. I guess after Settings I switched that strategy.

SychO9: > The drawback of injecting them in the construct is that they will be resolved when the command is registered

The commands are only resolved after the entire application has booted, right before symfony console application takes over, so I don't believe there will be any issues with injecting them.
https://github.com/flarum/core/blob/4acff91f8063fcced9bf8c9a76fbb510d06823c0/src/Console/Server.php#L30-L38

luceos: Thanks 😍 great feedback!

SychO9 (approved):

PR #2996, opened 2021-07-31
Rewrite AdminPage.js into Typescript

davwheat:
**Progresses flarum/framework#3533**

- Rewrites the `AdminPage` component into Typescript
- Add missing a11y attributes using random IDs
- Add `Page` component attrs interface to be used in `AdminPage`

**Confirmed**

**Required changes:**


js/src/admin/components/AdminPage.tsx
```diff
@@ -0,0 +1,290 @@
...
+ */
+export type SettingsComponentOptions = HTMLInputSettingsComponentOptions | SwitchSettingComponentOptions | SelectSettingComponentOptions;
+
+/**
+ * Valid attrs that can be returned by the `headerInfo` function
+ */
+export type AdminHeaderAttrs = AdminHeaderOptions & Partial<Omit<Mithril.Attributes, 'class'>>;
+
+export default class AdminPage extends Page {
+  settings: Record<string, () => string> = {};
```
SychO9, line 94: Can we import the mithril `Stream` type instead and use that ? because `() => string` isn't accurate.

```ts
import type StreamType from 'mithril/stream';

export default class AdminPage extends Page {
  settings: Record<string, StreamType<string>> = {};
```

SychO9 (changes requested):

js/src/admin/components/AdminPage.tsx
davwheat: Oops! Done! 👍

js/src/admin/components/AdminPage.tsx
```diff
@@ -0,0 +1,290 @@
...
+  onsaved(): void {
+    this.loading = false;
+
+    app.alerts.show({ type: 'success' }, app.translator.trans('core.admin.settings.saved_message'));
+  }
+
+  /**
+   * Returns a function that fetches the setting from the `app` global.
+   */
+  setting(key: string, fallback: string = ''): () => string {
```
SychO9, line 248: :P

```suggestion
  setting(key: string, fallback: string = ''): Stream<string> {
```

SychO9 (changes requested):

davwheat:
Oops... Missed that one 🙈. Thanks! :)

SychO9 (approved):
Looks Good!

js/src/admin/components/AdminPage.tsx
```diff
@@ -0,0 +1,290 @@
...
+   * Returns a function that fetches the setting from the `app` global.
+   */
+  setting(key: string, fallback: string = ''): Stream<string> {
+    this.settings[key] = this.settings[key] || Stream<string>(app.data.settings[key] || fallback);
+
+    return this.settings[key];
+  }
+
+  /**
+   * Returns a list of settings that have been modified, but not yet saved.
```
askvortsov1, line 255: Not a list

askvortsov1 (changes requested):

js/src/admin/components/AdminPage.tsx
```diff
@@ -0,0 +1,290 @@
...
+   *    default: 'option1',
+   * }
+   *
+   * @example
+   *
+   * () => {
+   *   return <p>My cool component</p>;
+   * }
+   */
+  buildSettingComponent(entry: ((this: typeof this) => Mithril.Children) | SettingsComponentOptions): Mithril.Children {
```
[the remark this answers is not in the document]
SychO9: I love the idea!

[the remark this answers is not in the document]
davwheat: I like the idea too. Maybe make an issue for it? Refactoring out functionality to another area of core is outside the scope of this PR.

js/src/admin/components/AdminPage.tsx
```diff
@@ -0,0 +1,290 @@
...
+        {this.header(vnode)}
+        <div className="container">{this.content(vnode)}</div>
+      </div>
+    );
+  }
+
+  /**
+   * Returns the content of the AdminPage.
+   */
+  content(vnode: Mithril.Vnode<Record<string, unknown>, this>): Mithril.Children {
```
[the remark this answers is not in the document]
davwheat: We can't do that without making the whole class abstract. Not sure if that could affect any functionality in some way.

js/src/@types/global.d.ts
```diff
@@ -1,3 +1,31 @@
+/**
+ * A type that matches any valid value for the `type` attribute on an
+ * HTML `<input>` element.
+ */
+declare type HTMLInputTypes =
```
askvortsov1, line 5: Why?

js/src/admin/components/AdminPage.tsx
```diff
@@ -0,0 +1,264 @@
...
+   * Will be used as the class for the AdminPage.
+   *
+   * Will also be appended with `-header` and set as the class for the `AdminHeader` component.
+   */
+  className: string;
+}
+
+interface CommonSettingsItemOptions extends Mithril.Attributes {
+  setting: string;
+  label: string | ReturnType<typeof app.translator.trans>;
```
datitisev, line 27: Not sure how I feel about explicitly stating that it can be output of trans method... I mean that's not wrong but I don't think that's how we should refer to it 🤔 

Label also can be any mithril child... I think we should allow for that (whatever the type is specifically, I always forget)
Same with help perhaps?

js/src/@types/global.d.ts
davwheat: I considered it to be more of a general type as opposed to being specific to that component. There's a good chance we might want to use it elsewhere (or another Flarum ext might). It's mainly for convenience.

askvortsov1: Maybe we should move it into the [generic form builder](https://github.com/flarum/core/pull/2996#discussion_r691807155) when we have that? I'm hesitant to put stuff in global that doesn't absolutely need to be there.

js/src/admin/components/AdminPage.tsx
```diff
@@ -40,18 +40,21 @@ export interface HTMLInputSettingsComponentOptions extends CommonSettingsItemOpt
   type: HTMLInputTypes;
 }
 
+const BooleanSettingTypes = ['bool', 'checkbox', 'switch', 'boolean'] as const;
```
askvortsov1, line 43: nit: is there no option for something like "const readonly string[]"?

askvortsov1 (approved):

js/src/admin/components/AdminPage.tsx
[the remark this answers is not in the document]
askvortsov1: I think we should do that, it is an abstract class after all

js/src/admin/components/AdminPage.tsx
davwheat: That removes the entire point of `as const`.

davwheat:
Made a few changes. Will merge soon™️ if noone objects.

PR #2997, opened 2021-07-31
Add ability to set email confirmation state to false

davwheat:

Allows admins to set email confirmation state to `false`.

Should we instead add a `deactivate` method to `User.php` and fire a deactivation event, similar to what we do for activation?

**Confirmed**

### DIFF
--- a/src/User/Command/ConfirmEmailHandler.php
+++ b/src/User/Command/ConfirmEmailHandler.php
@@ -41,10 +41,11 @@ class ConfirmEmailHandler
         /** @var EmailToken $token */
         $token = EmailToken::validOrFail($command->token);
 
+        /** @var \Flarum\User\User */
         $user = $token->user;
         $user->changeEmail($token->email);
 
-        $user->activate();
+        $user->confirmEmail();
 
         $user->save();
         $this->dispatchEventsFor($user);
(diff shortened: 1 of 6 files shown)


luceos:
I see that User::activate fires an event, would it make sense to have a similar event (and perhaps method on User) for deactivation?

Other than that this PR looks fine.

davwheat:
Implemented both. We now have `User::deactivate` and a `Deactivated` event.

askvortsov1:
I am not convinced of the use case for this. As mentioned over chat, I would be in favor of a general activation / deactivation mechanism where email confirmation is one factor, but until then, what's the use case for marking an email as not confirmed?

luceos:
I was reading through the changes and got confused by the fact that activation and deactivation relies on a column about email verification. To that extend I agree, for now we might keep this logic and extend or modify it later? Imo this logic almost warrants it's own database column `activated_at`?

davwheat:
I think we should add this for now, and then change the activation logic later.

askvortsov1:
Very well, I think we can do this, but could we rename the methods to `confirmEmail` and `unconfirmEmail` (and deprecate `activate`)? Then the API would be consistent with the behavior we want to work towards in the future.

askvortsov1 (changes requested):
Looking over this again, I would be happy to approve this if we changed the method names as described above.

askvortsov1:

Issue #2600, opened 2021-02-15
CSS Variables


SychO9:
This isn't a proposal to use this feature -*yet*- but writing down my thoughts on the improvements it can bring and the challenges that come with using it.

------

This feature has been around for quiet some time now, not when most of flarum's frontend design was built though. (https://caniuse.com/css-variables)

## Pros
It's very flexible, dark mode for instance would be a simple addition of a class instead of recompiling the CSS.

Using local scoped variables can also solve a lot of problems, clean the code and make it easier to customize. 

Instead of using `color` and `background` as inline styles for the Tag hero and labels to use the custom color, we should just define a CSS variable inline and let the main CSS code use that, this makes it possible to customize these elements in more ways a lot easier.

For example if I wanted to create a theme where the Tag hero's color is the text color instead, I would have to add custom JS code to get that custom value first, but not if it's define as a CSS variable:

```jsx
// Before
<header className={'Hero TagHero' + (color ? ' TagHero--colored' : '')}
        style={color ? {color: '#fff', backgroundColor: color} : ''}>
// After
<header className={'Hero TagHero' + (color ? ' TagHero--colored' : '')}
        style={color ? {"--hero-bg": color} : ''}>
```

```less
// Before
.Hero {
  background: @hero-bg;
  color: @hero-color;
}
// After
.Hero {
  background: var(--hero-bg);
  color: var(--hero-color);
}
```

## Cons
In our codebase, we use some LESS color functions such as `fade`,`lighten`,`darken`,`mix` ..etc to create color variations. We would have to do something about those if we want to be able to properly make use of this feature and make things like darkmode a simple class change, otherwise while we could still only use it minimally in certain places to improve the code, there wouldn't be much value in using it.

There two approaches I can think of, the second shouldn't even be an option, but I'll mention it nonetheless.

**Declaring the variations as variables**
Go with a simple approach, which is to define these color variations as CSS variables first, and then make use of them in their respective places. For instance hovering over a primary button shows a darker primary color, in our current code that's `darken(fadein(@primary-color, 5%), 5%)`, we would in that case define (for example) a `--color-primary-dark` variable with that value and use that.

The problem is with LESS mixins, such as `Button--color(@color, @background)`, it creates variations of the background color, which is more dynamic than color variations created in normal places. We would have to find a compromise between declaring the variation variable and using the mixin.
* [x] Investigate mixins using color variations and combining that with css variables.

**Dynamically use variations with `hsl` by breaking down the colors**
[...]

PR #3001, opened 2021-08-06
[1.x] Custom Colorising with CSS Custom Properties

### LINKED ISSUE #2816: Improve custom colorizing by using CSS variables
Flarum has a lot of custom colorization, here are some places that come to mind:

* [x] Tag Labels.
* [x] Tag Heroes.
* [x] Profile Heroes.
* [x] SideNav Tag Icons.
* [x] Tag Tiles.
* [x] Start Discussion Button.
* [x] Badges.
* [x] Default Avatars.

The challenge with customizing such elements is that changing how colors are presented is very difficult (unless you do some monkey patching with JS, which can be tedious).

For example, the tag hero has an inline style of
```css
color: #fff;
background-color: lime;
```

If I wanted to switch up the colors or maybe add a border with the tag color, that wouldn't be possible (not in a clean way at least) and I would also have to use `!important` 😱.

We should instead declare inline custom CSS variables and use those, allowing something like this:

```css
color: var(--tag-color);
background: #fff;
```

SychO9:
**Fixes #2816**
**Progresses #2600**

This PR introduces usage of CSS variables instead if custom colors passed as inline styles, especially as a `background` property. The PR is not meant to convert all of the codebase to use CSS custom properties, that will take place in another PR(s), where the root CSS variables will also be properly organised.

The changes to the sideNav links and button colors are done for the tags extension's custom colorising.

This should be 100% backwards compatible and should not break any custom styling in existing communities.

General code quality and anything I missed.

**Confirmed**

**Required changes:**

- [x] Related core extension PRs:
  * **flarum/tags#139**
  * **flarum/sticky#29**
  * **flarum/lock#29**
  * **flarum/subscriptions#42**

### DIFF
--- a/js/src/forum/components/UserCard.js
+++ b/js/src/forum/components/UserCard.js
@@ -30,7 +30,7 @@ export default class UserCard extends Component {
     const badges = user.badges().toArray();
 
     return (
-      <div className={'UserCard ' + (this.attrs.className || '')} style={color ? { backgroundColor: color } : ''}>
+      <div className={'UserCard ' + (this.attrs.className || '')} style={color ? { '--usercard-bg': color } : ''}>
         <div className="darkenBackground">
           <div className="container">
             {controls.length
--- a/js/src/forum/components/UserCard.js
+++ b/js/src/forum/components/UserCard.js
@@ -30,7 +30,7 @@ export default class UserCard extends Component {
     const badges = user.badges().toArray();
 
     return (
-      <div className={'UserCard ' + (this.attrs.className || '')} style={color ? { '--usercard-bg': color } : ''}>
+      <div className={'UserCard ' + (this.attrs.className || '')} style={color && { '--usercard-bg': color }}>
         <div className="darkenBackground">
           <div className="container">
             {controls.length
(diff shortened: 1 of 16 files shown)


davwheat:
Thank you. Thank you! THANK YOUUUU! 🥰

js/src/forum/components/UserCard.js
```diff
@@ -30,7 +30,7 @@ export default class UserCard extends Component {
     const badges = user.badges().toArray();
 
     return (
-      <div className={'UserCard ' + (this.attrs.className || '')} style={color ? { backgroundColor: color } : ''}>
+      <div className={'UserCard ' + (this.attrs.className || '')} style={color ? { '--usercard-bg': color } : ''}>
```
davwheat, line 33: This is a more common pattern used in JSX. I know it was here before, but we might as well use it.

```suggestion
      <div className={'UserCard ' + (this.attrs.className || '')} style={color && { '--usercard-bg': color }}>
```

davwheat (approved):

tankerkiller125 (approved):

PR #3005, opened 2021-08-08
Rewrite ItemList; update ItemList typings

### LINKED ISSUE #3030: ItemList item `content` can change type unexpectedly
## Bug Report

**Current Behavior**
ItemList items content property can change type unexpectedly.

**Steps to Reproduce**
```ts
const items = new ItemList()

items.add('myItem', 1, 10)
items.add('myItem2', 'hello', 0)

console.log(items.items)
// {
//    myItem: { content: 1, priority: 10 }, 
//    myItem1: { content: 'hello', priority: 10 }, 
// }

typeof items.items.myItem === "object" // false
typeof items.items.myItem === "number" // true

const arr = items.toArray()

console.log(items.items)
// Note that primitives have changed to their object counterparts
// {
//    myItem: { content: Number { 1 }, priority: 10 }, 
//    myItem1: { content: String { "hello" }, priority: 10 }, 
// }

typeof items.items.myItem === "object" // true
typeof items.items.myItem === "number" // false
```

**Expected Behavior**
Item content types should not change without warning.

davwheat:
**Fixes #3030**

- Converts the class to a generic
  - We now provide a typed ItemClass, so the `.get()`, `.toArray()`, etc return the correct type based on the contents of the ItemLIst
  - Extensions which use Typescript may need to modify their code to address new typing errors as a result. This does not break the code, however.
- New `ItemList` methods:
  - `setContent()` for setting the content of an item
  - `setPriority()` for setting the priority of an item
  - `getPriority()` for getting the priority of an item
- `replace()` **is now deprecated**
  - to set content, use the `setContent(key, content)` method
  - to set priority, use the `setPriority(key, priority)` method
- `ItemList.items` **is now deprecated**
  - instead, use `toObject()` to get a map of itemNames to `{ content, priority, itemName }` records
- `ItemList.toArray()` has a new parameter: `keepPrimitives`
  - Setting to `true` will prevent content of primitive types, such as numbers/strings/etc., from being converted into Object representations of themselves
  - Setting to `false` (default) will retain the current behaviour of converting to Object-ified types
  - All object values, regardless of `keepPrimitives`, are proxied to provide read-only access to an `itemName` property.

**Confirmed**

**Required changes:**

### DIFF
--- a/js/src/@types/global.d.ts
+++ b/js/src/@types/global.d.ts
@@ -1,3 +1,8 @@
+declare type Writable<T> = { -readonly [P in keyof T]: T[P] };
+declare type DeepWritable<T> = { -readonly [P in keyof T]: DeepWritable<T[P]> };
+
+declare type DeepReadonly<T> = { -readonly [P in keyof T]: DeepReadonly<T[P]> };
+
 /**
  * @deprecated Please import `app` from a namespace instead of using it as a global variable.
  *
--- a/js/src/@types/global.d.ts
+++ b/js/src/@types/global.d.ts
@@ -1,7 +1,7 @@
 declare type Writable<T> = { -readonly [P in keyof T]: T[P] };
 declare type DeepWritable<T> = { -readonly [P in keyof T]: DeepWritable<T[P]> };
 
-declare type DeepReadonly<T> = { -readonly [P in keyof T]: DeepReadonly<T[P]> };
+declare type DeepReadonly<T> = { readonly [P in keyof T]: DeepReadonly<T[P]> };
 
 /**
  * @deprecated Please import `app` from a namespace instead of using it as a global variable.
--- a/js/src/common/utils/ItemList.ts
+++ b/js/src/common/utils/ItemList.ts
@@ -311,5 +311,3 @@ export default class ItemList<T> {
     });
   }
 }
-
-window.ItemList = ItemList;
--- a/js/src/common/utils/ItemList.ts
+++ b/js/src/common/utils/ItemList.ts
@@ -80,9 +80,7 @@ export default class ItemList<T> {
     return this;
   }
 
-  // TODO: [Flarum 2.0] Remove deprecated `.replace()` syntax.
-
-  //? This is the deprecated syntax.
+  // TODO: [Flarum 2.0] Remove deprecated `.replace()` method.
   /**
    * Replace an item and/or priority in the list, only if it is already present.
    *
@@ -90,8 +88,8 @@ export default class ItemList<T> {
    *
    * If the provided `key` is not present, nothing will happen.
    *
-   * @deprecated Please use the `changePriority` method to set the priority for
-   * an item. This will be removed in Flarum 2.0.
+   * @deprecated Please use the `set` and `setPriority` methods to replace items
+   * and their priorities. This will be removed in Flarum 2.0.
    *
    * @param key The key of the item in the list
    * @param content The item's new content
@@ -106,27 +104,6 @@ export default class ItemList<T> {
    * @example <caption>Replace content and priority.</caption>
    * items.replace('myItem', <p>My new value.</p>, 10);
    */
-  replace(key: string, content?: T | null, priority?: number): this;
-
-  //? This is the new syntax.
-  /**
-   * Replaces an item's content, if the provided item key exists.
-   *
-   * If the provided `key` is not present, nothing will happen.
-   *
-   * @param key The key of the item in the list
-   * @param content The item's new content
-   *
-   * @example <caption>Replace item content.</caption>
-   * items.replace('myItem', <p>My new value.</p>);
-   *
-   * @example <caption>Replace item content and priority.</caption>
-   *          items
-   *            .replace('myItem', <p>My new value.</p>)
-   *            .changePriority('myItem', 10);
-   */
-  replace(key: string, content: T): this;
-
   replace(key: string, content: T | null = null, priority: number | null = null): this {
     if (!this.has(key)) return this;
 
@@ -142,6 +119,27 @@ export default class ItemList<T> {
   }
 
   /**
+   * Replaces an item's content, if the provided item key exists.
+   *
+   * If the provided `key` is not present, nothing will happen.
+   *
+   * @param key The key of the item in the list
+   * @param content The item's new content
+   *
+   * @example <caption>Replace item content.</caption>
+   * items.replace('myItem', <p>My new value.</p>);
+   *
+   * @example <caption>Replace item content and priority.</caption>
+   *          items
+   *            .replace('myItem', <p>My new value.</p>)
+   *            .setPriority('myItem', 10);
+   */
+  set(key: string, content: T): this {
+    // Saves on bundle size to call the deprecated method internally
+    return this.replace(key, content);
+  }
+
+  /**
    * Replaces an item's priority, if the provided item key exists.
    *
    * If the provided `key` is not present, nothing will happen.
@@ -150,14 +148,14 @@ export default class ItemList<T> {
    * @param priority The item's new priority
    *
    * @example <caption>Replace item priority.</caption>
-   * items.changePriority('myItem', 10);
+   * items.setPriority('myItem', 10);
    *
    * @example <caption>Replace item priority and content.</caption>
    *          items
-   *            .changePriority('myItem', 10)
+   *            .setPriority('myItem', 10)
    *            .replace('myItem', <p>My new value.</p>);
    */
-  changePriority(key: string, priority: number): this {
+  setPriority(key: string, priority: number): this {
     if (this.has(key)) {
       this._items[key].priority = priority;
     }
--- a/js/src/common/utils/ItemList.ts
+++ b/js/src/common/utils/ItemList.ts
@@ -139,8 +139,14 @@ export default class ItemList<T> {
    *          items
    *            .setContent('myItem', <p>My new value.</p>)
    *            .setPriority('myItem', 10);
+   *
+   * @throws If the provided `key` is not present in the ItemList.
    */
-  set(key: string, content: T): this {
+  setContent(key: string, content: T): this {
+    if (!this.has(key)) {
+      throw new Error(`[ItemList] Cannot set content of Item. Key \`${key}\` is not present.`);
+    }
+
     // Saves on bundle size to call the deprecated method internally
     return this.replace(key, content);
   }
@@ -160,11 +166,15 @@ export default class ItemList<T> {
    *          items
    *            .setPriority('myItem', 10)
    *            .setContent('myItem', <p>My new value.</p>);
+   *
+   * @throws If the provided `key` is not present in the ItemList.
    */
   setPriority(key: string, priority: number): this {
-    if (this.has(key)) {
-      this._items[key].priority = priority;
+    if (!this.has(key)) {
+      throw new Error(`[ItemList] Cannot set priority of Item. Key \`${key}\` is not present.`);
     }
+
+    this._items[key].priority = priority;
 
     return this;
   }
--- a/js/src/common/utils/ItemList.ts
+++ b/js/src/common/utils/ItemList.ts
@@ -21,6 +21,17 @@ class Item<T> {
 }
 
 /**
+ * Options for new `replace` syntax.
+ *
+ * Removes the need for passing `null` to parameters.
+ */
+interface IItemListReplaceOptions<T> {
+  _useNewSyntax: true;
+  content?: T;
+  priority?: number;
+}
+
+/**
  * The `ItemList` class collects items and then arranges them into an array
  * by priority.
  */
@@ -68,12 +79,50 @@ export default class ItemList<T> {
   /**
    * Replace an item in the list, only if it is already present.
    *
-   * If `content` or `priority` are `null`, these values will not be replaced.
+   * Only replaces values which are present in the object.
+   *
+   * @example <caption>Replace priority and not content.</caption>
+   * items.replace('myItem', { priority: 10 });
+   *
+   * @example <caption>Replace content and not priority.</caption>
+   * items.replace('myItem', { content: <p>My new value.</p> });
+   *
+   * @example <caption>Replace content and priority.</caption>
+   * items.replace('myItem', { content: <p>My new value.</p>, priority: 10 });
    */
-  replace(key: string, content: T | null = null, priority: number | null = null): this {
+  replace(key: string, options: IItemListReplaceOptions<T>): this;
+
+  // TODO: [Flarum 2.0] Remove deprecated `.replace()` syntax.
+  /**
+   * Replace an item in the list, only if it is already present.
+   *
+   * If `content` or `priority` are `null`, these values will not be replaced.
+   *
+   * @deprecated Please use the new object-based syntax. This syntax will be removed in Flarum 2.0.
+   *
+   * @example <caption>Replace priority and not content.</caption>
+   * items.replace('myItem', null, 10);
+   *
+   * @example <caption>Replace content and not priority.</caption>
+   * items.replace('myItem', <p>My new value.</p>);
+   *
+   * @example <caption>Replace content and priority.</caption>
+   * items.replace('myItem', <p>My new value.</p>, 10);
+   */
+  replace(key: string, contentOrOptions?: T | null, priority?: number | null): this;
+
+  replace(key: string, contentOrOptions: T | null | IItemListReplaceOptions<T> = null, priority: number | null = null): this {
     if (this.items[key]) {
-      if (content !== null) {
-        this.items[key].content = content;
+      if (contentOrOptions !== null) {
+        // Being called with new object-based syntax.
+        if ('_useNewSyntax' in contentOrOptions) {
+          if (contentOrOptions.content !== undefined) this.items[key].content = contentOrOptions.content;
+          if (contentOrOptions.priority !== undefined) this.items[key].priority = contentOrOptions.priority;
+
+          return this;
+        }
+
+        this.items[key].content = contentOrOptions;
       }
 
       if (priority !== null) {
--- a/js/src/common/utils/ItemList.ts
+++ b/js/src/common/utils/ItemList.ts
@@ -240,17 +240,15 @@ export default class ItemList<T> {
       const item = this._items[key];
       item.key = i;
 
-      if (objectifyContent) {
+      if (objectifyContent || typeof item.content === 'object') {
         // Convert content to object, then proxy it
-        return { ...item, content: this.createItemContentProxy(Object(item.content), key) };
+        return {
+          ...item,
+          content: this.createItemContentProxy(typeof item.content === 'object' ? item.content : Object(item.content), key),
+        };
       } else {
-        if (typeof item.content === 'object') {
-          // If the content already is an object, proxy it
-          return { ...item, content: this.createItemContentProxy(item.content, key) };
-        } else {
-          // ...otherwise just return a clone of the item.
-          return { ...item };
-        }
+        // ...otherwise just return a clone of the item.
+        return { ...item };
       }
     });
 
--- a/js/src/common/utils/ItemList.ts
+++ b/js/src/common/utils/ItemList.ts
@@ -147,7 +147,7 @@ export default class ItemList<T> {
   /**
    * Merge another list's items into this one.
    *
-   * The list passed to this function will overwrite items whichalready exist
+   * The list passed to this function will overwrite items which already exist
    * with the same key.
    */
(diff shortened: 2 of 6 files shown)


js/src/common/utils/ItemList.ts
```diff
@@ -1,35 +1,51 @@
-class Item {
-  content: any;
+class Item<T> {
```
askvortsov1, line 1: Yay!

js/src/common/utils/ItemList.ts
```diff
@@ -1,35 +1,51 @@
...
+      });
   priority: number;
+
+  /**
+   * Set when calling `.toArray()`
+   */
   key?: number;
 
-  constructor(content: any, priority?: number) {
+  constructor(content: T, priority: number) {
```
askvortsov1, line 17: Why not default this? Or do we want to enforce this as recommended code style?

js/src/common/utils/ItemList.ts
```diff
@@ -88,13 +146,18 @@ export default class ItemList {
 
   /**
    * Merge another list's items into this one.
+   *
+   * The list passed to this function will overwrite items whichalready exist
```
askvortsov1, line 150:
```suggestion
   * The list passed to this function will overwrite items which already exist
```

askvortsov1 (changes requested):
To me, "replace" implies a new content. Why not restrict "replace" to only do content, and have a separate "changePriority" method?

js/src/common/utils/ItemList.ts
davwheat: The default of 0 is provided to the constructor by the `add` method on ItemList.

js/src/common/utils/ItemList.ts
```diff
@@ -1,35 +1,51 @@
...
     this.priority = priority;
   }
 }
 
+/**
+ * Options for new `replace` syntax.
+ *
+ * Removes the need for passing `null` to parameters.
+ */
+interface IItemListReplaceOptions<T> {
```
datitisev, line 28: I think we should export this, right? Same with the class above? Is there any reason not to?

davwheat: Yeah, there's no reason not to.

I don't know why we didn't used to export the Item class. Was there any reason?

datitisev: I have no idea 🤷

askvortsov1: > Yeah, there's no reason not to.
> 
> I don't know why we didn't used to export the Item class. Was there any reason?

The Item class is an implementation detail, and should not be instantiated outside of this file. Likewise, IMO, interfaces describing the structure of function inputs shouldn't be exported either.

davwheat: I agree with you on not exporting the Item, but I'm on the fence about exporting the interface. The interface purely serves a purpose in typing, so exporting it could allow areas of core and extensions to encapsulate the function and modify it as needed, while still ensuring their new function is correctly typed.

Without exporting it, some hacky-ness is needed around the use of `Parameters<ItemList["replace"]>[1]` or something similar.

davwheat:
Removing the functionality of changing priority from the `replace` function is definitely a breaking change.

If I wanted to add a `changePriority` method, I'd likely choose to have it just encapsulate the `replace` function:

```ts
changePriority = (key: string, priority: number) => this.replace(key, { _useNewSyntax: true, priority });
```

I think of replace as replacing the entire item as a new one, rather than just the content. Maybe this needs some further discussion?

js/src/common/utils/ItemList.ts
askvortsov1: Good point, let's export the interface but not the item class then.

askvortsov1:
> I think of replace as replacing the entire item as a new one, rather than just the content.

Well, what does "entire item" mean? I would argue that priority is moreso a property of the relation between items and their container than a property of the item itself. When someone asks how they can replace something with something else, that generally implies replacing in place, but keeping the same order. On a behavioral level, I think "replace" and "rearrange" make sense as separate, simple operations. Would be happy to hear counterarguments though. 

Absolutely agree that we shouldn't remove this functionality until 2.0. We could deprecate it though?

davwheat:
Good idea. So...

- `replace` will replace content with deprecated support for priority replacement
- `setPriority` (or another name?) will set the priority of the current key

Part of me wonders if we can take a jQuery approach and have something like `items.select('key').replace('content').setPriority(10)` 🤔

askvortsov1:
> Good idea. So...
> 
> - `replace` will replace content with deprecated support for priority replacement
> - `setPriority` (or another name?) will set the priority of the current key
> 
> Part of me wonders if we can take a jQuery approach and have something like `items.select('key').replace('content').setPriority(10)` 🤔

Let's start with the first part, and look into `select` later. Prioritization!

askvortsov1:
> Good idea. So...
> 
> - `replace` will replace content with deprecated support for priority replacement
> - `setPriority` (or another name?) will set the priority of the current key
> 
> Part of me wonders if we can take a jQuery approach and have something like `items.select('key').replace('content').setPriority(10)` 🤔

Let's start with the first part, and look into `select` later. Prioritization!

js/src/common/utils/ItemList.ts
```diff
@@ -1,35 +1,51 @@
-class Item {
-  content: any;
+class Item<T> {
+  content:
+    | T
+    | (T & {
+        /**
+         * Set when calling `.toArray()`
+         */
+        itemName: string;
```
askvortsov1, line 8: I don't think we should mutate internal state when generating toArray output. Could we avoid doing so, and as a result, simplify these types?

js/src/common/utils/ItemList.ts
askvortsov1: This should be obsolete as we want to deprecate changing priority with replace.

js/src/common/utils/ItemList.ts
```diff
@@ -1,35 +1,51 @@
...
    */
   isEmpty(): boolean {
-    for (const i in this.items) {
-      if (this.items.hasOwnProperty(i)) {
-        return false;
-      }
-    }
-
-    return true;
+    return !!Object.keys(this.items).length;
```
askvortsov1, line 48:
```suggestion
    return Object.keys(this.items).length === 0;
```

js/src/common/utils/ItemList.ts
```diff
@@ -1,9 +1,18 @@
...
+class Item<T> {
+  content: T;
   priority: number;
+
+  /**
+   * The index of this item in the latest generated array.
+   *
+   * Set when calling `.toArray()`.
+   *
+   * **NOTE:** If you modify the item list after `.toArray()` is called,
```
askvortsov1, line 10: I don't think we should consider this to be part of the public API. Do we even need to set this on the internal representation?

js/src/common/utils/ItemList.ts
```diff
@@ -13,65 +22,141 @@ class Item {
...
-    return true;
+    return Object.keys(this._items).length === 0;
   }
 
   /**
    * Check whether an item is present in the list.
    */
   has(key: string): boolean {
-    return !!this.items[key];
+    return Object.keys(this._items).includes(key);
```
askvortsov1, line 56:
```suggestion
    return key in this._items;
```

js/src/common/utils/ItemList.ts
```diff
@@ -13,65 +22,141 @@ class Item {
...
+   *
+   * If the provided `key` is not present, nothing will happen.
+   *
+   * @param key The key of the item in the list
+   * @param content The item's new content
+   *
+   * @example <caption>Replace item content.</caption>
+   * items.replace('myItem', <p>My new value.</p>);
+   *
+   * @example <caption>Replace item content and priority.</caption>
```
askvortsov1, line 120: This example doesnt seem to apply here

js/src/common/utils/ItemList.ts
```diff
@@ -13,65 +22,141 @@ class Item {
...
+  replace(key: string, content?: T | null, priority?: number): this;
 
-      if (priority !== null) {
-        this.items[key].priority = priority;
-      }
+  //? This is the new syntax.
+  /**
+   * Replaces an item's content, if the provided item key exists.
+   *
+   * If the provided `key` is not present, nothing will happen.
```
askvortsov1, line 112: Maybe we should error in this case?

js/src/common/utils/ItemList.ts
```diff
@@ -81,51 +166,147 @@ export default class ItemList {
...
+   *
+   * **NOTE:** Modifying any objects in the final array may also update the
+   * content of the original ItemList.
+   */
+  toArray(keepPrimitives: true): (T extends object ? T & Readonly<{ itemName: string }> : T)[];
 
-    for (const i in this.items) {
-      if (this.items.hasOwnProperty(i) && this.items[i] instanceof Item) {
-        this.items[i].content = Object(this.items[i].content);
+  /**
```
askvortsov1, line 216: Can we get rid of this 3rd comment?

askvortsov1 (changes requested):
I'm looking over this again, and while I'm a fan of most of the changes (especially typing!), `toObject` and `toArray` seem way, way too complex for me

> * This proxy prevents direct modification of the internal state to prevent people breaking the class.

Why can't we just return an object of `key => content`? I don't think we should expose any internal state. This also lets us avoid all the complicated proxy stuff. On a side note, proxies are powerful, but we shouldn't use them unless we absolutely have to: it's a lot of added complexity.

> .toArray()'s items are now proxied, meaning that values which were objects are no longer modified to add the itemName property. This is now provided through the Proxy.

If we're returning an array of primitives, we shouldn't have an itemName at all, so proxying is unnecessary. If we're returning an array of objects, we can use spread syntax to clone the object and cleanly attach the itemName. Proxies would be best to avoid.

js/src/common/utils/ItemList.ts
davwheat: That would be a breaking change suited for 2.0.

js/src/common/utils/ItemList.ts
davwheat: This can cause misleading results:
```ts
const x = {};
'constructor' in x // true
```

askvortsov1: Didn't know that, interesting!

js/src/common/utils/ItemList.ts
askvortsov1: Let's open an issue for re-evaluating throwing errors / warnings for this util

davwheat:
> Why can't we just return an object of key => content?

Because we need to be able to access the internal state of Items to correctly `merge()` two ItemLists.

> If we're returning an array of primitives, we shouldn't have an `itemName` at all

We don't. `itemName` is added to object values. Primitives do not receive the `itemName` proxied property.

If we have an `ItemList` of translations (they are either strings or string arrays), then we have an `ItemList<string | string[]>`.

```ts
const items = new ItemList<string | string[]>();

items.add('a', app.translator.trans('key1'), 10); // 'test'
items.add('b', app.translator.trans('key2', { test: 'hello' }), 9); // ['test', 'hello']

items.toArray();
// [ String('test') { itemName: 'key1' }, Array ['test', 'hello'] { itemName: 'key2' } ]

items.toArray(false);
// [ 'test', Array ['test', 'hello'] { itemName: 'key2' } ]
```

> we can use spread syntax to clone the object

That will work fine for maps, but not other objects. If our ItemList holds dates, for example, all the date info is completely stripped when cloning with spread. It renders the ItemList completely useless.

We only proxy the content, as that is what is returned in the array. If we use spread to clone the `Item` object, then adding `itemName` to the content will end up mutating the internal state which is exactly what you said we should avoid.

```ts
let x = new Date();
let y = { ...x, itemName: 'date1' };

console.log(y);
// { itemName: 'date1' }

File: js/src/common/utils/ItemList.ts (hunk 9 above)
Comment (davwheat) [reply js/src/common/utils/ItemList.ts:120]: It was meant to illustrate the chaining and show an example of how to use the new syntax instead of the deprecated version.

File: js/src/common/utils/ItemList.ts (hunk 7 above)
Comment (davwheat) [reply js/src/common/utils/ItemList.ts:10]: We don't use it internally, no. There's a chance someone somewhere uses it, but if you think we should just remove it, I'm fine with that.

File: js/src/common/utils/ItemList.ts (hunk 11 above)
Comment (davwheat) [reply js/src/common/utils/ItemList.ts:216]: Yes! I thought it was needed to provide JSDoc when no parameters were given, but it isn't. :)

**Comment by askvortsov1** [comment]
> Because we need to be able to access the internal state of Items to correctly `merge()` two ItemLists.

I still don't feel right about this: maybe we should tsignore and use the private `_items` property directly, since the caller is within the ItemList code? Maybe a `(as this)` or something case could help there?

I'm starting to think that the whole `itemName`-being-set approach might not be the best way to do this: what if instead, we returned an array of [itemName, content] tuples? I realize this would be a breaking change so we couldn't do it yet, but in the long term it would be a lot less hacky.

**Comment by davwheat** [comment]
> what if instead, we returned an array of [itemName, content] tuples?

I'm not sure which way I'd prefer to do this:

```
return { itemNames: [ 'item1', 'item2', 'item3' ], items: [ <Item1 />, <Item2 />, <Item3 /> ] }
```

...or as tuples.

This way is more similar to the current interface, in the sense that we can do `const { items } = itemList.toArray()` without worrying about `map`-ing an array of tuples to just the content to render it.

File: js/src/common/utils/ItemList.ts (hunk 7 above)
Comment (askvortsov1) [reply js/src/common/utils/ItemList.ts:11]: Let's remove it.

**Comment by askvortsov1** [comment]
> > what if instead, we returned an array of [itemName, content] tuples?
> 
> I'm not sure which way I'd prefer to do this:
> 
> ```
> return { itemNames: [ 'item1', 'item2', 'item3' ], items: [ <Item1 />, <Item2 />, <Item3 /> ] }
> ```
> 
> ...or as tuples.
> 
> This way is more similar to the current interface, in the sense that we can do `const { items } = itemList.toArray()` without worrying about `map`-ing an array of tuples to just the content to render it.

Now that I think about it, we cant do either because BC. Let's keep the proxy stuff for now, and hope we eventually come up with something better

File: js/src/common/utils/ItemList.ts [hunk 12]
```diff
@@ -81,51 +169,143 @@ export default class ItemList {
...
-      if (items.items.hasOwnProperty(i) && items.items[i] instanceof Item) {
-        this.items[i] = items.items[i];
+  merge<K>(otherList: ItemList<K>): ItemList<T | K> {
+    const otherItems = otherList.toObject();
+
+    Object.keys(otherItems).forEach((key) => {
+      const val = otherItems[key];
+
+      if (val instanceof Item) {
+        (this as ItemList<T | K>)._items[key] = val;
```
askvortsov1, line 190: Doesn't this lose priority though?

askvortsov1 (commented):
> I still don't feel right about this: maybe we should tsignore and use the private _items property directly, since the caller is within the ItemList code? Maybe a (as this) or something case could help there?

Bump on this btw

js/src/common/utils/ItemList.ts
```diff
@@ -81,51 +169,143 @@ export default class ItemList {
...
    * Merge another list's items into this one.
+   *
+   * The list passed to this function will overwrite items which already exist
+   * with the same key.
    */
-  merge(items: this): this {
-    for (const i in items.items) {
-      if (items.items.hasOwnProperty(i) && items.items[i] instanceof Item) {
-        this.items[i] = items.items[i];
+  merge<K>(otherList: ItemList<K>): ItemList<T | K> {
```
askvortsov1, line 183: Can't we require that the other item list has the same type as this one statically?

davwheat: Requiring that seems like a bad idea. Merging two ItemLists with different types seems like something that should be allowed.

js/src/common/utils/ItemList.ts
davwheat: No, it's copying the whole Item class from the other ItemList and adding it to this one.

davwheat:
Rebased. Will address review comments shortly.

davwheat:
> I still don't feel right about this: maybe we should tsignore and use the private _items property directly, since the caller is within the ItemList code? Maybe a (as this) or something case could help there?

Turns out TS is more than fine with accessing private properties as long as its within the ItemClass' implementation of itself.

davwheat:
Ok, I've updated the OP again to contain my latest changes. I think I'm pretty settled on these and the new API overall. It should be backwards compatible **apart from**:

- `ItemList.items` cannot be modified directly to add/remove items
- `Item.key` is no longer present. This _could_ affect ordering for some ItemLists where multiple Items had the same priority, and may break

js/src/common/utils/ItemList.ts
askvortsov1: > Requiring that seems like a bad idea. Merging two ItemLists with different types seems like something that should be allowed.

My only concern is that we lose out on catching errors where incompatible ItemLists are merged by accident.

Idea: could we have the generic here be something like:

`merge<N extends K = K>(otherList: ItemList<N>): ItemList<N> {`

Then, by default, `merge(otherList)` would require `otherList` to be of `K`, but `merge<K|T>(otherList)` where `otherList \in ItemList<T> would work.

js/src/common/utils/ItemList.ts
```diff
@@ -81,51 +172,129 @@ export default class ItemList {
...
+          content: this.createItemContentProxy(isObject(item.content) ? item.content : Object(item.content), key),
+        };
+      } else {
+        // ...otherwise just return a clone of the item.
+        return { ...item };
       }
-    }
+    });
+
+    return items.sort((a, b) => b.priority - a.priority).map((item) => item.content);
```
askvortsov1, line 248: Why remove the check that returns items in order of being added? Or is that already given?

js/src/common/utils/ItemList.ts
```diff
@@ -81,51 +172,129 @@ export default class ItemList {
...
+   * items.add('b', 'My cool value', 20);
+   * items.add('a', 'My value', 10);
+   * items.toObject();
+   * // {
+   * //   a: { content: 'My value', priority: 10, itemName: 'a' },
+   * //   b: { content: 'My cool value', priority: 20, itemName: 'b' },
+   * // }
+   */
+  toObject(): DeepReadonly<Record<string, IItemObject<T>>> {
+    const keyItemMap = Object.keys(this._items).reduce((map, key) => {
```
askvortsov1, line 269: Why the intermediate variable?

js/src/common/utils/ItemList.ts
```diff
@@ -81,51 +172,129 @@ export default class ItemList {
...
+
+      map[key] = obj;
+
+      return map;
+    }, {} as Record<string, IItemObject<T>>);
+
+    return keyItemMap;
+  }
+
+  protected createItemContentProxy<C extends object>(content: C, key: string): Readonly<C & { itemName: string }> {
```
askvortsov1, line 284: Mixed feelings about making this protected. Could we go private instead, or at least add a JSDoc declaring it as internal?

askvortsov1 (changes requested):

js/src/common/utils/ItemList.ts
davwheat: Looking back on this now, I agree that we should just type that the ItemLists should be the same.

The correct type for the merged itemlist should be set when the itemlist is first initialised. We shouldn't be changing the type mid-way through, as that causes a multitude of problems.

js/src/common/utils/ItemList.ts
davwheat: We wanted to remove `key` from the Item class, which was used internally for this purpose.

`sort` should default to the order already present if `0` is returned due to the nature of its sort algorithm.

In the below test, 'b' and 'd' both have a priority (`p`) of 3. 'b' is present before 'd', in the initial declaration, and still is after the sort is complete.

image

askvortsov1 (approved):

js/src/@types/global.d.ts
```diff
@@ -1,3 +1,8 @@
+declare type Writable<T> = { -readonly [P in keyof T]: T[P] };
+declare type DeepWritable<T> = { -readonly [P in keyof T]: DeepWritable<T[P]> };
+
+declare type DeepReadonly<T> = { -readonly [P in keyof T]: DeepReadonly<T[P]> };
```
davwheat, line 4: I think it should be this, actually.

```suggestion
declare type DeepReadonly<T> = { readonly [P in keyof T]: DeepReadonly<T[P]> };
```

SychO9 (approved):

PR #3006, opened 2021-08-09
Rewrite application files to TS

davwheat:

- Rewrites Application JS files to TS
- Rewrites some closely related files that affected the typing abilities of Application

Any more specific/accurate types we could use?

**Confirmed**


SychO9 (dismissed):

datitisev:
There's a `resolver` optional key apart from component & path, I think?

askvortsov1:
> There's a `resolver` optional key apart from component & path, I think?
> 
> Yes - https://github.com/flarum/core/blob/eb4b18a979c7406cbf154a107662652d282fe415/js/src/common/utils/mapRoutes.js#L19
> 
> Either `component` or `resolver` are needed, not both.
> If `component` is passed, you can also have `resolverClass`.
> 
> https://github.com/flarum/core/blob/eb4b18a979c7406cbf154a107662652d282fe415/js/src/common/utils/mapRoutes.js#L22

Let's enforce this with a union type! Yay for type safety!

askvortsov1 (commented):
Why not just convert Application to TS? That could be one of the most valuable things to type.

davwheat:
I planned for this to just be an intermediary step, but screw it.

TYPESCRIPT, HERE WE COME! 🚀

js/src/common/Application.tsx
```diff
@@ -24,18 +24,80 @@ import PageState from './states/PageState';
...
+    }
+  | {
+      /**
+       * An instance of a route resolver.
+       */
+      resolver: RouteResolver<Attrs, Comp, RouteArgs>;
+    }
+);
+
+export interface RouteResolver<
```
askvortsov1, line 71: Why can't we use Mithril's `RouteResolver`?

js/src/common/Application.tsx
```diff
@@ -45,44 +107,31 @@ export default class Application {
...
    *
    * @example
-   * app.routes.discussion = {path: '/d/:id', component: DiscussionPage.component()};
-   *
-   * @type {Object}
-   * @public
+   * app.routes.discussion = { path: '/d/:id', component: DiscussionPage };
    */
-  routes = {};
+  routes: Record<string, RouteItem<ComponentAttrs, Component<ComponentAttrs & { routeName: string }>, Record<string, unknown>>> = {};
```
askvortsov1, line 112: Can we use a type alias (or several!) for this?

js/src/common/Application.tsx
```diff
@@ -94,73 +143,68 @@ export default class Application {
...
+    apiDocument: Record<string, unknown> | null;
+    locale: string;
+    locales: Record<string, string>;
+    resources: Record<string, unknown>[];
+    session: { userId: number; csrfToken: string };
+    [key: string]: unknown;
+  };
 
-  initialRoute;
+  title: string = '';
```
askvortsov1, line 202: These should probably be private

js/src/common/Application.tsx
```diff
@@ -195,44 +239,43 @@ export default class Application {
...
+    const appHeaderEl = document.querySelector('.App-header')!;
 
     // Add a class to the body which indicates that the page has been scrolled
     // down. When this happens, we'll add classes to the header and app body
     // which will set the navbar's position to fixed. We don't want to always
     // have it fixed, as that could overlap with custom headers.
-    const scrollListener = new ScrollListener((top) => {
-      const $app = $('#app');
-      const offset = $app.offset().top;
+    const scrollListener = new ScrollListener((top: number) => {
```
askvortsov1, line 258: Yay native js!

js/src/common/Application.tsx
```diff
@@ -247,36 +290,34 @@ export default class Application {
 
   /**
    * Determine the current screen mode, based on our media queries.
-   *
-   * @returns {String} - one of "phone", "tablet", "desktop" or "desktop-hd"
    */
-  screen() {
+  screen(): 'phone' | 'tablet' | 'desktop' | 'desktop-hd' {
```
askvortsov1, line 294: Maybe this one should be an exported type alias, since it's used in a lot of external code?

js/src/common/Application.tsx
```diff
@@ -287,11 +328,13 @@ export default class Application {
...
    * @see https://mithril.js.org/request.html
-   * @param {Object} options
+   *
+   * @param options
    * @return {Promise}
-   * @public
    */
-  request(originalOptions) {
+  request<ResponseType>(
+    originalOptions: Mithril.RequestOptions<ResponseType> & { errorHandler: (errorMessage: string) => void; url: string }
```
askvortsov1, line 336: Complex one-liner types are hard to parse: can we declare these up top and/or split these into a composition of simpler type declarations?

js/src/common/Application.tsx
```diff
@@ -300,27 +343,34 @@ export default class Application {
     // prevent redraws from occurring.
     options.background = options.background || true;
 
-    extend(options, 'config', (result, xhr) => xhr.setRequestHeader('X-CSRF-Token', this.session.csrfToken));
+    extend(options, 'config', (_: undefined, xhr: Parameters<Required<Mithril.RequestOptions<ResponseType>>['config']>[0]) => {
```
askvortsov1, line 346: Same here

js/src/common/Application.tsx
```diff
@@ -300,27 +343,34 @@ export default class Application {
...
 
     // If the method is something like PATCH or DELETE, which not all servers
     // and clients support, then we'll send it as a POST request with the
     // intended method specified in the X-HTTP-Method-Override header.
-    if (options.method !== 'GET' && options.method !== 'POST') {
+    if (options.method && !['GET', 'POST'].includes(options.method)) {
       const method = options.method;
-      extend(options, 'config', (result, xhr) => xhr.setRequestHeader('X-HTTP-Method-Override', method));
+
+      extend(options, 'config', (_: undefined, xhr: Parameters<Required<Mithril.RequestOptions<ResponseType>>['config']>[0]) => {
```
askvortsov1, line 356: Especially since it's reused twice

js/src/common/Application.tsx
```diff
@@ -330,15 +380,15 @@ export default class Application {
       let responseText;
 
       if (original) {
-        responseText = original(xhr.responseText);
+        responseText = original(xhr, options);
```
askvortsov1, line 383: This seems like a change in behavior

js/src/common/Application.tsx
```diff
@@ -399,7 +450,7 @@ export default class Application {
           content,
           controls: isDebug && [
             <Button className="Button Button--link" onclick={this.showDebug.bind(this, error, formattedError)}>
-              Debug
+              {app.translator.trans('core.lib.debug_button')}
```
askvortsov1, line 453: Nice catch!

js/src/common/utils/mapRoutes.ts
```diff
@@ -6,20 +8,23 @@ import DefaultResolver from '../resolvers/DefaultResolver';
...
  *
  * @see https://mithril.js.org/route.html#signature
- * @param {Object} routes
- * @param {String} [basePath]
- * @return {Object}
  */
-export default function mapRoutes(routes, basePath = '') {
-  const map = {};
+export default function mapRoutes(
+  routes: Record<string, RouteItem<Record<string, unknown>, Component<{ routeName: string; [key: string]: unknown }>, Record<string, unknown>>>,
```
askvortsov1, line 13: Here's another batch where extracting + composing into simpler types would be easier to parse. And ideally, reuse between here and Application

js/src/common/utils/mapRoutes.ts
```diff
@@ -6,20 +8,23 @@ import DefaultResolver from '../resolvers/DefaultResolver';
...
+  > = {};
 
   for (const routeName in routes) {
     const route = routes[routeName];
 
     if ('resolver' in route) {
       map[basePath + route.path] = route.resolver;
     } else if ('component' in route) {
-      const resolverClass = 'resolverClass' in route ? route.resolverClass : DefaultResolver;
+      const resolverClass = 'resolverClass' in route ? route.resolverClass! : DefaultResolver;
```
askvortsov1, line 27: Could we do a null coalesce here?
```suggestion
      const resolverClass = route.resolverClass ?? DefaultResolver;
```

js/src/forum/ForumApplication.ts
```diff
@@ -18,78 +18,65 @@ import DiscussionListState from './states/DiscussionListState';
...
 
     routes(this);
 
-    /**
-     * An object which controls the state of the cached discussion list, which
-     * is used in the index page and the slideout pane.
-     *
-     * @type {DiscussionListState}
-     */
     this.discussions = new DiscussionListState({});
```
askvortsov1, line 80: Could we move this to the definition? `discussions: DiscussionListState = new DiscussionListState({});`

askvortsov1 (changes requested):

js/src/common/Application.tsx
davwheat: This is more specific to our use case, in terms of typing attributes and the fact that Flarum uses its Component class for basically every component.

js/src/common/Application.tsx
davwheat: Yes, it was. The behaviour didn't match with the official Mithril typings, as we expected to pass the response text rather than the XHR object.

I'm changing the union type to a custom interface, and changing this back to fix it.

askvortsov1 (approved):

davwheat:
Rebased onto master to fix the conflicts

SychO9 (approved):

davwheat:
LET'S GET TYPESCRIPTIN'

PR #3007, opened 2021-08-09
Rewrite ModalManager and state to Typescript

davwheat:
**Progresses flarum/framework#3533**

- Rewrites files to TS
  - Fixes modal attrs being marked as required 
- Adds new `isModalOpen()` method

Should we make `closeTimeout` private, or just mark it as `@internal`? I don't see any reason that people would use it.

**Confirmed**


Pi-Bouf (approved):

js/src/common/states/ModalManagerState.ts
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * Class used to manage modal state.
+ *
+ * Accessible on the `app` object via `app.modal` property.
+ */
+export default class ModalManagerState {
+  /**
+   * @internal
+   */
+  modal: null | {
```
askvortsov1, line 12: Maybe this should be generic? Then, could we enforce that the attrs correspond to the modal class in question?

js/src/common/states/ModalManagerState.ts
```diff
@@ -0,0 +1,75 @@
...
+    componentClass: typeof Modal;
+    attrs?: Record<string, unknown>;
+  };
+
+  private closeTimeout?: number;
+
+  constructor() {
+    this.modal = null;
```
askvortsov1, line 20: Could we default this outside of the constructor?

js/src/common/states/ModalManagerState.ts
```diff
@@ -0,0 +1,75 @@
...
+    });
+  }
+
+  /**
+   * Checks if a modal is currently open.
+   *
+   * @returns `true` if a modal dialog is currently open, otherwise `false`.
+   */
+  isModalOpen(): boolean {
+    return !!this.modal;
```
askvortsov1, line 73:
```suggestion
    return this.modal != null;
```

nit: this feels a bit safer

askvortsov1 (changes requested):

js/src/common/states/ModalManagerState.ts
datitisev: It doesn't matter because we render the modal based on its boolean value anyways (`modal ?` or `modal &&` in ModalManager)... I'd actually prefer we check if its falsy instead of null in case somehow it gets set to a falsy value that isn't null, as then this method would return true but a modal would not be open.

askvortsov1: > somehow it gets set to a falsy value

How? Isn't that the goal of having these typings?

datitisev: Typings don't prevent from someone else setting something though? In either case I think it'd be a good idea to keep the logic the same - either change both to `!= null` or falsy, there shouldn't be differences there.

askvortsov1: Let's keep this as falsy then

js/src/common/states/ModalManagerState.ts
davwheat: How would that be done, though? We can't make the class generic as it is only initialised once when the frontend boots in the browser.

We could make the `show` method generic to enforce the second argument (props) matching the Modal class (if we then convert that to TS too), but that wouldn't enforce this property.

askvortsov1: Looking back at this, I wonder why we didn't have this accept a vnode of type Modal, instead of combining class + attrs dynamically? Then, we could have enforced static prop requirements.

Either way, it's good for now.

js/src/common/states/ModalManagerState.ts
askvortsov1: bump

askvortsov1 (approved):

js/src/common/components/Modal.tsx
```diff
@@ -1,33 +1,42 @@
...
+}
+
 /**
  * The `Modal` component displays a modal dialog, wrapped in a form. Subclasses
  * should implement the `className`, `title`, and `content` methods.
- *
- * @abstract
  */
-export default class Modal extends Component {
+export default abstract class Modal<ModalAttrs extends Record<string, unknown> = {}> extends Component<ModalAttrs & IInternalModalAttrs> {
```
datitisev, line 20: What's the reasoning behind splitting the internal & other attrs? Don't all methods use `ModalAttrs & IInternalModalAttrs` anyways (both in this class & parent Component)?

This just seems like it overcomplicates the code for no reason, I'd rather just have a `ModalAttrs` interface that extends `ComponentAttrs`

js/src/common/components/Modal.tsx
```diff
@@ -78,70 +85,59 @@ export default class Modal extends Component {
 
   /**
    * Get the class name to apply to the modal.
-   *
-   * @return {String}
-   * @abstract
    */
-  className() {}
+  abstract className(): string;
```
datitisev, line 89: Not sure how I feel about making these abstract... if an extension doesn't implement any of these methods it'll break - I envision this happening mostly with `className`. If they have a subclass that is also treated as an abstract Modal class, then that would also break.

What I'm saying is, this is a breaking change 🙃

js/src/common/components/ModalManager.tsx
```diff
@@ -1,29 +1,35 @@
...
       <div className="ModalManager modal fade">
-        {modal
-          ? modal.componentClass.component({
-              ...modal.attrs,
-              animateShow: this.animateShow.bind(this),
-              animateHide: this.animateHide.bind(this),
-              state: this.attrs.state,
-            })
-          : ''}
+        {modal &&
```
datitisev, line 21: This should be `!!modal`, right?

js/src/common/components/Modal.tsx
davwheat: > I'd rather just have a `ModalAttrs` interface that extends `ComponentAttrs`

`ComponentAttrs` includes native HTML attributes, which don't apply in this situation. Extending with it would add typing for `className`, `id`, etc, which aren't inherently valid Modal attributes.

js/src/common/components/ModalManager.tsx
davwheat: It doesn't make a difference, does it? Typescript doesn't seem to mind.

datitisev: I meant because of the changes you do in other places to avoid showing falsy values.

js/src/common/components/Modal.tsx
datitisev: Ah, I see. Can't we just have a ModalAttrs interface that contains teh `IInternalModalAttrs` attrs and extends `Record<string, unknown>` so we don't need to see `ModalAttrs & IInternalModalAttrs` everywhere? Or nah?

davwheat: I don't like the idea of extending with `Record<string, unknown>` as that allows props that haven't actually been typed on the modal.

https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgJICUD2mzIgD0hABMBnZdCBTKYgHlLClAHMAaZAVxAGsRMA7iAB8yAN4BYAFDJZyALYBPAApRMABwBcyRsxAtpAX2nTQkWIhSpVG8gSJk0WHOOlyFigPJgAFtBtayCCc8gBG0EYmUtQgjMj42tZq6uQAvK4yckoB2gBEPsAAhLlsbllevv7J2gCMpZmyvsCkqKQAcjgAKorqEMR5oZy4zTpgwAA248iISOqQxMihisidAMolkVJAA

datitisev: Doesn't doing `ModalAttrs extends Record<string, unknown> = {}` achieve the same thing? 🤔

davwheat: Oooooooooops, yes it does. I meant to remove that.

dsevillamartin (changes requested):

SychO9 (approved):

PR #3011, opened 2021-08-11
feat: Allow registering settings as `Less` config vars through Settings Extender

### LINKED ISSUE #2819: `Less` Configuration Variables
There are times where I wanted to add a setting which I needed to use in `Less` code similarly to how the primary color is used. It would be great to have:

1. The ability to add a `Less` configuration variable through an extender (Theme extender makes sense).
2. The ability to make it so that a setting that is changed clears the cache, at least the CSS cached assets, and this is for 1 where a configuration value is likely to be a setting the admin can edit, in which case a new set of CSS files need to be compiled with the new value. Right now configuration values such as primary color trigger a cache clear because the setting name starts with `theme_` which I used as a around before for my own settings, but I think it'd be best to improve this and make it a system. (Theme extender also makes sense atm).

SychO9:
**Fixes #2819**

This PR adds a `registerLessConfigVar` method to the settings extender that allows passing those settings values to the `LESS` code similarly to how they are serialized to the frontend. It also makes sure such settings always flushes the CSS cache when saving new values.

In the issue linked, the plan was to initially introduce this in a `Theme` extender so that, but the `Settings` extender make more sense, since all we want is to directly use settings as config vars.

General code quality, anything I might've missed.

**Confirmed**


tests/integration/extenders/SettingsTest.php
```diff
@@ -163,6 +163,38 @@ public function custom_setting_default_passed_to_callback()
...
+    public function custom_less_var_works_if_registered()
+    {
+        $this->extend(
+            (new Extend\Frontend('forum'))
+                ->css(__DIR__.'/../../fixtures/less/config.less'),
+            (new Extend\Settings())
+                ->registerLessConfigVar('custom-config-setting', 'custom-prefix.custom_setting')
+        );
+
+        $response = $this->send($this->request('GET', '/'));
```
davwheat, line 194: How tricky would it be to actually check that the value is correctly declared in the compiled CSS assets?

```
$response = $this->send($this->request('GET', '/assets/forum.css'));

// I don't know the syntax at all...
$this->assertTrue(true, strpos($response->getData(), '--custom-config-setting:customValue;') === false
```

SychO9: For some reason the request to the file didn't work, though `forum.css` is no small file, so it could be memory consuming.

If the variable doesn't actually work, the LESS parser will error out and produce a 500 error, that should be enough for this test.

src/Frontend/FrontendServiceProvider.php
```diff
@@ -57,6 +57,31 @@ public function register()
...
+                    'default' => '#000',
+                ],
+                'config-secondary-color' => [
+                    'key' => 'theme_secondary_color',
+                    'default' => '#000',
+                ],
+                'config-dark-mode'       => [
+                    'key' => 'theme_dark_mode',
+                    'callback' => function ($value) {
+                        return $value ? 'true' : 'false';
```
askvortsov1, line 74: Couldn't we cast to string?

tests/integration/extenders/SettingsTest.php
askvortsov1: > For some reason the request to the file didn't work, though forum.css is no small file, so it could be memory consuming.

It shouldn't, if you're trying to get it via a request, since static asset handling is done at the web server level.  Could we directly read it from the file system?

src/Frontend/FrontendServiceProvider.php
SychO9: Nope :/

```php
php > var_dump((string) true);
string(1) "1"
php > var_dump((string) false);
string(0) ""
```

askvortsov1: 😔

askvortsov1 (approved):

src/Frontend/FrontendServiceProvider.php
```diff
@@ -83,17 +108,18 @@ public function addBaseCss(SourceCollector $sources)
...
+                $var = $vars[$name];
+                $value = $settings->get($var['key'], $var['default'] ?? null);
+
+                if (isset($var['callback'])) {
+                    $value = $var['callback']($value);
+                }
 
-            return array_reduce(array_keys($vars), function ($string, $name) use ($vars) {
-                return $string."@$name: {$vars[$name]};";
+                return $string."@$name: {$value};";
```
davwheat, line 122: One of the things that worries me here is the possibility of injecting stuff dangerously (e.g. `0; @my-other-var: 1px`), but tbh that's not our problem.

It might be worth adding that injecting stuff here is possible, and so not to blatantly trust any user input.

davwheat (approved):

src/Frontend/FrontendServiceProvider.php
SychO9: True, a bad input could result in the parser failing with a 500 error, possibly also locking the admin out if used in the admin styles as well. I'll adjust the docblock. Really need to look into having the parser gracefully fail instead, a badly compiled CSS is better than an error blocking access.

src/Extend/Settings.php
```diff
@@ -43,6 +44,29 @@ public function serializeToForum(string $attributeName, string $key, $callback =
         return $this;
     }
 
+    /**
+     * Register a setting as a LESS configuration variable.
```
SychO9, line 48:
```suggestion
     * Register a setting as a LESS configuration variable. User input should not be trusted and should be validated for correct values.
```

src/Frontend/FrontendServiceProvider.php
askvortsov1: > a bad input could result in the parser failing with a 500 error,

Could we throw https://github.com/flarum/core/blob/b8754c7d7db07bbea25d15e285da9dadeece72ce/src/Forum/ValidateCustomLess.php at it?

SychO9: Oh wow didn't know that class existed, it attempts a LESS compilation with the new settings, but only when the `custom_less` setting has been modified, we should use that instead of `RecompileFrontendAssets`.

askvortsov1: Yeah, we could configure that to respond to changes of all setting-linked variables, or build in a LESS-error-reporting system so build fails don't shut down the site.

davwheat (approved):

src/Settings/SettingsServiceProvider.php
```diff
@@ -22,7 +22,10 @@ class SettingsServiceProvider extends AbstractServiceProvider
     public function register()
     {
         $this->container->singleton('flarum.settings.default', function () {
-            return new Collection();
+            return new Collection([
+                'theme_primary_color' => '#4D698E',
+                'theme_secondary_color' => '#4D698E',
+            ]);
```
SychO9, line 28: Would we want to actually remove the defaults from the installation step in favor of this ? 🤔

askvortsov1: That's the goal (as well as getting rid of data migrations). Probably in a separate PR though.

askvortsov1 (approved):

PR #3012, opened 2021-08-11
Limit height of code blocks

pierres:
Users often post lengthy code or configuration listings which makes following the actual discussion difficult. Therefore we limit a code block by half a screen while still being able to scroll through the code listing itself.

### DIFF
--- a/less/forum/Post.less
+++ b/less/forum/Post.less
@@ -149,7 +149,7 @@
       overflow-x: auto;
       // Backwards compatibility for browsers that don't support `max()`
       max-height: 50vh;
-      max-height: max(50vh, 250px);
+      max-height: ~"max(50vh, 250px)";
     }
   }
   h1, h2, h3, h4, h5, h6 {
--- a/less/forum/Post.less
+++ b/less/forum/Post.less
@@ -147,6 +147,8 @@
       border-radius: 0;
       display: block;
       overflow-x: auto;
+      // Backwards compatibility for browsers that don't support `max()`
+      max-height: 50vh;
       max-height: max(50vh, 250px);
     }
   }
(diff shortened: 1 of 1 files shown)


less/forum/Post.less
```diff
@@ -147,6 +147,7 @@
       border-radius: 0;
       display: block;
       overflow-x: auto;
+      max-height: 50vh;
```
davwheat, line 150: I think adding a minimum width might be a good idea, just in case. Maybe 300px?

pierres: This would waste space for e.g. single line code block which are not uncommon. E.g.: https://forum.archlinux.de/d/34082-networkmanager-failed-nach-update

davwheat: Ah yes, my bad, what I was thinking of was using:

```css
max-height: 50vh;
max-height: max(50vh, 300px);
```

That way very small devices don't get tiiiiny code blocks, and big devices are still limited to 50vh. What do you think?

pierres: Sure, this would stop the code block from getting too small. I guess you would need to switch to landscape mode on a small device to trigger this issue though.

SychO9 (approved):
Thank you! I've wanted this done for a while 😍 gets hard to scroll when people post logs.

davwheat (approved):
@SychO9 Double-check you're happy with the compatibility change I made, otherwise good to merge 👍 

Thank you @pierres!

SychO9:
@davwheat fine by me 👍

davwheat:
Ohh hold off a moment... LESS doesn't like `max`

davwheat:
Tests should pass now... 🤞

SychO9:
I always miss that when using vanilla CSS functions, LESS almost always has the same functions implemented that calculate the result when compiling.

davwheat:
Me too. It's a real pain 😕

pierres:
Thanks a lot. Next time I'll try not to forget about LESS and older browsers :-)

PR #3014, opened 2021-08-11
[A11Y] Make checkboxes focusable

### LINKED ISSUE #2973: [A11Y] Cannot focus switches (`display: none`)

davwheat:
**Fixes #2973**

- Make checkbox inputs appear over their display elements so that they are accessible to mobile screen-readers
- Add `aria-hidden` to the purely visual checkbox elements
- Restore browser focus rings to the checkbox display elements

Nothing, really.

**Screenshot**

https://user-images.githubusercontent.com/7406822/129100197-7769a3b1-e8ed-4e18-83ea-51db2293ec56.mp4

**Confirmed**

### DIFF
--- a/less/common/Checkbox.less
+++ b/less/common/Checkbox.less
@@ -2,19 +2,42 @@
   display: block;
   cursor: pointer;
   margin: 0;
+  position: relative;
 
   input[type="checkbox"] {
-    .visually-hidden();
+    position: absolute;
+    z-index: 1;
+
+    opacity: 0;
+
+    top: 0;
+    bottom: 0;
+    left: 0;
+    right: 0;
+
+    width: 100%;
+    height: 100%;
+
+    cursor: pointer;
+
     .add-keyboard-focus-ring-nearby("+ .Checkbox-display");
   }
 }
 .Checkbox--switch {
-  padding-left: 65px;
+  @left-pad: 65px;
+
+  padding-left: @left-pad;
   margin: 5px 0;
+
+  input[type="checkbox"] {
+    top: -4px;
+    width: 50px;
+    height: 28px;
+  }
 
   .Checkbox-display {
     float: left;
-    margin-left: -65px;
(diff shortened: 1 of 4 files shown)


less/common/Checkbox.less
```diff
@@ -3,8 +3,9 @@
   cursor: pointer;
   margin: 0;
 
-  & input[type="checkbox"] {
-    display: none;
+  input[type="checkbox"] {
```
askvortsov1, line 6: Why not &?

SychO9: it doesn't make a difference, it's the same.

less/common/scaffolding.less
```diff
@@ -159,3 +159,13 @@ blockquote ol:last-child {
...
+
+.visually-hidden {
+  clip: rect(0 0 0 0);
+  clip-path: inset(50%);
+  height: 1px;
+  overflow: hidden;
+  position: absolute;
+  white-space: nowrap;
+  width: 1px;
+}
```
SychO9, line 171: Can we make the checkbox the same dimensions as the switch itself ? we could make them parameters and then pass those values in (we'd have to relatively position the checkbox's parent and add `top:0;left:0;margin-top:-5px` as well to the checkbox)

Something I read here: https://www.sarasoueidan.com/blog/inclusively-hiding-and-styling-checkboxes-and-radio-buttons/#hiding-the-checkboxes-inclusively

davwheat: Good catch. I didn't realise that was how it worked!

davwheat:
Confirmed that Talkback on Android works correctly with the checkboxes.

Interestingly, `aria-hidden=true` needs to be set as opposed to just `aria-hidden` for talkback to pick up on it. I'm going to need to fix some other bits of core to reflect that.

askvortsov1 (approved):

SychO9 (approved):

PR #3018, opened 2021-08-12
[A11Y] Add focus traps to modals and nav drawer

### LINKED ISSUE #2663: [A11Y] Triggering click on drawer button can cause layered backdrops
## Bug Report

**Current Behavior**
On mobile viewports, focusing the nav drawer toggle button and triggering a click on it multiple times (e.g. via space) causes multiple backdrops to be shown, but only one of these closes when closing the drawer.

1l96mO

**Steps to Reproduce**
1. Shrink viewport to mobile size
2. Use TAB to focus the drawer toggle button
3. Hit SPACE twice
4. Close the drawer
5. Backdrop remains

**Expected Behavior**
Only one backdrop should be created.

**Possible solution(s)**

I have two ideas:

1. We prevent this situation happening by checking if the drawer is currently open on button click. This fixes this situation.
2. We put a check in the Drawer class to prevent it being opened if App has `drawerOpen`. This fixes it in case any exts open the drawer or if something else can cause this. We may also wish to consider removing all `.drawer-backdrop` as a precautionary measure when closing, too.

### LINKED ISSUE #2665: [A11Y] Focus can leave open nav drawer on mobile
## Bug Report

**Current Behavior**
Using TAB to shift focus around the mobile nav drawer allows focus to leave the drawer to the page content behind it.

**Steps to Reproduce**
1. Shrink viewport
2. Open drawer
3. Tab to last drawer element
4. Press tab again
5. Focus is in background page

**Expected Behavior**
Focus should be trapped inside the nav drawer

**Possible Solution**
We need a focus trapper.

We could use a [pre-existing one](https://www.npmjs.com/package/focus-trap) from npm, but [it's 2.9 kB in size](https://bundlephobia.com/result?p=focus-trap) (and steadily increasing), which is a fair amount for what it does.

As that supports IE9, and we have jQuery (for now), we might be able to make our own version based off of that package, but using jQuery snippets instead, which should cut down on bundle size a ton.

davwheat:
**Fixes #2663**
**Fixes #2665**

> ### **Please merge #3007 before this PR!**

- Add focus trap to modals
- Add focus trap to drawer on phone viewports
- Auto close drawer when leaving phone viewport
- Replace deprecated JQuer (`$.click()` -> `$.on('click')`)
- Use `requestAnimationFrame` instead of `setTimeout`

**Screenshot**

**Confirmed**

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -9,6 +9,7 @@
     "clsx": "^1.1.1",
     "color-thief-browser": "^2.0.2",
     "dayjs": "^1.10.7",
+    "focus-trap": "^6.7.1",
     "jquery": "^3.6.0",
     "jquery.hotkeys": "^0.1.0",
     "mithril": "^2.0.4",
--- a/js/src/common/components/ModalManager.tsx
+++ b/js/src/common/components/ModalManager.tsx
@@ -1,8 +1,12 @@
 import Component from '../Component';
-import { createFocusTrap, FocusTrap } from '../utils/focusTrap';
 
-import type Mithril from 'mithril';
+import { createFocusTrap, FocusTrap } from '../utils/focusTrap';
+import { tabbable } from 'tabbable';
+
 import type ModalManagerState from '../states/ModalManagerState';
+import type Mithril from 'mithril';
+
+window.tabbable = tabbable;
 
 interface IModalManagerAttrs {
   state: ModalManagerState;
@@ -14,7 +18,12 @@ interface IModalManagerAttrs {
  * overwrite the previous one.
  */
 export default class ModalManager extends Component<IModalManagerAttrs> {
-  focusTrap: FocusTrap | undefined;
+  protected focusTrap: FocusTrap | undefined;
+
+  /**
+   * Whether a modal is currently shown by this modal manager.
+   */
+  protected modalShown: boolean = false;
 
   view(vnode: Mithril.VnodeDOM<IModalManagerAttrs, this>): Mithril.Children {
     const modal = this.attrs.state.modal;
@@ -45,12 +54,23 @@ export default class ModalManager extends Component<IModalManagerAttrs> {
 
   onupdate(vnode: Mithril.VnodeDOM<IModalManagerAttrs, this>): void {
     super.onupdate(vnode);
+
+    requestAnimationFrame(() => {
+      try {
+        if (this.modalShown) this.focusTrap!.activate?.();
+        else this.focusTrap!.deactivate?.();
+      } catch {
+        // We can expect errors to occur here due to the nature of mithril rendering
+      }
+    });
   }
 
   animateShow(readyCallback: () => void): void {
     if (!this.attrs.state.modal) return;
 
     const dismissible = !!this.attrs.state.modal.componentClass.isDismissible;
+
+    this.modalShown = true;
 
     // If we are opening this modal while another modal is already open,
     // the shown event will not run, because the modal is already open.
@@ -68,14 +88,12 @@ export default class ModalManager extends Component<IModalManagerAttrs> {
         keyboard: dismissible,
       })
       .modal('show');
-
-    this.focusTrap.activate?.();
   }
 
   animateHide(): void {
     // @ts-expect-error: No typings available for Bootstrap modals.
     this.$().modal('hide');
 
-    this.focusTrap.deactivate?.();
+    this.modalShown = false;
   }
 }
--- a/js/src/common/components/ModalManager.tsx
+++ b/js/src/common/components/ModalManager.tsx
@@ -1,4 +1,5 @@
 import Component from '../Component';
+import { createFocusTrap, FocusTrap } from '../utils/focusTrap';
 
 import type Mithril from 'mithril';
 import type ModalManagerState from '../states/ModalManagerState';
@@ -13,7 +14,9 @@ interface IModalManagerAttrs {
  * overwrite the previous one.
  */
 export default class ModalManager extends Component<IModalManagerAttrs> {
-  view() {
+  focusTrap: FocusTrap | undefined;
+
+  view(vnode: Mithril.VnodeDOM<IModalManagerAttrs, this>): Mithril.Children {
     const modal = this.attrs.state.modal;
 
     return (
@@ -29,13 +32,19 @@ export default class ModalManager extends Component<IModalManagerAttrs> {
     );
   }
 
-  oncreate(vnode: Mithril.VnodeDOM<IModalManagerAttrs, this>) {
+  oncreate(vnode: Mithril.VnodeDOM<IModalManagerAttrs, this>): void {
     super.oncreate(vnode);
 
     // Ensure the modal state is notified about a closed modal, even when the
     // DOM-based Bootstrap JavaScript code triggered the closing of the modal,
     // e.g. via ESC key or a click on the modal backdrop.
     this.$().on('hidden.bs.modal', this.attrs.state.close.bind(this.attrs.state));
+
+    this.focusTrap = createFocusTrap(this.element as HTMLElement);
+  }
+
+  onupdate(vnode: Mithril.VnodeDOM<IModalManagerAttrs, this>): void {
+    super.onupdate(vnode);
   }
 
   animateShow(readyCallback: () => void): void {
@@ -59,10 +68,14 @@ export default class ModalManager extends Component<IModalManagerAttrs> {
         keyboard: dismissible,
       })
       .modal('show');
+
+    if (this.focusTrap) this.focusTrap.activate();
   }
 
   animateHide(): void {
     // @ts-expect-error: No typings available for Bootstrap modals.
     this.$().modal('hide');
+
+    if (this.focusTrap) this.focusTrap.deactivate();
   }
 }
--- a/js/src/common/components/ModalManager.tsx
+++ b/js/src/common/components/ModalManager.tsx
@@ -1,12 +1,9 @@
 import Component from '../Component';
 
 import { createFocusTrap, FocusTrap } from '../utils/focusTrap';
-import { tabbable } from 'tabbable';
 
 import type ModalManagerState from '../states/ModalManagerState';
 import type Mithril from 'mithril';
-
-window.tabbable = tabbable;
 
 interface IModalManagerAttrs {
   state: ModalManagerState;
--- a/js/src/common/components/ModalManager.tsx
+++ b/js/src/common/components/ModalManager.tsx
@@ -69,13 +69,13 @@ export default class ModalManager extends Component<IModalManagerAttrs> {
       })
       .modal('show');
 
-    if (this.focusTrap) this.focusTrap.activate();
+    this.focusTrap.activate?.();
   }
 
   animateHide(): void {
     // @ts-expect-error: No typings available for Bootstrap modals.
     this.$().modal('hide');
 
-    if (this.focusTrap) this.focusTrap.deactivate();
+    this.focusTrap.deactivate?.();
   }
 }
--- a/js/src/common/utils/Drawer.js
+++ b/js/src/common/utils/Drawer.js
@@ -1,5 +1,4 @@
 import { createFocusTrap } from './focusTrap';
-import { throttle } from './throttleDebounce';
 
 /**
  * The `Drawer` class controls the page's drawer. The drawer is the area the
@@ -13,7 +12,7 @@ export default class Drawer {
   focusTrap;
 
   /**
-   * @type {HTMLElement}
+   * @type {HTMLDivElement}
    */
   appElement;
 
@@ -28,8 +27,10 @@ export default class Drawer {
     });
 
     this.appElement = document.getElementById('app');
-
     this.focusTrap = createFocusTrap('#drawer', { allowOutsideClick: true });
+    this.drawerAailableMediaQuery = window.matchMedia(
+      `(max-width: ${getComputedStyle(document.documentElement).getPropertyValue('--screen-phone-max')})`
+    );
   }
 
   /**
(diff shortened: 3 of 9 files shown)


js/src/common/components/ModalManager.js
```diff
@@ -1,40 +1,72 @@
...
+
+  onupdate(vnode) {
+    super.onupdate(vnode);
+
+    requestAnimationFrame(() => {
+      try {
+        if (this.modalShown) this.focusTrap.activate?.();
+        else this.focusTrap.deactivate?.();
+      } catch {
+        // We can expect errors to occur here due to the nature of mithril rendering
```
askvortsov1, line 60: Could we catch a more specific kind of exception?

js/src/common/components/ModalManager.js
```diff
@@ -1,40 +1,72 @@
...
+      }
+    });
   }
 
   animateShow(readyCallback) {
     const dismissible = !!this.attrs.state.modal.componentClass.isDismissible;
 
+    this.modalShown = true;
```
askvortsov1, line 68: Maybe these should happen before the jquery calls? That way, if there's rerendering triggered by jquery (is that a thing?), it'll be applied correctly.

js/src/common/utils/Drawer.js
```diff
@@ -1,20 +1,41 @@
...
         this.hide();
       }
     });
+
+    this.focusTrap = createFocusTrap('#drawer', { allowOutsideClick: true });
+  }
+
+  resizeHandler(e) {
+    if (this.isOpen()) {
+      if (document.documentElement.style.getPropertyValue('--flarum-screen') !== 'phone') {
```
askvortsov1, line 30: Can we use the `app.screen` encapsulation?

js/src/common/utils/Drawer.js
```diff
@@ -1,20 +1,41 @@
...
+  /**
+   * @type {import('./focusTrap').FocusTrap}
+   */
+  focusTrap;
+
   constructor() {
     // Set up an event handler so that whenever the content area is tapped,
     // the drawer will close.
-    $('#content').click((e) => {
+    $('#content').on('click', (e) => {
```
askvortsov1, line 18: Maybe we should just use the native JS method?

js/src/common/utils/Drawer.js
```diff
@@ -1,20 +1,41 @@
...
         e.preventDefault();
         this.hide();
       }
     });
+
+    this.focusTrap = createFocusTrap('#drawer', { allowOutsideClick: true });
+  }
+
+  resizeHandler(e) {
+    if (this.isOpen()) {
```
askvortsov1, line 29: Can we use `&&` instead of nesting?

js/src/common/utils/Drawer.js
```diff
@@ -1,20 +1,41 @@
...
       if (this.isOpen()) {
         e.preventDefault();
         this.hide();
       }
     });
+
+    this.focusTrap = createFocusTrap('#drawer', { allowOutsideClick: true });
+  }
+
+  resizeHandler(e) {
```
askvortsov1, line 28: When are these resizes actually an issue? I can imagine changes in height easily (keyboard open), but why would we have changes in width?

js/src/common/utils/Drawer.js
```diff
@@ -41,7 +62,10 @@ export default class Drawer {
 
     const $app = $('#app');
 
-    if (!$app.hasClass('drawerOpen')) return;
+    this.focusTrap.deactivate();
+    window.removeEventListener('resize', this.throttledResizeHandler);
```
askvortsov1, line 66: Are we calling this onremove too? Or just on hide?

js/src/common/utils/Drawer.js
```diff
@@ -50,7 +74,7 @@ export default class Drawer {
 
     $app.removeClass('drawerOpen');
 
-    if (this.$backdrop) this.$backdrop.remove();
+    this.$backdrop.remove?.();
```
askvortsov1, line 77: Nice

askvortsov1 (changes requested):

js/src/common/components/ModalManager.js
```diff
@@ -1,40 +1,72 @@
 import Component from '../Component';
+import { createFocusTrap } from '../utils/focusTrap';
+import { tabbable } from 'tabbable';
+
+window.tabbable = tabbable;
```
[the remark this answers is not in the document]
davwheat: No, this was from debugging -- I just totally forgot to remove it.

js/src/common/components/ModalManager.js
```diff
@@ -1,40 +1,72 @@
...
       <div className="ModalManager modal fade">
-        {modal
-          ? modal.componentClass.component({
-              ...modal.attrs,
-              animateShow: this.animateShow.bind(this),
-              animateHide: this.animateHide.bind(this),
-              state: this.attrs.state,
-            })
-          : ''}
+        {!!modal &&
```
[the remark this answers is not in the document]
davwheat: No, `!undefined === true`, meaning that `!!undefined` would be `false`.

Mithril will attempt to render the result of `&&` as a text node if the left hand side resolves to false but is not actually a boolean.

For example:
```js
return <>{0 && <MyElement />}</>
```

Would render a '0' into the DOM.

I've used `!!` here to maintain consistency with other areas of core where we use `&&` to conditionally render components. I've been swapping out ternaries with this method as it's slightly more efficient.

js/package.json
```diff
@@ -10,6 +10,7 @@
     "color-thief-browser": "^2.0.2",
     "dayjs": "^1.10.4",
     "expose-loader": "^1.0.3",
+    "focus-trap": "^6.6.0",
```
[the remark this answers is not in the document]
davwheat: Seems to be fine for me:

image

js/src/common/components/ModalManager.js
```diff
@@ -1,40 +1,72 @@
...
+            animateShow: this.animateShow.bind(this),
+            animateHide: this.animateHide.bind(this),
+            state: this.attrs.state,
+          })}
       </div>
     );
   }
 
+  /**
+   * @param {import('mithril').VnodeDOM<{}, this>} vnode
```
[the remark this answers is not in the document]
davwheat: I'm a little ahead of you on this one! I made a TS conversion PR a couple of days ago :P

https://github.com/flarum/core/pull/3007

js/src/common/utils/Drawer.js
davwheat: Ooh, I didn't realise that existed! Thanks for pointing it out!

js/src/common/utils/Drawer.js
davwheat: See: https://github.com/flarum/core/pull/3018/commits/438577dae7a79f78cd2e6481d112340af5870917

The focus would end up staying trapped in the header elements if the drawer was opened, then the window was resized to be larger. There'd then be no way to close the drawer (and thus escape the focus trap) without making the window smaller again.

js/src/common/utils/Drawer.js
davwheat: I think only on hide should be fine. The Drawer class isn't a component, so we don't have access to the `onremove` lifecycle hook here, and, as such, this class never gets "recreated" in the same way mithril components can.

js/src/common/components/ModalManager.js
davwheat: I thought about that, but thought the added code needed to check if the exception was the one we'd expect wasn't worth it.

JS doesn't have error classes like PHP and other languages, meaning the only way to catch this specific error would be comparing the error message against a constant.

image

js/src/common/utils/Drawer.js
datitisev: This changes the logic. You want to check for $backdrop, the current code checks that the `remove` button is truthy, which will error if `$backdrop` is undefined

image

davwheat: Now I look at it, why do we need this at all?

From what I can see, after you remove the element with jQuery, the variable doesn't get cleared.

The only scenario would be if `hide` was somehow called before `show`. Is that even possible?

```js
let x = $('<div />').appendTo('body')
!!x // true
x.remove()
!!x // true
x.remove()
// ...
```

datitisev: I'd still keep the check just in case. 🤷 No reason not to have it, right?

davwheat: True. I've combined the two as if the value is anything other than `undefined`, we get an error if we only have a `?` after `$backdrop.`

js/src/common/components/ModalManager.js
davwheat: I'm not sure I understand what you're saying. Are you referring to `this.modalShown = false;` being called after removing the `in` class?

jQuery shouldn't ever cause a redraw, unless you're using it to call a redraw of course, or after any interaction event is fired , and subsequently handled, on a mithril-mounted DOM node.

js/src/common/utils/Drawer.js
davwheat: I discovered, thanks to a kind person in `#extend` in Discord, this: https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/onchange

No need for resize handlers! We can detect when a media query changes state in JS!

```ts
const phoneMaxWidth = document.documentElement.getPropertyValue('--screen-phone-max')
const drawerAailableMediaQuery = window.matchMedia(`(max-width: ${phoneMaxWidth})`);

drawerAailableMediaQuery.addEventListener('change', (e) => {
  if (e.matches) {
    // Is phone
  } else {
    // Is not phone
  }
})
```

js/src/common/components/ModalManager.js
askvortsov1: > JS doesn't have error classes like PHP and other languages

Never thought I'd see you admit an upside of PHP :wink:

js/src/common/utils/Drawer.js
askvortsov1: Isnt that also effectively a resize handler, but one that only responds to the exact change we care about?

js/src/common/components/ModalManager.js
davwheat: Never ever mention it again. 😄

js/src/common/utils/Drawer.js
```diff
@@ -1,28 +1,72 @@
...
+    document.getElementById('content').addEventListener('click', (e) => {
       if (this.isOpen()) {
         e.preventDefault();
         this.hide();
       }
     });
+
+    this.appElement = document.getElementById('app');
+    this.focusTrap = createFocusTrap('#drawer', { allowOutsideClick: true });
+    this.drawerAailableMediaQuery = window.matchMedia(
```
datitisev, line 31:
```suggestion
    this.drawerAvailableMediaQuery = window.matchMedia(
```

js/src/common/components/ModalManager.js
datitisev: Hm I don't like the idea of catching all errors and not just the one we're trying to avoid spamming...

js/src/forum/utils/KeyboardNavigatable.ts
```diff
@@ -51,20 +51,35 @@ export default class KeyboardNavigatable {
...
-   * This will be triggered by the Return and Tab keys..
+   * This will be triggered by the Return key.
    */
   onSelect(callback: KeyboardEventHandler): KeyboardNavigatable {
     const handler: KeyboardEventHandler = (e) => {
       e.preventDefault();
       callback(e);
     };
 
-    this.callbacks.set(9, handler);
```
datitisev, line 62: This is technically a breaking change as it changes behavior, is it not? Not sure how we want to handle it, if at all.

datitisev (approved):
Not tested locally, but code looks good aside from a few comments.

datitisev (changes requested):
I didn't mean to approve.

js/src/forum/utils/KeyboardNavigatable.ts
davwheat: It is, yes. We *could* add a second param to override the original behaviour, which might be a better idea.

datitisev: I'm in favor of replacing the second param with a diff method for 2.0. Just want to void breaking outside of major releases.

davwheat:
YOU WILL GET CO-AUTHORSHIP OF THAT COMMIT IF IT IS THE LAST THING I DO

SychO9 (approved):

davwheat:
Rebase in progress...

@datitisev you still have "requested changes" marked on this PR -- I think I addressed your comment about a breaking change, but I'll re-request your review and you can either approve, dismiss the request, or request other changes.

js/src/common/components/ModalManager.tsx
```diff
@@ -13,7 +15,14 @@ interface IModalManagerAttrs {
  * overwrite the previous one.
  */
 export default class ModalManager extends Component<IModalManagerAttrs> {
-  view() {
+  protected focusTrap: FocusTrap | undefined;
+
+  /**
+   * Whether a modal is currently shown by this modal manager.
+   */
+  protected modalShown: boolean = false;
```
davwheat, line 23: /cc reviewers @askvortsov1 @SychO9 

What are your opinions on the accessibility modifiers here? Should these be public, private or protected? I went with protected for now, as that gives the best of both worlds, but I do wonder if it should be public instead.

SychO9: I prefer it protected, I don't think we'd be breaking any extensions either.

davwheat: The only thing I thought was if an extension closes a modal without using `animateHide()`, then `modalShown` won't be updated, and the focus trap will still exist and be active.

SychO9: how would that happen ?

davwheat: If they directly called `$('.ModalManager').modal('hide')`, or whatever the right selector is. I can't actually think of any reason why they'd do it, so I'm fine with `protected`.

SychO9: yea, there are so many ways extension devs could do things the wrong way, I don't think we can ever do anything about that except have good docs explaining best practices.

askvortsov1: Let's keep it as protected. We offer a public API for closing the modal, anything else is explicitly going outside of that public API, and therefore not guarunteed not to break.

datitisev (approved):

PR #3021, opened 2021-08-12
[A11Y] Replace invalid link elements with buttons

### LINKED ISSUE #3519: [A11Y] Convert empty links to button elements
image

davwheat:
**Fixes flarum/framework#3519**

- Replace links with no or dead hrefs with `<button>` elements
- Add new `.Button--reset` class which removes all native styles from buttons
- Don't use `.box-shadow()` mixin. It's unneeded and bloats the CSS.

Anywhere I've missed? Should the Link be removed from "dead" authors?

**Screenshot**

image

**Confirmed**


askvortsov1 (commented):
To me, `reset` implies that it's using browser native button styling. Is there a convention behind this naming? Maybe Button--text?

SychO9:
I haven't reviewed yet, but we have `Button--link` and `Button--text`, don't one of these do the job ?

davwheat:
Unfortunately not. `Button--link` keeps the button padding, and `Button--text` results in incorrect vertical aligning and undesired underlining on hover/focus (which doesn't match the previous behaviour with `<a>` without `href` attributes.

We definitely don't want the underlining in some places (post scrubber for example), but other areas I'd be open to ("links" in modals).

SychO9:
Ah I see, makes sense.

btw, I believe we can do `all: unset` here :D found out about it recently!

askvortsov1 (approved):

dsevillamartin:
Not tested the changes or really looked at CSS, mainly focused on JS. 
My thoughts are basically on cleaning this up rather than anything else -

We could potentially have a <`Component`/`Tag`> variable that switches between `Link` and `span` to avoid this code duplication. Don't know if that'd be better or worse - potentially create a separate component that chooses the tag (including `<Button>` reset for `onclick`)? 🤔 

That way it would avoid writing `Button Button--reset` every time (unless we explicitly want it) or handling falsy `href`/`onclick`'s, as the component would do that too.

js/src/forum/components/PostMeta.js
```diff
@@ -29,9 +30,9 @@ export default class PostMeta extends Component {
 
     return (
       <div className="Dropdown PostMeta">
-        <a className="Dropdown-toggle" onclick={selectPermalink} data-toggle="dropdown">
+        <Button className="Button Button--reset Dropdown-toggle" onclick={selectPermalink} data-toggle="dropdown">
```
SychO9, line 33: the timestamp looks colored at all times, can we instead use `Button--text` here and just add a `vertical-align: bottom;` to `.Button--text` ? it's a lot more consistent.

js/src/forum/components/LogInModal.js
```diff
@@ -125,11 +125,17 @@ export default class LogInModal extends Modal {
...
+        <Button class="Button Button--reset" onclick={this.forgotPassword.bind(this)}>
+          {app.translator.trans('core.forum.log_in.forgot_password_link')}
+        </Button>
       </p>,
 
       app.forum.attribute('allowSignUp') ? (
-        <p className="LogInModal-signUp">{app.translator.trans('core.forum.log_in.sign_up_text', { a: <a onclick={this.signUp.bind(this)} /> })}</p>
+        <p className="LogInModal-signUp">
+          {app.translator.trans('core.forum.log_in.sign_up_text', {
+            a: <Button class="Button Button--reset" onclick={this.signUp.bind(this)} />,
```
SychO9, line 136: these look weird as well, colored with no effect on hover, the underline hovering effect of the link was nice.
perhaps we should add a new `Button--textlink` that would mimic an exact link?

SychO9 (commented):
The `Button--reset` looks fine with the scrubber buttons, but odd with the rest.

SychO9 (changes requested):
Already left a review 👼🏼

askvortsov1:

PR #3027, opened 2021-08-15
[A11Y] Explicitly state `aria-hidden` value; hide icons from screenreaders

davwheat:

- Explicitly state the value of `aria-hidden` (needed for Talkback on Android)
- Hide icons from screenreaders as these are purely visual elements

🤷 

**Confirmed**

  - tested on Android 11

### DIFF
--- a/js/src/admin/components/UserListPage.tsx
+++ b/js/src/admin/components/UserListPage.tsx
@@ -286,7 +286,7 @@ export default class UserListPage extends AdminPage {
 
           return (
             <div class="UserList-email" key={user.id()} data-email-shown="false">
-              <span class="UserList-emailAddress" aria-hidden onclick={() => setEmailVisibility(true)}>
+              <span class="UserList-emailAddress" aria-hidden="true" onclick={() => setEmailVisibility(true)}>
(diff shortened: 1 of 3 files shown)


SychO9 (approved):

askvortsov1 (approved):
The string value suffices and we don't need a raw boolean, right?

EDIT: Actually, I think I would prefer using a boolean regardless. It's a bit more explicit as to the intended meaning.

davwheat:
Sure. Mithril converts the boolean to a string when rendering to the DOM, but both ways are fine. I think the boolean might actually be slightly smaller in terms of JS bundle :P

davwheat:
Actually it seems this will not work. Mithril converts the `true` boolean into an empty attribute value as opposed to the string `"true"`, meaning this doesn't fix `aria-hidden` on Android.

askvortsov1 (approved):
Fair enough. Maybe we should open an issue with Mithril?

davwheat:
No, its intended behaviour.

That way you can set an attribute depending on a conditional, like `checked={this.state.checked}`.

PR #3032, opened 2021-08-18
Remove format hook on commit; update CI to only build with valid formatting

davwheat:
**Depends on https://github.com/flarum/action-build/pull/7**

Here's one that @luceos will like. 😄 

We've had discussions internally, and I think the majority of us want to see this gone. It doesn't work properly on Windows due to Windows being annoying, and it's just pretty slow in general.

- @luceos:
  - "if you commit often it just fucks up your work process and pulls you out of your zone"
  - "can someone please take a look at prettier so that it doesn't run if there are no JS changes, please?! i am literally going crazy"
  - "i see no reason to reduce available developer time by forcing these hooks onto them"
- @datitisev:
  - "this damn husky precommit command is annoying af when rebasing"
  - "prevents fast commits when trying to separate work done"
- @SychO9:
  - "I think we should disable the automatic hook, and when the workflow fails, we can locally manually run `npm run format`"
- @tankerkiller125:
  - "I was always of the opinion that all of that stuff should just run as Github CI actions, I don't remember why but that idea was shot down at the time"
- @davwheat:
  - "i am in agreement [with the use of GitHub Actions for formatting instead.] we've shifted to Actions for a lot of stuff more recently i think"

Our CI already performs format checking on all our PRs, so if formatting is not correct we'll be sure to know. This git hook just slows down our development and causes added frustration when creating new commits, especially for those of us who use IDE extensions that format with Prettier.

----

**In the rare chance we commit unformatted JS to master,** our build CI shouldn't run with the changes from this PR.

- If prettier fails, no build job will run.
- If prettier succeeds on `master`, JS will be built and committed
- If prettier succeeds on PRs/other branches, JS will be built but not committed

🤷


luceos:
Whatever you did. Merge it, just .. merge it.

tankerkiller125 (approved):

davwheat:
Need to fix the CI job first! :P

tankerkiller125:
I mean... The CI job had a check mark which indicated it was working to me.... But hey, regardless I have faith that you'll have it fixed and working right :P

davwheat:
Weirdly, GitHub fails silently on invalid workflow files which is a little worrying.

It's all fixed now, though!

askvortsov1 (approved):
Begone!!!!

Can you also add the new build script to the flarum-cli skeleton?

askvortsov1 (approved):
Still good to merge IMO

PR #3037, opened 2021-08-20
Pass filter params to getApiDocument

imorland:
Companion PR to https://github.com/flarum/tags/pull/141

Currently, any filter params that are passed toForum\ Content\Index are ignored. As discussed with @askvortsov1 here is a solution...

### DIFF
--- a/src/Forum/Content/Index.php
+++ b/src/Forum/Content/Index.php
@@ -68,6 +68,7 @@ class Index
         $sort = Arr::pull($queryParams, 'sort');
         $q = Arr::pull($queryParams, 'q');
         $page = max(1, intval(Arr::pull($queryParams, 'page')));
+        $filters = Arr::pull($queryParams, 'filter', []);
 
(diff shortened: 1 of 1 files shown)


src/Forum/Content/Index.php
```diff
@@ -77,6 +78,8 @@ public function __invoke(Document $document, Request $request)
             'page' => ['offset' => ($page - 1) * 20, 'limit' => 20]
         ];
 
+        $params['filter'] = array_merge($filters, $params['filter']);
```
askvortsov1, line 81: Since $params['filter'] is empty here, let's just set `'filter' => $filters` in the `$params = ...` code above?

askvortsov1 (changes requested):

src/Forum/Content/Index.php
imorland: Done :)

askvortsov1 (approved):

PR #3038, opened 2021-08-20
Move email confirmation to POST request

### LINKED ISSUE #2969: Confirmation links are only valid once
## Bug Report

**Current Behavior**
When an confirmation link has been opened once, subsequent attempts will result in an error such as "This confirmation link has already been used or is invalid.". This is still confusing to some users. Especially since some email clients such as Outlook Online will attempt to "preview" links by opening them (which is **utter insanity**, but that's another matter), so when the user eventually clicks the link, they will always see a (suspicious looking) error message.

**Steps to Reproduce**
1. Have email client that previews links (Outlook Online)
2. Start email verification
3. Click on link in mail
4. See error message in browser

**Environment**
- Flarum version: 1.0.4
- Browser: insane mail client

**Possible Solution(s)**
- Ideal: Nuke Microsoft datacenters and ban their decision-makers from ever working in the tech sector again
- If there's a way to format the mail in a way to stop previews, that would be best
- Two-step verification? Must open mail and click on a button
- Don't show an error message, redirect directly to forum

**Additional Context**
This could be considered a sequel of #1337 :)

Thanks!

datitisev:
**Fixes #2969**

- Create blade view for GET request
- Keep POST request at same URL (don't want to change to /confirm because that could be breaking if another ext uses that route)

- Do we want more blade views or to move this to JS routes?

**Screenshot**
image

**Confirmed**

### DIFF
--- a/locale/core.yml
+++ b/locale/core.yml
@@ -545,6 +545,11 @@ core:
 
   # Translations in this namespace are used in views other than Flarum's normal JS client.
   views:
+    # Translations in this namespace are displayed by the Confirm Email interface.
+    confirm_email:
+      submit_button: => core.ref.confirm_email
+      text: Click the button below to confirm your account's email.
+      title: => core.ref.confirm_email
 
     # Translations in this namespace are displayed by the basic HTML content loader.
     content:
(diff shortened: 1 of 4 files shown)


askvortsov1 (approved):
Not tested locally, but code looks good

SychO9 (approved):
Looks good, tested locally.

PR #3040, opened 2021-08-21
Replace username with display name in more places

clarkwinkelmann:
Replaces instances of `username` which have not been converted to use display name yet.

The following were fixed:

- Confirmation and error alert box after deleting a user
- Display name in notification email TO header
- Author in discussion nojs view

I don't think any of these had an open issue.

For the alert change I have kept the existing name `username` for the translation variable. I think that's what we have already done in all previous instances.

**Confirmed**

### DIFF
--- a/js/src/forum/utils/UserControls.js
+++ b/js/src/forum/utils/UserControls.js
@@ -124,7 +124,7 @@ export default {
    * @param {string} type
    */
   showDeletionAlert(user, type) {
-    const { username, email } = user.data.attributes;
+    const { displayName: username, email } = user.data.attributes;
     const message = {
       success: 'core.forum.user_controls.delete_success_message',
       error: 'core.forum.user_controls.delete_error_message',
--- a/js/src/forum/utils/UserControls.js
+++ b/js/src/forum/utils/UserControls.js
@@ -124,13 +124,18 @@ export default {
    * @param {string} type
    */
   showDeletionAlert(user, type) {
-    const { displayName: username, email } = user.data.attributes;
     const message = {
       success: 'core.forum.user_controls.delete_success_message',
       error: 'core.forum.user_controls.delete_error_message',
     }[type];
 
-    app.alerts.show({ type }, app.translator.trans(message, { username, email }));
+    app.alerts.show(
+      { type },
+      app.translator.trans(message, {
+        user,
+        email: user.email(),
+      })
+    );
   },
 
   /**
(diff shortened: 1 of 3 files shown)


js/src/forum/utils/UserControls.js
```diff
@@ -124,7 +124,7 @@ export default {
    * @param {string} type
    */
   showDeletionAlert(user, type) {
-    const { username, email } = user.data.attributes;
+    const { displayName: username, email } = user.data.attributes;
```
askvortsov1, line 127: Shouldn't this be `username: displayName`?

clarkwinkelmann: I tried both and this one works :joy: 

The first name (`displayName`) is the one extracted from `user.data.attributes` and it's then renamed as `username` to keep the remaining of the code like it was before.

I did a few searches to find the correct syntax, this article explains the same https://wesbos.com/destructuring-renaming

I did consider refactoring the code to use a more readable syntax, but then it's two lines that are mostly redundant with each other...

clarkwinkelmann: Alternate solution, I could change this line to just extract `displayName` and edit line 133 to be `username: displayName`, which might be easier to read?

askvortsov1: I like that solution. Although you are right: I got this mixed up with graphql aliases.

SychO9:
This can be considered part of **https://github.com/flarum/core/issues/1734**

askvortsov1 (approved):
LGTM, only nit is in the comment thread.

SychO9 (approved):

js/src/forum/utils/UserControls.js
clarkwinkelmann: I have refactored those lines. Actually I have no idea why a double spread operator was used on the model internal data instead of retrieving the values from the model in the proper way. I'm now passing the user to the translator directly, so the translator can take care of defining `username`. And I just extract the email separately.

js/src/forum/utils/UserControls.js
```diff
@@ -124,13 +124,18 @@ export default {
...
     const message = {
       success: 'core.forum.user_controls.delete_success_message',
       error: 'core.forum.user_controls.delete_error_message',
     }[type];
 
-    app.alerts.show({ type }, app.translator.trans(message, { username, email }));
+    app.alerts.show(
+      { type },
+      app.translator.trans(message, {
+        user,
```
SychO9, line 135: isn't this meant to be `username: user.displayName()` ?

clarkwinkelmann: It's now using our "magic" translator behavior that automatically provides a `username` variable with the output of the `username()` helper function if you pass a user (which actually returns the display name). This is actually more consistent with the other uses of the translator.

SychO9: oh wow, I actually wasn't aware it was a thing lol

PR #3042, opened 2021-08-21
Add extra error handling for avatar file size & large payload

### LINKED ISSUE #2888: Avatar upload fails with large file sizes
## Bug Report

**Current Behavior**
When uploading a large file, the form data isn't sent (at least not properly? different browsers show different things) and the server returns with 
> POST https://flarum-new.localhost/api/users/1/avatar 422
> The :attribute field is required.

**Steps to Reproduce**
1. Go to user profile
2. Upload a large image (I used https://images.pexels.com/photos/842711/pexels-photo-842711.jpeg?cs=srgb&dl=pexels-christian-heitz-842711.jpg&fm=jpg)

**Expected Behavior**
No error.

**Screenshots**
<details>
<summary>Large Image</summary>

image

Chromium browsers do seem to show the request data, but the error still occurs.
image

</details>

<details>
<summary>Small Image</summary>

image

Chromium shows the same thing.
image

</details>

**Additional Context**
I found this while trying to debug why `fof/profile-image-crop` failed with large images. I don't know if this is a limitation of FormData or an issue with Mithril request or our code...

> I found the issue with profile-image-crop and large images.
It's not a problem with the ext but with core. The actual error escapes me.
Looking at the request, the FormData isn't passed for large file sizes (both with ext enabled & disabled) - and I cannot figure out why.
i tested this by adding my own data to FormData in core for image upload, and using an image that fails with both ext enabled & disabled
the reason it sometimes fails with it enabled but not disabled is because the resize/crop sometimes makes the image bigger

dsevillamartin:
**Fixes flarum/framework#2888**

- Add separate error messages during upload for `UPLOAD_ERR_INI_SIZE` & `UPLOAD_ERR_FORM_SIZE`, and `UPLOAD_ERR_PARTIAL`
- Add frontend handling of 413 status code (Payload Too Large)

- Do we perhaps want to say "failed to upload" for the other errors and just "required" for `UPLOAD_ERR_NO_FILE`? That'd make more sense as only NO_FILE is a required issue, and the others are internal so we can say failed.
- This file logic, validation and/or error handling should probably be extracted so that other areas & extensions can use it... out of scope for this however.

**Screenshot**
image

**Confirmed**

### DIFF
--- a/src/User/AvatarValidator.php
+++ b/src/User/AvatarValidator.php
@@ -33,15 +33,15 @@ class AvatarValidator extends AbstractValidator
         $error = $file->getError();
 
         if ($error !== UPLOAD_ERR_OK) {
-            if ($error == UPLOAD_ERR_INI_SIZE || $error == UPLOAD_ERR_FORM_SIZE) {
+            if ($error === UPLOAD_ERR_INI_SIZE || $error === UPLOAD_ERR_FORM_SIZE) {
                 $this->raise('file_too_large');
             }
 
-            if ($error == UPLOAD_ERR_PARTIAL) {
-                $this->raise('file_uploaded');
+            if ($error === UPLOAD_ERR_NO_FILE) {
+                $this->raise('required');
             }
 
-            $this->raise('required');
+            $this->raise('file_uploaded');
         }
     }
 
--- a/src/User/AvatarValidator.php
+++ b/src/User/AvatarValidator.php
@@ -41,7 +41,7 @@ class AvatarValidator extends AbstractValidator
                 $this->raise('required');
             }
 
-            $this->raise('file_uploaded');
+            $this->raise('file_upload_failed');
         }
     }
 
--- a/src/User/AvatarValidator.php
+++ b/src/User/AvatarValidator.php
@@ -92,9 +92,4 @@ class AvatarValidator extends AbstractValidator
     {
         return ['jpg', 'png', 'bmp', 'gif'];
     }
-
-    protected function getRules()
-    {
-        return ['avatar' => 'required|file'];
-    }
 }
--- a/src/User/AvatarValidator.php
+++ b/src/User/AvatarValidator.php
@@ -30,7 +30,17 @@ class AvatarValidator extends AbstractValidator
 
     protected function assertFileRequired(UploadedFileInterface $file)
     {
-        if ($file->getError() !== UPLOAD_ERR_OK) {
+        $error = $file->getError();
+
+        if ($error !== UPLOAD_ERR_OK) {
+            if ($error == UPLOAD_ERR_INI_SIZE || $error == UPLOAD_ERR_FORM_SIZE) {
+                $this->raise('file_too_large');
+            }
+
+            if ($error == UPLOAD_ERR_PARTIAL) {
+                $this->raise('file_uploaded');
+            }
+
             $this->raise('required');
         }
     }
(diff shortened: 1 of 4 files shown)


src/User/AvatarValidator.php
```diff
@@ -82,4 +92,9 @@ protected function getAllowedTypes()
     {
         return ['jpg', 'png', 'bmp', 'gif'];
     }
+
+    protected function getRules()
+    {
+        return ['avatar' => 'required|file'];
+    }
```
SychO9, line 99: Don't we manually check for these ?

Actually we only manually check that it's required, I don't think we check that it's an actual file 🤔 so it would make sense, is this correct?

askvortsov1:
Should we add a translation key for the attribute? The screenshot shows `:attribute` which doesn't seem right.

SychO9:
> Should we add a translation key for the attribute? The screenshot shows `:attribute` which doesn't seem right.

https://github.com/flarum/core/issues/2933

askvortsov1 (approved):

src/User/AvatarValidator.php
datitisev: Oh this is code from some testing j was doing... This change shouldnt be here.

datitisev:
Any thoughts on

> Do we perhaps want to say "failed to upload" for the other errors and just "required" for UPLOAD_ERR_NO_FILE? That'd make more sense as only NO_FILE is a required issue, and the others are internal so we can say failed.

To clarify, that'd change it to show required message when no file is provided, and upload failure message on the other errors, which would be clearer as to the issue.

src/User/AvatarValidator.php
SychO9: Alright then, I've logged what I was referring to separately so as not to go out of scope for this PR.

SychO9:
> To clarify, that'd change it to show required message when no file is provided, and upload failure message on the other errors, which would be clearer as to the issue.

I would be in favor.

dsevillamartin:
**EDIT:** Just saw you made an issue (#3044). 😂 

After further testing, not providing an avatar seems to result in 500 because there's `$attributes['avatar']` is not an `UploadedFileInterface`... not sure exactly when the error would be `UPLOAD_ERR_NO_FILE` 🤷‍♂️ 

Not sure if that should be addressed in this PR - to keep the code clean we'd need to change signature of `assertFileRequired`, which would be potentially breaking (not sure if we'd classify that as breaking tbh) to allow null

SychO9:
> After further testing, not providing an avatar seems to result in 500 because there's `$attributes['avatar']` is not an `UploadedFileInterface`... not sure exactly when the error would be `UPLOAD_ERR_NO_FILE` 
> 
> Not sure if that should be addressed in this PR - to keep the code clean we'd need to change signature of `assertFileRequired`, which would be potentially breaking (not sure if we'd classify that as breaking tbh) to allow null

Let's deal with that separately https://github.com/flarum/core/issues/3044, essentially we'd just want to introduce a new `assetFile` method that accepts any type as parameter and checks that the parameter is an instance of the interface.

datitisev:
Anything missing with this PR? Looking back at it the locale key should probably be changed from `file_uploaded` to `file_upload_failed`... the former doesn't make sense given the context.

SychO9 (approved):
I don't remember, it looks good though, let's just change that key and merge.

askvortsov1 (approved):

PR #3057, opened 2021-08-25
Preload FontAwesome, JS and CSS

### LINKED ISSUE #2741: Implement preloading and prefetching
## Feature request

Preloading and prefetching could result in ~100ms-200ms off the top of load times on Flarum forums.

We could preload:
- JS
- CSS
- FontAwesome font files

We could preconnect to:
- Twemoji CDN

davwheat:
**Fixes #2741**

- Adds support for preloads in `Document`
- Adds `preload()` and `preloads()` methods to Frontend extender
- Adds default preloads for FontAwesome fonts
- Adds default preloads for core JS and CSS assets

The benefits of preloading FontAwesome are that these would normally only be fetched when CSS parsing is complete, which is often.

Preloading JS and CSS only makes a difference if we're preloading FontAwesome already. Browsers classify preloads as higher priority requests, meaning they would stall downloads of our JS/CSS until these are done. Adding these as preloads prevents this, and they are all fetched at once.

**Overall, preloading results in a consistently faster page load time.** Almost always 100-150ms faster, sometimes reaching 200ms+. If/when themes and other extensions begin preloading their assets, this could increase further.

My failure to write/comprehend PHP and my blatant disregard for helpers and simplicity.

**Screenshot**

image

## With preloads

image

## Without preloads

image

**Confirmed**

**Required changes:**

- [ ] Related documentation PR:

### DIFF
--- a/src/Extend/Frontend.php
+++ b/src/Extend/Frontend.php
@@ -281,11 +281,11 @@ class Frontend implements ExtenderInterface
         $container->resolving(
             "flarum.frontend.$this->frontend",
             function (ActualFrontend $frontend, Container $container) {
-                $frontend->content(ContainerUtil::wrapCallback(function (Document $document) {
+                $frontend->content(function (Document $document) use ($container) {
                     foreach ($this->preloads as $preload) {
-                        $document->preloads[] = is_callable($preload) ? $preload($document) : $preload;
+                        $document->preloads[] = is_callable($preload) ? ContainerUtil::wrapCallback($preload, $container) : $preload;
                     }
-                }, $container));
+                });
             }
         );
     }
--- a/src/Extend/Frontend.php
+++ b/src/Extend/Frontend.php
@@ -34,7 +34,7 @@ class Frontend implements ExtenderInterface
     private $routes = [];
     private $removedRoutes = [];
     private $content = [];
-    private $preloads = [];
+    private $preloadArrs = [];
 
     /**
      * @param string $frontend: The name of the frontend.
@@ -154,7 +154,7 @@ class Frontend implements ExtenderInterface
      */
     public function preloads($preloads): self
     {
-        $this->preloads = array_merge($this->preloads, (array) $preloads);
+        $this->preloadArrs[] = $preloads;
 
         return $this;
     }
@@ -274,7 +274,7 @@ class Frontend implements ExtenderInterface
 
     private function registerPreloads(Container $container): void
     {
-        if (empty($this->preloads)) {
+        if (empty($this->preloadArrs)) {
             return;
         }
 
@@ -282,8 +282,8 @@ class Frontend implements ExtenderInterface
             "flarum.frontend.$this->frontend",
             function (ActualFrontend $frontend, Container $container) {
                 $frontend->content(function (Document $document) use ($container) {
-                    foreach ($this->preloads as $preloads) {
-                        $preloads = is_callable($preloads) ? ContainerUtil::wrapCallback($preloads, $container)($document) : $preloads;
+                    foreach ($this->preloadArrs as $preloadArr) {
+                        $preloads = is_callable($preloadArr) ? ContainerUtil::wrapCallback($preloadArr, $container)($document) : $preloadArr;
                         $document->preloads = array_merge($document->preloads, $preloads);
                     }
                 });
--- a/src/Extend/Frontend.php
+++ b/src/Extend/Frontend.php
@@ -282,8 +282,9 @@ class Frontend implements ExtenderInterface
             "flarum.frontend.$this->frontend",
             function (ActualFrontend $frontend, Container $container) {
                 $frontend->content(function (Document $document) use ($container) {
-                    foreach ($this->preloads as $preload) {
-                        $document->preloads[] = is_callable($preload) ? ContainerUtil::wrapCallback($preload, $container)($document) : $preload;
+                    foreach ($this->preloads as $preloads) {
+                        $preloads = is_callable($preloads) ? ContainerUtil::wrapCallback($preloads, $container)($document) : $preloads;
+                        $document->preloads = array_merge($document->preloads, $preloads);
                     }
                 });
             }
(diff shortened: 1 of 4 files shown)


src/Extend/Frontend.php
```diff
@@ -236,6 +272,24 @@ function (ActualFrontend $frontend, Container $container) {
...
+        if (empty($this->preloads)) {
+            return;
+        }
+
+        $container->resolving(
+            "flarum.frontend.$this->frontend",
+            function (ActualFrontend $frontend, Container $container) {
+                $frontend->content(ContainerUtil::wrapCallback(function (Document $document) {
+                    foreach ($this->preloads as $preload) {
+                        $document->preloads[] = is_callable($preload) ? $preload($document) : $preload;
```
davwheat, line 286: Am I right in thinking that wrapping with `wrapCallback` isn't needed here as it's already inside one of these?

SychO9: The content callback arg doesn't need to be wrapped in it. `wrapCallback` makes sure invokable classes work and are able to use dependency injection, it makes no difference to closures (callback functions) so it's not needed there.

davwheat: Gotcha. That should be right now :P

src/Extend/Frontend.php
```diff
@@ -236,6 +272,24 @@ function (ActualFrontend $frontend, Container $container) {
...
+        if (empty($this->preloads)) {
+            return;
+        }
+
+        $container->resolving(
+            "flarum.frontend.$this->frontend",
+            function (ActualFrontend $frontend, Container $container) {
+                $frontend->content(function (Document $document) use ($container) {
+                    foreach ($this->preloads as $preload) {
+                        $document->preloads[] = is_callable($preload) ? ContainerUtil::wrapCallback($preload, $container) : $preload;
```
SychO9, line 286:
```suggestion
                        $document->preloads[] = is_callable($preload) ? ContainerUtil::wrapCallback($preload, $container)($document) : $preload;
```

askvortsov1 (approved):

askvortsov1:
Very excited to see this! Let's fix tests, then good to go IMO

src/Extend/Frontend.php
```diff
@@ -124,11 +126,45 @@ public function content($callback): self
...
+     *   ]
+     * ]);
+     * ```
+     *
+     * @param callable|array $preloads
+     * @return self
+     */
+    public function preloads($preloads): self
+    {
+        $this->preloads = array_merge($this->preloads, $preloads);
```
SychO9, line 157: We can cast it to an array which will transform a callable into an item inside a new array and keep arrays as arrays.

```suggestion
        $this->preloads = array_merge($this->preloads, (array) $preloads);
```

davwheat:
Weirdly, it seems to work fine locally.

davwheat:
I don't have time to work on core for the next few days, so if either of you have any ideas, please just push straight to the branch :)

src/Extend/Frontend.php
```diff
@@ -236,6 +272,24 @@ function (ActualFrontend $frontend, Container $container) {
...
+            return;
+        }
+
+        $container->resolving(
+            "flarum.frontend.$this->frontend",
+            function (ActualFrontend $frontend, Container $container) {
+                $frontend->content(function (Document $document) use ($container) {
+                    foreach ($this->preloads as $preload) {
+                        $document->preloads[] = is_callable($preload) ? ContainerUtil::wrapCallback($preload, $container)($document) : $preload;
+                    }
```
askvortsov1, line 287: This logic seems flawed. Shouldn't we call the callback here instead of just wrapping (since we're already in `resolving`), and flatten the preloads as we load them into the document?

askvortsov1 (changes requested):

src/Extend/Frontend.php
SychO9: we are calling the callback here

src/Extend/Frontend.php
```diff
@@ -236,6 +272,24 @@ function (ActualFrontend $frontend, Container $container) {
...
+                });
```
askvortsov1, line 288:
```suggestion
                    foreach ($this->preloads as $preloads) {
                        $preloads = is_callable($preloads) ? ContainerUtil::wrapCallback($preloads, $container)($document) : $preloads;
                        $document->preloads = array_merge($document->preloads, $preloads);
                    }
                });
```

askvortsov1:
@davwheat @SychO9 issue was with a cast when `preloads` was called. Fixed now, and changed naming to be more descriptive of structure.

src/Frontend/FrontendServiceProvider.php
```diff
@@ -54,6 +54,48 @@ public function register()
...
+                            'as' => 'font',
+                            'type' => 'font/woff2',
+                            'crossorigin' => ''
+                        ], [
+                            'href' => $filesystem->url('fonts/fa-regular-400.woff2'),
+                            'as' => 'font',
+                            'type' => 'font/woff2',
+                            'crossorigin' => ''
+                        ]
+                    ];
```
askvortsov1, line 72: Maybe this should be a singleton? Preferably with a naming like "additional preloads"?

askvortsov1 (changes requested):

src/Frontend/FrontendServiceProvider.php
davwheat: To sate my ~~desire~~ need to learn PHP, what do we gain by doing that?

tankerkiller125: By making this a singleton it would allow someone to say replace or otherwise remove these pre-loads if they wanted. For example if they replace the built-in FontAwesome with FontAwesome Pro.

askvortsov1: It's also clearer semantically, because we're not just including some random hard-coded values as part of implementation details.

askvortsov1 (approved):
Approving tentative on the singleton comment

tankerkiller125 (approved):
Also going to approve tenitive the singleton comment being fixed

askvortsov1 (approved):
LGTM

PR #3067, opened 2021-09-08
Use `author` filter parametzer instead of `q` with gambit to get a user's discussions on the `DiscussionsUserPage`

iPurpl3x:
Changed the way the request for listing discussions on the user profile is constructed: before a gambit was used with the `q` filter parameter, now the `author` parameter is used instead. This makes it possible for extensions to use other filter parameters for this request because if the `q` parameter is used, all the other filter parameters are ignored.

Checking if `DiscussionsUserPage` still works as expected.

**Confirmed**


iPurpl3x:
I had to recreate this PR because the git history was wrong and included other changes. Here is the correct PR : https://github.com/flarum/core/pull/3068

Issue #2934, opened 2021-06-22
Global banners


iPurpl3x:
## Feature Request

**Is your feature request related to a problem? Please describe.**
Yes, the problem is this: [comment from `fof/terms`](https://github.com/FriendsOfFlarum/terms/blob/e303f03ab91ea08f9dc6da087c98d46591df25b7/js/src/forum/addUpdateAlert.js#L34). The problem is that when it comes to very customized forums, there are a lot of pages where the banner has to be added explicitly.

**Describe the solution you'd like**
A detailed description of your proposed solution. Include:
- How the feature would work/behave
  Add something in core to make it possible to show these kind of banners globally. Just like alertEmailConfirmation. 
- Any potential drawbacks
  I don't think so

**Justify why this feature belongs in Flarum's core, rather than in a third-party extension**
It's an extensibility feature that would be beneficial to many users and developers. 

**Describe alternatives you've considered**
As said above, the alternative is to add the banner to each page manually. Which is difficult to maintain.

SychO9:
You're referring to the hero I presume ? if so the goal is to improve that with #2248

iPurpl3x:
More or less @SychO9, the here is one thing and your proposal would already improve the extensibility a lot!

But here I'm referring to those yellow banners/alerts that appear when needing to confirm the email on registration, or the one that appears when the terms have changed (fof/terms). There might be other cases but I cannot bring them to my mind right now...

SychO9:
Oh I see, yes that makes sense!

PR #3073, opened 2021-09-17
Add config.php option not to boot extensions

askvortsov1:
If the `boot_extensions` key is set to false, core will not boot any extensions in the backend, or run any initializers in the frontend. This allows admins to recover their forum without SSH access by disabling extension boot, going to the admin panel, and turning off offending extensions.

**Questions/Concerns for Reviewers**:

- Should we somehow signal to admins that this mode is enabled, possibly via alert? Otherwise, it might be confusing why extensions are indicated as enabled in the admin dashboard. This is challenging, as the alerts we currently have are in `content.blade.php`, and that view does not have access to payload/config.
- Should this require maintenance mode? If we disable all extensions (e.g. tags), some restricted discussions / content will immediately become public. This could lead to leaks via web scraping.

### DIFF
--- a/js/src/common/Application.js
+++ b/js/src/common/Application.js
@@ -168,7 +168,9 @@ export default class Application {
   }
 
   boot() {
-    this.initializers.toArray().forEach((initializer) => initializer(this));
+    if (this.data.attributes.bootExtensions) {
+      this.initializers.toArray().forEach((initializer) => initializer(this));
+    }
 
     this.store.pushPayload({ data: this.data.resources });
 
--- a/src/Api/Serializer/ForumSerializer.php
+++ b/src/Api/Serializer/ForumSerializer.php
@@ -77,6 +77,7 @@ class ForumSerializer extends AbstractSerializer
             'baseUrl' => $url = $this->url->to('forum')->base(),
             'basePath' => parse_url($url, PHP_URL_PATH) ?: '',
             'debug' => $this->config->inDebugMode(),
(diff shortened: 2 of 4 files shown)


SychO9 (commented):
Damn, completely forgot about this..

> Should we somehow signal to admins that this mode is enabled, possibly via alert? Otherwise, it might be confusing why extensions are indicated as enabled in the admin dashboard. This is challenging, as the alerts we currently have are in content.blade.php, and that view does not have access to payload/config.

Yes, we can do something similar to what we currently do with [email confirmation alerts](https://github.com/flarum/core/blob/eb4b18a979c7406cbf154a107662652d282fe415/js/src/forum/utils/alertEmailConfirmation.js#L11), we add a div and mount into it from the frontend. Hacky, but it's something we will improve with 2.0's structure refactor anyway: https://github.com/flarum/core/issues/2934.

> Should this require maintenance mode? If we disable all extensions (e.g. tags), some restricted discussions / content will immediately become public. This could lead to leaks via web scraping.

I believe so, not booting all extensions might lead to unexpected results depending on what/how extensions do what they do, so only admins should be able to access the app. Problem is, the maintenance mode is currently not allowing admins in either, so we have to investigate and fix that first: https://github.com/flarum/core/issues/1732.

luceos:
I like this idea: a) yes add a warning in both forum and admin b) enable maintenance too

I'd like to suggest to name the config key `safe_mode` though.

askvortsov1:

PR #3080, opened 2021-09-24
Filter composer icon array to only valid values

### LINKED ISSUE #3078: Invalid `composer.json` data in enabled or disabled extensions brick admin dashboard
## Bug Report

**Current Behavior**
With some extensions, the admin frontend is inaccessible.

One example seems to be `fof/spamblock`. This extension has no effect on v1.0.4, but renders the admin frontend inaccessible on v1.1.

This appears to relate to #3059.

image

**Steps to Reproduce**
1. Install `fof/spamblock`
2. Access the admin frontend

**Expected Behavior**

**Environment**
- Flarum version: dev-master
- Webserver: Apache
- Hosting environment: Docker
- PHP version: 8.0.0

SychO9:
**Fixes #3078**

If an extension has objects within the icon object in composer.json, it leads to the error reported. So we filter out the values to make sure that doesn't occur.

**Confirmed**

### DIFF
--- a/src/Extension/Extension.php
+++ b/src/Extension/Extension.php
@@ -283,6 +283,14 @@ class Extension implements Arrayable
     {
         $properties = $this->getIcon();
 
+        if (empty($properties)) {
+            return '';
+        }
+
+        $properties = array_filter($properties, function ($item) {
+            return ! is_array($item);
+        });
+
         unset($properties['name']);
 
         return implode(';', array_map(function (string $property, string $value) {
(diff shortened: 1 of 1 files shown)


src/Extension/Extension.php
```diff
@@ -283,6 +283,14 @@ public function getIconStyles(): string
     {
         $properties = $this->getIcon();
 
+        if (empty($properties)) {
+            return '';
+        }
+
+        $properties = array_filter($properties, function ($item) {
+            return ! is_array($item);
```
davwheat, line 291: Would it be better to check with `is_string()`?

SychO9: oh definitely 🤦🏼 

```suggestion
            return is_string($item);
```

davwheat (approved):

askvortsov1 (approved):

PR #3084, opened 2021-09-28
#2492 - Groups filtering & retrieve single endpoint

### LINKED ISSUE #2492: Groups GET API endpoint ignores parameters
## Bug Report

**Current Behavior**
No matter the GET params, `/api/groups` always return all groups. None of the usual parameters like `filter[q]` and `page` work. There is also no way to get one group by its ID (`GET /groups/{id}` doesn't exist like it does for users and discussions).

**Steps to Reproduce**
1. Open https://discuss.flarum.org/api/groups?page[limit]=1
2. See more than 1 entry

MatusMak:
**Fixes flarum/framework#2492**

- Added dedicated `GET /api/groups/{id}` endpoint for retrieving single record
- Added basic filtering, sorting and pagination to `GET /api/groups` endpoint

- Overal code style to confirm compliance with established conventions

**Confirmed**

**Required changes:**


askvortsov1 (changes requested):
Fantastic work, thank you so much for taking up this issue and diving so deep into the codebase! I left a quick change comment in the code.

More generally speaking, I have mixed feelings about the "name" filter, as that feels more like a limited search than a filter. Seeing as it currently requires a precise match, is there a concrete use case? If not, maybe we should leave that one out for now.

MatusMak:
Yeah, you are right, I'm probably lacking a use case for the exact match filter. I could rewrite it to LIKE match, but that can be a limited searcher as you said.

I wanted to introduce further scoping, so that it can potentially cover use case described in flarum/issue-archive#280. But I guess that it's so rare 99.9% of forums won't need this.

So I guess the question is whether full text match would have any benefits or not. Let me know & I will make the changes :)

askvortsov1:
To be honest, I think that for now, it might be best to hold off on any name searching functionality until we better understand use cases. The PR is already pretty big, but more importantly, if we set an API now and decide to make changes later, that would be a breaking change. @flarum/core what are y'all thoughts?

> I wanted to introduce further scoping, so that it can potentially cover use case described in flarum/issue-archive#280. But I guess that it's so rare 99.9% of forums won't need this.

flarum/issue-archive#280 is complicated because completely paginated groups has big implications across our architecture. This PR is definitely a huge step towards that direction, but I don't think we should aim to tackle it here.

This all being said, thank you again so much for the contribution!

askvortsov1 (approved):
Excellent, thank you very much!

SychO9 (changes requested):
Thank you for the PR!

askvortsov1 (approved):

SychO9 (approved):
Code looks good!

PR #3026, opened 2021-08-15
CSS Code Housekeeping

SychO9:
**Progresses #2600**

**Changes proposed in this pull request:**
General cleanup of certain bits of code, all safe and doesn't change anything.
* Removed dead code.
* Used CSS Variables in certain scopes to improve customizability.

**Reviewers should focus on:**
🤷🏼‍♂️ 

**Confirmed**

- [x] Frontend changes: tested on a local Flarum installation.

PR #3085, opened 2021-09-28
Revert "CSS Code Housekeeping"

askvortsov1:
Reverts flarum/core#3026

This PR broke some components on Safari:

image

image
image

### DIFF
--- a/less/admin/AdminNav.less
+++ b/less/admin/AdminNav.less
@@ -52,12 +52,20 @@
 }
 
 @media @phone, @tablet {
-  .AdminNav {
+  .App-nav .AdminNav {
     .Dropdown-menu {
       > li {
+        .ExtensionListTitle {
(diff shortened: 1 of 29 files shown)


SychO9:
do we know if it's a specific change ? I can't think of any change in that PR that could affect those icons, most of those changes was just minor cleanup, none of which really directly affect any of the dropdown elements.

askvortsov1:
> do we know if it's a specific change ? I can't think of any change in that PR that could affect those icons, most of those changes was just minor cleanup, none of which really directly affect any of the dropdown elements.

I'm not 100% sure, all I know is that with browserstack, it worked before this commit but didn't work afterwards. Maybe some of the vendor prefixes were actually needed? If you have time, we can try and salvage out the working parts, but if not, might be better to revert entirely and revisit next cycle.

Issue #3022, opened 2021-08-12
Dashes in table prefix will prevent installation


clarkwinkelmann:
## Bug Report

**Current Behavior**
If the database table prefix includes a dash, one or multiple migrations will fail with a syntax error and prevent completion of the installation.

**Steps to Reproduce**
1. Install Flarum according to official instructions
2. Use prefix `flarum-` (the dash is the important part, can be anywhere in the prefix)
3. Installation fails

```
SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '-flags CHANGE time created_at DATETIME NOT NULL' at line 1 (SQL: ALTER TABLE flarum-flags CHANGE time created_at DATETIME NOT NULL)
```

**Expected Behavior**
Either the dash should not cause an error, or a validation error should reject the use of a dash.

**Screenshots**

**Environment**
- Flarum version: 1.0.4
- Hosting environment: local
- PHP version: 7.4
- MySQL mysql  Ver 14.14 Distrib 5.7.33,

**Possible Solution**
This error is also present in Laravel. If I create a new Laravel app, set the db prefix to `laravel-`, then attempt to use `Blueprint::renameColumn`, I will get the same error. It seems to happen for any field type and not just datetime.

I'm unsure if this is an issue with Laravel or Doctrine DBAL. Since `Blueprint::renameColumn` only works with DBAL installed, it needs DBAL. The error happens with both DBAL 2.x which Flarum uses and DBAL 3.x which is the latest.

The only issue I managed to find that might be related to this is this one about database names https://github.com/doctrine/dbal/issues/1735

I cannot find any explanation of what Laravel considers a valid database prefix. Maybe a dash was never supposed to be a valid prefix. We could try to open an issue at https://github.com/laravel/framework/issues to know what they think of it?

Otherwise we could just add validation to prevent the use of a dash in the prefix, at least via the web UI.

**Additional Context**
Reported first here https://discuss.flarum.org/d/28603-installation-page-sqlstate42000-syntax-error-or-access-violation

The migration that fails is this one, but I suspect it's just the first with a `renameColumn` that's also not in the `install.dump` file https://github.com/flarum/flags/blob/master/migrations/2018_06_27_101500_change_flags_rename_time_to_created_at.php I'm pretty sure all `renameColumn` instructions will cause the same error.

[...]

PR #3089, opened 2021-10-01
Disallow dashes in database prefix

mbicknese:
As a temporary fix it has been requested to disallow dashes in the database prefix. The installation process fails when the prefix does include a dash.

#3022

--

**#3022*

The install process will now throw an error when the database prefix contains a dash.

**Confirmed**

### DIFF
--- a/src/Install/DatabaseConfig.php
+++ b/src/Install/DatabaseConfig.php
@@ -87,8 +87,8 @@ class DatabaseConfig implements Arrayable
         }
 
         if (! empty($this->prefix)) {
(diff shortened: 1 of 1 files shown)


askvortsov1 (approved):
LGTM, thank you very much for the contribution!

davwheat (approved):

askvortsov1:
We're currently in feature freeze for our 1.1 release, so we'll merge this in a bit. Thanks again!

mbicknese:
Wow, you're fast. I'll continue to look into the bigger issue so we can revert this change later on.

PR #3090, opened 2021-10-02
[A11Y] Improve accessibility for discussion reply count on post stream

### LINKED ISSUE #2664: [A11Y] New reply counter is focus-able when not acting as a button
## Bug Report

**Current Behavior**
The reply counter is focusable when 0 new replies are present.

image

**Possible solutions**
1. Remove `tabindex="0"` when the new reply count is 0.
2. Change to `button` element (also allows a11y tech to actually press it in the first place), and use the more syntactically correct disabled` prop.

davwheat:
**Fixes #2664**

- Add new accessibility label to reply count
  - "2 unread replies. Mark replies as read"
  - "8 replies"
- Use `<button>` element when Mark as read feature is active, otherwise use `<span>`

Nothing in particular.

**Confirmed**


davwheat:
I forgot that I already added a screenreader-only class when I made checkboxes focusable in #3014, so I've removed the one I added in this PR.

PR #3094, opened 2021-10-06
Add README documentation to ExtensionPage

imorland:
Adds a button to display an extensions `README` file within a modal directly from the `ExtensionPage`

This is _almost_ a "lift and shift" from `blomstra/readme`. Did I forget any core conventions when migrating this? 

**Screenshot**
image
image

**Confirmed**

### DIFF
--- a/js/src/admin/components/ExtensionPage.js
+++ b/js/src/admin/components/ExtensionPage.js
@@ -11,6 +11,7 @@ import LoadingModal from './LoadingModal';
 import ExtensionPermissionGrid from './ExtensionPermissionGrid';
 import isExtensionEnabled from '../utils/isExtensionEnabled';
 import AdminPage from './AdminPage';
+import ReadmeModal from './ReadmeModal';
 
 export default class ExtensionPage extends AdminPage {
   oninit(vnode) {
@@ -194,6 +195,21 @@ export default class ExtensionPage extends AdminPage {
         );
       }
     });
+
+    const extension = this.extension;
+    items.add(
+      'readme',
+      Button.component(
+        {
+          icon: 'fab fa-readme',
+          class: 'Readme-link',
+          onclick() {
+            app.modal.show(ReadmeModal, { extension });
+          },
+        },
+        app.translator.trans('core.admin.extension.readme.button_label')
+      )
+    );
 
     return items;
   }
--- a/js/src/admin/components/ReadmeModal.js
+++ b/js/src/admin/components/ReadmeModal.js
@@ -1,0 +1,57 @@
+import app from '../../admin/app';
+import Modal from '../../common/components/Modal';
+import LoadingIndicator from '../../common/components/LoadingIndicator';
+import Placeholder from '../../common/components/Placeholder';
+
+export default class ReadmeModal extends Modal {
+  oninit(vnode) {
+    super.oninit(vnode);
+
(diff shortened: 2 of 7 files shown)


locale/core.yml
```diff
@@ -131,6 +131,10 @@ core:
       open_modal: Open Settings
       permissions_title: Permissions
       uninstall_button: Uninstall
+      readme:
```
davwheat, line 134: `readme` should be sorted alphabetically in the keys here (directly below `permissions_title`).

The keys inside `readme` should also be sorted alphabetically.

js/src/admin/components/ExtensionPage.js
```diff
@@ -195,6 +196,21 @@ export default class ExtensionPage extends AdminPage {
       }
     });
 
+    const extension = this.extension;
+    items.add(
+      'readme',
+      Button.component(
+        {
+          icon: 'fab fa-readme',
+          class: 'Readme-link',
```
davwheat, line 205: Maybe `Button--link` could work here instead of styling it as a link manually? Haven't checked, though.

js/src/admin/components/ExtensionPage.js
```diff
@@ -195,6 +196,21 @@ export default class ExtensionPage extends AdminPage {
...
+          onclick() {
+            app.modal.show(ReadmeModal, { extension });
+          },
+        },
+        app.translator.trans('core.admin.extension.readme.button_label')
+      )
+    );
```
davwheat, line 212: Could we give this a priority in the ItemList?

js/src/admin/components/ReadmeModal.js
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  className() {
+    return 'ReadmeModal Modal--large';
+  }
+
+  title() {
+    const name = this.displayName;
+    return app.translator.trans('core.admin.extension.readme.title', {
+      displayName: name,
```
davwheat, line 25: Could we do away with the extra `name` variable here and directly use `this.displayName()`?

davwheat (changes requested):
All looks good on the whole, just a few nitpicks.

davwheat (approved):

askvortsov1 (changes requested):
I like the frontend a lot, and think this would be a great feature. From a code organization perspective, I'm not sure that the controller directly is the best place for this implementation to live. I'm also hesitant to add new endpoints that are far from complying with the JSON:API spec if not necessary. Left some ideas in the code.

src/Api/Controller/ReadmeController.php
```diff
@@ -0,0 +1,69 @@
...
+        } else {
+            $readme->content = null;
+        }
+
+        if ($readme->content) {
+            $xml = Fatdown::parse($readme->content);
+            $readme->content = Fatdown::render($xml);
+        }
+
+        return $readme;
```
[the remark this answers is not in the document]
imorland: I'm hesitant to include fetching the `README` content as part of the `Extension` class, as _potentially_ this could add a fair amount of data to it that's not always required. Therefore a seperated way of fetching the content is required, I feel.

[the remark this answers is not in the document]
davwheat: An `ExtensionReadme` model could work, I think?

PR #3097, opened 2021-10-07
[huntr] adding cache control headers to the admin area

luceos:
**Fixes https://huntr.dev/bounties/73176569-c102-47ad-9e95-17dd344a8847/**

This PR forces the `Cache-Control: no-store, max-age=0` header to the response in the Admin Area. This forces cache to be ignored upon browsing back and forth between pages using the browser controls. Although absolutely no fail safe, it should provide better protection against serving cached pages once an admin has signed out.

For now, I don't think we need to push this onto the Forum frontend, but we do need to start looking into cache control soon. Perhaps I can tease @BartVB to write some directions in the headers to apply for us as cache control is quite an expertise.

### DIFF
--- a/src/Admin/Middleware/DisableBrowserCache.php
+++ b/src/Admin/Middleware/DisableBrowserCache.php
@@ -18,8 +18,8 @@ class DisableBrowserCache implements Middleware
 {
     public function process(Request $request, Handler $handler): Response
     {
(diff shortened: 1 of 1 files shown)


davwheat:
Gotcha. Looks good to me.

Issue #2951, opened 2021-07-02
`formatNumber()` ignores current language convention while formatting numbers


rob006:
Polish language uses space as thousands separator, so `100000` should be formatted as `100 000`. But it looks like `formatNumber()` has really simple implementation with hardcoded comma as separator:

https://github.com/flarum/core/blob/d4e3254395fa03a0653218a543e23ad9de76ae77/js/src/common/utils/formatNumber.ts#L9-L11

It may be a good idea to switch to something like [`Intl.NumberFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat) or `toLocaleString()` instead - both `new Intl.NumberFormat('pl').format(number)` and `number.toLocaleString('pl')` returns correct format.

wolfgang-demeter:
I second that! In Germany it should read `100.000`, and not `100,000`.

askvortsov1:
If this is already included in browsers, the fix should be quite easy and shouldn't have a bundle size cost.

PR #3099, opened 2021-10-08
Added ES6 local support for formatNumber helper as per #2951

Braunson:
https://github.com/flarum/core/issues/2951

Updated `formatNumber` helper to use ES6 method to format numbers via provided locale. Unsure how to grab the current user's locale to inject when the helper is used in `PostStreamScrubber.js`. Would `app.session.user.locale` work if it's not necessarily set?

**Confirmed**

- [x] Frontend change


Braunson:
Updated to support the user's local and make the `locale` parameter optional.

Issue #3043, opened 2021-08-21
[Tags] Event posts reference [Deleted] tag


jordanjay29:
## Bug Report

**Current Behavior**
When nested secondary tags are removed from a discussion, the event posts can no longer reference them and [Deleted] appears instead. 

**Steps to Reproduce**
1. Create new discussion
2. Add to a nested secondary tag
3. Remove nested secondary tag
4. Tag event posts now reference [Deleted]

**Expected Behavior**
Nested secondary tags should have their proper names and links displayed just like Primary and Secondary tags do. 

**Screenshots**
![image](https://user-images.githubusercontent.com/7871706/130325108-2e81db95-c414-44b5-8098-8f910567b405.png)

**Environment**
- Flarum v1.0.4 on demo site, 20210821

SychO9:
I'm assuming that it's because not all tags are in the JS store now, so some tags will show as deleted in event posts.

datitisev:
@SychO9 Exactly.

What might help here is have some store method to fetch a model if it hasn't already be loaded - as a helper, by ID. Then we could just put that in oninit for the Tag component (or whatever it's called) and add a loading state 🤔 
Though one would have to account for ones that don't exist anymore and actually have been deleted... that'd be more complicated.
I'm just rambling.

SychO9:
yeah, we might also be able to take care of it in the backend side, when posts are requested, I imagine we might be able filter out tag event posts and get all tag ids from those event posts then have those models included in the response.

datitisev:
I think implementing both would make sense 🤔 but yeah you're probably right.

PR #3100, opened 2021-10-11
Better post loading support

askvortsov1:
Needed for https://github.com/flarum/core/issues/3043.

Some posts are loaded dynamically, and won't be immediately available. In this case, we show a loading indicator instead of displaying content. In this PR:

- We redraw post content if loading state has chnaged
- We show a loading indicator while loading


davwheat:
I thought we had skeleton posts for things like this don’t we?

Oh this is just generic loading indicator support for child classes to use if desired, I think. Never mind then!

davwheat (approved):

SychO9 (approved):

PR #3102, opened 2021-10-13
Test against php 8.1

sergiy-petrov:


matteocontrini:
> Not a comment about the PR content itself, but I do wonder why we don't automate tests against MySQL 8, as we do support it.



> We probably should.

Is there an issue to track this?

PR #3116, opened 2021-10-21
Allow using callbacks with `load` relations extender

SychO9:
**Needed for flarum/tags#149**

Some more complex relations can require eager loading with a callback, it is currently not possible, which is what this PR is for.

More explanation about the specific issue this helps solve in the related tags PR.

This all needs to be backwards compatible of course, is it the case or am I missing something ?

**Necessity**

**Confirmed**


src/Api/Controller/AbstractSerializeController.php
```diff
@@ -194,6 +220,27 @@ protected function loadRelations(Collection $models, array $relations): void
 
         if (! empty($relations)) {
             $relations = array_unique($relations);
+        }
+
+        $callables = [];
+
+        foreach ($relations as $k => $relation) {
```
askvortsov1, line 227: It's been a bit challenging to understand this. Instead of unsetting `$relations`, could we have `callableRelations` and `nonCallableRelations`, and generate both from `relations`? That way, it's a computational transformation and not mutation-based.

askvortsov1 (changes requested):

src/Api/Controller/AbstractSerializeController.php
```diff
@@ -222,26 +222,27 @@ protected function loadRelations(Collection $models, array $relations, ServerReq
...
                 $load = $addedRelationCallables[$relation];
 
-                $callables[$relation] = function ($query) use ($load, $request, $relations) {
+                $callableRelations[$relation] = function ($query) use ($load, $request, $relations) {
                     $load($query, $request, $relations);
                 };
-
-                unset($relations[$k]);
+            } else {
+                $nonCallableRelations[$k] = $relation;
```
askvortsov1, line 236: What is $k here? Maybe we could just append directly without setting it to a particular key value? If so, that would simplify the for loop too.

askvortsov1 (changes requested):

src/Api/Controller/AbstractSerializeController.php
```diff
@@ -194,7 +220,29 @@ protected function loadRelations(Collection $models, array $relations): void
...
+
+        foreach ($relations as $k => $relation) {
+            if (isset($addedRelationCallables[$relation])) {
+                $load = $addedRelationCallables[$relation];
+
+                $callableRelations[$relation] = function ($query) use ($load, $request, $relations) {
+                    $load($query, $request, $relations);
+                };
+            } else {
+                $nonCallableRelations[$k] = $relation;
```
SychO9, line 236:
```suggestion
        foreach ($relations as $relation) {
            if (isset($addedRelationCallables[$relation])) {
                $load = $addedRelationCallables[$relation];

                $callableRelations[$relation] = function ($query) use ($load, $request, $relations) {
                    $load($query, $request, $relations);
                };
            } else {
                $nonCallableRelations[] = $relation;
```

askvortsov1 (approved):

tankerkiller125 (approved):

PR #3127, opened 2021-10-26
feat: Default Settings Repository

### LINKED ISSUE #3056: Default Settings Via Extender
See https://github.com/flarum/core/issues/2897#issuecomment-905902877. I'll fill this in with more details later.

SychO9:
**Fixes #3056**
**Needed for #3011**

* Introduces a `default` settings extender to register setting default values.
* Deprecates the `addSettings` migration helper.

I don't think we're missing anything about us switching to this approach ?

**Necessity**

**Confirmed**


src/Settings/DefaultSettingsRepository.php
```diff
@@ -0,0 +1,53 @@
...
+        if (isset($this->defaults[$key])) {
+            throw new RuntimeException("Cannot modify immutable default setting $key.");
+        }
+
+        $this->defaults[$key] = $value;
+    }
+
+    public function get($key, $default = null)
+    {
+        return $this->inner->get($key, $this->defaults[$key] ?? $default);
```
askvortsov1, line 36: Can we add a comment saying that global config overrules local config because it's deprecated?

src/Settings/SettingsServiceProvider.php
```diff
@@ -20,12 +20,21 @@ class SettingsServiceProvider extends AbstractServiceProvider
...
+        $this->container->singleton(DefaultSettingsRepository::class);
+
+        $this->container->alias(DefaultSettingsRepository::class, 'flarum.settings.default');
+
         $this->container->singleton(SettingsRepositoryInterface::class, function (Container $container) {
-            return new MemoryCacheSettingsRepository(
-                new DatabaseSettingsRepository(
-                    $container->make(ConnectionInterface::class)
+            $defaults = $container->make(DefaultSettingsRepository::class);
+            $defaults->setInner(
```
askvortsov1, line 29: Maybe `inner` could be a constructor arg, for more consistency with `MemoryCacheSettingsRepository`?

askvortsov1 (changes requested):
Looking over this again, there's still one thing I'm unsure about.

There's a difference between a setting not being set (in which case `default` should be used), and a setting having the value of `null` (in which case `null` should be returned). Ideally, our system should be able to distinguish between the two, but because of the current implementation with nullable defaults, the difference between the two is encapsulated within `get` and not available to the caller. In this PR, you pass the configured default value through the repo layers, which solves the problem, but depends on the `default` argument that we are hoping to remove.

For v2, when we do remove the argument, we could throw an exception in `DatabaseSettingsRepository` if the setting is not found. This doesn't depend on the default arg, so it fits well with our plans. It would be a breaking change though, which is why we can't do it now.

src/Settings/SettingsServiceProvider.php
SychO9: That's how I initially implemented it, but the thing that's different about `DefaultSettingsRepository` is that it's not only acting as a `SettingsRepositoryInterface` but also as a repository for default setting values. That means we need to define the class itself as a singleton so that we can extend it in the extender to add defaults, and a constructor arg inner wouldn't work with that.

askvortsov1: Maybe we could use some kind of mutable collection to store the underlying map? Although that's a lot of boilerplate, probably not worth it.

askvortsov1 (approved):

SychO9:
> There's a difference between a setting not being set (in which case default should be used), and a setting having the value of null (in which case null should be returned). Ideally, our system should be able to distinguish between the two, but because of the current implementation with nullable defaults, the difference between the two is encapsulated within get and not available to the caller. In this PR, you pass the configured default value through the repo layers, which solves the problem, but depends on the default argument that we are hoping to remove.

Good point! hadn't thought of that, and agree with the plan for 2.0

src/Settings/SettingsServiceProvider.php
SychO9: Wait, that's a great idea, if we can't bind an array, we definitely could bind a collection, that would cleanup the default settings repository implementation.

askvortsov1 (approved):
This is really nice now!

PR #3140, opened 2021-10-29
Remove color validation in appearance admin page & add color indicator

### LINKED ISSUE #2121: Colour names and RGB passes validation
On the tags extension, entering the colour name will change the preview at the top to the colour of your choice but does not pass validation. Only HEX codes work. This used to work back in BETA 6 but was broken in future BETA versions and to this day still is broken in BETA 12

datitisev:
**Fixes #2121**

- Remove color validation in appearance page
- Add color indicator to inputs

- Color validation can be done with dummy HTML elements, but from the issue comments I don't think we want to validate at all.
- After making this I thought this might be a good use for a new component like `ColorInput` or something? That way it also avoids using `::after` for the box... though a component would have to be an entire form group.
  - Perhaps a setting type instead? Though `type="color"` is also a valid input type.
  - FYI, `input::after` is not valid so that's why I had to add it to Form-group

**Screenshot**

image

**Necessity**

**Confirmed**

### DIFF
--- a/js/src/admin/components/AdminPage.tsx
+++ b/js/src/admin/components/AdminPage.tsx
@@ -272,9 +272,11 @@ export default abstract class AdminPage<CustomAttrs extends IPageAttrs = IPageAt
 
         if ((ColorSettingTypes as readonly string[]).includes(type)) {
           Tag = ColorInput;
+        } else {
+          componentAttrs.type = type;
         }
 
-        settingElement = <Tag id={inputId} aria-describedby={helpTextId} type={type} bidi={this.setting(setting)} {...componentAttrs} />;
+        settingElement = <Tag id={inputId} aria-describedby={helpTextId} bidi={this.setting(setting)} {...componentAttrs} />;
       }
     }
 
--- a/js/src/admin/components/AdminPage.tsx
+++ b/js/src/admin/components/AdminPage.tsx
@@ -61,6 +61,7 @@ interface CommonSettingsItemOptions extends Mithril.Attributes {
   label: Mithril.Children;
   help?: Mithril.Children;
   className?: string;
+  formGroupAttrs?: Mithril.Attributes;
 }
 
 /**
@@ -218,7 +219,7 @@ export default abstract class AdminPage<CustomAttrs extends IPageAttrs = IPageAt
       return entry.call(this);
     }
 
-    const { setting, help, type, label, ...componentAttrs } = entry;
+    const { setting, help, type, label, formGroupAttrs = {}, ...componentAttrs } = entry;
 
     const value = this.setting(setting)();
 
--- a/js/src/admin/components/AdminPage.tsx
+++ b/js/src/admin/components/AdminPage.tsx
@@ -10,7 +10,7 @@ import Stream from '../../common/utils/Stream';
 import saveSettings from '../utils/saveSettings';
 import AdminHeader from './AdminHeader';
 import generateElementId from '../utils/generateElementId';
-import ColorInput from '../../common/components/ColorInput';
+import ColorPreviewInput from '../../common/components/ColorPreviewInput';
 
 export interface AdminHeaderOptions {
   title: string;
@@ -77,7 +77,7 @@ export interface HTMLInputSettingsComponentOptions extends CommonSettingsItemOpt
 const BooleanSettingTypes = ['bool', 'checkbox', 'switch', 'boolean'] as const;
 const SelectSettingTypes = ['select', 'dropdown', 'selectdropdown'] as const;
 const TextareaSettingTypes = ['textarea'] as const;
-const ColorSettingTypes = ['color-input', 'color-preview', 'custom-color'] as const;
+const ColorPreviewSettingType = 'color-preview';
 
 /**
  * Valid options for the setting component builder to generate a Switch.
@@ -107,9 +107,12 @@ export interface TextareaSettingComponentOptions extends CommonSettingsItemOptio
 
 /**
  * Valid options for the setting component builder to generate a ColorInput.
+=======
+ * Valid options for the setting component builder to generate a ColorPreviewInput.
+>>>>>>> c10922974 (Rename component to ColorPreviewInput, remove aliases in admin & export in compat)
  */
-export interface ColorSettingComponentOptions extends CommonSettingsItemOptions {
-  type: typeof ColorSettingTypes[number];
+export interface ColorPreviewSettingComponentOptions extends CommonSettingsItemOptions {
+  type: typeof ColorPreviewSettingType;
 }
 
 /**
@@ -120,7 +123,7 @@ export type SettingsComponentOptions =
   | SwitchSettingComponentOptions
   | SelectSettingComponentOptions
   | TextareaSettingComponentOptions
-  | ColorSettingComponentOptions;
+  | ColorPreviewSettingComponentOptions;
 
 /**
  * Valid attrs that can be returned by the `headerInfo` function
@@ -270,8 +273,8 @@ export default abstract class AdminPage<CustomAttrs extends IPageAttrs = IPageAt
       } else {
         let Tag: any = 'input';
 
-        if ((ColorSettingTypes as readonly string[]).includes(type)) {
-          Tag = ColorInput;
+        if (type === ColorPreviewSettingType) {
+          Tag = ColorPreviewInput;
         } else {
           componentAttrs.type = type;
         }
--- a/js/src/admin/components/AdminPage.tsx
+++ b/js/src/admin/components/AdminPage.tsx
@@ -10,6 +10,7 @@ import Stream from '../../common/utils/Stream';
 import saveSettings from '../utils/saveSettings';
 import AdminHeader from './AdminHeader';
 import generateElementId from '../utils/generateElementId';
+import ColorInput from '../../common/components/ColorInput';
 
 export interface AdminHeaderOptions {
   title: string;
@@ -77,6 +78,7 @@ export interface HTMLInputSettingsComponentOptions extends CommonSettingsItemOpt
 const BooleanSettingTypes = ['bool', 'checkbox', 'switch', 'boolean'] as const;
 const SelectSettingTypes = ['select', 'dropdown', 'selectdropdown'] as const;
 const TextareaSettingTypes = ['textarea'] as const;
+const ColorSettingTypes = ['color-input', 'color-preview', 'custom-color'] as const;
 
 /**
  * Valid options for the setting component builder to generate a Switch.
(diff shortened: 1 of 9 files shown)


SychO9 (commented):
I love this!

> After making this I thought this might be a good use for a new component like ColorInput or something? That way it also avoids using ::after for the box... though a component would have to be an entire form group.

I think a `ColorInput` would be perfect, though why would it have to be a `Form-group` ? if it's about wrapping the input itself, I guess we can just wrap it in a `Color-input` div and have that within the form group. Would be consistent with `Search-input` as well.

> Perhaps a setting type instead? Though type="color" is also a valid input type.

my experience with `type=color` (at least on chrome) is that it's not great, it does provide a native color picker which is nice but.. can be a pain to use.

datitisev:
> I think a ColorInput would be perfect, though why would it have to be a Form-group

Nevermind, it doesn't. I thought the CSS stylings were from it lol. But there's also FormControl class.

> my experience with type=color (at least on chrome) is that it's not great, it does provide a native color picker which is nice but.. can be a pain to use.

What I meant with mentioning `type=color` is that I wouldn't want to add an option (like select, bool) to `buildSettingComponent` that replaces the `color` type as that is technically a valid HTML input type. I know we don't want to use it but I don't think we would want to replace its uses. Would be potential breaking change as well, I guess.

SychO9:
> What I meant with mentioning type=color is that I wouldn't want to add an option (like select, bool) to buildSettingComponent that replaces the color type as that is technically a valid HTML input type. I know we don't want to use it but I don't think we would want to replace its uses. Would be potential breaking change as well, I guess.

Oh yeah that's a good point. Hmm.. `colour` no that's a joke..

I think it'd be fine to go with whatever naming, `custom-color` or something along the lines, as long as we can use it.

less/admin/AppearancePage.less
```diff
@@ -24,12 +25,17 @@
...
-    }
+  .Form-group::after {
+    content: '';
+    display: inline-block;
+    position: absolute;
+    right: 13px;
+    bottom: 8px;
+    width: 20px;
+    height: 20px;
+    background-color: var(--input-value);
```
davwheat, line 36: I wonder if we could add some embedded SVG warning icon here as a backup for the variable based background, in case the entered value is invalid. That can act as some visual validation instead of the not so great JS validation from before.

datitisev: Hm, interesting. I mean, if we wanted validation, we can use HTML element shenanigans with dummy elements and `color` style modifications... but I do like the image part.

It could be a ::before element that is behind the element when I create the ColorInput component.

davwheat: It wouldn't even need to be that. If we just had a `background` property directly above this one, it'd be applied when the `var` holds an invalid value.

(Though I think this `background-color` may need to become `background` too, or maybe we'd have to use a slightly hacky `background-image: linear-gradient(from var(...), to var(...))` value instead.)

davwheat: So...

```css
background-image: url (/* embedded SVG */);
background-image: linear-gradient(from var(--input-value), to var(--input-value))
```

datitisev: I don't want to embed an SVG, I'd rather use the stuff already loaded like the FA icons, without duplicating their SVG data.

davwheat: Hmm yeah, in that case FA icon styles as a `::before` is the way to go, I think?

SychO9: > Hmm yeah, in that case FA icon styles as a ::before is the way to go, I think?

Or maybe have the box be a dom element and place it there with CSS, the element would contain the icon behind and flexed to be centered

davwheat: mmm flex... 🤤 

But yeah, that might be the better way to go.

datitisev:
Made it a component instead + the validation.

One can click through both the box & icon. I tested adding a padding so text doesn't go underneath but then you can't click through the box to the input as it's no longer there... and so I decided against it.

image

My only concern with the code is how it obtains the `value` currently. I couldn't think of a good way to accomplish that.

davwheat:
> My only concern with the code is how it obtains the value currently. I couldn't think of a good way to accomplish that.

The only other way I can think of is a data attribute, which you can then use with the `attr(data-value)` function.

datitisev:
> The only other way I can think of is a data attribute, which you can then use with the attr(data-value) function.

Oh, true, but I wasn't referring to the CSS value. I meant the input value from the JS side that is then passed - I don't like using the `attrs.bidi?.() || attrs.value` as it looks odd and could potentially misrepresent the value at some point? Not sure.

Tried to override `oninput` but then that'd be on every view render and I didn't know if that was a good idea either (could also be done in initAttrs I guess and then set as a temporary `attr` when passed?)

SychO9:
> One can click through both the box & icon. I tested adding a padding so text doesn't go underneath but then you can't click through the box to the input as it's no longer there... and so I decided against it.

I think if we wrap it in a `label` we could still clock on the box/icon and have the input be focused, I'll test this later to confirm though.

davwheat:
Should do, most definitely with `pointer-events: none;` on the preview and invalid box.

datitisev:
> Should do, most definitely with pointer-events: none; on the preview and invalid box.

They already have this :P 

> I think if we wrap it in a label we could still click on the box/icon and have the input be focused, I'll test this later to confirm though.

:+1:

datitisev:
Re-requested reviews just to clear them :+1:

SychO9 (changes requested):
Looks good!

a nitpick though, the class naming,

as per the [convention](https://docs.flarum.org/contributing#css) we follow, `Color-input--preview` should be `Color-input-preview`, same for `Color-input--icon`. I'd also rename `Color-input` to `ColorInput`.

SychO9 (approved):

js/src/admin/components/AdminPage.tsx
```diff
@@ -75,6 +76,7 @@ export interface HTMLInputSettingsComponentOptions extends CommonSettingsItemOpt
 
 const BooleanSettingTypes = ['bool', 'checkbox', 'switch', 'boolean'] as const;
 const SelectSettingTypes = ['select', 'dropdown', 'selectdropdown'] as const;
+const ColorSettingTypes = ['color-input', 'color-preview', 'custom-color'] as const;
```
askvortsov1, line 79: Do we need to have `color-input` in here? The other two are self explanatory, but this one is a bit confusing

askvortsov1 (changes requested):
Phenomenal, thank you so much! I'm very excited to see continued growth of Flarum's component library :)

Only thing is, we should probably export it in compat so extensions can use it too.

Can't wait till we never have to deal with `compat` again...

js/src/admin/components/AdminPage.tsx
davwheat: I'd prefer only having one value, to be honest. For 2.0 we should deprecate the others.

datitisev: If we make it just `color-preview`, I assume that should be changed to this?

```js
const ColorPreviewSettingType = 'color-preview' as const
```

askvortsov1: Makes sense to me, although all the current checks for characters will need to be changed.

Also, is `as const` still necessary then?

datitisev: Oh, looks like it is not. I had no idea what it did until now tbh.

js/src/admin/components/AdminPage.tsx
```diff
@@ -258,7 +268,15 @@ export default abstract class AdminPage<CustomAttrs extends IPageAttrs = IPageAt
       if ((TextareaSettingTypes as readonly string[]).includes(type)) {
         settingElement = <textarea id={inputId} aria-describedby={helpTextId} bidi={this.setting(setting)} {...componentAttrs} />;
       } else {
-        settingElement = <input id={inputId} aria-describedby={helpTextId} type={type} bidi={this.setting(setting)} {...componentAttrs} />;
+        let Tag: any = 'input';
```
askvortsov1, line 271: Is there no way of typing this more specifically?

askvortsov1 (approved):

js/src/admin/components/AdminPage.tsx
datitisev: There probably is...  not 100% sure on what it'd be.

askvortsov1: If you look at the hyperscript signatures under mithril, maybe the type of the first argument to `m` will do?

Would be nice to have, but it's not a blocker.

davwheat: I think I typed this to `string | Component` because almost all JSX elements will be subclasses of our Component class, or strings.

js/src/common/components/ColorPreviewInput.tsx
```diff
@@ -0,0 +1,26 @@
...
+    attrs.type ||= 'text';
+
+    return (
+      <div className="ColorInput">
+        <input className={classList('FormControl', className)} {...attrs} />
+
+        <span className="ColorInput-icon">{icon('fas fa-exclamation-circle')}</span>
+
+        <div className="ColorInput-preview" style={{ '--input-value': value }} />
+      </div>
```
davwheat, line 23: One topic of note here is accessibility.

I didn't think about this earlier, but we would want some way of alerting a screenreader to the fact that the input is invalid.

A11Y isn't a massive one of our focuses on the admin panel, so I'm fine with not implementing this in the time being, but some aria roles to indicate element metadata would be nice:

```suggestion
      <div className="ColorInput">
        <input className={classList('FormControl', className)} {...attrs} />

        <span className="ColorInput-icon" role="presentation">{icon('fas fa-exclamation-circle')}</span>

        <div className="ColorInput-preview" style={{ '--input-value': value }} role="presentation" />
      </div>
```

davwheat (changes requested):

js/src/admin/components/AdminPage.tsx
askvortsov1: Maybe `Mithril.ComponentTypes | string`?

davwheat: @askvortsov1 I think something like that could be helpful in our global typings.

```ts
declare type ElementTag<Attrs = Record<string, unknown>, State = Record<string, unknown>> = string | ComponentTypes<Attrs, State>

Comment (askvortsov1) [reply js/src/admin/components/AdminPage.tsx:271]: Sounds good. I suppose we could add that in this PR?

Comment (davwheat) [reply js/src/admin/components/AdminPage.tsx:271]:
```suggestion
        let Tag: VnodeElementTag = 'input';
```

davwheat (approved):

SychO9 (approved):

PR #3143, opened 2021-10-30
Update ConnectToDatabase.php

koriaking:
flarum/issue-archive#49 

Because the default SQL_REQUIRE_PRIMARY_KEY cluster parameter for MySQL (for example, in Digital Ocean) is ON, we can insert a different value when installing a new forum and fix the error below

SQLSTATE[HY000]: General error: 3750 Unable to create or change a table without a primary key, when the system variable 'sql_require_primary_key' is set. Add a primary key to the table or unset this variable to avoid this message. Note that tables without a primary key can cause performance problems in row-based replication, so please consult your DBA before changing this setting.

**Fixes #0000**

**Screenshot**

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/src/Install/Steps/ConnectToDatabase.php
+++ b/src/Install/Steps/ConnectToDatabase.php
@@ -48,7 +48,10 @@ class ConnectToDatabase implements Step
                 throw new RangeException('MySQL version too low. You need at least MySQL 5.6.');
             }
         }
-
+        
+        /** resolves #3029 */
+        $pdo->exec('SET SESSION sql_require_primary_key=0');
+        
         ($this->store)(
             new MySqlConnection(
                 $pdo,


askvortsov1 (commented):
Thank you for the PR! Although I appreciate the contribution, I'm not sure that this is the approach we want to take here: although modifying our current DB setup to ensure primary keys on every table is a pain, I think it's also somewhat reasonable, as that has implications for scalability. On the other hand, perhaps configuring things so that they are consistent everywhere isn't a bad idea? 

@luceos @imorland would appreciate your thoughts on this

luceos:
I honestly am against this. Because there's no way to (easily) confirm and most importantly keep confirming (forward compatibility) that this works across all MySQL'ish databases; MySQL, Mariadb, Percona and cloud instances. It's better to create something inside your local extender to circumvent this inside your code for your specific hosting environment than trying to shape the software to your specific use case.

askvortsov1:
Thanks again, but I don't think we will be taking this approach.

PR #3145, opened 2021-11-01
Fix post policy for PHP 8

### LINKED ISSUE #3144: Edit post "until reply" policy broken on PHP8.0
## Bug Report

**Current Behavior**
Setting the "Allow post editing" setting to "Until next reply" doesn't work on PHP 8.0

**Steps to Reproduce**
1. Use PHP 8.0
2. Set "Allow post editing" setting to "Until next reply"
3. Create additional replies
4. Can still edit post

**Expected Behavior**
Behavior should be identical to older PHP versions

**Screenshots**

**Environment**
- Flarum version: 1.1.0
- PHP version: 8.0

**Possible Solution**
The problem is here https://github.com/flarum/core/blob/9e1b05a358055db1dc8093036a7e006c254b921c/src/Post/Access/PostPolicy.php#L59-L61

The third condition always evaluates to `true` on PHP 8.0 when the setting value is a string that doesn't convert to a number.

```
$ php8.0 -r "var_dump(60 < 'reply');"
bool(true)
$ php7.4 -r "var_dump(60 < 'reply');"
bool(false)
```

I think we should add an `is_numeric()` test inside the third condition.

**Additional Context**
Reported here https://discuss.flarum.org/d/29347-post-is-editable-if-it-has-reply

clarkwinkelmann:
**Fixes #3144**

Add unit tests and fix post policy for PHP 8

I'm really not sure if that's the best way to built the tests, but it does cover the original issue.

Would this best be covered by a test of the edit post API endpoint directly instead of testing just the gate?

Should I namespace the test differently?

Should I use more test methods instead of grouping the time tests?

**Screenshot**

**Necessity**

**Confirmed**

### DIFF
--- a/src/Post/Access/PostPolicy.php
+++ b/src/Post/Access/PostPolicy.php
@@ -58,7 +58,7 @@ class PostPolicy extends AbstractPolicy
 
             if ($allowEditing === '-1'
                 || ($allowEditing === 'reply' && $post->number >= $post->discussion->last_post_number)
-                || ($post->created_at->diffInMinutes(new Carbon) < $allowEditing)) {
+                || (is_numeric($allowEditing) && $post->created_at->diffInMinutes(new Carbon) < $allowEditing)) {
                 return $this->allow();
             }
         }
--- a/tests/integration/post/PostPolicyTest.php
+++ b/tests/integration/post/PostPolicyTest.php
@@ -1,0 +1,112 @@
+<?php
+
+/*
+ * This file is part of Flarum.
+ *
+ * For detailed copyright and license information, please view the
+ * LICENSE file that was distributed with this source code.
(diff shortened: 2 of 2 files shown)


SychO9 (approved):

clarkwinkelmann:
I guess I could make the tests a tiny bit faster by re-using the models instead of calling `findOrFail()` each time. In particular the same user which is retrieved 4 times per test without any real need. I could also do the same for the posts but there's less gain there with 2 reuse per post. That's still a lot less DB requests. What do you think?

askvortsov1 (approved):
LGTM! I think I would prefer variables referring to the model instances. There is a small performance improvement, but I think it'll also make the code a bit more readable.

PR #3146, opened 2021-11-02
Declare & Use CSS Custom Properties

### LINKED ISSUE #2600: CSS Variables

SychO9:
**Fixes #2600**

* Declare CSS custom properties for Less variables where possible.
* Use CSS custom properties instead of Less variables where possible.
* Declare color variations as CSS custom properties before using relevant mixins.
* Make some hard-coded colors into CSS custom properties.

Most of the changes are simple replacements of the common variables. What you should pay attention to is the following:

* `root.css` declaration of custom properties.
* Backwards compatibility.
* Anything that I might've missed about this.
* Mixins and the way they work (`.Button--color()`, `.light-contents()`)
For example the `Button--color` mixin used to be directly called as such: 

```less
.Button {
    .Button--color(@control-color, @control-bg);
}

.Button--primary {
    .Button--color(@body-bg, @primary-color);
}
```
Now the mixin is called like this:
```less
:root {
    .Button--color-vars(@control-color, @control-bg, 'button');
    .Button--color-vars(@body-bg, @primary-color, 'button-primary');
}

.Button {
    .Button--color-auto('button');
}

.Button--primary {
    .Button--color-auto('button-primary');
}
```

This is done so that the color variations created by the mixin are declared as CSS custom properties first, as it would make it possible for us to switch to a darkmode with a simple class name change.

**Necessity**

**Confirmed**

**Required changes:**
- [ ] Related documentation PR: (Remove if irrelevant): Some things here should be documented for theme designers and extension devs.


less/common/App.less
```diff
@@ -130,16 +130,16 @@
...
+    background: var(--header-bg);
+    width: var(--drawer-width);
     position: fixed;
     left: 0;
     top: 0;
     bottom: 0;
-    box-shadow: 0 2px 6px @shadow-color;
-    transform: translate(-@drawer-width - 6px, 0);
+    box-shadow: 0 2px 6px var(--shadow-color);
+    transform: translate(calc(~"0px - var(--drawer-width) - 6px"), 0);
```
askvortsov1, line 140: Why not `-6px - var(...)`?

less/admin/UsersListPage.less
```diff
@@ -59,10 +59,10 @@
       }
 
       &--shaded {
-        background: darken(@body-bg, 3%);
+        background: var(--body-bg-dark);
```
askvortsov1, line 62: I wonder if `darker` might be better than `dark`, since it's not completely dark. Or maybe `shaded` for everything?

less/common/mixins/light-contents.less
```diff
@@ -1,26 +1,26 @@
 // This is a mixin which styles components (buttons, inputs, etc.) for use on
 // dark backgrounds.
-.light-contents(@color: #fff, @control-bg: fade(#000, 10%), @control-color: #fff) {
+.light-contents(@color: #fff; @control-bg: fade(#000, 10%); @control-color: #fff; @name: 'light-content') {
```
askvortsov1, line 3: Shouldn't the variables args be removed here, since it should pull those from the corresponding `light-contents-vars`?

less/common/App.less
```diff
@@ -221,7 +221,7 @@
 @media @phone {
   .App-drawer {
     & when (@config-colored-header = true) {
-      .light-contents(@header-color, @header-control-bg, @header-control-color);
+      .light-contents(@name: 'header-light');
```
askvortsov1, line 224: Maybe `header-colored` instead of `header-light`?

less/common/Badge.less
```diff
@@ -35,5 +35,5 @@
 }
 
 .Badge--hidden {
-  background: #888;
+  --badge-bg: var(--badge-hidden-bg);
```
askvortsov1, line 38: Nice to see less hardcoding!

less/common/Button.less
```diff
@@ -51,9 +51,10 @@
   white-space: nowrap;
   line-height: 20px;
   padding: 8px 13px;
-  border-radius: @border-radius;
+  border-radius: var(--border-radius);
   .user-select(none);
-  .Button--color(@control-color, @control-bg, 'button');
+  color: var(--button-color);
```
askvortsov1, line 56: Why drop the mixin here?

less/common/Checkbox.less
```diff
@@ -89,34 +89,34 @@
...
-  color: @muted-more-color !important;
+  color: var(--muted-more-color) !important;
 
   &.on .Checkbox-display {
-    background: fade(#58a400, 0.5);
+    background: transparent;
   }
   &.off .Checkbox-display {
-    background: fade(@control-bg, 0.5);
+    background: transparent;
```
askvortsov1, line 117: Are these changes negligible?

less/common/Modal.less
```diff
@@ -107,37 +107,20 @@
...
+    background: var(--muted-more-color);
   }
 }
 .Modal-footer {
   border: 0;
   padding: 20px;
   text-align: center;
-  color: @muted-color;
-}
-.Modal-loading {
```
askvortsov1, line 119: Dead code, right? Not finding any references to it, but still wanted to make sure.

less/common/mixins/button-color.less
```diff
@@ -0,0 +1,41 @@
...
+  --button-bg-disabled: var(~"--@{name}-bg-disabled");
+}
+
+/**
+ * Legacy mixin, allows customizing button colors inline without declaring custom CSS properties at the root.
+ * For a better theming experience, declaring the custom CSS properties at the root first using `Button--color-vars()`
+ * then using `Button--color-auto()` is recommended.
+ */
+.Button--color(@color; @background) {
+  .Button--color-vars(@color, @background, 'button');
```
askvortsov1, line 40: Shouldn't this do `Button--color-auto` too, so that it applies style in addition to declaring the css vars for it?

less/forum/DiscussionPage.less
```diff
@@ -69,17 +69,17 @@
 
 @media @tablet-up {
   .DiscussionPage-list {
-    left: -@pane-width - 6px;
+    left: calc(~"0px - var(--pane-width) - 6px");
```
askvortsov1, line 72: Same here, could this be -6px - var?

askvortsov1 (changes requested):

askvortsov1:
This is huge! Really excited to see us taking this big step towards themability and modern architecture.

less/common/App.less
SychO9: oh true that should work

less/common/mixins/light-contents.less
SychO9: backwards compatibility ;), if we remove the args here, any extensions using the mixin will break.

less/common/mixins/button-color.less
SychO9: The vars we're declaring here are local and they are overriding existing vars, that is `--button-color`, `--button-bg`. So by overriding they are applied.

less/common/Modal.less
SychO9: Yup, dead code, I tracked it to the commit where the JSX was removed, I've seen this happen quiet a few times where the relevant UI is removed but the CSS is forgotten, we should try to watch out for that ourselves too.

less/common/Checkbox.less
SychO9: yea, I've come across 3 similar instances at least, where the alpha channel is nearly 0, which doesn't really make sense, because that's not something the eye can really see, so it might as well be set to transparent.

less/common/Button.less
SychO9: The mixin used to target subelements and add color variations to them, `> .Button-label`, `&:hover`, `&:active` ..etc

with the flexibility of CSS variables, we apply those by default, and we only override the color variables locally by declaring them on a local level.

less/common/App.less
```diff
@@ -130,16 +130,16 @@
...
+    background: var(--header-bg);
+    width: var(--drawer-width);
     position: fixed;
     left: 0;
     top: 0;
     bottom: 0;
-    box-shadow: 0 2px 6px @shadow-color;
-    transform: translate(-@drawer-width - 6px, 0);
+    box-shadow: 0 2px 6px var(--shadow-color);
+    transform: translate(calc(~"-6px - var(--drawer-width)"), 0);
```
davwheat, line 140:
```suggestion
    transform: translateX(calc(~"-6px - var(--drawer-width)"));
```

davwheat (approved):
Looks good to me! Just one minor suggestion.

less/common/mixins/light-contents.less
askvortsov1: Gotcha. Should we deprecate it?

askvortsov1 (approved):

less/common/mixins/light-contents.less
SychO9: Well, we still use it ourselves in the code, unless you mean creating a new one with a different name and then deprecating this one.

I think it's fine to leave though, the first two args are still used if someone wants to not use CSS vars and keep it inline, unless we'd rather enforce the CSS var usage, which might be better.

askvortsov1 (approved):
Excellent!

PR #3147, opened 2021-11-02
feat: allow use of any tag in `listItems` helper

davwheat:

This PR allows the use of any tag with the `listItems` helper function. This is useful to easily map an ItemList into separate DOM nodes outside of a `<ul>` element, without the need of code duplication.

Not everything needs to be a list. In fact, using `<ul>` and `<li>` too much is a common issue across the web, and is the reason why WebKit (Safari on macOS and iOS) actually remove all semantics from list tags when their `list-style` is set to `none`. [(Source)](https://www.scottohara.me/blog/2019/01/12/lists-and-safari.html)

This change could help to prevent such "list-itis" within Flarum and its extensions.

**Screenshot**

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/js/src/common/helpers/listItems.tsx
+++ b/js/src/common/helpers/listItems.tsx
@@ -1,6 +1,7 @@
 import type Mithril from 'mithril';
 import Separator from '../components/Separator';
 import classList from '../utils/classList';
+import type Component from '../Component';
 
 function isSeparator(item): boolean {
   return item.tag === Separator;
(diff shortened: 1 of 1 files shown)


js/src/common/helpers/listItems.tsx
```diff
@@ -3,13 +3,13 @@ import Separator from '../components/Separator';
...
 
-function isSeparator(item): boolean {
+function isSeparator(item: Mithril.Vnode): boolean {
   return item.tag === Separator;
 }
 
-function withoutUnnecessarySeparators(items: Array<Mithril.Vnode>): Array<Mithril.Vnode> {
-  const newItems = [];
-  let prevItem;
+function withoutUnnecessarySeparators(items: Mithril.Vnode[]): Mithril.Vnode[] {
```
askvortsov1, line 10: Should this be `Vnode[]` or `Children`?

davwheat: Either should work to pass into the function.

```suggestion
function withoutUnnecessarySeparators(items: Mithril.Children): Mithril.Children {
```

js/src/common/helpers/listItems.tsx
```diff
@@ -1,14 +1,15 @@
 import type Mithril from 'mithril';
 import Separator from '../components/Separator';
 import classList from '../utils/classList';
+import type * as Component from '../Component';
 
-function isSeparator(item): boolean {
+function isSeparator(item: Mithril.Vnode): boolean {
```
davwheat, line 6:
```suggestion
function isSeparator(item: Mithril.Children): boolean {
```

js/src/common/helpers/listItems.tsx
```diff
@@ -1,14 +1,15 @@
...
   return item.tag === Separator;
 }
 
-function withoutUnnecessarySeparators(items: Array<Mithril.Vnode>): Array<Mithril.Vnode> {
-  const newItems = [];
-  let prevItem;
+function withoutUnnecessarySeparators(items: Mithril.Vnode[]): Mithril.Vnode[] {
+  const newItems: Mithril.Vnode[] = [];
+  let prevItem: Mithril.Vnode;
```
davwheat, line 12:
```suggestion
  const newItems: Mithril.Children = [];
  let prevItem: Mithril.Child;
```

askvortsov1 (approved):

SychO9 (approved):

PR #3149, opened 2021-11-03
chore: replace `for ... in` with `Array.reduce`

davwheat:

Replace `for ... in` loop with `Array.reduce`.

It's a reduction in bundle size (one whole byte!!), and I think it benefits from the readability and helps to set our precedent for more modern helper methods within Flarum.

**Necessity**

**Confirmed**


askvortsov1 (approved):
Now do it in OCaml

SychO9 (approved):

davwheat:
@askvortsov1 yeah think I'll pass on that one... 😂

PR #3157, opened 2021-11-08
Add aria-label attribute to navigation drawer button.

grimmdude:
Adds `aria-label` attribute to navigation drawer button to improve accessibility and satisfy this warning in Chrome.

Screen Shot 2021-11-08 at 8 12 14 AM

Affects every page that the navigation drawer button is visible.

**Necessity**

**Confirmed**

### DIFF
--- a/js/src/common/components/Navigation.js
+++ b/js/src/common/components/Navigation.js
@@ -93,7 +93,7 @@ export default class Navigation extends Component {
         drawer.show();
       },
       icon: 'fas fa-bars',
-      ariaLabel: 'Open Navigation Drawer',
+      ariaLabel: app.translator.trans('core.lib.nav.drawer_button'),
     });
   }
 }
--- a/js/src/common/components/Navigation.js
+++ b/js/src/common/components/Navigation.js
@@ -93,7 +93,7 @@ export default class Navigation extends Component {
         drawer.show();
       },
       icon: 'fas fa-bars',
-      ariaLabel: app.translator.trans('core.lib.nav.drawer_button'),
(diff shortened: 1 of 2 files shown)


davwheat (changes requested):
Thank you for the PR!

We'll need to use Flarum's translator for this.

Could you add a key in the `core.yml` file inside the locale folder (make sure it's added in the correct place alphabetically), then get the text with `app.translator.trans('core.<key>')` in the JS.

The best place would be under `core.lib`:

https://github.com/flarum/core/blob/2daee924c56a258a651697f7af2cf12f6a3d50f5/locale/core.yml#L501

grimmdude:
Got it, thanks @davwheat

locale/core.yml
```diff
@@ -539,6 +539,10 @@ core:
     loading_indicator:
       accessible_label: => core.ref.loading
 
+    # These translations are used in the loading navigation component.
```
davwheat, line 542:
```suggestion
    # These translations are used in the navigation header.
```

js/src/common/components/Navigation.js
```diff
@@ -93,6 +93,7 @@ export default class Navigation extends Component {
         drawer.show();
       },
       icon: 'fas fa-bars',
+      ariaLabel: app.translator.trans('core.lib.nav.drawer_button'),
```
davwheat, line 96:
```suggestion
      'aria-label': app.translator.trans('core.lib.nav.drawer_button'),
```

davwheat (changes requested):
Oops, `ariaLabel` also needs to be `aria-label`. Missed that when looking on my phone.

davwheat (approved):

Issue #3158, opened 2021-11-08
Scrolling notification list infinitely repeats


askvortsov1:
See https://discuss.flarum.org/d/29401-notification-list-is-infinitely-duplicated-leads-to-incorrect-count

https://youtu.be/uGpVSmVuDy4

PR #3159, opened 2021-11-08
Use calculated offset when loading page in PaginatedListState

askvortsov1:
**Fixes https://github.com/flarum/core/issues/3158**

Previously, the old offset was overriding the new one, so the same page was loaded over and over again. 

**Necessity**

**Confirmed**


SychO9 (approved):
Makes sense!

davwheat (approved):

PR #3161, opened 2021-11-11
Add 'Close' & 'Back' aria-label attributes to buttons.

grimmdude:
Adds `aria-label` attribute to modal close & back buttons to improve accessibility and satisfy these warnings.

Screen Shot 2021-11-11 at 10 08 32 AM

Affects every modal and the hero welcome component.

**Necessity**

**Confirmed**

### DIFF
--- a/js/src/common/components/Navigation.js
+++ b/js/src/common/components/Navigation.js
@@ -53,6 +53,7 @@ export default class Navigation extends Component {
         e.preventDefault();
         history.back();
       },
+      'aria-label': app.translator.trans('core.lib.close_button'),
     });
   }
 
--- a/js/src/common/components/Navigation.js
+++ b/js/src/common/components/Navigation.js
@@ -53,7 +53,7 @@ export default class Navigation extends Component {
         e.preventDefault();
         history.back();
       },
-      'aria-label': app.translator.trans('core.lib.close_button'),
+      'aria-label': app.translator.trans('core.lib.nav.back'),
     });
   }
 
--- a/js/src/common/components/Navigation.js
+++ b/js/src/common/components/Navigation.js
@@ -47,13 +47,12 @@ export default class Navigation extends Component {
       className: 'Button Navigation-back Button--icon',
       href: history.backUrl(),
       icon: 'fas fa-chevron-left',
-      title: previous.title,
+      'aria-label': previous.title,
       onclick: (e) => {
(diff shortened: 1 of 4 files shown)


js/src/common/components/Navigation.js
```diff
@@ -53,6 +53,7 @@ export default class Navigation extends Component {
         e.preventDefault();
         history.back();
       },
+      'aria-label': app.translator.trans('core.lib.close_button'),
```
davwheat: I think it would be more appropriate to have it as "Back" rather than "Close".

grimmdude: Agreed @davwheat , my mistake.  Just pushed that change.

js/src/forum/components/WelcomeHero.js
```diff
@@ -27,6 +27,7 @@ export default class WelcomeHero extends Component {
             icon: 'fas fa-times',
             onclick: slideUp,
             className: 'Hero-close Button Button--icon Button--link',
+            'aria-label': app.translator.trans('core.lib.close_button'),
```
davwheat, line 30: I'd be a fan of a more descriptive accessible label here, like "Hide welcome message". What do you think?

locale/core.yml
```diff
@@ -500,6 +500,7 @@ core:
   # Translations in this namespace are used by the forum and admin interfaces.
   lib:
     debug_button: Debug
+    close_button: => core.ref.close
```
davwheat, line 503: I'd prefer this to be listed under a `modal` key to ensure we don't use this in other components too. We try to keep to separate translation keys for each instance of text.

js/src/common/components/Navigation.js
```diff
@@ -53,6 +53,7 @@ export default class Navigation extends Component {
         e.preventDefault();
         history.back();
       },
+      'aria-label': app.translator.trans('core.lib.nav.back'),
```
SychO9, line 56: It'd be better to replace the title parameter above with `aria-label` instead, `previous.title` is already being implicitly used right now as `aria-label` and is more descriptive.

SychO9 (commented):
Thank you for the PR!

SychO9 (approved):
Looks good to me.

askvortsov1 (approved):
Thank you!

PR #3162, opened 2021-11-11
Enforce 65k character limit when attempting to update setting values.

### LINKED ISSUE #3081: Error accessing the forum after saving custom CSS with more than +68000 characters
## Bug Report

**Current Behavior**
Error when accessing the forum after saving in the admin panel a custom CSS code of +68000 characters (doesn't mean it's exactly that number, in my case I've experienced it when reaching that number of characters).

This causes that when saving the CSS code the last lines are cut, and therefore in my case for example this caused a `ParseError: missing closing'}'` error in the entire forum because it just cut the code before a `}`

**Steps to Reproduce**
1. Go to admin panel ->Appearance -> Edit custom CSS
2. Paste the CSS code
3. Save (no error will appear) 
After this if you reopen the Edit Custom CSS box you will see the cut code
4. Go to the forum and you will see an error (depending on where the CSS code is cut) or lack of changes

**Expected Behavior**
Avoid cutting the CSS code when it has ±68000 characters through the admin panel.

**Environment**
- Flarum version: 1.0.4

**Possible Solution (by @clarkwinkelmann)**

> We have a step to validate the Less but the code probably gets truncated when inserted in the database, so the validation won't catch it https://github.com/flarum/core/blob/master/src/Forum/ValidateCustomLess.php
> 
> We should probably add max length validation for all settings. This probably has never been an issue in the past because most settings are short strings and there was never any risk of reaching the `TEXT` limitation.

grimmdude:
**Fixes #3081**

Throws a validation error if attempting to update setting values greater than 65k characters.

Changing settings.

**Screenshot**

Screen Shot 2021-11-11 at 11 55 06 AM

Screen Shot 2021-11-11 at 11 55 13 AM

**Necessity**

**Confirmed**

### DIFF
--- a/src/Settings/SettingsServiceProvider.php
+++ b/src/Settings/SettingsServiceProvider.php
@@ -10,10 +10,6 @@
 namespace Flarum\Settings;
 
 use Flarum\Foundation\AbstractServiceProvider;
-use Flarum\Frontend\RecompileFrontendAssets;
-use Flarum\Forum\ValidateCustomLess;
-use Flarum\Locale\LocaleManager;
-use Flarum\Settings\Event\Saved;
 use Flarum\Settings\Event\Saving;
 use Flarum\Settings\SettingsValidator;
 use Illuminate\Contracts\Container\Container;
@@ -50,39 +46,12 @@ class SettingsServiceProvider extends AbstractServiceProvider
     }
 
 
-    public function boot(Container $container, Dispatcher $events, SettingsValidator $settingsValidator)
+    public function boot(Dispatcher $events, SettingsValidator $settingsValidator)
     {
         $events->listen(
-            Saved::class,
-            function (Saved $event) use ($container) {
-                $recompile = new RecompileFrontendAssets(
-                    $container->make('flarum.assets.forum'),
-                    $container->make(LocaleManager::class)
-                );
-                $recompile->whenSettingsSaved($event);
-
-                $validator = new ValidateCustomLess(
-                    $container->make('flarum.assets.forum'),
-                    $container->make('flarum.locales'),
-                    $container,
-                    $container->make('flarum.less.config')
-                );
-                $validator->whenSettingsSaved($event);
-            }
-        );
-
-        $events->listen(
             Saving::class,
-            function (Saving $event) use ($container, $settingsValidator) {
+            function (Saving $event) use ($settingsValidator) {
                 $settingsValidator->assertValid($event->settings);
-
-                $validator = new ValidateCustomLess(
-                    $container->make('flarum.assets.forum'),
-                    $container->make('flarum.locales'),
-                    $container,
-                    $container->make('flarum.less.config')
-                );
-                $validator->whenSettingsSaving($event);
             }
         );
     }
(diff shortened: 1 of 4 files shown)


askvortsov1 (changes requested):
Thank you for the PR! I think it might be more consistent with our current approach if the validation logic was put into a listener for `Saving::class` rather than directly in the controller.

askvortsov1 (changes requested):
Sorry I didn't mention this previously, but it would be preferable to wrap logic in a class, maybe `ValidateSettingLength`?

askvortsov1 (changes requested):
A few more comments, sorry for the back-and-forth:

- We register `ValidateCustomLess` in the `ForumServiceProvider` because it's diretly relevant to the forum frontend (custom LESS isn't used on the forum side). However, I think it would be better to register this code in the [SettingsServiceProvider](https://github.com/flarum/core/blob/5f110f73e7858741907665b1ec8b5c372be99796/src/Settings/SettingsServiceProvider.php#L17-L17), as this validation logic applies to all settings.
- As a convention, all our validators extend `AbstractValidator` (https://github.com/flarum/core/blob/312ff057f85d79548c35ec0f57459382d2f0d378/src/Foundation/AbstractValidator.php#L17-L17). I would prefer to either follow this pattern here, or for the new class to be an event listener (via invokable class).

grimmdude:
No problem @askvortsov1, thanks for the input.

grimmdude:
@askvortsov1, let me know what you think about that.  Moved those settings event listeners to `SettingsServiceProvider` and used a more conventional approach for the new validation.

src/Settings/SettingsServiceProvider.php
```diff
@@ -41,4 +48,42 @@ public function register()
...
+                $validator = new ValidateCustomLess(
+                    $container->make('flarum.assets.forum'),
+                    $container->make('flarum.locales'),
+                    $container,
+                    $container->make('flarum.less.config')
+                );
+                $validator->whenSettingsSaving($event);
+            }
+        );
+    }
```
askvortsov1, line 88: The logic for validating custom less and for recompiling frontend assets should remain in `ForumServiceProvider`, as it's specific to the forum. The only event listening in this file should be running the settings validator on saved settings.

src/Settings/SettingsValidator.php
```diff
@@ -0,0 +1,37 @@
...
+{
+    /**
+     * Make a new validator instance for this model.
+     *
+     * @param array $attributes
+     * @return \Illuminate\Validation\Validator
+     */
+    protected function makeValidator(array $attributes)
+    {
+        // These rules apply to all settings.
```
askvortsov1, line 24:
```suggestion
        // Entries in the default DB settings table are limited to 65,000
        // characters. We validate against this to avoid confusing errors.
```

askvortsov1 (changes requested):

askvortsov1:
> @askvortsov1, let me know what you think about that.  Moved those settings event listeners to `SettingsServiceProvider` and used a more conventional approach for the new validation.

Almost exactly what I intended, but in my previous comment I think I forgot to state that only the length validation logic should be moved; everything else is forum frontend dependent so it makes sense to keep it as is.

src/Settings/SettingsServiceProvider.php
grimmdude: OK, I misunderstood your original comment.

askvortsov1 (approved):
Thank you!

tankerkiller125 (approved):

PR #3163, opened 2021-11-11
Adjust boolean check of is_email_confirmed to suppress extraneous user activation events

### LINKED ISSUE #2713: Activated event is sent every time an email is confirmed instead of just once
## Bug Report

**Current Behavior**
The `Activated` event is dispatched every time an email change is confirmed, instead of just on the first email verification.

**Steps to Reproduce**
1. Use an extension that reacts to the activation event, for example Audit Log
2. Create a new account and activate it
3. Change email and confirm new email
4. Notice two logs for account activation

**Expected Behavior**
The `Activated` event shouldn't be triggered again once the account is already activated.

**Environment**
- Flarum version: beta 16, but also affects previous versions

**Possible Solution**
The problem is here

https://github.com/flarum/core/blob/509adf228a18745b61a62b63cbdce3198ac17ff3/src/User/User.php#L368

We use an explicit check `!== true`, but the value isn't cast to `bool`, so we're always comparing `1 !== true` and always dispatch the event.

I suggest we add an Eloquent attribute cast to make this value an actual boolean.

Another option could be to just use `!$this->is_email_confirmed` since the column is non-nullable anyway.

**Additional Context**
This is not a big issue in itself, but if any extension wanted to use that event to configure something for the user account, it would be nice if the event didn't get triggered again. Since the user can essentially trigger it at will, a malicious example could be the registration event used to reset the user balance or strikes in an external database, which could be abused.

grimmdude:
**Fixes #2713**

Instead of explicilty checking for a `true` value, adjust to check if `is_email_confirmed` _evaluates_ to `true`.  This matches the approach currently being used elsewhere in the `User` model, and will prevent the `Activated` event from being fired every time a user confirms their email.

Existing usage of this approach:
https://github.com/flarum/core/blob/5f110f73e7858741907665b1ec8b5c372be99796/src/User/User.php#L725-L727

Activating an already activated user.

**Necessity**

**Confirmed**

### DIFF
--- a/src/User/User.php
+++ b/src/User/User.php
@@ -358,7 +358,7 @@ class User extends AbstractModel
      */
     public function activate()
     {
-        if ($this->is_email_confirmed !== true) {
+        if (!$this->is_email_confirmed) {
             $this->is_email_confirmed = true;
 
             $this->raise(new Activated($this));
--- a/src/User/User.php
+++ b/src/User/User.php
@@ -358,7 +358,7 @@ class User extends AbstractModel
      */
     public function activate()
     {
-        if (!$this->is_email_confirmed) {
+        if (! $this->is_email_confirmed) {
             $this->is_email_confirmed = true;
 
             $this->raise(new Activated($this));


askvortsov1 (approved):
Thank you for (yet another) PR! I think this might be the most PRs we've seen from a new contributor in one day 😁. Approving pending style fixes.

SychO9 (approved):
Thank you for the PR!

Eventually, we should use an eloquent attribute cast to boolean, but probably in a major release.

PR #3167, opened 2021-11-19
feat/fix: allow replacing of blade template namespaces

davwheat:

@imorland deserves the credit for this, but I'll happily take it (unless it doesn't work, then blame him!).

- Allow replacing Blade template namespaces

This is, for example, needed for overriding the `forum.blade.php` file to inject custom HTML where needed, and not only where the `headerHtml` and `footerHtml` is included.

This PR will need tests! I'll try my best... 🙃

**_Is this a feature or a bug fix?_** Replacing blade views used to be possible during beta, but *I believe* when the shift was made to extenders, this functionality was lost.

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/src/Extend/View.php
+++ b/src/Extend/View.php
@@ -18,6 +18,7 @@ class View implements ExtenderInterface, LifecycleInterface
 {
     private $namespaces = [];
     private $replaceNamespaces = [];
+    private $prependNamespaces = [];
 
     /**
      * Register a new namespace of Laravel views.
--- a/src/Extend/View.php
+++ b/src/Extend/View.php
@@ -44,6 +44,8 @@ class View implements ExtenderInterface, LifecycleInterface
 
     /**
      * Override an existing namespace of Laravel views.
+     * 
+     * **If you're only replacing one or a few views, you probably want to `->prepend()` the new views instead.**
      *
      * Views are PHP files that use the Laravel Blade syntax for creation of server-side generated HTML.
      * Flarum core uses them for error pages, the installer, HTML emails, and the skeletons for the forum and admin sites.
@@ -61,11 +63,33 @@ class View implements ExtenderInterface, LifecycleInterface
         return $this;
     }
 
+    /**
+     * Extend an existing namespace of Laravel views.
+     *
+     * Views are PHP files that use the Laravel Blade syntax for creation of server-side generated HTML.
+     * Flarum core uses them for error pages, the installer, HTML emails, and the skeletons for the forum and admin sites.
+     * To replace views within a namespace, you will need to put them in a folder in your extension, and register that folder under an existing namespace.
+     *
+     * @param  string  $namespace: The name of the namespace.
+     * @param  string|string[]  $hints: This is a path (or an array of paths) to the folder(s)
+     *                               where view files are stored, relative to the extend.php file.
+     * @return self
+     */
+    public function prepend(string $namespace, $hints): self
+    {
+        $this->prependNamespaces[$namespace] = $hints;
+
+        return $this;
+    }
+
     public function extend(Container $container, Extension $extension = null)
     {
         $container->resolving(Factory::class, function (Factory $view) {
             foreach ($this->namespaces as $namespace => $hints) {
                 $view->addNamespace($namespace, $hints);
+            }
+            foreach ($this->prependNamespaces as $namespace => $hints) {
+                $view->prependNamespace($namespace, $hints);
             }
             foreach ($this->replaceNamespaces as $namespace => $hints) {
                 $view->replaceNamespace($namespace, $hints);
--- a/src/Extend/View.php
+++ b/src/Extend/View.php
@@ -15,7 +15,6 @@ use Illuminate\Contracts\Container\Container;
 use Illuminate\Contracts\View\Factory;
 
 /**
- * 
  * Views are PHP files that use the Laravel Blade syntax for creation of server-side generated HTML.
  *
  * Flarum's core uses them for error pages, the installer, HTML emails, and the skeletons for the forum and admin sites.
--- a/src/Extend/View.php
+++ b/src/Extend/View.php
@@ -22,7 +22,6 @@ use Illuminate\Contracts\View\Factory;
 class View implements ExtenderInterface, LifecycleInterface
 {
     private $namespaces = [];
-    private $replaceNamespaces = [];
     private $prependNamespaces = [];
 
     /**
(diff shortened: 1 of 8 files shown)


askvortsov1:
I'll want to see tests before reviewing, sorry.

Also, do ALL files in a namespace get replaced? That could be a lot of code duplication.

SychO9:
> This is, for example, needed for overriding the forum.blade.php file to inject custom HTML where needed, and not only where the headerHtml and footerHtml is included.

You can still replace that view using the `Frontend` extender's `content` method, example of replacing the `admin.blade.php`:
https://github.com/afrux/asirem/blob/main/extend.php#L38

Now that probably doesn't/can't be applied to other views such as errors, but I reckon there's probably a better way to go about replacing a specific view than replacing a whole namespace? especially if

> Also, do ALL files in a namespace get replaced? That could be a lot of code duplication.

askvortsov1:
We'll want to support both `prependNamespace` and `replaceNamespace`, although we might want to start with just `prepend` since that's more useful.

src/Extend/View.php
```diff
@@ -41,12 +43,58 @@ public function namespace(string $namespace, $hints): self
...
     }
 
+    /**
+     * Override an existing namespace of Laravel views.
+     *
+     * **If you're only replacing one or a few views, you probably want to `->prepend()` the new views instead.**
+     *
+     * Views are PHP files that use the Laravel Blade syntax for creation of server-side generated HTML.
+     * Flarum core uses them for error pages, the installer, HTML emails, and the skeletons for the forum and admin sites.
+     * To replace views, you will need to put them in a folder in your extension, and register that folder under an existing namespace.
```
askvortsov1, line 53: Let's factor out this explanation to the extender's docblock. Too much repetition IMO.

tests/integration/extenders/ViewTest.php
```diff
@@ -36,4 +36,30 @@ public function custom_view_namespace_can_be_added_by_extender()
...
+    public function can_prepend_view_namespace_by_extender()
+    {
+        $this->extend(
+            (new Extend\View)
+                ->prepend('flarum', dirname(__FILE__, 3).'/fixtures/views')
+        );
+
+        $this->assertEquals('<html><body>Hello World!</body></html>', trim($this->app()->getContainer()->make(Factory::class)->make('flarum::test')->render()));
+    }
 }
```
askvortsov1, line 65: What's the difference between these two? Can we add tests showing that `prepend` keeps the originals for other views, whereas `replace` replaces them?

askvortsov1 (changes requested):

tests/integration/extenders/ViewTest.php
davwheat: They both keep them, at least when trying replace on the Flarum namespace, which is why I'm very confused.

askvortsov1: Here's the problem:

https://github.com/flarum/core/blob/a57ef5a3d500f15be2253843209b5726cb36b22e/vendor/illuminate/support/ServiceProvider.php#L156-L177

The way we attach the Flarum paths is we [register the namespace on boot](https://github.com/flarum/core/blob/a57ef5a3d500f15be2253843209b5726cb36b22e/src/Frontend/FrontendServiceProvider.php#L147-L147), by calling the above function. But that uses `callAfterResolving`, which IIRC means that the namespace is added after the replace/prepend calls.

davwheat: Is there any way to easily fix that, or do you think it might be better to keep it like this, and mention so in the docblock.

IMO, this way results in someone being less likely to mess up Flarum, and someone can still remove all the views through the `extend()` method, no?

davwheat: Is there any way to easily fix that, or do you think it might be better to keep it like this, and mention so in the docblock.

IMO, this way results in someone being less likely to mess up Flarum, and someone can still remove all the views through the `extend()` method, no?

SychO9: isn't `prepend` enough for the use case we're dealing with though, I think we should just use introduce that and if there's ever a use case for a full replacement, we can worry about it then.

askvortsov1: Fair enough, that sounds good to me.

tests/integration/extenders/ViewTest.php
```diff
@@ -36,4 +36,36 @@ public function custom_view_namespace_can_be_added_by_extender()
...
+        );
+
+        /** @var Factory */
+        $viewFactory = $this->app()->getContainer()->make(Factory::class);
+
+        $this->assertEquals('<html><body>Hello World!</body></html>', trim($viewFactory->make('flarum::test')->render()));
+
+        // Original namespace hint has been replaced
+        $this->assertFalse($viewFactory->exists('flarum::frontend.app'));
+    }
```
davwheat, line 57:
```suggestion
```

askvortsov1 (changes requested):
Sorry, a few more requests:

Can we add a test case showing that you can override a view used by flarum, such as the homepage, in a full integration test?

Maybe we should rename `prepend` on the extender to `override` or `extendNamespace`?

davwheat:
Out of those options, I prefer `extendNamespace` because we aren't *technically* overriding even if it seems that way. It also gives us the ability to easily slot in `overrideNamespace` in the future if the need arises.

src/Forum/ForumServiceProvider.php
```diff
@@ -131,11 +131,11 @@ public function register()
...
     }
 
-    public function boot(Container $container, Dispatcher $events, Factory $view)
+    public function boot(Container $container, Dispatcher $events, ViewFactory $views)
     {
-        $this->loadViewsFrom(__DIR__.'/../../views', 'flarum.forum');
+        $views->addNamespace('flarum.forum', __DIR__.'/../../views');
 
-        $view->share([
+        $views->share([
```
SychO9, line 138: aren't these changes unnecessary now that we're not implementing override ?

davwheat: Yes, but wouldn't this make it easier for extensions to override Flarum's views, and be more consistent with how developers would expect views to be added?

SychO9: how so ? it does the same thing as before, the difference is just that before it added the namespace after the view factory was resolved, only reason I'm mentioning this is because I don't know why that was done that way, perhaps there is a good reason to avoid resolving the view factory too early. That's why I generally like to avoid unnecessary changes. Otherwise I really don't know to be honest. Just a rather stay safe since it's not a necessary change type of thing.

askvortsov1 (approved):

SychO9 (approved):

PR #3172, opened 2021-11-20
chore: use `@php` in Blade templates

davwheat:

As per title.

None

**Necessity**

**Confirmed**

### DIFF
--- a/views/layouts/basic.blade.php
+++ b/views/layouts/basic.blade.php
@@ -1,9 +1,8 @@
-{{-- TODO: Change below to @php when Laravel is upgraded --}}
 <?php
   $primaryColor = $settings->get('theme_primary_color', '#000');
 ?>
 
-<!doctype html>
+<!DOCTYPE html>
 <html>
   <head>
     <meta charset="utf-8">
--- a/views/layouts/basic.blade.php
+++ b/views/layouts/basic.blade.php
@@ -1,6 +1,6 @@
-<?php
+@php
   $primaryColor = $settings->get('theme_primary_color', '#000');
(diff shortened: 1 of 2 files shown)


askvortsov1:
Eh, not sure I agree with this. If we're taking action here, we might as well fix the issue.

davwheat:
I think "issue" is stretching it. As far as I can see, `@php` is just alternate syntax to `<?php`, isn't it?

views/layouts/basic.blade.php
```diff
@@ -1,9 +1,8 @@
-{{-- TODO: Change below to @php when Laravel is upgraded --}}
 <?php
   $primaryColor = $settings->get('theme_primary_color', '#000');
 ?>
```
davwheat, line 3:
```suggestion
@php
  $primaryColor = $settings->get('theme_primary_color', '#000');
@endphp
```

askvortsov1 (approved):

views/layouts/basic.blade.php
```diff
@@ -1,9 +1,8 @@
-{{-- TODO: Change below to @php when Laravel is upgraded --}}
-<?php
+@php
   $primaryColor = $settings->get('theme_primary_color', '#000');
-?>
+@endphp
```
SychO9, line 3: from local testing it seems like declaring variables within `@php` doesn't actually work for some reason 🤔

SychO9: oh, it works without the comment on top nvm

SychO9 (approved):

PR #3174, opened 2021-11-22
Typescript for models

askvortsov1:
**Supercedes https://github.com/flarum/core/pull/2411**
**Part of flarum/framework#3310**
**Part of flarum/issue-archive#96**

Sorry in advance, this is a long one.

Convert Model.ts, Store.ts, and all models to TypeScript.

As can be seen in the "Typecheck" GH action, this PR is sound from a type integrity perspective. There are, however a few concerns I have.

First and foremost, please take a look at the util types/interfaces I've added to `Store` and `Model`, their naming, and whether everything that needs to be exported is actually being exported.

Secondly, is there anything missing in the `Store` util types that should be there in order to comply with the JSON:API spec?

Thirdly, the types for `computed` are very crude. Is there a better way to do this in a typesafe way?

Fourthly, there currently isn't anything that enforces integrity of whether a model is saved being consistent with `exists`. One suggestion was to add a `Saved extends boolean` generic param to `Model`, and use conditional types to govern the types of `exists` and `data`. This makes things really messy though, as every subclass also needs to specify this, and `this` as a return type doesn't support swapping types when needed.

Fithly, dealing with singular vs plural types with `Store.pushPayload` is a pain. See `Application.preloadedApiDocument` and `Store.find` for an example. If anyone has ideas for a cleaner way to do this (or a way to constrain issues to `Store.pushPayload`) that would be fantastic.

Finally, and this is the biggest one, I've typed all model attributes as safely as I can: `| null` is added to all nullable fields, and `| undefined` is added to all fields not included on the model's `BaseSerializer`. This is a bit clunky and overly broad, but it's safe. One idea I had to get around this is the concept of `SerializerProfiles`: this would be a generic argument to models representing which fields are present on a given serializer, and their types. It would then be used to generate all typings on all attributes. This sounds great in concept, but when trying to implement it, I had to use a lot of very ugly mapped and conditional types, as well as some very verbose generics and naming. As a result, I don't think it's worth going down that route.

**Necessity**

**Confirmed**

**Required changes:**


js/src/common/Model.ts
```diff
@@ -0,0 +1,383 @@
...
+
+interface ModelAttributes {
+  [key: string]: unknown;
+}
+
+interface ModelRelationships {
+  [relationship: string]: {
+    data: ModelIdentifier | ModelIdentifier[];
+  };
+}
```
askvortsov1, line 18: Should these be exported?

js/src/common/Model.ts
```diff
@@ -0,0 +1,383 @@
...
+   * The data store that this resource should be persisted to.
+   */
+  protected store: Store | null;
+
+  /**
+   * @param data A resource object from the API.
+   * @param store The data store that this model should be persisted to.
+   * @public
+   */
+  constructor(data: ModelData = {}, store = null) {
```
askvortsov1, line 75: Why do we allow `store` to be nullable? Shouldn't we at least default it to `app.store`?

js/src/common/Model.ts
```diff
@@ -0,0 +1,383 @@
...
+  attribute<T = unknown>(attribute: string): T {
+    return this.data?.attributes?.[attribute] as T;
+  }
+
+  /**
+   * Merge new data into this model locally.
+   *
+   * @param data A resource object to merge into this model
+   */
+  pushData(data: ModelData | { relationships?: SaveRelationships }): this {
```
askvortsov1, line 103: I did a lot of refactoring on this method to make it more understandable, less hacky, and more typesafe. This one needs a bit of extra review.

js/src/common/Model.ts
```diff
@@ -0,0 +1,383 @@
...
+  /**
+   * Merge new attributes into this model, both locally and with persistence.
+   *
+   * @param attributes The attributes to save. If a 'relationships' key
+   *     exists, it will be extracted and relationships will also be saved.
+   * @public
+   */
+  save(
+    attributes: SaveAttributes,
+    options: Omit<FlarumRequestOptions<ApiPayloadSingle>, 'url'> & { meta?: any } = {}
```
askvortsov1, line 165: Do we actually use `meta` anywhere? If so, should we formalize our usage of it via an interface somewhere?

js/src/common/models/Discussion.ts
```diff
@@ -0,0 +1,146 @@
...
+import ItemList from '../utils/ItemList';
+import Badge from '../components/Badge';
+import Mithril from 'mithril';
+import Post from './Post';
+import User from './User';
+
+export default class Discussion extends Model {
+  title() {
+    return Model.attribute<string>('title').call(this);
+  }
```
SychO9, line 13: wouldn't it be more friendly for attributes/relationships to write instead:
```suggestion
  title = Model.attribute<string>('title');
```

SychO9 (commented):
Only reviewed about 65% of this so far :P

js/src/common/Model.ts
SychO9: I wouldn't change it unless something is broken and requires the change.

js/src/common/models/Discussion.ts
datitisev: @SychO9 This breaks monkey patching I believe. Class properties are not in `.prototype` and thus cannot be extended easily (if at all (?))

From Clark's PR
> The difference between `Object.assign`/`mixin` and the proposed solution is that the attributes are no longer part of the model `prototype` and as such cannot be easily overridden by extensions. However I don't think any extension ever attempted that and I'm not sure why the "mixin" style was used at all in the past.

More @ https://github.com/flarum/core/pull/2411#issuecomment-847427695

SychO9: ah shame :(, the easier the code is the easier it is for new extension devs to get into it. But being able to monkey patch these is important as well.

js/src/common/Model.ts
askvortsov1: On further inspection, I want to change it to `app.store`. It's just simpler this way.

js/src/common/models/Discussion.ts
askvortsov1: Personally, I don't think model attributes should really be monkey patched (that's a VERY unsafe approach to development), just that for now we need to maintain BC.

datitisev: It will create a difference between ext and core attributes, as we extend Models with `prototype`, and once we switch over to class properties, those would be separate from the prototype. Don't think this would cause issues, but it might be confusing. 

And yeah, think I'm for changing it in a breaking release as it's very ugly and repetitive.

askvortsov1: I'm hoping that by the time 2.0 rolls around, we will have proper JS extenders for this.

js/src/common/Store.ts
```diff
@@ -0,0 +1,243 @@
...
+
+export interface ApiPayloadPlural {
+  data: SavedModelData[];
+  included?: SavedModelData[];
+  links?: {
+    first: string;
+    next?: string;
+    prev?: string;
+  };
+}
```
SychO9, line 48: The meta I mentioned before was about the response payload here btw, not the request, didn't even know the request had that in the spec.

Pi-Bouf (approved):

SychO9 (approved):

js/src/common/models/Discussion.ts
```diff
@@ -0,0 +1,146 @@
...
+
+  /**
+   * Get the Badge components that apply to this discussion.
+   */
+  badges(): ItemList<Mithril.Children> {
+    const items = new ItemList<Mithril.Children>();
+
+    if (this.isHidden()) {
+      items.add('hidden', Badge.component({ type: 'hidden', icon: 'fas fa-trash', label: app.translator.trans('core.lib.badge.hidden_tooltip') }));
```
davwheat, line 134: JSX pretty please :) (Will need to rename the file though.)

```suggestion
      items.add('hidden', <Badge type="hidden" icon="fas fa-trash", label={app.translator.trans('core.lib.badge.hidden_tooltip')} />);
```

js/src/common/models/User.ts
```diff
@@ -0,0 +1,165 @@
...
+  /**
+   * Get the Badge components that apply to this user.
+   */
+  badges(): ItemList<Mithril.Children> {
+    const items = new ItemList<Mithril.Children>();
+    const groups = this.groups();
+
+    if (groups) {
+      groups.forEach((group) => {
+        items.add(`group${group?.id()}`, GroupBadge.component({ group }));
```
davwheat, line 119:
```suggestion
        items.add(`group${group?.id()}`, <GroupBadge group={group} />);
```

js/src/common/models/User.ts
```diff
@@ -0,0 +1,165 @@
...
+        // See https://github.com/lokesh/color-thief/issues/40
+        if (e instanceof TypeError) {
+          user.avatarColor = [255, 255, 255];
+        } else {
+          throw e;
+        }
+      }
+      user.freshness = new Date();
+      m.redraw();
+    };
```
davwheat, line 150: This seems to prevent TS running wild:

```suggestion
    image.addEventListener('load', function (this: HTMLImageElement) {
      try {
        const colorThief = new ColorThief();
        user.avatarColor = colorThief.getColor(this);
      } catch (e) {
        // Completely white avatars throw errors due to a glitch in color thief
        // See https://github.com/lokesh/color-thief/issues/40
        if (e instanceof TypeError) {
          user.avatarColor = [255, 255, 255];
        } else {
          throw e;
        }
      }
      user.freshness = new Date();
      m.redraw();
    });
```

davwheat (changes requested):

davwheat (approved):

PR #3175, opened 2021-11-22
Add typing files for our translator libraries

askvortsov1:
Title says it all.


davwheat (changes requested):
Should this not go in our global types folder?

askvortsov1:
> Should this not go in our global types folder?

No, as we don't export these libraries to be globally available

SychO9 (approved):

davwheat (approved):

PR #3177, opened 2021-11-22
Convert routes to Typescript

askvortsov1:
Converts `routes.js` files to Typescript, resolves issues with `app.route.X` not found.

### DIFF
--- a/js/src/forum/ForumApplication.ts
+++ b/js/src/forum/ForumApplication.ts
@@ -10,7 +10,7 @@ import Composer from './components/Composer';
 import DiscussionRenamedNotification from './components/DiscussionRenamedNotification';
 import CommentPost from './components/CommentPost';
 import DiscussionRenamedPost from './components/DiscussionRenamedPost';
-import routes from './routes';
+import routes, { makeRouteHelpers } from './routes';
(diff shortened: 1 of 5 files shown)


js/src/forum/ForumApplication.ts
```diff
@@ -73,10 +73,14 @@ export default class ForumApplication extends Application {
...
   discussions: DiscussionListState = new DiscussionListState({});
 
+  route: typeof Application.prototype.route & ReturnType<typeof makeRouteHelpers>;
+
   constructor() {
     super();
 
     routes(this);
+
+    this.route = Object.assign(Application.prototype.route.bind(this), makeRouteHelpers(this));
```
datitisev, line 83: We may want to use `this.constructor.route` instead? That way we don't make it specific to Application in case for some reason some extension wants to modify it. Even if not for that, just to make sure we're referencing the current subclass just in case.

askvortsov1: That exact syntax didn't work, but I changed it to use prototypes

SychO9 (approved):

davwheat (approved):

PR #3178, opened 2021-11-22
fix: Modal typings complicate inheritance

SychO9:
Not all modals contain forms needing these methods: `content` `onsubmit`.
Also exports the modal attrs interface so that inheritance works without errors.

I suppose a better solutions would be to have two separate modal components, a `Modal` and a `FormModal` but that would be breaking.

**Necessity**

**Confirmed**

### DIFF
--- a/js/src/common/components/Modal.tsx
+++ b/js/src/common/components/Modal.tsx
@@ -78,6 +78,9 @@ export default abstract class Modal<ModalAttrs extends IInternalModalAttrs = IIn
     }
   }
 
+  /**
+   * @todo split into FormModal and Modal in 2.0
+   */
   view() {
     if (this.alertAttrs) {
       this.alertAttrs.dismissible = false;
@@ -124,9 +127,7 @@ export default abstract class Modal<ModalAttrs extends IInternalModalAttrs = IIn
   /**
    * Get the content of the modal.
    */
-  content(): Mithril.Children {
-    return null;
-  }
+ abstract content(): Mithril.Children;
 
   /**
    * Handle the modal form's submit event.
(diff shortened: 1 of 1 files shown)


js/src/common/components/Modal.tsx
```diff
@@ -129,7 +129,9 @@ export default abstract class Modal<ModalAttrs = {}> extends Component<ModalAttr
   /**
    * Handle the modal form's submit event.
    */
-  abstract onsubmit(e: Event): void;
+  onsubmit(e: Event): void {
```
davwheat, line 132: ...while we're here! 😛

```suggestion
  onsubmit(e: SubmitEvent): void {
```

davwheat (approved):

davwheat (approved):

js/src/common/components/Modal.tsx
```diff
@@ -124,12 +124,16 @@ export default abstract class Modal<ModalAttrs = {}> extends Component<ModalAttr
   /**
    * Get the content of the modal.
    */
-  abstract content(): Mithril.Children;
+  content(): Mithril.Children {
```
askvortsov1, line 127: Shouldn't we require `content` of some form?

SychO9: `content` is called inside the form which is inside the default view, when you're implementing a non form modal, you'll implement the view method because of that. you won't need the default form JSX in the default view.

SychO9: oh wait no, you're right, you can't and shouldn't be reimplementing the view method, you're have to recreate the base jsx, content needs to be required, but that also means even if it isn't a form modal it'll be within a form, which isn't logical.

Something to fix in 2.0, I'll re-require the content method

askvortsov1 (approved):

PR #3180, opened 2021-11-23
fix: incorrect typings for Modal `hide()` method

davwheat:

The `hide()` method of the Modal class will have an event passed as the first param when clicking the Modal's close button due to how the onclick handler is set.

Should we change the `this.hide.bind(this)` to `() => this.hide()` instead of this? Could be breaking, but I'd argue this param was never public API, really.

**Necessity**

**Confirmed**


askvortsov1:
> Should we change the `this.hide.bind(this)` to `() => this.hide()` instead of this? Could be breaking, but I'd argue this param was never public API, really.

Is there a use case for using the original mouse event while closing the modal? I don't see one, so this might be preferable to keep things simpler.

askvortsov1 (approved):

SychO9 (approved):

Issue #3055, opened 2021-08-25
Avatar Upload throws a 500 error when file mimetype is incorrect but extension is correct


SychO9:
## Bug Report

**Current Behavior**
If you upload a file that isn't an image with its extension edited to that of one, the server returns a 500 error.

**Expected Behavior**
Should throw a validation error.

**Environment**
- Flarum version: 1.1-dev

PR #3181, opened 2021-11-25
Improve avatar upload experience

askvortsov1:
Fixes https://github.com/flarum/core/issues/3055

- On the frontend, accept only image types as a hint to the OS file picker
- On the backend, use `getimagesize` as an additional validation guard against non-image files with image extensions. This isn't necessary for security, but results in less confusing error mesages.

### DIFF
--- a/src/User/AvatarValidator.php
+++ b/src/User/AvatarValidator.php
@@ -67,8 +67,8 @@ class AvatarValidator extends AbstractValidator
         $guessedExtension = MimeTypes::getDefault()->getExtensions($file->getClientMediaType())[0] ?? null;
 
         $stream = $file->getStream();
-        $tmpFilename = $stream->getMetadata("uri");
-        $notImage = !getimagesize($tmpFilename);
+        $tmpFilename = $stream->getMetadata('uri');
+        $notImage = ! getimagesize($tmpFilename);
 
         if ($notImage || ! in_array($guessedExtension, $allowedTypes)) {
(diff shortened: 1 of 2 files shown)


davwheat (approved):

src/User/AvatarValidator.php
```diff
@@ -66,7 +66,11 @@ protected function assertFileMimes(UploadedFileInterface $file)
 
         $guessedExtension = MimeTypes::getDefault()->getExtensions($file->getClientMediaType())[0] ?? null;
 
-        if (! in_array($guessedExtension, $allowedTypes)) {
+        $stream = $file->getStream();
+        $tmpFilename = $stream->getMetadata('uri');
+        $notImage = ! getimagesize($tmpFilename);
+
+        if ($notImage || ! in_array($guessedExtension, $allowedTypes)) {
```
SychO9, line 73: `getimagesize` doesn't seem to be a reliable way of checking if the file is an image (https://www.php.net/manual/en/function.getimagesize.php)

I'm thinking it might be better to catch errors when creation the intervention image object, logging the error in the logs, and returning a better response to the end user.

askvortsov1: Changed to use that logic

SychO9: I wouldn't get rid of the mimes check though 🤔 ? the try catch should act as a second check.

tankerkiller125 (approved):

PR #3183, opened 2021-12-01
Use revision versioner to allow custom asset versioning

luceos:

Allows asset compilation to use a different version hasher implementation. This allows easier implementation of rev-manifest.json storage locations. Eg on our stack if we override where assets are stored, the rev-manifest.json is stored there too (s3). This causes each request to read from the s3 bucket as well. A better location for something like that would be either the settings table or redis. For normal local-file based Flarum instances, the default rev-manifest handler is sufficient.

Is this okay?

**Confirmed**

**Required changes:**

### DIFF
--- a/src/Frontend/Compiler/FileVersioner.php
+++ b/src/Frontend/Compiler/FileVersioner.php
@@ -14,7 +14,10 @@ use Illuminate\Support\Arr;
 
 class FileVersioner implements VersionerInterface
 {
-    protected Filesystem $filesystem;
+    /**
+     * @var Filesystem
+     */
+    protected $filesystem;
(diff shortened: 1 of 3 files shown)


askvortsov1:
Makes sense to me, but unfortunately we still have to stick with PHPDocs for class parameters until we fully drop 7.3.

askvortsov1 (commented):
Convert the property type to a docblock

luceos:
Habit be bad, will do so asap.

luceos:
> Convert the property type to a docblock

Done, anything else?

askvortsov1 (approved):

tankerkiller125 (approved):

Issue #2742, opened 2021-03-27
Allow extending XHR response error handling globally


matteocontrini:
As discussed [here](https://discuss.flarum.org/d/24975-handling-xhr-status-codes-globally), the handling of specific response status codes when doing client XHR requests cannot be currently extended.

It should be possible to show a custom error message depending on the response properties. A possible use-case is to handle a situation where a WAF returned a browser challenge to a API request, and the user wouldn't notice and instead receive a generic error message.

It's probably not a common use-case, so this is low priority and maybe not even a breaking change, but still something that could be improved and that I would find useful.

askvortsov1:
This should be relatively simple, just a matter of rewriting https://github.com/flarum/core/blob/1c4817a0b39a2b1e4a7690e8065df4ef6e1894aa/js/src/common/Application.js#L286-L286 so the error handling logic is in a separate, overridable/extensible method.

matteocontrini:
Still nice to have IMO.

matteocontrini:
So if I wanted to show a specific alert message for a specific status code what should I do now? Override the whole `requestErrorCatch` method? 🤔

askvortsov1:
> So if I wanted to show a specific alert message for a specific status code what should I do now? Override the whole `requestErrorCatch` method? thinking

For now, yes. I'd just check for the error code matching the one you're looking for, and if it's anything else, default to the original implementation.

I suppose at a later time we could refactor this so that current handling is more of a factory and you can add handlers for custom error codes, but I'd like to get a system of JS extenders in place before we start on that.

PR #3184, opened 2021-12-01
Split up application error handling

askvortsov1:
Fixes https://github.com/flarum/core/issues/2742

- Use a `catch` instead of a `then` with a vacuous `onfulfilled`
- Instead of setting a default `errorHandler` that throws the error, just run the fallback handler if no error handler is explicitly provided
- Split `requestErrorCatch` and `requestErrorDefaultHandler` into separate methods. 
- Clean up some messy duplicate code with `formattedErrors` and `content` on 422 errors
- Properly typehint `AlertAttrs`

### DIFF
--- a/js/src/common/utils/RequestError.ts
+++ b/js/src/common/utils/RequestError.ts
@@ -1,5 +1,5 @@
 import type Mithril from 'mithril';
-import { AlertAttrs } from '../components/Alert';
+import type { AlertAttrs } from '../components/Alert';
 
 export type InternalFlarumRequestOptions<ResponseType> = Mithril.RequestOptions<ResponseType> & {
   url: string;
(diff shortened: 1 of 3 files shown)


Pi-Bouf (approved):

js/src/common/utils/RequestError.ts
```diff
@@ -1,7 +1,7 @@
 import type Mithril from 'mithril';
+import { AlertAttrs } from '../components/Alert';
```
davwheat, line 2:
```suggestion
import type { AlertAttrs } from '../components/Alert';
```

davwheat (approved):

SychO9 (approved):

PR #3186, opened 2021-12-02
Move colorItems to ItemList

imorland:
At present, if an extension or theme wishes to add/remove any items from `Colors`, dom traversal is required.
To allow for extensions/themes to better integrate with the `Appearance Page`, each element under `Colors` is now built within an `ItemList`.

**Screenshot**
Visibly, no changes
image

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/js/src/admin/components/AppearancePage.js
+++ b/js/src/admin/components/AppearancePage.js
@@ -82,7 +82,7 @@ export default class AppearancePage extends AdminPage {
     items.add('helpText', <div className="helpText">{app.translator.trans('core.admin.appearance.colors_text')}</div>, 80);
 
     items.add(
-      'color-input',
+      'theme-colors',
       <div className="AppearancePage-colors-input">
         {this.buildSettingComponent({
           type: 'color-preview',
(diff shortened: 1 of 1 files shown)


davwheat (approved):

SychO9:
if an extension/theme is adding a new color variable, shouldn't that be in the extension's own settings page 🤔 ?

imorland:
A couple of potential use cases:

- `fof/nightmode` may want to remove the `dark mode` toggle, as this _might_ be confusing to have that active/available when "auto night mode" is active via that extension

- A possible future extension could replace the text input for color hex with a color picker

js/src/admin/components/AppearancePage.js
```diff
@@ -102,6 +76,53 @@ export default class AppearancePage extends AdminPage {
     ];
   }
 
+  colorItems() {
+    const items = new ItemList();
+
+    items.add('helpText', <div className="helpText">{app.translator.trans('core.admin.appearance.colors_text')}</div>, 80);
+
+    items.add(
+      'color-input',
```
askvortsov1, line 85:
```suggestion
      'theme-colors',
```

askvortsov1 (approved):

SychO9 (approved):

askvortsov1:
A few thoughts before merging.

I'm fine with this change for now, but I want to see a big revamp of the admin dashboard as part of 2.0. To the point where we might want to revert some of the changes we made in v1.0. Some particular concerns:

- It's not clear whether extensions should be putting settings on their own pages, or extending other extension pages / core pages.
- Some extensions clearly need their own pages, others really don't. There's no visual indication in the UI which is which.
- Some extensions are really headless plugins rather than full-fledged modules, it would be nice to distinguish between the two.

PR #3189, opened 2021-12-07
Make StatusWidget tools dropdown extensible

rafaucau:

This PR allows you to more easily modify the tools dropdown in the `StatusWidget` component using `ItemList`.

Until now, if you wanted to add something to this dropdown, you had to push a child:
```js
 extend(StatusWidget.prototype, 'items', (items) => {
    const tools = items.get('tools');
    tools.children.push(<Button>Test</Button>);
  });
```
If you want to remove / replace a given item it gets harder.
Now such code is enough:
```js
 extend(StatusWidget.prototype, 'toolsItems', (items) => {
    items.add('test', <Button>Test</Button>);
  });
```

Confirm that everything works.

**Screenshot**

image

**Necessity**

**Confirmed**

### DIFF
--- a/js/src/admin/components/StatusWidget.js
+++ b/js/src/admin/components/StatusWidget.js
@@ -26,13 +26,24 @@ export default class StatusWidget extends DashboardWidget {
         buttonClassName="Button"
         menuClassName="Dropdown-menu--right"
       >
-        <Button onclick={this.handleClearCache.bind(this)}>{app.translator.trans('core.admin.dashboard.clear_cache_button')}</Button>
+        {this.toolsItems().toArray()}
       </Dropdown>
     );
 
     items.add('version-flarum', [<strong>Flarum</strong>, <br />, app.forum.attribute('version')]);
     items.add('version-php', [<strong>PHP</strong>, <br />, app.data.phpVersion]);
     items.add('version-mysql', [<strong>MySQL</strong>, <br />, app.data.mysqlVersion]);
+
+    return items;
+  }
+
+  toolsItems() {
+    const items = new ItemList();
+
+    items.add(
+      'clearCache',
+      <Button onclick={this.handleClearCache.bind(this)}>{app.translator.trans('core.admin.dashboard.clear_cache_button')}</Button>
+    );
 
     return items;
   }


js/dist-typings/admin/components/StatusWidget.d.ts
```diff
@@ -1,5 +1,6 @@
 export default class StatusWidget extends DashboardWidget {
     items(): ItemList<any>;
+    toolsItems(): ItemList<any>;
```
davwheat, line 3: This file will be automatically updated. Could we remove the changes to it from this PR, please?

js/src/admin/components/StatusWidget.js
```diff
@@ -37,6 +37,17 @@ export default class StatusWidget extends DashboardWidget {
     return items;
   }
 
+  toolsItems() {
+    const items = new ItemList();
```
davwheat, line 41:
```suggestion
    const items = new ItemList<Mithril.Children>();
```

(Also need to add `import type Mithril from 'mithril';` to the top of the file)

davwheat (changes requested):
Thank you for the PR! Just some minor comments from me.

I know @askvortsov1 doesn't want many changes to the Admin dashboard at the moment, since it'll be getting a rewrite in the near future, so I'd appreciate his input.

Personally, I think we can make changes up until v1.2. That way, when rewriting, we can use the existing dashboard as a benchmark to ensure features or extensibility aren't accidentally dropped between the two, and to make the overall migration easier.

davwheat:
Also it looks like you've rebased this incorrectly, as there are some changes to the file versioner since I began writing my review. Could you make sure those are fixed, too?

rafaucau:
Wait, because I wanted to correct a typo in the commit name and it messed up and attached someone else's files :D

rafaucau:
Uhhh, managed to fix my commits 🤦‍♂️

js/src/admin/components/StatusWidget.js
rafaucau: But this is a JS file, not a TS file

js/dist-typings/admin/components/StatusWidget.d.ts
rafaucau: Ok, in the new commits it is no longer there

js/src/admin/components/StatusWidget.js
davwheat: Ermmmmm you saw nothing! It's been a long night, sorry. 🙈

davwheat (approved):

askvortsov1 (approved):
We might change things eventually, but for now I'm glad for this to be more extensible.

PR #3190, opened 2021-12-09
feat: add Less custom function extender, `is-extension-enabled` function

davwheat:

- create a new custom less function function on our Theme extender
- creates a new Less function for applying styles only when a specific extension is enabled

This prevents stylesheet bloats from extensions which extend multiple extensions, by only adding styles which are needed.

**Extender usage:**
```php
    (new Extend\Theme())
        ->addCustomLessFunction('test', function () {
            return new \Less_Tree_Quoted('', 'true');
        })
```

**Is ext enabled func usage:**

```less
// Is extension enabled function:
body when (is-extension-enabled('flarum-tags') = true) {
  // enabled
}
body when (is-extension-enabled('flarum-tags') = false) {
  // not enabled
}
```

(no longer applicable)
Would a simpler syntax be better, such as:

```less
.if-extension-enabled(flarum-tags, {
  // ...
});
```

**Necessity**

**Confirmed**

**Required changes:**


less/common/mixins/extension-styles.less
```diff
@@ -0,0 +1,7 @@
+.if-extension-enabled(@extension) {
+  // Placeholder mixin
```
askvortsov1, line 2: Does this still need an implementation?

askvortsov1 (changes requested):
> Would a simpler syntax be better, such as:

I think so. What's the benefit of doing it the other way?

less/common/mixins/extension-styles.less
davwheat: No. If no extension defines this, the Less compilation will error, so we need to make sure it is defined at least once.

davwheat:
> > Would a simpler syntax be better, such as:
> 
> I think so. What's the benefit of doing it the other way?

There's no inherent benefit of either way, I don't believe. Just syntactically different.

davwheat:
So I've pushed an alternate syntax which I think may be better.

The original syntax required that the mixin be declared in the root scope, meaning that

```less
// This would work
.if-extension-enabled(@ext) when (@ext = flarum-tags) {
  body { background: black; }
}

// ...but this would NOT
body {
  .if-extension-enabled(@ext) when (@ext = flarum-tags) {
    background: black;
  }
}
```

The new syntax, however, does support this:

```less
// This would work
.when-extension-enabled(flarum-tags, {
  body { background: black; }
});

// ...and so does this
body {
  .when-extension-enabled(flarum-tags, {
    background: black;
  });
}
```

**However,** the newer method will be significantly less performant, since it performs a loop over all enabled extensions each time it is called, rather than only doing it once.

We could reduce that by setting a less variable per enabled extension (e.g. `@extension-enabled--flarum-tags`), then use Less's variable-variables support to check for these, but that seems a bit hacky, imo.

SychO9:
What worries me about this is that, even without this, the Less compiler breaks the forum sometimes by running out of memory, it has been reported multiple times before, I've even experienced it locally, so I'm personally very reluctant to introducing anything that could reduce its performance even less.

How about registering a custom function instead ? I haven't actually tested the difference (*yet*) but I'm assuming it would be a more efficient solution. (tested locally and it works).

```less
.App {
  & when(is-extension-enabled('flarum-tags') = true) {
    opacity: 0.3;
  }
}
```

```php
$parser->registerFunction('is-extension-enabled', function (Less_Tree_Quoted $extensionId) {
    return new Less_Tree_Quoted('', in_array($extensionId->value, ['flarum-tags', 'flarum-blog']) ? 'true' : 'false');
});
```

davwheat:
That does seem like a much better idea.

When I saw the documentation of less-php state that custom functions weren't supported, I assumed it meant completely, but if we can define them in PHP, that's perfect!

davwheat:
@SychO9 Should be done now :)

askvortsov1 (approved):
Overall makes sense, but is the `= true` / `= false` really necessary? Is there no "not" symbol

src/Frontend/FrontendServiceProvider.php
```diff
@@ -135,6 +135,15 @@ function (Container $container) {
...
                 ],
+                'extensions-enabled'     => [
+                    'key' => 'extensions_enabled',
+                    'callback' => function ($value) {
+                        /** @var array */
+                        $extensionsEnabled = json_decode($value, true);
+
+                        return implode(',', $extensionsEnabled);
+                    },
+                ],
```
SychO9, line 146: unneeded less variable

src/Extend/Theme.php
```diff
@@ -52,6 +53,32 @@ public function overrideFileSource(string $file, string $newFilePath, string $ex
...
 
+    /**
+     * This method allows you to add custom Less functions.
+     *
+     * **Example usage:**
+     * ```php
+     * use
+     *
+     * (new Extend\Theme)
+     *     ->addLessFunction('is_flarum', function (Less_Tree_Quoted $text) {
```
SychO9, line 64: emphasis on the naming of the function needing to be dashed case, I think that's the only way it worked locally for me.

SychO9: additionally, I would rather simplify the callback to only take a normal mixed type value and return any mixed type value, and we transform those values properly, to have a layer of abstraction between the less parser's implementation and the purpose of the extender.

SychO9 (changes requested):

src/Extend/Theme.php
```diff
@@ -52,6 +53,53 @@ public function overrideFileSource(string $file, string $newFilePath, string $ex
...
+     * @param string $functionName Name of the function identifier.
+     * @param callable $callable The PHP function to run when the Less function is called.
+     * @return self
+     */
+    public function addCustomLessFunction(string $functionName, callable $callable): self
+    {
+        if (! is_callable($callable)) {
+            return $this;
+        }
+
```
askvortsov1, line 78: Shouldn't this be impossible due to dynamic type checking?

src/Extend/Theme.php
```diff
@@ -52,6 +53,53 @@ public function overrideFileSource(string $file, string $newFilePath, string $ex
...
+                return new \Less_Tree_Dimension($return);
+            }
+
+            // Strings
+            if (is_string($return)) {
+                return new \Less_Tree_Quoted('', $return);
+            }
+
+            // Anything else
+            return new \Less_Tree_Quoted('', (string) $return);
```
askvortsov1, line 97: What if the function returns `Less_Tree_Quoted already, as per the example in the docblock?

src/Frontend/Assets.php
```diff
@@ -42,6 +44,11 @@ class Assets
      */
     protected $assetsDir;
 
+    /**
+     * @var SettingsRepositoryInterface
+     */
+    protected $settings;
```
askvortsov1, line 50: Were you planning on adding this to the constructor?

src/Frontend/Assets.php
```diff
@@ -62,12 +69,18 @@ class Assets
      */
     protected $fileSourceOverrides = [];
 
-    public function __construct(string $name, Filesystem $assetsDir, string $cacheDir = null, array $lessImportDirs = null)
+    /**
+     * @var array
+     */
+    protected $customFunctions = [];
+
+    public function __construct(string $name, Filesystem $assetsDir, string $cacheDir = null, array $lessImportDirs = null, array $customFunctions = null)
```
askvortsov1, line 77: Why not have `[]` the default for `customFunctions`?

src/Frontend/Assets.php
```diff
@@ -173,6 +186,22 @@ protected function makeLessCompiler(string $filename): LessCompiler
...
+                $raw = $settings->get('extensions_enabled');
+                $extensionsEnabled = json_decode($raw);
+
+                return new Less_Tree_Quoted('', in_array($extensionId->value, $extensionsEnabled) ? 'true' : 'false');
+            }
+        ];
+
+        $this->customFunctions = array_merge($coreFunctions, $this->customFunctions);
+
+        $compiler->setCustomFunctions($this->customFunctions);
```
askvortsov1, line 203: Why not move this logic to a singleton on the Frontend service provider? That's a better place to have config like this IMO.

src/Frontend/Compiler/LessCompiler.php
```diff
@@ -99,6 +109,12 @@ protected function compile(array $sources): string
             }
         }
 
+        foreach ($this->customFunctions as $name => $callback) {
+            if (is_callable($callback)) {
```
askvortsov1, line 113: Again, don't they have to be?

tests/integration/extenders/ThemeTest.php
```diff
@@ -104,4 +104,32 @@ public function theme_extender_override_file_source_works_by_failing_when_necess
...
+    {
+        $this->extend(
+            (new Extend\Frontend('forum'))
+                ->css(__DIR__.'/../../fixtures/less/custom_function.less'),
+            (new Extend\Theme)
+                ->addCustomLessFunction('is-flarum', function ($text) {
+                    return strtolower($text) === 'flarum' ? 'true' : 100;
+                })
+                ->addCustomLessFunction('is-gt', function ($a, $b) {
+                    return $a > $b === 'flarum' ? true : false;
```
askvortsov1, line 121: Why is `=== 'flarum' in here?

askvortsov1: Also shouldn't the returned values be strings, not raw bools?

src/Extend/Theme.php
```diff
@@ -52,6 +53,53 @@ public function overrideFileSource(string $file, string $newFilePath, string $ex
...
+        }
+
+        $this->customFunctions[$functionName] = function (...$args) use ($callable) {
+            $argVals = array_map(function ($arg) {
+                return $arg->value;
+            }, $args);
+
+            $return = $callable(...$argVals);
+
+            // Numbers
```
askvortsov1, line 86: Should there be a case for booleans?

askvortsov1 (changes requested):

src/Extend/Theme.php
```diff
@@ -52,6 +53,53 @@ public function overrideFileSource(string $file, string $newFilePath, string $ex
...
     }
 
+    /**
+     * This method allows you to add custom Less functions.
+     *
+     * **Example usage:**
+     * ```php
+     * (new Extend\Theme)
+     *     ->addCustomLessFunction('is-flarum', function (mixed $text) {
+     *         return new Less_Tree_Quoted('', strtolower($text) === 'flarum' ? 'true' : 'false')
```
davwheat, line 63:
```suggestion
     *         return strtolower($text) === 'flarum'
```

src/Extend/Theme.php
davwheat: Yes...
```suggestion
```

src/Extend/Theme.php
davwheat, line 86:
```suggestion
            // Booleans
            if (is_bool($return)) {
                return new \Less_Tree_Quoted('', $return ? 'true' : 'false');
            }

            // Numbers
```

src/Extend/Theme.php
davwheat: They shouldn't return Less tree items. Docblock should have been updated.

src/Frontend/Assets.php
davwheat: Yes, but I ran into some issues I didn't understand, so swapped to resolve. If you can get it working... 🤞

src/Frontend/Assets.php
davwheat:
```suggestion
    public function __construct(string $name, Filesystem $assetsDir, string $cacheDir = null, array $lessImportDirs = null, array $customFunctions = [])
```

src/Frontend/Assets.php
```diff
@@ -62,12 +69,18 @@ class Assets
...
+     */
+    protected $customFunctions = [];
+
+    public function __construct(string $name, Filesystem $assetsDir, string $cacheDir = null, array $lessImportDirs = null, array $customFunctions = null)
     {
         $this->name = $name;
         $this->assetsDir = $assetsDir;
         $this->cacheDir = $cacheDir;
         $this->lessImportDirs = $lessImportDirs;
+        $this->customFunctions = $customFunctions ?? [];
```
davwheat, line 83:
```suggestion
        $this->customFunctions = $customFunctions;
```

src/Frontend/Compiler/LessCompiler.php
davwheat, line 113:
```suggestion
```

src/Frontend/Compiler/LessCompiler.php
```diff
@@ -99,6 +109,12 @@ protected function compile(array $sources): string
...
+                $parser->registerFunction($name, $callback);
+            }
```
davwheat, line 115:
```suggestion
```

tests/integration/extenders/ThemeTest.php
davwheat: 1. I have no idea.
2. No, we should transform them appropriately, hence the test.

davwheat:
```suggestion
                    return $a > $b;
```

src/Frontend/Assets.php
davwheat: Done.

src/Extend/Theme.php
```diff
@@ -52,6 +53,54 @@ public function overrideFileSource(string $file, string $newFilePath, string $ex
...
     }
 
+    /**
+     * This method allows you to add custom Less functions.
+     *
+     * **Example usage:**
+     * ```php
+     * (new Extend\Theme)
+     *     ->addCustomLessFunction('is-flarum', function (mixed $text) {
+     *         return strtolower($text) === 'flarum'
```
askvortsov1, line 63: I feel like we should document which return types are valid, and maybe specify that in the docblock?

src/Extend/Theme.php
```diff
@@ -52,6 +53,54 @@ public function overrideFileSource(string $file, string $newFilePath, string $ex
...
+            if (! is_string($return) && is_numeric($return)) {
+                return new \Less_Tree_Dimension($return);
+            }
+
+            // Strings
+            if (is_string($return)) {
+                return new \Less_Tree_Quoted('', $return);
+            }
+
+            // Anything else
```
askvortsov1, line 97: I wonder if this should error instead of trying to cast: for example, arrays would cause problems here.

src/Frontend/Assets.php
```diff
@@ -225,4 +235,9 @@ public function addFileSourceOverrides(array $fileSourceOverrides)
     {
         $this->fileSourceOverrides = array_merge($this->fileSourceOverrides, $fileSourceOverrides);
     }
+
+    public function addCustomLessFunctions(array $customFunctions)
+    {
+        $this->customFunctions = array_merge($this->customFunctions, $customFunctions);
+    }
```
askvortsov1, line 242: Why make this plural out of curiosity?

src/Frontend/Assets.php
```diff
@@ -173,6 +179,10 @@ protected function makeLessCompiler(string $filename): LessCompiler
             $compiler->setFileSourceOverrides($this->fileSourceOverrides);
         }
 
+        $this->customFunctions = array_merge(resolve('flarum.frontend.default_preloads'), $this->customFunctions);
+
+        $compiler->setCustomFunctions($this->customFunctions);
```
askvortsov1, line 184: Why not `extend` the singleton and then inject the full set of custom functions into the compiler's constructor? I fear that this will end up overriding the functions added by other instances of the same extender.

askvortsov1 (changes requested):

src/Frontend/Assets.php
davwheat: Because you can pass an array with multiple functions in.

E.g.

```php
addCustomLessFunctions([
  'funcA' => ...,
  'funcB' => ...,
])

File: src/Extend/Theme.php (hunk 15 above)
Comment (davwheat) [reply src/Extend/Theme.php:63]: Good idea

File: src/Frontend/Assets.php (hunk 18 above)
Comment (davwheat) [reply src/Frontend/Assets.php:184]: Mainly because I don't know how I'd go about doing that whatsoever.

File: src/Frontend/Assets.php (hunk 17 above)
Comment (askvortsov1) [reply src/Frontend/Assets.php:242]: Right, but why not

`addCustomLessFunction()->addCustomLessFunction()->addCustomLessFunction()`

Comment (SychO9) [reply src/Frontend/Assets.php:242]: I think you're confusing this method for the extender one @askvortsov1 lol

File: src/Frontend/Assets.php (hunk 18 above)
Comment (askvortsov1) [reply src/Frontend/Assets.php:184]: See https://github.com/flarum/core/blob/dc48e2327b1212c3b22fa57b23550f64b24632be/src/Extend/Auth.php#L63-L75  for an example of extending a singleton, and https://github.com/flarum/core/blob/dc48e2327b1212c3b22fa57b23550f64b24632be/src/Frontend/FrontendServiceProvider.php#L31-L31 for where `Assets` is instantiated (and so where to add constructor args).

Comment (davwheat) [reply src/Frontend/Assets.php]: Think it should be done now...

File: src/Extend/Theme.php [hunk 19]
```diff
@@ -52,8 +54,61 @@ public function overrideFileSource(string $file, string $newFilePath, string $ex
...
+            if (is_string($return)) {
+                return new \Less_Tree_Quoted('', $return);
+            }
+
+            // Numbers
+            if (is_numeric($return)) {
+                return new \Less_Tree_Dimension($return);
+            }
+
+            throw new RuntimeException('Custom Less functions must only return a string, number or boolean.');
```
askvortsov1, line 100: Nitpick: maybe include the name of the function?

askvortsov1 (approved):

src/Frontend/FrontendServiceProvider.php
```diff
@@ -109,6 +111,22 @@ function (Container $container) {
...
+        $this->container->singleton(
+            'flarum.frontend.custom_less_functions',
+            function (Container $container) {
+                $extensionsEnabled = json_decode($container->make(SettingsRepositoryInterface::class)->get('extensions_enabled'));
+
+                // Please note that these functions do not go through the same transformation which the Theme extender's
+                // `addCustomLessFunction` method does. You'll need to use the correct Less tree return type, and get
+                // parameter values with `$arg->value`.
+                return [
+                    'is-extension-enabled' => function ($extensionId) use ($extensionsEnabled) {
```
SychO9, line 123:
```suggestion
                    'is-extension-enabled' => function ($extensionId) use (\Less_Tree_Quoted $extensionsEnabled) {
```

src/Extend/Theme.php
```diff
@@ -52,8 +54,61 @@ public function overrideFileSource(string $file, string $newFilePath, string $ex
...
+    public function addCustomLessFunction(string $functionName, callable $callable): self
+    {
+        $this->customFunctions[$functionName] = function (...$args) use ($callable) {
+            $argVals = array_map(function ($arg) {
+                return $arg->value;
+            }, $args);
+
+            $return = $callable(...$argVals);
+
+            // Booleans
```
SychO9, line 85:
```suggestion
```

src/Extend/Theme.php
```diff
@@ -52,8 +54,61 @@ public function overrideFileSource(string $file, string $newFilePath, string $ex
...
+            if (is_bool($return)) {
+                return new \Less_Tree_Quoted('', $return ? 'true' : 'false');
+            }
+
+            // Strings
```
SychO9, line 90:
```suggestion
```

src/Extend/Theme.php
```diff
@@ -52,8 +54,61 @@ public function overrideFileSource(string $file, string $newFilePath, string $ex
...
+            if (is_string($return)) {
+                return new \Less_Tree_Quoted('', $return);
+            }
+
+            // Numbers
```
SychO9, line 95:
```suggestion
```

SychO9 (approved):
LGTM!

just nitpicks, because they're fun :p

src/Frontend/FrontendServiceProvider.php
SychO9, line 123:
```suggestion
                    'is-extension-enabled' => function (\Less_Tree_Quoted $extensionId) use ($extensionsEnabled) {
```

src/Extend/Theme.php
```diff
@@ -52,8 +54,61 @@ public function overrideFileSource(string $file, string $newFilePath, string $ex
...
+     *
+     * @see https://leafo.net/lessphp/docs/#custom_functions
+     *
+     * @param string $functionName Name of the function identifier.
+     * @param callable $callable The PHP function to run when the Less function is called.
+     * @return self
+     */
+    public function addCustomLessFunction(string $functionName, callable $callable): self
+    {
+        $this->customFunctions[$functionName] = function (...$args) use ($callable) {
```
davwheat, line 78:
```suggestion
        $this->customFunctions[$functionName] = function (...$args) use ($callable, $functionName) {
```

src/Extend/Theme.php
davwheat:
```suggestion
            throw new RuntimeException('Custom Less function `'.$functionName.'` must only return a string, number or boolean.');
```

PR #3193, opened 2021-12-12
fix: getPlainContent causes external content to be fetched

imorland:
`getPlainContent` renders the element using `jQuery`, which when an image is included in the content, causes this image to be fetched from the server unneccessarily. 

Impact can be seen for example when a post `excerpt` is present, either from `flarum/sticky` or `ianm/synopsis`. If several posts are stickied, or if synopsis is enabled, many images or videos, etc are fetched.

Taking advice from @davwheat , I've refactored this util to not use jQuery at all.

`js` is not my strong suit, this appears to work, but please let me know if I've missed something!

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/js/src/common/utils/string.ts
+++ b/js/src/common/utils/string.ts
@@ -37,7 +37,7 @@ export function getPlainContent(string: string): string {
     });
   });
 
-  return element.textContent.replace(/\s+/g, ' ').trim();
+  return element.innerText.replace(/\s+/g, ' ').trim();
 }
 
 /**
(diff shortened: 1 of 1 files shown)


js/src/common/utils/string.ts
```diff
@@ -28,11 +28,16 @@ export function slug(string: string): string {
...
-  dom.find(getPlainContent.removeSelectors.join(',')).remove();
+  getPlainContent.removeSelectors.forEach((selector) => {
+    const el = element.querySelectorAll(selector);
+    el.forEach((e) => {
+      e.remove();
+    });
+  });
 
-  return dom.text().replace(/\s+/g, ' ').trim();
+  return element.textContent.replace(/\s+/g, ' ').trim();
```
davwheat, line 40: We should use innerText here, as it is aware of styling, so won't include hidden elements.

davwheat (changes requested):

davwheat (approved):

askvortsov1 (approved):

PR #3194, opened 2021-12-12
Make SlugManager available to blade template

imorland:
**Fixes #0000**

I discovered that, it is not possible to use `SlugManager` within a mail template. Example use case: https://github.com/FriendsOfFlarum/subscribed/blob/be01ea2b2259733c5e59b970e8bb6172243982bc/resources/views/emails/userCreated.blade.php#L5

This template errors if the user slugger is not `default`.

A solution is ready, but requires this PR: https://github.com/FriendsOfFlarum/subscribed/pull/15

Make `SlugManager` available to `blade` templates

I also considered adding an alias, something like `flarum.slug.manager` here, and resolving from that in the template, but I think the above solution is probably more dev friendly
https://github.com/flarum/core/blob/master/src/Http/HttpServiceProvider.php#L64-L66

**Screenshot**

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/src/Frontend/FrontendServiceProvider.php
+++ b/src/Frontend/FrontendServiceProvider.php
@@ -148,7 +148,8 @@ class FrontendServiceProvider extends AbstractServiceProvider
 
         $views->share([
             'translator' => $container->make('translator'),
-            'url' => $container->make(UrlGenerator::class)
+            'url' => $container->make(UrlGenerator::class),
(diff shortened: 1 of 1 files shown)


askvortsov1 (approved):
Yep, this looks good to me! Agreed that there's no need to force demand-time resolution.

SychO9 (approved):

PR #3195, opened 2021-12-12
feat: Allow switching the ImageManager driver

imorland:
I recently identified an issue where image uploads were rendered with different color gamuts if they had been processed by Intervention in any way (resized, etc), if they had an embedded icc color profile. By switching to `imagick`, this profile was respected. Therefore, I'd like to introduce the option to use either driver everywhere in Flarum.

We use the `ImageManager` from Intervention in plently of places, both in core and extensions. Intervention supports two drivers - `gd` (default) and `imagick`. [ref](http://image.intervention.io/getting_started/configuration)

This PR adds the possibility to switch to `imagick`, if this is required.

By default `gd` will still be used, however, adding the following to `config.php`, one can specify the alternative:
```php
'intervention' =>
  array (
    'driver' => 'imagick',
  )
```

Extensions should update their code and refrain from `new ImageManager()` and resolve the singleton from the container instead.

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/src/Filesystem/FilesystemServiceProvider.php
+++ b/src/Filesystem/FilesystemServiceProvider.php
@@ -80,7 +80,7 @@ class FilesystemServiceProvider extends AbstractServiceProvider
             }
 
             if (! Arr::has(self::INTERVENTION_DRIVERS, $driver)) {
-                throw new RuntimeException('intervention/image: invalid driver');
+                throw new RuntimeException("intervention/image: $driver is not valid");
             }
 
             return new ImageManager([
--- a/src/Filesystem/FilesystemServiceProvider.php
+++ b/src/Filesystem/FilesystemServiceProvider.php
@@ -12,6 +12,7 @@ namespace Flarum\Filesystem;
 use Flarum\Foundation\AbstractServiceProvider;
 use Flarum\Foundation\Config;
 use Flarum\Foundation\Paths;
+use Flarum\Foundation\ValidationException;
 use Flarum\Http\UrlGenerator;
 use Illuminate\Contracts\Container\Container;
 use Illuminate\Filesystem\Filesystem;
@@ -20,6 +21,10 @@ use Intervention\Image\ImageManager;
 
 class FilesystemServiceProvider extends AbstractServiceProvider
 {
+    const gdDriver = 'gd';
+    const imagickDriver = 'imagick';
+    const interventionDrivers = ['gd' => 'gd', 'imagick' => 'imagick'];
+    
     /**
      * {@inheritdoc}
      */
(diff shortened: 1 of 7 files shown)


src/Filesystem/FilesystemServiceProvider.php
```diff
@@ -20,6 +21,10 @@
 
 class FilesystemServiceProvider extends AbstractServiceProvider
 {
+    const gdDriver = 'gd';
+    const imagickDriver = 'imagick';
+    const interventionDrivers = ['gd' => 'gd', 'imagick' => 'imagick'];
```
askvortsov1, line 24: Let's make this protected?

src/Filesystem/FilesystemServiceProvider.php
```diff
@@ -20,6 +21,10 @@
 
 class FilesystemServiceProvider extends AbstractServiceProvider
 {
+    const gdDriver = 'gd';
```
askvortsov1: For consts, let's use UPPER_CASE

matteocontrini:
Pay attention to the fact that ImageMagick by default not only doesn't strip the ICC profile but EXIF data in general, which means that GPS data contained in an image's EXIF could be unintentionally leaked to everyone.

askvortsov1:
Could we inject the `ImageManager` as a constructor dependency where possible?

> Pay attention to the fact that ImageMagick by default not only doesn't strip the ICC profile but EXIF data in general, which means that GPS data contained in an image's EXIF could be unintentionally leaked to everyone.

We'd still be using `gd` as the default though? I feel like this is something we might want to mention in documentation rather than adding a layer to address it ourselves.

imorland:
> I feel like this is something we might want to mention in documentation rather than adding a layer to address it ourselves.

Agreed. I'll look at adding something for this in `fof/upload` to protect GPS data, etc

matteocontrini:
> Agreed. I'll look at adding something for this in `fof/upload` to protect GPS data, etc

If it helps, this is what I've been doing in a private extension in production for years I think:

```php
$exif = $img->getCore()->getImageProfiles('exif', false);
if (count($exif) > 0) {
    $img->getCore()->removeImageProfile('exif');
}
```

The `exif` profile includes for example the GPS data but not the ICC data.

askvortsov1 (approved):

src/Filesystem/FilesystemServiceProvider.php
```diff
@@ -60,5 +66,26 @@ public function register()
...
+            $intervention = $config->offsetGet('intervention');
+            $driver = Arr::get($intervention, 'driver', self::INTERVENTION_DRIVERS['gd']);
+
+            // Check that the imagick library is actually available, else default back to gd.
+            if ($driver === self::INTERVENTION_DRIVERS['imagick'] && ! extension_loaded(self::INTERVENTION_DRIVERS['imagick'])) {
+                $driver = self::INTERVENTION_DRIVERS['gd'];
+            }
+
+            if (! Arr::has(self::INTERVENTION_DRIVERS, $driver)) {
+                throw new ValidationException(['intervention' => 'invalid driver']);
```
SychO9, line 83: this should be a runtime exception, since it's not user input, it's application configuration.

SychO9 (changes requested):

SychO9 (approved):

PR #3203, opened 2021-12-14
feat: Add events for Notification read / read all

imorland:
Adds events for `Notification` read and readall actions, for extensions to listen for.

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/src/Notification/Command/ReadNotificationHandler.php
+++ b/src/Notification/Command/ReadNotificationHandler.php
@@ -17,7 +17,7 @@ use Illuminate\Contracts\Events\Dispatcher;
 class ReadNotificationHandler
 {
     /**
-     * @var DIspatcher
+     * @var Dispatcher
      */
     protected $events;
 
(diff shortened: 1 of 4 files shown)


davwheat:
Would it be worth also passing the time to the event? There's a chance that any event handlers may wish to use it (for whatever reason), and the time might not be the same as what was set on the DB table.

imorland:
Yeah, I guess that could be useful. Didn't think of that as part of the use case I hace in mind, but usel nonetheless.

src/Notification/Command/ReadNotificationHandler.php
```diff
@@ -10,10 +10,25 @@
...
 
 use Carbon\Carbon;
+use Flarum\Notification\Event\Read;
 use Flarum\Notification\Notification;
+use Illuminate\Contracts\Events\Dispatcher;
 
 class ReadNotificationHandler
 {
+    /**
+     * @var DIspatcher
```
davwheat, line 20:
```suggestion
     * @var Dispatcher
```

davwheat (approved):

src/Notification/Command/ReadNotificationHandler.php
imorland: See, I was just testing to see if anyone noticed 🙈

PR #3206, opened 2021-12-15
fix: replace throw with debug warning to fix breaking change

davwheat:
**Extension developers, coming from the warning message:**
In Flarum 2.0, you will no longer be able to push data to the app store which does not have a matching model class _from an API payload_, such as those from `app.store.find` and the forum's preloaded API document.

You should define a class in your extension which inherits from [`Model`](https://github.com/flarum/core/blob/master/js/src/common/Model.ts), and preferably defining attribute functions instead of accessing `.data.attributes`.

Call `app.store.pushObject` manually will also work as this, by default, allows unregistered objects.

For more info about creating frontend models, check out the Flarum documentation: https://docs.flarum.org/extend/models#frontend-models

---

Many community extensions include unregistered objects in the forum payload. These extensions would suffer from a breaking change for v1.2, which is not appropriate.

This PR replaces the throw with a debug warning, with a link to this PR for reference.

image

Would registering a type `{}` be a workaround or feature? Technically, with the current JS, someone could do `app.store.models.myModel = Object` and the code would happily call `new Object()` allowing them to keep the same functionality. Should we mention this in the comment at the top of the PR (either to discourage it, or to state it as a workaround)?

**Necessity**

**Confirmed**

### DIFF
--- a/js/src/common/Store.ts
+++ b/js/src/common/Store.ts
@@ -1,5 +1,6 @@
 import app from '../common/app';
 import { FlarumRequestOptions } from './Application';
+import fireDebugWarning from './helpers/fireDebugWarning';
 import Model, { ModelData, SavedModelData } from './Model';
 
 export interface MetaInformation {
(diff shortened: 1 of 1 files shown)


imorland (approved):
Tested locally, no issues from me

askvortsov1 (approved):

PR #3208, opened 2021-12-16
fix: Empty settings don't return defaults

imorland:
This has been an annoyance forever, as an extention developer.

For example, we have a setting, a string let's say. We can provide a default value for that, great. However, if that setting is saved as simply an empty string `''`, I would expect the provided default to be returned. That's not currently the case.

At present, the logic for deciding if the default should be returned is based on `is_null`, which only works if that setting key does not exist on the table. By using `empty()`, we can check for `null` or empty string at the same time, and return the default, if provided.

Are there any unexpected side effects of doing this?

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/src/Settings/DatabaseSettingsRepository.php
+++ b/src/Settings/DatabaseSettingsRepository.php
@@ -27,7 +27,7 @@ class DatabaseSettingsRepository implements SettingsRepositoryInterface
 
     public function get($key, $default = null)
     {
-        if (is_null($value = $this->database->table('settings')->where('key', $key)->value('value'))) {
+        if (empty($value = $this->database->table('settings')->where('key', $key)->value('value'))) {
             return $default;
         }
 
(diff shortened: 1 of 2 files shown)


davwheat:
The only downside I can think of is if someone wants to save a setting as an empty string which is, by default, not an empty string. I think that case is very rare, if it even happens, and if someone wanted to do this, they could just set the default value in a migration instead.

SychO9:
`empty` also returns true with int `0` and string `"0"`, I understand the case you're describing, but I don't think this is the best way to solve it.

Perhaps we can somehow specify which values for a setting would lead to it being set null/removed so that it then falls back to the default.

imorland:
> `empty` also returns true with int `0` and string `"0"`

Ahh, yes I forgot about that. 

Ok, so this isn't the "quick win" I thought it might have been. I'll close this and create an issue where we can discuss options, etc

PR #3212, opened 2021-12-17
fix: breaking change in Search component - renaming of `state` property

davwheat:
**Extension developers:**
Accessing the `state` property of the `Search` component is now deprecated, and will be removed in Flarum 2.0. Please replace usages of this property with `searchState`. You do not need to change any of your code's implementation -- just a simple rename of the property works fine.

When doing so, be aware that you **must** update your `flarum/core` constraint to `^1.2.0` in order to not cause errors on older versions of Flarum.

_Old example:_
```tsx
if (typeof this.state.getValue() === 'undefined') {
  this.state.setValue('');
}
```

_New example:_
```tsx
if (typeof this.searchState.getValue() === 'undefined') {
  this.searchState.setValue('');
}
```

----

A breaking change was implemented as part of the model TS refactor to the Search component. Any components which inherited from this class would be met with forum-breaking errors when accessing the `state` property, as this was unexpectedly renamed to `searchState`.

This PR implements getters and setters for the old `state` property, which get/set the new `searchState` property. This also fires a debug warning with a deprecation notice, and a link to this PR for extension developers to read.

_As an example, this change would break FoF/Byobu._

**Necessity**

**Confirmed**

### DIFF
--- a/js/src/common/Store.ts
+++ b/js/src/common/Store.ts
@@ -1,6 +1,6 @@
 import app from '../common/app';
 import { FlarumRequestOptions } from './Application';
-import fireDebugWarning from './helpers/fireDebugWarning';
+import { fireDeprecationWarning } from './helpers/fireDebugWarning';
 import Model, { ModelData, SavedModelData } from './Model';
 
 export interface MetaInformation {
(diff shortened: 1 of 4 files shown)


SychO9 (approved):

Pi-Bouf (approved):

tankerkiller125 (approved):

iPurpl3x:
Is it normal that the warning is shown even if there are no extensions installed?

davwheat:
Yes, it's something we need to fix, but actually undoing this change was the highest on our priority list.

PR #3215, opened 2021-12-20
chore: use consistent empty vdom return types in core

davwheat:

- consistently use `null` or `false` as empty return types

At the moment, core uses a mix of `''`, `false`, `[]` and `null` to all represent empty content within components or the vdom. Using `''` or `[]` are generally bad, as this ends up polluting the vdom. `''` adds unneeded text vnodes to the vdom, and `[]` adds empty fragments to the vdom, both of which take up memory and can decrease performance during redraws. In general, these impacts are likely negligible, but we should set a good example for extension developers.

Have I missed anywhere?

**Necessity**

**Confirmed**

### DIFF
--- a/js/src/admin/components/DashboardWidget.js
+++ b/js/src/admin/components/DashboardWidget.js
@@ -20,6 +20,6 @@ export default class DashboardWidget extends Component {
    * @return {VirtualElement}
    */
   content() {
-    return [];
+    return null;
   }
 }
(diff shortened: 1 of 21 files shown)


js/src/admin/components/DashboardWidget.js
```diff
@@ -20,6 +20,6 @@ export default class DashboardWidget extends Component {
    * @return {VirtualElement}
    */
   content() {
-    return [];
+    return null;
```
askvortsov1, line 23: This one could be breaking if someone was extending this method. But that's probably fine since I don't think this is public API (it's not an itemlist).

askvortsov1 (approved):

js/src/admin/components/DashboardWidget.js
davwheat: Yeah, IMO extending functions like this is risky and we could break it at any time. On the other hand, everything should be itemlists!

askvortsov1:

PR #3217, opened 2021-12-20
fix: `classes` function including empty strings

davwheat:

If the existing classes passed has one or more spaces in a row (e.g. `MyCoolPost   MyCoolPost--megaCool`), then the `classes` array would fill up with empty strings (e.g. `['MyCoolPost', '', '', 'MyCoolPost--megaCool']`) despite these not actually being classes.

This PR filters out empty strings from the array to prevent this.

**Screenshot**

**Necessity**

**Confirmed**

### DIFF
--- a/js/src/forum/components/Post.js
+++ b/js/src/forum/components/Post.js
@@ -132,7 +132,9 @@ export default class Post extends Component {
       'Post--loading': this.loading,
       'Post--by-actor': user === app.session.user,
       'Post--by-start-user': user?.id() === discussion.attribute('startUserId'),
-    }).split(' ').filter(x => !!x);
+    })
+      .split(' ')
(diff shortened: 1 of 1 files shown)


askvortsov1:
Should this go into our wrapper of class list?

davwheat:
Actually this whole function can be refactored to use classList.

askvortsov1 (commented):
Does this still return a string array though? What if people are extending it to add more classes?

davwheat:
> Does this still return a string array though?

Oops.

> Should this go into our wrapper of class list?

I don't think so. If people use `classList`/`clsx` correctly, there won't be multi-space gaps between class names.

askvortsov1 (approved):

askvortsov1:

PR #3219, opened 2021-12-21
fix: hide `WelcomeHero` when content is empty

davwheat:
**Fixes https://discuss.flarum.org/d/22061**

- hide welcomehero when content is empty

Is there a better way to do this? Should `this.hidden` be replaced with a method `isHidden()` instead?

**Necessity**

**Confirmed**


SychO9 (approved):

SychO9:
also I do like the idea of a `isHidden` method instead. For extensibility.

davwheat:
It would be a breaking change, I think, unless we still respect `.hidden`... I'll give it a shot.

SychO9 (approved):
Nice!

askvortsov1 (approved):

PR #3225, opened 2021-12-24
feat: add Less stylesheet formatting

davwheat:
This needs some discussion. IMO, this is the best way to go about formatting our Less stylesheets, but it also adds JS-related code to our repo root, rather than keeping it contained within the `js` folder. We've also said in the past, however, that having stuff in the `js` subdirectory modifying files outside that folder is confusing and unclear.

If we decide to do this, I'll create the GH Actions workflow for it, too, and fix the existing workflows.

- adds Yarn workspace to root of repo, with Prettier
- format Less with Prettier

Is there anything else we want to format, too? Should JS code formatting also move to the repo root script?

**Necessity**

**Confirmed**


askvortsov1:
Can we hold off on this discussion until after we figure out monorepo stuff?

askvortsov1:

PR #3228, opened 2021-12-27
Some v1.2 frontend fixes

askvortsov1:
- Fix error when not allowed to search
- Fix login form inputs unclickable on mobile
- Use translations for page titles in frontend for consistency with backend
- Fix consecutive shows of same modal with different attrs (Fixes https://discuss.flarum.org/d/25182-friendsofflarum-oauth/259)

### DIFF
--- a/js/src/common/Application.tsx
+++ b/js/src/common/Application.tsx
@@ -9,6 +9,7 @@ import Translator from './Translator';
 import Store, { ApiPayload, ApiResponse, ApiResponsePlural, ApiResponseSingle, payloadIsPlural } from './Store';
 import Session from './Session';
 import extract from './utils/extract';
+import extractText from './utils/extractText';
 import Drawer from './utils/Drawer';
 import mapRoutes from './utils/mapRoutes';
 import RequestError, { InternalFlarumRequestOptions } from './utils/RequestError';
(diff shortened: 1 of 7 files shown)


davwheat (approved):

davwheat:
Re-running tests because they were taking 30+ mins to install deps 🤣

locale/core.yml
```diff
@@ -628,11 +633,6 @@ core:
       submit_button: => core.ref.save_changes
       title: => core.ref.reset_your_password
 
-    # Translations in this namespace are used to format page meta titles.
-    meta_titles:
-      with_page_title: "{pageNumber, plural, =1 {{pageTitle} - {forumName}} other {{pageTitle}: Page # - {forumName}}}"
-      without_page_title: "{pageNumber, plural, =1 {{forumName}} other {Page # - {forumName}}}"
-
```
datitisev, line 635: Is this not a breaking change for language packs?

js/src/common/Application.tsx
```diff
@@ -365,9 +365,21 @@ export default class Application {
...
+      forumName: this.forum.attribute('title'),
+      // Until we add page numbers to the frontend, this is constant at 1
+      // so that the page number portion doesn't show up in the URL.
+      pageNumber: 1,
+    };
+
+    const title =
+      onHomepage || !this.title
+        ? app.translator.trans('core.lib.meta_titles.without_page_title', params)
+        : app.translator.trans('core.lib.meta_titles.with_page_title', params);
```
datitisev, line 381: These need to use `extractText`

image

datitisev (changes requested):
Search & input fixes tested working.

Updated title logic needs fixing & question about BC.

locale/core.yml
askvortsov1: Hmm that's a tough one. As per the definition in https://discuss.flarum.org/d/26242-draft-public-api-policy, our public API is defined in terms of the interfaces of the code, not necessarily of behavior. This will require language packs to update, but in the meantime it won't break forums using those language packs, so I don't see it as illegal.

SychO9: It's fine, we introduced these in this cycle anyway, so it doesn't matter, it never made it to a release.

askvortsov1: The argument is that even requiring translations means that extensions will need to add translations to avoid showing the 'core.lib' text. However, since that's easily addressable and doesn't break functionality, I think it's fine.

SychO9: I do agree with that.

js/src/common/components/ModalManager.tsx
```diff
@@ -24,16 +24,21 @@ export default class ModalManager extends Component<IModalManagerAttrs> {
...
       <div className="ModalManager modal fade">
-        {!!modal &&
-          modal.componentClass.component({
-            ...modal.attrs,
-            animateShow: this.animateShow.bind(this),
-            animateHide: this.animateHide.bind(this),
-            state: this.attrs.state,
-          })}
+        {!!Tag && (
+          <div>
```
davwheat, line 32: What's the need for this extra `<div>`?

askvortsov1: Accidentially left that in, I thought I had removed that.

davwheat (approved):

locale/core.yml
datitisev: Sounds good, I figured it wasn't that big of a deal as it's not breaking.

datitisev (approved):
Confirmed fix for app title & modal opening (specifically oauth with sign up modal) working

PR #3230, opened 2021-12-27
perf: only update last seen time when current > 120s ago

### LINKED ISSUE #3231: `last_used`/`last_seen` updated on all API requests
Access tokens `last_used` date and User `last_seen` at updated on each request, including for API GET requests, causing ALL requests to always use write DB connections and causing massive usage on databases

davwheat:
**Ref flarum/issue-archive#121. Fixes flarum/framework#3231.**

- Only updates the last seen timestamp when the current value is > 120 seconds ago

This, in a dev environment with Clockwork, reduces the overall request time by approx 5 milliseconds.

We might want to consider a better solution, such as a heartbeat ping from users every x minutes/seconds to update this and other states, rather than this method.

Would it be better to use the `isOnline` timeout value of 5 mins instead?

Also, is it worth adding tests for this?

**Necessity**

**Confirmed**

### DIFF
--- a/src/Http/AccessToken.php
+++ b/src/Http/AccessToken.php
@@ -64,6 +64,12 @@ class AccessToken extends AbstractModel
     protected static $lifetime = 0;
 
     /**
+     * Difference from the current `last_activity_at` attribute value before `updateLastSeen()`
+     * will update the attribute on the DB. Measured in seconds.
+     */
+    private const LAST_ACTIVITY_UPDATE_DIFF = 60;
+
+    /**
      * Generate an access token for the specified user.
      *
      * @param int $userId
@@ -98,7 +104,7 @@ class AccessToken extends AbstractModel
         $now = Carbon::now();
 
         // Only update if the current timestamp >120s ago
-        if ($this->last_activity_at === null || $this->last_activity_at->diffInSeconds($now) > 120) {
+        if ($this->last_activity_at === null || $this->last_activity_at->diffInSeconds($now) > AccessToken::LAST_ACTIVITY_UPDATE_DIFF) {
             $this->last_activity_at = $now;
         }
 
--- a/src/Http/AccessToken.php
+++ b/src/Http/AccessToken.php
@@ -97,6 +97,7 @@ class AccessToken extends AbstractModel
     {
         $now = Carbon::now();
 
+        // Only update if the current timestamp >120s ago
         if ($this->last_activity_at === null || $this->last_activity_at->diffInSeconds() > 120) {
             $this->last_activity_at = $now;
         }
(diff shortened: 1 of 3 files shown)


luceos:
A heartbeat makes sense. Especially if extensions can replace the logic. Websockets have this logic by default and can be configured to set the last seen at value from their end (definitely going to add it to realtime now).

luceos (approved):
This is what blomstra uses also.

davwheat:
I agree with you on that. We should consider whether it's worth merging this for the time being and implementing a heartbeat for next release, or whether to completely hold off.

How significant is the DB load caused by this at the moment for Blomstra?

askvortsov1:
> I agree with you on that. We should consider whether it's worth merging this for the time being and implementing a heartbeat for next release, or whether to completely hold off.
> 
> How significant is the DB load caused by this at the moment for Blomstra?

Debouncing it by 5 minutes should already hopefully reduce the load, so I'd be in favor of merging it now and adding a heartbeat for the next release.i

src/Http/AccessToken.php
```diff
@@ -95,7 +95,12 @@ public static function generate($userId)
      */
     public function touch(ServerRequestInterface $request = null)
     {
-        $this->last_activity_at = Carbon::now();
+        $now = Carbon::now();
+
+        // Only update if the current timestamp >120s ago
+        if ($this->last_activity_at === null || $this->last_activity_at->diffInSeconds($now) > 120) {
```
askvortsov1, line 101: Minor nitpick, but can we set it to >=, or maybe even the diff / 2? It would be preferable to keep this somewhat overlapping.

askvortsov1 (approved):
The 120 value (or whatever we end up going with) should be in a constant somewhere, not hard coded. Otherwise LGTM!

davwheat:
I've increased the diff to 3 mins (User) and 1.5 mins (access token).

I think this is a good trade-off between performance and ensuring that data is relatively up-to-date.

5 mins wouldn't be good as there's a higher chance of someone inadvertently appearing offline if they make a request just before 5 mins pass from their last.

src/User/User.php
```diff
@@ -564,7 +570,12 @@ public function setPreference($key, $value)
      */
     public function updateLastSeen()
     {
-        $this->last_seen_at = Carbon::now();
+        $now = Carbon::now();
+
+        // Only update if the current timestamp is from a while ago
```
askvortsov1, line 575: Let's drop this comment and the correspondone one in AccessToken.php. Otherwise good to merge!

askvortsov1 (approved):

src/Http/AccessToken.php
```diff
@@ -95,7 +101,12 @@ public static function generate($userId)
      */
     public function touch(ServerRequestInterface $request = null)
     {
-        $this->last_activity_at = Carbon::now();
+        $now = Carbon::now();
+
+        // Only update if the current timestamp >120s ago
```
davwheat, line 106:
```suggestion
```

src/User/User.php
davwheat:
```suggestion
```

PR #3233, opened 2021-12-28
perf: include request info in first accesstoken touch

davwheat:
**Fixes: #3234**

Prevents two separate DB updates as a result of sequential token `->touch()`es.

image

If a user selects the Remember option, the last activity timestamp is updated in that middleware, then the IP and user agent is updated in the session auth middleware.

These two separate requests can be merged by actually providing the request info to the touch method in the remember middleware.

N/A

**Necessity**

**Confirmed**


askvortsov1 (approved):
Makes sense, thanks!

tankerkiller125 (approved):

PR #3236, opened 2021-12-28
fix: don't fire deprecation warnings for Mithril-originating action

davwheat:

Don't fire a deprecation warning when Search state is set to `undefined`.

Mithril does this when instantiating every component, which is triggering this dep warning when no extension or core have done anything wrong.

**Necessity**

**Confirmed**

### DIFF
--- a/js/src/forum/components/Search.tsx
+++ b/js/src/forum/components/Search.tsx
@@ -76,7 +76,7 @@ export default class Search<T extends SearchAttrs = SearchAttrs> extends Compone
     return this.searchState;
   }
   protected set state(state: SearchState) {
-    fireDeprecationWarning('`state` property of the Search component is deprecated', '3212');
+    state !== undefined && fireDeprecationWarning('`state` property of the Search component is deprecated', '3212');
     this.searchState = state;
   }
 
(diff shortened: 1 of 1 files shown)


SychO9 (approved):

askvortsov1 (approved):

askvortsov1 (approved):
Can we add a quick comment though?

js/src/forum/components/Search.tsx
```diff
@@ -76,7 +76,7 @@ export default class Search<T extends SearchAttrs = SearchAttrs> extends Compone
     return this.searchState;
   }
   protected set state(state: SearchState) {
-    fireDeprecationWarning('`state` property of the Search component is deprecated', '3212');
+    state !== undefined && fireDeprecationWarning('`state` property of the Search component is deprecated', '3212');
```
davwheat, line 79:
```suggestion
    // Workaround to prevent triggering deprecation warnings due to Mithril
    // setting state to undefined when creating components
    state !== undefined && fireDeprecationWarning('`state` property of the Search component is deprecated', '3212');
```

askvortsov1 (approved):

PR #3238, opened 2021-12-28
fix: prevent unwarranted a11y warnings on custom Button subclasses

davwheat:

UploadImageButton has its own content in the `view` method, so we don't need the empty children a11y warning.

N/A

**Necessity**

**Confirmed**

### DIFF
--- a/js/src/common/components/Button.tsx
+++ b/js/src/common/components/Button.tsx
@@ -67,6 +67,15 @@ export interface IButtonAttrs extends ComponentAttrs {
  * styles can be applied by providing `className="Button"` to the Button component.
  */
 export default class Button<CustomAttrs extends IButtonAttrs = IButtonAttrs> extends Component<CustomAttrs> {
+  /**
+   * Prevents firing of accessibility warnings for this Button component.
+   * 
+   * This should be set to `true` when creating a custom Button class component which
+   * does not expect any children to be provided to it because, for example, it has a
+   * custom `view` method with its own content baked-in.
+   */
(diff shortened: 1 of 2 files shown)


tankerkiller125 (approved):

js/src/common/components/Button.tsx
```diff
@@ -107,7 +116,7 @@ export default class Button<CustomAttrs extends IButtonAttrs = IButtonAttrs> ext
 
     const { 'aria-label': ariaLabel } = this.attrs;
 
-    if (!ariaLabel && !extractText(vnode.children) && !this.element?.getAttribute?.('aria-label')) {
```
askvortsov1, line 110: Instead of doing this, could we check whether the view method is overriden?

[Playground Link](https://www.typescriptlang.org/play?#code/MYGwhgzhAECC0G8BQ1XQGYFcB2wAUAlIimqQE4CmALpmdtAORgMDcJqAvku9AJYQAxHMADyANwpkyvACYyK2QsVLlqtelQAW-AHRZc0AIQBeY3B0AHMgHsqtgJ4WKe4Ty5ckoSDABC0CgAeVAoyMPDIpPr4RBEqqJQ0dIwARqxuSB7A1tgQ1iDOINYA5nh42BQA7nCEBDr8QrjiktJyCjWe2bn5OoUlZZXQPjV1gsJNUrLyigQEQA)

askvortsov1 (changes requested):

askvortsov1 (approved):

PR #3239, opened 2021-12-29
fix: error in funding `composer.json` block bricks frontend

davwheat:

Any error in `composer.json`'s `funding` block syntax will brick the admin frontend.

Oddly, this error seems to persist after correcting the error, `composer update` and a cache clear on my local dev forum. I had to implement this to allow me to toggle the extension, which then fixed the error.

N/A

**Necessity**

**Confirmed**

### DIFF
--- a/src/Extension/Extension.php
+++ b/src/Extension/Extension.php
@@ -263,11 +263,11 @@ class Extension implements Arrayable
             return $icon;
         }
 
-        $file = $this->path.'/'.$file;
+        $file = $this->path . '/' . $file;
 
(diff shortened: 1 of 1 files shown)


askvortsov1 (changes requested):
Instead of this, could we use something like `Arr::get($funding, '0.url')`?

davwheat:
That was a fair bit simpler.

askvortsov1 (approved):

tankerkiller125 (approved):

PR #3249, opened 2022-01-04
FIx broken post/discussion soft delete

askvortsov1:
Before the Model typescript rewrite, `pushAttributes` supported including relationship objects, which is hacky but incorrect behavior. With the rewrite, this functionality was broken.

This PR deprecates the functionality, adds a deprecated BC layer with a debug warning, and removes instances of incorrect usage.

### DIFF
--- a/js/src/common/Model.ts
+++ b/js/src/common/Model.ts
@@ -121,7 +121,7 @@ export default abstract class Model {
           fireDeprecationWarning('Providing models as attributes to `Model.pushData()` or `Model.pushAttributes()` is deprecated.', '3249');
           delete data.attributes[key];
           data.relationships ||= {};
-          data.relationships[key] = { data: Model.getIdentifier(val)};
+          data.relationships[key] = { data: Model.getIdentifier(val) };
         }
       }
 
--- a/js/src/common/Model.ts
+++ b/js/src/common/Model.ts
@@ -1,5 +1,6 @@
 import app from '../common/app';
 import { FlarumRequestOptions } from './Application';
+import fireDebugWarning from './helpers/fireDebugWarning';
 import Store, { ApiPayloadSingle, ApiResponseSingle, MetaInformation } from './Store';
 
 export interface ModelIdentifier {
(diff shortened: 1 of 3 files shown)


js/src/common/Model.ts
```diff
@@ -111,6 +112,19 @@ export default abstract class Model {
 
     if ('attributes' in data) {
       this.data.attributes ||= {};
+
+      // @deprecated
+      // Filter out relationships that got in by accident.
+      for (const key in data.attributes) {
+        const val = data.attributes[key];
+        if (val && val instanceof Model) {
+          fireDebugWarning('Model.pushData() was passed a Model instance in the `attributes` field. This is not supported.');
```
davwheat, line 121: Could we use the deprecation warning helper here instead, and add a message to the top of the PR explaining how this should be done instead? (e.g. https://github.com/flarum/core/pull/3212)

```suggestion
          fireDeprecationWarning('Providing models as attributes to `Model.pushData()` or `Model.pushAttributes()` is deprecated.', '3249');
```

davwheat (changes requested):

js/src/common/Model.ts
```diff
@@ -111,6 +112,19 @@ export default abstract class Model {
 
     if ('attributes' in data) {
       this.data.attributes ||= {};
+
+      // @deprecated
+      // Filter out relationships that got in by accident.
+      for (const key in data.attributes) {
```
davwheat, line 118: Could this be better as `Object.entries(data.attributes).map(([key, val]) => {})`?

askvortsov1: I usually prefer functional transformations, but in this case since we're deleting attributes I think an imperative iteration is better.

js/src/common/Model.ts
```diff
@@ -1,5 +1,6 @@
 import app from '../common/app';
 import { FlarumRequestOptions } from './Application';
+import fireDebugWarning from './helpers/fireDebugWarning';
```
davwheat, line 3:
```suggestion
import { fireDeprecationWarning } from './helpers/fireDebugWarning';
```

davwheat (approved):
Just need to update the import. Approved after that.

PR #3133, opened 2021-10-27
[2.0] feat!: use `ItemList` for the DiscussionListState sorting map

davwheat:
**Changes proposed in this pull request:**

It is currently very difficult to change the ordering of the discussion list sorting options. This change replaces the existing object map with an ItemList, allowing us to use priorities instead.

**Reviewers should focus on:**

**Screenshot**

**Necessity**

- [X] Has the problem that is being solved here been clearly explained?
- [x] If applicable, have various options for solving this problem been considered?
- [x] For core PRs, does this need to be in core, or could it be in an extension?
- [X] Are we willing to maintain this for years / potentially forever?

**Confirmed**

- [x] Frontend changes: tested on a local Flarum installation.
- [ ] Backend changes: tests are green (run `composer test`).
- [ ] Core developer confirmed locally this works as intended.
- [x] Tests have been added, or are not appropriate here.

**Required changes:**

- [ ] Related documentation PR: (Remove if irrelevant)
- [ ] Related core extension PRs: (Remove if irrelevant)

PR #3254, opened 2022-01-06
[fix] Extract sortmap options to container

imorland:
**Partially addresses flarum/issue-archive#70 **

Make the backend sort options list a container binding so extensions can modify it.
Add a new extender `Extend\Sortmap` to register additional sort mapping options

I don't think we should go further for the moment (ie serializing to the frontend, etc) until https://github.com/flarum/core/pull/3133 is merged also. This change proposed here will at least allow extensions to properly add additional sort options

Documentation - what level is detail would you like to see here?

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/src/Extend/Sortmap.php
+++ b/src/Extend/Sortmap.php
@@ -16,7 +16,7 @@ class Sortmap implements ExtenderInterface
 {
     private $sortFields = [];
 
-    public function addSort(string $key, string $value):self
+    public function addSort(string $key, string $value): self
     {
         $this->sortFields[$key] = $value;
 
--- a/src/Extend/Sortmap.php
+++ b/src/Extend/Sortmap.php
@@ -1,0 +1,32 @@
+<?php
+
+/*
+ * This file is part of Flarum.
+ *
+ * For detailed copyright and license information, please view the
+ * LICENSE file that was distributed with this source code.
+ */
+
+namespace Flarum\Extend;
+
+use Flarum\Extension\Extension;
+use Illuminate\Contracts\Container\Container;
+
+class Sortmap implements ExtenderInterface
+{
+    private $sortFields = [];
+
+    public function addSort(string $key, string $value):self
+    {
+        $this->sortFields[$key] = $value;
+
+        return $this;
+    }
+
+    public function extend(Container $container, Extension $extension = null)
+    {
+        $container->extend('flarum.forum.sortmap', function (array $sortFields) {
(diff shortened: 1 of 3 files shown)


askvortsov1:
I agree with this idea in theory, but for v2.0 I'm hoping that this will be completely included in extenders, and therefore we'd need to uproot all extensions using this feature.

luceos:
@askvortsov1 upon reading this PR I was actually considering something else as well. Especially because I'm not a fan of "ioc binding all the things" anymore. What did you have in mind, where would you store the sorting?

askvortsov1:
> @askvortsov1 upon reading this PR I was actually considering something else as well. Especially because I'm not a fan of "ioc binding all the things" anymore. What did you have in mind, where would you store the sorting?

It would probably still be in the ioc, which isn't a bad thing IMO since it is global-state site configuration. But the IOC ideally wouldn't be the public API way of intracting with it.

luceos:
I'd personally see every related to sort inside one class responsible per model. So the `Discussion` model having a `DiscussionSearchParams` which includes the logic for the sort mapping as well. 

I would be okay with merging this in as long as an issue is attached to v2.0 as a reminder to remove/review this. Are we okay to move forward with this @askvortsov1 ?

src/Extend/Sortmap.php
```diff
@@ -0,0 +1,32 @@
...
+ * For detailed copyright and license information, please view the
+ * LICENSE file that was distributed with this source code.
+ */
+
+namespace Flarum\Extend;
+
+use Flarum\Extension\Extension;
+use Illuminate\Contracts\Container\Container;
+
+class Sortmap implements ExtenderInterface
```
askvortsov1, line 15: Let's drop this class for now until we figure out a cleaner way to generalize this for all models. Other than that, good to go IMO.

src/Forum/Content/Index.php
```diff
@@ -70,7 +70,7 @@ public function __invoke(Document $document, Request $request)
         $page = max(1, intval(Arr::pull($queryParams, 'page')));
         $filters = Arr::pull($queryParams, 'filter', []);
 
-        $sortMap = $this->getSortMap();
+        $sortMap = resolve('flarum.forum.sortmap');
```
askvortsov1, line 73: Is there no way we can inject this in the constructor?

askvortsov1 (changes requested):

askvortsov1:

PR #3266, opened 2022-01-24
Trims blank spaces on email address 

ajuchacko:
Removes blank space on email address while both registering or updating user's email address. Although I've not tested this small case, I've made necessary changes.
flarum/issue-archive#446 

Thank you

### DIFF
--- a/src/User/Command/EditUserHandler.php
+++ b/src/User/Command/EditUserHandler.php
@@ -17,7 +17,6 @@ use Flarum\User\UserRepository;
 use Flarum\User\UserValidator;
 use Illuminate\Contracts\Events\Dispatcher;
 use Illuminate\Support\Arr;
-use Illuminate\Support\Str;
 use Illuminate\Validation\ValidationException;
 
 class EditUserHandler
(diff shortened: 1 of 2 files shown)


askvortsov1:
Thank you about the PR! I like the idea, but I think a general solution (along the lines of https://github.com/laravel/framework/blob/277c2fbd0cebd2cb194807654d870f4040e288c0/src/Illuminate/Foundation/Http/Middleware/TrimStrings.php) might be better. That would also fix flarum/issue-archive#247. @flarum/core any thoughts on this?

tankerkiller125:
I agree with Sasha that the TrimStrings middleware would be way better for this, it would also resolve potential other issues we might have around this kind of thing.

The only issue I could possibly see with the TrimStrings middleware would be if people are using a space at the beginning or end of their password.

luceos:
Thank you @ajuchacko for your contribution. I think I agree with the rest and implementing a HTTP middleware makes more sense in the long run.

Is this something you could take care of, that would be bliss!

PR #3271, opened 2022-01-30
feat: Replace ColorPreviewInput preview box with 'color' input

dsevillamartin:
- Replace the preview box with a `color` input that allows for easier modification of colors while still having a text field
- Extracts `id` from attrs so that it's not used for two inputs

- Do we want a check for whether `color` input is supported? It seems to have good support, but on iOS it is not supported in `UIWebView` (deprecated, removed in iOS 13 I believe)
- Do we want to make this optional or just change the functionality to be usable? I don't see any reason for it to be an option, but we could do that.

**Screenshot**

(Ignore the 'Primary color' label)

https://user-images.githubusercontent.com/6401250/151707985-e4d4a3bb-9d07-4e5c-8513-394bd5f71485.mp4

https://user-images.githubusercontent.com/6401250/151708016-5721f29d-773d-46fb-9069-6e2ff4cfcc70.mp4

**Necessity**

**Confirmed**

- ~~Tests have been added, or are not appropriate here.~~

### DIFF
--- a/js/src/common/components/ColorPreviewInput.tsx
+++ b/js/src/common/components/ColorPreviewInput.tsx
@@ -5,23 +5,20 @@ import classList from '../utils/classList';
 import icon from '../helpers/icon';
 
 export default class ColorPreviewInput extends Component {
-  value?: string;
-
   view(vnode: Mithril.Vnode<ComponentAttrs, this>) {
-    const { className, ...attrs } = this.attrs;
-    const value = attrs.bidi?.() || attrs.value;
+    const { className, id, ...attrs } = this.attrs;
 
     attrs.type ||= 'text';
 
     return (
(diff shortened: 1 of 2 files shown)


askvortsov1:
It's not working for me locally on Chrome
image

datitisev:
@askvortsov1 That's odd. It works for me.

While developing this, I had a few issues when CSS sometimes didn't get properly recompiled and I had to manually clear the Flarum cache. Perhaps that is what is happening to you?

This is in Chrome (downloaded to make sure Chromium Edge was working and Chrome wasn't)

image

askvortsov1 (approved):
Working now, LGMT!

SychO9 (approved):

PR #3278, opened 2022-02-06
fix: composer header overlap with controls

davwheat:

- fixes overlap with long discussion titles and composer's controls

**Screenshot**

### Before
image
image

### After
image
image

**Necessity**

**Confirmed**


askvortsov1 (approved):

askvortsov1:
At some point we do need to set up prettier for less

davwheat:
When monorepo comes :)

luceos (approved):

PR #3282, opened 2022-02-08
fix: sort posts by number instead of creation datetime

### LINKED ISSUE #3281: Use `number` for sorting posts instead of `createdAt`
First mentioned here: https://github.com/flarum/core/pull/2384#pullrequestreview-509977549

We currently order posts by `createdAt` time rather than using the dedicated `number` field. This is quite unintuitive, since `number` is used for deep-linking in the frontend, and not sorting by number can result in situations where scrolling down decreases the deep-link variable in the URL.

This also prevents extensions re-ordering a discussion's posts without rewriting timestamps.

**Places which need changing:**

https://github.com/flarum/core/blob/c622070366740e8cf9bf637eb5afc37747eb1467/js/src/forum/states/PostStreamState.js#L290-L294

https://github.com/flarum/core/blob/c622070366740e8cf9bf637eb5afc37747eb1467/js/src/forum/components/DiscussionPage.tsx#L213

https://github.com/flarum/core/blob/c622070366740e8cf9bf637eb5afc37747eb1467/src/Api/Controller/ShowDiscussionController.php#L135

https://github.com/flarum/core/blob/c622070366740e8cf9bf637eb5afc37747eb1467/src/Api/Controller/ShowDiscussionController.php#L189

https://github.com/flarum/core/blob/c622070366740e8cf9bf637eb5afc37747eb1467/src/Api/Controller/ListPostsController.php#L44

not sure if there are more...

davwheat:
**Fixes #3281**

Replaces instances of `createdAt` sorting with `number` sorting for posts.

Anywhere I've missed? Not too familiar with this area of core, personally. Does this affect/break anything else?

While this has the slim chance of breaking one or more extensions, I'm in favour of taking this approach sooner rather than later for the reasoning of sanity. If we have a dedicated `number` field, we should use that instead of `createdAt` as it is what most extension developers would assume is used in the first place. The fact that this hasn't been reported before suggests that no extensions have run into this issues, thus it shouldn't affect any common extensions.

**Screenshot**

Times are no longer seen as the source of ordering:

**Necessity**

**Confirmed**

### DIFF
--- a/src/Api/Controller/ListPostsController.php
+++ b/src/Api/Controller/ListPostsController.php
@@ -41,7 +41,7 @@ class ListPostsController extends AbstractListController
     /**
      * {@inheritdoc}
      */
-    public $sortFields = ['number'];
+    public $sortFields = ['number', 'createdAt'];
 
     /**
      * @var PostFilterer
--- a/src/Api/Controller/ListPostsController.php
+++ b/src/Api/Controller/ListPostsController.php
@@ -41,7 +41,7 @@ class ListPostsController extends AbstractListController
     /**
      * {@inheritdoc}
      */
-    public $sortFields = ['createdAt'];
+    public $sortFields = ['number'];
 
     /**
      * @var PostFilterer
--- a/js/src/forum/components/DiscussionPage.tsx
+++ b/js/src/forum/components/DiscussionPage.tsx
@@ -210,7 +210,7 @@ export default class DiscussionPage<CustomAttrs extends IDiscussionPageAttrs = I
             record.relationships.discussion.data.id === discussionId
         )
         .map((record) => app.store.getById<Post>('posts', record.id))
-        .sort((a?: Post, b?: Post) => (a?.createdAt()?.getTime() ?? 0) - (b?.createdAt()?.getTime() ?? 0))
+        .sort((a?: Post, b?: Post) => (a?.number() ?? 0) - (b?.number() ?? 0))
         .slice(0, 20);
     }
 
--- a/js/src/forum/states/PostStreamState.js
+++ b/js/src/forum/states/PostStreamState.js
@@ -289,7 +289,7 @@ class PostStreamState {
 
     if (loadIds.length) {
       return app.store.find('posts', loadIds).then((newPosts) => {
-        return loaded.concat(newPosts).sort((a, b) => a.createdAt() - b.createdAt());
+        return loaded.concat(newPosts).sort((a, b) => a.number() - b.number());
       });
     }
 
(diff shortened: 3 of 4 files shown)


src/Api/Controller/ListPostsController.php
```diff
@@ -41,7 +41,7 @@ class ListPostsController extends AbstractListController
     /**
      * {@inheritdoc}
      */
-    public $sortFields = ['createdAt'];
+    public $sortFields = ['number'];
```
askvortsov1, line 44: This should probably still be supported, in case someone wants to sort by createdAt from an external API.

askvortsov1 (changes requested):
The change makes sense, left one comment below. The only potential concern that comes to mind is that this might break the merge discussions extension, but I don't recall if that renumbers posts. @datitisev any thoughts on that?

datitisev:
@askvortsov1 I think most errors that come from merge discussions is during the renumbering process, as it is required otherwise posts have the same discussion id & number, which I believe makes them have duplicate indexes? Not sure.

Either way, that shouldn't stop this PR. If it causes issues then we'll fix them there, I think this is more important to have.

davwheat:
>  The only potential concern that comes to mind is that this might break the merge discussions extension, but I don't recall if that renumbers posts.

That's exactly where this stemmed from, actually. @imorland was working on a new feature to allow a choice for merging posts (in-line based on post creation time and updating the numbers OR append to the end of the discussion) which is what lead us to discover this issue.

src/Api/Controller/ListPostsController.php
davwheat:
```suggestion
    public $sortFields = ['number', 'createdAt'];
```

askvortsov1 (approved):
Sounds good then!

tankerkiller125 (approved):

luceos:
Make sure that performance isn't impacted. Other than that I'm aok with this.

davwheat:
Just a thought, do we need to define an `isDefaultSort` method, since `createdAt` isn't a default sort anymore, and the default method only checks if the sort if in the `sortFields` array?

askvortsov1:
> Just a thought, do we need to define an `isDefaultSort` method, since `createdAt` isn't a default sort anymore, and the default method only checks if the sort if in the `sortFields` array?

That shouldn't be necessary: `isDefaultSort` determines whether the user has requested a sort, not whether the sort is in the set of supported sort columns

davwheat:
Ah yeah, it seems `sortIsDefault()` tells the searcher/filterer to ignore what the API was provided as a sort if it's an illegal value and just use the default instead.

E.g. if you said `sort=-random`, it'd default back to `number`.

That's my understanding, at least.

Issue #3283, opened 2022-02-09
Replace LESS processing package


datitisev:
As I was going to replace `components/font-awesome` with the official `fortawesome/font-awesome` package to update to FA 6, I came across the following:

```php
ParseError: Unexpected input in _sizing.less on line 15, column 5
13| 
14| // step-based scale (with alignment)
15| each(.fa-sizes(), {
16|   .@{fa-css-prefix}-@{key} {
17|     .fa-size(@value);
18|   }
```

The `wikimedia/less.php` package had its last release on Dec 2020. There are issues open for updating to Less.js 3/4 and closed ones that refer to that regarding this issue with `each` (https://github.com/wikimedia/less.php/issues/49).

It does not look like they are actively maintaining it (just closing PRs, as they stated in some comment) so we may have to replace the package with another.

datitisev:
All LESS PHP processors are outdated. What we eventually want to do is most likely move to SCSS (e.g. `scssphp/scssphp` had an update last month).

PR #3284, opened 2022-02-13
Up font-awesome version to ^6.0.0

iamdarkle:
I have pushed a PR (https://github.com/components/font-awesome/pull/55) for the FA6 update in `components/font-awesome`, the new version is already tagged (https://github.com/components/font-awesome/releases/tag/6.0.0). This PR is just to push the requirement to version 6.0.0.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,7 +38,7 @@
     "require": {
         "php": ">=7.3",
         "axy/sourcemap": "^0.1.4",
-        "components/font-awesome": "^5.14.0",
+        "components/font-awesome": "^6.0.0",
         "dflydev/fig-cookies": "^3.0.0",
         "doctrine/dbal": "^2.7",
         "dragonmantank/cron-expression": "^3.1.0",


datitisev:
So I actually attempted to do this. This is what I found through this process:

- There seems to be no reason to use `components/font-awesome` as `fortawesome/font-awesome` is the official package which provides the same code
- FontAwesome v6 uses `each` in its less which is unsupported by `wikimedia/less.php`, the package that we use to process LESS (#3283)

    ```php
    ParseError: Unexpected input in _sizing.less on line 15, column 5
    13| 
    14| // step-based scale (with alignment)
    15| each(.fa-sizes(), {
    16|   .@{fa-css-prefix}-@{key} {
    17|     .fa-size(@value);
    18|   }
    ```
- There are no up-to-date LESS PHP processors

In short, simply updating to v6 causes errors. And so we have to do something more complex before Flarum 2.0 to be able to update.

iamdarkle:
I hadn't seen that issue, thanks for trying it out, I didn't think it would cause those problems... 😬
I'll keep an eye out for progress on the workaround, which also sounds much better in the long run, so I'll close the PR as it doesn't make much sense.

PR #3287, opened 2022-02-21
Fix permission check in user edit modal

clarkwinkelmann:
Fixes the permission check that shows/hides fields in the edit modal.

The correct logic was already used in `EditUserModal.data`, but in `EditUserModal.fields`, permissions were checked against the current session user rather than the edited user.

This causes no issue on a bare Flarum installation because all users have the same `canEditCredentials` boolean value since it's extracted from the actor's permissions. But if the gate has been customized by an extension this leads to inconsistent behavior. This is most noticeable if you customize the gate to disallow a user to edit themselves:

```php
class UserPolicy extends AbstractPolicy {
    public function editCredentials($actor, $user) {
        if ($actor->id === $user->id) {
            return $this->deny();
        }
    }
}
```

With this gate in place, moderators don't see the fields when editing any user, but they should see it on users that aren't themselves. This is fixed with the PR.

If there's a reason for the old implementation, it completely eludes me.

**Necessity**

**Confirmed**

### DIFF
--- a/js/src/common/components/EditUserModal.tsx
+++ b/js/src/common/components/EditUserModal.tsx
@@ -61,7 +61,7 @@ export default class EditUserModal<CustomAttrs extends IEditUserModalAttrs = IEd
   fields() {
     const items = new ItemList();
 
-    if (app.session.user?.canEditCredentials()) {
+    if (this.attrs.user.canEditCredentials()) {
       items.add(
         'username',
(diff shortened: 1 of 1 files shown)


askvortsov1 (approved):
Probably an oversight, fix looks good to me!

davwheat (approved):

PR #3288, opened 2022-02-21
Fix active search jumping to last item during loading

### LINKED ISSUE #3082: Search-box highlight should focus on the first item
## Bug Report

**Current Behavior**
The position of the highlight of the search-box is kind of Random? It often focus on Search Users, which is not very kind to forum user.

**Expected Behavior**
The highlight should focus on "Search all discussions"

**Screenshots**
<details>
  <summary>Show</summary>
  <image src="https://user-images.githubusercontent.com/35162802/135006695-3630b449-fcb3-4d53-a40d-e10be208449d.png"/>
  <image src="https://user-images.githubusercontent.com/35162802/135006770-edfc3552-80a1-4cba-9020-fba46eff7eaa.png"/>
</details>

**Environment**
- Flarum version: 1.0.4
- Website URL: https://discuss.flarum.org, https://discuss.flarum.org.cn

Different search results of different language keywords

clarkwinkelmann:
**Fixes #3082**

This PR solves the mysterious jump to last result that sometimes happens during search.

The problem was that `getCurrentNumericIndex` would return `-1` if the current item was not found, and since during result loading there is never any result, `-1` would be sent to `setIndex`. And `setIndex` contains logic to loop through to the last index intended for the UP/DOWN logic. It's then a race between the `onupdate` hook and the results loading. A first `onupdate` would run to hide the search results, then a second one would run to display the results again. But if the results have already loaded when the first hook runs, the loop through would happen.

This is a pretty simple change to make sure `0` is always returned in place of `-1`.

**Necessity**

**Confirmed**

### DIFF
--- a/js/src/forum/components/Search.tsx
+++ b/js/src/forum/components/Search.tsx
@@ -324,9 +324,10 @@ export default class Search<T extends SearchAttrs = SearchAttrs> extends Compone
 
   /**
    * Get the position of the currently selected search result item.
+   * Returns zero if not found.
    */
   getCurrentNumericIndex(): number {
-    return this.selectableItems().index(this.getItem(this.index));
+    return Math.max(0, this.selectableItems().index(this.getItem(this.index)));
   }
 
   /**


askvortsov1 (approved):
Argh that's a frustrating one, good job finding the fix! Solution makes sense to me.

davwheat (approved):

PR #3289, opened 2022-02-21
Fix pinned pane broken after discussion creation

clarkwinkelmann:
Fixes an issue I have experienced regularly on Discuss and other forums.

After a discussion has been created and the page has been redirected to the new discussion, the pinned pane would appear but the page content would not shift to the right, resulting in content being hidden by the pane.

This is because the `DiscussionPage` looks at the number of discussions in the list to decide whether to render the pane, but at that exact moment it's empty because a refresh has been triggered by `DiscussionComposer`.

The solution is to also look for the loading state, because if the discussions are loading it means the same thing, that we were coming from a page that shows the discussion list.

The bug can also be triggered by directly running the following code in console while on the homepage: `app.discussions.refresh() && m.route.set(app.route.discussion(app.store.all('discussions')[0]));`

The issue is less noticable when the pane isn't pinned, in that situation the pane was simply unavailable.

I have also removed an invalid parameter to `app.discussions.refresh()` where the bug "originates" from. This wasn't causing any issue except IDE warnings due to the typings not matching.

**Screenshot**
Before fix:
Screen Shot 2022-02-21 at 19 44 37

After fix:
Screen Shot 2022-02-21 at 19 44 22

**Necessity**

**Confirmed**

### DIFF
--- a/js/src/forum/components/DiscussionComposer.js
+++ b/js/src/forum/components/DiscussionComposer.js
@@ -101,7 +101,7 @@ export default class DiscussionComposer extends ComposerBody {
       .save(data)
       .then((discussion) => {
         this.composer.hide();
-        app.discussions.refresh({ deferClear: true });
+        app.discussions.refresh();
         m.route.set(app.route.discussion(discussion));
       }, this.loaded.bind(this));
   }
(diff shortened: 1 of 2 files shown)


davwheat (approved):
Crikey, you're on fire today! 👀

askvortsov1 (approved):

luceos:
Darn this has been so annoying!

PR #3295, opened 2022-03-07
fix: unread filter gambit does not return results when > 999 read discussions for a user

imorland:
When making use of the `unread` filter or gambit, I noticed that when a user has more than 999 previously read discussions, no results are returned.

Taking some advice from @askvortsov1, this PR deprecates `getReadIds()` which returned an array of IDs, in favour of a new utility function `getReadIdsQuery` - returning a `Builder`. This should also be a little more performant in the constructed SQL statement.

From a core standpoint, `getReadIds` is only used by the unread filter/gambit. So I think this is pretty safe. Is there anything else that's not been considered here?

**Necessity**

**Confirmed**

### DIFF
--- a/src/Discussion/DiscussionRepository.php
+++ b/src/Discussion/DiscussionRepository.php
@@ -49,10 +49,7 @@ class DiscussionRepository
      */
     public function getReadIds(User $user)
     {
-        return Discussion::leftJoin('discussion_user', 'discussion_user.discussion_id', '=', 'discussions.id')
-            ->where('discussion_user.user_id', $user->id)
-            ->whereColumn('last_read_post_number', '>=', 'last_post_number')
-            ->pluck('id')
+        return $this->getReadIdsQuery($user)
             ->all();
     }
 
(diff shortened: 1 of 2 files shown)


luceos (approved):
🎉

clarkwinkelmann:
Could there be a benefit to have this as an eloquent query scope, so that extension can also use this as part of relationship queries or in other contexts where there's an existing query builder instance?

The method on the repository could still make sense to return a prepared query that calls the scope and adds the `->select('id')` part.

src/Discussion/DiscussionRepository.php
```diff
@@ -54,6 +56,20 @@ public function getReadIds(User $user)
             ->all();
```
askvortsov1, line 56: We could implement this via the replacement for reduced code duplication.

askvortsov1 (approved):

src/Discussion/DiscussionRepository.php
imorland: Absolutely. done!

davwheat (approved):

askvortsov1 (approved):

PR #3349, opened 2022-03-16
feat: Graceful frontend extension initialization failure

SychO9:
**Part of flarum/issue-archive#85**

Instead of allowing the application's frontend to crash, we can catch the frontend errors when initializing extensions and gracefully fail to minimize the impact and improve user experience.

The approach, error message formulation, general code quality.

**Screenshot**
Before | After
-- | --
Screenshot from 2022-03-16 14-17-21 |  Screenshot from 2022-03-16 14-12-25

**Necessity**

**Confirmed**

### DIFF
--- a/framework/core/js/src/common/Application.tsx
+++ b/framework/core/js/src/common/Application.tsx
@@ -33,6 +33,7 @@ import type Mithril from 'mithril';
 import type Component from './Component';
 import type { ComponentAttrs } from './Component';
 import Model, { SavedModelData } from './Model';
+import fireApplicationError from "./helpers/fireApplicationError";
 
 export type FlarumScreens = 'phone' | 'tablet' | 'desktop' | 'desktop-hd';
 
@@ -262,7 +263,22 @@ export default class Application {
   }
 
   public boot() {
-    this.initializers.toArray().forEach((initializer) => initializer(this));
+    const caughtInitializationErrors: CallableFunction[] = [];
+
+    this.initializers.toArray().forEach((initializer) => {
+      try {
+        initializer(this);
+      } catch (e) {
+        const extension = initializer.itemName.includes('/')
+          ? initializer.itemName.replace(/(\/flarum-ext-)|(\/flarum-)/g, '-')
+          : initializer.itemName;
+
+        caughtInitializationErrors.push(() => fireApplicationError({
+          userTitle: extractText(app.translator.trans('core.lib.error.extension_initialiation_failed_message', { extension })),
+          consoleTitle: `${extension} failed to initialize`,
+        }, e));
+      }
+    });
 
     this.store.pushPayload({ data: this.data.resources });
 
@@ -273,6 +289,8 @@ export default class Application {
     this.mount();
 
     this.initialRoute = window.location.href;
+
+    caughtInitializationErrors.forEach(handler => handler());
   }
 
   // TODO: This entire system needs a do-over for v2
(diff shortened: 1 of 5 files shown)


framework/core/js/src/common/Application.tsx
```diff
@@ -262,7 +263,22 @@ export default class Application {
   }
 
   public boot() {
-    this.initializers.toArray().forEach((initializer) => initializer(this));
+    const caughtInitializationErrors: CallableFunction[] = [];
+
+    this.initializers.toArray().forEach((initializer) => {
+      try {
+        initializer(this);
+      } catch (e) {
```
askvortsov1, line 271: Shouldn't we rethrow `e` so the original error is accessible?

framework/core/js/src/common/Application.tsx
```diff
@@ -262,7 +263,22 @@ export default class Application {
...
+        const extension = initializer.itemName.includes('/')
+          ? initializer.itemName.replace(/(\/flarum-ext-)|(\/flarum-)/g, '-')
+          : initializer.itemName;
+
+        caughtInitializationErrors.push(() => fireApplicationError({
+          userTitle: extractText(app.translator.trans('core.lib.error.extension_initialiation_failed_message', { extension })),
+          consoleTitle: `${extension} failed to initialize`,
```
askvortsov1: Why keep this untranslated?

framework/core/js/src/common/helpers/fireApplicationError.ts
```diff
@@ -0,0 +1,24 @@
...
+ * Fire a Flarum error which is shown in the JS console for everyone and in an alert for the admin.
+ *
+ * @param options
+ * {
+ *   userTitle: a user friendly title of the error, should be localized.
+ *   consoleTitle: an error title that goes in the console, doesn't have to be localized.
+ * }
+ * @param errors
+ */
+export default function fireApplicationError(options: { userTitle: string, consoleTitle: string }, ...errors: any) {
```
askvortsov1: Ooo I like this!

askvortsov1 (changes requested):

framework/core/js/src/common/Application.tsx
SychO9: The error is still accessible since we display it with `console.error` in the introduced helper. (see screenshot above)
If we throw the execution halts and the app will display the nojs content:

https://github.com/flarum/framework/blob/b64003cba5ae77575babccc1803773d12fe5e2f7/framework/core/views/frontend/app.blade.php#L27-L36

framework/core/js/src/common/Application.tsx
SychO9: Didn't think it would need translation since it's a console message along with English error messages. These will most of the time be copy-pasted to extension authors.

dsevillamartin:
Seems to work fine.

---

I do wonder if there's something we could do to make it able to be monkey patched by e.g. Sentry in order to report these errors. Only way these errors would be caught now would be with console logging enabled.

Think that's an issue more in general with exporting functions, however. I believe we've had some issues relating to that in the past? Unsure. Just some thoughts.

framework/core/js/src/common/Application.tsx
dsevillamartin: I agree with that there's no reason to translate this.

SychO9:
Good point for a good use case. But yeah if it were to be done, monkey patching the function itself would be the best approach, which yeah isn't possible right now, but will be once the export registry is tackled, which is on the 1.x roadmap.

framework/core/js/src/common/Application.tsx
askvortsov1: Makes sense!

askvortsov1 (approved):

dsevillamartin (approved):

PR #3351, opened 2022-03-18
WIP fix(tags)!: reduce tag filter query from 3+ second

luceos:
This PR removes the subqueries inserted for each selected tag. As this
caused the query to take over 3 seconds on communities with hundred
million posts.

```sql
and (`discussions`.`id` in (select `discussion_id` from `discussion_tag` where `tag_id` = 105))
```

The above part takes on average 3.85 seconds, with a join it takes 0.00
seconds.

BREAKING CHANGE: extending the query might have different outcome as we
now use a join and no subquery.

**Necessity**

**Confirmed**

**Required changes:**

---

The full query that caused us to investigate:

```sql
select `discussions`.*
from `discussions`
where (`discussions`.`id` not in (select `discussion_id`
                                  from `discussion_tag`
                                  where `tag_id` not in (select `tags`.`id`
                                                         from `tags`
                                                         where (`tags`.`id` in (select `perm_tags`.`id`
                                                                                from `tags` as `perm_tags`
                                                                                where (`perm_tags`.`is_restricted` = 1 and
                                                                                       `perm_tags`.`id` in ('202', '205', '206', '89'))
                                                                                   or `perm_tags`.`is_restricted` = 0) and
                                                                (`tags`.`parent_id` in (select `perm_tags`.`id`
                                                                                        from `tags` as `perm_tags`
                                                                                        where (`perm_tags`.`is_restricted` = 1 and
                                                                                               `perm_tags`.`id` in ('202', '205', '206', '89'))
                                                                                           or `perm_tags`.`is_restricted` = 0) or
                                                                 `tags`.`parent_id` is null)))))
  and (`discussions`.`is_private` = 0 or (((`discussions`.`is_approved` = 0 and (`discussions`.`user_id` = 106739 or
                                                                                 ((`discussions`.`id` not in
                                                                                   (select `discussion_id`
                                                                                    from `discussion_tag`
                                                                                    where `tag_id` not in
                                                                                          (select `tags`.`id`
                                                                                           from `tags`
                                                                                           where (`tags`.`id` in
                                                                                                  (select `perm_tags`.`id`
                                                                                                   from `tags` as `perm_tags`
                                                                                                   where (`perm_tags`.`is_restricted` = 1 and `perm_tags`.`id` in ('89'))) and
                                                                                                  (`tags`.`parent_id` in
                                                                                                   (select `perm_tags`.`id`
                                                                                                    from `tags` as `perm_tags`
                                                                                                    where (`perm_tags`.`is_restricted` = 1 and `perm_tags`.`id` in ('89'))) or
                                                                                                   `tags`.`parent_id` is null))))) and
                                                                                  exists(select *
                                                                                         from `tags`
                                                                                                  inner join `discussion_tag` on `tags`.`id` = `discussion_tag`.`tag_id`
                                                                                         where `discussions`.`id` = `discussion_tag`.`discussion_id`)))) or
                                           (`discussions`.`id` in (select distinct `recipients`.`discussion_id`
                                                                   from `recipients`
                                                                   where `recipients`.`removed_at` is null
                                                                     and (`recipients`.`user_id` = 106739))))))
  and (`discussions`.`hidden_at` is null or `discussions`.`user_id` = 106739 or
       ((`discussions`.`id` not in (select `discussion_id`
                                    from `discussion_tag`
                                    where `tag_id` not in (select `tags`.`id`
                                                           from `tags`
                                                           where (`tags`.`id` in (select `perm_tags`.`id`
                                                                                  from `tags` as `perm_tags`
                                                                                  where (`perm_tags`.`is_restricted` = 1 and 0 = 1)) and
                                                                  (`tags`.`parent_id` in (select `perm_tags`.`id`
                                                                                          from `tags` as `perm_tags`
                                                                                          where (`perm_tags`.`is_restricted` = 1 and 0 = 1)) or
                                                                   `tags`.`parent_id` is null))))) and exists(select *
                                                                                                              from `tags`
                                                                                                                       inner join `discussion_tag` on `tags`.`id` = `discussion_tag`.`tag_id`
                                                                                                              where `discussions`.`id` = `discussion_tag`.`discussion_id`)))
  and (`discussions`.`comment_count` > 0 or `discussions`.`user_id` = 106739 or
       ((`discussions`.`id` not in (select `discussion_id`
                                    from `discussion_tag`
                                    where `tag_id` not in (select `tags`.`id`
                                                           from `tags`
                                                           where (`tags`.`id` in (select `perm_tags`.`id`
                                                                                  from `tags` as `perm_tags`
                                                                                  where (`perm_tags`.`is_restricted` = 1 and 0 = 1)) and
                                                                  (`tags`.`parent_id` in (select `perm_tags`.`id`
                                                                                          from `tags` as `perm_tags`
                                                                                          where (`perm_tags`.`is_restricted` = 1 and 0 = 1)) or
                                                                   `tags`.`parent_id` is null))))) and exists(select *
                                                                                                              from `tags`
                                                                                                                       inner join `discussion_tag` on `tags`.`id` = `discussion_tag`.`tag_id`
                                                                                                              where `discussions`.`id` = `discussion_tag`.`discussion_id`)))
  and (`discussions`.`id` in (select `discussion_id` from `discussion_tag` where `tag_id` = 193))
order by `is_stickiest` desc, `is_sticky` desc, `last_posted_at` desc
limit 21;
```


luceos:
I have tested this extensively locally and am now working towards testing this out in a staging environment.

askvortsov1 (approved):
From a code perspective, looks good. IMO it's good to merge after your testing.

SychO9 (dismissed):
Code looks good, haven't tested locally, we should introduce integration tests for gambits.

Another improvement could be to grab all tag IDs at once instead of calling `getIdForSlug` every time (query for every tag filter), impact would only be noticeable with a large number of tag filters applied.

luceos:
> Another improvement could be to grab all tag IDs at once instead of calling `getIdForSlug` every time (query for every tag filter), impact would only be noticeable with a large number of tag filters applied.

The performance hit isn't from the tags per se, but from the subquery that selects all discussions belonging to those tags. So your optimalisation by itself would have a far less positive outcome than using a join in my improvement.

---

Hold off from merging this, I want to push this and the post number changes into one branch for testing before we merge it.

BartVB:
If I understand correctly Sami means moving `getIdForSlug()` out of the `foreach` (or even getting rid of the foreach altogether if possible) _and_ using a join.

The performance impact of that additional change will be much smaller, but doing a `foreach` in a query is suboptimal (to use an euphemism). Not the biggest issue in this case, but removing extraneous queries one at a time would really help in the long run. Would also set an example for extension developers.

luceos:
What would probably be a better solution would be to use an in-memory (array or replacable) cache on `getIdForSlug` altogether. That would reduce the number of queries to 1 for all similar requests over the whole product, automatically including extensions.

What do you think? I am also not sure it should be part of this PR, but it could.

SychO9:
It was just a side-comment about something I noticed, let's not do anything about it in this PR 👍🏼

tankerkiller125:
@luceos I've been taking a quick look at places where database caching of some kind would be beneficial, this is for sure one of the use cases where a replaceable cache would be good. (Redis or other KV store would be preferable for HA setups/reasons)

askvortsov1:
Either way, I think this is good to merge (squash please!), assuming the tests you're running are going well.

luceos:
No confirmation from actual staging environment yet, holding off from merging.

BartVB:
I don't think using a cache for `getIdForSlug` would be the best way forward. Those queries are as light as they get, using Redis wouldn't help in that case because time is spent on network latency, not processing. Network latency is approx the same for a DB server and Redis.

IMO that energy could be better spent in reducing the amount of N+1 queries in Flarum. There are quite a few places where queries are run in a `foreach()`. Also in this case most of the performance increase would come from reducing the number of hits on a database/caching layer.

> let's not do anything about it in this PR

That sounds like a smart move.

The join/sorting issue
------

When it comes to the query changes in this PR; it doesn't bring the performance increase that we hoped for.  For some reason MySQL still uses `discussion_tag` as the primary table for joins which results in horrible performance when sorting.

The optimized query (which returns in a couple of milliseconds) has:
```sql
left join discussion_tag dt on (discussions.id = dt.discussion_id and tag_id = 193)
```

I had expected the optimizer to figure out that both are the same, but apparently it doesn't (at least not on Percona 8.0.26).

The queries produced by this PR contain:
```sql
left join `discussion_tag` on `discussions`.`id` = `discussion_tag`.`discussion_id`
[...]
and (`discussion_tag`.`tag_id` = 105)
```

This is the first row of an EXPLAIN:

```
           id: 1                                                                                                                  
  select_type: PRIMARY
        table: discussion_tag
   partitions: NULL  
         type: ref                                               
possible_keys: PRIMARY,discussion_tag_tag_id_foreign
          key: discussion_tag_tag_id_foreign
      key_len: 4                                                                                                                  
          ref: const
         rows: 90522 
     filtered: 100.00                                                                                                             
        Extra: Using where; Using index; Using temporary; Using filesort
```

If it's using `discussion_tag` then it can't use `discussions_is_sticky_last_posted_at_index` for sorting. That means that it needs to create a temporary table with all results (90k rows in this case) and then sort them manually.

I can think of some SQL hacks to work around this, but I'm hoping that there is an Eloquent way to do this?

extensions/tags/src/Query/TagFilterGambit.php
```diff
@@ -55,23 +55,23 @@ protected function constrain(Builder $query, $rawSlugs, $negate)
...
+                        $query->orWhereNull('discussion_tag.tag_id');
+                    } elseif ($slug === 'untagged' && $negate) {
+                        $query->orWhereNotNull('discussion_tag.tag_id');
+                    } elseif ($id = $this->tags->getIdForSlug($slug)) {
+                        $query->orWhere(
+                            'discussion_tag.tag_id',
+                            $negate ? '!=' : '=',
+                            $id
+                        );
+                    }
```
SychO9: Let's try this instead:
```suggestion
        $query
            ->distinct()
            ->join('discussion_tag', function (\Illuminate\Database\Query\JoinClause $join) use ($slugs, $negate) {
                $join
                    ->on('discussions.id', '=', 'discussion_tag.discussion_id')
                    ->where(function (\Illuminate\Database\Query\JoinClause $join) use($slugs, $negate) {
                        foreach ($slugs as $slug) {
                            if ($slug === 'untagged' && ! $negate) {
                                $join->orWhereNull('discussion_tag.tag_id');
                            } elseif ($slug === 'untagged' && $negate) {
                                $join->orWhereNotNull('discussion_tag.tag_id');
                            } elseif ($id = $this->tags->getIdForSlug($slug)) {
                                $join->orWhere(
                                    'discussion_tag.tag_id',
                                    $negate ? '!=' : '=',
                                    $id
                                );
                            }
                        }
                    });
            });
```

SychO9 (commented):
yeah, that's pretty bad, it traverses all 90k rows (all existing rows) for the results.

I tested and suggested a block of code that should apply the desired query. However, there is actually an issue I noticed about using a join, if you search with multiple tag gambits/filters, it will attempt to apply the same join again, which means join column conditions won't be unique with the `discussion_tag` table having been previously joined. We need to use a unique table alias every time.

extensions/tags/src/Query/TagFilterGambit.php
BartVB: Alas that doesn't result in a significant speedup. But if I change `->join()` into `->leftJoin()` then the request times drop dramatically. Seems to be doing what we want!

As said; I'm not exactly proficient with Eloquent but can't the `->where()` be replaced with `->on()`? That way `\Illuminate\Database\Query\JoinClause` isn't needed. That would be a bit more readable/cleaner because in the end it's not really a `WHERE` but a part of the `ON` part of that `LEFT JOIN`.

Great work!

askvortsov1 (approved):
LGTM! Please squash when merging

SychO9 (changes requested):
We can't merge this yet until we figure out how best to uniquely alias the joined table (try applying the filter/gambit multiple times to see what I mean `tag:support tag:feedback`)

> However, there is actually an issue I noticed about using a join, if you search with multiple tag gambits/filters, it will attempt to apply the same join again, which means join column conditions won't be unique with the discussion_tag table having been previously joined. We need to use a unique table alias every time.

luceos:
> We can't merge this yet until we figure out how best to uniquely alias the joined table (try applying the filter/gambit multiple times to see what I mean `tag:support tag:feedback`)
> 
> > However, there is actually an issue I noticed about using a join, if you search with multiple tag gambits/filters, it will attempt to apply the same join again, which means join column conditions won't be unique with the discussion_tag table having been previously joined. We need to use a unique table alias every time.

Yeah we noticed it's broken. Will fix asap. Sorry about that.

luceos:
@SychO9 I've aliased the table, should be unique for whatever passthrough. Unless the same slugs were used twice.

luceos:
According to @BartVB we cannot merge this yet.

BartVB:
This issue can be reduced to:

```sql
select discussions.id 
  from discussions 
    left join discussion_tag on discussions.id = discussion_tag.discussion_id
  where discussion_tag.tag_id = 105 
  order by is_sticky desc, last_posted_at desc
  limit 21;
```
Which is slow.

Where it went wrong is that `LEFT JOIN on tag_id = 105` is not the same as `LEFT JOIN ... WHERE tag_id = 105`. In the first instance you would of course get all rows from the discussions table, rows where `tag_id=105` get extra information from the `discussion_tag` table added. That's quick because MySQL hardly needs to filter so it uses `discussions` as the first join table. Because of that it can use an index for sorting. Problem is that it will not return the correct results.

If you add the `where` clause or use an `inner join` then you'll get the correct results, but `tag_id` has low cardinality so MySQL first takes matching rows from `discussion_tag` and then joins that with `discussions`. That would make sense if we didn't need to sort... Without the index quite a few rows need to be sorted, depending on how popular a tag is. I.e. if we change the query to `WHERE tag_id = 206` then the query returns instantly because that tag_id only has 6 discussions, instead of the 123.101 discussions in tag_id = 105.

So in a lot of cases (where there are no a lot of rows in a tag) the MySQL optimiser choses right.

It's possible to force MySQL to use the `discussions_is_sticky_last_posted_at_index` index. For instance with a STRAIGHT_JOIN:

```sql
select discussions.id 
  from discussions 
    straight_join discussion_tag on discussions.id = discussion_tag.discussion_id
  where discussion_tag.tag_id = 105 
  order by is_sticky desc, last_posted_at desc
  limit 21;
```

or

```sql
select discussions.id 
  from discussions force index (discussions_is_sticky_last_posted_at_index)
    left join discussion_tag on discussions.id = discussion_tag.discussion_id
  where discussion_tag.tag_id = 105 
  order by is_sticky desc, last_posted_at desc
  limit 21;
```

But doing that would make queries slow for queries that involve tags with only a few discussions because a very large part of the index needs to be read and skipped in those cases.

So it's back to the drawing board it seems...

PR #3354, opened 2022-03-19
fix: don't duplicate path in URL to post on subdirectory installs.

askvortsov1:


luceos:
Feels like this issue solves **one** instance of a problem, a problem that can cascade into other areas including 3rd party extensions. What is exactly the cause? Is baseUrl sometimes the domain + path and sometimes just the path?

dsevillamartin:
@luceos `baseUrl` has the URL in `config.php`, so it will always have the subfolder if there is one. `app.route` will return an URL with the subfolder as well because it is needed for navigation to the correct page (`/subfolder/d/...`).

framework/core/js/src/forum/components/PostMeta.js
```diff
@@ -55,6 +55,8 @@ export default class PostMeta extends Component {
    * @returns {string}
    */
   getPermalink(post) {
-    return app.forum.attribute('baseUrl') + app.route.post(post);
+    // URL.hostname transformation is necessary to avoid duplicating
+    // path in subdirectory installs.
+    return URL(app.forum.attribute('baseUrl')).hostname + app.route.post(post);
```
davwheat: This will not keep URLs which contain ports, nor the protocol. We should use `.origin` instead.


```suggestion
    return new URL(app.forum.attribute('baseUrl')).origin + app.route.post(post);
```

davwheat (changes requested):

SychO9 (approved):

davwheat (approved):

PR #3348, opened 2022-03-12
Assorted Typing Fixes

askvortsov1:
Moving one step closer to having all the hundreds of CI jobs on this monorepo be green.

PR #3356, opened 2022-03-20
fix: `Post--by-start-user` not working

SychO9:
This is an odd one, https://github.com/flarum/framework/pull/3170 was supposed to fix this, but I'm not sure where I got `startUserId` from, that has never been part of the discussion frontend attributes. (and the title of that fix refers to the wrong class as well).

This PR loads the author relationship by default to compare against.

We could also just introduce the attribute above, instead of loading the relationship.

**Necessity**

**Confirmed**

### DIFF
--- a/framework/core/js/src/forum/components/Post.tsx
+++ b/framework/core/js/src/forum/components/Post.tsx
@@ -132,7 +132,7 @@ export default abstract class Post<CustomAttrs extends IPostAttrs = IPostAttrs> 
       classes.push('Post--by-actor');
     }
 
-    if (user && user?.id() === discussion.attribute('startUserId')) {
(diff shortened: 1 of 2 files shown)


davwheat (approved):

tankerkiller125 (approved):

askvortsov1 (approved):

luceos:
Does this PR need to have upstream changes for tests to succeed?

SychO9:
it will be merged after https://github.com/flarum/framework/pull/3348 is merged.

askvortsov1:
Yep, although it will need to be updated since `Post` was converted to TypeScript in #3348

PR #3359, opened 2022-03-28
a11y: use `feed` aria role for DiscussionList

### LINKED ISSUE #3363: [A11Y] Apply `role="feed"` to post/discussion streams

davwheat:
**FIxes flarum/framework#3363**

- Uses `role=feed` on discussion list

Reference: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/feed_role

> If the number of articles is known, set [aria-setsize](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-setsize) on the articles themselves. However, if the total number is extremely large, indefinite, or changes often, set aria-setsize="-1" to indicate that the size of the feed is not known.

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/framework/core/js/src/forum/components/DiscussionList.js
+++ b/framework/core/js/src/forum/components/DiscussionList.js
@@ -49,11 +49,12 @@ export default class DiscussionList extends Component {
         <ul role="feed" aria-busy={isLoading} className="DiscussionList-discussions">
           {state.getPages().map((pg, pageNum) => {
             return pg.items.map((discussion, itemNum) => (
-              <li key={discussion.id()}
-                  data-id={discussion.id()} 
-                  role="article"
-                  aria-setsize="-1"
-                  aria-posinset={pageNum * pageSize + itemNum}
+              <li
+                key={discussion.id()}
+                data-id={discussion.id()} 
+                role="article"
+                aria-setsize="-1"
+                aria-posinset={pageNum * pageSize + itemNum}
               >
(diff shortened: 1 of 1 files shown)


framework/core/js/src/forum/components/DiscussionList.js
```diff
@@ -39,13 +42,15 @@ export default class DiscussionList extends Component {
...
+    const pageSize = state.pageSize;
+
     return (
-      <div className={'DiscussionList' + (state.isSearchResults() ? ' DiscussionList--searchResults' : '')}>
-        <ul className="DiscussionList-discussions">
-          {state.getPages().map((pg) => {
-            return pg.items.map((discussion) => (
-              <li key={discussion.id()} data-id={discussion.id()}>
-                {DiscussionListItem.component({ discussion, params })}
+      <div class={classList('DiscussionList', { 'DiscussionList--searchResults': state.isSearchResults() })}>
```
askvortsov1, line 48: Shouldn't we still use `className`?

davwheat: It makes no difference in Mithril, and newer stuff we've been doing has been using class, or supporting passing either to components (see the Button TS rewrite).

askvortsov1 (approved):

SychO9 (approved):

PR #3378, opened 2022-04-04
feat(emoji): support emoji 14 & unicode 14

dsevillamartin:
- Update `twemoji` & `simple-emoji-map` for emoji 14 & unicode 14
    - `forum.js`: 63.9 KB to 74.4 KB (wouldn't it be nice to lazy load these for the composer... 😅)

The shortcodes of emojis changed in `emojibase-data` v6 or v7, which `simple-emoji-map` uses. 
Do we care that there are differences there? In other words, do we want to keep using the legacy shortnames?

The following file would allow the continued use of the old emoji shortcodes with support for the new ones.
(Or use a set other than `emojibase` entirely - https://github.com/milesj/emojibase/tree/emojibase-data%407.0.1/packages/data/en/shortcodes)
<details>

<summary><code>.simple-emoji-map</code></summary>

```json
{
    "shortcodes": {
        "dataset": "emojibase-legacy",
        "fallbackDataset": "emojibase"
    },
}
```

</details>

**Screenshot**
Current vs new emojis (some with new shortnames):

Artboard

**Necessity**

- ~~Has the problem that is being solved here been clearly explained?~~

**Confirmed**

- ~~Tests have been added, or are not appropriate here.~~

### DIFF
--- a/extensions/emoji/js/package.json
+++ b/extensions/emoji/js/package.json
@@ -4,18 +4,18 @@
     "version": "0.0.0",
     "prettier": "@flarum/prettier-config",
     "dependencies": {
-        "simple-emoji-map": "^0.4.1",
-        "twemoji": "^13.1.0"
+        "simple-emoji-map": "^0.5.1",
+        "twemoji": "^14.0.2"
     },
     "devDependencies": {
-        "prettier": "^2.5.1",
-        "flarum-webpack-config": "^2.0.0",
-        "webpack": "^5.65.0",
-        "webpack-cli": "^4.9.1",
         "@flarum/prettier-config": "^1.0.0",
         "flarum-tsconfig": "^1.0.2",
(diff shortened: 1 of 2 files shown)


davwheat (approved):

davwheat:
I don't think there's a need to support the old shortcodes, but if there isn't any big difference in bundle size, why not?

askvortsov1 (approved):

dsevillamartin:
> but if there isn't any big difference in bundle size, why not?

Don't think there's a big diff, but there will probs be differences with names of new emojis vs old emojis regarding what one would expect emojis to be called.

I feel towards using the new set as that way we don't have to bother making sure we keep using the legacy shortcodes.

askvortsov1:
Sounds good, go ahead and merge when ready!

PR #3379, opened 2022-04-05
fix: notification timestamp mismatched from other core code

davwheat:

When running Flarum with a customised config on a server which uses a non-UTC timezone, there is a mismatch with notifications since these are handled as UTC, while other timestamps are handled in local time. This PR changes notifications to use local time, like everything else in core.

Does this present an issue for existing communities? I argue not, as noone else has noticed this until now, so a fix would have just as small of an effect as the current issue has.

**Necessity**

**Confirmed**


imorland (approved):

askvortsov1 (approved):

BartVB:
That was quick :)

PR #3380, opened 2022-04-06
fix: restricted sibling tags appearing for unauthorized members on the sidebar

SychO9:
Loading the `parent.children` tag relation grabs the models without checking for permissions.
This PR makes sure that the relation is passed through a permission check first (repository class code) and then if a `parent.children.parent` relation is also included, it manually attached that relation instead of trying to eager load it.

* --> add `parent.children` to relation permission check array.
* --> check if a `parent.children.parent` relation is passed in the controller.
* --> manually attach the `parent` to `children`.
* --> remove eager loading of `parent.children.parent`.
* --> integration tests.

**Necessity**

**Confirmed**


imorland:
Thank you for putting this together @SychO9 - I just had a chance to test it locally.

It works fine from the perspective of the siblings are not longer visible (where applicable), but I do run into this when reloading on the parent tag:

```
Warning: Attempt to read property "children" on null in /var/www/framework/extensions/tags/src/Api/Controller/ShowTagController.php on line 67

Warning: foreach() argument must be of type array|object, null given in /var/www/framework/extensions/tags/src/Api/Controller/ShowTagController.php on line 67
```

SychO9:
ah good catch, we really need our test suite to catch warnings, I'll fix the error and take care of tweaking the tests separately as part of (https://github.com/flarum/framework/issues/3383)

PR #3382, opened 2022-04-07
fix: Check for null before attempting ->getRelations()

imorland:
Follows up on https://github.com/flarum/mentions/pull/79

With this implementation, when a 3rd party extension uses the renderer for a user mention, and does not pass any `context`, an error occurs:

`Call to a member function getRelations() on null in /opt/flarum/vendor/flarum/mentions/src/Formatter/FormatUserMentions.php:48`

This is because `isset` does not first check that `$context` is not null, before attempting to evaluate `->getRelations()`

This PR looks to address this. Locally tested.


davwheat (approved):
Makes sense!

SychO9 (approved):

PR #3385, opened 2022-04-09
fix: languages with combining characters cannot be searched

SychO9:
The regular expression used in the Fulltext search gambit attempts to remove non-words to avoid triggering MySQL boolean mode, however, it also removes special characters that combine with words in languages such as Telugu and Devanagari.

This pull request tweaks the regular expression to take those into consideration and allow searching in those languages. (Before this fix, searching for example for: `नागरी` resulted in actually searching for `न गर `)

**Necessity**

**Confirmed**

### DIFF
--- a/framework/core/src/Discussion/Search/Gambit/FulltextGambit.php
+++ b/framework/core/src/Discussion/Search/Gambit/FulltextGambit.php
@@ -25,7 +25,7 @@ class FulltextGambit implements GambitInterface
         // Replace all non-word characters with spaces.
         // We do this to prevent MySQL fulltext search boolean mode from taking
         // effect: https://dev.mysql.com/doc/refman/5.7/en/fulltext-boolean.html
-        $bit = preg_replace('/[^\p{L}\p{N}_]+/u', ' ', $bit);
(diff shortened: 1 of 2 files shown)


SychO9:
for the `\p{M}` refer to https://www.regular-expressions.info/unicode.html

askvortsov1 (approved):
Wow... This is kind of staggering, not gonna lie: if this has been the main reason completely breaking UTF-8 search, this might genuinely be one of the most impactful PRs in Flarum history.

askvortsov1 (commented):
Only thought: perhaps we should also test that we can search titles for other languages?

askvortsov1 (approved):

davwheat (approved):

luceos (approved):
At least there's tests now 🙈

justoverclockl:
this will affect also filter options? because actually filter doesn't work with other languages:

```
 app.store
      .find('discussions', {
        'filter[q]':  ' తెలుగు భాష',
```

SychO9:
when you use `filter[q]` you're performing a search, so yes.

PR #3386, opened 2022-04-10
feat: add utf-8 slug driver for discussions

askvortsov1:
TSIA.

Question is whether or not we want this. I honestly don't see a solid reason not to include it, since we're hoping to better support non-ascii languages. The only issue is that when copying the URL from the browser bar, the resulting value will be percent-encoding, but this is a browser issue, not one that we can do anything about. With this in mind, I think it's reasonable to include this, but not set it as the default option.


davwheat (approved):

davwheat:
Seems like a smart enough idea to me.

framework/core/src/Discussion/Utf8SlugDriver.php
```diff
@@ -0,0 +1,44 @@
...
+    protected $discussions;
+
+    public function __construct(DiscussionRepository $discussions)
+    {
+        $this->discussions = $discussions;
+    }
+
+    public function toSlug(AbstractModel $instance): string
+    {
+        $title = preg_replace('![-\s]+!u', '-', $instance->title);
```
SychO9, line 30: we should probably ~~escape~~ remove non-word characters though I think

SychO9 (commented):
I don't see why not!

perhaps one integration test would be good to have.

framework/core/src/Discussion/Utf8SlugDriver.php
tankerkiller125: Would \W and \D be better so that anything that's not a Digit or Word Character is replaced? From my knowledge these properly support UTF-8?

SychO9: I think this is the same as https://github.com/flarum/framework/blob/main/framework/core/src/Discussion/Search/Gambit/FulltextGambit.php#L28, since we wouldn't want to be removing the characters we're trying to support.

PR #3388, opened 2022-04-11
chore: further cleans up all packages into one mono repo

luceos:
placeholder, this needs a todo


askvortsov1:
I haven't been able to fully review this since it doesn't load in my browser, but there seems to be a LOT of excessive renaming and restructuring of folder paths. Is this really necessary? It makes the PR much larger / more difficult to review, and it'll make it a lot harder to get file change history (since most git clients don't necessarily follow renames).

luceos:
Following PSR makes a ton of sense for the long run though, and from my point of view we'd better introduce mono repo properly as soon as possible than having to live with a half hearted pseudo state. In this state contributing has become harder, if not nearly impossible. Our goal with mono repo was to make things simpler, if we have to give up our history to do so I'll gladly take the blame on that one.

- Folder renames are required for PSR-4 compliant namespacing.
- Moved tests per extension to one tests directory for the framework with tests per extension (`extensions/tags/tests/integration` => `tests/integration/Tags`) so that tests can be executed using one phpunit file.
- Removed all obsolete `require-dev` keys.
- Removed all `extra` and other configuration settings from extension (read only) composer files.

(more soon)

SychO9:
I don't think we can mimic Symfony or Laravel, each extension test suite is and should stay isolated, the package manager for example goes to extra lengths to test installing composer packages.

I'm not saying to make any changes either, but we need to tackle this gradually. We can start by answering the question: what do we need the most from a central `composer.json` right now? maybe we can achieve it differently

askvortsov1:
^^^

> Removed all extra and other configuration settings from extension (read only) composer files.

These should really stay: all the flarum cli config shouldn't be removed

> Folder renames are required for PSR-4 compliant namespacing.

If we add a separate key for each namespace, we don't need to rename anything

luceos:
The idea remains to run all tests for everything at once and to enable installing core+bundled extensions in one swoop. But the fact that you think Package Manager is more stand alone should have implied to keep it outside of the mono repo, perhaps the same counts for the testing package. Because in the end, the mono repo is about Flarum, testing the testing package while changing core or a bundled extension should not affect the testing package. That a change to the testing package cascades into the framework, sure, but that can be achieved by keeping the testing package out of the framework. The same principle can apply to everything in js-packages.

askvortsov1:
> The idea remains to run all tests for everything at once and to enable installing core+bundled extensions in one swoop. But the fact that you think Package Manager is more stand alone should have implied to keep it outside of the mono repo, perhaps the same counts for the testing package. Because in the end, the mono repo is about Flarum, testing the testing package while changing core or a bundled extension should not affect the testing package. That a change to the testing package cascades into the framework, sure, but that can be achieved by keeping the testing package out of the framework. The same principle can apply to everything in js-packages.

I'm sorry but I still don't understand what problem you're trying to solve with this.

I want to keep all our code together so that we have everything that might be running on a live Flarum instance in one place. I want to do this so that ongoing work is centralized, we don't have to do a song and dance to keep issues/planning organized, we don't need to check 20 different repositories for PRs, we don't need to coordinate PRs for one change across multiple repos, and maybe most importantly, all CI runs together so that we minimize the risk of breaking extensions with core changes. I also want to maintain a structure not too different from that of standalone repositories.

I am aware that Laravel and Symfony have their own standards and conventions when it comes to monorepos. We are not trying to solve the same problems that they are. At the moment, the problems I see with our setup are:

- you can't install dev versions of packages in the monorepo
- we need to run hundreds of GH Actions jobs to run backend tests. It would be nice to consolidate this. That doesn't necessarily mean putting all tests in the same place, just running all tests together.

The vast majority of changes you seem to be intending here are to stick to Laravel's convention rather than to tackle the issues we actually face here. Please correct me if I'm misunderstanding something here.

luceos:
Okay so we need to find a solution that works without losing the structure we have. Let me think about this and come up with a second attempt.

askvortsov1:
> Okay so we need to find a solution that works without losing the structure we have. Let me think about this and come up with a second attempt.

Aside from the 2 I mentioned above, what problems do you think we should be considering when deciding how to proceed?

luceos:
I cannot use a branch of the framework inside a flarum community.

askvortsov1:
> * you can't install dev versions of packages in the monorepo

Wouldn't that fall under this point though? If you could install dev versions of the monorepo, you could also install branches onto communities.

luceos:
I've force pushed the composer.json changes that ref the direct namespace, while adding the deps and helper. Next up would be to fix the backend tests.

- [ ] remove workbench declarations in extensions
- [ ] phpunit definition in root
- [ ] remove phpunit files in extensions (?)

luceos:
@askvortsov1 I don't know how I feel about the changes to the Extension Manager. It will create a difference between production and framework, tests cover a different scenario than what is going to happen in a real hosting environment. Why exactly was this change necessary?

askvortsov1:
ExtensionManager needs to figure out which extensions exist in the system. It does so by checking the vendor/composer/installed.json file. That file only contains records for individually installed packages, NOT for packages installed as part of a repo / available via "replaces". Therefore, monorepo extensions are not visible to ExtensionManager. Therefore, all extensions in a monorepo are not available. So the forum is broken.

Apologies for the lack of eloquence, currently closing out an all-nighter.

davwheat:
Sleep!

luceos:
I don't like this at all to be honest. The framework exists to make things easier for developers, this extension manager modification makes the code more complex. I firmly believe we need to find a better way to solve this issue from the side of the framework repo, not any of the packages (including core). It just adds noise, noise to suppress a fault in our framework set up.

I think we need a better way to inject this data into Flarum without modifying as much of the code. The term `monorepo` shouldn't - in my opinion - come up in the code at all. I am sorry to be standfast on this, but this seems to steer away too much from the quality that we try to reach for.

I will come up with some ideas this week.

askvortsov1:
> I don't like this at all to be honest. The framework exists to make things easier for developers, this extension manager modification makes the code more complex. I firmly believe we need to find a better way to solve this issue from the side of the framework repo, not any of the packages (including core). It just adds noise, noise to suppress a fault in our framework set up.
> 
> I think we need a better way to inject this data into Flarum without modifying as much of the code. The term `monorepo` shouldn't - in my opinion - come up in the code at all. I am sorry to be standfast on this, but this seems to steer away too much from the quality that we try to reach for.
> 
> I will come up with some ideas this week.

If there's some way for `replaces` entries to show up in `composer/installed.json`, that would be much nicer. But I'm not convinced that there's a way to do so. Would be nice if there is though!

askvortsov1:
@luceos updated to not hardcode the monorepo structure. I suspect this is as close as we're gonna get to a good solution, at least for now.

luceos (commented):
Looking good.

SychO9 (approved):
I've spent some time playing around with composer, I thought we could do something with path repositories but doesn't seem like we can.

I'm ok with this approach for now. Looks good 👍🏼

PR #3402, opened 2022-04-18
fix: undefined element to anchorScroll triggers errors

davwheat:

Sentry has been picking up looooots of errors from here, mainly because `element` is `null` or `undefined`, and this should prevent most of them popping up.

Nothing in particular.

**Necessity**

**Confirmed**


dsevillamartin:
In case we want to investigate root cause instead of this fix -- as we "decided" in the past, https://github.com/flarum/framework/pull/1881#discussion_r322475850.

https://github.com/flarum/framework/blob/09bed537b35c5f22333e526e50740c7aa35f1e45/framework/core/js/src/forum/states/PostStreamState.ts#L245

davwheat:
VS Code seemed to indicate this worked, but I guess not lol

davwheat:
Yeah, my original fix made no sense... I've redone it.

Do we want to silently return for the first one? This means that the callback will never run, which could be an issue.

I'm _guessing_ that the reason this happens is when you scroll very fast, a new page will start loading, and an old page may be removed, so if the anchor was around an item on that old page it could generate this error after the callback is complete.

dsevillamartin:
Not sure about behavior, but i believe $(element) is always truthy, even if element doesn't exist? Unless I'm completely wrong here, i think it's .offset() that is falsy if the element doesn't exist.

davwheat:
Yes, sorry, I'm not with it this morning 🙈

PR #3407, opened 2022-04-24
chore: fix package manager tests

SychO9:
Fixes the backend tests for the package manager so we can move forward with the queue implementation.

### DIFF
--- a/extensions/package-manager/tests/phpunit.integration.xml
+++ b/extensions/package-manager/tests/phpunit.integration.xml
@@ -19,10 +19,7 @@
     <testsuites>
         <testsuite name="Flarum Integration Tests">
             <directory suffix="Test.php">./integration</directory>
-             <exclude>./integration/tmp</exclude>
+            <exclude>./integration/tmp</exclude>
         </testsuite>
     </testsuites>
-    <php>
-        <env name="FLARUM_TEST_TMP_DIR_LOCAL" value="tests/integration/tmp" force="true" />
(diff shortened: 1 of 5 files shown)


extensions/package-manager/tests/phpunit.integration.xml
```diff
@@ -19,10 +19,7 @@
     <testsuites>
         <testsuite name="Flarum Integration Tests">
             <directory suffix="Test.php">./integration</directory>
-             <exclude>./integration/tmp</exclude>
+            <exclude>./integration/tmp</exclude>
         </testsuite>
     </testsuites>
-    <php>
-        <env name="FLARUM_TEST_TMP_DIR_LOCAL" value="tests/integration/tmp" force="true" />
-    </php>
```
askvortsov1, line 27: Why move this from here to the composer.json?

askvortsov1 (approved):

extensions/package-manager/tests/phpunit.integration.xml
SychO9: it's not actually moving, rather, it's no longer necessary to set the env variable since you added explicit exclusion of the vendor directory within the `tmp` directory a few months back.

The addition in `composer.json` just makes sure that it uses one of the two env variable values, or defaults back to the default value we use which is `tests/integration/tmp`

tankerkiller125 (approved):

PR #3411, opened 2022-05-06
fix: incorrect tooltip docs for `tooltipVisible` attr

davwheat:

- fixes incorrect info in docblock for `tooltipVisible` attr

**Necessity**

**Confirmed**


clarkwinkelmann (approved):
Sounds good!

There are still other issues that prevent my code from working but this part looks all good!

askvortsov1 (approved):

PR #3414, opened 2022-05-06
fix: first posts having `NULL` as value for number column

SychO9:
**Introduced in #3358** 

We didn't account for new discussions (first posts) at that point the max returns `NULL` since there are no posts for that discussion yet. So this PR introduces a fallback value of 0.

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,3 @@
 node_modules
+vendor
+composer.lock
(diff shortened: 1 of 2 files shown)


luceos (commented):
Wouldn't a null coalescence on PHP level be much easier?

SychO9:
That's not possible since we're not evaluating the expression. The expression is part of the *post* insertion database query.

luceos:
> That's not possible since we're not evaluating the expression. The expression is part of the _post_ insertion database query.

Nice. Looking good to me!

luceos (approved):

davwheat (approved):
Confirmed working locally

PR #3418, opened 2022-05-10
feat: Queue package manager commands

SychO9:
Sorry for the one single massive commit, had to import changes from before monorepo migration.

* Adds option to queue the commands in the background.
* Adds UI to view queued commands.

**Screenshot**

**Necessity**

**Confirmed**

Controller communicating with a dispatcher object which chooses wether to run the command synchronously or asynchronously


extensions/subscriptions/src/Listener/SendNotificationWhenReplyIsPosted.php
```diff
@@ -27,6 +27,12 @@ public function __construct(Queue $queue)
 
     public function handle(Posted $event)
     {
+        $discussion = $event->post->discussion;
+
+        /** @var \Psr\Log\LoggerInterface $log */
+        $log = resolve('log');
+        $log->info("running subscriptions send notification when reply is posted listener. last_post_number: $discussion->last_post_number");
+
```
imorland, line 35: Hey @SychO9 just noticed this whilst looking at a seperate issue.

I'm not so sure we'd really want this output to the log, was this perhaps an accidental inclustion from some debugging or something?

SychO9: oh nice catch, that was for debugging in a different PR, it somehow slipped in while rebasing 🤔.

PR #3420, opened 2022-05-12
fix(regression): PostStreamState loadCount static member

SychO9:
Calls the static member directly.


davwheat (approved):

askvortsov1:
I think the problem is that `this` isn't bound. Can we call .bind(this) in lines 70 and 71 to solve this without compromising extensibility?

askvortsov1 (approved):

PR #3421, opened 2022-05-12
Make it possible to reply (or create a discussion) with another post type than `comment`

iPurpl3x:
Nothing breaking, just added the possibility to handle a `type` attribute coming from an API call to either create a discussion or reply to one. When the `type` attribute is provided, the according post model is used to create the post. This makes it possible to start a discussion with a post that is not of type `comment`. 

Obviously the custom type need to be registered using the `Post` extender and the `reply()` method needs to be implemented with the same parameters as for the `reply()` method of the `CommentPost` model.

When no `type` is found, the post will be considered as a `comment` by default.

Nothing is broken, the new ability to create a post with a custom type works as expected.

**Necessity**

**Confirmed**

### DIFF
--- a/framework/core/src/Post/Command/PostReplyHandler.php
+++ b/framework/core/src/Post/Command/PostReplyHandler.php
@@ -15,6 +15,7 @@ use Flarum\Foundation\DispatchEventsTrait;
 use Flarum\Notification\NotificationSyncer;
 use Flarum\Post\CommentPost;
 use Flarum\Post\Event\Saving;
+use Flarum\Post\Post;
 use Flarum\Post\PostValidator;
 use Illuminate\Contracts\Events\Dispatcher;
 use Illuminate\Support\Arr;
@@ -78,10 +79,16 @@ class PostReplyHandler
             $actor->assertCan('reply', $discussion);
         }
 
+        // Get the post type from the query data. When none is provided, we'll
+        // assume it's a comment.
+        $type = Arr::get($command->data, 'attributes.type', 'comment');
+        // Get the post model class that should be used to create it.
+        $postModel = Post::getModels()[$type];
+
         // Create a new Post entity, persist it, and dispatch domain events.
         // Before persistence, though, fire an event to give plugins an
         // opportunity to alter the post entity based on data in the command.
-        $post = CommentPost::reply(
+        $post = $postModel::reply(
             $discussion->id,
             Arr::get($command->data, 'attributes.content'),
             $actor->id,


framework/core/src/Post/Command/PostReplyHandler.php
```diff
@@ -78,10 +79,16 @@ public function handle(PostReply $command)
             $actor->assertCan('reply', $discussion);
         }
 
+        // Get the post type from the query data. When none is provided, we'll
+        // assume it's a comment.
+        $type = Arr::get($command->data, 'attributes.type', 'comment');
+        // Get the post model class that should be used to create it.
+        $postModel = Post::getModels()[$type];
```
askvortsov1, line 86: I wonder if this should be tied to some security / permissions model? I could envision someone using a custom API client to create post types that should only be allowed by the system.

framework/core/src/Post/Command/PostReplyHandler.php
```diff
@@ -78,10 +79,16 @@ public function handle(PostReply $command)
...
+
         // Create a new Post entity, persist it, and dispatch domain events.
         // Before persistence, though, fire an event to give plugins an
         // opportunity to alter the post entity based on data in the command.
-        $post = CommentPost::reply(
+        $post = $postModel::reply(
```
askvortsov1, line 91: Not all post models will have this interface for the `reply` method, so this isn't structurally sound.

askvortsov1 (changes requested):
I like the idea in theory, but I think we might need a bit more complicated approach, due to reasons I point out in comments below.

clarkwinkelmann (changes requested):
At the very least it needs more error handing.

Line 86 will throw a PHP array access warning if the key doesn't exist.

Line 91 will throw a 500 error if the method doesn't exist. Checking for an interface could be a solution.

I think both permission and validation could be achieved with a single interface. The interface could be called something like "text repliable (via API?)" But interface cannot be for static methods if I remember correctly? That's not a deal breaker, the method could be moved to non-static and the model instantiated before call.

I do have some concerns for other extensions as well. If an extension extends the `Post\Saving` event. would it be supposed to handle those posts like if they were instances of `CommentPost` ? Since many extensions have logic explicitely for `CommentPost`, setting a different type here might bypass a lot of logic from other extensions, which could be intentional but also not, depending on the use case.

This looks like something an extension could implement relatively easily, use `Saving` event and look for a specific `type` value in payload, otherwise ignore. Does having it core really make anything easier? If any of those special types has special handling, the `Saving` event will have to be used anyway.

luceos:
Moved to a roadmap, continuous effort task. We have some ideas on how to approach this, but we first need to decide on the best path forward and then draft a technical plan for implementation.

PR #3424, opened 2022-05-14
Make unread notifications method public

BurnNoticeSpy:
Makes the notifications method public in `framework/core/src/User/User.php`

### DIFF
--- a/framework/core/src/User/User.php
+++ b/framework/core/src/User/User.php
@@ -447,7 +447,7 @@ class User extends AbstractModel
      *
      * @return \Illuminate\Database\Eloquent\Collection
      */
-    protected function getUnreadNotifications()
+    public function getUnreadNotifications()
     {
         static $cached = null;
 


luceos (approved):

tankerkiller125 (approved):

imorland (approved):

davwheat (changes requested):
I am not comfortable merging this with the current caching system present on the method. It seems rather dangerous to publicly expose this, especially without clear documentation about the caching, since it'll only provide the correct data for the first user which it is requested for.

At the very least, I'd expect a map of User IDs to data, but I'd really want a more reliable method of caching instead.

BurnNoticeSpy:
So is this caching style done elsewhere? Can it be handled separately from this PR?

Thoughts on the danger of making this public from the other reviewers?

luceos:
I think I've come to agree this stance and to prevent this PR from dragging along, we should discuss options inside an issue before we move forward with a solution.

If you're up for taking the lead on this discussion @BurnNoticeSpy feel free to open an issue on this repository.

Thanks for the effort by the way, this is not about you or your code. This is about guaranteeing stability, security and compatibility in extensibility which we'll be responsible to maintain moving forward.

PR #3432, opened 2022-05-20
fix(mentions): post mentions with deleted author not rendering

### LINKED ISSUE #3427: Cannot mention post with deleted author
## Bug Report

**Current Behavior**
If you try mentioning a post whose author is [deleted], the preview will render correctly, but after saving the mention will stay as plain text.

**Steps to Reproduce**
1. Create post and delete author, either by deleting the user or setting `user_id` null in the database
2. Use reply button or `@` syntax to mention the post with deleted author
3. Preview shows mention as expected
4. Submit post
5. Final post doesn't contain mention

**Expected Behavior**
It should be possible to mention those posts. In fact it works fine if the author is deleted after the reply is created. But if the reply is created afterwards, it doesn't work.

**Screenshots**
Preview:
image
Submitted:
image

**Environment**
- Flarum version: 1.2.1, likely still in 1.3.0 because the master code still contains the issue right now

**Possible Solution**
The problem is in https://github.com/flarum/framework/blob/471947dfa78743f9feefd775346103f4d930dd07/extensions/mentions/src/ConfigureMentions.php#L124-L135 which only saves a mention if both the post and author exist.

I think we should just move `$tag->setAttribute('displayname', $post->user->display_name);` to a separate `if( $post->user)`. Not sure if we should put a different default value for `displayname`. Its value will already be set to `[deleted]` in `Formatter/FormatPostMentions` anyway.

The preview JavaScript code doesn't have that limitation https://github.com/flarum/framework/blob/471947dfa78743f9feefd775346103f4d930dd07/extensions/mentions/js/src/forum/utils/textFormatter.js#L23-L33

SychO9:
**Fixes #3427**

The issue details the problem and solution very well.
Remember to clear the cache when testing locally.

**Necessity**

**Confirmed**

### DIFF
--- a/extensions/mentions/src/ConfigureMentions.php
+++ b/extensions/mentions/src/ConfigureMentions.php
@@ -125,10 +125,13 @@ class ConfigureMentions
     {
         $post = CommentPost::find($tag->getAttribute('id'));
 
-        if ($post && $post->user) {
+        if ($post) {
             $tag->setAttribute('discussionid', (int) $post->discussion_id);
             $tag->setAttribute('number', (int) $post->number);
-            $tag->setAttribute('displayname', $post->user->display_name);
+
+            if ($post->user) {
+                $tag->setAttribute('displayname', $post->user->display_name);
+            }
 
             return true;
         }


askvortsov1 (approved):
Makes sense, having this logic separated is cleaner.

davwheat (approved):

PR #3434, opened 2022-05-20
docs: ItemList set methods throw errors when keys are not present

davwheat:

- Corrects errors in TSDoc annotations of ItemList

N/A

**Necessity**

**Confirmed**

### DIFF
--- a/framework/core/js/src/common/utils/ItemList.ts
+++ b/framework/core/js/src/common/utils/ItemList.ts
@@ -127,7 +127,7 @@ export default class ItemList<T> {
   /**
    * Replaces an item's content, if the provided item key exists.
    *
-   * If the provided `key` is not present, nothing will happen.
+   * If the provided `key` is not present, an error will be thrown.
    *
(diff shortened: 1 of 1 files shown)


SychO9 (approved):

imorland:
I'm personally in two minds about this.

I've been using these methods on the assumption that the docs are correct 🙈 , so part of me says that in the year that this has been released, the core behaviour should match the docs.

On the other side, throwing an error is helpful for tracking down typo's, etc

Thinking a while... 🤔

davwheat:
I mean we *could* ignore missing keys in prod, and error in dev?

askvortsov1 (approved):

askvortsov1:
> I mean we _could_ ignore missing keys in prod, and error in dev?

I believe we have a helper for this?

davwheat:
> I believe we have a helper for this?

Indeed we do! It's just a case of if it's worthwhile doing so.

PR #3439, opened 2022-05-23
fix: settings extender working only with first instances

### LINKED ISSUE #3438: Default settings values are not returned
## Bug Report

**Current Behavior**
When calling `$settings->get('some-key-that-is-not-already-in-the-database')`, having already defined a default value in an extensions' `extend.php`, the value returned is `null`

**Steps to Reproduce**
- In `extend.php`
```php
new (Extend\Settings())
  ->default('some-key-that-is-not-already-in-the-database', 'my-default-value'),
  ```
  - Call that settings key
  ```php
 $settings->get('some-key-that-is-not-already-in-the-database')
  ```
  - Notice the value is `null`

**Expected Behavior**
The default value as defined in `extend.php` should be returned.

**Environment**
- Flarum version: 1.3
- Website URL: local
- Webserver: n/a
- Hosting environment: n/a
- PHP version: 8.0.13
- Browser: n/a

**Additional Context**

SychO9:
**Fixes #3438**

This was an oversight, after extending the container binding it wasn't returned.
The extender still worked, but only partially, it didn't register all defaults. I'm not entirely sure yet how that even worked before.
I have tweaked the tests to catch this.

**Necessity**

**Confirmed**

### DIFF
--- a/framework/core/src/Extend/Settings.php
+++ b/framework/core/src/Extend/Settings.php
@@ -95,6 +95,8 @@ class Settings implements ExtenderInterface
 
                     $defaults->put($key, $value);
                 }
+
+                return $defaults;
             });
         }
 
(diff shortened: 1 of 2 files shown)


imorland (approved):

imorland:
Confirmed working! Thank you @SychO9

davwheat (approved):

PR #3442, opened 2022-05-31
Bug fix - Select the correct users to notify for subscribed posts

OwenMelbz:
Sorry, I don't have time to do a full tested and reviewed PR feedback form to due our project deadlines.

However this has been requested via: https://discuss.flarum.org/d/30928-subscription-notifications-not-firing

The issue was that the notification would only get dispatched to users who's last read number was the same as the latest post.

However we need to send the notification to all users who have a read number LESS than the current post to make sure they are notified.

This PR hopes to solve that.

### DIFF
--- a/extensions/subscriptions/src/Job/SendReplyNotification.php
+++ b/extensions/subscriptions/src/Job/SendReplyNotification.php
@@ -49,7 +49,7 @@ class SendReplyNotification implements ShouldQueue
         $notify = $discussion->readers()
             ->where('users.id', '!=', $post->user_id)
             ->where('discussion_user.subscription', 'follow')
-            ->where('discussion_user.last_read_post_number', $this->lastPostNumber)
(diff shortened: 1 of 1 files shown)


SychO9:
Hey! Thanks for the PR!

While this does solve the bug, it is a change in behavior, I believe doing this would send reply notifications not just once but every time there is a new post reply. The initial behavior was to just send a notification on the first reply only.

Ref: https://github.com/flarum/framework/issues/3443

Need to look into this a bit further 🤔

luceos:
@OwenMelbz thanks for the pull request. After some further research it seemed your changes weren't the right one. We have merged a fix and will deploy this as soon as possible as a patch.

Although your contribution wasn't merged, we are still very appreciative of the initial research and assistance in trying to resolve the issue. Thank you!

PR #3445, opened 2022-06-01
fix(subscriptions): reply notifications not working

### LINKED ISSUE #3443: Reply notifications not being sent out to users
## Bug Report

**Current Behavior**
When someone replies to a discussion you follow, you do not receive a notification.

**Steps to Reproduce**
1. Follow a discussion you have read.
2. Reply with a different user
3. No notification is sent out.

**Expected Behavior**
A notification should be sent out.

**Environment**
- Flarum version: Introduced in 1.3.0

**Possible Solution**
**Context:** If the actor has read the post number 6 which at the time was the last, and someone replies, the _subscriptions_ extension sends out notifications to those following who have last read post number 6. It doesn't send notifications to all those who have last read less than post number 7 (newest reply) to avoid repetitive notifications.

**Additional Context**
It does this by listening to the `Posted` event and using the `$event->post->discussion->last_post_number` (should be 6), it expects to run before `DiscussionMetadataUpdater` subscriber (updates `last_post_number` to 7), however in 1.3 this PR (https://github.com/flarum/framework/pull/3373) changes the behavior and leads to extension listeners to run after all core listeners. So what's happening now is that the `last_post_number` used in the subscriptions listener has already been updated to 7 in this example.

**v1.2** | **v1.3**
-- | --
Screenshot from 2022-05-31 14-23-56 | Screenshot from 2022-05-31 14-22-04

**Potential Solutions**
* We could introduce a listener priority system to make sure the listener runs before the discussion metadata updater listener.
* We could go for a `last_post_number - 1` change in the subscriptions listener. This wouldn't be a solution if we were still using `post_number_index`, but since we now count new post numbers based on previous existing post number, it might be fine to do.

SychO9:
**Fixes #3443**

As mentioned in the issue the reply notification job now receives an updated `$lastPostNumber` and because new post numbers are now guaranteed to be the max+1 (last post number before it + 1) then we can assume the job can use `$lastPostNumber - 1`.

**Necessity**

**Confirmed**

### DIFF
--- a/extensions/subscriptions/composer.json
+++ b/extensions/subscriptions/composer.json
@@ -51,7 +51,7 @@
                 "prettier": true,
                 "typescript": false,
                 "bundlewatch": false,
-                "backendTesting": false,
+                "backendTesting": true,
                 "editorConfig": true,
                 "styleci": true
             }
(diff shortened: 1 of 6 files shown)


davwheat:
Could we run into issues if two people post simultaneously and the queue processes the jobs late? I'd assume that the last post number would then refer to a later post than what we want, wouldn't it?

SychO9:
If I understand you correctly, the post number would still point to the correct value since it is provided to the job object at posting time. So no matter when the job processes it'll have the correct number because it isn't looking it up but using what it was given:

https://github.com/flarum/framework/blob/0a3f449f9e034c76ba3048a07c6cb7c2aebb509b/extensions/subscriptions/src/Job/SendReplyNotification.php#L38-L42

davwheat:
Oh yes, sorry, I was being a total idiot.

davwheat (approved):

SychO9:
Never!

PR #3446, opened 2022-06-02
feat: split frontend asset generation into separate steps for more extensibility

davwheat:

This PR splits out the logic within frontend asset generation into separate protected methods on the class to allow for more simple extending of asset generation.

For example, this is particularly useful for FoF Nightmode. Instead of completely duplicating logic found in this class, it can now override methods in order to add its own CSS compiler, and will no longer need to manually force commit assets.

It also means that future changes to core are less likely to impact extensions hooking into this class (which caused https://github.com/FriendsOfFlarum/nightmode/pull/60, for example).

You can see the impact this has in this diff: https://github.com/FriendsOfFlarum/nightmode/compare/72a75dec1ed85be8a3ec0c0442a87a6688157e84..dw/flarum-1.4-rewrite

I'm not totally familiar with this area of core, so am unsure if this will have additional unforeseen (by me) impacts. Local testing with the Nightmode branch above does not raise any issues for me.

**Necessity**

**Confirmed**

### DIFF
--- a/framework/core/src/Frontend/Content/Assets.php
+++ b/framework/core/src/Frontend/Content/Assets.php
@@ -46,7 +46,7 @@ class Assets
      * @return $this
      * @throws \Illuminate\Contracts\Container\BindingResolutionException
      */
-    public function forFrontend(string $name): static
+    public function forFrontend(string $name): Assets
     {
         $this->assets = $this->container->make('flarum.assets.'.$name);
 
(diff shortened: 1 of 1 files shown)


SychO9 (approved):

luceos:
@davwheat I cannot see your diff, is this something we need to introduce in nightmode now? I.. don't see the benefit of what was merged here or the necessary rewrite for nightmode.

davwheat:
I think the necessary nightmode changes may have already been merged.

I'm not totally sure what the intention was after this got merged anymore.

PR #3448, opened 2022-06-03
fix(tags): tag selection modal shown when no tags can be selected

### LINKED ISSUE #3444: The tag prompt shows up even for a `0 to 0` tag requirements
## Bug Report

Discovered on Nightly by working around here: https://discuss.flarum.org/d/30939-eliminate-choose-tags-for-your-discussion-screen

**Current Behavior**
If you require 0 tag (0 to 0 for both primary & secondary), the user can post without any tag but will see the tag prompt and will have to confirm without having the possibility to select anything.

**Steps to Reproduce**
1. As admin, set both tag requirements from 0 to 0 for both tag requirements
2. As a user, create a new discussion and publish it
3. You have to see and validate the tag prompt for nothing

**Expected Behavior**
If tag requirements are from 0 to 0, I should be able to publish without having to see the tag prompt.

**Screenshots**
Screenshot 2022-06-01 at 22-52-33 Flarum_nightly

**Environment**
Nightly

**Possible Solution**
I think there should be a new scenario here:

`tag requirements === 0` : no prompt at all (except for admins if tags do exist)

SychO9:
**Fixes #3444**

Checks that `maxPrimaryTags` and `maxSecondaryTags` are not nil before showing tag selection modal.

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/extensions/tags/js/src/forum/addTagComposer.js
+++ b/extensions/tags/js/src/forum/addTagComposer.js
@@ -65,10 +65,15 @@ export default function () {
     const chosenSecondaryTags = chosenTags.filter((tag) => tag.position() === null);
     const selectableTags = getSelectableTags();
 
+    const minPrimaryTags = parseInt(app.forum.attribute('minPrimaryTags'));
+    const minSecondaryTags = parseInt(app.forum.attribute('minSecondaryTags'));
+    const maxPrimaryTags = parseInt(app.forum.attribute('maxPrimaryTags'));
+    const maxSecondaryTags = parseInt(app.forum.attribute('maxSecondaryTags'));
+
     if (
-      (!chosenTags.length ||
-        chosenPrimaryTags.length < app.forum.attribute('minPrimaryTags') ||
-        chosenSecondaryTags.length < app.forum.attribute('minSecondaryTags')) &&
+      ((!chosenTags.length && maxPrimaryTags !== 0 && maxSecondaryTags !== 0) ||
+        chosenPrimaryTags.length < minPrimaryTags ||
+        chosenSecondaryTags.length < minSecondaryTags) &&
       selectableTags.length
     ) {
       app.modal.show(TagDiscussionModal, {


extensions/tags/js/src/forum/addTagComposer.js
```diff
@@ -62,9 +62,14 @@ export default function () {
...
-          || (chosenSecondaryTags.length < app.forum.attribute('minSecondaryTags'))
+    const minPrimaryTags = parseInt(app.forum.attribute('minPrimaryTags'));
+    const minSecondaryTags = parseInt(app.forum.attribute('minSecondaryTags'));
+    const maxPrimaryTags = parseInt(app.forum.attribute('maxPrimaryTags'));
+    const maxSecondaryTags = parseInt(app.forum.attribute('maxSecondaryTags'));
+
+    if (((!chosenTags.length && maxPrimaryTags !== 0 && maxSecondaryTags !== 0)
+          || (chosenPrimaryTags.length < minPrimaryTags)
+          || (chosenSecondaryTags.length < minSecondaryTags)
         ) && selectableTags.length) {
```
davwheat: Does this not work, too? I think only removing the check that no tags are selected is enough.

```suggestion
    const minPrimaryTags = parseInt(app.forum.attribute('minPrimaryTags'));
    const minSecondaryTags = parseInt(app.forum.attribute('minSecondaryTags'));

    if (selectableTags.length && (chosenPrimaryTags.length < minPrimaryTags || chosenSecondaryTags.length < minSecondaryTags)) {
```

SychO9: Sorry I forgot about this PR 😞 

Your change makes sense but it misses another scenario, the current behavior which we don't want to lose is that: If the maximum is not nil, and the user chooses no tags (with min being 0) the tag selection modal should show up because they can select something.

davwheat (approved):

PR #3246, opened 2022-01-01
refactor: Modal rewrite to use native HTML Dialog element + polyfill

davwheat:
**Fixes flarum/issue-archive#124**

**Changes proposed in this pull request:**

Replaces Bootstrap modals with HTML5 Dialog element and polyfill.

This reduces our bundle, and improves accessibility.

**Reviewers should focus on:**

Testing extension modal implementations work with these changes.

**Necessity**

- [x] Has the problem that is being solved here been clearly explained?
- [x] If applicable, have various options for solving this problem been considered?
- [x] For core PRs, does this need to be in core, or could it be in an extension?
- [x] Are we willing to maintain this for years / potentially forever?

**Confirmed**

- [x] Frontend changes: tested on a local Flarum installation.
- [x] Core developer confirmed locally this works as intended.
- [x] Tests have been added, or are not appropriate here.

PR #3456, opened 2022-06-10
feat(modals): support stacking modals, remove bootstrap modals dependency

### LINKED ISSUE #3516: [A11Y] Add missing attributes to modal
**Note:** this could be performed during a migration to `<dialog>` HTML elements.

jaspervriends:
Fixes #3516

In this PR I've worked on making it possible to have multiple dialogs opened at the same time and make them stackable. I have partly used code of #3246 and altered it. It also removed the bootstrap modal dependency.

However, I did not proceed using the native HTML Dialog element for this PR due to lack of animations (for example, the backdrop is not animatable) and responsiveness with the interface.

A third parameter has been added to the `app.modal.show` function which is a boolean and defaults to `false`. This means that all already existing modals will keep their intended behaviour.

I also copied the new 'dismissable' options from the other PR, such as an option to hide the dialog on escape key, clicking on the backdrop or using the close button and managing them seperately per modal. This means there is a deprecation too which was mentioned in the code.

- Current modals are opening and closing as intended
- Stacking multiple modals works
- Responsiveness is still good

**Necessity**

**Confirmed**

### DIFF
--- a/framework/core/js/src/common/components/ModalManager.tsx
+++ b/framework/core/js/src/common/components/ModalManager.tsx
@@ -64,7 +64,7 @@ export default class ModalManager extends Component<IModalManagerAttrs> {
             onclick={this.handlePossibleBackdropClick.bind(this)}
             ontransitionend={this.onBackdropTransitionEnd.bind(this)}
             data-showing={!!this.attrs.state.modalList.length}
-            style={{'--modal-count': this.attrs.state.modalList.length}}
+            style={{ '--modal-count': this.attrs.state.modalList.length }}
           />
         )}
       </>
--- a/framework/core/js/src/common/components/ModalManager.tsx
+++ b/framework/core/js/src/common/components/ModalManager.tsx
@@ -196,7 +196,7 @@ export default class ModalManager extends Component<IModalManagerAttrs> {
     }
   }
 
-  protected handlePossibleBackdropClick(e: MouseEvent | TouchEvent): void {
+  protected handlePossibleBackdropClick(e: MouseEvent): void {
     if (!this.attrs.state.modal || !this.attrs.state.modal.componentClass.dismissibleOptions.viaBackdropClick) return;
 
     this.animateHide();
--- a/framework/core/js/src/common/components/ModalManager.tsx
+++ b/framework/core/js/src/common/components/ModalManager.tsx
@@ -28,11 +28,14 @@ export default class ModalManager extends Component<IModalManagerAttrs> {
     return this.attrs.state.modalList.map((modal: any) => {
       const Tag = modal?.componentClass;
 
-      // Make stackable modals
-      const zIndex = 9999 + modal.key;
-
       return (
-        <div className="ModalManager modal" modal-key={modal.key} style={{ zIndex }}>
+        <div
+          className="ModalManager modal"
+          data-modal-key={modal.key}
+          role="dialog"
+          style={{ '--modal-number': modal.key }}
+          aria-hidden={this.attrs.state.modal !== modal && 'true'}
+        >
           {!!Tag && (
             <Tag
               key={modal?.key}
@@ -74,8 +77,8 @@ export default class ModalManager extends Component<IModalManagerAttrs> {
         }
 
         // Activate focus trap if there's a new dialog which is not trapped yet
-        if (this.dialogElement && this.lastSetFocusTrap !== dialogKey) {
-          this.focusTrap = createFocusTrap(this.dialogElement as HTMLElement, { allowOutsideClick: true });
+        if (this.activeDialogElement && this.lastSetFocusTrap !== dialogKey) {
+          this.focusTrap = createFocusTrap(this.activeDialogElement as HTMLElement, { allowOutsideClick: true });
 
           this.focusTrap!.activate?.();
         }
@@ -91,25 +94,25 @@ export default class ModalManager extends Component<IModalManagerAttrs> {
   /**
    * Get current active dialog
    */
-  private get dialogElement(): HTMLElement {
-    return document.body.querySelector(`div[modal-key="${this?.attrs?.state?.modal?.key}"] .Modal`) as HTMLElement;
+  private get activeDialogElement(): HTMLElement {
+    return document.body.querySelector(`div[data-modal-key="${this?.attrs?.state?.modal?.key}"] .Modal`) as HTMLElement;
   }
 
   /**
    * Get backdrop element of active dialog
    */
-  private get backdropElement(): HTMLElement {
-    return document.body.querySelector(`div[modal-key="${this?.attrs?.state?.modal?.key}"] .backdrop`) as HTMLElement;
+  private get activeBackdropElement(): HTMLElement {
+    return document.body.querySelector(`div[data-modal-key="${this?.attrs?.state?.modal?.key}"] .backdrop`) as HTMLElement;
   }
 
   animateShow(readyCallback: () => void): void {
     if (!this.attrs.state.modal) return;
 
-    this.dialogElement.addEventListener('transitionend', () => readyCallback(), { once: true });
+    this.activeDialogElement.addEventListener('transitionend', () => readyCallback(), { once: true });
 
     // Fade in
     requestAnimationFrame(() => {
-      this.dialogElement.classList.add('in');
+      this.activeDialogElement.classList.add('in');
     });
   }
 
@@ -117,7 +120,7 @@ export default class ModalManager extends Component<IModalManagerAttrs> {
     if (this.modalClosing) return;
     this.modalClosing = true;
 
-    this.backdropElement.addEventListener(
+    this.activeBackdropElement.addEventListener(
       'transitionend',
       () => {
         this.modalClosing = false;
@@ -131,8 +134,8 @@ export default class ModalManager extends Component<IModalManagerAttrs> {
       { once: true }
     );
 
-    this.dialogElement.classList.remove('in');
-    this.dialogElement.classList.add('out');
+    this.activeDialogElement.classList.remove('in');
+    this.activeDialogElement.classList.add('out');
   }
 
   protected handleEscPress(e: KeyboardEvent): any {
--- a/framework/core/js/src/common/components/ModalManager.tsx
+++ b/framework/core/js/src/common/components/ModalManager.tsx
@@ -71,7 +71,15 @@ export default class ModalManager extends Component<IModalManagerAttrs> {
 
     requestAnimationFrame(() => {
       try {
-        if (!this.attrs.state.isModalOpen()) return;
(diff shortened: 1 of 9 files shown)


framework/core/js/src/common/components/ModalManager.tsx
```diff
@@ -10,90 +10,154 @@ interface IModalManagerAttrs {
...
-            state={this.attrs.state}
-          />
-        )}
-      </div>
-    );
+    return this.attrs.state.modalList.map((modal: any) => {
+      const Tag = modal?.componentClass;
+
+      // Make stackable modals
+      const zIndex = 9999 + modal.key;
```
davwheat: Is this needed? Elements with the same zIndex should stack as per normal DOM hierarchy rules, where the last element is on top.

If we are going to do it this way, we should use the CSS variable for modal z-index and either:

1. Provide an custom attribute, such as `--modal-number`, then use `calc()` to set the z-index

2. Fetch the value of the modal z-index attribute on the `::root` selector and use that instead of a hardcoded value.

I expect number 1 will be easier.

framework/core/js/src/common/components/ModalManager.tsx
```diff
@@ -10,90 +10,154 @@ interface IModalManagerAttrs {
...
+
+      return (
+        <div className="ModalManager modal" modal-key={modal.key} style={{ zIndex }}>
```
davwheat: I would prefer to use `data-modal-key`.

framework/core/js/src/common/components/ModalManager.tsx
```diff
@@ -10,90 +10,154 @@ interface IModalManagerAttrs {
...
       } catch {
         // We can expect errors to occur here due to the nature of mithril rendering
       }
     });
   }
 
+  /**
+   * Get current active dialog
+   */
+  private get dialogElement(): HTMLElement {
```
davwheat: Could we rename this to more suitable match the doc block?


```suggestion
  private get activeDialogElement(): HTMLElement {
```

framework/core/js/src/common/components/ModalManager.tsx
```diff
@@ -10,90 +10,154 @@ interface IModalManagerAttrs {
...
+    return document.body.querySelector(`div[modal-key="${this?.attrs?.state?.modal?.key}"] .Modal`) as HTMLElement;
+  }
+
+  /**
+   * Get backdrop element of active dialog
+   */
+  private get backdropElement(): HTMLElement {
```
davwheat:
```suggestion
  private get activeBackdropElement(): HTMLElement {
```

framework/core/js/src/common/states/ModalManagerState.ts
```diff
@@ -19,44 +25,57 @@ export default class ModalManagerState {
...
    *
    * @example <caption>Show a modal from a lifecycle method (`oncreate`, `view`, etc.)</caption>
    * // This "hack" is needed due to quirks with nested redraws in Mithril.
    * setTimeout(() => app.modal.show(MyCoolModal, { attr: 'value' }), 0);
+   *
+   * @example <caption>Stacking modals</caption>
+   * app.modal.show(MyCoolStackedModal, { attr: 'value' }, true);
    */
-  show(componentClass: UnsafeModalClass, attrs: Record<string, unknown> = {}): void {
+  show(componentClass: UnsafeModalClass, attrs: Record<string, unknown> = {}, stackModal: Boolean = false): void {
```
davwheat: We should use `boolean` primitive rather than the `Boolean` object.

```suggestion
  show(componentClass: UnsafeModalClass, attrs: Record<string, unknown> = {}, stackModal: boolean = false): void {
```

framework/core/js/src/common/components/ModalManager.tsx
davwheat: We should also add `aria-hidden="true"` to all models except the active one.

Also `role="dialog"`.

davwheat (commented):
Just some comments so far. Haven't had time to properly review yet.

jaspervriends:
Thanks for the feedback, will work on it later this week

jaspervriends:
Processed feedback :)

davwheat (dismissed):
I've addressed some things with commits on this PR, but would like a sanity check for my changes from another core dev before looking to merge this.

Thank you for your work on this! This is a big improvement for Flarum's modals system! :)

framework/core/js/src/forum/components/LogInModal.tsx
```diff
@@ -146,7 +146,7 @@ export default class LogInModal<CustomAttrs extends ILoginModalAttrs = ILoginMod
     const email = this.identification();
     const attrs = email.includes('@') ? { email } : undefined;
 
-    app.modal.show(ForgotPasswordModal, attrs);
+    app.modal.show(ForgotPasswordModal, attrs, true);
```
SychO9: I don't believe this should be stackable 🤔

framework/core/less/common/Modal.less
```diff
@@ -6,55 +6,85 @@
   overflow: hidden;
```
SychO9: `.modal-open` definition is dead code now I think

framework/core/js/src/common/states/ModalManagerState.ts
```diff
@@ -19,62 +26,101 @@ export default class ModalManagerState {
...
+   * compatibility reasons, until Flarum 2.0.
+   *
+   * @example <caption>Show a modal</caption>
+   * app.modal.show(MyCoolModal, { attr: 'value' });
+   *
+   * @example <caption>Show a modal from a lifecycle method (`oncreate`, `view`, etc.)</caption>
+   * // This "hack" is needed due to quirks with nested redraws in Mithril.
+   * setTimeout(() => app.modal.show(MyCoolModal, { attr: 'value' }), 0);
+   */
+  show(componentClass: UnsafeModalClass, attrs?: Record<string, unknown>): void;
```
SychO9: in 2.0 we'll probably keep the new `stackModal` argument set to `false` by default (or `true`). So I think we could dro this signature and the deprecation mark. Thoughts?

SychO9 (dismissed):
Code looks good! a few comments and a bug I noticed. Thanks a lot for the PR!

There's an issue with the scrolling behavior, I think a video will better explain it:
https://user-images.githubusercontent.com/20267363/178266180-69aaa829-b528-4808-999f-4693bd15b3a5.mp4

framework/core/js/src/forum/components/LogInModal.tsx
davwheat: Sorry, was using this for testing! 🙈

davwheat:
oops... that was the wrong git command...

have asked jasper for write access to his fork so i can resurrect this. have a backup of the code at https://github.com/flarum/framework/tree/jv/stackable-modals

davwheat:
RE the scrolling bug: isn't this how the modals currently work in core?

https://user-images.githubusercontent.com/7406822/178990288-fede20f7-ccbc-4bf0-a72f-979446491969.mp4

SychO9:
yea but that isn't how they work in this branch 🙈 `0:15` of my uploaded video shows behavior in this branch

davwheat:
OH I didn't watch for long enough! Right, I'll get on that.

davwheat:
@SychO9 Fixed, I hope.

SychO9:
🙈 now you can't scroll inside a modal

framework/core/less/common/Modal.less
```diff
@@ -45,41 +48,7 @@
...
-      transition: none;
-    }
-
-    + .ModalManager[data-visibility-state="exiting"] {
-      .backdrop {
-        transition: none;
-      }
-    }
-  }
+  pointer-events: none;
```
SychO9: so the scrollbar shows up correctly but with `pointer-events: none` I can't scroll. But removing this also means you can't click outside to close 🤔

framework/core/js/src/common/components/ModalManager.tsx
```diff
@@ -26,47 +26,62 @@ export default class ModalManager extends Component<IModalManagerAttrs> {
...
+            </div>
+          );
+        })}
+
+        {this.attrs.state.backdropShown && (
+          <div
+            class="Modal-backdrop backdrop"
+            onclick={this.handlePossibleBackdropClick.bind(this)}
+            ontransitionend={this.onBackdropTransitionEnd.bind(this)}
+            data-showing={!!this.attrs.state.modalList.length}
```
SychO9, line 70: Ideally, the backdrop would show behind the latest open modal and above the others. I think a simple CSS trick would do it: (other comment below for follow up)

```suggestion
            data-showing={!!this.attrs.state.modalList.length}
            style={{'--modal-count': this.attrs.state.modalList.length}}
```

framework/core/less/common/Modal.less
```diff
@@ -14,24 +14,27 @@
...
+  background: var(--overlay-bg);
+  position: fixed;
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  opacity: 0;
+  transition: opacity 0.2s ease-out;
+  z-index: 1;
+  z-index: ~"calc(var(--zindex-modal) - 1)";
```
SychO9:
```suggestion
  z-index: ~"calc(var(--zindex-modal) + var(--modal-count)) - 2";
```

SychO9 (commented):
Using one backdrop seems great! I think only small tweaks are left.

luceos:
The patience in this PR is admirable.

davwheat:
PS: JS checks fail in some cases due to attempting to commit and push dist JS to Jasper's repo since the PR is based on the `main` branch.

SychO9:
Under known issues: https://caniuse.com/pointer-events
> Moving the scrollbar on an object with pointer-events: none; works in Firefox, but doesn't work in either Chrome or IE.

What I did is introduce invisible backdrops under each modal so that we don't have to use `pointer-events` and be able to easily exit the modal while maintaining scrollabilty (*is that a word?*).

davwheat:
Ahhhh that's a great find!

I didn't think of that idea and don't think I ever would have!

framework/core/js/src/common/components/ModalManager.tsx
```diff
@@ -192,7 +196,7 @@ export default class ModalManager extends Component<IModalManagerAttrs> {
     }
   }
 
-  protected handlePossibleBackdropClick(e: MouseEvent): void {
+  protected handlePossibleBackdropClick(e: MouseEvent | TouchEvent): void {
```
SychO9:
```suggestion
  protected handlePossibleBackdropClick(e: MouseEvent): void {
```

davwheat (approved):

davwheat:
Rebased on `main`.

SychO9 (approved):

davwheat:
phew that was a big one 😄

PR #3458, opened 2022-06-11
Deploy dependabot

flawedworld:
This will configure dependabot to update all dependencies in all folders in the repo. Can see it in action on my fork of the repo. It will be quite spammy initially, due to having multiple projects in a monorepo (It's limited to 10 pulls per location), but once it's initial pulls are merged, it should be relatively quiet. It's generally a good idea to update all dependencies irrespective of if they have security content as the former isn't necessarily made public or obvious and it's good codebase hygiene.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,273 @@
+version: 2
+updates:
+  # CI
+  - package-ecosystem: github-actions
+    directory: "/"
+    schedule:
+      interval: daily
+    target-branch: main
+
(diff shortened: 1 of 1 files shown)


SychO9:
We actually disabled it not long ago because of the amount of PRs was too crazy as this is a monorepo with multiple packages. However we will probably switch it back on soon once we do some housekeeping on the monorepo: https://github.com/flarum/framework/pull/3464

Thanks for the PR though!

PR #3463, opened 2022-06-13
Move workflows to monorepo

flawedworld:
https://github.com/flarum/framework/issues/3462

### DIFF
--- a/.github/workflows/REUSABLE_backend.yml
+++ b/.github/workflows/REUSABLE_backend.yml
@@ -1,0 +1,112 @@
+name: Flarum Backend Jobs
+
+on:
+  workflow_call:
+    inputs:
+      enable_backend_testing:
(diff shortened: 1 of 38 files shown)


luceos:
It seems the generic workflow causes another workflow to be added. It might be necessary to move it one level up or one level down to be ignored. Perhaps GitHub Actions even has a flag to ignore a file.

flawedworld:
>It seems the generic workflow causes another workflow to be added.

I don't see where that is happening.

SychO9:
Not sure why the tests are failing here :thinking: but I've tried looking into it in a different branch https://github.com/flarum/framework/pull/3490 and they seem to work fine there, I'm gonna close this in favor of that one, it only moves the workflows as a start, PHP 8.2 will be done in a separate PR where we can look into potential failures. Thanks for the PR though!

PR #3464, opened 2022-06-13
chore: with yarn workspaces we only need a single `yarn.lock`

SychO9:
Struggled with this while trying to audit fix locally, we don't actually need the separate lock files with yarn workspaces, only the root one is used making the rest unnecessary. This will allow us to re-enable dependabot which we disabled due to the high amount of PRs it posted.

**Necessity**

**Confirmed**


davwheat (approved):

SychO9:
ah shame, seems like the default value on `inputs.cache_dependency_path` doesn't actually get evaluated properly since at that point `inputs.js_package_manager` is unset. I'll re-introduce the env variable to do the evaluation there.

davwheat:
That's interesting behaviour, but makes sense.

Sounds good!

PR #3479, opened 2022-06-17
Use arrow functions

MathiasReker:
Anonymous functions with one-liner return statement must use arrow functions.

**Fixes #0000**

**Screenshot**

**Necessity**

**Confirmed**

**Required changes:**


davwheat:
I don't think this is supported in PHP 7.3, which Flarum supports.

MathiasReker:
You are correct, I overlooked this. Thanks

PR #3493, opened 2022-06-19
fix: post mention notification errors with _no content_ subjects

### LINKED ISSUE #3492: Mentioning an event post breaks the notification panel
## Bug Report

**Current Behavior**

If a mention refers to an event post instead of a comment (probably by mistake), the generated notification breaks the notifications panel.

**Steps to Reproduce**
1. Generate an event post (e.g. rename a discussion)
2. Login with another user and mention that post by using its ID (e.g. `@"matteocontrini"#p66032` where `66032` is the ID of the post corresponding to the rename event)
3. Login with the mentioned user and observe that an exception is thrown when trying to render the notification

**Expected Behavior**
It shouldn't break for sure, but it probably shouldn't generate a notification in general.

**Screenshots**

image

image

**Environment**
- Flarum version: 1.2.1

SychO9:
**Fixes #3492**

Just checks before using the `contentPlain` result.

**Necessity**

**Confirmed**

### DIFF
--- a/extensions/mentions/js/src/forum/components/PostMentionedNotification.js
+++ b/extensions/mentions/js/src/forum/components/PostMentionedNotification.js
@@ -23,6 +23,8 @@ export default class PostMentionedNotification extends Notification {
   }
 
   excerpt() {
-    return truncate(this.attrs.notification.subject().contentPlain(), 200);
+    const contentPlain = this.attrs.notification.subject().contentPlain();
+
+    return contentPlain && truncate(contentPlain, 200);
   }
 }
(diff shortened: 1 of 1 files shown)


extensions/mentions/js/src/forum/components/PostMentionedNotification.js
```diff
@@ -23,6 +23,8 @@ export default class PostMentionedNotification extends Notification {
   }
 
   excerpt() {
-    return truncate(this.attrs.notification.subject().contentPlain(), 200);
+    const contentPlain = this.attrs.notification.subject().contentPlain();
+
+    return contentPlain && truncate(contentPlain, 200);
```
davwheat: Would this be simpler, do you think? I don't want introduce any issues, like returning false, undefined, etc, and code not accounting for this.

```suggestion
    return truncate(this.attrs.notification.subject().contentPlain() || '', 200);
```

SychO9: Yup good thinking!

davwheat (approved):

PR #3495, opened 2022-06-20
feat: add tag selector custom setting component for `buildSettingComponent`

davwheat:

Creates a new `TagSelector` component, designed for use within `buildSettingComponent` in conjuction with core PR #3494.

This allows extensions to more easily select tag model instances within their settings pages.

Checks may fail until #3494 is merged.

**Screenshot**

https://user-images.githubusercontent.com/7406822/174515268-493ee6e2-19ee-4a27-a7a6-7ce54ab8b4cb.mp4

**Necessity**

**Confirmed**

**Required changes:**

- [x] Related core PR: #3494
- [ ] Related core PR: #3456

### DIFF
--- a/extensions/tags/js/src/admin/addCustomSettingComponent.tsx
+++ b/extensions/tags/js/src/admin/addCustomSettingComponent.tsx
@@ -1,0 +1,13 @@
+import AdminPage, { CommonSettingsItemOptions } from 'flarum/admin/components/AdminPage';
+import { extend } from 'flarum/common/extend';
+import TagSelectorComponent, { ITagSelectorAttrsSpecial } from './components/TagSelector';
+
+export default function addCustomSettingComponent() {
+  extend(AdminPage.prototype, 'customSettingComponents', function (items) {
+    items.add('flarum-tags.tag-selector', (attrs: ITagSelectorAttrsSpecial & CommonSettingsItemOptions) => {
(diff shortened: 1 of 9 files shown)


davwheat:
Moving Prettier to another PR

luceos:
I like the idea of such a component very much. But I'm afraid that the design is too different from what we've used before. There's only a limited number of tags visible in the dropdown, they're structured differently and there might be quite a lot to scroll through. I'd prefer to see the frontend component re-used so that it would accept parameters and can be used in the admin and extensions.

Other maintainers please advice.

SychO9:
I love the idea, though I agree with Daniël, I think we can reuse the TagSelectorModal component, although there might be some significant work needed to make it usable in a different context than what it was built for, but that would be perfect IMHO.

davwheat:
The main issue I had is the fact that it's within a modal. This significantly limits use cases as noone else would be able to use the component within a modal due to current one-modal limitations in core.

We can definitely explore alternate designs -- maybe something like a textbox with autocomplete to more closely mirror the forum modal?

SychO9:
Good point, maybe we can consider it after https://github.com/flarum/framework/pull/3456? it would be great if we can avoid reinventing the wheel.

SychO9:
Superseded by https://github.com/flarum/framework/pull/3686

PR #3503, opened 2022-07-06
feat(subscriptions): add option to send notifications when not caught up

davwheat:

This PR adds a new configurable option to Flarum Subscriptions to allow the sending of notifications even when a user is not caught up to a discussion yet. This could be seen as a easier migration step from one forum software to another, where users are more used to receiving emails for **all** new posts, rather than just the first new one.

Due to the notification subject for subscriptions being the discussion itself rather than the post, notifications in the frontend are still grouped into one single notification, so this change really only affects emails. Changing the subject would be a breaking change, though, as would introducing a new notification which replaces the old one.

This should retain existing behaviour by default, using the memory-based settings defaults.

**Necessity**

**Confirmed**


extensions/subscriptions/src/Job/SendReplyNotification.php
```diff
@@ -43,14 +44,22 @@ public function __construct(Post $post, $lastPostNumber)
...
 
-        $notify = $discussion->readers()
+        $query = $discussion->readers()
             ->where('users.id', '!=', $post->user_id)
-            ->where('discussion_user.subscription', 'follow')
-            ->where('discussion_user.last_read_post_number', $this->lastPostNumber - 1)
-            ->get();
+            ->where('discussion_user.subscription', 'follow');
+
+        if ($settings->get('flarum-subscriptions.dont_notify_unless_caught_up', true)) {
```
SychO9: isn't the fallback here unnecessary since we specify the default using the extender?

SychO9 (commented):
Looks good, two thoughts though:

- I think when it comes to notification behavior it might be better to introduce this as a user preference instead of a global admin setting.
- `dont_notify_unless_caught_up` (and the locale value) is hard to understand unless you read further or look at the code. Could we change it to `notify_first_new_unread_post_only` (and the locale text) ?

extensions/subscriptions/src/Job/SendReplyNotification.php
davwheat: True -- I just added it from habit!

```suggestion
        if ($settings->get('flarum-subscriptions.dont_notify_unless_caught_up')) {
```

davwheat:
This code is rather... interesting.

I'm not sure how to go about doing validation for user preferences -- if I could be pointed in the right direction, I'd appreciate that.

I also think my DB queries might not be very performant for large communities. Will seek feedback on that from some others. If that's the case, it might be wise adding a new column on the users table to store this preference, thus allowing all the new logic in the PHP to be moved to part of the DB query instead.

The code does all work, though, it's just not that great. 😅

davwheat:
I think the best option is to either move the criteria option to a column on the user table, or a new table with a FK. A new col on the users table would not be a very performant migration for very large communities, though.

luceos:
I've looked at this PR again to see whether we should "just do it", but when I'm looking at the logic to match preference to see whom to send notifications to I can only cringe. This logic seems overly convoluted and confusing. I know this isn't really a great PR review comment, sorry about that. We are all doing our best to find solutions to things that are (much) requested, but I'm unsure that this is the right approach.

This request almost warrants a completely different implementation for handling subscriptions and notifications..

extensions/subscriptions/src/Job/SendReplyNotification.php
```diff
@@ -43,18 +46,54 @@ public function __construct(Post $post, $lastPostNumber)
 
     public function handle(NotificationSyncer $notifications)
     {
+        /**
+         * @var SettingsRepositoryInterface
+         */
+        $settings = resolve(SettingsRepositoryInterface::class);
```
SychO9:
```suggestion
    public function handle(NotificationSyncer $notifications, SettingsRepositoryInterface $settings)
    {
```

extensions/subscriptions/js/src/forum/addUserPreferences.tsx
```diff
@@ -0,0 +1,50 @@
+import app from 'flarum/forum/app';
+import { extend } from 'flarum/common/extend';
+
+import SettingsPage from 'flarum/forum/components/SettingsPage';
+import FieldSet from 'flarum/common/components/FieldSet';
+import Select from 'flarum/common/components/Select';
+
+import type Mithril from 'mithril';
+
+const keys = ['all_new', 'first_new'];
```
SychO9: I think a boolean field would be easier, `Notify about every new post instead of only the last in a discussion`.

wdyt?

SychO9 (commented):
> This logic seems overly convoluted and confusing.

Our preferences system really needs a bit of a refactor. Storing them in a column is not flexible.

I think the complication lies in the: preference vs admin default vs admin enforcement. I don't think admin enforcement here is really necessary. I think we should also drop the admin default for a better _admin default user preferences_ system in the future which would integrate seamlessly with reading user preferences.

That would probably cut down on the complexity.

extensions/subscriptions/js/src/forum/addUserPreferences.tsx
davwheat: I would agree that this makes more sense for this implementation.

I opted to use a select drop down in order to potentially allow for future extensibility by extensions to add their own notification options if they wished. Having said that, though, the way I wrote this doesn't actually allow for that anyway! 😅

SychO9: Yea, the hardest thing about core is always considering extensibility, but considering the slight complexity in this feature, keeping it simple will be beneficial to us 👍🏼.

extensions/subscriptions/locale/en.yml
```diff
@@ -49,7 +52,6 @@ flarum-subscriptions:
 
   # Translations in this namespace are used in emails sent by the forum.
   email:
-
     # These translations are used in emails sent when a post is made in a subscribed discussion
     new_post:
```
davwheat, line 52: We need to find a way to switch out the bottom portion of email translations depending on user prefs.

I couldn't find any way to get the user within the notif blueprint, so I'm not sure how I'd change this without having two separate blueprints. @imorland did this in a Blomstra extension and said he can bring over the changes a couple of weeks ago but I forgot to remind him. 🙈

extensions/subscriptions/locale/en.yml
```diff
@@ -30,6 +28,11 @@ flarum-subscriptions:
...
       new_post_text: "{username} posted"
 
+    # These translations are used on the user preferences page
+    user_preferences:
+      heading: Discussion subscriptions
+      notify_for_all_posts_label: Notify about every new post instead of only the last in a discussion
+
     # These translations are used in the Settings page.
     settings:
       follow_after_reply_label: Automatically follow discussions that I reply to
```
SychO9: for that page I believe we put translations under `settings` as that's the settings page. (*I think, I don't remember if user preferences had their own separate namespace*)

extensions/subscriptions/locale/en.yml
SychO9: what are you referring to I'm not following 🤔

SychO9 (commented):
Looks good, just a couple of comments, and I took the liberty to push integration tests.

davwheat:
> what are you referring to I'm not following 🤔 

At the bottom of the email translation, it says that the user won't be notified for new.replies until the thread has been read. That needs to be removed based on user prefs.

extensions/subscriptions/src/Job/SendReplyNotification.php
```diff
@@ -41,20 +45,45 @@ public function __construct(Post $post, $lastPostNumber)
...
+
+        $followers = $discussion->readers()
+            ->select('users.id', 'users.preferences', 'discussion_user.last_read_post_number')
             ->where('users.id', '!=', $post->user_id)
-            ->where('discussion_user.subscription', 'follow')
-            ->where('discussion_user.last_read_post_number', $this->lastPostNumber - 1)
-            ->get();
+            ->where('discussion_user.subscription', 'follow');
+
+        $followers->chunk(150, function (Collection $followers) use (&$usersToNotify) {
```
SychO9, line 60: I made a minor change here (diff shows it as more than it is lol) where we [chunk](https://laravel.com/docs/9.x/queries#chunking-results) results in case of large communities. @davwheat does that seem sensible?

davwheat: Definitely.

extensions/subscriptions/locale/en.yml
davwheat: Makes sense tbh. We use the same for "Automatically follow discussions that I reply to" :D

SychO9:
> > what are you referring to I'm not following thinking
> 
> At the bottom of the email translation, it says that the user won't be notified for new.replies until the thread has been read. That needs to be removed based on user prefs.

oh, yea I see it now, I don't think we can do a per preference phrase, the blueprint specifies the same email body to all users, it isn't aware of individual ones, so we can only try to tweak the wording of the email maybe so that it hints to preferences?

extensions/subscriptions/js/src/forum/addUserPreferences.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+import SettingsPage from 'flarum/forum/components/SettingsPage';
+import FieldSet from 'flarum/common/components/FieldSet';
+import Switch from 'flarum/common/components/Switch';
+
+export default function addUserPreferences() {
+  extend(SettingsPage.prototype, 'settingsItems', function (items) {
+    items.add(
+      'fof-subscriptions',
+      <FieldSet label={app.translator.trans('flarum-subscriptions.forum.user_preferences.heading')} className="Settings-subscriptions">
```
davwheat:
```suggestion
      <FieldSet label={app.translator.trans('flarum-subscriptions.forum.settings.preferences_heading')} className="Settings-subscriptions">
```

extensions/subscriptions/js/src/forum/addUserPreferences.tsx
```diff
@@ -0,0 +1,25 @@
...
+        <Switch
+          id="flarum_subscriptions__notify_for_all_posts"
+          state={!!this.user!.preferences()?.['flarum-subscriptions.notify_for_all_posts']}
+          onchange={(val: boolean) => {
+            this.user!.savePreferences({ 'flarum-subscriptions.notify_for_all_posts': val });
+          }}
+        >
+          {app.translator.trans('flarum-subscriptions.forum.user_preferences.notify_for_all_posts_label')}
```
davwheat:
```suggestion
          {app.translator.trans('flarum-subscriptions.forum.settings.notify_for_all_posts_label')}
```

extensions/subscriptions/locale/en.yml
davwheat:
```suggestion
    # These translations are used in the Settings page.
    settings:
      follow_after_reply_label: Automatically follow discussions that I reply to
      notify_for_all_posts_label: Notify about every new post instead of only the last in a discussion
      preferences_heading: Discussion subscriptions
```

SychO9 (changes requested):
One last thing, the UI needs tobe consistent, the option should go under `Automatically follow discussions that I reply to`, we can do that in `addSubscriptionSettings()` instead without adding a new section in user settings.

extensions/subscriptions/js/src/forum/addUserPreferences.tsx
```diff
@@ -0,0 +1,25 @@
+import app from 'flarum/forum/app';
```
SychO9: you forgot to remove this file 👼🏼 and the call to the function

davwheat: whatttttt i would neverrr 🙈🙈🙈🙈

(do not code right before your flight is about to start boarding)

davwheat:
lets try that again...

extensions/subscriptions/locale/en.yml
```diff
@@ -33,6 +31,8 @@ flarum-subscriptions:
...
+      notify_for_all_posts_label: Notify about every new post instead of only the last in a discussion
+      preferences_heading: Discussion subscriptions
```
SychO9:
```suggestion
```

extensions/subscriptions/js/src/forum/addSubscriptionSettings.tsx
```diff
@@ -23,5 +24,20 @@ export default function () {
...
+          id="flarum_subscriptions__notify_for_all_posts"
+          state={!!this.user!.preferences()?.['flarum-subscriptions.notify_for_all_posts']}
+          onchange={(val: boolean) => {
+            this.user!.savePreferences({ 'flarum-subscriptions.notify_for_all_posts': val });
+          }}
+        >
+          {app.translator.trans('flarum-subscriptions.forum.settings.notify_for_all_posts_label')}
+        </Switch>
+      </FieldSet>
+    );
```
SychO9: I'd remove the heading, it doesn't really match well

```suggestion

    items.add(
      'notifyForAllPosts',
      <Switch
        id="flarum_subscriptions__notify_for_all_posts"
        state={!!this.user!.preferences()?.['flarum-subscriptions.notify_for_all_posts']}
        onchange={(val: boolean) => {
          this.user!.savePreferences({ 'flarum-subscriptions.notify_for_all_posts': val });
        }}
      >
        {app.translator.trans('flarum-subscriptions.forum.settings.notify_for_all_posts_label')}
      </Switch>
    );
```

Screenshot from 2022-08-30 19-21-37

extensions/subscriptions/js/src/forum/addSubscriptionSettings.tsx
```diff
@@ -2,9 +2,10 @@ import app from 'flarum/forum/app';
 import { extend } from 'flarum/common/extend';
 import SettingsPage from 'flarum/forum/components/SettingsPage';
 import Switch from 'flarum/common/components/Switch';
+import FieldSet from 'flarum/common/components/FieldSet';
```
SychO9:
```suggestion
```

SychO9 (changes requested):

SychO9 (approved):

PR #3515, opened 2022-07-10
refactor: convert core modals to TypeScript

SychO9:
**Part of #3533**

Converts all JS modals to TypeScript.

**Necessity**

**Confirmed**


davwheat (approved):

framework/core/js/src/forum/components/RenameDiscussionModal.tsx
```diff
@@ -70,8 +81,9 @@ export default class RenameDiscussionModal extends Modal {
           this.loading = false;
           m.redraw();
         });
-    } else {
-      this.hide();
     }
+
+    this.hide();
+    return;
```
davwheat: Do we need this?

SychO9: typescript complains about return type difference in the conditional, I made it explicit instead 🙈 https://github.com/flarum/framework/pull/3515/commits/8b2ed266c94c7727f990500c7b7a52f558249210

PR #3526, opened 2022-07-12
chore: use github issue template forms

SychO9:
Uses GitHub issue forms so that it's easier and more clear.

*This is just a sample from GitHub docs*
GitHub sample issue form

**Necessity**

**Confirmed**


davwheat (approved):
👑

PR #3530, opened 2022-07-14
feat: extender to add custom less variables

davwheat:

Adds new theming extender which allows for the definition of custom Less variables. This is somewhat similar to the current extender which allows the adding of functions.

This can help with things such as determining if styling should be applied at asset-compile-time rather than in runtime with Javascript. This will help reduce extra styles being present in stylesheets, and extra JS being present in our dist scripts.

Variables and values defined here will be added immediately after core's own settings keys `@config-...` directly into the Less source files.

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/framework/core/src/Extend/Theme.php
+++ b/framework/core/src/Extend/Theme.php
@@ -104,11 +104,8 @@ class Theme implements ExtenderInterface
     /**
      * Defines a new Less variable to be accessible in all Less files.
      *
-     * This can be useful for styling based on a backend setting, for example.
-     *
-     * Remember that Less is not recompiled when settings values change, so you
-     * may need to hook into a Saved event to recompile stylesheets when the
-     * value of one of your extension's settings changes in the database.
+     * If you want to expose a setting from your database to Less, you should use
+     * the `registerLessConfigVar` extender from `Extend\Settings` instead.
      *
      * Please note the value returned from the callable will be inserted directly
      * into the Less source. If it is unsafe in some way (e.g., contains a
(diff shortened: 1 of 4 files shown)


SychO9:
What's a use case for this? we have [`registerLessConfigVar`](https://github.com/flarum/framework/blob/main/framework/core/src/Extend/Settings.php#L80) in the `Settings` extender and I'm wondering if that's enough for your use case considering the docblock.

davwheat:
That would be enough for most use cases, but for this the idea was to define a full URI to an extension asset, which can't be done purely in Less alone since it can't account for subdirectory installs.

framework/core/src/Extend/Theme.php
```diff
@@ -100,12 +101,53 @@ public function addCustomLessFunction(string $functionName, callable $callable):
...
+     *     $settings = resolve(SettingsRepositoryInterface::class);
+     *     return boolval($settings->get('my-extension.show_thing', false));
+     *   })
+     * ```
+     *
+     * @param string $variableName Name of the variable identifier.
+     * @param callable $value The PHP function to run, which returns the value for the variable.
+     *
+     * @return self
+     */
```
SychO9: can we explicitly mention to use the `registerLessConfigVar` in the case of exposing a setting? probably also use a different example to avoid confusion

SychO9 (commented):
makes sense. then let's just tweak the docblock to not confuse the two extenders.

SychO9 (approved):

PR #3531, opened 2022-07-14
perf(statistics): rewrite for performance on very large communities

davwheat:

The Statistics extension is currently a big issue on large communities, due to the way it's written.

Currently, all statistics data is loaded on the backend during page load, and bundled with the page's payload itself. On a community with about 1 million discussions, this case mean the admin dashboard takes over 10s to load. In addition, statistics data is loaded no matter which dashboard page is navigated to, causing unnecessary database queries and server load.

Here's our proposal:
- [x] Move statistics data out of initial payload, and into an XHR request
- [x] Split XHR request into lifetime and timed
- [x] Add new periods: previous 7 and 28 days
- [x] If needed, move widget to the extension page instead of dashboard
  - [ ] ~~prevents an XHR request on every admin dash visit~~
  - [ ] ~~could also instead have a button labelled "show statistics" which loads it on the main dash~~
  - [x] if we move, should add message stating as such so admins aren't confused, with link to ext page
  - [x] add partial stats widget to dashboard, with link to original
- [x] Cache stats server-side to prevent recalculation with multiple hits in succession

There are probably some other things we can do to also improve performance, but I feel this is a good starting point.

**Necessity**

**Confirmed**

**Required changes:**


extensions/statistics/less/admin.less
```diff
@@ -84,11 +119,12 @@
   position: relative; /* for absolutely positioned tooltip */
 
   /* https://www.smashingmagazine.com/2015/11/using-system-ui-fonts-practical-guide/ */
-  font-family: -apple-system, BlinkMacSystemFont,
-  'Segoe UI', 'Roboto', 'Oxygen', 'Ubuntu', 'Cantarell',
-  'Fira Sans', 'Droid Sans', 'Helvetica Neue', sans-serif;
+  font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", "Roboto", "Oxygen",
+    "Ubuntu", "Cantarell", "Fira Sans", "Droid Sans", "Helvetica Neue",
+    sans-serif;
```
luceos: I know this is nitpicking, but what happened here? Quotes? Shouldn't really be in this PR I think?

davwheat: Oh, that was a side-effect of IDE code formatting. I can revert it.

davwheat:
```suggestion
  font-family: -apple-system, BlinkMacSystemFont,
  'Segoe UI', 'Roboto', 'Oxygen', 'Ubuntu', 'Cantarell',
  'Fira Sans', 'Droid Sans', 'Helvetica Neue', sans-serif;
```

luceos (changes requested):
I don't understand the need to use a setting for the timezone. And instead of the method returning the actual actor timezone, it returns the setting or the server timezone. I think only the frontend uses timezones right now, everything on the backend should always be uct. Can you clarify?

davwheat:
In this PR, my main objective was improving performance.

The use of a setting, and returning the server timezone is just the original extension behaviour, which I didn't want to tamper with to reduce the chance of introducing bugs along with this PR.

I think it's easier for the PR to be reviewed based on the changes made compared to the current content and behaviour, and then continue to rewrite the behaviour in a second PR. I don't want this becoming one big huge PR which basically scraps the extension entirely and starts again.

That being said, I _believe_ the time zone things exist to account for the server not running on UTC, but I've not checked this.

luceos (approved):

extensions/statistics/locale/en.yml
```diff
@@ -9,12 +9,16 @@ flarum-statistics:
 
     # These translations are used in the Statistics dashboard widget.
     statistics:
-      active_users_text: "{count} active"
```
SychO9, line 12: This is dead code I guess? can't see it used anywhere

extensions/statistics/src/Api/Controller/ShowStatisticsData.php
```diff
@@ -0,0 +1,169 @@
...
+            return $cachedVal;
+        }
+
+        $val = array_map(function ($entity) {
+            return $entity[0]->count();
+        }, $this->entities);
+
+        $this->cache->put('flarum-subscriptions.lifetime_stats', $val, static::$lifetimeStatsCacheTtl);
+
+        return $val;
```
SychO9: can we inject an `\Illuminate\Contracts\Cache\Repository::class` for the cache instead and do this:

```suggestion
        return $this->cache->remember('flarum-subscriptions.lifetime_stats', self::$lifetimeStatsCacheTtl, function () {
            return array_map(function ($entity) {
                return $entity[0]->count();
            }, $this->entities);
        });
```

extensions/statistics/src/Api/Controller/ShowStatisticsData.php
```diff
@@ -0,0 +1,169 @@
...
+            return $cachedVal;
+        }
+
+        $val = array_map(function ($entity) {
+            return $this->getTimedCounts($entity[0], $entity[1]);
+        }, $this->entities);
+
+        $this->cache->put('flarum-subscriptions.timed_stats', $val, static::$timedStatsCacheTtl);
+
+        return $val;
```
SychO9: same here

SychO9 (changes requested):
* For the indexes, do we need both of them, or does only one of them make a difference?
* If we are caching the data I don't think we need to move out the full stats? I'm not entirely sure UX wise if it's less appealing.

extensions/statistics/locale/en.yml
davwheat: Yeah, I couldn't either

davwheat:
> For the indexes, do we need both of them, or does only one of them make a difference?

Yes, we need both. `type` on its own for the total count, and the composite for the time-based counting. Without both, each can reach up to about 5-6 seconds for 1M models.

davwheat:
For the caching, the reason I still opted to move the full set of data out was the sheer amount of load the statistics gathering could put on the SQL database. I don't think it's necessary to perform multi-second long queries when only loading the admin dashboard, when most of the time the user won't be visiting the page for forum statistics.

SychO9 (approved):
I always be writing respository as well..

PR #3534, opened 2022-07-14
feat(likes): Option to prevent users liking their own posts

imorland:

Introduces an option to prevent forum users liking their own posts. Default to `allow`, as per current behavior. 

I've literally lost track of how many times I've been asked to override this, so I believe it makes sense to include the option directly in the extension 🤔 

I'm not 100% convinced on the translation copy. Does this wording make sense?

**Screenshot**
image

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/extensions/likes/extend.php
+++ b/extensions/likes/extend.php
@@ -7,6 +7,8 @@
  * LICENSE file that was distributed with this source code.
  */
 
+namespace Flarum\Likes;
+
 use Flarum\Api\Controller;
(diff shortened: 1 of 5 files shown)


davwheat (approved):
Copy seems fine to me

SychO9 (commented):
Code looks good, haven't tested it locally yet. Could we add an integration test for this?

davwheat (approved):

PR #3540, opened 2022-07-16
feat: use `position: sticky` for discussion side nav

### LINKED ISSUE #3541: Use sticky position instead of fixed for the scrubber
Currently the scrubber is fixed to the window which can create issues where the scrubber overlaps with additional bottom elements like a custom footer for example. [Sticky](https://caniuse.com/?search=sticky) has been around for a while now, and does a better job by only fixing the scrubber element within the discussion container.

We should look into it, properly test it and see if we can safely switch to it as that would be a big improvement to the UI.

**Reference:** https://discuss.flarum.org/d/26312-fix-footer-error

davwheat:
**Fixes #3541**

Swaps `position: fixed` for `position: sticky` for the Discussion sidebar. This helps to increase the amount of screen real estate for extension to add controls to the sidebar, when a large discussion hero is used.

Removal of the `overflow-x: hidden` is needed for sticky to work correctly. This shouldn't pose any issues for us, but bad extensions could cause x overflow to occur now.

> We _could_ keep it, but set to `visible` when the selector is `.App.App-discussion` instead.

**Screenshot**

https://user-images.githubusercontent.com/7406822/179372878-d3c889d2-c1f4-478a-81aa-74885f5114a5.mp4

https://user-images.githubusercontent.com/7406822/179372873-b98ce48a-9381-49c6-959e-370e239cfd62.mp4

**Necessity**

**Confirmed**

### DIFF
--- a/framework/core/less/common/App.less
+++ b/framework/core/less/common/App.less
@@ -3,7 +3,6 @@
   padding-top: var(--header-height);
   padding-bottom: 50px;
   min-height: 100vh;
-  max-width: 100vw;
 
   @media @phone {
     padding-top: var(--header-height-phone);
--- a/framework/core/less/common/App.less
+++ b/framework/core/less/common/App.less
@@ -2,8 +2,8 @@
   position: relative !important;
   padding-top: var(--header-height);
   padding-bottom: 50px;
-  overflow-x: hidden;
   min-height: 100vh;
+  max-width: 100vw;
 
   @media @phone {
     padding-top: var(--header-height-phone);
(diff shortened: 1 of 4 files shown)


framework/core/less/common/App.less
```diff
@@ -2,8 +2,8 @@
   position: relative !important;
   padding-top: var(--header-height);
   padding-bottom: 50px;
-  overflow-x: hidden;
   min-height: 100vh;
+  max-width: 100vw;
```
SychO9: I prefer moving`overflow-h: hidden` to the body, it doesn't mess with the sticky and continues to prevent horizontal scrolling

SychO9 (changes requested):
🤩

framework/core/less/common/App.less
davwheat: Woah, I did not know that! Time to edit tens of other projects now, I guess... 🙈

SychO9 (approved):
Looks good!

PR #3548, opened 2022-07-19
fix: potential static caching memory exhaustion

davwheat:

Removes static caching and model filtering in favour of adding extra where clauses to a relation.

Why was the static caching utilised in the first place?

**Necessity**

**Confirmed**

### DIFF
--- a/framework/core/src/User/User.php
+++ b/framework/core/src/User/User.php
@@ -439,7 +439,7 @@ class User extends AbstractModel
      */
     public function getUnreadNotificationCount()
     {
-        return $this->queryUnreadNotifications()->count();
+        return $this->unreadNotifications()->count();
     }
 
     /**
@@ -447,7 +447,7 @@ class User extends AbstractModel
      *
      * @return \Illuminate\Database\Eloquent\Relations\HasMany
      */
-    protected function queryUnreadNotifications()
+    protected function unreadNotifications()
     {
         return $this->notifications()
             ->whereIn('type', $this->getAlertableNotificationTypes())
@@ -463,7 +463,7 @@ class User extends AbstractModel
      */
     protected function getUnreadNotifications()
     {
-        return $this->queryUnreadNotifications()->get();
+        return $this->unreadNotifications()->get();
     }
 
     /**
@@ -473,7 +473,7 @@ class User extends AbstractModel
      */
     public function getNewNotificationCount()
     {
-        return $this->queryUnreadNotifications()
+        return $this->unreadNotifications()
             ->where('created_at', '>', $this->read_notifications_at)
             ->count();
     }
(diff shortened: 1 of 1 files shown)


framework/core/src/User/User.php
```diff
@@ -439,7 +439,21 @@ public function getAlertableNotificationTypes()
...
-        return $this->getUnreadNotifications()->count();
+        return $this->queryUnreadNotifications()->count();
+    }
+
+    /**
+     * Return query builder for all notifications that have not been read yet.
+     *
+     * @return \Illuminate\Database\Eloquent\Relations\HasMany
+     */
+    protected function queryUnreadNotifications()
```
davwheat: should probably just be `unreadNotifications()`

SychO9 (commented):
This seems to be the only relevant discussion: https://github.com/flarum/framework/issues/500. As far as I can tell the static caching was only done for the sake of avoiding two counting queries. But I'd rather have two count queries than do this caching and risk causing bugs/memory exhaustion issues as we saw in realtime ext.

So I'm in favor of moving forward with this.

framework/core/src/User/User.php
```diff
@@ -470,9 +473,9 @@ protected function getUnreadNotifications()
      */
     public function getNewNotificationCount()
     {
-        return $this->getUnreadNotifications()->filter(function ($notification) {
-            return $notification->created_at > $this->read_notifications_at ?: 0;
-        })->count();
+        return $this->queryUnreadNotifications()
```
davwheat:
```suggestion
        return $this->unreadNotifications()
```

framework/core/src/User/User.php
```diff
@@ -449,18 +463,7 @@ public function getUnreadNotificationCount()
...
-            $cached[$this->id] = $this->notifications()
-                ->whereIn('type', $this->getAlertableNotificationTypes())
-                ->whereNull('read_at')
-                ->where('is_deleted', false)
-                ->whereSubjectVisibleTo($this)
-                ->get();
-        }
-
-        return $cached[$this->id];
+        return $this->queryUnreadNotifications()->get();
```
davwheat:
```suggestion
        return $this->unreadNotifications()->get();
```

framework/core/src/User/User.php
```diff
@@ -439,7 +439,21 @@ public function getAlertableNotificationTypes()
      */
     public function getUnreadNotificationCount()
     {
-        return $this->getUnreadNotifications()->count();
+        return $this->queryUnreadNotifications()->count();
```
davwheat:
```suggestion
        return $this->unreadNotifications()->count();
```

framework/core/src/User/User.php
davwheat:
```suggestion
    protected function unreadNotifications()
```

luceos (approved):

PR #3553, opened 2022-07-23
chore: Setup PHPStan Level 5

### LINKED ISSUE #3559: PHPStan Setup
PHP doesn't give us that much built-in type safety. With PHPStan, we could perform static analysis to confirm:

- [Closure typing](https://phpstan.org/writing-php-code/phpdoc-types#callables)
- [Array shapes](https://phpstan.org/writing-php-code/phpdoc-types#array-shapes)
- [Typed arrays](https://phpstan.org/writing-php-code/phpdoc-types#general-arrays)
- [Generics](https://phpstan.org/blog/generics-in-php-using-phpdocs)

While this probably won't be 100% ideal (especially when it comes to editor integration), it would be a **HUGE** step up from where we are now.

SychO9:
**Closes #3559**

* Sets up PHPStan infrastructure.
* Adapts Docblocks and certain backend typings.
* Fixes some issues caught by phpstan.

* Pay most attention to codebase edits.
* Generally they are just minor docblock improvements.
* ___The most notable change is___ the replacement of `\Illuminate\Database\ConnectionInterface` by `\Illuminate\Database\Connection`. Ideally we would stick to injecting the interface as that is more correct to separate the implementation from the contract. But in most of our codebase, we use methods that only exist within the implementation and not the interface. We could just leave it and fool phpstan with docblocks, but I thought it would be better to be explicit about what we expect injected. Thoughts are very welcome about this especially.

**Necessity**

**Confirmed**


framework/core/src/Api/Serializer/ForumSerializer.php
```diff
@@ -39,7 +40,7 @@ class ForumSerializer extends AbstractSerializer
     protected $url;
 
     /**
-     * @var Cloud
+     * @var Filesystem
```
luceos: This must be the Cloud interface as previously declared. Only the Cloud interface supports retrieving URL's.. Not sure why this change is required?

framework/core/src/Api/Serializer/ForumSerializer.php
```diff
@@ -132,6 +133,9 @@ protected function getFaviconUrl()
 
     public function getAssetUrl($assetPath): string
     {
-        return $this->assetsFilesystem->url($assetPath);
+        /** @var Cloud $assetsFilesystem */
+        $assetsFilesystem = $this->assetsFilesystem;
```
luceos: That's the hack. Not a pretty one. Because enforcing the Cloud interface guarantees the url method to exist. The standard Filesystem interface does not have one and there are definitely implementation that do not implement both.

framework/core/src/Database/DatabaseMigrationRepository.php
```diff
@@ -30,10 +30,10 @@ class DatabaseMigrationRepository implements MigrationRepositoryInterface
     /**
      * Create a new database migration repository instance.
      *
-     * @param  \Illuminate\Database\ConnectionInterface $connection
-     * @param  string                                   $table
+     * @param  Connection $connection
+     * @param  string $table
      */
-    public function __construct(ConnectionInterface $connection, $table)
+    public function __construct(Connection $connection, $table)
```
luceos: The reason an interface is used is due to ioc. The interface is always bound, you cannot be sure about this abstract class. I don't feel this is a good change. And I definitely recommend against it.

framework/core/src/Database/DatabaseServiceProvider.php
```diff
@@ -23,6 +23,7 @@ class DatabaseServiceProvider extends AbstractServiceProvider
     public function register()
     {
         $this->container->singleton(Manager::class, function (Container $container) {
+            /** @var \Illuminate\Container\Container $container */
```
luceos: Is this change required? The Container is already type hinted as the argument of the method.

framework/core/src/Database/Migrator.php
```diff
@@ -39,24 +39,21 @@ class Migrator
...
      * The output interface implementation.
      *
-     * @var OutputInterface
+     * @var OutputInterface|null
      */
     protected $output;
+
     /**
-     * @var ConnectionInterface|MySqlConnection
+     * @var Connection
```
luceos: Please let's not make this change.

framework/core/src/Database/Migrator.php
```diff
@@ -39,24 +39,21 @@ class Migrator
...
      * Create a new migrator instance.
-     *
-     * @param  MigrationRepositoryInterface  $repository
-     * @param  ConnectionInterface           $connection
-     * @param  Filesystem                    $files
      */
     public function __construct(
         MigrationRepositoryInterface $repository,
-        ConnectionInterface $connection,
+        Connection $connection,
```
luceos: Please let's not make this change.

framework/core/src/Extend/View.php
```diff
@@ -65,7 +65,7 @@ public function extendNamespace(string $namespace, $hints): self
 
     public function extend(Container $container, Extension $extension = null)
     {
-        $container->resolving(Factory::class, function (Factory $view) {
+        $container->resolving(Factory::class, function (\Illuminate\View\Factory $view) {
```
luceos: Please import, even if we have to use an alias 

`use Illuminate\View\Factory as IlluminateFactory;`

We never really use FQCN's in code without importing.

framework/core/src/Extension/ExtensionManager.php
```diff
@@ -333,13 +333,13 @@ public function migrate(Extension $extension, $direction = 'up')
...
      * @param Extension $extension
-     * @return array Notes from the migrator.
+     * @return void
      *
      * @internal
      */
     public function migrateDown(Extension $extension)
     {
-        return $this->migrate($extension, 'down');
+        $this->migrate($extension, 'down');
```
luceos, line 342: Won't this break output handling, eg for showing the output from the migrate command? Removing the ability to read the output from migrating is, at least from an extensibility standpoint, not helpful.

framework/core/src/Foundation/Application.php
```diff
@@ -140,10 +140,10 @@ protected function registerBaseBindings()
          * Use container inside flarum instead.
          */
         $this->container->instance('app', $this->container);
-        $this->container->alias('app', \Illluminate\Container\Container::class);
+        $this->container->alias('app', \Illuminate\Container\Container::class);
```
luceos, line 143: Please import ;) Should have happened before.

framework/core/src/Foundation/Application.php
```diff
@@ -140,10 +140,10 @@ protected function registerBaseBindings()
...
 
         $this->container->instance('container', $this->container);
-        $this->container->alias('container', \Illluminate\Container\Container::class);
+        $this->container->alias('container', \Illuminate\Container\Container::class);
```
luceos, line 146: Please import ;) Should have happened before.

framework/core/src/Foundation/Console/InfoCommand.php
```diff
@@ -51,7 +52,7 @@ public function __construct(
         ExtensionManager $extensions,
         Config $config,
         SettingsRepositoryInterface $settings,
-        ConnectionInterface $db,
+        Connection $db,
```
luceos: Please no.

framework/core/src/Foundation/ErrorHandling/HandledError.php
```diff
@@ -28,7 +28,7 @@ class HandledError
 
     public static function unknown(Throwable $error)
     {
-        return new static($error, 'unknown', 500);
+        return new self($error, 'unknown', 500);
```
luceos: When extending the handler, this will cause this call to return an instance of this class not the extended class. Why is this change required?

framework/core/src/Frontend/Compiler/FileVersioner.php
```diff
@@ -27,8 +27,8 @@ public function __construct(Filesystem $filesystem)
 
     public function putRevision(string $file, ?string $revision)
     {
-        if ($this->filesystem->has(static::REV_MANIFEST)) {
-            $manifest = json_decode($this->filesystem->read(static::REV_MANIFEST), true);
+        if ($this->filesystem->exists(static::REV_MANIFEST)) {
```
luceos, line 30: From previous experience it might be better to just to one call `get` and if that returns anything, use that. Doing an `exists` and then a `get` is two calls. In scalable environments where the filesystem is s3 this surmounts to a lot of calls fast.

framework/core/src/Frontend/Compiler/FileVersioner.php
```diff
@@ -44,8 +44,8 @@ public function putRevision(string $file, ?string $revision)
 
     public function getRevision(string $file): ?string
     {
-        if ($this->filesystem->has(static::REV_MANIFEST)) {
-            $manifest = json_decode($this->filesystem->read(static::REV_MANIFEST), true);
+        if ($this->filesystem->exists(static::REV_MANIFEST)) {
```
luceos, line 47: See other remark about exists + get

framework/core/src/Frontend/Compiler/JsCompiler.php
```diff
@@ -72,7 +75,7 @@ protected function delete(string $file)
     {
         parent::delete($file);
 
-        if ($this->assetsDir->has($mapFile = $file.'.map')) {
+        if ($this->assetsDir->exists($mapFile = $file.'.map')) {
```
luceos, line 75: Same applies here. If this throws an error, just capture/dismiss that.

framework/core/src/Frontend/Compiler/RevisionCompiler.php
```diff
@@ -24,6 +24,7 @@ class RevisionCompiler implements CompilerInterface
      * @var Filesystem
```
luceos: We are relying on url in the code. The interface should be Cloud.

framework/core/src/Frontend/Compiler/RevisionCompiler.php
```diff
@@ -145,7 +149,6 @@ protected function save(string $file, array $sources): bool
 
     /**
      * @param SourceInterface[] $sources
-     * @return string
```
luceos, line 148: If we have phpdoc we always phpdoc everything, so as long as we cannot typehint the SourceInterface on the actual method, we should keep the return phpdoc too.

framework/core/src/Frontend/Compiler/RevisionCompiler.php
```diff
@@ -158,18 +161,13 @@ protected function compile(array $sources): string
...
-     */
     protected function format(string $string): string
     {
         return $string;
     }
 
     /**
      * @param SourceInterface[] $sources
-     * @return string
```
luceos, line 172: Rather not remove return type hint as long as we cannot get rid of all phpdoc.

framework/core/src/Frontend/Compiler/RevisionCompiler.php
```diff
@@ -196,12 +194,9 @@ public function flush()
         }
     }
 
-    /**
-     * @param string $file
-     */
     protected function delete(string $file)
     {
-        if ($this->assetsDir->has($file)) {
+        if ($this->assetsDir->exists($file)) {
```
luceos, line 196: Can't we do one call and capture an error?

framework/core/src/Install/Steps/EnableBundledExtensions.php
```diff
@@ -60,7 +60,12 @@ class EnableBundledExtensions implements Step
      */
     private $enabledExtensions;
 
-    public function __construct(ConnectionInterface $database, $vendorPath, $assetPath, $enabledExtensions = null)
+    /**
+     * @var Migrator|null
+     */
+    private $migrator;
+
+    public function __construct(Connection $database, $vendorPath, $assetPath, $enabledExtensions = null)
```
luceos: Rather not change the db interface.

framework/core/src/Install/Steps/RunMigrations.php
```diff
@@ -27,7 +27,7 @@ class RunMigrations implements Step
      */
     private $path;
 
-    public function __construct(ConnectionInterface $database, $path)
+    public function __construct(Connection $database, $path)
```
luceos: Rather not change the interface.

framework/core/src/Notification/Notification.php
```diff
@@ -220,7 +220,7 @@ public function scopeMatchingBlueprint(Builder $query, BlueprintInterface $bluep
      */
     public static function notify(array $recipients, BlueprintInterface $blueprint)
     {
-        $attributes = static::getBlueprintAttributes($blueprint);
+        $attributes = self::getBlueprintAttributes($blueprint);
```
luceos: Same applies here. Self will return this class when called, not the class extending this. As such for extensibility this is bad.

framework/core/src/Post/Post.php
```diff
@@ -94,7 +94,7 @@ public static function boot()
         static::creating(function (self $post) {
             $post->type = $post::$type;
 
-            /** @var ConnectionInterface $db */
+            /** @var Connection $db */
```
luceos: Rather not.

framework/core/src/Queue/QueueServiceProvider.php
```diff
@@ -52,7 +52,7 @@ public function register()
 
         // Extensions can override this binding if they want to make Flarum use
         // a different queuing backend.
-        $this->container->singleton('flarum.queue.connection', function (Container $container) {
+        $this->container->singleton('flarum.queue.connection', function (\Illuminate\Container\Container $container) {
```
luceos: 😕 why aren't we importing?

framework/core/src/User/EmailToken.php
```diff
@@ -76,12 +76,12 @@ public function user()
      *
      * @param \Illuminate\Database\Eloquent\Builder $query
      * @param string $id
-     * @return static
+     * @return self
```
luceos: Why is static invalid? Perhaps this then should be EmailToken, so that it better matches when something extends this class.

luceos (changes requested):
Okay this was a big one.

Most of this looks great, a few things came up that we need to take a decision on:

- I prefer not switching from the DatabaseInterface to the Database abstract. The interface is guaranteed to be bound into ioc, that's the whole idea of the interface. Using the implementation everywhere seems off to me. If it really helps we can typehint the property with that class as well `DatabaseInterface|Connection`, but dropping the interface - especially in construct arguments typehints - is going to bite us in the long run.
- I noticed some places where full class names were used and not imported, let's try to import all the things.
- Some changes to go from filesystem Cloud interface to Filesystem aren't correct. Only the Cloud interface has the `url` method we need. We cannot drop that or we'll break things.
- I saw that we are doing `exists` and `get` or `delete` sometimes. If possible it would be a huge deal in cutting latency if we can use only `get` and `delete` and capture any thrown errors to reduce api calls to cloud based storage.
- In some changes we moved from `self` to `static`. Now I can be wrong, but I thought that doing that enforces the returned class to be **this** implementation and not **any** class extending this implementation. This would be a huge deal in extensibility and being able to apply raw/higher level overrides.

SychO9:
_(@luceos gonna reply here globally here about some of the issues raised here)_

**Database connection interface**
that's the thing with `ConnectionInterface`, I dislike having to inject the actual implementation, but we either do that because that's what our code actually expects because it uses methods exclusive to the implementation or we hack around it.

```php
// The interface
interface ConnectionInterface { ... }

// Illuminate implementation
abstract class Connection extends ConnectionInterface
{
    public function getSchemaBuilder(); // doesn't exist in the interface
}

// Solution 1: actually inject the implementation because that's what we actually expect.
class Example
{
    public function __construct(Connection $db) { ... }

    public function example()
    {
        $this->db->getSchemaBuilder()->...
    }
}

// Solution 2: hack around it everywhere we expect implementation-specific instances.
class Example
{
    public function __construct(ConnetionInterface $db) { ... }

    public function example()
    {
        /** @var Connection */
        $db = $this->db;

        $db->getSchemaBuilder()->...
    }
}
```

---

**Filesystem vs Cloud** (https://github.com/flarum/framework/pull/3553#discussion_r933761636, https://github.com/flarum/framework/pull/3553#discussion_r933761734)
Illuminate's filesystem `Factory` contract `disk` method returns a `Filesystem` contract, not `Cloud`. And since PHP doesn't have generics, we're gonna have to hack it with a docblock one way or the other.

---

**self vs static**
PHPStan complains about using `static` as it can be unsafe when extending and lead to fatal errors. I'm personally not a fan of using static unless it really makes sense. Otherwise, we could add a global ignore for phpstan's error about the static keyword.

---

**`exists()` vs `get()`**
I understand the mentioned performance gain, but I believe that is outside the scope of this PR. We only change `has` to `exists` and others because the older methods are deprecated.

tankerkiller125:
@SychO9 in regards to generics PHPStan has a way to work around that issue https://phpstan.org/blog/generics-in-php-using-phpdocs

NVM, I just saw your mention of the docblock 🤦

luceos (approved):

luceos:
Thanks for the clarification. I hope this doesn't bite is in the behind in the long run 💪

SychO9:
I do think it might be better to use stubs to tell phpstan that the interface does have the methods even though that's not the case.

Same for the filesystem, perhaps using a stub to change the return type to `Cloud` might be better.

Not sure yet, need to consider this more.

framework/core/src/Database/DatabaseServiceProvider.php
SychO9: The typehint is for the contract, laravel's manager expects the implementation class (another case of difficulty injecting contract vs implementation)

framework/core/src/Extension/ExtensionManager.php
SychO9: ~~the migrate method doesn't return anything (void).~~ Nvm it returns int or void

framework/core/src/Frontend/Compiler/RevisionCompiler.php
SychO9: For what reason though 🤔

SychO9:
**New changes:**
- Reverts to injecting `ConnectionInterface` instead of `Connection` and ignores related errors, while our codebase expects the actual implementation and it would make sense to inject it instead. I feel like it would be more sane to keep the interface instead.
- Uses a stub to modify the `Filesystem` contract to have the disk return type changed to `Cloud`. That's what our codebase expects everywhere.
- Also revets `static` to `self` changes and ignores errors. I think this needs more discussion in the long run, while we do this for the sake of extensibility, it just seems to be used anywhere and everywhere, even in private API.

luceos:
> * Also revets `static` to `self` changes and ignores errors. I think this needs more discussion in the long run, while we do this for the sake of extensibility, it just seems to be used anywhere and everywhere, even in private API.

You mean you reverted `self` to `static`?

I'm okay with this if that's the case 👍 Thanks for considering my argumentation.

SychO9:
yup `self` to `static`my bad

framework/core/src/Foundation/Application.php
askvortsov1: Illllllllllllllllllllllllllluminate

framework/core/src/Frontend/Compiler/RevisionCompiler.php
askvortsov1: My gut feeling leans towards agreeing with Sami. PHPdoc has 2 benefits:

- Visual documentation
- Static typechecking (through PHPStan)

Both of these should be accomplished by the PHP-native return typehint.

composer.json
```diff
@@ -178,5 +178,11 @@
                 "extension.neon"
             ]
         }
+    },
+    "scripts": {
+        "analyse:phpstan": "phpstan analyse"
+    },
+    "scripts-descriptions": {
+        "analyse:phpstan": "Run static analysis"
     }
```
askvortsov1, line 187: Sweet, so we can run this for the entire monorepo, all at once?

framework/core/src/Api/Serializer/ForumSerializer.php
```diff
@@ -133,6 +133,8 @@ protected function getFaviconUrl()
 
     public function getAssetUrl($assetPath): string
     {
-        return $this->assetsFilesystem->url($assetPath);
+        $assetsFilesystem = $this->assetsFilesystem;
+
+        return $assetsFilesystem->url($assetPath);
```
askvortsov1: Is this still necessary with the stubbing changes?

framework/core/src/Database/DatabaseServiceProvider.php
```diff
@@ -22,7 +23,7 @@ class DatabaseServiceProvider extends AbstractServiceProvider
      */
     public function register()
     {
-        $this->container->singleton(Manager::class, function (Container $container) {
+        $this->container->singleton(Manager::class, function (ContainerImplementation $container) {
```
askvortsov1, line 26: We can't inject the interface here because of the custom `'flarum'` key?

framework/core/src/Discussion/Discussion.php
```diff
@@ -224,8 +225,8 @@ public function setLastPost(Post $post)
      */
     public function refreshLastPost()
     {
-        /** @var Post $lastPost */
         if ($lastPost = $this->comments()->latest()->first()) {
+            /** @var Post $lastPost */
             $this->setLastPost($lastPost);
```
askvortsov1, line 230: nit: `latestPost` feels like better naming.

askvortsov1: Also, is the relationship not typed on the Eloquent level?

framework/core/src/Discussion/DiscussionRepository.php
```diff
@@ -28,40 +28,39 @@ public function query()
...
     }
 
     /**
      * Get the IDs of discussions which a user has read completely.
      *
-     * @deprecated 1.3 Use `getReadIdsQuery` instead
-     *
      * @param User $user
-     * @return array
+     * @return \Illuminate\Database\Eloquent\Collection<Discussion>
```
askvortsov1, line 46: Should this be imported instead of using the fully qualified name?

framework/core/src/Discussion/DiscussionRepository.php
```diff
@@ -28,40 +28,39 @@ public function query()
...
      * Get a query containing the IDs of discussions which a user has read completely.
      *
      * @param User $user
-     * @return Builder
+     * @return Builder<Discussion>
      */
     public function getReadIdsQuery(User $user): Builder
     {
-        return Discussion::leftJoin('discussion_user', 'discussion_user.discussion_id', '=', 'discussions.id')
+        return Discussion::query()
```
askvortsov1: Why not `$this->query()` as before?

framework/core/src/Extend/ErrorHandling.php
```diff
@@ -76,7 +76,7 @@ public function type(string $exceptionClass, string $errorType): self
      * contain "details" - arbitrary data with more context for to the error.
      *
      * @param string $exceptionClass: The ::class attribute of the exception class.
-     * @param string $errorType: The ::class attribute of the handler class.
+     * @param string $handlerClass: The ::class attribute of the handler class.
```
askvortsov1, line 79: I believe that we can use `class-string<T>` for these. Although maybe best left for a followup PR so we can get this landed.

https://phpstan.org/blog/generics-by-examples

framework/core/src/Extension/Extension.php
```diff
@@ -498,7 +496,7 @@ public function migrate(Migrator $migrator, $direction = 'up')
         }
 
         if ($direction == 'up') {
-            return $migrator->run($this->getPath().'/migrations', $this);
+            $migrator->run($this->getPath().'/migrations', $this);
```
askvortsov1, line 499: Shouldn't this also be a return? If the migrator run fails, feels like we should return that as well.

framework/core/src/Foundation/Application.php
askvortsov1, line 143: These probably should be imported though, unless I missed another explanation of why not?

askvortsov1 (changes requested):
re: the exists + get into one issue: if we want to do this, we should consider making either a wrapper interface/class or an explicit util method. Probably the latter.

framework/core/src/User/User.php
```diff
@@ -311,14 +311,15 @@ public function changeAvatarPath($path)
...
      * @param string|null $value
      * @return string
      */
     public function getAvatarUrlAttribute(string $value = null)
     {
         if ($value && strpos($value, '://') === false) {
-            return resolve(Factory::class)->disk('flarum-avatars')->url($value);
+            $disk = resolve(Factory::class)->disk('flarum-avatars');
+
+            return $disk->url($value);
```
askvortsov1: Same as above, is this still necessary?

framework/core/src/User/User.php
```diff
@@ -410,15 +411,6 @@ public function hasPermissionLike($match)
...
 
-    private function checkForDeprecatedPermissions($permission)
-    {
-        foreach (['viewDiscussions', 'viewUserList'] as $deprecated) {
-            if (strpos($permission, $deprecated) !== false) {
-                trigger_error('The `viewDiscussions` and `viewUserList` permissions have been renamed to `viewForum` and `searchUsers` respectively. Please use those instead.', E_USER_DEPRECATED);
-            }
-        }
-    }
-
```
askvortsov1, line 421: We might as well keep this in, but it feels outside the scope of this PR.

php-packages/phpstan/stubs/Illuminate/Contracts/Filesystem/Cloud.stub
```diff
@@ -0,0 +1,7 @@
+<?php
+
+namespace Illuminate\Contracts\Filesystem;
+
+interface Cloud extends Filesystem
```
askvortsov1, line 5: We'll want to add phpdoc to the filesystem driver such that it only allows class strings implementing Cloud

askvortsov1 (changes requested):
Another general comment. From a maintainer's / reviewer's perspective, it can be hard to tell why certain things are overriden. I think it would be helpful in the future if there were comments on each stub file explaining what the goal of overriding / stubbing the class in question is.

composer.json
SychO9: Yes, everything is currently setup for it to run on all the monorepo at once, including the workflow.

framework/core/src/Database/DatabaseServiceProvider.php
SychO9: no, because we instantiate a laravel `Manager` class that expects the implementation rather than the contract. Another case of many where it's hard to use laravel's internals because it doesn't expect the contracts, we can use stubs to fix this as well if we want to.

framework/core/src/Discussion/Discussion.php
SychO9: > nit: `latestPost` feels like better naming.

in what sense? the last post is referred to last post everywhere.

framework/core/src/Discussion/DiscussionRepository.php
SychO9: phpDocs seem to be the exception looking at our codebase

framework/core/src/Extend/ErrorHandling.php
SychO9: Yup we can do a lot more with features like those other the higher we up the level of phpstan.

framework/core/src/Extension/Extension.php
SychO9: the `run` method of the migrator doesn't actually return anything, same goes for methods called within `run`.

framework/core/src/Foundation/Application.php
SychO9: Felt out of scope, there are so many instances like these, and this PR does so much already, I'd rather we do anything else that isn't necessarily related to setting up phpstan in other PRs.

framework/core/src/User/User.php
SychO9: The method is not called anywhere, that's why phpstan complains about it. I'm pretty sure that when we added it we forgot to actually make use of it.

php-packages/phpstan/stubs/Illuminate/Contracts/Filesystem/Cloud.stub
SychO9: Sorry I don't follow 🤔

framework/core/src/Database/DatabaseServiceProvider.php
askvortsov1: Let's keep it as implementation then; better to be accurate where we can/

framework/core/src/Discussion/Discussion.php
askvortsov1: Nevermind, didn't look into the context.

framework/core/src/Extension/Extension.php
askvortsov1: Fair enough!

framework/core/src/Foundation/Application.php
askvortsov1: Good point, let's aim to get this merged shortly after bundling the release so that we can build on it going forward.

framework/core/src/User/User.php
askvortsov1: Yeah fair enough, let's kill it.

php-packages/phpstan/stubs/Illuminate/Contracts/Filesystem/Cloud.stub
askvortsov1: I meant we should change the extender's phpdocs so that they use `class-string` of an interface annotations instead of just string. If/when we roll out PHPStan to extensions, that should ensure that they are being used properly.

askvortsov1 (approved):

PR #3564, opened 2022-07-25
feat: publish assets on admin dashboard cache clear

davwheat:

- passes error to frontend when cache clear command returns non-zero exit code
- publish core and extension assets when clearing from the admin dashboard

The reason for this is two-fold:
- clear cache on admin page is often a "go-to" solution for Flarum weirdness, and we should include publishing ext assets in this as it grows more popular with recent FoF Webpack chunking, etc
- publishing assets is important, but otherwise not possible on non-ssh capable shared hosting environments

Not familiar with this area of core. Did I do it properly? 🙈 

**Necessity**

**Confirmed**

### DIFF
--- a/framework/core/src/Api/Controller/ClearCacheController.php
+++ b/framework/core/src/Api/Controller/ClearCacheController.php
@@ -11,6 +11,7 @@ namespace Flarum\Api\Controller;
 
 use Flarum\Foundation\Console\AssetsPublishCommand;
 use Flarum\Foundation\Console\CacheClearCommand;
+use Flarum\Foundation\IOException;
 use Flarum\Http\RequestUtil;
 use Laminas\Diactoros\Response\EmptyResponse;
 use Psr\Http\Message\ServerRequestInterface;
@@ -40,6 +41,7 @@ class ClearCacheController extends AbstractDeleteController
 
     /**
      * {@inheritdoc}
+     * @throws IOException|\Flarum\User\Exception\PermissionDeniedException
      */
     protected function delete(ServerRequestInterface $request)
     {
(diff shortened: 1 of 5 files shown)


iamdarkle:
This is brilliant actually. Goodbye `php flarum assets:publish` I won't miss you 👋

imorland (commented):
Tested locally, all seems to work as intended.

One thought however; should we consider renaming the controller? `ClearCacheController` no longer accurately describes what's happening here now? `ClearCacheAndAssetsController` maybe? 🤔 

Also, should we perhaps update the `Please wait` UI, to something like `Please wait, whilst we clear the cache and re-publish assets`?

image

davwheat:
Wouldn't renaming it be a breaking change?

I do agree that the front end portion could do with a bit of love, though.

framework/core/src/Api/Controller/ClearCacheController.php
```diff
@@ -38,11 +45,24 @@ protected function delete(ServerRequestInterface $request)
...
         RequestUtil::getActor($request)->assertAdmin();
 
-        $this->command->run(
+        $exitCode = $this->command->run(
             new ArrayInput([]),
             new NullOutput()
         );
 
+        if ($exitCode !== 0) {
+            throw new \Exception('Clearing cache failed. Try running `php flarum cache:clear` from the command line to see more info.');
```
SychO9: We can't throw a broad exception to the end user, it will just result in a server error without any meaningful message 🤔.

Let me push something for this real quick. (sorry for hijacking 🙈 i know for a fact exceptions are a bit of a pain)

SychO9 (changes requested):

framework/core/src/Api/Controller/ClearCacheController.php
davwheat: Oh, no worries. Rather someone does it properly! :P

SychO9 (approved):
I'm fine with the controller name as is, I think it's still appropriate since the endpoint and its description is clearing cache, even if we might do other things as well.

imorland (approved):

PR #3565, opened 2022-07-26
feat: Restart the queue worker after cache clearing, ext enable/disable, save settings

### LINKED ISSUE #3563: Restart the queue on cache clear, extension enable/disable and extension settings save

imorland:
**Fixes #3563 **

Adds a subscriber to trigger the queue to restart after certain events

Tested locally with a standard redis based queue. Doesn't work with `blomstra/horizon`, although I think there's a seperate issue with that extension, which I'll start to investigate now

**Necessity**

**Confirmed**

### DIFF
--- a/framework/core/src/Queue/QueueRestarter.php
+++ b/framework/core/src/Queue/QueueRestarter.php
@@ -1,8 +1,14 @@
 <?php
 
+/*
+ * This file is part of Flarum.
+ *
+ * For detailed copyright and license information, please view the
+ * LICENSE file that was distributed with this source code.
+ */
+
 namespace Flarum\Queue;
 
-use Carbon\Carbon;
 use Flarum\Extension\Event\Disabled;
 use Flarum\Extension\Event\Enabled;
 use Flarum\Foundation\Event\ClearingCache;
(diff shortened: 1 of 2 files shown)


framework/core/src/Queue/QueueRestarter.php
```diff
@@ -0,0 +1,46 @@
...
+        /** @var RestartCommand $command */
+        $command = resolve(RestartCommand::class);
+
+        $command->setLaravel($container);
+
+        $command->run(
+            new ArrayInput([]),
+            new NullOutput
+        );
+    }
```
SychO9: Let's inject the dependencies into the class:
```suggestion
    /**
     * @var Container
     */
    protected $container;
    
    /**
     * @var RestartCommand
     */
    protected $command;

    public function __construct(Container $container, RestartCommand $command)
    {
        $this->container = $container;
        $this->command = $command;
    }

    public function subscribe(Dispatcher $events)
    {
        $events->listen([
            ClearingCache::class, Saved::class,
            Enabled::class, Disabled::class
        ], [$this, 'restart']);
    }

    public function restart()
    {
        $this->command->setLaravel($this->container);

        $this->command->run(
            new ArrayInput([]),
            new NullOutput
        );
    }
```

SychO9 (changes requested):

SychO9 (approved):

PR #3566, opened 2022-07-26
feat: expose assets base url to frontend forum model

davwheat:

Currently extensions have to guess the assets base URL or serialize it to the frontend themselves, resulting in either a lot of extra needless data, or extensions assuming assets will always be at `<baseUrl>/assets`.

This can cause issues if, for example, the assets are on a cloud provider like S3.

E.g., https://github.com/FriendsOfFlarum/profile-image-crop/blob/02e2e524d3355495c4ed0915f342a7e7bf8e2565/js/src/forum/components/ProfileImageCropModal.js#L55

**Necessity**

**Confirmed**

### DIFF
--- a/framework/core/src/Api/Serializer/ForumSerializer.php
+++ b/framework/core/src/Api/Serializer/ForumSerializer.php
@@ -92,7 +92,7 @@ class ForumSerializer extends AbstractSerializer
             'canViewForum' => $this->actor->can('viewForum'),
             'canStartDiscussion' => $this->actor->can('startDiscussion'),
             'canSearchUsers' => $this->actor->can('searchUsers'),
-            'assetsBaseUrl' => $this->assetsFilesystem->url(''),
+            'assetsBaseUrl' => rtrim($this->assetsFilesystem->url(''), '/'),
         ];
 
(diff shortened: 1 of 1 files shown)


luceos (approved):
That we didn't do this sooner 🙈

imorland (approved):
Good call 🥇

imorland:
Actually, thinking some more about this, do we _really_ need this? Does this not encourage path building in the frontend, rather than having anything required by the frontend to be constructed in the backend and passed thru a serializer, etc?

davwheat:
Yes, but I don't think that's a bad thing.

We shouldn't be constructing every possible path we ever want to use on the backend, otherwise, we'll end up with possibly multiple attributes per extension, bloating network requests.

davwheat:
Following some discussion through Discord, Ian has withdrawn his concern for now.

He has raised a good point about how this could be potentially misused in some cases. For example, constructing a URL for an avatar image URL should be done in the backend, not with this attribute and concatenating a path.

For now, the opinion is to roll with it and adjust in the run-up to Flarum 2.0, if needed.

PR #3567, opened 2022-07-28
feat: clear password & email tokens when appropriate

SychO9:
* Delete password tokens when the password is changed (40e60c61b844e6db79e5bdaad52f64831b87113a).
* Delete password tokens after email change (edf19f18ce24ee043044cde8e67a13324b37f255).
* Delete email tokens after password reset (8bcc6e6198ca5fe67fc92a9846de269b5b7d7cad).

For users with permission to edit credentials who can change an email without confirmation, email tokens aren't deleted, I kept that behavior unchanged, not sure if we should delete the tokens in that case as well.

**Necessity**

**Confirmed**

### DIFF
--- a/framework/core/src/User/TokensClearer.php
+++ b/framework/core/src/User/TokensClearer.php
@@ -1,0 +1,31 @@
+<?php
+
+/*
+ * This file is part of Flarum.
+ *
+ * For detailed copyright and license information, please view the
+ * LICENSE file that was distributed with this source code.
+ */
+
+namespace Flarum\User;
+
+use Flarum\User\Event\PasswordChanged;
+use Illuminate\Contracts\Events\Dispatcher;
+
+class TokensClearer
+{
+    public function subscribe(Dispatcher $events): array
+    {
+        return [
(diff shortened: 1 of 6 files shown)


framework/core/src/User/TokensClearer.php
```diff
@@ -0,0 +1,40 @@
...
+
+use Flarum\User\Event\EmailChanged;
+use Flarum\User\Event\PasswordChanged;
+use Illuminate\Contracts\Events\Dispatcher;
+
+class TokensClearer
+{
+    public function subscribe(Dispatcher $events): void
+    {
+        $events->listen(PasswordChanged::class, [$this, 'clearPasswordTokens']);
```
luceos:
```suggestion
        $events->listen([PasswordChanged::class, EmailChanged::class], [$this, 'clearPasswordTokens']);
```

This helps keep method declaration in one spot. Line 22 needs to be removed though.

Not sure which format you prefer, but using an array of events in `listen()` is supported.

luceos (approved):
Look at the one comment I left. Once that is resolved, either way, feel free to merge.

PR #3568, opened 2022-07-28
fix: prevent admin/mod deleting themselves

imorland:

Currently, a user with permission to delete other users (admins, mods, etc) may also delete themselves. This happened to me today as a forum admin user 🙈 

This PR proposes to prevent a user with elevated delete user permissions from deleting themselves.

**Necessity**

**Confirmed**

### DIFF
--- a/framework/core/src/User/Access/UserPolicy.php
+++ b/framework/core/src/User/Access/UserPolicy.php
@@ -39,4 +39,15 @@ class UserPolicy extends AbstractPolicy
             return $this->allow();
         }
     }
+
+    /**
+     * @param User $actor
+     * @param User $user
+     */
+    public function delete(User $actor, User $user)
+    {
+        if ($user->id === $actor->id) {
+            return $this->deny();
+        }
+    }
 }


davwheat (dismissed):
Makes sense 🤣

wojtekxtx:
> Makes sense 🤣

Absolutely not! Not to mention that, at least within EU, due to GDPR, not allowing user to delete his/hers account **is illegal**.

wojtekxtx (commented):
If this passess, any usage of Flarum will become illegal.

davwheat:
This can be overridden in a GDPR extension.

Additionally, Flarum provides no current way to delete your account, so this change doesn't inherently make Flarum "illegal".

GDPR only requires users to be able to delete their info if they request it, which can be done via email. It doesn't have to be self-serve.

imorland:
I should clarify, this is only for when a user has **delete user** permissions. It's sole purpose is to prevent self deletion by accident.

wojtekxtx:
> This can be overridden in a GDPR extension.

Even if so, than I think this shouldnt be merged. `Core` should be law-compliant the same way as extension/s

imorland:
> > This can be overridden in a GDPR extension.
> 
> 
> 
> Even if so, than I think this shouldnt be merged. `Core` should be law-compliant the same way as extension/s

This has nothing to do with the scenario you are talking about

wojtekxtx:
> I should clarify, this is only for when a user has **delete user** permissions. It's sole purpose is to prevent self deletion by accident.

In the light of GDPR user should be able to delete their account ( by clicking a button / link is the easiest form to implement ).
If Flarum dont have this ability, it means its not compliant with GDPR. So, any usage is risky. Period.

matteocontrini:
> If Flarum dont have this ability, it means its not compliant with GDPR.

GDPR is not about having buttons in the UI to delete accounts, so, well, no. This PR is not about that feature anyway 😅

wojtekxtx:
> > If Flarum dont have this ability, it means its not compliant with GDPR.
> 
> GDPR is not about having buttons in the UI to delete accounts

Of course GDPR does not spewcify the exact way of users deleting their accounts, but specifies that **users have to have the ability to delete both their account and its data ( posts/profile/attachements etc)**

imorland:
> > I should clarify, this is only for when a user has **delete user** permissions. It's sole purpose is to prevent self deletion by accident.
> 
> 
> 
> In the light of GDPR user should be able to delete their account ( by clicking a button / link is the easiest form to implement ).
> 
> If Flarum dont have this ability, it means its not compliant with GDPR. So, any usage is risky. Period.

GDPR is an important piece of legislation, and something that Flarum is currently in the process of implementing fully. 

This PR has no bearing on that aspect, so your concern, although valid, is out of scope for this PR

imorland:
Updated the description here to hopefully clarify the purpose of this change a little better 🤞

luceos:
Core already has a check to prevent deletion of the last admin. I honestly think one should be able to delete their own account with the permission granted to you should you feel the need. Otherwise other elevated users can keep your account hostage.

What do you think?

davwheat:
It's a fair point, actually.

Maybe this is more of a case for the frontend to require user confirmation before deleting a user where the ID matches the current session instead?

imorland:
> Core already has a check to prevent deletion of the last admin. I honestly think one should be able to delete their own account with the permission granted to you should you feel the need. Otherwise other elevated users can keep your account hostage.
> 
> What do you think?

Yes, there is the check to prevent deletion of the last admin - and rightly so. The permission to _hard delete_ other users, in my opinion is just that - **other users**, be that for reasons of spam/trolls or manual GDPR RTBF requests, etc. I don't believe this particular functionality was ever intended to be a "self deletion" offering, and the permission simply does not exist for being able to delete self, but not others.

For this reason, I'd be in favour of continuing to propose this change. 

Certainly a "are you sure" prompt would be good to have as well, perhaps as a seperate PR

SychO9:
I believe there is a browser confirmation before *any* user deletion.

imorland:
> I believe there is a browser confirmation before _any_ user deletion.

There is, but a **specific** one warning about deleting yourself would be useful, if this PR is not accepted

SychO9:
I like how GitHub handles deletions of repositories/orgs. You have to enter the name of the repo/org before continuing with the action. Could we maybe take a similar route? you would have a modal where you are warned and have to type the username.

imorland:
> I like how GitHub handles deletions of repositories/orgs. You have to enter the name of the repo/org before continuing with the action. Could we maybe take a similar route? you would have a modal where you are warned and have to type the username.

Yeah, I'm a great fan of how GutHub handle this. Would be happy with that 👍

davwheat:
How would this work when using nicknames and having special characters in the nick? Would be very hard to type it out then.

SychO9:
> How would this work when using nicknames and having special characters in the nick? Would be very hard to type it out then.

it should be username, not nickname. also you can copy paste too, as long as it clearly shows what you're getting yourself into.

luceos:
I prefer this solution over this PR as well. Glad about the train of thought here 💪

PR #3571, opened 2022-07-28
fix: normal logout affects all sessions

rob006:

Previously all user tokens were deleted, which logouts from all sessions. After this change logout only works for current session.

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/framework/core/src/Forum/Controller/LogOutController.php
+++ b/framework/core/src/Forum/Controller/LogOutController.php
@@ -102,11 +102,12 @@ class LogOutController implements RequestHandlerInterface
             return new HtmlResponse($view->render());
         }
 
+        $accessToken = $session->get('access_token');
         $response = new RedirectResponse($url);
 
         $this->authenticator->logOut($session);
 
-        $actor->accessTokens()->delete();
+        $actor->accessTokens()->where('token', $accessToken)->delete();
 
         $this->events->dispatch(new LoggedOut($actor));
 


davwheat:
I think that if we choose to make this change, we should have a new option in a user's settings page to "log out from all devices" for security reasons. Maybe also a count of active session tokens, too?

rob006:
I don't think that having "log out from all devices" feature is a blocker for this PR. Such feature could be handled by extension, while current logout behavior is just really weird and I doubt it was intentional.

SychO9:
If we make this change though, a user effectively has no way of logging out from other signed-in devices, which is a security concern. So the idea of first introducing a way to logout from all sessions, then make this change, makes sense. I believe this is also a feature in our security roadmap.

davwheat:
Yeah, the potential security implications is my concern for this PR, hence why I see it as blocking.

matteocontrini:
There was #2074 by @clarkwinkelmann that introduced a more advanced GUI for managing sessions. Ideally that would be the best solution, I guess.

SychO9:
Yup, I'll be bringing that back to life in the next few days as part of the security roadmap. So let's close this for now.

SychO9:
@rob006 could you please revive this now that we have #3605 merged?

rob006:
@SychO9 I restored branch, but I can't reopen this PR.

SychO9 (approved):

luceos (approved):

PR #3573, opened 2022-07-30
chore: unify JS actions into one (rewritten `flarum/action-build`)

SychO9:
**Relies on https://github.com/flarum/action-build/pull/10**

Unifies all frontend workflows into one.

**Necessity**

**Confirmed**


SychO9:
Tested on a single extension repository to make sure things are still good on that end ✅

SychO9:
Bundlewatch is disabled for now since we use a single frontend workflow. Maybe we can specifically have it only measure core dist files, or if possible all dist files at once?

davwheat:
I think disabling Bundlewatch for now is the best. We can look into re-enabling it later.

davwheat:
Rebased to (hopefully) pass PHP tests.

davwheat (approved):

SychO9:
Here goes nothing..

PR #3576, opened 2022-08-01
feat: add createTableIfNotExists migration helper

imorland:
Adds an additional `Migration` helper - `createTableIfNotExists` - self explanitory, really.

I considered modifying the existing `createTable` function, with an option 3rd param to specify if this should be conditional, but felt that a seperate, although very similar function, would be clearer in it's intended functionality. Hopefully others will agree?

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/framework/core/src/Database/Migration.php
+++ b/framework/core/src/Database/Migration.php
@@ -39,6 +39,25 @@ abstract class Migration
     }
 
     /**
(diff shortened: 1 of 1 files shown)


luceos (approved):
I actually think this makes sense.

SychO9 (approved):

PR #3578, opened 2022-08-03
perf: Add `loading="lazy"` attribute for avatars

rob006:

This enables lazy loading for avatars. It should slightly improve loading time and allows to avoid loading these images when they're hidden by custom styles.

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/framework/core/js/src/common/helpers/avatar.tsx
+++ b/framework/core/js/src/common/helpers/avatar.tsx
@@ -12,7 +12,7 @@ export interface AvatarAttrs extends ComponentAttrs {}
  */
 export default function avatar(user: User | null, attrs: ComponentAttrs = {}): Mithril.Vnode {
   attrs.className = 'Avatar ' + (attrs.className || '');
-  attrs.loading = attrs.loading || 'lazy';
+  attrs.loading ??= 'lazy';
   let content: string = '';
 
   // If the `title` attribute is set to null or false, we don't want to give the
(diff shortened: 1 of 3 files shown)


SychO9 (approved):

rafaucau:
Not all avatars should be lazy loaded, as it affects Largest Concertful Paint. This can even degrade performance.
E.g. The avatar of a logged-in user in the header and the first few posts should not have this attribute and might instead be preloaded.
https://web.dev/optimize-lcp/#optimize-the-priority-the-resource-is-given

> ⚠️Warning
>Never lazy-load your LCP image, as that will always lead to unnecessary resource load delay, and will have a negative impact on LCP.

By the way, images should have `width` and `height` attributes or CSS `aspect-ratio`.
https://web.dev/optimize-cls/#images-without-dimensions

rob006:
I don't think it will have any measurable effect on LCP due to SPA nature of Flarum - images are already the last thing that browser needs to load. 

> By the way, images should have `width` and `height` attributes or CSS `aspect-ratio`.

AFAIK all avatars have width and height set by styles.

framework/core/js/src/common/helpers/avatar.tsx
```diff
@@ -12,6 +12,7 @@ export interface AvatarAttrs extends ComponentAttrs {}
  */
 export default function avatar(user: User | null, attrs: ComponentAttrs = {}): Mithril.Vnode {
   attrs.className = 'Avatar ' + (attrs.className || '');
+  attrs.loading = attrs.loading || 'lazy';
```
davwheat:
```suggestion
  attrs.loading ??= 'lazy';
```

davwheat (commented):
I'm in agreement that LCP isn't an issue for this. Avatars are relatively small, so won't be classed as Flarum's LCP image, but I think it's worth overriding `loading` on the `UserPage`.

davwheat (approved):

PR #3584, opened 2022-08-05
feat: Add custom class for email confirmation alert

rob006:

Currently this alert uses generic classes and it is quite hard to distinguish this particular alert from any other alerts. Custom class should simplify for example adding custom styles for this alert.

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/framework/core/js/src/forum/utils/alertEmailConfirmation.js
+++ b/framework/core/js/src/forum/utils/alertEmailConfirmation.js
@@ -59,9 +59,9 @@ export default function alertEmailConfirmation(app) {
     }
   }
 
-  m.mount($('<div class="Alert--emailConfirmation"/>').insertBefore('#content')[0], {
+  m.mount($('<div class="App-notices"/>').insertBefore('#content')[0], {
     view: () => (
-      <ContainedAlert dismissible={false} controls={[<ResendButton />]}>
+      <ContainedAlert dismissible={false} controls={[<ResendButton />]} className="Alert--emailConfirmation">
         {app.translator.trans('core.forum.user_email_confirmation.alert_message', { email: <strong>{user.email()}</strong> })}
       </ContainedAlert>
     ),
--- a/framework/core/js/src/forum/utils/alertEmailConfirmation.js
+++ b/framework/core/js/src/forum/utils/alertEmailConfirmation.js
@@ -59,7 +59,7 @@ export default function alertEmailConfirmation(app) {
     }
   }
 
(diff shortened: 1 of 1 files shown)


davwheat (commented):
I haven't looked as I'm on my phone, but is it possible to put the class attribute on the `ContainedAlert` JSX element instead? I feel that makes more sense as I think the way this alert is added right now could do with some improvement, to be honest.

rob006:
From styling perspective it is better to add this class on the most outermost wrapper. If we want to add this on `ContainedAlert`, then IMO we should get rid of this div wrapper.

davwheat:
Maybe we could add a class to both? If we restructure how this alert is added, then I don't want to accidentally drop this class and break some styling.

framework/core/js/src/forum/utils/alertEmailConfirmation.js
```diff
@@ -59,7 +59,7 @@ export default function alertEmailConfirmation(app) {
     }
   }
 
-  m.mount($('<div/>').insertBefore('#content')[0], {
+  m.mount($('<div class="Alert--emailConfirmation"/>').insertBefore('#content')[0], {
     view: () => (
       <ContainedAlert dismissible={false} controls={[<ResendButton />]}>
```
SychO9: Agree with @davwheat. How about:

```suggestion
  m.mount($('<div class="App-notices"/>').insertBefore('#content')[0], {
    view: () => (
      <ContainedAlert dismissible={false} controls={[<ResendButton />]} className="Alert--emailConfirmation">
```

davwheat: This seems like a better approach, imo.

SychO9: @rob006 this should do it for your use case while still sticking to our conventions, any objection to making the change and merging?

rob006: HTML structure is fine, but I'm not sure how it would scale. If we would have another type of alert, does it mean we would have multiple `div.App-notices` containers?

SychO9: No there should always be the one only `App-notices` container, unless an extension is really set on hacking that.

For 2.0 it will most likely be made extensible to easily have multiple possible notices as we've seen it to be a plausible use case.

SychO9 (approved):

davwheat (approved):

PR #3587, opened 2022-08-06
feat: access tokens user management UI

SychO9:
**Picks up from #2651 and #2074**

Introduces the following:
* UI to manage access tokens (actor only).
  * Revoke developer tokens (only the actor can delete his own and no one else).
  * Create developer tokens (based on permission).
  * View developer token value.
  * See other active sessions (token values are not exposed from the API for sessions/system-created tokens).
    * Terminate a remote active session.
    * Terminate all other sessions (not including the current session, user should log out).

After reading the above behavior ^, any opinions if anything should be done differently? such as:
* Should admins/permission-given-users be able to terminate/revoke other users' sessions/dev-tokens?
* Is this large enough to move to a separate bundled extension? (I personally think not).

Checkout integration tests method names for further behavior understanding.

**Video**
https://user-images.githubusercontent.com/20267363/183262689-0de50d68-91db-48d6-a5f9-c252d789c553.mp4

**Screenshot**
Screenshot from 2022-08-06 19-53-36

**Necessity**

**Confirmed**


clarkwinkelmann (dismissed):
It looks really good!

I tried it out locally and didn't find anything major.

I would suggest adding a confirmation popup when Revoking a developer token. Maybe that would be useful for Terminate and Terminate all others as well.

In the developer tokens part I see the text says "on [...]" and then the string ends. Looking at the source it looks like it should describe the device. I did my test with cURL (UA `curl/7.68.0`). I think there should be some fallback "N/A" value or the raw UA?

image

I would also suggest replacing the hidden token value by a hard-coded number of stars. Showing `****` would probably be enough, no need for such a long string.

And finally I'd suggest wrapping the revealed token value in `<code></code>` to make it stand out better and potentially make it easier to select in some browsers (?)

PS: video in first post doesn't play in Firefox. But works in Chrome

SychO9:
Will do! thanks for the review!

tankerkiller125:
> - Should admins/permission-given-users be able to terminate/revoke other users' sessions/dev-tokens?

In my view the answer to this is yes, if the system get's compromised or something it would be good for it to be possible for the admin to revoke all sessions and/or tokens. HOWEVER, whether this should be a individual user view where the admin can close just a single users sessions, or just a button on the admin side that wipes all of the sessions is another matter.

I do think that token removal should be on an individual user level though in terms of admin deletion. As my view of it is that it's a way for admins to moderate tokens (if say someone is abusing them) 

> - Is this large enough to move to a separate bundled extension? (I personally think not).

It should not be a separate extension, I view it as a critical security feature that should be part of core.

SychO9:
> In my view the answer to this is yes, if the system get's compromised or something it would be good for it to be possible for the admin to revoke all sessions and/or tokens. HOWEVER, whether this should be a individual user view where the admin can close just a single users sessions, or just a button on the admin side that wipes all of the sessions is another matter.

> I do think that token removal should be on an individual user level though in terms of admin deletion. As my view of it is that it's a way for admins to moderate tokens (if say someone is abusing them)

Then maybe we should create a `moderate other users' sessions/tokens` to allow a moderator group access to a user's list of sessions/tokens (but make sure to hide the actual token value of course from non-actors). We'd allow viewing basic information and allow terminating/revoking. Does that sound good?

tankerkiller125:
That seems like a reasonable idea, I wouldn't want anyone seeing the values other than the users anyway. Most sites only allow the viewing of a token once anyway even for the users.

SychO9:
Updated with mentioned changes + now there is a `moderate access tokens permission` to be able to see active sessions/dev tokens of other users and terminate/revoke them (without being able to see the actual token value). There are tests that should cover all scenatios.

luceos:
> just a button on the admin side that wipes all of the sessions is another matter

I think we ought to add this to the roadmap. A security tab in the admin is probably a good idea. You don't want to make protecting your community complicated. Potentially this could be a cli command too, it would reduce the "whoops I pressed this button, what does it do again" scenario, while still keeping it in reach.

framework/core/js/src/forum/components/SecurityPage.tsx
```diff
@@ -0,0 +1,80 @@
...
+import listItems from '../../common/helpers/listItems';
+import type Mithril from "mithril";
+import extractText from "../../common/utils/extractText";
+import AccessTokensList from "./AccessTokensList";
+
+/**
+ * The `SecurityPage` component displays the user's security control panel, in
+ * the context of their user profile.
+ */
+export default class SecurityPage<CustomAttrs extends IUserPageAttrs = IUserPageAttrs> extends UserPage<CustomAttrs> {
```
askvortsov1: Maybe `UserSecurityPage`? I could see us / extensions having security pages for other systems in the future

framework/core/js/src/forum/components/SecurityPage.tsx
```diff
@@ -0,0 +1,80 @@
...
+  /**
+   * Build an item list for the user's settings controls.
+   */
+  settingsItems() {
+    const items = new ItemList<Mithril.Children>();
+
+    ['accessToken', 'session'].forEach((section) => {
+      const sectionName = `${section}Items` as 'accessTokenItems' | 'sessionItems';
+      // Camel-case to snake-case
+      const sectionLocale = section.replace(/[A-Z]/g, letter => `_${letter.toLowerCase()}`);
```
askvortsov1: This feels like it should be a util. Also, whatcha got against camels? :camel:

framework/core/js/src/common/components/DataSegment.tsx
```diff
@@ -0,0 +1,18 @@
+import Component, {ComponentAttrs} from "../Component";
+import type Mithril from "mithril";
+
+export interface IDataSegmentAttrs extends ComponentAttrs {
+  label: Mithril.Children;
+  value: Mithril.Children;
+}
+
+export default class DataSegment<CustomAttrs extends IDataSegmentAttrs = IDataSegmentAttrs> extends Component<CustomAttrs> {
```
askvortsov1: The naming of this component feels ambiguous. At the least, we should add a docblock discussing intended usage

framework/core/js/src/common/helpers/listItems.tsx
```diff
@@ -25,7 +25,7 @@ type ModdedChildren = ModdedChild | ModdedChildArray;
  */
 export type ModdedChildrenWithItemName = ModdedChildren & { itemName?: string };
 
-function isVnode(item: ModdedChildren): item is Mithril.Vnode {
+export function isVnode(item: ModdedChildren): item is Mithril.Vnode {
   return typeof item === 'object' && item !== null && 'tag' in item;
 }
```
askvortsov1: If we want to reuse this guard, it should be its own util

framework/core/js/src/common/models/AccessToken.ts
```diff
@@ -25,8 +25,10 @@ export default class AccessToken extends Model {
   lastUserAgent() {
     return Model.attribute<string>('lastUserAgent').call(this);
   }
-
   isCurrent() {
     return Model.attribute<boolean>('isCurrent').call(this);
   }
```
askvortsov1, line 30: Not relevant to this file, but looks like some dist recompilation changes crept in here

framework/core/js/src/forum/components/AccessTokensList.tsx
```diff
@@ -105,17 +115,35 @@ export default class AccessTokensList<CustomAttrs extends IAccessTokensListAttrs
...
+        onclick={() => this.revoke(token)}>
         {app.translator.trans(`core.forum.security.${deleteKey}`)}
       </Button>
     );
 
     return items;
   }
+
+  revoke(token: AccessToken) {
+    this.loading[token.id()!] = true;
```
askvortsov1, line 171: Do we need a redraw after this, or will the click automatically trigger one?

framework/core/js/src/forum/components/AccessTokensList.tsx
```diff
@@ -105,17 +115,35 @@ export default class AccessTokensList<CustomAttrs extends IAccessTokensListAttrs
...
+      'session': 'terminate_session',
       'token': 'revoke_access_token'
     }[this.attrs.type];
 
     items.add(
       'revoke',
-      <Button className="Button Button--danger">
+      <Button
+        className="Button Button--danger"
+        disabled={token.isCurrent()}
```
askvortsov1: Good idea, but we should add a tooltip of some sort to reinforce why this button doesn't work, in addition to the "Current Active Session" text

framework/core/src/Api/Controller/DeleteAccessTokenController.php
```diff
@@ -32,8 +32,11 @@ protected function delete(ServerRequestInterface $request)
 
         $token = AccessToken::query()->findOrFail($id);
 
+        /** @var Session $session */
+        $session = $request->getAttribute('session');
+
         // Current session should only be terminated through logout.
-        if ($token->token === $request->getAttribute('session')->token()) {
+        if ($session && $token->token === $session->get('access_token')) {
```
askvortsov1, line 39: I can't wait till we can fully move to 8.0!

https://wiki.php.net/rfc/nullsafe_operator

@SychO9 is there a PHPStan level that will automatically detect stuff like this?

framework/core/src/Http/Access/AccessTokenPolicy.php
```diff
@@ -17,8 +17,8 @@ class AccessTokenPolicy extends AbstractPolicy
 {
     public function revoke(User $actor, AccessToken $token)
     {
-        if ($token->user_id === $actor->id) {
-            return $this->allow();
-        }
+        // NOTE: even admins can't delete or view other user's tokens,
+        // that's why we deny otherwise.
+        return $token->user_id === $actor->id ? $this->allow() : $this->deny();
```
askvortsov1: I can imagine a use case for logging out a user. Admins can reset passwords / access the DB, why this particular behavior here?

framework/core/tests/integration/api/access_tokens/DeleteTest.php
```diff
@@ -0,0 +1,128 @@
...
+    public function user_cannot_delete_other_users_tokens(int $authenticatedAs, array $canDeleteIds)
+    {
+        foreach ($canDeleteIds as $id) {
+            $response = $this->send(
+                $this->request('DELETE', "/api/access-tokens/$id", compact('authenticatedAs'))
+            );
+
+            $this->assertEquals(404, $response->getStatusCode());
+        }
+    }
```
askvortsov1: We should consider splitting the admin case out into a separate test, so that it's easier to tweak / understand behavior.

Also, data providers are a cool new feature1

php-packages/testing/src/integration/BuildsHttpRequests.php
```diff
@@ -49,7 +49,9 @@ protected function requestAsUser(Request $req, int $userId): Request
             'type' => 'session'
         ]);
 
-        return $req->withAddedHeader('Authorization', "Token {$token}");
+        return $req
+            ->withAddedHeader('Authorization', "Token {$token}")
+            ->withAttribute('tests_token', $token);
```
askvortsov1, line 55: Can we add a comment explaining why this change was made? Guessing it's so that we can read the added token in tests

framework/core/js/src/forum/components/AccessTokensList.tsx
```diff
@@ -26,7 +26,8 @@ export default class AccessTokensList<CustomAttrs extends IAccessTokensListAttrs
   oninit(vnode: Mithril.Vnode<CustomAttrs, this>) {
     super.oninit(vnode);
 
-    this.tokens = this.attrs.tokens;
+    // Sort by current first.
```
askvortsov1: Maybe "Show current token first"?

framework/core/locale/core.yml
```diff
@@ -452,11 +452,12 @@ core:
       created: Created
       hide_access_token: Hide Token
       last_activity: Last activity
+      terminate_all_other_sessions: Terminate all other sessions
       terminate_session: Terminate
       title: => core.ref.security
       revoke_access_token: Revoke
       session_heading: Active Sessions
-      session_terminated: Session Terminated.
+      session_terminated: "{count, plural, one {Session terminated} other {Sessions terminated}}."
```
askvortsov1: Should we say how many?

framework/core/js/src/forum/components/SecurityPage.tsx
```diff
@@ -100,4 +115,21 @@ export default class SecurityPage<CustomAttrs extends IUserPageAttrs = IUserPage
...
+
+    return app.request({
+      method: 'DELETE',
+      url: app.forum.attribute('apiUrl') + '/sessions',
+    }).then(() => {
+      this.loading = false;
+      this.tokens = this.tokens!.filter((token) => !token.isSessionToken() || token.isCurrent());
+      app.alerts.show(
+        { type: 'success' },
+        app.translator.trans('core.forum.security.session_terminated', { count: 2 })
```
askvortsov1: Why is 2 hardcoded here? Can't we count how many have actually been terminated?

framework/core/migrations/2022_08_06_000000_change_access_tokens_last_activity_at_to_nullable.php
```diff
@@ -0,0 +1,25 @@
...
+ * LICENSE file that was distributed with this source code.
+ */
+
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Schema\Builder;
+
+return [
+    'up' => function (Builder $schema) {
+        $schema->table('access_tokens', function (Blueprint $table) {
+            $table->dateTime('last_activity_at')->nullable()->change();
```
askvortsov1, line 16: Why?

framework/core/migrations/2022_08_06_000000_change_access_tokens_last_activity_at_to_nullable.php
```diff
@@ -0,0 +1,25 @@
...
+        });
+    },
+
+    'down' => function (Builder $schema) {
+        $schema->table('access_tokens', function (Blueprint $table) {
+            $table->dateTime('last_activity_at')->nullable(false)->change();
```
askvortsov1: This feels like it'll explode and cause an error: if there are null entries already, what should happen to them?

framework/core/src/Api/Controller/AbstractSerializeController.php
```diff
@@ -122,6 +123,10 @@ public function handle(ServerRequestInterface $request): ResponseInterface
             }
         }
 
+        if (! $this->serializer) {
+            throw new InvalidArgumentException("Serializer required for controller: " . static::class);
+        }
+
```
askvortsov1: I really wish we had good support for splitting commits / stacked PRs. This would make sense in a separate commit.

framework/core/src/Api/Controller/CreateAccessTokenController.php
```diff
@@ -0,0 +1,69 @@
...
+        ])->validate();
+
+        $token = DeveloperAccessToken::generate($actor->id);
+
+        $token->title = $title;
+        $token->last_activity_at = null;
+
+        $token->save();
+
+        // $this->events->dispatch(new DeveloperTokenCreated($token));
```
askvortsov1: Why commented?

framework/core/js/src/forum/components/AccessTokensList.tsx
```diff
@@ -157,6 +157,6 @@ export default class AccessTokensList<CustomAttrs extends IAccessTokensListAttrs
   }
 
   tokenValueDisplay(token: AccessToken) {
-    return this.showingTokens[token.id()!] ? token.token() : token.token().replace(/./g, '*');
+    return this.showingTokens[token.id()!] ? token.token() : token.token()!.replace(/./g, '*');
```
askvortsov1: Why not a nullsafe instead? `token.token()?.replace...`

framework/core/js/src/forum/components/AccessTokensList.tsx
```diff
@@ -157,6 +157,6 @@ export default class AccessTokensList<CustomAttrs extends IAccessTokensListAttrs
   }
 
   tokenValueDisplay(token: AccessToken) {
```
askvortsov1: Let's typehint the return, since it's a public API method?

framework/core/src/Http/Filter/UserFilter.php
```diff
@@ -0,0 +1,32 @@
...
+ * For detailed copyright and license information, please view the
+ * LICENSE file that was distributed with this source code.
+ */
+
+namespace Flarum\Http\Filter;
+
+use Flarum\Filter\FilterInterface;
+use Flarum\Filter\FilterState;
+
+class UserFilter implements FilterInterface
```
askvortsov1, line 21: Feels a bit odd that nothing in this file mentions / relates to access tokens

askvortsov1 (dismissed):
Overall, fantastic work! A few UI comments:

- The "Terminate" button should show a tooltip if disabled, to make it 100% clear that the token in question is currently active. 
- Last Activity shouldn't be used if null (although I have mixed feelings about making it nullable): image
- "View" vs "Hide Token" is inconsistent; it should probably be "View Token".
- We might want to move password reset here, in a separate PR

framework/core/js/src/forum/components/AccessTokensList.tsx
davwheat: Browser events automatically trigger redraws :)

framework/core/js/src/forum/components/AccessTokensList.tsx
davwheat: I would rather have an indicator next to the token showing its in use, like our online user indicator? (Or in addition to the tooltip.)

Maybe similar to GitHub's sessions page:

image

framework/core/js/src/common/helpers/listItems.tsx
SychO9: you might have an outdated branch locally or something, the export was removed after it was accidentally added before your review 🤔

framework/core/js/src/forum/components/SecurityPage.tsx
SychO9: Current naming made sense when onl the logged in user could access (like the `SettingsPage`), but now that mods/admins can you're right. I looked at other user page componenets and the pattern seems to be `PostsUserPage` `DiscussionsUserPage` ..etc so perhaps `SecurityUserPage` ? (*though it sounds worse than `UserSecurityPage`*)

framework/core/js/src/forum/components/AccessTokensList.tsx
SychO9: I think even with an indicator a tooltip would still make sense.

framework/core/src/Http/Access/AccessTokenPolicy.php
SychO9: somehow you're looking at old code 🤔 cause this behavior was changed (and comment dropped) after me and matt's covo above before your review

framework/core/src/Api/Controller/CreateAccessTokenController.php
SychO9: Same, this was changed before lol

framework/core/migrations/2022_08_06_000000_change_access_tokens_last_activity_at_to_nullable.php
SychO9: My thinking is, since now developer tokens *could* be created by the end-user now, they wouldn't be instantly used. So the last activity *should* be null to indicate that it was never used until it actually is.

framework/core/migrations/2022_08_06_000000_change_access_tokens_last_activity_at_to_nullable.php
SychO9: That's true 🤔 It must remain null, good catch!

framework/core/src/Api/Controller/DeleteAccessTokenController.php
SychO9: not as far as I can tell unfotunately

framework/core/js/src/forum/components/AccessTokensList.tsx
SychO9: surprisingly though, seems like tooltips don't work on disabled buttons. Logged an issue for now.

framework/core/js/src/common/components/DataSegment.tsx
SychO9: Yea, not sure what a better naming would be.

framework/core/src/Api/Controller/AbstractSerializeController.php
SychO9: We can split this

framework/core/js/src/forum/components/AccessTokensList.tsx
SychO9: old code 👼🏼

framework/core/src/Http/Filter/UserFilter.php
SychO9: If it wasn't a breaking change, it would make sense to move access token domain to its own namespace.

framework/core/migrations/2022_08_06_000000_change_access_tokens_last_activity_at_to_nullable.php
tankerkiller125: Sycho's thought here is the correct one. The last activity should always be null until it's actually used. That's a standard practice you'll find across almost all applications/services.

framework/core/js/src/common/components/DataSegment.tsx
```diff
@@ -0,0 +1,24 @@
...
+ * A generic component for displaying a label and value inline.
+ * Created to avoid reinventing the wheel.
+ *
+ * `label: value`
+ */
+export default class DataSegment<CustomAttrs extends IDataSegmentAttrs = IDataSegmentAttrs> extends Component<CustomAttrs> {
+  view(vnode: Mithril.Vnode<CustomAttrs, this>): Mithril.Children {
+    return (
+      <div className="DataSegment">
+        <div className="DataSegment-label">{this.attrs.label}:</div>
```
davwheat: I think it might be valuable having the `:` translated here, in case some locales often use a different separator to a colon.

Preferably the translation should be passed the label in full, rather than just containing a colon.

framework/core/js/src/forum/components/AccessTokensList.tsx
```diff
@@ -0,0 +1,182 @@
...
+import app from '../app';
+import icon from '../../common/helpers/icon';
+import uaParser from 'ua-parser-js';
+import Button from '../../common/components/Button';
+import humanTime from '../../common/helpers/humanTime';
+import ItemList from '../../common/utils/ItemList';
+import DataSegment from '../../common/components/DataSegment';
+import extractText from '../../common/utils/extractText';
+import classList from '../../common/utils/classList';
+import Tooltip from '../../common/components/Tooltip';
```
davwheat: Just sorting the imports a bit

```suggestion
import app from '../app';
import Component, { ComponentAttrs } from '../../common/Component';
import icon from '../../common/helpers/icon';
import uaParser from 'ua-parser-js';
import Button from '../../common/components/Button';
import humanTime from '../../common/helpers/humanTime';
import ItemList from '../../common/utils/ItemList';
import DataSegment from '../../common/components/DataSegment';
import extractText from '../../common/utils/extractText';
import classList from '../../common/utils/classList';
import Tooltip from '../../common/components/Tooltip';
import type Mithril from 'mithril';
import type AccessToken from '../../common/models/AccessToken';
```

framework/core/js/src/forum/components/AccessTokensList.tsx
```diff
@@ -0,0 +1,182 @@
...
+
+export interface IAccessTokensListAttrs extends ComponentAttrs {
+  tokens: AccessToken[];
+  type: 'session' | 'token';
```
davwheat: Would it make more sense to call it a developer token rather than just token? Technically session tokens are also "tokens" in this sense

framework/core/js/src/forum/components/AccessTokensList.tsx
```diff
@@ -0,0 +1,182 @@
...
+  tokenView(token: AccessToken): Mithril.Children {
+    return (
+      <div
+        className={classList('AccessTokensList-item', {
+          'AccessTokensList-item--active': token.isCurrent(),
+        })}
+      >
+        <div className="AccessTokensList-item-icon">{icon(this.attrs.icon || 'fas fa-key')}</div>
+        <div className="AccessTokensList-item-info">{this.tokenInfoItems(token).toArray()}</div>
+        <div className="AccessTokensList-item-actions">{this.tokenActionItems(token).toArray()}</div>
```
davwheat: Could we ItemList this, just in case? You never know what an extension might want to do :P

framework/core/js/src/forum/components/AccessTokensList.tsx
```diff
@@ -0,0 +1,182 @@
...
+      <div className="AccessTokensList-item-lastActivityAt">
+        <DataSegment
+          label={app.translator.trans('core.forum.security.last_activity')}
+          value={
+            token.lastActivityAt()
+              ? [
+                  humanTime(token.lastActivityAt()),
+                  token.lastIpAddress() && [' — ', token.lastIpAddress()],
+                  this.attrs.type === 'token' && token.lastUserAgent() && [' — ', <span className="AccessTokensList-item-title-sub">{device}</span>],
+                ]
```
davwheat:
```suggestion
              ? (
                  <>
                    {humanTime(token.lastActivityAt())}
                    {token.lastIpAddress() && ` — ${token.lastIpAddress()}`}
                    {this.attrs.type === 'token' && token.lastUserAgent() && <> — <span className="AccessTokensList-item-title-sub">{device}</span></>}
                  </>
                )
```

framework/core/js/src/forum/components/AccessTokensList.tsx
```diff
@@ -0,0 +1,182 @@
...
+    this.loading[token.id()!] = true;
+
+    return token.delete().then(() => {
+      this.loading[token.id()!] = false;
+      this.attrs.ondelete && this.attrs.ondelete(token);
+      const key = this.attrs.type === 'session' ? 'session_terminated' : 'token_revoked';
+      app.alerts.show({ type: 'success' }, app.translator.trans(`core.forum.security.${key}`, { count: 1 }));
+      m.redraw();
+    });
+  }
```
davwheat: I think an async function here is cleaner.

```suggestion
  async revoke(token: AccessToken) {
    if (!confirm(extractText(app.translator.trans('core.forum.security.revoke_access_token_confirmation')))) return;

    this.loading[token.id()!] = true;

    await token.delete()

    this.loading[token.id()!] = false;
    this.attrs.ondelete?.(token);
    const key = this.attrs.type === 'session' ? 'session_terminated' : 'token_revoked';
    app.alerts.show({ type: 'success' }, app.translator.trans(`core.forum.security.${key}`, { count: 1 }));

    m.redraw();
  }
```

framework/core/js/src/forum/components/NewAccessTokenModal.tsx
```diff
@@ -0,0 +1,67 @@
+import Modal, { IInternalModalAttrs } from '../../common/components/Modal';
+import Mithril from 'mithril';
+import Button from '../../common/components/Button';
+import app from '../app';
+import AccessToken from '../../common/models/AccessToken';
+import Stream from '../../common/utils/Stream';
+import { SaveAttributes } from '../../common/Model';
```
davwheat: Sort, and use type imports

```suggestion
import app from '../app';
import Modal, { IInternalModalAttrs } from '../../common/components/Modal';
import Button from '../../common/components/Button';
import Stream from '../../common/utils/Stream';
import type AccessToken from '../../common/models/AccessToken';
import type { SaveAttributes } from '../../common/Model';
import type Mithril from 'mithril';
```

framework/core/js/src/forum/components/NewAccessTokenModal.tsx
```diff
@@ -0,0 +1,67 @@
...
+  title(): Mithril.Children {
+    return app.translator.trans('core.forum.security.new_access_token_modal.title');
+  }
+
+  content(): Mithril.Children {
+    return (
+      <div className="Modal-body">
+        <div className="Form Form--centered">
+          <div className="Form-group">
+            <input
```
davwheat: Missing label? If we want to use the modal title, we should add an `aria-labelled-by`, or we can create a visually-hidden (`.sr-only`) label element wrapping this.

framework/core/js/src/forum/components/NewAccessTokenModal.tsx
```diff
@@ -0,0 +1,67 @@
...
+            <Button className="Button Button--primary Button--block" type="submit" loading={this.loading}>
+              {app.translator.trans('core.forum.security.new_access_token_modal.submit_button')}
+            </Button>
+          </div>
+        </div>
+      </div>
+    );
+  }
+
+  submitData(): SaveAttributes {
```
davwheat, line 43: We should probably make `SaveAttributes` a generic to pull the attrs from the Model definition 🙈 

Not really for this PR though

framework/core/js/src/forum/components/NewAccessTokenModal.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+    app.store
+      .createRecord<AccessToken>('access-tokens')
+      .save(this.submitData())
+      .then((token) => {
+        this.attrs.onsuccess(token);
+        app.modal.close();
+      })
+      .finally(this.loaded.bind(this));
+  }
```
davwheat, line 63: I think this is also cleaner as async but we have to maintain the superclass API :(

We could define and execute an async anonymous function inside it, but I think that would look a bit ugly.

framework/core/js/src/forum/components/UserPage.tsx
```diff
@@ -159,6 +159,20 @@ export default class UserPage<CustomAttrs extends IUserPageAttrs = IUserPageAttr
       );
     }
 
+    if (app.session.user!.id() === user.id() || app.forum.attribute<boolean>('canModerateAccessTokens')) {
+      if (app.session.user!.id() !== user.id()) {
+        items.add('separator', <Separator />, -90);
```
davwheat: We should try and be more specific to prevent clashing of keys with extensions

```suggestion
        items.add('security-separator', <Separator />, -90);
```

framework/core/js/src/forum/components/UserSecurityPage.tsx
```diff
@@ -0,0 +1,190 @@
...
+import FieldSet from '../../common/components/FieldSet';
+import listItems from '../../common/helpers/listItems';
+import type Mithril from 'mithril';
+import extractText from '../../common/utils/extractText';
+import AccessTokensList from './AccessTokensList';
+import AccessToken from '../../common/models/AccessToken';
+import LoadingIndicator from '../../common/components/LoadingIndicator';
+import Button from '../../common/components/Button';
+import NewAccessTokenModal from './NewAccessTokenModal';
+import camelCaseToSnakeCase from '../../common/utils/camelCaseToSnakeCase';
```
davwheat:
```suggestion
import app from '../../forum/app';
import UserPage, { IUserPageAttrs } from './UserPage';
import ItemList from '../../common/utils/ItemList';
import FieldSet from '../../common/components/FieldSet';
import listItems from '../../common/helpers/listItems';
import extractText from '../../common/utils/extractText';
import AccessTokensList from './AccessTokensList';
import LoadingIndicator from '../../common/components/LoadingIndicator';
import Button from '../../common/components/Button';
import NewAccessTokenModal from './NewAccessTokenModal';
import camelCaseToSnakeCase from '../../common/utils/camelCaseToSnakeCase';
import type AccessToken from '../../common/models/AccessToken';
import type Mithril from 'mithril';
```

framework/core/js/src/forum/components/UserSecurityPage.tsx
```diff
@@ -0,0 +1,190 @@
...
+  developerTokensItems() {
+    const items = new ItemList<Mithril.Children>();
+
+    items.add(
+      'accessTokenList',
+      this.tokens === null ? (
+        <LoadingIndicator />
+      ) : (
+        <AccessTokensList
+          key={this.tokens.length}
```
davwheat: Do we need this key?

framework/core/js/src/forum/components/UserSecurityPage.tsx
```diff
@@ -0,0 +1,190 @@
...
+  sessionsItems() {
+    const items = new ItemList<Mithril.Children>();
+
+    items.add(
+      'sessionsList',
+      this.tokens === null ? (
+        <LoadingIndicator />
+      ) : (
+        <AccessTokensList
+          key={this.tokens.length}
```
davwheat: Same here

framework/core/js/src/forum/components/UserSecurityPage.tsx
```diff
@@ -0,0 +1,190 @@
...
+    );
+
+    if (this.user!.id() === app.session.user!.id()) {
+      items.add(
+        'terminateAllOtherSessions',
+        <Button
+          className="Button"
+          onclick={this.terminateAllOtherSessions.bind(this)}
+          loading={this.loading}
+          disabled={!this.tokens?.find((token) => token.isSessionToken() && !token.isCurrent())}
```
davwheat: Is it worth adding a tooltip to explain why this is disabled, like we do for the current session button?

framework/core/less/forum/UserSecurityPage.less
```diff
@@ -0,0 +1,80 @@
...
+
+  > ul > li {
+    margin-bottom: 25px;
+  }
+}
+
+.AccessTokensList {
+  display: flex;
+  flex-direction: column;
+  gap: 1px;
```
davwheat: Flex gap isn't supported on iOS 14.0 and below. We should probably use:

```css
> *:not(:first-child) {
  margin-left: 1px;
}
```

framework/core/less/forum/UserSecurityPage.less
```diff
@@ -0,0 +1,80 @@
...
+      &-sub {
+        font-style: italic;
+      }
+    }
+
+    &-actions {
+      display: flex;
+      align-items: center;
+      margin-left: auto;
+      gap: 8px;
```
davwheat:
```suggestion
      > *:not(:first-child) {
        margin-left: 8px;
      }
```

framework/core/less/forum/UserSecurityPage.less
```diff
@@ -0,0 +1,80 @@
...
+}
+
+@media @phone {
+  .AccessTokensList {
+    gap: 8px;
+
+    &-item {
+      flex-wrap: wrap;
+      gap: 16px;
+      padding: 16px;
```
davwheat:
```suggestion
      padding: 16px;
      
      > *:not(:first-child) {
        margin-left: 16px;
      }
```

framework/core/less/forum/UserSecurityPage.less
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  &--empty {
+    color: var(--control-color);
+  }
+}
+
+@media @phone {
+  .AccessTokensList {
+    gap: 8px;
```
davwheat:
```suggestion
    > *:not(:first-child) {
      margin-left: 8px;
    }
```

davwheat (changes requested):

framework/core/less/forum/UserSecurityPage.less
SychO9: why must you destroy my hopes and dreams

davwheat: just sharing the safari love 🥰🥰

framework/core/js/src/forum/components/UserSecurityPage.tsx
SychO9: mithril doesn't redraw the list when the tokens change (revoke/create). By using a key with the array length it can tell that it needs to be redrawn.

Not sure why this is necessary here though, I might be missing something about this.

davwheat: I'm not sure either... I can't see any reason why the key would be needed here.

I don't think the key being the array length is a good idea, though, since if (for some reason) the number of tokens stayed the same, but one of them had changed to another model or was edited in some way, we wouldn't trigger a redraw.

I can't think of a better option which doesn't have the same limitation, though.

SychO9: fixed: https://github.com/flarum/framework/pull/3587/commits/61ceb8c8c84f204e5ceb7127244ea427e05e26a8

framework/core/js/src/common/utils/camelCaseToSnakeCase.ts
```diff
@@ -0,0 +1,3 @@
+export default function camelCaseToSnakeCase(str: string): string {
+  return str.replace(/[A-Z]/g, (letter) => `_${letter.toLowerCase()}`);
+}
```
davwheat: Would this be better in our `string` utils file?

imorland (approved):

framework/core/js/src/forum/components/AccessTokensList.tsx
```diff
@@ -0,0 +1,194 @@
+import app from '../app';
+import Component, { ComponentAttrs } from '../../common/Component';
+import icon from '../../common/helpers/icon';
+import uaParser from 'ua-parser-js';
```
davwheat: I am having concerns about our bundle size nowadays. Could we potentially lazy-load this?

framework/core/js/src/forum/components/AccessTokensList.tsx
```diff
@@ -0,0 +1,194 @@
...
+          {token.isCurrent()
+            ? [' — ', <span className="AccessTokensList-item-title-sub">{app.translator.trans('core.forum.security.current_active_session')}</span>]
+            : null}
+        </div>
+      );
+    } else {
+      items.add(
+        'title',
+        <div className="AccessTokensList-item-title">
+          <span className="AccessTokensList-item-title-main">{[token.title() || '/', token.token() && [' — ', this.tokenValueDisplay(token)]]}</span>
```
davwheat: I don't understand what the ` || '/'` is for? It seems like this is some fallback value for the token name, but I don't think that fallback makes much sense if I understand it correctly.

This entire inline JSX expression seems rather complex. Could we pull it out into a separate method like `generateTokenTitle` and call that here instead?

Also, I think we should make it (including the separating dash) translatable as I think some cultures and languages don't really use dashes.

Something like this?

```ts
generateTokenTitle(token: Token) {
  const name = token.title() || app.translator.translate('core.forum.security.token_title_placeholder');
  const value = this.tokenValueDisplay(token);

  return app.translator.translate('core.forum.security.token_item_title', { name, value });
}
```

framework/core/js/src/forum/components/UserPage.tsx
```diff
@@ -159,6 +159,20 @@ export default class UserPage<CustomAttrs extends IUserPageAttrs = IUserPageAttr
       );
     }
 
+    if (app.session.user!.id() === user.id() || app.forum.attribute<boolean>('canModerateAccessTokens')) {
+      if (app.session.user!.id() !== user.id()) {
+        items.add('security-separator', <Separator />, -90);
+      }
```
davwheat:
```suggestion
    const managingOwnTokens = app.session.user!.id() === user.id();

    if (managingOwnTokens || app.forum.attribute<boolean>('canModerateAccessTokens')) {
      if (!managingOwnTokens) {
        items.add('security-separator', <Separator />, -90);
      }
```

framework/core/js/src/forum/components/UserSecurityPage.tsx
```diff
@@ -0,0 +1,195 @@
...
+        this.state.setLoading(false);
+
+        // Count terminated sessions first.
+        const count = this.state.getOtherSessionTokens().length;
+
+        this.state.removeOtherSessionTokens();
+        app.alerts.show({ type: 'success' }, app.translator.trans('core.forum.security.session_terminated', { count }));
+        m.redraw();
+      });
+  }
```
davwheat:
```suggestion
  async terminateAllOtherSessions() {
    if (!confirm(extractText(app.translator.trans('core.forum.security.terminate_all_other_sessions_confirmation')))) return;

    this.state.setLoading(true);

    try {
      await app.request({
        method: 'DELETE',
        url: app.forum.attribute('apiUrl') + '/sessions',
      });
    } catch (error) {
      this.state.setLoading(false);
      app.alerts.show({ type: 'error' }, app.translator.trans('core.forum.security.session_termination_failed'));
      return;
    }

    this.state.setLoading(false);

    // Count terminated sessions first.
    const count = this.state.getOtherSessionTokens().length;

    this.state.removeOtherSessionTokens();
    app.alerts.show({ type: 'success' }, app.translator.trans('core.forum.security.session_terminated', { count }));

    m.redraw();
  }
```

framework/core/locale/core.yml
```diff
@@ -461,6 +463,33 @@ core:
...
+      new_access_token_button: => core.ref.new_token
+      new_access_token_modal:
+        submit_button: Create Token
+        title: => core.ref.new_token
+        title_placeholder: Title
+      empty_text: It looks like there is nothing to see here.
+      revoke_access_token: Revoke
+      revoke_access_token_confirmation: => core.ref.generic_confirmation_message
+      sessions_heading: Active Sessions
+      session_terminated: "{count, plural, one {Session terminated} other {# Sessions terminated}}."
```
davwheat, line 486:
```suggestion
      session_terminated: "{count, plural, one {Session terminated} other {# Sessions terminated}}."
      session_termination_failed: "An error occurred while terminating your sessions."
```

framework/core/src/Api/Controller/DeleteAccessTokenController.php
```diff
@@ -0,0 +1,52 @@
...
+use Illuminate\Support\Arr;
+use Laminas\Diactoros\Response\EmptyResponse;
+use Psr\Http\Message\ServerRequestInterface;
+
+class DeleteAccessTokenController extends AbstractDeleteController
+{
+    /**
+     * {@inheritdoc}
+     */
+    protected function delete(ServerRequestInterface $request)
```
davwheat, line 26: Is there any reason we don't use the Command and CommandHandler system for this like we do with most other models?

davwheat (changes requested):
This is amazing. Just a couple of things, really.

Also, I think the frontend additions need to be added to their respective `compat `files?

framework/core/js/src/forum/components/UserSecurityPage.tsx
SychO9: First is more extensible though, since it returns the request promise

framework/core/src/Api/Controller/DeleteAccessTokenController.php
SychO9: it's not a must, and this is due to the fact the access token was created on a simple controller in the first place. Wecan improve this in the future if we wish.

framework/core/js/src/forum/components/AccessTokensList.tsx
SychO9: good point, switched to parsing the UA on the backend side

framework/core/js/src/common/components/DataSegment.tsx
davwheat: Maybe `LabelValue`-something?

framework/core/js/src/forum/components/AccessTokensList.tsx
```diff
@@ -0,0 +1,196 @@
...
+    const items = new ItemList<Mithril.Children>();
+
+    if (this.attrs.type === 'session') {
+      items.add(
+        'title',
+        <div className="AccessTokensList-item-title">
+          <span className="AccessTokensList-item-title-main">{token.device()}</span>
+          {token.isCurrent()
+            ? [' — ', <span className="AccessTokensList-item-title-sub">{app.translator.trans('core.forum.security.current_active_session')}</span>]
+            : null}
```
davwheat: Just a bit simpler
```suggestion
          {!!token.isCurrent()
            && [' — ', <span className="AccessTokensList-item-title-sub">{app.translator.trans('core.forum.security.current_active_session')}</span>]}
```

davwheat (approved):

davwheat:
Overall, I think this is great. `DataSegment` could be renamed, but it's quite a difficult component to name imo.

SychO9:
`LabelValue` is a lot better indeed

PR #3589, opened 2022-08-09
fix(compilation): versioner not inject into compilers

### LINKED ISSUE #3583: Compilers are not build by the container
### Current Behavior

It is currently not possible to bind a custom versioner to `VersionerInterface`, as the compiliers are not build by the container.

https://github.com/flarum/framework/blob/main/framework/core/src/Frontend/Compiler/RevisionCompiler.php#L47

The compilers instead are instantiated manually in `Assets`:
https://github.com/flarum/framework/blob/main/framework/core/src/Frontend/Assets.php#L164
https://github.com/flarum/framework/blob/main/framework/core/src/Frontend/Assets.php#L159

`Assets` should instead resolve the compilers with arguments

### Steps to Reproduce

Bind a custom versioner in the container

`$this->container->bind(VersionerInterface::class, MyCustomVersioner:class);`

Notice that this does not have any effect.

### Expected Behavior

For the custom versioner to be injected

### Screenshots

_No response_

### Environment

- Flarum version: x.y.z
- PHP version: x.y.z

```
```

### Possible Solution

_No response_

### Additional Context

_No response_

luceos:
**Fixes #3583**

While adding the revision versioner, I forgot to actually inject this versioner into the compilers. This PR solves that. 

PS this change does not impact fof/nightmode as it doesn't override the method.

I noticed we have zero tests (right?) for the Assets and Compilers, I don't feel these should be introduced now. But would it make sense to add an issue for that to be completed inside a 1.x release?

**Screenshot**

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/framework/core/src/Frontend/Assets.php
+++ b/framework/core/src/Frontend/Assets.php
@@ -156,12 +156,18 @@ class Assets
 
     protected function makeJsCompiler(string $filename)
     {
-        return new JsCompiler($this->assetsDir, $filename);
+        return resolve(JsCompiler::class, [
+            'assetsDir' => $this->assetsDir,
+            'filename' => $filename
+        ]);
     }
 
     protected function makeLessCompiler(string $filename): LessCompiler
     {
-        $compiler = new LessCompiler($this->assetsDir, $filename);
+        $compiler = resolve(LessCompiler::class, [
+            'assetsDir' => $this->assetsDir,
+            'filename' => $filename
+        ]);
 
         if ($this->cacheDir) {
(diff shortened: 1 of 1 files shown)


davwheat (approved):

SychO9 (approved):
I'm fine with this for 1.x, in 2.0 it would be good to refactor this frontend compilation code to be more flexible. We've been more and more relying on using `resolve` even though core prefers not to, we can decrease that with breaking changes in 2.0 though.

imorland (approved):
Checks out fine with the intended use case from an extension 👍

PR #3590, opened 2022-08-09
feat: Admin debug mode warning

imorland:
Adds a `debug` warning widget to the admin dashboard. When `debug` is not enabled, the widget is not displayed.

How's the `less`? I'm terrible...

**Screenshot**
image

**Necessity**

**Confirmed**

**Required changes:**


framework/core/js/src/admin/components/DebugWarningWidget.js
```diff
@@ -0,0 +1,20 @@
...
+  content() {
+    return (
+      <div>
+        <h4 className="DebugWarningWidget-title">
+          {icon('fas fa-exclamation-triangle')} {app.translator.trans('core.admin.debug-warning.label')}
+        </h4>
+        <p className="DebugWarningWidget-detail">{app.translator.trans('core.admin.debug-warning.detail')}</p>
+      </div>
+    );
+  }
```
SychO9: We can probably use the `Alert` component directly here and then make any necessary CSS tweaks.

imorland: I think I follow.. updated, that better?

SychO9:
I pushed what I had in mind would be a good opportunity to further improve the alert component so that it can be used in similar ways without hassle or custom style changes. This is the result (`error` type alert too much?)

Screenshot from 2022-08-11 18-21-14

imorland:
Was just looking at this, I think the changes to the `Alert` component are useful 👍 

Only thought is, for _this_ use case, I think it should be a `warning`, rather than `error`?

SychO9:
> Was just looking at this, I think the changes to the `Alert` component are useful +1
> 
> Only thought is, for _this_ use case, I think it should be a `warning`, rather than `error`?

agreed 🙈

framework/core/js/src/admin/components/DebugWarningWidget.js
```diff
@@ -0,0 +1,21 @@
...
+
+export default class DebugWarningWidget extends DashboardWidget {
+  className() {
+    return 'DebugWarningWidget';
+  }
+
+  content() {
+    return (
+      <Alert
+        type="error"
```
SychO9:
```suggestion
        type="warning"
```

framework/core/locale/core.yml
```diff
@@ -59,6 +59,14 @@ core:
       title: Dashboard
       tools_button: Tools
 
+    # These translations are usin in the debug warning widget.
+    debug-warning:
+      detail: |
+        When <code>debug</code> mode is active, Flarum will rebuild it's <code>js</code> and <code>less</code> assets on every request, and could also potentially leak other information, such as database secrets, environment variables, etc.
+
+        It is highly recommended to disable <code>debug</code> in your <code>config.php</code> file when debugging is not required.
+      label: Debug mode active
```
SychO9:
```suggestion
    debug-warning:
      detail: |
        When <code>debug</code> mode is active, Flarum will rebuild its <code>JavaScript</code> and <code>CSS</code> assets on every request, and could also potentially leak other information, such as database secrets, environment variables, etc.

        It is highly recommended to disable <code>debug</code> in your <code>config.php</code> file in production.
      label: Debug mode active
```

imorland:
Updated screenshot with latest changes

SychO9 (approved):

framework/core/js/src/common/components/Alert.tsx
```diff
@@ -51,6 +58,12 @@ export default class Alert<T extends AlertAttrs = AlertAttrs> extends Component<
 
     return (
       <div {...attrs}>
+        {title ? (
+          <div class="Alert-title">
+            {icon ? <span className="Alert-title-icon">{iconHelper(icon)}</span> : null}
+            <span className="Alert-title-text">{title}</span>
+          </div>
+        ) : null}
```
davwheat:
```suggestion
        {!!title && (
          <div class="Alert-title">
            {!!icon && <span class="Alert-title-icon">{iconHelper(icon)}</span>}
            <span class="Alert-title-text">{title}</span>
          </div>
        )}
```

framework/core/locale/core.yml
```diff
@@ -59,6 +59,14 @@ core:
       title: Dashboard
       tools_button: Tools
 
+    # These translations are usin in the debug warning widget.
+    debug-warning:
+      detail: |
+        When <code>debug</code> mode is active, Flarum will rebuild its <code>JavaScript</code> and <code>CSS</code> assets on every request, and could also potentially leak other information, such as database secrets, environment variables, etc.
+
+        It is highly recommended to disable <code>debug</code> in your <code>config.php</code> file in production.
```
davwheat: Do we have a docs page we could link to about this, just for clarification?

davwheat (changes requested):

framework/core/locale/core.yml
imorland: We have https://docs.flarum.org/troubleshoot/#step-0-activate-debug-mode

Perhaps this is what you had in mind?

davwheat: Yeah, I think that works.

davwheat (approved):

PR #3591, opened 2022-08-09
fix: user has wrong discussion read status

### LINKED ISSUE #3524: Deleting last posts can cause the discussion to remain read
## Bug Report

**Current Behavior**
If user A has read all of discussion B, and the last posts of that discussion are deleted. The discussion will remain marked as read for a while for user A even with new posts.

**Steps to Reproduce**
1. Create a discussion with many posts as an admin.
2. Read the discussion as user A.
3. Delete last 2 posts as an admin.
4. Create a new post as an admin.
5. Notice discussion still marked as read as user A.

**Expected Behavior**
Discussion should be marked as unread.

**Environment**
- Flarum version: 1.3

**Possible Solution**
What needs to be done is to update the discussion_user table on post deletion, so that last_read_post_number of the discussion for any user where that value was equal to the deleted post number is lowered. And we should add an integration test for this as well.

**Additional Context**
[Add any other context about the problem here.](https://discuss.flarum.org/d/31144-possibly-a-bug)

SychO9:
**Fixes #3524**

Updates the user state of a discussion when a last post is deleted. The idea is to decrement the `last_read_post_number` to the actual new lat post number, so that users can still be notified of replies and read status is correctly computed as both rely on the `last_read_post_number` field.

**Necessity**

**Confirmed**

### DIFF
--- a/extensions/subscriptions/tests/integration/api/discussions/ReplyNotificationTest.php
+++ b/extensions/subscriptions/tests/integration/api/discussions/ReplyNotificationTest.php
@@ -51,7 +51,7 @@ class ReplyNotificationTest extends TestCase
         /** @var User $mainUser */
         $mainUser = User::query()->find(1);
 
-        $this->assertEquals(0, $this->getUnreadNotificationCount($mainUser));
+        $this->assertEquals(0, $mainUser->getUnreadNotificationCount());
 
         $this->send(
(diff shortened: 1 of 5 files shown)


davwheat (approved):

luceos (approved):
Far better PR than mine, well done!

PR #3597, opened 2022-08-12
test: prevent running both `push` and `pull_request` actions at the same time

SychO9:
Not a 100% perfect solution, however as us maintainers always push branches directly to the original repository and make PRs from those, this is plenty enough. Instead of having 300+ actions in each PR which takes time, we'll just have the normal 171 `push` actions.

This makes sure to still run `pull_request` actions from contributors as they'd be forking the repo.

**Necessity**

**Confirmed**

### DIFF
--- a/.github/workflows/REUSABLE_backend.yml
+++ b/.github/workflows/REUSABLE_backend.yml
@@ -88,7 +88,9 @@ jobs:
 
     name: 'PHP ${{ matrix.php }} / ${{ matrix.db }} ${{ matrix.prefixStr }}'
 
-    if: inputs.enable_backend_testing
+    if: >-
+      inputs.enable_backend_testing &&
+      ((github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name != github.repository) || github.event_name != 'pull_request')
 
     steps:
       - uses: actions/checkout@master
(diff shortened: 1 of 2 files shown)


davwheat (commented):
Would it be easier if we only ran push checks on pushes to main?

SychO9:
> Would it be easier if we only ran push checks on pushes to main?

I don't think we want to do that, a lot of times we all push things to branches and open PRs only after tests are successful or if we're just testing something in workflows or in actual tests. Overall I think we need push event on any branch.

davwheat:
Yeah that's fair. I was basing it off my recent preferred option of opening draft PRs instead.

davwheat:
Could this affect the checks portion of PRs, though? I'm not sure if they list push and PR triggered workflows or just PR ones.

SychO9:
> Could this affect the checks portion of PRs, though? I'm not sure if they list push and PR triggered workflows or just PR ones.

Not as far as I've seen, the checks portion in PRs lists any and all checks from the last commit. So for example this PR itself only shows push workflows (since the change is taking effect in this PR already and there are no pull_request workflows). Normally there'd be double, push + pull_request and would take much longer.

davwheat (approved):

PR #3601, opened 2022-08-15
perf(statistics): split timed data into per-model XHR requests

davwheat:

- splits up timed data into multiple XHR requests so info about the models are loaded separately
- also kills off the timezone offset option that was undocumented and seemingly not useful

**Necessity**

**Confirmed**

### DIFF
--- a/extensions/statistics/js/src/admin/components/StatisticsWidget.tsx
+++ b/extensions/statistics/js/src/admin/components/StatisticsWidget.tsx
@@ -25,11 +25,14 @@ export default class StatisticsWidget extends DashboardWidget {
 
   chart: any;
 
-  timedData: any;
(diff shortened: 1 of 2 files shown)


luceos (approved):

SychO9 (changes requested):
While the `flarum-statistics.timezone` setting doesn't seem to actually be a thing, I don't think we want to remove the rest of the conversion to user timezone.

davwheat:
From my understanding, the code gets the user timezone in PHP (`date_default_timezone_get`), which should always be UTC for Flarum, right?

SychO9:
you're right, all of the conversion relies on that value, which is essentially the server default anyway.

SychO9 (approved):

PR #3610, opened 2022-08-22
feat: customizable session driver

SychO9:
**Part of the security roadmap**

* Makes the session driver interchangeable (currently, a driver can be selected by setting `session.driver` in `config.php`, in the future probably through an advanced admin page UI).
* Introduces the needed extension API.
* Adds tests.
* Adds session driver to info command.

* Currently if the driver is configured to a driver that doesn't exist an exception is thrown, would we rather silently fallback to the default? I feel like that might be more convenient as an admin might disable the extension that provides the driver and crashing with an exception wouldn't be nice.

**Necessity**

**Confirmed**


luceos:
> * Currently if the driver is configured to a driver that doesn't exist an exception is thrown, would we rather silently fallback to the default? I feel like that might be more convenient as an admin might disable the extension that provides the driver and crashing with an exception wouldn't be nice.

I think that makes the most sense. It would however be necessary to add an entry to the logs (error/critical) that provides feedback about this failure to the webmaster.

luceos (approved):

framework/core/src/User/SessionServiceProvider.php
```diff
@@ -20,12 +22,42 @@ class SessionServiceProvider extends AbstractServiceProvider
...
-            return new FileSessionHandler(
-                $container['files'],
-                $container['config']['session.files'],
-                $container['config']['session.lifetime']
-            );
+        $this->container->singleton('flarum.session.drivers', function () {
+            return [];
+        });
+
+        $this->container->singleton('session', function (Container $container) {
```
askvortsov1, line 29: Why `session` and not `session_manager`?

framework/core/src/User/SessionManager.php
```diff
@@ -0,0 +1,45 @@
...
+
+class SessionManager extends IlluminateSessionManager
+{
+    public function handler(string $driver = null): SessionHandlerInterface
+    {
+        $config = $this->container->make(Config::class);
+
+        /** @var Store $driver */
+        try {
+            $driverInstance = parent::driver($driver ?? $config['session.driver']);
```
askvortsov1:
```suggestion
            $requestedDriver = $driver ?? $config['session.driver'];
            $driverInstance = parent::driver($requestedDriver);
```

askvortsov1: Just for clarity

framework/core/src/User/SessionManager.php
```diff
@@ -0,0 +1,45 @@
...
+use Flarum\Foundation\Config;
+use Illuminate\Session\SessionManager as IlluminateSessionManager;
+use Illuminate\Session\Store;
+use InvalidArgumentException;
+use Psr\Log\LoggerInterface;
+use SessionHandlerInterface;
+
+class SessionManager extends IlluminateSessionManager
+{
+    public function handler(string $driver = null): SessionHandlerInterface
```
askvortsov1: driverName maybe?

framework/core/src/User/SessionServiceProvider.php
```diff
@@ -20,12 +22,42 @@ class SessionServiceProvider extends AbstractServiceProvider
...
+            return $manager;
+        });
+
+        $this->container->alias('session', SessionManager::class);
+
+        $this->container->singleton('session.handler', function (Container $container): SessionHandlerInterface {
+            /** @var SessionManager $manager */
+            $manager = $container->make('session');
+
+            return $manager->handler();
```
askvortsov1, line 60: What's the point of being able to request a custom handler if we always go for the default one?

framework/core/src/User/SessionManager.php
```diff
@@ -0,0 +1,45 @@
...
+        $config = $this->container->make(Config::class);
+
+        /** @var Store $driver */
+        try {
+            $driverInstance = parent::driver($driver ?? $config['session.driver']);
+        } catch (InvalidArgumentException $e) {
+            if (! $driver) {
+                // If we're expecting the default driver, and it's not available,
+                // then we'll fall back to the default driver.
+                $driverInstance = parent::driver($this->getDefaultDriver());
```
askvortsov1: Sorry, this isn't making a lot of sense to me. The way it looks, we have 3 layers:

1. If the caller has specified a driver, try to get that. We don't seem to be using this ability, why do we support it?
2. Otherwise, get the driver as set up in `config.php`
3. If that fails (which seems likely, since most configs won't explicitly configure a driver), try to get the default driver, and log an error?

The messages / text being logged seems to be confused between the "default driver" and "config.php configured driver"

askvortsov1 (changes requested):

framework/core/src/User/SessionServiceProvider.php
SychO9: Laravel's binding is called `session` and our codebase seems consistent in using laravel binding names when it makes sense, like the filesystem ones.

framework/core/src/User/SessionManager.php
SychO9: Yeah, I think it's over-engineered. Initially, I started by overriding the `driver` method of the parent class which allows specifying a driver name. But switched to a new method t explicitly return a session handler interface. So we can simplify this further.

framework/core/src/Foundation/Console/InfoCommand.php
```diff
@@ -168,4 +186,27 @@ private function identifyDatabaseVersion(): string
...
+        // Get class name
+        $handlerName = get_class($this->sessionHandler);
+        // Drop the namespace
+        $handlerName = Str::afterLast($handlerName, '\\');
+        // Lowercase the class name
+        $handlerName = strtolower($handlerName);
+        // Drop everything like queue SyncQueue, RedisQueue
+        $handlerName = str_replace('sessionhandler', '', $handlerName);
+
+        return $configuredDriver !== $handlerName ? "$handlerName <comment>(code override, configured to <options=bold,underscore>$configuredDriver</>)</comment>" : $configuredDriver;
```
askvortsov1: Let's use an if(configuredDriver === handlerName) { return configuredDriver} return ...;

That should improve readability a bit/

framework/core/src/Foundation/Console/InfoCommand.php
```diff
@@ -168,4 +186,27 @@ private function identifyDatabaseVersion(): string
...
+
+    /**
+     * Some extensions/packages might be overriding the session.handler binding.
+     * So we have to check if the driver used is configured or hardcoded.
+     */
+    private function identifySessionDriver(): string
+    {
+        $defaultDriver = $this->session->getDefaultDriver();
+        $driver = Arr::get($this->config, 'session.driver', $defaultDriver);
+        $this->session->driver($driver);
```
askvortsov1: Not a big fan of effectful code in otherwise pure functions, can we add a comment explaining what this does, if we need it?

framework/core/src/User/SessionManager.php
```diff
@@ -0,0 +1,48 @@
...
+        /** @var Store $driverInstance */
+        try {
+            $driverInstance = parent::driver($driverName);
+        } catch (InvalidArgumentException $e) {
+            $defaultDriverName = $this->getDefaultDriver();
+            $driverInstance = parent::driver($defaultDriverName);
+
+            // But we will log a critical error to the webmaster.
+            $this->container->make(LoggerInterface::class)->critical(
+                "The configured session driver [$driverName] is not available. Falling back to [$defaultDriverName]. Please check your configuration."
```
askvortsov1:
```suggestion
                "The configured session driver [$driverName] is not available. Falling back to default [$defaultDriverName]. Please check your configuration."
```

askvortsov1 (changes requested):

framework/core/src/Foundation/Console/InfoCommand.php
SychO9: good catch actually, should have adapted this when I made the other changes.

framework/core/src/Foundation/Console/InfoCommand.php
SychO9: should be much much more readable now.

askvortsov1 (approved):
Looks good to me!

PR #3616, opened 2022-08-24
fix: password reset leaks user existence

SychO9:
**Part of the security roadmap**

* Removes unknown user error, throws a validation error instead.
* Moves password reset logic to a job to prevent leak by duration.

I wonder if we should drop the command and move it to the controller now that we also added a job. Would that be a possible breaking change?

**Necessity**

**Confirmed**


framework/core/src/User/Command/RequestPasswordResetHandler.php
```diff
@@ -90,29 +64,19 @@ public function handle(RequestPasswordReset $command)
...
 
-        if ($validation->fails()) {
-            throw new ValidationException($validation);
-        }
-
         $user = $this->users->findByEmail($email);
 
-        if (! $user) {
-            throw new ModelNotFoundException;
+        // Prevents leaking user existence.
```
davwheat: How does this not leak user existence?

We still throw an error if the user is not found, it just happens to be a validation error instead. This too causes the issue of confusing genuine users who have entered a valid email and are then told the email entered isn't valid.
 
If a user _knows_ a valid email is entered, they will still know that the user does/doesn't exist based on if we still return an error or not.

Wouldn't it be better to not return an error and instead use the ever more popular approach of "if this user exists, we'll send them an email"?

SychO9: 🤦🏼 you're totally right

framework/core/locale/core.yml
```diff
@@ -350,7 +350,7 @@ core:
     forgot_password:
       dismiss_button: => core.ref.okay
       email_placeholder: => core.ref.email
-      email_sent_message: We've sent you an email containing a link to reset your password. Check your spam folder if you don't receive it within the next minute or two.
+      email_sent_message: If this user exists, we've sent an email containing a link to reset your password. Check your spam folder if you don't receive it within the next minute or two.
```
davwheat: The wording feels a bit clunky. I think it's partially because we switch from "if this user exists" to "your password" in one sentence. How does this sound?

```suggestion
      email_sent_message: If this email is registered, we'll send an email containing a link to reset their password. Check your spam folder if you don't receive it within the next minute or two.
```

davwheat (commented):
We might want to consider sending emails to users when their emails aren't registered (ratelimited, obviously) so that a user doesn't sit aimlessly for minutes if they enter one of two emails they might have used, for example.

I know some other services do this, with something like "You've requested a password reset, but we didn't find a user associated with this email address", etc.

SychO9:
hmm 🤔 I don't know if we would want to send emails to non-registered users, but yeah would definitely require rate limiting which would need to be implemented separately as it would be a cache-based implementation possibly reusing the laravel one and allowing for easy re-usage.

davwheat:
Yeah, fair point. We should be extra careful about spam to non-users.

davwheat (approved):

framework/core/locale/core.yml
```diff
@@ -350,7 +350,7 @@ core:
     forgot_password:
       dismiss_button: => core.ref.okay
       email_placeholder: => core.ref.email
-      email_sent_message: We've sent you an email containing a link to reset your password. Check your spam folder if you don't receive it within the next minute or two.
+      email_sent_message: If this email is registered, we'll send an email containing a link to reset their password. Check your spam folder if you don't receive it within the next minute or two.
```
askvortsov1:
```suggestion
      email_sent_message: If the email you entered is registered with this site, we'll send you an email containing a link to reset your password. Check your spam folder if you don't receive it within the next minute or two.
```

askvortsov1: Minor wording nit: wanted to expand on what "registered" means, which email we're talking about, and a consistent "you"/"your" usage instead of "their".

framework/core/src/User/Command/RequestPasswordResetHandler.php
```diff
@@ -9,77 +9,33 @@
...
     }
 
     /**
      * @param RequestPasswordReset $command
-     * @return \Flarum\User\User
-     * @throws ModelNotFoundException
+     * @return void
+     * @throws ValidationException
      */
     public function handle(RequestPasswordReset $command)
```
askvortsov1: Yeah, let's kill this and go straight to controller. It's not an API route, so we don't really lose abstractability, and the nested controller > handler > job is a bit too much complexity IMO.

askvortsov1 (changes requested):

askvortsov1 (approved):

luceos (approved):

PR #3619, opened 2022-08-26
fix(regression): bad post actions alignment introduced in #3540

SychO9:
**Fixes a regression introduced in #3540**

* We can't reintroduce the clearfix, it messes with first post height (#3540).
* We don't want to un-hide `Post-footer` when it's empty (#2926), but we can do so in normal posts only as it's the only reliable way atm. It doesn't pose a problem anyway unless extensions that add items leave those items hanging in the dom even if they're empty.

**Screenshot**
Screenshot from 2022-08-26 13-51-25

**Necessity**

**Confirmed**

### DIFF
--- a/framework/core/less/forum/DiscussionPage.less
+++ b/framework/core/less/forum/DiscussionPage.less
@@ -33,7 +33,7 @@
     &, > ul {
       width: 150px;
     }
-    
+
     > ul {
       > li {
         margin-bottom: 10px;
@@ -51,9 +51,7 @@
       }
     }
   }
-}
 
-@media @tablet-up {
   .DiscussionPage-stream {
     margin-right: 225px;
   }
--- a/framework/core/less/forum/DiscussionPage.less
+++ b/framework/core/less/forum/DiscussionPage.less
@@ -23,16 +23,24 @@
   }
 }
 @media @tablet-up {
+  .DiscussionPage-discussion {
+    > .container {
+      display: grid;
+      grid-gap: 75px;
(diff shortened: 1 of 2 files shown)


davwheat:
Could we investigate the idea of using flex/grid instead of float for the stream/nav elements? This should let us reimplement the clearfix, while not affecting the height of the first post.

This (obviously implemented with a media query in core):
image

Plus removing the `margin-right` on the stream seems to do the trick.

We can then add back clearfix:

image

And then actions show correctly:

image

SychO9:
@davwheat I'm worried it might easily break any existing custom styling, I think it would make more sense to do that in 2.0 when we refactor it all?

davwheat:
I'm somewhat of the opinion that custom theming should be done with the new Less extenders we have to override files. In this case, as there are no DOM changes, there's no issue here.

Doing full rethemes without overriding the Less source results in even more CSS bloat than we currently have, which is an issue imo.

We've also done other Flex/Grid refactors in 1.x (Notifications List in 1.3, for example)

framework/core/less/forum/DiscussionPage.less
```diff
@@ -51,11 +59,9 @@
       }
     }
   }
-}
 
-@media @tablet-up {
   .DiscussionPage-stream {
-    margin-right: 225px;
+    order: -1;
```
davwheat: I'd prefer to use explicit grid columns or even grid areas here, to be more explicit in what is desired.

davwheat (approved):

PR #3623, opened 2022-09-01
fix: add missing `offline` config when install fresh Flarum

datlechin:
**Fixes #0000**

On the doc page https://docs.flarum.org/config/ it mentions `offline` option but currently in core there is no such option when fresh install Flarum.

So this PR adds `offline` configuration option when fresh install Flarum.

**Screenshot**

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/framework/core/src/Install/Console/FileDataProvider.php
+++ b/framework/core/src/Install/Console/FileDataProvider.php
@@ -20,6 +20,7 @@ use Symfony\Component\Yaml\Yaml;
 class FileDataProvider implements DataProviderInterface
 {
     protected $debug = false;
+    protected $offline = false;
     protected $baseUrl = null;
     protected $databaseConfiguration = [];
     protected $adminUser = [];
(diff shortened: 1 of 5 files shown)


luceos:
I remember discussing this a few years ago and back then the decision was to keep the configuration as uncluttered as possible. Adding this key makes sense, but - similar to debug - can cause confusion what they're for. As configuration files usually aren't reflecting all possible options anyhow, I'm not sure how adding this is beneficial.. It would make more sense to have a dedicated docs page with all keys properly explained.

askvortsov1:
Agreed with @luceos's points above, this is why we have https://docs.flarum.org/config. Perhaps we should instead link to those docs in the generated config file?

Either way, I don't think there's a use case for configuring this at install, especially since offline mode is very limited at the moment (unless I missed something in the last few months that I was away).

luceos:
> Agreed with @luceos's points above, this is why we have https://docs.flarum.org/config. Perhaps we should instead link to those docs in the generated config file?
> 
> Either way, I don't think there's a use case for configuring this at install, especially since offline mode is very limited at the moment (unless I missed something in the last few months that I was away).

Nothing changed on your absence.

PR #3625, opened 2022-09-02
feat: dispatch event to `flarum/installation-packages` on release

### LINKED ISSUE #3561: Automated workflow to generate installable flarum packages

SychO9:
Auto trigger build of installation packages on [flarum/installation-packages](https://github.com/flarum/installation-packages).

* Needs a `PACKAGES_BUILD_TOKEN` secret added first. The token must have permission to trigger the `workfow_dispatch` event on the target repository.
* Then needs to be tested.

**Necessity**

**Confirmed**


luceos:
Is the idea behind your workflow to run after the release, cq the version has been tagged? Because the prepare release workflow only gathers changes, commits, contributors and donators. It doesn't actually tag (yet) because tags need changelogs..

SychO9:
Argh.. true, you run the `prepare release` workflow first, then tag.

That's fine we can always trigger the tarball build manually as well.

SychO9:
Actually, let me just switch to a separate workflow that runs on release

.github/workflows/build-install-packages.yml
```diff
@@ -0,0 +1,29 @@
+name: Build Install Packages
+
+on:
+  release:
+    types: [released]
+
+env:
+  VERSION: ${{ github.event.release.tag_name }}
+  PHP_VERSIONS: '7.3 7.4 8.0 8.1'
+  INSTALL_PACKAGES_INPUTS: '{ "flarum_version": "{0}", "php_versions": "{1}" }'
```
askvortsov1, line 10: Where are these magic variables coming from? Is there no named equivalent?

.github/workflows/build-install-packages.yml
```diff
@@ -0,0 +1,29 @@
+name: Build Install Packages
+
+on:
+  release:
+    types: [released]
+
+env:
+  VERSION: ${{ github.event.release.tag_name }}
+  PHP_VERSIONS: '7.3 7.4 8.0 8.1'
```
askvortsov1: Can we use a list?

.github/workflows/build-install-packages.yml
SychO9: it's the exact values of the env variables above it.

.github/workflows/build-install-packages.yml
SychO9: the value can only be string, integer or boolean, and I don't think there's benefit to declaring it as a JSON array only to reconvert it to a bash list.

SychO9:
```suggestion
  PHP_VERSIONS: '8.1 8.2 8.3'
```

PR #3629, opened 2022-09-06
feat: increase margin between notifications list control icons

askvortsov1:
This feels cleaner, and reduces the chances of a misclick.

Increase button margin

This leaves slightly less room for other buttons, but the difference is very minor, while the benefits outweigh IMO.

**Screenshot**
Before:
image

After:
image

**Necessity**

**Confirmed**

### DIFF
--- a/framework/core/less/forum/NotificationList.less
+++ b/framework/core/less/forum/NotificationList.less
@@ -2,7 +2,7 @@
   overflow: hidden;
 
   .App-primaryControl > button:not(:last-of-type) {
-    margin-right: 8px;
+    margin-right: 12px;
   }
 
   &-header {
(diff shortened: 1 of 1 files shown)


framework/core/less/forum/NotificationList.less
```diff
@@ -2,7 +2,7 @@
   overflow: hidden;
 
   .App-primaryControl > button:not(:last-of-type) {
-    margin-right: 4px;
+    margin-right: 8px;
```
SychO9: I think even slightly more would be good.
```suggestion
    margin-right: 12px;
```

askvortsov1: image

SychO9 (approved):

PR #3641, opened 2022-09-13
fix: email notifications are sent when unconfirmed

luceos:
This fixes an issue with Flarum where users are getting notified through email, even though they haven't confirmed their email address. On discuss we have seen flags from sendgrid where mails have become undeliverable, my process is to mark these users as not having confirmed their email address hoping this prevents them from getting notified. It does not.

Does this solution suffice?

**Necessity**

**Confirmed**

**Required changes:**


askvortsov1 (changes requested):
Is there a reason why we don't want to encapsulate this logic into the shouldEmail method?

luceos:
@askvortsov1 not really. I wasn't sure this method on the user should be modified. In hindsight it might be the better place.

PR #3643, opened 2022-09-16
Make it possible to extend SetupScript

iPurpl3x:

Make it possible to extend `Flarum\Testing\integration\Setup\SetupScript` and added a public method to add settings to the initial installation pipeline.

New usage example:

```php
(new SetupScript())
    ->addSettings([...])
    ->run();
```

As documented, the new method can be used to add settings to the Flarum installation.
Should be used only when it is really needed.
This can be useful in rare cases where the settings are required to be set already when extensions Extenders are executed. In those cases, setting the settings with the `setting()` method of the `TestCase` will not work.

The new methods work and don't break any existing code.

**Necessity**

**Confirmed**

### DIFF
--- a/php-packages/testing/src/integration/Setup/SetupScript.php
+++ b/php-packages/testing/src/integration/Setup/SetupScript.php
@@ -66,7 +66,19 @@ class SetupScript
     /**
      * @var DatabaseConfig
      */
-    private $dbConfig;
+    protected $dbConfig;
+
+    /**
+     * Settings to be applied during installation.
+     * @var array
+     */
+    protected $settings = ['mail_driver' => 'log'];
+
+    /**
+     * Extensions to activate.
+     * @var array
+     */
+    protected $extensions = [];
 
     public function __construct()
     {
@@ -124,8 +136,8 @@ class SetupScript
                 'password',
                 'admin@machine.local'
             ))
-            ->settings(['mail_driver' => 'log'])
-            ->extensions([])
+            ->settings($this->settings)
+            ->extensions($this->extensions)
             ->build();
 
         // Run the actual configuration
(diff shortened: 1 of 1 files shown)


SychO9:
What's the use case for this? I'd rather make sure the current codebase doesn't allow you to achieve your use case before making the change.

iPurpl3x:
Hi @SychO9 

In my use-case, I needed to activate some extensions (like `flarum/tags`) because else, no matter if I did `$this->extension('flarum-tags');` in the TestCase or not, the test would fail because a migration of the extension I am writing the test in, required `flarum/tags` to be enabled, as it alters the `tag_user` table. Seems like migrations are executed before `$this->extension(...);`

Also, I realized that making settings with `$this->setting(...)` was not helping to use `glowingblue/redis-setup`. But when I set `['glowingblue-redis.enableCache' => true]` in the setup, it works as expected.

This is now my `setup.php` code:

```php
$setup = (new SetupScript())
	->addSettings([
		'glowingblue-redis.enableCache' => true,
	])
	->addExtensions([
		'flarum-tags',
		'flarum-likes',
		'glowingblue-redis-setup',
	])
	->run();
```

SychO9:
The `extension` and `setting` methods will have no effect if you are calling them after the application has already been booted, which is very likely what's happening in your tests, because otherwise, migrations run just fine for specified extensions, and some for settings.

Lookout for anything in your tests that could be causing your tests to boot the application earlier than you want, like resolving a class, or making manual database changes or an early request ..etc

iPurpl3x:
Yes I get that, so I need to change the code here : https://github.com/glowingblue/flarum-ext-redis-setup/blob/master/src/Extend/EnableRedis.php#L40 ?

And what about the migrations that fail ?

SychO9:
I don't know how that ties into the test code so I can't tell, maybe if you share the test code in the discord server

iPurpl3x:
I made the update as discussed over Discord @SychO9. If you want to make any adjustments to the docblock I wrote, let me know.

SychO9 (approved):

PR #3645, opened 2022-09-16
feat: introduce frontend extenders

SychO9:
Prior to this, frontend extenders never actually worked before, nor could any extension use them at all as they weren't exported in `compat` either. This PR does the following:
* Refactor extenders in core for them to actually work:
   - They are now in the `common/extenders` directory to avoid conflicts with the `extend.ts` module).
   - An `IExtender` interface has been introduced.
   - Th `Routes` and `PostTypes` extenders have been adapted to the new changes and the `Model` extender was removed for a separate PR.
* Make use of the `Routes` and `PostTypes` extenders in bundled extensions.
* The extenders can be accessed by exposing an `extend` module from the extension, which contains an array of extenders.

* JS Tests are only failing because `check-typings` is running before new typings are built.
* The bundled extension changes were made here as an example, but I can extract them into another PR if wanted. The commits are separated though, so it should be easy to review.

**Necessity**

**Confirmed**


luceos:
I don't feel prefixing interfaces with an `I` is the way to go. We are trying to mimic the ideology of Laravel mostly and Laravel explicitly moved away from `ISomething` and `SomethingInterface` to `Contracts\Something`. Is there a way to do this here as well? For instance naming the `IExtender` something like `Contract\Extends` or `Contract\Extensible`. This, in my experience with Laravel, seemed to be a more eloquent way of using interfaces honestly. What do you think?

SychO9:
> I don't feel prefixing interfaces with an I is the way to go.

We are already using that convention with other frontend interfaces, so it would be inconsistent tbh.

> We are trying to mimic the ideology of Laravel mostly and Laravel explicitly moved away from ISomething and SomethingInterface to Contracts\Something.

That's on the backend side though, and we don't actually mimic Laravel, we explicitly follow the [PSR-2 coding style](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md) which promotes `FooInterface`  `AbstractFoo` ...etc

I've changed the class name to `ExtenderInterface` though

framework/core/js/src/common/extenders/Routes.ts
```diff
@@ -0,0 +1,36 @@
...
+   * @param path The path of the route.
+   * @param component must extend `Page` component.
+   */
+  add(name: string, path: `/${string}`, component: any): Routes {
+    this.routes[name] = { path, component };
+
+    return this;
+  }
+
+  helper(name: string, callback: HelperRoute): Routes {
```
askvortsov1, line 23: Is the name "helper" consistent with the concrete implementations included in core's routing helpers? Can we somehow refer to that, even if the type doesn't adds any additional structure or safety, if only for documentation?

framework/core/js/src/common/extenders/Routes.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    return this;
+  }
+
+  extend(app: Application, extension: ExtensionModuleInterface) {
+    Object.assign(app.routes, this.routes);
+
+    if (Object.keys(this.helpers).length) {
+      Object.assign(app.route, this.helpers);
+    }
```
askvortsov1: Why is this conditional? Assigning in an empty object should be a no-op, so why not do it regardless?

askvortsov1:
> > I don't feel prefixing interfaces with an I is the way to go.
> 
> We are already using that convention with other frontend interfaces, so it would be inconsistent tbh.

I would prefer being consistent across the frontend. Different languages have different coding styles, I prefer to stick to convention in each as opposed to forcing an odd middle ground.

framework/core/js/src/common/extenders/Routes.ts
SychO9: It's referred to as `routeHelpers` in core, so I think `helper` makes sense.

https://github.com/flarum/framework/blob/a595665bfbf7820c58a3911df154ced435b6fbfd/framework/core/js/src/forum/routes.ts#L41-L51

framework/core/js/src/common/extenders/Routes.ts
SychO9: I honestly don't recall why I specifically did this here and not elsewhere, it is unnecessary.

askvortsov1 (approved):
Great step towards a more consistent, declarative API. Sorry for the review delay!

PR #3646, opened 2022-09-16
feat: frontend `Model` extender

SychO9:
**Relies on #3645**

Follow up to the linked PR. Adds a `Model` frontend extender and makes use of it in bundled extensions.

**Necessity**

**Confirmed**


luceos (approved):
I love how much simpler the extensibility becomes in this way 👍

extensions/flags/js/src/forum/extend.ts
```diff
@@ -1,4 +1,12 @@
 import Extend from 'flarum/common/extenders';
+import Post from 'flarum/common/models/Post';
 import FlagsPage from './components/FlagsPage';
+import Flag from './models/Flag';
 
-export default [new Extend.Routes().add('flags', '/flags', FlagsPage)];
+export default [
+  new Extend.Routes().add('flags', '/flags', FlagsPage),
+
+  new Extend.Model(Flag).register('flags'),
```
askvortsov1: Does this mean that one class could be registered for multiple models?

Would attributes / relationships registered on a model be set for all the name bindings it's associated with?

extensions/suspend/js/src/forum/extend.ts
```diff
@@ -0,0 +1,11 @@
+import Extend from 'flarum/common/extenders';
+import User from 'flarum/common/models/User';
+import Model from 'flarum/common/Model';
+
+export default [
+  new Extend.Model(User)
+    .attribute<boolean>('canSuspend')
+    .attribute<Date, string>('suspendedUntil', Model.transformDate)
+    .attribute<string>('suspendReason')
+    .attribute<string>('suspendMessage'),
```
askvortsov1: Perhaps some of these should be nullable?

askvortsov1 (commented):
Do we have any way to declare stubs / bindings so that these supplemental attributes are available through typing?

I know it would need to be done explicitly, but feels like it should be possible.

framework/core/js/src/common/extenders/Model.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export default class Model implements IExtender {
+  private model: { new (): ActualModel };
+  private type: string | null = null;
+
+  public constructor(model: { new (): ActualModel }) {
+    this.model = model;
+  }
+
+  public attribute<T, O = unknown>(name: string, transform: ((attr: O) => T) | null = null): Model {
```
askvortsov1, line 13: We might want to consider attaching things like this in the `extend` stage, so that we can perform safety checks / conflict resolution / "compilation analysis" of some sort.

davwheat:
> Do we have any way to declare stubs / bindings so that these supplemental attributes are available through typing?



I think we've been using a `shims.d.ts` file for adding typings for a while, now.

E.g...

- https://github.com/flarum/framework/blob/b28606b8ef885c67cedff29c45f1c2babf27166b/extensions/package-manager/js/src/admin/shims.d.ts
- https://github.com/blomstra/flarum-ext-fathom-analytics/blob/main/js/%40types/shims.d.ts

SychO9:
⬆️ typing-wise, extender or not, we still have to declare additional typings the same way, though let me check further

extensions/flags/js/src/forum/extend.ts
SychO9: It's a translation of `app.store.models[type] = class;`, and since registering attributes and relationships adds to the class prototype then it would be the case yeah, one could use a model for multiple resource types, although I cat think of a valid use case for that.

extensions/suspend/js/src/forum/extend.ts
SychO9: true

framework/core/js/src/common/extenders/Model.ts
SychO9: Agreed.

extensions/flags/js/src/forum/extend.ts
askvortsov1: Sorry, I forgot to follow up on this. Since the `id` of the model is the text name `'flags'`, that feels like a more obvious constructor argument for the extender. I can see an argument though that requiring the model class itself to be available makes sure it's exported properly, and is available to typing extension. I'm not opposed to doing it this way, just want to ensure that our extenders will have a consistent pattern of type signatures.

extensions/flags/js/src/forum/extend.ts
```diff
@@ -1,4 +1,12 @@
...
+
+  new Extend.Model(Post).hasMany<Flag>('flags').attribute<boolean>('canFlag'),
```
askvortsov1: Is there an option of putting each extender invocation on a separate line for readability, or will the formatter always force it to be like this?

extensions/tags/js/src/forum/extend.ts
```diff
@@ -1,14 +1,22 @@
...
 
 export default [
+  ...commonExtend,
+
   new Extend.Routes()
     .add('tags', '/tags', TagsPage)
     .add('tag', '/t/:tags', IndexPage)
     .helper('tag', (tag) => app.route('tag', { tags: tag.slug() })),
 
   new Extend.PostTypes().add('discussionTagged', DiscussionTaggedPost),
```
askvortsov1: I left a comment above, but this is what I was thinking about; we have `.add('name', Class)` for PostTypes, and `(Class).register('name')` for models. This feels inconsistent

askvortsov1 (changes requested):

extensions/flags/js/src/forum/extend.ts
SychO9: I see where you're coming from. Since the goal here is to add to the store the model class that maps from the model type string `'flags' => Flag` we could either have it as an optional constructor argument, or maybe even better, within a `Store` extender.

```ts
new Store().add('flags', Flag),
```

extensions/flags/js/src/forum/extend.ts
SychO9: Formatter forces it to be this way. We'd probably affect formatting everwhere if we try to do something about it.

askvortsov1: Could leave blank line comments after, maybe?

```
a.call() //
   .call2() //
   .call3(),
```

SychO9: That could work, I'll give it a go

PR #3647, opened 2022-09-17
fix: package manager failures not showing alerts

datlechin:
`LoadingModal` LoadingModal is not closed when the installation fails

Screenshot 2022-09-17 at 19 11 18

**Fixes #0000**

**Screenshot**

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/extensions/package-manager/js/src/admin/components/Installer.tsx
+++ b/extensions/package-manager/js/src/admin/components/Installer.tsx
@@ -64,7 +64,6 @@ export default class Installer extends Component<InstallerAttrs> {
         body: {
           data: this.data(),
         },
-        errorHandler,
       })
       .then((response) => {
         if (response.processing) {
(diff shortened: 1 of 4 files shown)


askvortsov1 (approved):
Change makes sense. Hopefully the request error handling system will show an error alert too? Not seeing that on the screenshot

datlechin:
@askvortsov1 the point is not showing error when validation failed

Screenshot 2022-09-17 at 19 53 35

SychO9 (commented):
Good catch, for the error though, it would be better to move calling the custom `errorHandler` within `catch` of every package manager request so that it can all fallback to the default error handler instead of having to reinvent the wheel. (*should have done that when I built this but oh well..*)

So basically, it's changing every instance of:

```js
app.request({
  method: ...,
  ...,
  errorHandler,
})
```

to 

```js
app.request({
  method: ...,
  ...,
}).catch(errorHandler)
```

and adapting the custom error handler if it needs any changes.

datlechin:
@SychO9 ah, I see, i have try on local
so we no need to use `errorHandler` on `request`
it works well

Screenshot 2022-09-17 at 20 51 33

SychO9:
the custom error handler is necessary for other type of errors which the default one doesn't understand 😇

SychO9 (approved):
Thanks 🤗

PR #3652, opened 2022-09-25
feat: contrast util with yiq calculator

ornanovitch:
**Needed for https://github.com/flarum/framework/pull/3653**

- The `isDark` utility converts a hex color to rgb, and then calcul a YIQ value in order to get the appropriate contrast value (is it dark or is it light?)
- In order to use this new snippet without any pain, I created two new color variables which aren't depending on the dark/light mode: `@text-on-light` and `@text-on-dark`

See #3653 for example

See https://www.w3.org/TR/AERT/#color-contrast for references

I don't know... Is the name getContrast ok regarding your naming conventions?

**Screenshot**

see https://github.com/flarum/framework/pull/3653

**Necessity**

**Confirmed**

- [x] Tests are not appropriate here.

### DIFF
--- a/framework/core/js/src/common/utils/isDark.ts
+++ b/framework/core/js/src/common/utils/isDark.ts
@@ -4,7 +4,7 @@
  * light?) See https://www.w3.org/TR/AERT/#color-contrast for references
  */
 
-export default function getContrast(hexcolor: String) {
+export default function isDark(hexcolor: String) {
 
   var hexnumbers = hexcolor.replace("#", "");
 
@@ -15,7 +15,7 @@ export default function getContrast(hexcolor: String) {
   const r = parseInt(hexnumbers.substr(0,2),16);
   const g = parseInt(hexnumbers.substr(2,2),16);
   const b = parseInt(hexnumbers.substr(4,2),16);
-  const contrast = ((r*299)+(g*587)+(b*114))/1000;
+  const yiq = ((r*299)+(g*587)+(b*114))/1000;
 
-  return contrast;
+  return (yiq >= 128) ? false : true;
 }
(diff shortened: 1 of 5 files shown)


davwheat (changes requested):
Ooh, you moved this here! Makes sense.

I left a comment about exporting an `isDark` util so we don't need to compare against 128 in our code in your draft PR. Could we do that here?

ornanovitch:
> I left a comment about exporting an `isDark` util so we don't need to compare against 128 in our code in your draft PR. Could we do that here?

You're right! I'll do it asap :)

ornanovitch:
@davwheat It seems a lot better, thanks for your advice!

davwheat (approved):

framework/core/js/src/common/utils/isDark.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * The `isDark` utility converts a hex color to rgb, and then calcul a YIQ
+ * value in order to get the appropriate brightness value (is it dark or is it
+ * light?) See https://www.w3.org/TR/AERT/#color-contrast for references. A YIQ
+ * value >= 128 is a light color.
+ */
+
+export default function isDark(hexcolor: String) {
+
```
davwheat:
```suggestion
```

framework/core/js/src/common/utils/isDark.ts
```diff
@@ -0,0 +1,22 @@
...
+  var hexnumbers = hexcolor.replace("#", "");
```
davwheat:
```suggestion
  let hexnumbers = hexcolor.replace("#", "");
```

imorland (approved):

ornanovitch:
:)

iPurpl3x:
@Ornanovitch, @imorland, @davwheat `isDark` throws an error and the app crashes if a color is set to `NULL` in the database…

SychO9:
please open a bug report with specifics.

iPurpl3x:
Here you go @SychO9 https://github.com/flarum/framework/issues/3764

Issue #111, opened 2015-06-19
Searches performed without the fulltext gambit fail


tobscure:
e.g. searching for `tag:blog` fails because the API wants to include relevantPosts but there are none.

PR #3653, opened 2022-09-25
fix(tags): tag text color contrast

### LINKED ISSUE #3633: [Tags] `var(--tag-color)` should depend on `var(--tag-bg)` rather than `var(--body-bg)`
### Current Behavior

When you create a colored tag with an extremely light or dark background, the tag's text color (defined by `var(--tag-color)`) will be more or less readable depending on your choice regarding the dark/light theme.

### Steps to Reproduce

For example, I just created a very light colored tag on Nightly. Here is the same tag appearance with light and dark theme:

| light mode | dark mode |
| ------------ | ------------ |
| Screenshot 2022-09-07 at 18-07-32 Flarum_nightly | Screenshot 2022-09-07 at 18-19-45 Flarum_nightly |

### Expected Behavior

I expect my colored tags to be readable in every condition. This is not literarily a bug, but this is a real a11y issue.

### Screenshots

_No response_

### Environment

- Nightly

_No response_

### Possible Solution

I'd like to propose a new snippet (let's call it `contrast.ts`. I think it's worth placing it directly into the core (`framework/core/js/src/common/utils/contrast.ts`), because it could be very useful to resolve more contrast issues in the future. I already began to develop a PoC some times ago, but I never finished it because I was writing a thesis. Now that I'm free, I'd like to propose a PR which adds this util, and a second PR which uses it in order to make `var(--tag-color)` depending directly on `var(--tag-bg)`. Then it could be used as a good example to improve other pieces of the flarum frontend when it comes to contrast issues.

WDYT?

### Additional Context

The simple snippet I worked on:

```ts
/**
 * The `getContrast` utility converts a hex color to rgb, and then calcul a YIQ
 * value in order to get the appropriate contrast value (is it dark or is it light?)
 * See https://www.w3.org/TR/AERT/#color-contrast for references
 */

export default function getContrast(hexcolor: String) {
  const hexnum = hexcolor.replace("#", "");
  const r = parseInt(hexnum.substr(0,2),16);
  const g = parseInt(hexnum.substr(2,2),16);
  const b = parseInt(hexnum.substr(4,2),16);
  const contrast = ((r*299)+(g*587)+(b*114))/1000;

  return contrast;
}
```

ornanovitch:
**Fixes #3633**

Depends on #3652 

Use [the new `isDark` utility](#3652) to add new classes (e.g. `tag-dark` or `tag-light`) for better contrasts between tag's bg and text.

My code style?

**Screenshot**

|                      | before                                                                                                                                                     | after                                                                                                                                               |
| ------               | ---------------------------------------------------------------------------------------------------------------------------------------------------        | -----                                                                                                                                               |
| discussion list      | Screenshot 2022-09-25 at 18-28-23 flarobe        | Screenshot 2022-09-25 at 18-27-31 flarobe |
| admin edit modal   | Screenshot 2022-09-26 at 10-10-57 flarobe | Screenshot 2022-09-26 at 10-11-28 flarobe |
| tags page            | Screenshot 2022-09-26 at 10-07-41 Tags - flarobe |  Screenshot 2022-10-18 at 20-07-46 Tags - flarobe |
| discussion list hero | Screenshot 2022-09-29 at 15-47-59 General - flarobe | Screenshot 2022-09-29 at 15-59-34 General - flarobe |
| discussion hero      | Screenshot 2022-09-29 at 16-24-49 hello the dark world - flarobe | Screenshot 2022-09-29 at 16-25-06 hello the dark world - flarobe |
| "start a discussion" btn | Screenshot 2022-09-29 at 16-27-45 General - flarobe | Screenshot 2022-09-29 at 16-39-41 General - flarobe |

**Necessity**

**Confirmed**


framework/core/js/src/common/utils/getContrast.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * The `getContrast` utility converts a hex color to rgb, and then calcul a YIQ
+ * value in order to get the appropriate contrast value (is it dark or is it
+ * light?) See https://www.w3.org/TR/AERT/#color-contrast for references
+ */
+
+export default function getContrast(hexcolor: String) {
```
davwheat: I think we should export a second `isDark` util here to avoid repeating `>= 128` and avoid using this magic number through our code.

davwheat (changes requested):

ornanovitch:
There is maybe a last element to work on, but it's less trivial, I need to think a bit more about it:

Screenshot 2022-10-15 at 14-23-02 zfeez - flarobe

 Other than that, I think my proposal is ready :)

luceos (approved):
@Ornanovitch first off, thanks for the determination to get this working, good job!

This looks okay, someone needs to confirm this works locally, though that seems to be okay.

extensions/tags/js/src/forum/addTagLabels.js
```diff
@@ -23,7 +24,13 @@ export default function () {
       const color = tags[0].color();
       if (color) {
         view.attrs.style = { '--hero-bg': color };
-        view.attrs.className += ' DiscussionHero--colored';
```
davwheat, line 26: We would need to keep this for backwards compatibility.

extensions/tags/js/src/forum/addTagLabels.js
```diff
@@ -23,7 +24,13 @@ export default function () {
...
+
+        if (isDark(color)) {
+          view.attrs.className += ' DiscussionHero--dark';
+        }
+        else {
+          view.attrs.className += ' DiscussionHero--light';
+        }
```
davwheat:
```suggestion
        view.attrs.className = classList(view.attrs.className, 'DiscussionHero--colored', isDark(color) ? 'DiscussionHero--dark' : 'DiscussionHero--light');
```

extensions/tags/js/src/forum/addTagLabels.js
```diff
@@ -1,6 +1,7 @@
 import { extend } from 'flarum/common/extend';
 import DiscussionListItem from 'flarum/forum/components/DiscussionListItem';
 import DiscussionHero from 'flarum/forum/components/DiscussionHero';
+import isDark from 'flarum/common/utils/isDark';
```
davwheat:
```suggestion
import isDark from 'flarum/common/utils/isDark';
import classList from 'flarum/common/utils/classList';

```

extensions/tags/js/src/forum/components/TagHero.js
```diff
@@ -1,4 +1,5 @@
 import Component from 'flarum/common/Component';
+import isDark from 'flarum/common/utils/isDark';
 import tagIcon from '../../common/helpers/tagIcon';
```
davwheat, line 3:
```suggestion
import tagIcon from '../../common/helpers/tagIcon';
import classList from 'flarum/common/utils/classList';
```

extensions/tags/js/src/forum/components/TagHero.js
```diff
@@ -7,7 +8,7 @@ export default class TagHero extends Component {
     const color = tag.color();
 
     return (
-      <header className={'Hero TagHero' + (color ? ' TagHero--colored' : '')} style={color ? { '--hero-bg': color } : ''}>
+      <header className={'Hero TagHero' + (color ? ' TagHero--colored' : '') + (isDark(color) ? ' TagHero--dark' : ' TagHero--light')} style={color ? { '--hero-bg': color } : ''}>
```
davwheat:
```suggestion
      <header className={classList('Hero', 'TagHero', { 'TagHero--colored': color }, isDark(color) ? 'TagHero--dark' : 'TagHero--light')} style={color ? { '--hero-bg': color } : ''}>
```

extensions/tags/js/src/forum/components/TagsPage.js
```diff
@@ -58,7 +59,7 @@ export default class TagsPage extends Page {
                 const children = sortTags(tag.children() || []);
 
                 return (
-                  <li className={'TagTile ' + (tag.color() ? 'colored' : '')} style={{ '--tag-bg': tag.color() }}>
+                    <li className={'TagTile ' + (tag.color() ? 'colored ' : '') + (isDark(tag.color()) ? 'tag-dark' : 'tag-light')} style={{ '--tag-bg': tag.color() }}>
```
davwheat:
```suggestion
                  <li className={classList('TagTile', { 'colored': tag.color() }, isDark(tag.color()) ? 'tag-dark' : 'tag-light')} style={{ '--tag-bg': tag.color() }}>
```

extensions/tags/js/src/forum/components/TagsPage.js
```diff
@@ -4,6 +4,7 @@ import Link from 'flarum/common/components/Link';
 import LoadingIndicator from 'flarum/common/components/LoadingIndicator';
 import listItems from 'flarum/common/helpers/listItems';
 import humanTime from 'flarum/common/helpers/humanTime';
+import isDark from 'flarum/common/utils/isDark';
```
davwheat:
```suggestion
import isDark from 'flarum/common/utils/isDark';
import classList from 'flarum/common/utils/classList';
```

extensions/tags/less/common/TagLabel.less
```diff
@@ -16,10 +16,21 @@
   }
 
   &.colored {
-    --tag-color: @body-bg;
```
davwheat, line 19: I think we should keep this line just in case.

extensions/tags/less/common/TagLabel.less
```diff
@@ -16,10 +16,21 @@
...
+      }
+    }
+
+    &.tag-light {
+      --tag-color: @text-on-light;
+
+      .TagLabel-text {
+        color: var(--tag-color) !important;
+      }
     }
```
davwheat:
```suggestion
    &.tag-dark {
      --tag-color: @text-on-dark;
    }

    &.tag-light {
      --tag-color: @text-on-light;
    }
    
    
    .TagLabel-text {
      color: var(--tag-color) !important;
    }
```

davwheat (changes requested):

extensions/tags/js/src/common/helpers/tagLabel.js
```diff
@@ -1,4 +1,5 @@
 import extract from 'flarum/common/utils/extract';
+import isDark from 'flarum/common/utils/isDark';
 import Link from 'flarum/common/components/Link';
```
davwheat, line 2:
```suggestion
import Link from 'flarum/common/components/Link';
import classList from 'flarum/common/utils/classList';
```

extensions/tags/js/src/common/helpers/tagLabel.js
```diff
@@ -14,6 +15,13 @@ export default function tagLabel(tag, attrs = {}) {
     if (color) {
       attrs.style['--tag-bg'] = color;
       attrs.className += ' colored';
+
+      if (isDark(color)) {
+        attrs.className += ' tag-dark';
+      }
+      else {
+        attrs.className += ' tag-light';
+      }
```
davwheat:
```suggestion
      attrs.className = classList(attrs.className, 'colored', isDark(color) ? 'tag-dark' : 'tag-light');
```

davwheat (changes requested):

davwheat:
I know not all of these changes are completely due to your edits (e.g., the use of `classList` instead of string concatenation), but they do represent changes we're trying to make within our code base, and they can be included within this PR without any judgement imo.

ornanovitch:
> I know not all of these changes are completely due to your edits (e.g., the use of `classList` instead of string concatenation), but they do represent changes we're trying to make within our code base, and they can be included within this PR without any judgement imo.

That's fine, thank you for your kind review! I'm not a js dev at all so I can't even argue about this^^ And I clearly understand the path of rationalization you are going through. I'll commit those changes in a few hours :)

ornanovitch:
Oooops, I merged the `dist` files... Is it a problem?

extensions/tags/js/src/forum/components/TagHero.js
```diff
@@ -1,13 +1,15 @@
...
+import classList from 'flarum/common/utils/classList';
 
 export default class TagHero extends Component {
   view() {
     const tag = this.attrs.model;
     const color = tag.color();
 
     return (
-      <header className={'Hero TagHero' + (color ? ' TagHero--colored' : '')} style={color ? { '--hero-bg': color } : ''}>
+      <header className={classList('Hero', 'TagHero', { 'TagHero--colored': color }, isDark(color) ? 'TagHero--dark' : 'TagHero--light')} style={color ? { '--hero-bg': color } : ''}>
```
ornanovitch: `style={color ? { '--hero-bg': color } : ''` has no effect anymore.. I'm investigating :mechanic:

ornanovitch: Never mind, wrong file... :facepalm:

ornanovitch:
~~all is fixed but I don't know what to do about the `dist` files...~~

I made it! No more dirty dist :+1: 

I give it a last review

SychO9:
Thanks for the PR!

Instead of creating multiple classes for each component, and since it's always a color change, I feel like it might be better to have two reusable classes, perhaps something along the lines of `.a11y-lightened-text` `.a11y-darkened-text`.

What do you think @davwheat @Ornanovitch ?

ornanovitch:
> Thanks for the PR!
> 
> Instead of creating multiple classes for each component, and since it's always a color change, I feel like it might be better to have two reusable classes, perhaps something along the lines of `.a11y-lightened-text` `.a11y-darkened-text`.
> 
> What do you think @davwheat @Ornanovitch ?

I need to sleep on it but that sounds like a good idea :)

ornanovitch:
Can someone try this `tag-contrast` branch and tell me if it works? I have some unexpected `TypeError` since my last commits, I think yarn is unhappy :crying_cat_face: I hope it's only a local problem

edit: all good, it's working, I had a local problem!

ornanovitch:
@SychO9 this is more tricky than what I expected, because the vast majority of those rules are too specific:

Those are variables redefinitions:
https://github.com/flarum/framework/blob/ca01b4c049bf638d2c00f1e0d92248a6112e5003/extensions/tags/less/common/TagLabel.less#L21-L27
https://github.com/flarum/framework/blob/ca01b4c049bf638d2c00f1e0d92248a6112e5003/extensions/tags/less/forum/TagHero.less#L1-L8

Those are background rules:
https://github.com/flarum/framework/blob/ca01b4c049bf638d2c00f1e0d92248a6112e5003/extensions/tags/less/common/TagLabel.less#L44-L48

Those rules need the `<a>` tag, not only a class:
https://github.com/flarum/framework/blob/ca01b4c049bf638d2c00f1e0d92248a6112e5003/extensions/tags/less/common/TagLabel.less#L52-L62

Those are mixins:
https://github.com/flarum/framework/blob/ca01b4c049bf638d2c00f1e0d92248a6112e5003/extensions/tags/less/forum/TagTiles.less#L111-L115

So I don't think it worth creating generic classes for now: it would only work for one rule I think (the New Discussion button). WDYT?

SychO9:
Yea, this is where CSS variables help with specificity issues. Let me push a commit later for this 👍🏼

SychO9 (changes requested):
Pushed a commit to make it more flexible to change color based on YIQ using CSS Variables without having to define modifier classes for every component.

I also change the `#111` & `#ddd` colors to `#000` & `#fff`. They easier to the eyes, the other colors looked really odd with some backgrounds.

Another thing I noticed while testing locally 🤔. Too many other colors are changing that I don't think we want them changed. That's due to the threshold of `128` which insures high contrast results. That's good for a11y but I believe we want to introduce a high contrast mode in the future for this specific case rather than change all default coloring.

Can we please increase the threshold to `138`?

ornanovitch:
Wow, what an improvement! Thank you :rocket: :rocket: :rocket: I need to learn about those css variables tricks.

> I also change the #111 & #ddd colors to #000 & #fff. They easier to the eyes, the other colors looked really odd with some backgrounds.

That's fine IMO

> Too many other colors are changing that I don't think we want them changed. That's due to the threshold of 128 which insures high contrast results. That's good for a11y but I believe we want to introduce a high contrast mode in the future for this specific case rather than change all default coloring.
>
> Can we please increase the threshold to 138?

I don't have strong opinion on this, but I'd like to test a little. Do you have examples of some unexpected color changes?

SychO9:
> I'd like to test a little. Do you have examples of some unexpected color changes?

I use tag colors from discuss, the white text color on support tag by default is fine, but isDark with the current threshold of 128 leads in dark text color. (same with 90% tags).

Like I mentioned before, it's actually better for a11y for people who need really high contrast, but the intention is to fix that in the future through a high contrast mode anyway.

ornanovitch:
Tested, it seems perfect to me :)

extensions/tags/less/common/TagLabel.less
```diff
@@ -16,7 +16,7 @@
   }
 
   &.colored {
-    --tag-color: @body-bg;
+    --tag-color: var(--contrast-color, var(--body-bg));
```
ornanovitch, line 19: @SychO9 is this some kind of fallback logic?

SychO9: Yes, it uses the contrast color if it is set (like using the text contrast classes) otherwise it'll use he default value. In this case the default is `body-bg`

ornanovitch: That's awesome, I learn a lot with this refactor, thank you!

SychO9 (approved):
The last commit I pushed preserves design colo consistency in both light & dark modes.
This looks ready to be merged now.

iamdarkle:
This is really cool, congrats on the long work on this!!

I'm glad I'll soon be able to get rid of my CSS mess that was trying to match the color with each of my tags.

askvortsov1 (approved):
Very cool!

PR #3655, opened 2022-10-10
feat: allow specifying extensions when installing an instance

SychO9:
Useful when doing migrations when more than the default extensions are required to migrate the data to flarum. This allows quickly spinning up a flarum database with the necessary schema.

This use case was also raised by the work being done in https://github.com/linc/nitro-porter

**Necessity**

**Confirmed**


luceos (approved):
What an easy fix/addition 👍

imorland (approved):

PR #3656, opened 2022-10-12
feat: send notifications of a new reply when post is approved

SychO9:
Replies held for approval naturally don't send out notifications to users, however currently after a post has been approved users are still not notified that a reply has been posted.

This PR tweaks subscriptions to send notifications once a reply has been approved.

**Reviewers should focus on**
The code in approval was extracted into a listener because no matter what listeners are always executed before subscribers even if the extension is set to load before.

**Necessity**

**Confirmed**


imorland (approved):
Looks good to me!

davwheat (approved):

PR #3658, opened 2022-10-21
[mentions] feat: group mentions

imorland:
**Addresses https://discuss.flarum.org/d/31473-mentioning-groups-within-posts **

Allows mentioning of groups

**Screenshot**
User notification settings:
image

Autocomplete:
image

Preview:
image

Post content:
image

Web notification:
image

Email notification:
```
[2022-10-21 17:54:02] flarum.INFO: Message-ID: <956025521ad75a6bc18b18347855b9d9@dev.lan>
Date: Fri, 21 Oct 2022 17:54:02 +0000
Subject: admin mentioned a group you're a member of in Example discussion
From: Flarum Dev Docker <noreply@localhost>
To: admin <admin@example.com>
MIME-Version: 1.0
Content-Type: text/plain; charset=utf-8
Content-Transfer-Encoding: quoted-printable

Hey admin!

admin mentioned a group you're a member of in Example discussion.

http://dev.lan/d/1/33

---

@"Admins"#g1 
```

**Necessity**

**Confirmed**

**Required changes:**


extensions/mentions/js/src/forum/utils/getMentionText.js
```diff
@@ -28,9 +32,12 @@ export default function getMentionText(user, postId) {
...
     const cleanText = getCleanDisplayName(user);
     return `@"${cleanText}"#${user.id()}`;
-  } else {
+  } else if (group === undefined) {
     // @"Display name"#pPostID
     const cleanText = getCleanDisplayName(user);
     return `@"${cleanText}"#p${postId}`;
+  } else if (postId === undefined) {
+    // @"Name Plural"#gGroupID
+    return `@"${group.namePlural()}"#g${group.id()}`;
```
SychO9: I'd prefer changing the conditionals to `if user, else if post, else if group, else exception` it's more readable and it takes into consideration the case where everything is undefined/falsy.

extensions/mentions/js/src/forum/utils/textFormatter.js
```diff
@@ -31,3 +31,17 @@ export function filterPostMentions(tag) {
     return true;
   }
 }
+
+export function filterGroupMentions(tag) {
+  const group = app.store.getById('groups', tag.getAttribute('id'));
+
+  if (group) {
+    tag.setAttribute('displayname', extractText(group.namePlural()));
```
SychO9: just a terminology nitpick, can we change all `diplayname` occurrences to groupname or just `name`? `displayname` is such a user-specific term in the codebase right now.

extensions/mentions/less/forum.less
```diff
@@ -97,6 +97,17 @@
   position: absolute;
   .Button--color(@tooltip-color, @tooltip-bg);
 }
+.GroupMention {
+  color: @body-bg;
+  i {
+    margin-left: 5px;
+  }
```
SychO9:
```suggestion

  .icon {
    margin-left: 5px;
  }
```

extensions/mentions/migrations/2022_10_21_000001_change_post_mentions_group_add_foreign_keys.php
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Schema\Builder;
+
+return [
+    'up' => function (Builder $schema) {
+        $schema->table('post_mentions_group', function (Blueprint $table) {
+            $table->foreign('post_id')->references('id')->on('posts')->onDelete('cascade');
+            $table->foreign('mentions_group_id')->references('id')->on('groups')->onDelete('cascade');
```
SychO9: let's merge this in just the table creation migration

extensions/mentions/migrations/2022_10_21_000002_add_created_at_to_post_mentions_group.php
```diff
@@ -0,0 +1,30 @@
...
+ * LICENSE file that was distributed with this source code.
+ */
+
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Schema\Builder;
+
+return [
+    'up' => function (Builder $schema) {
+        $schema->table('post_mentions_group', function (Blueprint $table) {
+            $table->timestamp('created_at')->nullable();
```
SychO9: let's merge this in just the table creation migration

extensions/mentions/src/Formatter/FormatGroupMentions.php
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * Configure rendering for group mentions.
+     *
+     * @param s9e\TextFormatter\Renderer $renderer
+     * @param mixed $context
+     * @param string|null $xml
+     * @return void
+     */
+    public function __invoke(Renderer $renderer, $context, string $xml)
```
SychO9:
```suggestion
     * @return string
     */
    public function __invoke(Renderer $renderer, $context, string $xml): string
```

extensions/mentions/src/Formatter/UnparseGroupMentions.php
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * For detailed copyright and license information, please view the
+ * LICENSE file that was distributed with this source code.
+ */
+
+namespace Flarum\Mentions\Formatter;
+
+use Symfony\Contracts\Translation\TranslatorInterface;
+
+class UnparseGroupMentions
```
SychO9: We don't need the unparsing logic if we only have a single format which is `@"name"#gID`, the unparsing logic was added for usernames to convert `@username` to `@"DisplayName"#uID`

SychO9 (changes requested):
Looks neat! a few comments:

* Groups are loaded from the store for autocomplete, do we know if all groups are within the store at that point? part of the requirements suggests an API search endpoint.
* I have yet to test locally, but I think we might be missing taking into consideration hidden groups.
* To avoid abuse, I reckon we need a `mention groups` permission,but since it wasn't listed in the initial requirements, we can add it inanother PR outside of this.

imorland:
> * Groups are loaded from the store for autocomplete, do we know if all groups are within the store at that point? part of the requirements suggests an API search endpoint.

🤦‍♂️ whoops, yes I missed that. I feel the search endpoint should probably be in core, rather than mentions, what is your opinion?


> * I have yet to test locally, but I think we might be missing taking into consideration hidden groups.

As it stands, yes, hidden groups are not returned for users who do not have permission to view them.

> * To avoid abuse, I reckon we need a `mention groups` permission,but since it wasn't listed in the initial requirements, we can add it inanother PR outside of this.

I'm happy to include that in this PR, as it makes complete sense to do that at the same time. Again, probably for core directly, rather than in mentions?

extensions/mentions/less/forum.less
```diff
@@ -108,6 +109,9 @@
     color: @body-bg;
   }
 }
+.Badge--group {
+  box-shadow: none;
+}
```
SychO9: Is this meant to remove shadows within autocomplete? if so, I prefer a more context oriented approach like
```less
.MentionsDropdown .Badge {...}
```

extensions/mentions/migrations/2022_10_21_000000_create_post_mentions_groups_table.php
```diff
@@ -15,6 +15,10 @@
     function (Blueprint $table) {
         $table->integer('post_id')->unsigned();
         $table->integer('mentions_group_id')->unsigned();
+        $table->timestamp('created_at')->nullable();
```
SychO9:
```suggestion
        $table->timestamp('created_at')->useCurrent()->nullable();
```

extensions/mentions/src/ConfigureMentions.php
```diff
@@ -144,24 +144,24 @@ private function configureGroupMentions(Configurator $config)
...
 
         $tag->template = '
             <xsl:choose>
                 <xsl:when test="@deleted != 1">
-                    <span class="GroupMention" style="background: {@color}">@<xsl:value-of select="@displayname"/><i class="icon {@icon}"></i></span>
+                    <span class="GroupMention" style="background: {@color}">@<xsl:value-of select="@groupname"/><i class="icon {@icon}"></i></span>
                 </xsl:when>
                 <xsl:otherwise>
-                    <span class="GroupMention GroupMention--deleted">@<xsl:value-of select="@displayname"/></span>
+                    <span class="GroupMention GroupMention--deleted" style="background: {@color}">@<xsl:value-of select="@groupname"/><i class="icon {@icon}"></i></span>
```
SychO9, line 158: is color available when the group is deleted?

SychO9 (changes requested):
> I feel the search endpoint should probably be in core, rather than mentions, what is your opinion?

Yup. core makes sense.

> I'm happy to include that in this PR, as it makes complete sense to do that at the same time. Again, probably for core directly, rather than in mentions?

`mention groups` permission is only relevant in the mentions extension, so it makes more sense in mentions than in core imo.

extensions/mentions/src/ConfigureMentions.php
imorland: This is needed for the _preview_ post:

With:
image


Without:
image

extensions/mentions/migrations/2022_10_21_000000_create_post_mentions_groups_table.php
imorland: After this change, the tests fail on `MySQL` 5.7 - I think to do with the comment on dbal in the deleted migration.

@SychO9 - https://github.com/flarum/framework/actions/runs/3311921966/jobs/5467972384

https://github.com/flarum/framework/pull/3658/commits/9347665baa1820ed0875c6b2dd22b14fbc2768e7

extensions/mentions/js/src/forum/addComposerAutocomplete.js
```diff
@@ -10,25 +10,39 @@ import highlight from 'flarum/common/helpers/highlight';
...
+          results.forEach((g) => {
+            if (!returnedGroupIds.has(g.id())) {
+              returnedGroupIds.add(g.id());
+              returnedGroups.push(g);
+            }
+          });
+        });
+      }
+
+      buildSuggestions();
```
SychO9, line 45: I believe `buildSuggestions` should be called once we're certain both API requests have finished, like before

extensions/mentions/js/src/forum/addComposerAutocomplete.js
```diff
@@ -124,12 +143,46 @@ export default function addComposerAutocomplete() {
...
+            <button
+              className={'PostPreview ' + className}
+              onclick={() => applySuggestion(replacement)}
+              onmouseenter={function () {
+                dropdown.setIndex($(this).parent().index());
+              }}
+            >
+              <span className="PostPreview-content">
+                <Badge
+                  class={`Avatar Badge Badge--group Badge--group--${group.id()} Badge-icon `}
```
SychO9: No longer needed
```suggestion
                  class={`Avatar Badge Badge--group--${group.id()} Badge-icon `}
```

extensions/mentions/migrations/2022_10_21_000000_create_post_mentions_groups_table.php
```diff
@@ -0,0 +1,24 @@
...
+    function (Blueprint $table) {
+        $table->integer('post_id')->unsigned();
+        $table->integer('mentions_group_id')->unsigned();
+        $table->timestamp('created_at')->nullable();
+        $table->primary(['post_id', 'mentions_group_id']);
+
+        $table->foreign('post_id')->references('id')->on('posts')->onDelete('cascade');
+        $table->foreign('mentions_group_id')->references('id')->on('groups')->onDelete('cascade');
+    }
+);
```
SychO9: > After this change, the tests fail on MySQL 5.7 - I think to do with the comment on dbal in the deleted migration.
>
> @SychO9 - https://github.com/flarum/framework/actions/runs/3311921966/jobs/5467972384
>
> https://github.com/flarum/framework/commit/9347665baa1820ed0875c6b2dd22b14fbc2768e7

Ah yeah, In that case
```suggestion
use Illuminate\Database\Schema\Blueprint;
use Illuminate\Database\Schema\Builder;

return [
    'up' => function (Builder $schema) {
        $schema->create('post_mentions_groups', function (Blueprint $table) {
            $table->integer('post_id')->unsigned();
            $table->integer('mentions_group_id')->unsigned();
            $table->timestamp('created_at')->nullable();
            $table->primary(['post_id', 'mentions_group_id']);

            $table->foreign('post_id')->references('id')->on('posts')->onDelete('cascade');
            $table->foreign('mentions_group_id')->references('id')->on('groups')->onDelete('cascade');
        });

        // do this manually because dbal doesn't recognize timestamp columns
        $connection = $schema->getConnection();
        $prefix = $connection->getTablePrefix();
        $connection->statement("ALTER TABLE `${prefix}post_mentions_group` MODIFY created_at TIMESTAMP NULL DEFAULT CURRENT_TIMESTAMP");
    },

    'down' => function (Builder $schema) {
        $schema->drop('post_mentions_groups');
    }
];
```

extensions/mentions/migrations/2022_10_21_000001_alter_created_at_set_current_timestamp_on_post_mentions_group.php
```diff
@@ -0,0 +1,23 @@
...
+/*
+ * This file is part of Flarum.
+ *
+ * For detailed copyright and license information, please view the
+ * LICENSE file that was distributed with this source code.
+ */
+
+use Illuminate\Database\Schema\Builder;
+
+return [
```
SychO9: let's delete this and have it all in one migration (previous comment code suggestion)

SychO9 (changes requested):
Sorry, I now realize we don't need the group searching logic at all. I looked at the code to make sure, and we actually preload all the groups always anyway, so the store always has the groups as it is needed by the frontend.

So we can just use the store and a simple mention groups permission in mentions.

You can see that we preload all the groups through the `ForumSerializer` in core.

imorland:
I think everything you've requested is now covered. Still needs tests, but I've started to add them. I'd appreciate if you do have any time if you could lend a hand with them if possible? 🙏

extensions/mentions/extend.php
```diff
@@ -80,13 +87,14 @@
         ->load([
             'mentionsUsers', 'mentionsPosts', 'mentionsPosts.user', 'mentionedBy',
             'mentionedBy.mentionsPosts', 'mentionedBy.mentionsPosts.user', 'mentionedBy.mentionsUsers',
+            'mentionsGroups'
         ]),
 
     (new Extend\ApiController(Controller\CreatePostController::class))
-        ->addInclude(['mentionsPosts', 'mentionsPosts.mentionedBy']),
+        ->addInclude(['mentionsPosts', 'mentionsPosts.mentionedBy', 'mentionsGroups']),
```
SychO9: I believe since groups are preloaded on the frontend we don't need them to be included in responses by default, s an optional include fits better.
```suggestion
        ->addInclude(['mentionsPosts', 'mentionsPosts.mentionedBy'])
        ->addOptionalInclude('mentionsGroups'),
```

extensions/mentions/extend.php
```diff
@@ -80,13 +87,14 @@
...
 
     (new Extend\ApiController(Controller\UpdatePostController::class))
-        ->addInclude(['mentionsPosts', 'mentionsPosts.mentionedBy']),
+        ->addInclude(['mentionsPosts', 'mentionsPosts.mentionedBy', 'mentionsGroups']),
```
SychO9: same here
```suggestion
        ->addInclude(['mentionsPosts', 'mentionsPosts.mentionedBy'])
        ->addOptionalInclude('mentionsGroups'),
```

extensions/mentions/js/src/forum/index.js
```diff
@@ -74,6 +84,8 @@ app.initializers.add('flarum-mentions', function () {
   getPlainContent.removeSelectors.push('a.PostMention');
 });
 
+User.prototype.canMentionGroups = Model.attribute('canMentionGroups');
+
```
SychO9: this should be inside the initializer, somewhere at the top

extensions/mentions/migrations/2022_10_21_000000_create_post_mentions_groups_table.php
```diff
@@ -0,0 +1,34 @@
+<?php
```
SychO9: nitpick: the name of the migration refers to a `post_mentions_groups` table, should be `post_mentions_group`

extensions/mentions/migrations/2022_10_25_000000_add_default_permissions.php
```diff
@@ -0,0 +1,15 @@
...
+ *
+ * For detailed copyright and license information, please view the
+ * LICENSE file that was distributed with this source code.
+ */
+
+use Flarum\Database\Migration;
+use Flarum\Group\Group;
+
+return Migration::addPermissions([
+    'mentionGroups' => Group::MEMBER_ID
```
SychO9: I wonder if this should be restricted by default 🤔 considering the possibility of abuse. Opinions from others would be appreciated @flarum/core

framework/core/src/Extend/Formatter.php
```diff
@@ -52,6 +52,7 @@ public function configure($callback): self
      * - \s9e\TextFormatter\Parser $parser
      * - mixed $context
      * - string $text: The text to be parsed.
+     * - ServerRequestInterface $request. This argument MUST either be nullable, or omitted entirely.
```
SychO9:
```suggestion
     * - ServerRequestInterface|null $request. This argument MUST either be nullable, or omitted entirely.
```

framework/core/src/Search/SearchServiceProvider.php
```diff
@@ -30,7 +30,7 @@ public function register()
         $this->container->singleton('flarum.simple_search.fulltext_gambits', function () {
             return [
                 DiscussionSearcher::class => DiscussionFulltextGambit::class,
-                UserSearcher::class => UserFulltextGambit::class
+                UserSearcher::class => UserFulltextGambit::class,
```
SychO9: nitpick: just to keep the git history clean lol
```suggestion
                UserSearcher::class => UserFulltextGambit::class
```

framework/core/tests/integration/api/groups/ListTest.php
```diff
@@ -74,9 +74,9 @@ public function filters_only_public_groups_for_admin()
             $this->request('GET', '/api/groups', [
                 'authenticatedAs' => 1,
             ])
-            ->withQueryParams([
-                'filter' => ['hidden' => 0],
-            ])
+                ->withQueryParams([
+                    'filter' => ['hidden' => 0],
+                ])
```
SychO9: nitpick: just to keep the git history clean
```suggestion
            ->withQueryParams([
                'filter' => ['hidden' => 0],
            ])
```

framework/core/tests/integration/api/groups/ListTest.php
```diff
@@ -95,9 +95,9 @@ public function filters_only_hidden_groups_for_admin()
             $this->request('GET', '/api/groups', [
                 'authenticatedAs' => 1,
             ])
-            ->withQueryParams([
-                'filter' => ['hidden' => 1],
-            ])
+                ->withQueryParams([
+                    'filter' => ['hidden' => 1],
+                ])
```
SychO9: nitpick: just to keep the git history clean lol
```suggestion
            ->withQueryParams([
                'filter' => ['hidden' => 1],
            ])
```

framework/core/tests/integration/api/groups/ListTest.php
```diff
@@ -114,9 +114,9 @@ public function filters_only_public_groups_for_guest()
     {
         $response = $this->send(
             $this->request('GET', '/api/groups')
-            ->withQueryParams([
-                'filter' => ['hidden' => 0],
-            ])
+                ->withQueryParams([
+                    'filter' => ['hidden' => 0],
+                ])
```
SychO9: nitpick: just to keep the git history clean
```suggestion
            ->withQueryParams([
                'filter' => ['hidden' => 0],
            ])
```

framework/core/tests/integration/api/groups/ListTest.php
```diff
@@ -133,9 +133,9 @@ public function hides_hidden_groups_when_filtering_for_guest()
     {
         $response = $this->send(
             $this->request('GET', '/api/groups')
-            ->withQueryParams([
-                'filter' => ['hidden' => 1],
-            ])
+                ->withQueryParams([
+                    'filter' => ['hidden' => 1],
+                ])
```
SychO9: nitpick: just to keep the git history clean
```suggestion
            ->withQueryParams([
                'filter' => ['hidden' => 1],
            ])
```

framework/core/tests/integration/api/groups/ListTest.php
```diff
@@ -153,9 +153,9 @@ public function paginates_groups_without_filter()
     {
         $response = $this->send(
             $this->request('GET', '/api/groups')
-            ->withQueryParams([
-                'page' => ['limit' => '2', 'offset' => '2'],
-            ])
+                ->withQueryParams([
+                    'page' => ['limit' => '2', 'offset' => '2'],
+                ])
```
SychO9: nitpick: just to keep the git history clean
```suggestion
            ->withQueryParams([
                'page' => ['limit' => '2', 'offset' => '2'],
            ])
```

framework/core/tests/integration/api/groups/ListTest.php
```diff
@@ -172,10 +172,10 @@ public function paginates_groups_with_filter()
...
         $response = $this->send(
             $this->request('GET', '/api/groups')
-            ->withQueryParams([
-                'filter' => ['hidden' => 1],
-                'page' => ['limit' => '1', 'offset' => '1'],
-            ])
+                ->withQueryParams([
+                    'filter' => ['hidden' => 1],
+                    'page' => ['limit' => '1', 'offset' => '1'],
+                ])
```
SychO9: nitpick: just to keep the git history clean
```suggestion
            ->withQueryParams([
                'filter' => ['hidden' => 1],
                'page' => ['limit' => '1', 'offset' => '1'],
            ])
```

framework/core/tests/integration/api/groups/ListTest.php
```diff
@@ -195,9 +195,9 @@ public function sorts_groups_by_name()
             $this->request('GET', '/api/groups', [
                 'authenticatedAs' => 1,
             ])
-            ->withQueryParams([
-                'sort' => 'nameSingular',
-            ])
+                ->withQueryParams([
+                    'sort' => 'nameSingular',
+                ])
```
SychO9: nitpick: just to keep the git history clean
```suggestion
            ->withQueryParams([
                'sort' => 'nameSingular',
            ])
```

SychO9 (changes requested):

extensions/mentions/migrations/2022_10_25_000000_add_default_permissions.php
luceos: Agreed, restrict by default. Don't enable this feature for every user (also mentioned this in the weekly).

extensions/mentions/migrations/2022_10_21_000000_create_post_mentions_group_table.php
```diff
@@ -0,0 +1,34 @@
...
+            $table->primary(['post_id', 'mentions_group_id']);
+
+            $table->foreign('post_id')->references('id')->on('posts')->onDelete('cascade');
+            $table->foreign('mentions_group_id')->references('id')->on('groups')->onDelete('cascade');
+        });
+
+        // do this manually because dbal doesn't recognize timestamp columns
+        $connection = $schema->getConnection();
+        $prefix = $connection->getTablePrefix();
+        $connection->statement("ALTER TABLE `${prefix}post_mentions_group` MODIFY created_at TIMESTAMP NULL DEFAULT CURRENT_TIMESTAMP");
```
clarkwinkelmann: Can we add a comment that explains why `TIMESTAMP` is needed instead of `DATETIME` ? I don't see anything in the code.

Is it the only way to auto-set the value to current timestamp? Any reason to do it in the database and not PHP?

clarkwinkelmann (changes requested):

extensions/mentions/migrations/2022_10_21_000000_create_post_mentions_group_table.php
luceos: I agree that this raw sql should not be part of the migration. Also, dbal is able to recognize timestamp columns just fine, I'm curious what issue you ran into that this was deemed necessary (setting a default)?

imorland: > I agree that this raw sql should not be part of the migration. Also, dbal is able to recognize timestamp columns just fine, I'm curious what issue you ran into that this was deemed necessary (setting a default)?

See https://github.com/flarum/framework/pull/3658#discussion_r1003169130 and https://github.com/flarum/framework/pull/3435

extensions/mentions/migrations/2022_10_25_000000_add_default_permissions.php
imorland: Removed. `admin` only by default now..

framework/core/src/Post/CommentPost.php
```diff
@@ -145,10 +146,11 @@ public function getParsedContentAttribute()
...
      *
      * @param string $value
+     * @param ServerRequestInterface $request
      */
-    public function setContentAttribute($value)
+    public function setContentAttribute($value, ServerRequestInterface $request = null)
     {
-        $this->attributes['content'] = $value ? static::$formatter->parse($value, $this) : null;
+        $this->attributes['content'] = $value ? static::$formatter->parse($value, $this, $request) : null;
     }
```
luceos: I'm slightly worried about this additional argument in this mutator. But I cannot find a reason why we cannot accept it either 🤦

luceos (approved):

extensions/mentions/migrations/2022_10_21_000000_create_post_mentions_group_table.php
clarkwinkelmann: My concern is mostly about the use of TIMESTAMP and not the raw SQL query. At the latest that code will need to be changed in [15 years because of the TIMESTAMP limit](https://en.wikipedia.org/wiki/Year_2038_problem) so I think an explanation of why a DATETIME couldn't be used would be useful to the person that refactors it in the future, without having to dig for the PR that originally introduced it. And even after reading the comments I'm still not sure what disqualifies DATETIME.

EDIT: Or is it planned to keep TIMESTAMPs as long as possible everywhere in Flarum and only change all of them at once at some point in the future? I think it would make sense to use DATETIME for new features as much as possible to reduce the changes required in the future.

framework/core/src/Post/CommentPost.php
clarkwinkelmann: I also don't like it very much, because of the limits it puts on us if we want to add more parameters later.

One solution could be to pass an array of options, one of which could be the request. The same array could then be passed to the formatter.

Another solution could be to ask the developers to use named parameters, and not make the order of the optional parameters part of the official API. But that requires PHP 8.0+

---

Speaking of the actual use case here, if the goal is to share the actor, I think extensions should be able to provide the actor without the need to build a request. What if an extension manually creates a post under a different user than the one currently logged in? What if it's created over a command line utility? Maybe it shouldn't even be called `actor`, but more something like `owner` to designate who wrote the content being saved in the current operation, no matter who actually performs the operation ?

EDIT: another use case I'm surprised we still don't have any extension for (but that's another topic) is re-parsing existing posts after you enable an extension. You don't want to build a fake request object for every single post of the database while looping through them. Maybe `$this->user` should be used by the `setContentAttribute` method directly, and we document the importance of attaching the user (if there's one) to the post before setting the content?

extensions/mentions/migrations/2022_10_21_000000_create_post_mentions_group_table.php
imorland: Thank you for being persistent with this @clarkwinkelmann - turns out I was just being lazy and following previous migrations and using `timestamp` for no other reason other than "that's how we've done it before" 🙈 

I've now switched to using `dateTime`, which also does not require the raw SQL as well, so win-win :)

framework/core/src/Post/CommentPost.php
SychO9: I agree with only passing the actor to `setContentAttribute` and using the post's author by default.

imorland: Whilst I do agree with @clarkwinkelmann 's comment, and certainly passing the `owner` or `actor` arg would allow for greater felixibility going forward, the reasoning behind implementing it this way is to match how this is done already on the formatter's `render()` method.

https://github.com/flarum/framework/blob/main/framework/core/src/Formatter/Formatter.php#L107

That said, I'll see about getting this change implemented as soon as I can

extensions/mentions/src/Formatter/CheckPermissions.php
```diff
@@ -0,0 +1,29 @@
...
+{
+    public function __invoke(Parser $parser, $content, string $text, ?User $actor): string
+    {
+        // Check user has `mentionGroups` permission, if not, remove the `GROUPMENTION` tag the parser
+
+        if ($actor) {
+            if ($actor->cannot('mentionGroups')) {
+                $parser->disableTag('GROUPMENTION');
+            }
+        }
```
SychO9:
```suggestion
        // Check user has `mentionGroups` permission, if not, remove the `GROUPMENTION` tag the parser
        if ($actor && $actor->cannot('mentionGroups')) {
            $parser->disableTag('GROUPMENTION');
        }
```

framework/core/src/Extend/Formatter.php
```diff
@@ -52,6 +52,7 @@ public function configure($callback): self
      * - \s9e\TextFormatter\Parser $parser
      * - mixed $context
      * - string $text: The text to be parsed.
+     * - ServerRequestInterface|null $request. This argument MUST either be nullable, or omitted entirely.
```
SychO9:
```suggestion
     * - \Flarum\User\User|null $actor. This argument MUST either be nullable, or omitted entirely.
```

framework/core/src/Post/CommentPost.php
```diff
@@ -44,9 +45,10 @@ class CommentPost extends Post
      * @param string $content
      * @param int $userId
      * @param string $ipAddress
+     * @param ServerRequestInterface|null $request
      * @return static
      */
-    public static function reply($discussionId, $content, $userId, $ipAddress)
+    public static function reply($discussionId, $content, $userId, $ipAddress, $request)
```
SychO9: Can we pass the actor instead since that's what we need? also it needs to be optional here otherwise it'll be a breaking change.

imorland: Good catch 🙈

imorland:
Note: PHPStan failure is fixed by https://github.com/flarum/framework/pull/3665 by @SychO9

SychO9 (approved):
Added a few more tests, one of them actually caught a bug where the editing of posts doesn't have the actor so group mentions don't work. pushed a fix at the same time.

I believe this is ready now.

imorland:
> Added a few more tests, one of them actually caught a bug where the editing of posts doesn't have the actor so group mentions don't work. pushed a fix at the same time.

Ah nice one, thank you!

PR #3660, opened 2022-10-29
Fix tag discussion count decreased by 2 when hiding before deleting

clarkwinkelmann:
**Fixes #0000**

Check whether the discussion deletion delta was already applied after discussion `Hidden` event before applying it again in `Deleted` event.

This issue was decreasing the discussion count for all tags that belong to the discussion by 2 when the discussion was hidden and then permanently deleted (which would be all of the time when using the web UI to delete discussions permanently). With this fix it will always decrease by one, either if you permanently delete in one go (via REST API) or soft delete and then delete via REST API or UI.

I found other issues with the tag discussion count calculation, I'm already submitting this one since it's the easier to fix and I don't think there will be much debate on this particular change.

I feel like the long term solution is to drop the delta logic entirely and compute the new total via SQL queries which would be able to account for a lot more situations and auto-fix the count if it somehow gets wrong still. 

**Necessity**

**Confirmed**

### DIFF
--- a/extensions/tags/src/Listener/UpdateTagMetadata.php
+++ b/extensions/tags/src/Listener/UpdateTagMetadata.php
@@ -66,7 +66,9 @@ class UpdateTagMetadata
      */
     public function whenDiscussionIsDeleted(Deleted $event)
     {
-        $this->updateTags($event->discussion, -1);
+        // If already soft deleted when permanently deleted, the -1 delta has already been applied in Hidden listener
+        $delta = $event->discussion->hidden_at ? 0 : -1;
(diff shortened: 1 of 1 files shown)


SychO9 (approved):

SychO9:
> I feel like the long term solution is to drop the delta logic entirely and compute the new total via SQL queries which would be able to account for a lot more situations and auto-fix the count if it somehow gets wrong still.

Delta logic can easily go wrong, especially with extensions. So I agree.

luceos (approved):

PR #3666, opened 2022-11-03
feat(phpstan): foundation for usage in extensions

### LINKED ISSUE #3559: PHPStan Setup

SychO9:
## Changes proposed in this pull request
In #3553, we added *phpstan* static code analysis, but only to **core** code. This PR aims to make the `flarum/phpstan` package usable by extension repositories. Which means (at the moment) being able to pick up on new model relations and attributes.

Static analysis is actually expanded to other extensions in a separate PR for ease of readability (#3667).

There are two crucial changes/additions introduced here.

### `Model::cast()` Extender
A new `cast(string $name, string $type)` extender is introduced, which allows adding [type casting](https://laravel.com/docs/8.x/eloquent-mutators#attribute-casting) to attributes added to existing models from extensions. This provides three advantages:
* The feature itself is a good addition to extensions.
* Our *phpstan package* can tell that an attribute added to a core model from an extension is a valid class property, since those are accessed though magic methods.
* Deprecates the old `dateAttribute` extender, since laravel has deprecated the use of the `$dates` property in models for date attributes, and recommends using `$casts` instead. I hope in another PR to also change model classes from directly using `$dates` property. 

### Extending PHPStan
This PR also extends *phpstan* so that it can tell from `extend.php` files what relations and attributes are added to existing models, so that they aren't reported as unknown. `extend.php` files are parsed and currently the package looks up calls to `cast` and `hasOne` `belongsTo` `hasMany` `belongsToMany` method calls to inform the phpstan utility about them.

## Reviewers should focus on:
* `cast` model extender.
* Deprecating `dateAttribute` extender.
* Generally PHPStan extension logic, I would focus less on the code for it.

## Follow-up Tasks
* Write documentation about using PHPStan in extensions.
* Add PHPStan infrastructure to Flarum CLI.

## Necessity

## Confirmed

### DIFF
--- a/framework/core/src/Extend/Model.php
+++ b/framework/core/src/Extend/Model.php
@@ -19,6 +19,7 @@ class Model implements ExtenderInterface
 {
     private $modelClass;
     private $customRelations = [];
+    private $casts = [];
 
     /**
      * @param string $modelClass: The ::class attribute of the model you are modifying.
@@ -34,17 +35,25 @@ class Model implements ExtenderInterface
      *
      * @param string $attribute
      * @return self
+     * @deprecated use `castAttribute` instead. Will be removed in v2.
      */
     public function dateAttribute(string $attribute): self
     {
-        Arr::set(
-            AbstractModel::$dateAttributes,
-            $this->modelClass,
-            array_merge(
-                Arr::get(AbstractModel::$dateAttributes, $this->modelClass, []),
-                [$attribute]
-            )
-        );
+        $this->castAttribute($attribute, 'datetime');
+
+        return $this;
+    }
+
+    /**
+     * Add a custom attribute type cast. Should not be applied to non-extension attributes.
+     *
+     * @param string $attribute: The new attribute name.
+     * @param string $cast: The cast type. See https://laravel.com/docs/8.x/eloquent-mutators#attribute-casting
+     * @return self
+     */
+    public function castAttribute(string $attribute, string $cast): self
+    {
+        $this->casts[$attribute] = $cast;
 
         return $this;
     }
@@ -184,5 +193,14 @@ class Model implements ExtenderInterface
         foreach ($this->customRelations as $name => $callback) {
             Arr::set(AbstractModel::$customRelations, "$this->modelClass.$name", ContainerUtil::wrapCallback($callback, $container));
         }
+
+        Arr::set(
+            AbstractModel::$customCasts,
+            $this->modelClass,
+            array_merge(
+                Arr::get(AbstractModel::$customCasts, $this->modelClass, []),
+                $this->casts
+            )
+        );
     }
 }
--- a/framework/core/src/Extend/Model.php
+++ b/framework/core/src/Extend/Model.php
@@ -35,7 +35,7 @@ class Model implements ExtenderInterface
      *
      * @param string $attribute
      * @return self
-     * @deprecated use `castAttribute` instead. Will be removed in v2.
+     * @deprecated use `cast` instead. Will be removed in v2.
      */
     public function dateAttribute(string $attribute): self
     {
--- a/framework/core/src/Extend/Model.php
+++ b/framework/core/src/Extend/Model.php
@@ -39,7 +39,7 @@ class Model implements ExtenderInterface
      */
     public function dateAttribute(string $attribute): self
     {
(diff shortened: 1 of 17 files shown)


framework/core/src/Extend/Model.php
```diff
@@ -34,17 +35,25 @@ public function __construct(string $modelClass)
      *
      * @param string $attribute
      * @return self
+     * @deprecated use `castAttribute` instead. Will be removed in v2.
```
SychO9:
```suggestion
     * @deprecated use `cast` instead. Will be removed in v2.
```

php-packages/phpstan/src/Attributes/ModelCastAttributeExtension.php
```diff
@@ -0,0 +1,87 @@
...
+
+    public function getProperty(ClassReflection $classReflection, string $propertyName): PropertyReflection
+    {
+        return $this->resolveCastAttributeProperty($this->findCastAttributeMethod($classReflection, $propertyName), $classReflection);
+    }
+
+    private function findCastAttributeMethod(ClassReflection $classReflection, string $propertyName): ?MethodCall
+    {
+        foreach ($this->extendersResolver->getExtenders() as $extender) {
+            if (! $extender->isExtender('Model')) {
```
askvortsov1, line 49: Maybe this string would make sense to put in a constant somewhere? Or could we reflect the model itself to get the unqualified class name?

php-packages/phpstan/src/Attributes/ModelCastAttributeExtension.php
```diff
@@ -0,0 +1,87 @@
...
+                continue;
+            }
+
+            foreach (array_merge([$classReflection->getName()], $classReflection->getParentClassesNames()) as $className) {
+                if ($className === 'Flarum\Database\AbstractModel') {
```
askvortsov1, line 54: Same here, might be cleaner to put all these strings in one place. Deriving it from the class would be nice, but I assume impossible since then phpstan ext would need to depend on core

php-packages/phpstan/src/Attributes/ModelDateAttributesExtension.php
```diff
@@ -0,0 +1,76 @@
...
+use Flarum\PHPStan\Extender\MethodCall;
+use Flarum\PHPStan\Extender\Resolver;
+use PHPStan\Reflection\ClassReflection;
+use PHPStan\Reflection\PropertiesClassReflectionExtension;
+use PHPStan\Reflection\PropertyReflection;
+use PHPStan\Type\NullType;
+use PHPStan\Type\ObjectType;
+use PHPStan\Type\UnionType;
+
+class ModelDateAttributesExtension implements PropertiesClassReflectionExtension
```
askvortsov1, line 22: There's a lot of code repetition; maybe we should consider some abstract class on top of `PropertiesClassReflectionExtension` for finding extenders?

php-packages/phpstan/src/Extender/FilesProvider.php
```diff
@@ -0,0 +1,45 @@
...
+    private $cachedExtenderFiles;
+    /** @var string[] */
+    private $paths;
+
+    public function __construct(array $paths)
+    {
+        $this->paths = $paths;
+    }
+
+    public function getExtenderFiles(): array
```
askvortsov1, line 24: Sometimes I wish PHP were Hack: https://docs.hhvm.com/hack/attributes/predefined-attributes#__memoize

php-packages/phpstan/src/Attributes/ModelDateAttributesExtension.php
SychO9: We can further iterate on this 👍🏼

PR #3668, opened 2022-11-03
feat: include current user as forum relation

imorland:
Except for `Guest` users, include the current user as a relation on `ForumSerializer`.

The need for this came about when I was asked to introduce a way of identifying a user on the API, without knowing the `user_id` in advance, but only an `AccessToken`.

**Necessity**

**Confirmed**

### DIFF
--- a/framework/core/src/Api/Controller/ShowForumController.php
+++ b/framework/core/src/Api/Controller/ShowForumController.php
@@ -25,7 +25,7 @@ class ShowForumController extends AbstractShowController
     /**
      * {@inheritdoc}
      */
-    public $include = ['groups', 'currentUser'];
+    public $include = ['groups', 'actor'];
 
     /**
      * {@inheritdoc}
--- a/framework/core/src/Api/Controller/ShowForumController.php
+++ b/framework/core/src/Api/Controller/ShowForumController.php
@@ -36,7 +36,7 @@ class ShowForumController extends AbstractShowController
 
         return [
             'groups' => Group::whereVisibleTo($actor)->get(),
-            'currentUser' => $actor->isGuest() ? null : $actor
+            'actor' => $actor->isGuest() ? null : $actor
         ];
     }
 }
--- a/framework/core/src/Api/Controller/ShowForumController.php
+++ b/framework/core/src/Api/Controller/ShowForumController.php
@@ -25,15 +25,18 @@ class ShowForumController extends AbstractShowController
     /**
      * {@inheritdoc}
      */
-    public $include = ['groups'];
+    public $include = ['groups', 'currentUser'];
 
     /**
      * {@inheritdoc}
      */
     protected function data(ServerRequestInterface $request, Document $document)
     {
+        $actor = RequestUtil::getActor($request);
+
         return [
-            'groups' => Group::whereVisibleTo(RequestUtil::getActor($request))->get()
+            'groups' => Group::whereVisibleTo($actor)->get(),
+            'currentUser' => $actor->isGuest() ? null : $actor
         ];
     }
 }
--- a/framework/core/src/Api/Serializer/ForumSerializer.php
+++ b/framework/core/src/Api/Serializer/ForumSerializer.php
@@ -15,6 +15,7 @@ use Flarum\Http\UrlGenerator;
 use Flarum\Settings\SettingsRepositoryInterface;
 use Illuminate\Contracts\Filesystem\Cloud;
 use Illuminate\Contracts\Filesystem\Factory;
+use Tobscure\JsonApi\Relationship;
 
 class ForumSerializer extends AbstractSerializer
 {
@@ -68,7 +69,7 @@ class ForumSerializer extends AbstractSerializer
     /**
      * {@inheritdoc}
      */
-    protected function getDefaultAttributes($model)
+    protected function getDefaultAttributes($model): array
     {
         $attributes = [
             'title' => $this->settings->get('forum_title'),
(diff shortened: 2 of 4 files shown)


framework/core/src/Api/Controller/ShowForumController.php
```diff
@@ -25,15 +25,18 @@ class ShowForumController extends AbstractShowController
     /**
      * {@inheritdoc}
      */
-    public $include = ['groups'];
+    public $include = ['groups', 'currentUser'];
```
SychO9:
```suggestion
    public $include = ['groups', 'actor'];
```

framework/core/src/Api/Controller/ShowForumController.php
```diff
@@ -25,15 +25,18 @@ class ShowForumController extends AbstractShowController
...
      * {@inheritdoc}
      */
     protected function data(ServerRequestInterface $request, Document $document)
     {
+        $actor = RequestUtil::getActor($request);
+
         return [
-            'groups' => Group::whereVisibleTo(RequestUtil::getActor($request))->get()
+            'groups' => Group::whereVisibleTo($actor)->get(),
+            'currentUser' => $actor->isGuest() ? null : $actor
```
SychO9:
```suggestion
            'actor' => $actor->isGuest() ? null : $actor
```

framework/core/src/Api/Serializer/ForumSerializer.php
```diff
@@ -135,4 +135,12 @@ public function getAssetUrl($assetPath): string
     {
         return $this->assetsFilesystem->url($assetPath);
     }
+
+    /**
+     * @return \Tobscure\JsonApi\Relationship|null
+     */
+    protected function currentUser($model)
```
SychO9:
```suggestion
    protected function actor($model)
```

framework/core/tests/integration/api/forum/ShowTest.php
```diff
@@ -31,6 +31,22 @@ protected function setUp(): void
...
+    {
+        $response = $this->send(
+            $this->request('GET', '/api')
+        );
+
+        $this->assertEquals(200, $response->getStatusCode());
+
+        $json = json_decode($response->getBody()->getContents(), true);
+
+        $this->assertArrayNotHasKey('currentUser', Arr::get($json, 'data.relationships'));
```
SychO9:
```suggestion
    public function guest_user_does_not_see_actor_relationship()
    {
        $response = $this->send(
            $this->request('GET', '/api')
        );

        $this->assertEquals(200, $response->getStatusCode());

        $json = json_decode($response->getBody()->getContents(), true);

        $this->assertArrayNotHasKey('actor', Arr::get($json, 'data.relationships'));
```

framework/core/tests/integration/api/forum/ShowTest.php
```diff
@@ -51,6 +67,8 @@ public function normal_user_sees_most_information()
         $this->assertEquals('http://localhost/api', Arr::get($json, 'data.attributes.apiUrl'));
 
         $this->assertArrayNotHasKey('adminUrl', Arr::get($json, 'data.attributes'));
+        $this->assertArrayHasKey('currentUser', Arr::get($json, 'data.relationships'));
+        $this->assertEquals(2, Arr::get($json, 'data.relationships.currentUser.data.id'));
```
SychO9:
```suggestion
        $this->assertArrayHasKey('actor', Arr::get($json, 'data.relationships'));
        $this->assertEquals(2, Arr::get($json, 'data.relationships.actor.data.id'));
```

SychO9 (changes requested):
Can we rename `currentUser` to actor instead ?

Also, since we're including the actor in the forum payload, we should look into if we need to adapt `CorePayload.php` accordingly.

imorland:
> we should look into if we need to adapt `CorePayload.php` accordingly.

That didn't even cross my mind 🙈 Good shout!

framework/core/src/Api/Serializer/ForumSerializer.php
```diff
@@ -135,4 +135,12 @@ public function getAssetUrl($assetPath): string
     {
         return $this->assetsFilesystem->url($assetPath);
     }
+
+    /**
+     * @return \Tobscure\JsonApi\Relationship|null
+     */
+    protected function actor($model)
```
davwheat: Could this be typed directly? Probably not. PHP still messes with me...

```suggestion
    protected function actor($model): ?\Tobscure\JsonApi\Relationship
```

imorland: Something is telling me no for PHP 7.3? Bu then again it's really early and my mind is not awake..

SychO9: it's valid 7.3 syntax, we should import the class though instead of using the full qualified name directly

davwheat (approved):

luceos (approved):
You have added multiple return type hints. IMO if a return type hint shows exactly the same information as the phpdoc, you should remove the phpdoc. There's no added value in a phpdoc (which is easily outdated) in this case.

SychO9:
Let's not forget to check this btw

> Also, since we're including the actor in the forum payload, we should look into if we need to adapt CorePayload.php accordingly

imorland:
> Let's not forget to check this btw
> 
> > Also, since we're including the actor in the forum payload, we should look into if we need to adapt CorePayload.php accordingly

Already checked and adapted :) https://github.com/flarum/framework/pull/3668/commits/08ca28cf744f846a7f613c239f812d061c97ad34

SychO9 (approved):
Woups didn't notice since it was above the comment

PR #3669, opened 2022-11-04
chore: Indicate cross-origin request in generic error message

clarkwinkelmann:
**Fixes #0000**

This is another change intended to help with troubleshooting requests. Most URL misconfiguration issues lead to CORS errors, but the generic request error message from Flarum doesn't allow to recognize those immediately since it covers a wide range of client-side and server-side errors.

This PR adds a "(cross-origin request)" indicator at the end of the generic "Ooops" message if the request URL origin and page origin are different. The regular message is unchanged if both origins match.

Here were some of the reflections behind the approach:

Since the error is purely client-side and doesn't expose anything sensitive, I think it makes perfect sense to expose it to every visitor, including when debug mode is OFF.

The message shouldn't be hidden behind a debug button, otherwise nobody would include it in their initial support request.

I don't think the message should mention the Flarum URL config, that would be a bit too specific since users *could* legitimately run Flarum and its API on different domains with proper CORS headers. It might also be too delicate to straight up say the website is misconfigured if anyone might encounter the message. The proposed wording is quite discreet, while still being an obvious clue to any Flarum support agent.

There's the possibility the `new URL()` call fails if the URL given to `m.request()` is completely invalid. I'm not sure if that's worth catching. The request was going to fail anyway if that happened.

**Screenshot**
image
image

**Necessity**

**Confirmed**

### DIFF
--- a/framework/core/locale/core.yml
+++ b/framework/core/locale/core.yml
@@ -550,7 +550,7 @@ core:
       dependent_extensions_message: "Cannot disable {extension} until the following dependent extensions are disabled: {extensions}"
       extension_initialiation_failed_message: "{extension} failed to initialize, check the browser console for further information."
       generic_message: "Oops! Something went wrong. Please reload the page and try again."
-      generic_cross_origin_message: "Oops! Something went wrong. Please reload the page and try again (cross origin request)."
+      generic_cross_origin_message: "Oops! Something went wrong (cross origin request). Please reload the page and try again."
       missing_dependencies_message: "Cannot enable {extension} until the following dependencies are enabled: {extensions}"
       not_found_message: The requested resource was not found.
       payload_too_large_message: The request payload was too large.
--- a/framework/core/locale/core.yml
+++ b/framework/core/locale/core.yml
@@ -550,6 +550,7 @@ core:
       dependent_extensions_message: "Cannot disable {extension} until the following dependent extensions are disabled: {extensions}"
       extension_initialiation_failed_message: "{extension} failed to initialize, check the browser console for further information."
       generic_message: "Oops! Something went wrong. Please reload the page and try again."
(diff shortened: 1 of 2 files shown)


clarkwinkelmann:
For anyone reading the email notifications, I have changed the position of the (cross origin request) text in the error message and updated the screenshots in the description. I think it makes more sense that way.

Should I add an hyphen in "cross-origin"? :thinking:

clarkwinkelmann:
I'm not sure why the PHP unit tests failed since I didn't change any PHP code :man_shrugging:

framework/core/locale/core.yml
```diff
@@ -550,6 +550,7 @@ core:
       dependent_extensions_message: "Cannot disable {extension} until the following dependent extensions are disabled: {extensions}"
       extension_initialiation_failed_message: "{extension} failed to initialize, check the browser console for further information."
       generic_message: "Oops! Something went wrong. Please reload the page and try again."
+      generic_cross_origin_message: "Oops! Something went wrong (cross origin request). Please reload the page and try again."
```
davwheat: If we choose to show this to all visitors, I think the CORS message should either be at the end of the text, or flow with the other text. (Also, yes, I think it should be hyphenated.)

```suggestion
      generic_cross_origin_message: "Oops! Something went wrong during a cross-origin request. Please reload the page and try again."
```

```suggestion
      generic_cross_origin_message: "Oops! Something went wrong. Please reload the page and try again. (cross-origin request)"
```

clarkwinkelmann: I like the first proposal ("during").

I fear that if it is at the end, there's more chance it will be omitted during a support request.

I'll wait a bit before making the change to hear if there's other feedback (or someone from the core team can do it).

SychO9: I like the first proposal as well.

SychO9 (approved):

davwheat (approved):

PR #3670, opened 2022-11-06
feat: Allow additional login params, Introduce `LogInValidator`

imorland:
Precursor to https://discuss.flarum.org/d/31904-enabling-cloudflare-turnstile-on-login

- Extract `params` into an extendable function so that additional params can be added in `LogInModal`
- `LogInController` runs validation of the supplied `params` against a new `LogInValidator` - empty ruleset by default.

**Necessity**

**Confirmed**

**Required changes:**


framework/core/js/src/forum/components/LogInModal.tsx
```diff
@@ -9,6 +9,7 @@ import ItemList from '../../common/utils/ItemList';
 import Stream from '../../common/utils/Stream';
 import type Mithril from 'mithril';
 import RequestError from '../../common/utils/RequestError';
+import { LoginParams } from '../../common/Session';
```
davwheat:
```suggestion
import type { LoginParams } from '../../common/Session';
```

davwheat (approved):

SychO9 (commented):
Hmm, an event doesn't really seem appropriate to me. I'd rather we create a login validator, empty by default, and extensions can extend the validator instead.

imorland:
🤔 yeah... That would make more sense, actually. On it...

davwheat:
Why not both?

That would allow for processing of data from a user logging in, and the more traditional approach to validation.

SychO9:
I don't know, it's not apparent to me what the processing of data could be, and if whatever it is it couldn't be handled using the exsiting extension API (like we did here with the validator), so I prefer dealing with one use case at a time to avoid over-engineering it.

SychO9 (approved):

PR #3671, opened 2022-11-06
feat: Allow additional reset password params, introduce `ForgotPasswordValidator`

imorland:
Related to https://github.com/flarum/framework/pull/3670

Allows for 3rd party extensions to extend the functionality of the Forgot Password modal. ie Adding a captcha, etc.

`ForgotPasswordModal`:
- Extracts fields to an extendable function
- Extract params into an extendable function so that additional params can be added
`ForgotPasswordController`
- Replaces the validator factory with a new `ForgotPasswordValidator`, with the same rule as existed previously
- This can be extended by 3rd party extensions to add any additional rules, where required.

**Necessity**

**Confirmed**

**Required changes:**


luceos (approved):
More extensibility 💪

davwheat (approved):

PR #3678, opened 2022-11-13
feat(jest): create jest config package for unit testing

SychO9:
This is part one of two PRs to introduce tests to our JavaScript codebase. This PR lays the foundation by introducing a `flarum-jest-config` package and the necessary changes to begin unit testing. An example core `abbreviateNumber` util test is added as well.

* Anything I might be missing about introducing Jest testing.

Part two is here: #3679

**Necessity**

**Confirmed**

**Required changes:**


SychO9:
need to add the workflow

js-packages/jest-config/README.md
```diff
@@ -0,0 +1,14 @@
+# Jest config for Flarum
+
+This package generates a [Jest](https://jestjs.io/) config object to run unit & integration tests on Flarum extensions.
```
askvortsov1:
```suggestion
This package provides a [Jest](https://jestjs.io/) config object to run unit & integration tests on Flarum extensions.
```

js-packages/webpack-config/index.js
```diff
@@ -72,33 +72,11 @@ module.exports = function (options = {}) {
         {
           // Matches .js, .jsx, .ts, .tsx
           // See: https://regexr.com/5snjd
-          test: /\.(j|t)sx?$/,
+          test: /\.[jt]sx?$/,
```
askvortsov1, line 75: A lot cleaner, yay!

askvortsov1 (approved):
Wonderful, thanks so much for making this work! I think the only uncertainty I have is whether it would be better to put test code next to the source code, either in the same folder or in a `__tests__` subdir. That way, it's easier to keep track of source code having corresponding tests, although coverage should help with that too.

SychO9:
> Wonderful, thanks so much for making this work! I think the only uncertainty I have is whether it would be better to put test code next to the source code, either in the same folder or in a __tests__ subdir. That way, it's easier to keep track of source code having corresponding tests, although coverage should help with that too.

I don't have any strong opinions about this at the moment. I'm okay with it either way. We will probably find out as we write more tests in the future.

askvortsov1:
Let's keep as is for now then, LGTM!

PR #3684, opened 2022-11-18
fix: evaluated page title content

SychO9:
Properly escapes HTML entities when setting page titles. Otherwise, any decoded title content is evaluated as HTML code.

**Necessity**

**Confirmed**

### DIFF
--- a/framework/core/js/src/common/Application.tsx
+++ b/framework/core/js/src/common/Application.tsx
@@ -417,16 +417,15 @@ export default class Application {
 
     title = count + title;
 
-    // We set the title directly to the title element's innerHTML to properly display raw entities,
-    // while preventing XSS attacks from user input.
-    // https://github.com/flarum/framework/issues/3514
-    // https://github.com/flarum/framework/pull/3684
-    const titleNodes = document.getElementsByTagName('title');
-    if (titleNodes.length) {
-      titleNodes[0].innerHTML = title;
-    } else {
-      document.title = title;
-    }
+    // We pass the title through a DOMParser to allow HTML entities
+    // to be rendered correctly, while still preventing XSS attacks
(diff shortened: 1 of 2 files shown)


clarkwinkelmann (approved):
As mentioned on Discord I don't really like the use of `innerHTML` since its behavior on `<title>` isn't precisely described in the official web spec, but I'm hopeful we will be able to drop it in a future release once https://github.com/flarum/framework/issues/3685 is fixed.

I have not tested this specific PR but I played around with the same changes on my dev-main install and the changes proposed work well in my local tests.

framework/core/js/src/common/Application.tsx
```diff
@@ -410,16 +410,23 @@ export default class Application {
...
+    // We set the title directly to the title element's innerHTML to properly display raw entities,
+    // while preventing XSS attacks from user input.
+    // https://github.com/flarum/framework/issues/3514
+    // https://github.com/flarum/framework/pull/3684
+    const titleNodes = document.getElementsByTagName('title');
+    if (titleNodes.length) {
+      titleNodes[0].innerHTML = title;
+    } else {
+      document.title = title;
+    }
```
davwheat:
```suggestion
    // We pass the title through a DOMParser to allow HTML entities
    // to be rendered correctly, while still preventing XSS attacks
    // from user input by using a script-disabled environment.
	// https://github.com/flarum/framework/issues/3514
	// https://github.com/flarum/framework/pull/3684
    const parser = new DOMParser();
    const safeTitle = parser.parseFromString(title, 'text/html').body.innerText;
    
    document.title = safeTitle;
```

davwheat (changes requested):

davwheat (approved):

imorland (approved):

PR #3690, opened 2022-11-21
fix: provide web friendly session driver name

imorland:
#3593 introduced some basic system stats to the admin panel. The value for `session driver` returned unuseful content for the web, but was still useful for the CLI response.

web

Pass a `forWeb` boolean (default to `false`) to `identifySessionDriver` when being called from `AdminPayload` and return a shortened version of the text.

web

cli

**Necessity**

**Confirmed**

### DIFF
--- a/framework/core/src/Admin/Content/AdminPayload.php
+++ b/framework/core/src/Admin/Content/AdminPayload.php
@@ -113,7 +113,7 @@ class AdminPayload
         }
 
         $document->payload['queueDriver'] = $this->appInfo->identifyQueueDriver();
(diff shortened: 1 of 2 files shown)


SychO9 (approved):

davwheat (approved):

luceos:
I was looking at this PR and thought that the ApplicationInfoProvider should not contain any formatting at all. Wouldn't it make more sense to have some kind of Renderer class or make the info command handle the display of these info values?

SychO9:
Yeah, I agree the class should only return the data and the formatting should be done by the data consumer.

PR #3696, opened 2022-11-24
fix: column `id` can be ambiguous in group filter with extensions

iPurpl3x:
Changed 'id' to 'groups.id' to avoid `Column 'id' in where clause is ambiguous` error.

This error normally doesn't occur, but if an extension adds an `id` column to the `group_user` table, then this error would occur. 

(Adding that column helps when one wants to `chunkById()` on the `group_user` model)

Nothing is broken by this minor change.

**Necessity**

**Confirmed**

### DIFF
--- a/framework/core/src/User/Query/GroupFilterGambit.php
+++ b/framework/core/src/User/Query/GroupFilterGambit.php
@@ -61,7 +61,7 @@ class GroupFilterGambit extends AbstractRegexGambit implements FilterInterface
             }
         }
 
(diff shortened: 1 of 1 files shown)


SychO9:
You made a typo, it's `groups.id`

iPurpl3x:
Good catch @SychO9! I fixed it.

imorland (approved):

SychO9 (approved):
In the long term, we should look into prefixing all columns with table names in filters/gambits to avoid these issues.

PR #3698, opened 2022-11-27
feat: expose {time} to eventPost data, fix renamed tooltip

imorland:
**Fixes #0000**

`DiscussionRenamedPost` used the old style tooltip method, plus the tooltip was triggered by hovering anywhere on the text line. We now use the `Tooltip` component, and only on the discussion title itself.

Adds `time` to `EventPost` attrs, enabling the use of it in translations for event posts easily. See attached screenshots for core/bundled usages.

**Screenshot**
Before:
image

After:
image

Updated translations using the `time` attr
image

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/extensions/lock/locale/en.yml
+++ b/extensions/lock/locale/en.yml
@@ -29,8 +29,8 @@ flarum-lock:
 
     # These translations are displayed between posts in the post stream.
     post_stream:
-      discussion_locked_text: "{username} locked the discussion."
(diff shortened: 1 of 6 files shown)


askvortsov1 (approved):
I wonder if `rel_time` might be a better name; my first thought was that there's an "at" missing (e.g. x did y AT {time}), but that's not the case with humanTime

SychO9 (approved):

PR #3699, opened 2022-11-27
perf: transpile js to es6

davwheat:

- output ES6 Javascript instead of ES5

**All extension JS source code will need to be recompiled with an updated Webpack config.**

This will reduce our tech debt and overall bundle size due to unneeded ES5 polyfills included by Babel despite us not supporting any ES5 browsers.

**Necessity**

**Confirmed**

**Required changes:**


SychO9:
@davwheat is this complete?

davwheat:
Yeah, I think so. I only marked it as draft because we weren't 2.x-dev ready then.

SychO9 (approved):

PR #3709, opened 2022-12-13
feat(core): PHP 8.2 Support

### LINKED ISSUE #3462: Test with PHP 8.2
Need to:
- [x] move workflow definitions from `.github` to the monorepo (here) (https://github.com/flarum/framework/pull/3490).
- [ ] start testing with PHP 8.2

imorland:
**Fixes #3462**

Adds support for PHP 8.2, there are still deprecations we can't solve (in packages like swift mailer) so we disable deprecation warnings i the workflow tests, otherwise 8.2 should be ok to use.

**Necessity**

**Confirmed**


imorland:
Wikimedia/less.php deprecation warnings are awaiting a PR/new release: https://github.com/wikimedia/less.php/pull/86

pierres:
FYI: less.php has a new release which is compatible with PHP 8.2

SychO9:
Nice, thanks @pierres!

pierres:
Is there any way I can support this effort? So far Flarum works fine withe plattform override. Some tests are failing due deprecations in third party libraries (e.g. s9e/text-formatter).  swiftmailer might also be more challenging as the package needs to be replaced with e.g. smyfony/mailer.

It might also be an option to temporary allow deprecations within third party libraries.

luceos:
> Is there any way I can support this effort? So far Flarum works fine withe plattform override. Some tests are failing due deprecations in third party libraries (e.g. s9e/text-formatter). swiftmailer might also be more challenging as the package needs to be replaced with e.g. smyfony/mailer.
> 
> It might also be an option to temporary allow deprecations within third party libraries.

I think we might need to wait for the upgrade of Flarum to Laravel 9 (or 10 would make more sense as that's going to be released soon). L 9+ will replace swiftmailer with symfony mailer already. As far as s9e goes, I think it should be just fine, if not JoshyPHP is usually extremely quick on making patches to support new php versions.

What you can do? I think what's missing here, is a detailed list of everything that is still not working for php 8.2 so that we can resolve them one by one. However if an upgrade to L9 (at the least) seems to be required (due to swiftmailer for instance) we will have to wait for Flarum 2.0 I'm afraid, too much would break..

SychO9:
https://github.com/s9e/TextFormatter/pull/212

SychO9 (approved):

SychO9:
There are still deprecation warnings from swift mailer that we can't do anything about. So moving forward for 1.x the recommendation will be to disable deprecation warnings (this is supposed to be the case for production instances by default anyway). Other than that, 8.2 is good to go.

PR #3712, opened 2022-12-17
feat: Admin User Search

### LINKED ISSUE #3711: Admin user search
**Description**
The user list on the admin panel currently shows all users ordered by ID. It would be a big improvement to be able to search through it based on username and use the available search gambits. (https://flarum.test/admin#/users)

**What needs to be done**
The logic/endpoint for searching on the backend is already available, so we just need to hit the API with a search query from the frontend.
* First, we need to add the search input above the table (https://github.com/flarum/framework/blob/0e2053da1f7e1b9e1eb63221a6a186bb6fee355e/framework/core/js/src/admin/components/UserListPage.tsx#L94)
* Then we need the input to store the updated value on a state variable, we an use a `Stream` property and the `bidi` attribute for that, here's an example with the `nameSingular` input value: https://github.com/flarum/framework/blob/cf818aae9e65d713d889b0dfe1ca7e0fd16b5ca4/framework/core/js/src/admin/components/EditGroupModal.tsx#L74
* Finally, we need to add the input value to the query (https://github.com/flarum/framework/blob/0e2053da1f7e1b9e1eb63221a6a186bb6fee355e/framework/core/js/src/admin/components/UserListPage.tsx#L340-L345) it would need to be under `filter: { q: query }`
* We might need to set a debounce to prevent xhr requests from firing too fast when typing.

OrdinaryJellyfish:
**Fixes #3711**

* Adds search input to admin users list

* If the search input works as intended

**Screenshot**

**Necessity**

**Confirmed**

### DIFF
--- a/framework/core/js/src/admin/components/UserListPage.tsx
+++ b/framework/core/js/src/admin/components/UserListPage.tsx
@@ -14,8 +14,11 @@ import type User from '../../common/models/User';
 import ItemList from '../../common/utils/ItemList';
 import classList from '../../common/utils/classList';
 import extractText from '../../common/utils/extractText';
+import Stream from '../../common/utils/Stream';
 
 import AdminPage from './AdminPage';
+import { IPageAttrs } from '../../common/components/Page';
+import { debounce } from '../../common/utils/throttleDebounce';
 
 type ColumnData = {
   /**
@@ -32,6 +35,8 @@ type ColumnData = {
  * Admin page which displays a paginated list of all users on the forum.
  */
 export default class UserListPage extends AdminPage {
+  userSearchQuery = Stream('');
+
   /**
    * Number of users to load per page.
    */
@@ -91,6 +96,7 @@ export default class UserListPage extends AdminPage {
     const columns = this.columns().toArray();
 
     return [
+      <input className="FormControl" placeholder={app.translator.trans('core.admin.users.search_placeholder')} bidi={this.userSearchQuery} />,
       <p class="UserListPage-totalUsers">{app.translator.trans('core.admin.users.total_users', { count: this.userCount })}</p>,
       <section
         class={classList(['UserListPage-grid', this.isLoadingPage ? 'UserListPage-grid--loadingPage' : 'UserListPage-grid--loaded'])}
@@ -339,9 +345,10 @@ export default class UserListPage extends AdminPage {
 
     app.store
       .find<User[]>('users', {
+        filter: { q: this.userSearchQuery() },
         page: {
           limit: this.numPerPage,
-          offset: pageNumber * this.numPerPage,
+          offset: pageNumber * this.numPerPage
         },
       })
       .then((apiData) => {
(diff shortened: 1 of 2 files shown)


framework/core/js/src/admin/components/UserListPage.tsx
```diff
@@ -14,8 +14,10 @@ import type User from '../../common/models/User';
 import ItemList from '../../common/utils/ItemList';
 import classList from '../../common/utils/classList';
 import extractText from '../../common/utils/extractText';
+import Stream from '../../common/utils/Stream';
```
SychO9: we can remove this now

framework/core/js/src/admin/components/UserListPage.tsx
```diff
@@ -32,6 +34,8 @@ type ColumnData = {
  * Admin page which displays a paginated list of all users on the forum.
  */
 export default class UserListPage extends AdminPage {
+  query: string = '';
```
SychO9: let's make this private

framework/core/js/src/admin/components/UserListPage.tsx
```diff
@@ -91,6 +95,18 @@ export default class UserListPage extends AdminPage {
...
     return [
+      <div className="Search-input">
+        <input
+          className="FormControl SearchBar"
+          type="search"
+          placeholder={app.translator.trans('core.admin.users.search_placeholder')}
+          oninput={(e: InputEvent) => {
+            this.isLoadingPage = true;
+            this.query = (e?.target as HTMLInputElement)?.value;
+            debounce(250, () => this.loadPage(this.pageNumber))();
```
SychO9: the debounce isn't actually working right now.
That's because we create a new denounced function on every re-render, and the only way debounce works is if we use the same debounce function we create.

To amend that, we can set the denounced function on a private property then we can use that instead.

```ts
private throttledSearch = debounce(250, () => this.loadPage(0));
```

framework/core/locale/core.yml
```diff
@@ -242,6 +242,7 @@ core:
     # These translations are used for the users list on the admin dashboard.
     users:
       description: A paginated list of all users on your forum.
+      search_placeholder: Search users
```
SychO9: We organize locale alphabetically, so let's push this right above title.

Additionally, since we have another entry that translates to `Search users` (*look up `search_users_label`*) let's create a common entry under the `ref:` group and reuse that one in the other two entries, namely (`search_placeholder` and `search_users_label`).

For an example on reusing ref entries, look up `=> core.ref`

SychO9 (commented):
Looking good!

Some comments

SychO9 (approved):
Looks good to me!

Design-wise I think the entire page could be improved, especially once we add in more features, but that can be done at a later time.

luceos (approved):
Excellent work @OrdinaryJellyfish 👍 

@SychO9 feel free to merge.

PR #3713, opened 2022-12-19
The negate field doesn't get used, which means you cant exclude tags

OwenMelbz:
Currently you cannot filter discussions/posts by excluding tag ids, as the `$negate` boolean is ignored.

`/api/posts?sort=-createdAt&filter[tag]=14` to include tag 14

and 

`/api/posts?sort=-createdAt&filter[tag]=-14` to exclude tag 14

-------

Unless there's a different way to handle this?


davwheat (approved):
Makes sense to me!

imorland (approved):

askvortsov1 (approved):
Makes sense, thanks!

PR #3718, opened 2023-01-09
fix: search results getting dismissed in iOS through touch

SychO9:
**Fixes #0000**

**Screenshot**

**Necessity**

**Confirmed**

**Required changes:**


askvortsov1 (approved):
Makes sense, don't have an iOS device though so can't test locally

SychO9:
This is actually untested so let's avoid merging for now

imorland:
@SychO9 Sorry for the delay, I've had a chance to test this locally today. 

The issue is not present in Safari (iOS 16.3) on either `main` or this branch. However, the issue is present when using the `Orion` browser, even with this change.

PR #3724, opened 2023-01-22
feat(seo): shift h1 tag from logo to discussion title

luceos:
Many times have we seen opponents of using h2 as the discussion title. Although my own SEO knowledge is limited, I have seen the importance of structuring pages according to the content you wish to prioritize. If we only take that into consideration there is zero reason for the app-wide identical logo to take precedence over any other heading.

- https://discuss.flarum.org/d/31209-discussion-page-is-not-seo-friendly

This change makes the logo a standard (visually identical) element, and makes the discussion hero title a h1.

- [ ] Do we want to do this (for 1.x)?
- [ ] Do we need to make the logo element a different one?

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/extensions/tags/views/frontend/content/tag.blade.php
+++ b/extensions/tags/views/frontend/content/tag.blade.php
@@ -2,7 +2,7 @@
 
 <div class="container">
     <h1>{{ $tag->name }}</h1>
-    <h2>{{ $tag->description }}</h2>
+    <p>{{ $tag->description }}</p>
 
     <ul>
         @foreach ($apiDocument->data as $discussion)
--- a/extensions/tags/views/frontend/content/tag.blade.php
+++ b/extensions/tags/views/frontend/content/tag.blade.php
@@ -1,8 +1,8 @@
 @inject('url', 'Flarum\Http\UrlGenerator')
 
 <div class="container">
-    <h2>{{ $tag->name }}</h2>
-    <p>{{ $tag->description }}</p>
+    <h1>{{ $tag->name }}</h1>
+    <h2>{{ $tag->description }}</h2>
 
     <ul>
         @foreach ($apiDocument->data as $discussion)
--- a/framework/core/js/src/forum/components/WelcomeHero.tsx
+++ b/framework/core/js/src/forum/components/WelcomeHero.tsx
@@ -39,8 +39,8 @@ export default class WelcomeHero extends Component<IWelcomeHeroAttrs> {
           />
 
           <div class="containerNarrow">
-            <h2 class="Hero-title">{app.forum.attribute('welcomeTitle')}</h2>
-            <div class="Hero-subtitle">{m.trust(app.forum.attribute('welcomeMessage'))}</div>
+            <h1 class="Hero-title">{app.forum.attribute('welcomeTitle')}</h1>
+            <h2 class="Hero-subtitle">{m.trust(app.forum.attribute('welcomeMessage'))}</h2>
           </div>
         </div>
       </header>
--- a/framework/core/js/src/forum/components/WelcomeHero.tsx
+++ b/framework/core/js/src/forum/components/WelcomeHero.tsx
@@ -40,7 +40,7 @@ export default class WelcomeHero extends Component<IWelcomeHeroAttrs> {
 
           <div class="containerNarrow">
             <h1 class="Hero-title">{app.forum.attribute('welcomeTitle')}</h1>
-            <h2 class="Hero-subtitle">{m.trust(app.forum.attribute('welcomeMessage'))}</h2>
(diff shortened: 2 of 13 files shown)


SychO9 (commented):
The `.Hero h2` to `.Hero h1` can be considered a breaking change styling-wise, but IMO a minor one that is easily fixable in custom designs.

Can you please also change other hero titles? (`WelcomeHero` `TagHero` and `UserCard`)

luceos:
Yeah I couldn't find a way to change this ti draft. Sorry.

luceos:
Looking at my code, it seems, making the change completely, does affect quite some rendered HTML. If we merge this for 1.7 (or 1.x in general) an entry for the Dev Upgrade Guide is required and a possible warning for end users to report issues related to it.

extensions/tags/views/frontend/content/tag.blade.php
```diff
@@ -1,8 +1,8 @@
 @inject('url', 'Flarum\Http\UrlGenerator')
 
 <div class="container">
-    <h2>{{ $tag->name }}</h2>
-    <p>{{ $tag->description }}</p>
+    <h1>{{ $tag->name }}</h1>
+    <h2>{{ $tag->description }}</h2>
```
SychO9: I would revert this, description should remain a paragraph rather than a header

```suggestion
    <h1>{{ $tag->name }}</h1>
    <p>{{ $tag->description }}</p>
```

framework/core/js/src/forum/components/WelcomeHero.tsx
```diff
@@ -39,8 +39,8 @@ export default class WelcomeHero extends Component<IWelcomeHeroAttrs> {
           />
 
           <div class="containerNarrow">
-            <h2 class="Hero-title">{app.forum.attribute('welcomeTitle')}</h2>
-            <div class="Hero-subtitle">{m.trust(app.forum.attribute('welcomeMessage'))}</div>
+            <h1 class="Hero-title">{app.forum.attribute('welcomeTitle')}</h1>
+            <h2 class="Hero-subtitle">{m.trust(app.forum.attribute('welcomeMessage'))}</h2>
```
SychO9: same here

```suggestion
            <h1 class="Hero-title">{app.forum.attribute('welcomeTitle')}</h1>
            <div class="Hero-subtitle">{m.trust(app.forum.attribute('welcomeMessage'))}</div>
```

SychO9 (commented):
> Looking at my code, it seems, making the change completely, does affect quite some rendered HTML

I would reduce changes, for example, admin side components don't matter.

extensions/tags/views/frontend/content/tag.blade.php
davwheat: Might be good to add `role="doc-subtitle"` to the `<p>` tag too :)

framework/core/js/src/forum/components/WelcomeHero.tsx
luceos: I don't agree. Looking at the structure of the page, the next nested item (h3's) are user cards etc. I think if you have a WelcomeHero on the page then the sub title is definitely of a higher order than the content of each discussion listed. SEO is (mainly) about prioritizing content, subtitle - look at the class - is the second most important text.

extensions/tags/views/frontend/content/tag.blade.php
luceos: I was wondering that for the logo element actually @davwheat and researched that (any thoughts?), thanks for providing this one, very useful!

luceos:
> I would reduce changes, for example, admin side components don't matter.

I think we should keep our structure uniform across frontends. If we make the change it's easier to explain priority changed across all frontends instead of just one.

framework/core/js/src/forum/components/WelcomeHero.tsx
SychO9: 🤔 That doesn't make sense though, the hero subtitle is not a heading, it's the content after the hero heading, the content of the hero, a paragraph here would make more sense. I need to read up on seo practices..

Subtitle being the class name is just naming.

davwheat: Subtitles also aren't really numbered headings. That causes confusion for accessibility. https://css-tricks.com/html-for-subheadings-and-headings/

rob006: > the next nested item (h3's) are user cards etc

I think this is another thing that should be changed. Discussion titles on discussions list should use H2 header. And posts inside discussion should not have automatic headers at all, since there is no good one (poster name is terrible header in that case), and it is better to rely on users to use proper headers inside of posts.

SychO9 (approved):

SychO9:
@rob006 please feel free to contribute the other changes you mention in another PR 🙏🏼.

PR #3726, opened 2023-02-04
Update forum.less to fix the misalignment of the choose tags button

iamdarkle:
**Fixes #0000**

This little PR is simply to remove a CSS rule.

I'm not sure of the reason for this `vertical-align: 2px;` (maybe at some point you had a good reason), but currently this makes the button display misaligned with respect to the Discussion Title input (although in this case it can go unnoticed), but in conjunction with other extensions it's quite noticeable as it's the only misaligned button, as you can see in the case of the screenshot with `flamarkt/taxonomies`

Is there any need to keep the `vertical-align: 2px;`?

**Screenshots**

Current with `vertical-align: 2px;`
Screenshot 2023-02-04 195405

Without `vertical-align: 2px;` (This PR)
Screenshot 2023-02-04 195425

To add other extensions with which the issue is clearly visible:
- `fof/polls`
- `fof/discussion-language`

**Necessity**

**Confirmed**

### DIFF
--- a/extensions/tags/less/forum.less
+++ b/extensions/tags/less/forum.less
@@ -36,7 +36,6 @@
 }
 .DiscussionComposer-changeTags {
   margin-right: 15px;
-  vertical-align: 2px;
 
   &.disabled {
     opacity: 0.5;


askvortsov1 (approved):
Less code and less bugs? Sounds like a win to me

PR #3727, opened 2023-02-07
fix(phpstan): adapt phpstan package for extension use

n-peugnet:
- fix "Stub file does not exist" error message as reported in <https://github.com/flarum/docs/pull/441#issuecomment-1419668760>
- fix "Ignored error pattern was not matched in reported errors" error
  messages as reported in <https://github.com/flarum/docs/pull/441#issuecomment-1420903036>

### DIFF
--- a/php-packages/phpstan/phpstan-baseline.neon
+++ b/php-packages/phpstan/phpstan-baseline.neon
@@ -1,17 +1,23 @@
 parameters:
 	ignoreErrors:
 		# Remove this group below with larastan 2.0 (i.e Flarum 2.0)
-		- "#Relation '[A-z_-]+' is not found in [A-z\_]+ model.#"
-		- '#^Parameter \#1 \$query of method [A-z_<>\\]+\:\:union\(\) expects [A-z_<> .,|\\]+ given\.$#'
-		- '#^Parameter \#1 \$query of method [A-z_<>\\]+\:\:joinSub\(\) expects [A-z_<> .,|\\]+ given\.$#'
+		- message: "#Relation '[A-z_-]+' is not found in [A-z\_]+ model.#"
+		  reportUnmatched: false
+		- message: '#^Parameter \#1 \$query of method [A-z_<>\\]+\:\:union\(\) expects [A-z_<> .,|\\]+ given\.$#'
+		  reportUnmatched: false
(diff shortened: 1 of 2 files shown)


php-packages/phpstan/phpstan-baseline.neon
```diff
@@ -1,17 +1,23 @@
...
 		# Remove this group below with larastan 2.0 (i.e Flarum 2.0)
-		- "#Relation '[A-z_-]+' is not found in [A-z\_]+ model.#"
-		- '#^Parameter \#1 \$query of method [A-z_<>\\]+\:\:union\(\) expects [A-z_<> .,|\\]+ given\.$#'
-		- '#^Parameter \#1 \$query of method [A-z_<>\\]+\:\:joinSub\(\) expects [A-z_<> .,|\\]+ given\.$#'
+		- message: "#Relation '[A-z_-]+' is not found in [A-z\_]+ model.#"
+		  reportUnmatched: false
+		- message: '#^Parameter \#1 \$query of method [A-z_<>\\]+\:\:union\(\) expects [A-z_<> .,|\\]+ given\.$#'
+		  reportUnmatched: false
+		- message: '#^Parameter \#1 \$query of method [A-z_<>\\]+\:\:joinSub\(\) expects [A-z_<> .,|\\]+ given\.$#'
+		  reportUnmatched: false
```
SychO9, line 9: rather than ignoring each error individually, let's use the global report unmatched parameter:
https://phpstan.org/user-guide/ignoring-errors#reporting-unused-ignores

SychO9 (changes requested):
Thanks for the PR!

php-packages/phpstan/phpstan-baseline.neon
n-peugnet: The reason why I chose to ignore each of them instead of using the global parameter, is because I am not sure if it only affects the `ignoreErrors` of the current file of if it sets it also for user defined ones. I for one, would like to keep it enabled globally, so that my own unmatched `ignoreErrors` are still reported.

I will quickly check this.

n-peugnet: Indeed, by setting the parameter globally in this file, it also affects the `ignoreErrors` defined in my own phpstan config file. So I would prefer to keep it as so.

SychO9 (approved):
Sounds reasonable!

PR #3732, opened 2023-02-11
feat(seo): Do not use h3 header for poster author in posts stream

rob006:
See https://github.com/flarum/framework/pull/3724#discussion_r1083536881

However this is more tricky, as some extensions rely on current markup. During brief testing I found 2 problems:

1. Best Answer extension uses h3 for displaying username and it relies on styles provided by core. That is why I kept old styles for h3 tag - additional styles should not be a big problem and we keep BC this way. Redundant styles could be removed on 2.0 release.
2. FoF Gamification crashed completely, because it searches for `h3` tag: 
   https://github.com/FriendsOfFlarum/gamification/blob/aad90d175fe55b3609de8c7649f9364765a37853/js/src/forum/components/RankingsPage.js#L76
   This could be quite easily fixed in extension, but BC break is obvious and there could be more extensions affected by this change. If we want to keep BC here, I don't think there is other solution than postponing this change to 2.0.

In future it would be better to add CSS classes everywhere and discourage relying on markup, as used tag may always change. Even some basic cases like targeting link using `a` selector may be problematic, because in future this link may be changed to button instead.

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/framework/core/js/src/forum/components/PostUser.js
+++ b/framework/core/js/src/forum/components/PostUser.js
@@ -22,9 +22,9 @@ export default class PostUser extends Component {
     if (!user) {
       return (
         <div className="PostUser">
-          <div class="PostUser-name">
+          <h3 class="PostUser-name">
             {avatar(user, { className: 'PostUser-avatar' })} {username(user)}
-          </div>
+          </h3>
         </div>
       );
     }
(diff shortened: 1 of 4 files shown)


luceos (approved):

SychO9:
Based on the PR description and on second thought, this would be better suited for 2.0 considering a lot of extensions add content next to the user name. If it was just about breaking styles in extensions that'd be fine, but JS is more problematic, and I'm pretty sure it's not just the FoF extensions mentioned.

rob006:
I could revert change of tag name, so it will be  `h3` again. But the rest of changes should be fine and I think it is worth to merge them to 1.x, so extensions authors could target elements using class name and prepare earlier for upcoming BC breaks.

SychO9:
@rob006 agreed, let's do that!

rob006:
Done.

PR #3733, opened 2023-02-11
fix: infinite scroll not initialized for notifications on big screens

rob006:

On QHD screens `70vh` is bigger than space taken by 10 notifications. As a result there is no scroll for this dropdown, and as a result infinite scroll do not work in this place. I mentioned this on Discord a few months ago: https://discord.com/channels/360670804914208769/707030467450372166/1010096800511836261

This PR sets max height of this dropdown to 800px and increases number of notifications loaded in this dropdown, to make sure it is always filled.

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/framework/core/js/src/forum/states/NotificationListState.ts
+++ b/framework/core/js/src/forum/states/NotificationListState.ts
@@ -4,7 +4,7 @@ import Notification from '../../common/models/Notification';
 
 export default class NotificationListState extends PaginatedListState<Notification> {
   constructor() {
-    super({}, 1, 10);
+    super({}, 1, 20);
   }
 
   get type(): string {
--- a/framework/core/less/forum/NotificationsDropdown.less
+++ b/framework/core/less/forum/NotificationsDropdown.less
@@ -3,7 +3,7 @@
     padding: 0;
 
     .NotificationList-content {
-      max-height: 70vh;
+      max-height: ~"min(70vh, 800px)";
       overflow: auto;
     }
   }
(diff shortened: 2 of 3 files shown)


iamdarkle:
Coincidentally, I experienced this for the first time last week, and it was on a 1080p screen, in a particular case where the whole list of notifications were from the same discussion.

askvortsov1 (changes requested):
The max-height change makes sense, but I'm hesitant to remove pagination-by-default from the notifications list controller like that. Couldn't that be a frontend change, with a different `limit` query param depending on screen size?

rob006:
> but I'm hesitant to remove pagination-by-default from the notifications list controller like that

This PR does not remove pagination, it changes page size from 10 to 20 items (this is default size from [AbstractSerializeController](https://github.com/flarum/framework/blob/1200c1f72f4feaa3c9a0054213c5930657c738f2/framework/core/src/Api/Controller/AbstractSerializeController.php#L60)).

SychO9:
rather than 20, 12 should be enough for this case? notifications listing performance is still not great, so ideally we should avoid doubling items per page.

rob006:
12 will not be enough in some cases. I have many notifications like this:

c7afa43a

In this scenario you need ate least 14-15 notifications to fill 800px of dropdown. And probably there could be cases when you have 4 notifications grouped like this - then you need even more notifications. 20 is just safe value, and performance is not terrible in my case (it is faster than discussions list or single discussion view in most cases).

askvortsov1 (approved):
I think this all seems reasonable. I don't see why notifications should have a special case for the default limit.

SychO9 (approved):
> In this scenario you need ate least 14-15 notifications to fill 800px of dropdown. 

Alright then

PR #3735, opened 2023-02-15
Add basic tests for PHP and JS Translators

rob006:
Failing tests to demonstrate the discrepancy between JS and PHP translators.

Related issues: https://github.com/flarum/framework/issues/3702, https://github.com/flarum/framework/issues/3734, https://github.com/flarum/framework/issues/3685

**Necessity**

**Confirmed**

**Required changes:**


rob006:
@askvortsov1 Is there any particular reason why https://www.npmjs.com/package/@ultraq/icu-message-formatter was chosen for ICU formatting? It seems to lack some basic features. I tested https://www.npmjs.com/package/format-message and https://www.npmjs.com/package/intl-messageformat and both looked more mature and reliable. And they have separate packages to parse message to token-like format - it could be used to handle rich-syntax support with less magic than current implementation, which relies on string operations on original message.

SychO9:
@askvortsov1 sorry for the second ping, but any chance you recall? I'd like to try to look into this before 1.8 is over

askvortsov1:
> @askvortsov1 sorry for the second ping, but any chance you recall? I'd like to try to look into this before 1.8 is over

Hey, sorry, missed this one. Iirc, it came down to package size: the ultraq one is 6kb, vs 22kb and 28kb respectively. That being said, with the (broken) rich text changes, it comes out to ~10kb, and I don't know that there's a trivial fix. It would probably be worth it to switch to a more mature, widely-used package.

PR #3744, opened 2023-02-22
feat: add user creation to users list page

davwheat:
Supersedes https://github.com/flarum/framework/pull/2998

Progresses https://github.com/flarum/framework/issues/3742

- adds new user creation modal to Admin
- adds button to open this on the users page

**Screenshot**

image

**Necessity**

**Confirmed**


luceos:
Do we want to allow setting a password by default? Wouldn't it make more sense to create a temporary one instead? And then hook into the password reset logic?

davwheat:
I think both are valid options, and I was thinking about that myself.

Maybe an option to "force user to reset password to log in"? Question is whether we want to automatically fire the reset pwd email or just state that the user should reset their own password?

Or should we have a separate "someone's made an account for you on xyz", but that's straying into extension territory imo.

davwheat:
I've opted to generate a random password on the client-side and provide that, but I do wonder if it's better to generate it in the backend and automate the forgotten password email instead? But I feel a forgotten password for a new user on a site they haven't used yet would be very confusing.

framework/core/locale/core.yml
```diff
@@ -51,6 +49,17 @@ core:
...
+    # These translations are used in the Create User modal.
+    create_user:
+      bulk_add_label: Keep modal open after creation?
+      email_placeholder: => core.ref.email
+      email_confirmed_label: Require user to confirm this email?
+      password_placeholder: => core.ref.password
+      submit_button: Create user
+      title: Create new user
+      use_random_password: Generate random password?
+      username_placeholder: => core.ref.username
```
SychO9: I think the switch labels would be better not being questions 🤔

framework/core/locale/core.yml
```diff
@@ -712,7 +720,6 @@ core:
 
   # Translations in this namespace are used in emails sent by the forum.
   email:
-
```
SychO9: formatting made its way in 🙈

framework/core/js/src/admin/components/CreateUserModal.tsx
```diff
@@ -0,0 +1,241 @@
...
+          type="password"
+          autocomplete="new-password"
+          placeholder={passwordLabel}
+          aria-label={passwordLabel}
+          bidi={this.password}
+          disabled={this.loading || this.password() === null}
+        />
+      </div>,
+      40
+    );
```
SychO9: Let's split the password generation switch from the password input form group into its own form group. more extensible that way too 👍🏼

framework/core/js/src/admin/components/CreateUserModal.tsx
```diff
@@ -0,0 +1,241 @@
...
+
+    items.add(
+      'bulkAdd',
+      <div className="Form-group CreateUserModal-bulkAdd">
+        <Switch name="bulkAdd" disabled={this.loading} state={this.bulkAdd()} onchange={(checked: boolean) => this.bulkAdd(checked)}>
+          {app.translator.trans('core.admin.create_user.bulk_add_label')}
+        </Switch>
+      </div>,
+      0
+    );
```
SychO9: I have been using laravel filaments lately and quite like how they use a `Create & create another button` (https://demo.filamentphp.com/shop/products/create) maybe we might consider a similar approach here? wdyt

imorland:
@davwheat what's the situation here? Do you have time to complete these changes so this can be merged, or shall we tackle it?

davwheat:
Someone else can take it, if they'd like. I don't have the time this week or next.

SychO9 (approved):

PR #3749, opened 2023-03-10
chore(notification): hide actions on empty state

datlechin:
Hide actions when notification is empty

**Fixes #0000**

**Screenshot**

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/framework/core/js/src/forum/components/NotificationList.js
+++ b/framework/core/js/src/forum/components/NotificationList.js
@@ -34,7 +34,7 @@ export default class NotificationList extends Component {
     const state = this.attrs.state;
 
     if (!state.hasItems()) {
-      return;
+      return items;
     }
 
     items.add(
--- a/framework/core/js/src/forum/components/NotificationList.js
+++ b/framework/core/js/src/forum/components/NotificationList.js
@@ -33,6 +33,10 @@ export default class NotificationList extends Component {
     const items = new ItemList();
     const state = this.attrs.state;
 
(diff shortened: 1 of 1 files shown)


SychO9 (approved):
Thanks for the PR!

I believe it's a fine change.

davwheat (changes requested):
We should still return the item list even if we exit that func early.

framework/core/js/src/forum/components/NotificationList.js
```diff
@@ -33,6 +33,10 @@ export default class NotificationList extends Component {
     const items = new ItemList();
     const state = this.attrs.state;
 
+    if (!state.hasItems()) {
+      return;
```
davwheat:
```suggestion
      return items;
```

davwheat (changes requested):

framework/core/js/src/forum/components/NotificationList.js
```diff
@@ -33,6 +33,10 @@ export default class NotificationList extends Component {
     const items = new ItemList();
     const state = this.attrs.state;
 
+    if (!state.hasItems()) {
+      return items;
+    }
+
```
SychO9, line 39: actually, it would probably be better to wrap the two available items in the conditional, to not completely prevent adding for example a link to notification preferences as an item from an extension?

SychO9 (approved):

PR #3757, opened 2023-03-12
chore: search input padding right to much

datlechin:
image

**Fixes #0000**

**Screenshot**

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/framework/core/less/common/Search.less
+++ b/framework/core/less/common/Search.less
@@ -75,7 +75,6 @@
     float: left;
     width: 225px;
     padding-left: 32px;
-    padding-right: 32px;
     transition: var(--transition), width 0.4s;
     box-sizing: inherit !important;
   }


davwheat:
I believe this was done to house the "clear" button. Could we move that button into the padding area instead of removing the padding?

datlechin:
@davwheat the 'clear' button created by browser. see the search input in the front
image

davwheat:
With the frontend forum search, a `.Search-clear` button appears created by Flarum itself

image

This is the reason we have the right padding. In reality, we should remove the `.Search-input` class from the users list search instead, since it doesn't use the Search component, but this will remove the search icon and such from it.

https://github.com/flarum/framework/blob/a6a067ad48cc9d7bcadb903663d7e69280c5d99f/framework/core/js/src/admin/components/UserListPage.tsx#L119-L130

---

@SychO9 since this issue isn't a major one, what do you think about using [a new CSS feature](https://caniuse.com/css-has) to fix this? (Only lacks support in Firefox right now.)

This will set padding-right to 8px when there's no `.Search-clear` button, so will preserve the correct padding on the forum, while allowing us to show the search icon on other fields.

```less
.Search-input:not(:has(.Search-clear)) input {
    padding-right: 8px;
}
```

image

image

PR #3759, opened 2023-03-13
feat: conditional extenders

SychO9:
Adds support for conditional extenders, needed especially when adding extenders when certain extensions are enabled. (needed for tag mentions).

Here are some usage examples:
```php
return [
    (new Extend\Conditional)
        ->whenExtensionEnabled('flarum-tags', [
                (new Extend\Formatter)
                    ->parse(FormatTagMentions::class)
                    ->unparse(UnparseTagMentions::class),

                (new Extend\ApiController(ListPostsController::class)
                    ->load('mentionsTags'),

                (new Extend\Model(Post::class)
                    ->belongsToMany('mentionsTags', Tag::class, 'post_mentions_tag', 'post_id', 'mentions_tag_id'),

                (new Extend\ApiSerializer(BasicPostSerializer::class))
                    ->hasMany('mentionsTags', TagSerializer::class),
        ]),
]
```

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/framework/core/src/Extend/Conditional.php
+++ b/framework/core/src/Extend/Conditional.php
@@ -1,5 +1,12 @@
 <?php
 
+/*
+ * This file is part of Flarum.
+ *
+ * For detailed copyright and license information, please view the
+ * LICENSE file that was distributed with this source code.
+ */
+
 namespace Flarum\Extend;
 
 use Flarum\Extension\Extension;
--- a/framework/core/src/Extend/Conditional.php
+++ b/framework/core/src/Extend/Conditional.php
@@ -1,0 +1,54 @@
+<?php
+
+namespace Flarum\Extend;
+
+use Flarum\Extension\Extension;
+use Flarum\Extension\ExtensionManager;
+use Illuminate\Contracts\Container\Container;
+
+class Conditional implements ExtenderInterface
+{
+    /**
+     * @var array<array{condition: bool|callable, extenders: ExtenderInterface[]}>
+     */
+    protected $conditions = [];
+
+    /**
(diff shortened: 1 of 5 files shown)


imorland (approved):
I've been considering this for a while now actually. Great job 👍

SychO9:
Inspired by @davwheat & @luceos 🤗

davwheat (approved):

framework/core/src/Extend/Conditional.php
```diff
@@ -0,0 +1,61 @@
...
+{
+    /**
+     * @var array<array{condition: bool|callable, extenders: ExtenderInterface[]}>
+     */
+    protected $conditions = [];
+
+    /**
+     * @param ExtenderInterface[] $extenders
+     */
+    public function whenExtensionEnabled(string $extensionId, array $extenders): self
```
luceos, line 26: The signature doesn't match the PR comment example. Your example uses a callable as the second argument, not an array. The array is the return response from the callable yes. I assume your example is wrong then..

Also the phpdoc is incorrect, there are two args.

luceos (changes requested):
Good job. Please do cleanups in different pr's next time, makes reviewing a ton easier 🙈

framework/core/src/Extend/Conditional.php
SychO9: yea updated the example thanks

phpdoc doesn't matter, we only add typings for what isn't typed, we can't vanilla type array as ExtenderInterface[], we've being doing this since 1.0 to avoid unnecessary duplication. And the phpdoc standard allows it.

PR #3765, opened 2023-03-17
feat: frontend content flexible order priorities

garygreen:
This allows access to page/controller loaded content when using the Frontend `content` extender - which allows developers to fully customise the document programatically, e.g. able to see `Discussion`, `Tag`, etc.

Previous to this PR this was not possible:

```php
        ->content(function(Document $document, Request $request) {
                  // Not possible to get the Discussion, or Tag, or anything that the controller loaded here from the `document`
                 $document->title; // always null, despite viewing a Discussion page.
                 dd($document);
        });
```

**Before** - `$document` - notice `title` is missing
image

**After**  - notice Discussion information is filled on the document and are able to access the model data.

image

**Use Cases**

This has many - one we are looking at is to add a `<meta>` SEO noindex tag to our Discussion page if we consider the posts there to have "thin content" value. With this PR, it's easily possible to get acccess to the Discussion document and relevant posts. Prior to this PR, it's not easily possible without manually loading the Discussion model, etc.

This potentially introduces a breaking behavioural change if people were relying on the old behaviour. I think it's a minor change though.

**Necessity**

**Confirmed**

**Required changes:**


framework/core/src/Forum/ForumServiceProvider.php
```diff
@@ -126,8 +126,14 @@ public function register()
             return $assets;
         });
 
-        $this->container->bind('flarum.frontend.forum', function (Container $container) {
-            return $container->make('flarum.frontend.factory')('forum');
+        $this->container->bind('flarum.frontend.forum', function (Container $container, array $args = []) {
```
SychO9: why a multi-dimensional array of `$args` instead of directly the content?

SychO9 (commented):
Hi, thanks for the PR!

This would fix [flarum/issue-archive#199](https://github.com/flarum/issue-archive/issues/199)

I don't know yet whether this is BC compatible or not, so need to check 🤔

garygreen:
@SychO9 It does indeed fix https://github.com/flarum/issue-archive/issues/199 - I hadn't seen that issue up until now, glad there is a discussion for this. This is a simple PR that allows devs to have complete final control over the state of the `Document`.

It might be a BC like you say, as some may rely on the current behaviour of `Document` missing lots of values, and expecting it to override stuff that was set. That is quite unlikely though, imo. If it is a concern then it can be documented in an upgrade guide, or this could target 2.0 (how far away is that? hopefully not too far because we need this PR!) 😂

framework/core/src/Forum/ForumServiceProvider.php
SychO9: we could pass the $content directly rather then an arrya I believe

SychO9 (commented):
I have added priorities to the PR so that it's a lot more flexible, any developer can now choose to add content before any other content or after using priority numbers.

SychO9 (approved):

PR #3767, opened 2023-03-18
fix(mentions): mentions keep firing after a reply

### LINKED ISSUE #3766: Mentions fire xhr after completing the mentions element
### Current Behavior

When you mention someone (perfectly visible when using "reply") and you continue typing after that, by dividing the mention and the rest of your text with a space, XHR's keep firing continuously. This causes a huge overload on the server.

### Steps to Reproduce

1. Reply to a post
2. Continue typing a reply
3. Inspect network tab

### Expected Behavior

No XHR should be fired after the mention block is finished.

### Screenshots

image

### Environment

- Flarum version: 1.7.1 and nightly
- Website URL: http://discuss.flarum.org and nightly.flarum.org

```
```

### Possible Solution

_No response_

### Additional Context

_No response_

luceos:
**Fixes #3766**

Identify whether we have a completed mention before the current text so that we can stop mentioning. The logic checks for any space and/or follow up text, then also check whether the display name mention exists.

Did I forget about any edge cases?
Are we okay with this if statement? Imo it increases readability wdyt?

**Screenshot**

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/extensions/mentions/js/src/forum/addComposerAutocomplete.js
+++ b/extensions/mentions/js/src/forum/addComposerAutocomplete.js
@@ -89,17 +89,30 @@ export default function addComposerAutocomplete() {
 
       if (selection[1] - cursor > 0) return;
 
+      // Identify whether we are working with a mention by looking at an @
+      // consider the following scenario's `@"d1spl4Y name"`, `@username`, but also
+      // post mentions `@"d1spl4y name"#p45`. We cancel searching when we see a space or newline
+      // but only if this isn't part of a display name.
+
       // Search backwards from the cursor for an '@' symbol. If we find one,
       // we will want to show the autocomplete dropdown!
       const lastChunk = this.attrs.composer.editor.getLastNChars(30);
-      absMentionStart = 0;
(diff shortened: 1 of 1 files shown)


garygreen:
> We cancel searching when we see a space or newline

It should probably stop searching also whenever a non-valid username character is pressed. E.g. consider if you typed:

```
@luceos.
```

When typing "." to end the sentence it should stop firing. It looks like this is already the case for mentions package - it will show dropdown immediately on typing of `@` and then any characters like `. , @ -` it will close the dropdown and stop autocompleting. How does that logic work? Would it be easier to hook into a `close` event of some kind for the mentions extension?

luceos:
> It should probably stop searching also whenever a non-valid username character is pressed

The mentions extension also supports display names though. Those can be anything, even UTF8 characters.

luceos:
@SychO9 can you look at this one, especially now that I'm reviewing #3769 this "bug" becomes a nuisance.

SychO9:
will look into it after #3769 is merged as it reactors this part of the code completely

SychO9:
Superseded by #3806

PR #3771, opened 2023-03-19
Adjust size of FontAwesome icons on extensions list to match size of custom images

rob006:
**Before:**

223952bb

**After:**

a2f51291

**Necessity**

**Confirmed**

### DIFF
--- a/framework/core/less/admin/ExtensionWidget.less
+++ b/framework/core/less/admin/ExtensionWidget.less
@@ -46,6 +46,11 @@
   color: var(--muted-color);
 }
 
+.ExtensionListItem-icon i {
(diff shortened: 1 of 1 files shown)


imorland (approved):

SychO9:
@rob006 are you using any custom themes? this does not seem to be the case in default flarum, font sizes are already made relative to the size

rob006:
@SychO9 You're right, this is caused by [`malago/flarum-achievements`](https://github.com/malago86/flarum-achievements/blob/f5b9cfed9b1243d8bc345f9a02d75041a26e68b2/less/admin.less#L12-L15):

7dd82bca

PR #3773, opened 2023-03-21
fix: empty string displayed as SelectDropdown title

imorland:
Reverts the change to using `extractText` introduced in https://github.com/flarum/framework/pull/3608/

Whilst using `extractText`, the button title is rendered as an empty string. This PR reverts that change so that expected behaviour is restored.

**Screenshot**

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/framework/core/js/src/common/components/SelectDropdown.tsx
+++ b/framework/core/js/src/common/components/SelectDropdown.tsx
@@ -1,6 +1,5 @@
 import Dropdown, { IDropdownAttrs } from './Dropdown';
 import icon from '../helpers/icon';
-import extractText from '../utils/extractText';
 import classList from '../utils/classList';
 import type Component from '../Component';
 import type Mithril from 'mithril';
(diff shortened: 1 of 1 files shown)


SychO9 (approved):
@imorland was his tested locally and with the affected extensions? I haven't tested it myself

framework/core/js/src/common/components/SelectDropdown.tsx
```diff
@@ -50,7 +49,8 @@ export default class SelectDropdown<CustomAttrs extends ISelectDropdownAttrs = I
     const activeChild = children.find(isActive);
     let label = (activeChild && typeof activeChild === 'object' && 'children' in activeChild && activeChild.children) || this.attrs.defaultLabel;
 
-    label = extractText(label);
+    // @ts-ignore
+    if (label instanceof Array) label = label[0];
```
davwheat:
```suggestion
    if (Array.isArray(label)) label = label[0];
```

davwheat (changes requested):

davwheat:
Could we also document what this issue was affecting (e.g., an example of an extension where this was a problem)?

imorland:
Example of `fof/discussion-language`:

Flarum `1.7.1`:
image

With this patch:
image

imorland:
> @imorland was his tested locally and with the affected extensions? I haven't tested it myself

Yes, tested locally - all good

framework/core/js/src/common/components/SelectDropdown.tsx
imorland: Whilst I don't mind making this change, the advice from @SychO9 was to revert to the previous code for safety (https://github.com/flarum/framework/pull/3608). 

Any comments/thoughts/suggestions welcome :)

SychO9: I don't mind this change either

PR #3774, opened 2023-03-23
fix: `isDark()` utility can receive null value

### LINKED ISSUE #3764: `isDark` throws an error and the app crashes if a color is set to `NULL` in the database
### Current Behavior

When a tag is saved in the database with the `color` column set to `NULL`, then when trying to display that tag in the UI, there will be an error as the `isDark` util will try to call `replace` on a `null` variable. The app crashes then because of this.

### Steps to Reproduce

1. Set the color column of a tag to `NULL`
2. Open that tag index page

### Expected Behavior

No error

### Environment

- Flarum version:1.7.1
- PHP version: 8.1

exside:
**Fixes #3764**

Shouldn't be many, was mainly an edge case when there were NULL values in the db that would crash this method.

**Necessity**

Well, it doesn't crash anymore =)
Decided to not make the method too fancypants for performance reasons (e.g. no regex checks for valid hex colors etc.)

**Confirmed**

**Required changes:**

### DIFF
--- a/framework/core/js/src/common/utils/isDark.ts
+++ b/framework/core/js/src/common/utils/isDark.ts
@@ -25,7 +25,7 @@ export default function isDark(hexcolor: string): boolean {
   const b = parseInt(hexnumbers.slice(4, 6), 16);
   const yiq = (r * 299 + g * 587 + b * 114) / 1000;
 
-  const threshold = parseInt(getComputedStyle(document.body).getPropertyValue("--yiq-threshold").trim()) || 128;
+  const threshold = parseInt(getComputedStyle(document.body).getPropertyValue('--yiq-threshold').trim()) || 128;
   
   return yiq < threshold;
 }
--- a/framework/core/js/src/common/utils/isDark.ts
+++ b/framework/core/js/src/common/utils/isDark.ts
@@ -8,18 +8,25 @@
  * to preserve design consistency.
  */
 export default function isDark(hexcolor: string): boolean {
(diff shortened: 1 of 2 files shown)


framework/core/js/src/common/utils/isDark.ts
```diff
@@ -8,18 +8,24 @@
  * to preserve design consistency.
  */
 export default function isDark(hexcolor: string): boolean {
```
SychO9:
```suggestion
export default function isDark(hexcolor: string | null): boolean {
```

could you also type the `textContrastClass` util the same way?

framework/core/js/src/common/utils/isDark.ts
```diff
@@ -8,18 +8,24 @@
...
+  // return if hexcolor is undefined or shorter than 4 characters, shortest hex form is #333;
+  // decided against regex hex color validation for performance considerations
+  if (!hexcolor || hexcolor.length <= 4) {
```
SychO9:
```suggestion
  if (!hexcolor || hexcolor.length < 4) {
```

SychO9 (changes requested):
Thanks for the PR!

framework/core/js/src/common/utils/isDark.ts
luceos: Is `hexcolor` here including the `#` or not? That would make a world of difference.

SychO9: it includes `#`

SychO9 (approved):

imorland (approved):

PR #3585, opened 2022-08-05
[likes] reduce performance impact, add count and recent likes

luceos:
**Fixes #0000**

**Changes proposed in this pull request:**

This PR supposedly adds two new values by default to each post whenever Likes is enabled. A count of likes and the most recent likes.

I ran into an issue with the count, but before I spend much more time on this let's figure out whether we want this at all.

(I know dist needs to be removed :) )

**Reviewers should focus on:**

**Screenshot**

**Necessity**

- [ ] Has the problem that is being solved here been clearly explained?
- [ ] If applicable, have various options for solving this problem been considered?
- [ ] For core PRs, does this need to be in core, or could it be in an extension?
- [ ] Are we willing to maintain this for years / potentially forever?

**Confirmed**

- [ ] Frontend changes: tested on a local Flarum installation.
- [ ] Backend changes: tests are green (run `composer test`).
- [ ] Core developer confirmed locally this works as intended.
- [ ] Tests have been added, or are not appropriate here.

**Required changes:**

- [ ] Related documentation PR: (Remove if irrelevant)
- [ ] Related core extension PRs: (Remove if irrelevant)

PR #3781, opened 2023-04-01
perf(likes): limit `likes` relationship results

SychO9:
**Supersedes #3585**

Similar to #3780, it limits the results of the likes relationship of a post, and paginates all results in the existing modal.
The PR makes sure that the actor ID is in the limited results of the relationship, so that we can show the proper translation value that has `You liked and others...`.

Changes here are based on the linked PR, so they can only be merged after that.

**Screenshot**
Screenshot from 2023-04-01 15-25-11
Screenshot from 2023-04-01 15-25-19

**Necessity**

**Confirmed**


luceos (approved):
Looking good, even seeing improvements locally with the discuss data.

imorland (approved):

PR #3783, opened 2023-04-05
Update Translator.php

fakerybakery:

- Suppress error thrown in PHP 8.2

Added `@` before `preg_match` to fix issues with several extensions which pass `null` to it.

**Necessity**

**Confirmed**


luceos:
Hello @fakerybakery,

Thanks for this PR. What error does this PR fix exactly?

imorland:
Thanks for the PR!

I noticed this recently too. However, I'd rather not simply suppress errors from `preg_match`. I've created an alternative solution for this in https://github.com/flarum/framework/pull/3801, so I'll close this now.

PR #3784, opened 2023-04-07
feat(permission): add delete own posts permission

OrdinaryJellyfish:
**Fixes https://discuss.flarum.org/d/31993-adding-permission-to-delete-own-posts**

* Add permission for users to delete their own posts as a time-based permission

* Does the permission work properly?
* Is it in the right spot on the permission grid?

**Screenshot**

image

**Necessity**

**Confirmed**

**Required changes:**

None

### DIFF
--- a/framework/core/js/src/admin/components/PermissionGrid.tsx
+++ b/framework/core/js/src/admin/components/PermissionGrid.tsx
@@ -399,7 +399,6 @@ export default class PermissionGrid<CustomAttrs extends IPermissionGridAttrs = I
       },
       60
     );
-    
 
     items.add(
       'userEditCredentials',
--- a/framework/core/js/src/admin/components/PermissionGrid.tsx
+++ b/framework/core/js/src/admin/components/PermissionGrid.tsx
@@ -367,6 +367,30 @@ export default class PermissionGrid<CustomAttrs extends IPermissionGridAttrs = I
     );
 
     items.add(
+      'hideOwnPosts',
+      {
+        icon: 'far fa-trash-alt',
+        label: app.translator.trans('core.admin.permissions.allow_delete_own_posts_label'),
+        setting: () => {
+          const minutes = parseInt(app.data.settings.allow_hide_own_posts, 10);
+
+          return SettingDropdown.component({
+            defaultLabel: minutes
+              ? app.translator.trans('core.admin.permissions_controls.allow_some_minutes_button', { count: minutes })
+              : app.translator.trans('core.admin.permissions_controls.allow_indefinitely_button'),
+            key: 'allow_hide_own_posts',
+            options: [
+              { value: '-1', label: app.translator.trans('core.admin.permissions_controls.allow_indefinitely_button') },
+              { value: '10', label: app.translator.trans('core.admin.permissions_controls.allow_ten_minutes_button') },
+              { value: 'reply', label: app.translator.trans('core.admin.permissions_controls.allow_until_reply_button') },
+            ],
+          });
+        },
+      },
+      60
+    );
+
+    items.add(
       'deletePosts',
       {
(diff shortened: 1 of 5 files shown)


framework/core/js/src/admin/components/PermissionGrid.tsx
```diff
@@ -366,6 +366,31 @@ export default class PermissionGrid<CustomAttrs extends IPermissionGridAttrs = I
       60
     );
 
+    items.add(
+      'hideOwnPosts',
```
SychO9: let's move this up to the participate section rather than moderation, under `allowPostEditing`.

SychO9 (changes requested):
Nice!

One thing I'd like us to make consistent is the `key` of the permission and locale label. Let's make tem consistent and use `allow_hide_own_posts` for te permission key AND the label `allow_hide_own_posts_label`, andeplace any `allow_delete_own_posts`

framework/core/src/Post/Access/PostPolicy.php
```diff
@@ -71,6 +71,14 @@ public function edit(User $actor, Post $post)
      */
     public function hide(User $actor, Post $post)
     {
-        return $this->edit($actor, $post);
+        if ($post->user_id == $actor->id && (! $post->hidden_at || $post->hidden_user_id == $actor->id) && $actor->can('reply', $post->discussion)) {
```
SychO9, line 74: I wonder if we should drop the bit about only being able to hide when the discussion is still open 🤔 
```suggestion
        if ($post->user_id == $actor->id && (! $post->hidden_at || $post->hidden_user_id == $actor->id)) {
```

OrdinaryJellyfish: I tried to keep the behavior close to the post-editing one for consistency so that's why that's there! I was wondering if it would be necessary or not.

SychO9: Yeah, not sure, let's keep it as is for now, then we'll see I guess unless other core devs have an opinion on is.

SychO9 (approved):
LGTM!

framework/core/src/Post/Access/PostPolicy.php
askvortsov1: I'd prefer to keep it. If a user can't participate in a discussion going forward, i don't think they should be able to participate retroactively either.

framework/core/js/src/admin/components/PermissionGrid.tsx
```diff
@@ -288,6 +288,31 @@ export default class PermissionGrid<CustomAttrs extends IPermissionGridAttrs = I
...
+        setting: () => {
+          const minutes = parseInt(app.data.settings.allow_hide_own_posts, 10);
+
+          return SettingDropdown.component({
+            defaultLabel: minutes
+              ? app.translator.trans('core.admin.permissions_controls.allow_some_minutes_button', { count: minutes })
+              : app.translator.trans('core.admin.permissions_controls.allow_indefinitely_button'),
+            key: 'allow_hide_own_posts',
+            options: [
+              { value: '-1', label: app.translator.trans('core.admin.permissions_controls.allow_indefinitely_button') },
```
askvortsov1, line 305: Perhaps this should be extracted into a util? Although probably not, if we restructure the permissions system during 2.0, that would become redundant.

askvortsov1 (approved):
Looks good to me, left a few comments though.

framework/core/js/src/admin/components/PermissionGrid.tsx
SychO9: I wonder if we will end up restructuring it in 2.0, we need some brainstorming .

PR #3785, opened 2023-04-10
fix: circular dependencies disable all involved extensions

SychO9:
Throws an exception wen circular dependencies are found, which prevents the extensions within the circular depdencies fro getting all disabled.

**Screenshot**
This is an example where I created a circular dependency with emoji, suspend, and likes:
image

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/framework/core/locale/core.yml
+++ b/framework/core/locale/core.yml
@@ -605,6 +605,7 @@ core:
 
     # These translations are displayed as error messages.
     error:
+      circular_dependencies_message: "Circular dependencies detected: {extensions}. Aborting. Please disable one of the extensions and try again."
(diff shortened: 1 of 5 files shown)


askvortsov1 (commented):
Should we also reject attempts to enable extensions that would cause circular dependencies?

SychO9:
That would already be rejected as well, whether the extension being enabled causes the circular dependency. Or whether there is an existing circular dependency with other already enabled extensions (due to an update). The error is thrown.

askvortsov1 (approved):
Ah right, this would run when enabled extensions are persisted. Looks good!

luceos (approved):

PR #3788, opened 2023-04-15
fix: notification subject discussion eager loading fails

### LINKED ISSUE #3778: fttb slow response time
### Current Behavior

我發現了一個BUG，當我的通知數來到5000+，fttb時間會來到2秒以上，但如果我清除了所有通知，則會恢復到400MS

我的母語是中文，擔心使用翻譯軟體會有差異，所以在這裡附上中文說明
My native language is Chinese, I am worried that there will be differences when using translation software, so I attach a Chinese explanation here

I found a bug where the fttb (first time to byte) time would exceed 2 seconds when my notification count reached 10000+. However, if I cleared all notifications, the fttb time would return to 400ms.

### Steps to Reproduce

Number of account notifications 5000+

### Expected Behavior

Slow initial server response time

### Screenshots

_No response_

### Environment

- Flarum version: 1.7.2
- Webserver: nginx
- Hosting environment: dedicated server
- PHP version: 8.1.7
- Browser: Chrome

```
```

### Possible Solution

_No response_

### Additional Context

_No response_

SychO9:
**Closes #3778**

In 1.7 the `property_exists` check was added to fix PHP 8.2 deprecations but missed the fact that dynamic properties cannot be checked like that.

I don't believe this is the problem described in the issue however, only 10 notifications are loaded at a time. So it could be a bad extension.

**Necessity**

**Confirmed**

### DIFF
--- a/framework/core/src/Api/Controller/ListNotificationsController.php
+++ b/framework/core/src/Api/Controller/ListNotificationsController.php
@@ -113,16 +113,16 @@ class ListNotificationsController extends AbstractListController
         $ids = [];
 
         foreach ($notifications as $notification) {
-            if ($notification->subject && $notification->subject->getAttribute('discussion_id')) {
-                $ids[] = $notification->subject->getAttribute('discussion_id');
+            if ($notification->subject && ($discussionId = $notification->subject->getAttribute('discussion_id'))) {
+                $ids[] = $discussionId;
             }
         }
 
         $discussions = Discussion::query()->find(array_unique($ids));
 
         foreach ($notifications as $notification) {
-            if ($notification->subject && $notification->subject->getAttribute('discussion_id')) {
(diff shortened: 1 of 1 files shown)


framework/core/src/Api/Controller/ListNotificationsController.php
```diff
@@ -113,16 +113,16 @@ private function loadSubjectDiscussions(array $notifications)
         $ids = [];
 
         foreach ($notifications as $notification) {
-            if ($notification->subject && property_exists($notification->subject, 'discussion_id')) {
-                $ids[] = $notification->subject->discussion_id;
+            if ($notification->subject && $notification->subject->getAttribute('discussion_id')) {
```
askvortsov1: Could we use `&& ($discussionId = $notification->subject->getAttribute('discussion_id'))` to avoid repeating it?

askvortsov1 (approved):

PR #3793, opened 2023-04-15
fix(a11y): reply placeholder not accessible

### LINKED ISSUE #3659: "Reply placeholder" is not focusable and is missing a `role="button"`
### Current Behavior

The `ReplyPlaceholder` component is inaccessible via keyboard navigation.

### Steps to Reproduce

1. Open a discussion
2. Try to tab navigate to the reply placeholder
3. alternatively, try to use other method to navigate to reply via keyboard

### Expected Behavior

Since the `article` element is clickable (i.e. interactable), it should also be accessible via keyboard.

### Screenshots

_No response_

### Environment

N/A, I found this as a user, not admin.

_No response_

### Possible Solution

I haven't written any jsx before, but it looks like `tabindex="0"` and `role="button"` should be added to the article tag in `ReplyPlaceholder.js`

```javascript
      <article className="Post ReplyPlaceholder" onclick={reply}>
        <header className="Post-header">
          {avatar(app.session.user, { className: 'PostUser-avatar' })} {app.translator.trans('core.forum.post_stream.reply_placeholder')}
        </header>
      </article>
```

According to [mdn: ARIA role button](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role) we then also need to add listeners for `space` and `enter` keydown-listeners

### Additional Context

_No response_

SychO9:
**Fixes #3659**

* Makes the placeholder into a button.

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/framework/core/js/src/forum/components/ReplyPlaceholder.js
+++ b/framework/core/js/src/forum/components/ReplyPlaceholder.js
@@ -38,11 +38,11 @@ export default class ReplyPlaceholder extends Component {
     };
 
     return (
-      <article className="Post ReplyPlaceholder" onclick={reply}>
-        <header className="Post-header">
+      <button className="Post ReplyPlaceholder" onclick={reply}>
+        <span className="Post-header">
           {avatar(app.session.user, { className: 'PostUser-avatar' })} {app.translator.trans('core.forum.post_stream.reply_placeholder')}
-        </header>
-      </article>
+        </span>
+      </button>
     );
   }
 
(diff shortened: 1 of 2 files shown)


askvortsov1 (approved):
Makes sense. I'm assuming this is tested locally, but might be nice for review to have before/after screenshots to ensure that nothing has changed

SychO9:
Yup, the style is not changed at all. Just the semantics.

PR #3797, opened 2023-04-16
chore: drop usage of jquery in `install` and `update` interfaces

### LINKED ISSUE #3457: Migrate update and install pages away from legacy external jQuery
https://github.com/flarum/framework/blob/main/framework/core/views/install/update.php#L20

https://github.com/flarum/framework/blob/main/framework/core/views/install/install.php#L69

SychO9:
**Fixes #3457**

* Just converts the JS code to vanilla JS.
* Tesed locally.

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/framework/core/views/install/install.php
+++ b/framework/core/views/install/install.php
@@ -67,37 +67,36 @@
 </form>
 
 <script>
-document.querySelector('form input').select();
+  window.onload = function() {
+    document.querySelector('form input').select();
 
-document.querySelector('form').addEventListener('submit', function(e) {
-  e.preventDefault();
+    document.querySelector('form').addEventListener('submit', function(e) {
+      e.preventDefault();
 
-  var button = this.querySelector('button');
-  button.textContent = 'Please Wait...';
-  button.disabled = true;
+      var button = this.querySelector('button');
+      button.textContent = 'Please Wait...';
+      button.disabled = true;
 
-  fetch('', {
-    method: 'POST',
-    body: new FormData(this)
-  })
-    .then(response => {
-      if (response.ok) {
-        window.location.reload();
-      } else {
-        response.text().then(errorMessage => {
-          var error = document.querySelector('#error');
-          error.style.display = 'block';
-          error.textContent = 'Something went wrong:\n\n' + errorMessage;
-          button.disabled = false;
-          button.textContent = 'Install Flarum';
+      fetch('', {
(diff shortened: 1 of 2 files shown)


framework/core/views/install/install.php
```diff
@@ -66,29 +66,38 @@
   </div>
 </form>
 
-<script src="https://code.jquery.com/jquery-2.1.4.min.js"></script>
 <script>
-$(function() {
-  $('form :input:first').select();
+document.querySelector('form input').select();
```
askvortsov1: Should this be in an `onload`?

framework/core/views/install/install.php
```diff
@@ -66,29 +66,38 @@
...
+        response.text().then(errorMessage => {
+          var error = document.querySelector('#error');
+          error.style.display = 'block';
+          error.textContent = 'Something went wrong:\n\n' + errorMessage;
+          button.disabled = false;
+          button.textContent = 'Install Flarum';
+        });
+      }
+    })
+    .catch(error => {
```
askvortsov1: Why are we catching the error to immediately rethrow it?

askvortsov1 (changes requested):

framework/core/views/install/install.php
SychO9: woups 🙈 was debugging

framework/core/views/install/update.php
```diff
@@ -18,37 +18,36 @@
...
-            error.textContent = 'Something went wrong:\n\n' + errorMessage;
-            button.disabled = false;
-            button.textContent = 'Update Flarum';
-          });
-        }
-      })
-      .catch(error => {
-        console.error('Error:', error);
-      });
+  window.onload = function() {
```
askvortsov1: https://stackoverflow.com/a/37401912 might be a better fit, actually

SychO9: let me just switch back to jQuery 🤣

askvortsov1: Aaaaaaaaa noooooo!!!!!!!

askvortsov1 (approved):

matteocontrini:
A bit late here, but I think fetch Promise rejections should be handled too. Otherwise network errors are ignored (they weren't, in the previous jQuery implementation).

PR #3802, opened 2023-04-20
chore[tags]: Change some methods from private to protected, to be able to extend the affected classes

iPurpl3x:
_see title_

**Necessity**

**Confirmed**

### DIFF
--- a/extensions/tags/src/Content/Tag.php
+++ b/extensions/tags/src/Content/Tag.php
@@ -113,7 +113,7 @@ class Tag
      *
      * @return array
      */
-    private function getSortMap()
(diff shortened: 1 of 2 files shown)


luceos:
@iPurpl3x out of curiosity, did you check the check boxes intended for maintainers? 🙈

imorland (approved):
I see no reason to not accept this PR, however be warned that much is likely to change here after the next release as we prepare for `2.0`

iPurpl3x:
@luceos I guess I did, but it wasn't clear to me whether I or a maintainer should check those checkboxes…

PR #3803, opened 2023-04-20
Update Application.tsx

johnvanhulsen:
Fix issue in mobile header. Scrolled and affix classnames on #App goes wrong without this. More info: https://discuss.flarum.org/d/32559-scrolled-and-affix-classnames-on-app-goes-wrong

**Fixes #0000**

Header (on mobile) with elements above the sticky #App

**Screenshot**

**QA**

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/framework/core/js/src/common/Application.tsx
+++ b/framework/core/js/src/common/Application.tsx
@@ -335,7 +335,7 @@ export default class Application {
     // which will set the navbar's position to fixed. We don't want to always
     // have it fixed, as that could overlap with custom headers.
     const scrollListener = new ScrollListener((top: number) => {
-      const offset = appEl.getBoundingClientRect().top + document.body.scrollTop;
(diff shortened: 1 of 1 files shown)


davwheat (approved):
Makes sense, and something I've run into before and simply disregarded.

askvortsov1 (changes requested):
Reading over this code, it feels weird to me, since we're now comparing "top + ... > top".
There's a few issues:

- `top` is `window.pageYOffset`, which is an alias for `window.scrollY`
- `document.body.scrollTop` is deprecated in favor of `window.scrollY`: https://stackoverflow.com/questions/19635188/why-is-body-scrolltop-deprecated

So it feels like we should instead remove `document.body.scrollTop`, so that `offset` is constant. But that feels like it would exhibit the same behavior as in the report?

SychO9:
Closing due to inactivity.

PR #3804, opened 2023-04-20
fix(tags): not all tags are loaded in the permission grid

### LINKED ISSUE #3800: Permission grid doesn't show all the tags
### Current Behavior

If you have many tags, the permission grid tag restrictions are incomplete. When you click "Restrict by tag", you'll only see a few tags (12 in my case). If you already have a restriction for a tag that doesn't get loaded, you won't see the column in the grid at all.

### Steps to Reproduce

I don't have precise reproduction steps, but you should try to:

- Create many tags so that they go beyond the limit of tags that are loaded in the frontend automatically (it's 12 here but I can't find that number in the source code).
- Navigate to the Permissions page in the admin, refresh the page and see that if you click "Restrict by tag" some tags are missing.
- Navigate to the Tags page so that the full tags list is loaded in the store, go back to Permissions and see that the complete list is now available.

### Expected Behavior

All tags should be loaded.

### Screenshots

_No response_

### Environment

- Flarum version: 1.7.2

I can provide it if relevant.

### Possible Solution

_No response_

### Additional Context

_No response_

SychO9:
**Fixes #3800**

* The code was just getting all tags from the store, this PR changes the state to make sure to consider first load.

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/extensions/tags/js/src/common/states/TagListState.ts
+++ b/extensions/tags/js/src/common/states/TagListState.ts
@@ -2,17 +2,27 @@ import app from 'flarum/common/app';
 import type Tag from '../../common/models/Tag';
 
 export default class TagListState {
-  loadedIncludes = new Set();
+  loadedIncludes?: Set<string>;
 
   async load(includes: string[] = []): Promise<Tag[]> {
-    const unloadedIncludes = includes.filter((include) => !this.loadedIncludes.has(include));
+    if (!this.loadedIncludes) {
+      return this.query(includes);
+    }
+
+    const unloadedIncludes = includes.filter((include) => !this.loadedIncludes!.has(include));
 
     if (unloadedIncludes.length === 0) {
       return Promise.resolve(app.store.all<Tag>('tags'));
     }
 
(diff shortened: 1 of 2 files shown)


extensions/tags/js/src/common/states/TagListState.ts
```diff
@@ -2,17 +2,27 @@ import app from 'flarum/common/app';
...
     if (unloadedIncludes.length === 0) {
       return Promise.resolve(app.store.all<Tag>('tags'));
     }
 
-    return app.store.find<Tag[]>('tags', { include: unloadedIncludes.join(',') }).then((val) => {
-      unloadedIncludes.forEach((include) => this.loadedIncludes.add(include));
+    return this.query(unloadedIncludes);
+  }
+
+  async query(includes: string[] = []): Promise<Tag[]> {
```
askvortsov1, line 21: Should we make this protected?

askvortsov1 (approved):

davwheat (approved):

PR #3808, opened 2023-04-27
perf: speed up post creation time

SychO9:
This PR brings performance improvements for the post creation endpoint.

* Visibility checking on the notification syncer (which is expensive) is dropped as it didn't actually fix what it was introduced for and was replaced in the past with visibility checking on the specific notification jobs. (https://github.com/flarum/framework/commit/ce6cb58612a6115e8280181192591b6b9dbac556).
* When post has a large amount of mentions, it retrieves each mention individually, this PR eager loads mentions once and uses the loaded collection instead (https://github.com/flarum/framework/commit/df4d3cac955fcab3c00bc9742648d822eeb7c26c).
* For visibility checking on mentions notifications, this PR also adds some needed eager loads (https://github.com/flarum/framework/commit/8aaa33067c8648ca6579e3e642847418a6e5f126).
* Right now, with group mentions, user mentions, and post mentions (without taking into account extensions), the endpoint has to do a lot of visibility checking, which is an expensive operation when applied to many users. This PR also moves the logic that triggers mention notification to a queueable job (https://github.com/flarum/framework/commit/6e1c22aa0ce2a5796d35a4f2bd4e6272693c478b).

**Screenshot**
Screenshot from 2023-04-27 14-43-45
↓
Screenshot from 2023-04-27 14-43-50
↓
Screenshot from 2023-04-27 16-11-58
↓
Screenshot from 2023-04-27 16-44-53
↓
Screenshot from 2023-04-27 17-47-14

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/extensions/mentions/src/Job/SendMentionsNotificationsJob.php
+++ b/extensions/mentions/src/Job/SendMentionsNotificationsJob.php
@@ -1,5 +1,12 @@
 <?php
 
+/*
+ * This file is part of Flarum.
+ *
+ * For detailed copyright and license information, please view the
+ * LICENSE file that was distributed with this source code.
+ */
+
 namespace Flarum\Mentions\Job;
 
 use Flarum\Mentions\Notification\GroupMentionedBlueprint;
--- a/extensions/mentions/src/Job/SendMentionsNotificationsJob.php
+++ b/extensions/mentions/src/Job/SendMentionsNotificationsJob.php
@@ -1,0 +1,101 @@
+<?php
+
+namespace Flarum\Mentions\Job;
+
+use Flarum\Mentions\Notification\GroupMentionedBlueprint;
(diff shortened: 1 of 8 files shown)


extensions/mentions/src/Job/SendMentionsNotificationsJob.php
```diff
@@ -0,0 +1,108 @@
...
+
+class SendMentionsNotificationsJob extends AbstractJob
+{
+    /**
+     * @var CommentPost
+     */
+    protected $post;
+
+    /**
+     * @var array
```
askvortsov1, line 29: Can we annotate `array<what>`?

askvortsov1 (approved):
Haven't tested locally, but code changes look good.

imorland (approved):

PR #3809, opened 2023-04-27
fix(bbcode): localize quote `wrote` string

### LINKED ISSUE #3807: Hardcoded "wrote" in flarum/bbcode
### Current Behavior

Hi, I'm interested on an old issue from 'issue-archive' repository I have the same problem and I need help with this:

https://github.com/flarum/issue-archive/issues/98

I would like to translate the string "wrote" in Spanish since I migrated an old forum from phpBB to Flarum and there were many quotes.

### Steps to Reproduce

Migrate content from another CMS and check discussions with quotes.

### Expected Behavior

A way of translating the word.

SychO9:
**Fixes #3807**

* Allows localizing the `wrote` bit of the quote bbcode.
* This PR takes into consideration previous attempts. The translated string is not cached with the renderer cache. It is resolved on new renders.

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/extensions/bbcode/extend.php
+++ b/extensions/bbcode/extend.php
@@ -9,22 +9,34 @@
 
 use Flarum\Extend;
 use s9e\TextFormatter\Configurator;
+use s9e\TextFormatter\Renderer;
 
-return (new Extend\Formatter)
-    ->configure(function (Configurator $config) {
-        $config->BBCodes->addFromRepository('B');
-        $config->BBCodes->addFromRepository('I');
-        $config->BBCodes->addFromRepository('U');
-        $config->BBCodes->addFromRepository('S');
-        $config->BBCodes->addFromRepository('URL');
-        $config->BBCodes->addFromRepository('IMG');
-        $config->BBCodes->addFromRepository('EMAIL');
-        $config->BBCodes->addFromRepository('CODE');
-        $config->BBCodes->addFromRepository('QUOTE');
-        $config->BBCodes->addFromRepository('LIST');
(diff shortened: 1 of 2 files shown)


extensions/bbcode/extend.php
```diff
@@ -9,22 +9,34 @@
...
+            $config->BBCodes->addFromRepository('B');
+            $config->BBCodes->addFromRepository('I');
+            $config->BBCodes->addFromRepository('U');
+            $config->BBCodes->addFromRepository('S');
+            $config->BBCodes->addFromRepository('URL');
+            $config->BBCodes->addFromRepository('IMG');
+            $config->BBCodes->addFromRepository('EMAIL');
+            $config->BBCodes->addFromRepository('CODE');
+            $config->BBCodes->addFromRepository('QUOTE', 'default', [
+                'authorStr' => '<xsl:value-of select="@author"/> <xsl:value-of select="$L_WROTE"/>'
```
davwheat, line 33: Can the author string be part of the translation instead? Some languages might need "wrote author" instead.

davwheat (changes requested):

extensions/bbcode/extend.php
SychO9: I don't believe that that's possible to do.

davwheat: Unfortunate... Either way, a locale ext could override this in their extend.php anyway, so

davwheat (approved):

imorland (approved):

PR #3813, opened 2023-04-29
Do not catch exceptions when testing Console commands

n-peugnet:

Configure the ConsoleApplication to not catch exceptions when testing Console commands.

This allows to easily test a case where a command should throw using `$this->expectException()`.

**QA**
For now I only tested it locally by modifying the vendor directory. It should not cause too much issues for existing testing suites, except if the way to check if a console was throwing was done by checking the output. As it is rather tedious I don't suspect that there are a lot of such test cases.

Edit: well there is one here :sweat_smile: 
https://github.com/flarum/framework/blob/af3f91ca5b30c2bfc7a7f9dd36839339372cc745/framework/core/tests/integration/extenders/ConsoleTest.php#L22-L29

This case has been simple to test as there is no other input, and Symfony seems to treat this exception differently, as usually it shows the file name and line of the throw statement. It can be written equally simply using `$this->expectExceptionMessage()`.

Edit2: In fact I think this test would even be better off using `$this->expectException()` as the goal is not to check the message but that the command is not found, and because this way the test is less brittle.

**Necessity**

**Confirmed**


askvortsov1 (approved):
This feels like a good change to me, especially with that error message coming from a dependency. That being said, I wonder if it would make sense to have an additional test case that would switch "catch exceptions" back on, so we can ensure that the UX in case of an error is reasonable. Preferably, this would be a flarum-level exception, not something from Symfony.

n-peugnet:
Please note that this is still a breaking change for the testsuites of extension developers, and it should be announced accordingly in the release notes.

SychO9:
Thanks for the PR!

I think this should be included in 2.0 since it has non BC changes.

askvortsov1:
Fair enough. I'm generally much less concerned about testing/static analysis breaking changes than user-facing ones, but since this isn't a very urgent change, 2.0 makes sense.

imorland (approved):

PR #3814, opened 2023-04-29
chore: improve test suite

SychO9:
* This improves the testing experience locally a little by not requiring us to run `composer install` every time we run tests on an extension on the monorepo.

**In the future**
* In the future when we have SQLite support, we can simplify it further by removing the need to run `composer setup` every single time as well.

**Necessity**

**Confirmed**

**Required changes:**


.github/workflows/REUSABLE_backend.yml
```diff
@@ -21,6 +21,12 @@ on:
         required: false
         default: '.'
 
+      # Used with mono-repos, defaults to `backend_directory`.
+      composer_directory:
```
askvortsov1: Note: maybe `composer_root_directory`?  Should be clearer that this is where composer is run from, not where composer is located.

framework/core/src/Extension/ExtensionManager.php
```diff
@@ -106,31 +106,18 @@ public function getExtensions()
...
-                if ($subextPaths = Arr::get($package, 'extra.flarum-subextensions', [])) {
-                    foreach ($subextPaths as $subExtPath) {
-                        $subPackagePath = "$packagePath/$subExtPath";
-                        $conf = json_decode($this->filesystem->get("$subPackagePath/composer.json"), true);
-
-                        if (Arr::get($conf, 'type') === 'flarum-extension') {
-                            $composerJsonConfs[$subPackagePath] = $conf;
-                        }
-                    }
+                if ($subExtConfs = $this->subExtensionConfsFromJson($package, $packagePath)) {
```
askvortsov1, line 76: Way cleaner. Nice!

php-packages/testing/src/integration/Extension/ExtensionManagerIncludeCurrent.php
```diff
@@ -112,4 +104,32 @@ protected function getAssetsFilesystem(): Cloud
...
         return new FilesystemAdapter(new FlysystemFilesystem(new Local($this->paths->public.'/assets'), ['url' => resolve('flarum.config')->url().'/assets']));
     }
+
+    protected function includeCurrentExtension($package, string $packagePath)
+    {
+        if (Arr::get($package, 'type') === 'flarum-extension') {
+            $current = new Extension($packagePath, $package);
+            $current->setInstalled(true);
+            $current->setVersion(Arr::get($package, 'version'));
+            $current->calculateDependencies([], []);
```
askvortsov1: Might be cleaner if we redid all these with $withCalculatedDependencies$, $withVersion$, etc. Also, the function param here takes 1 arg, and we're giving it 2.

php-packages/testing/src/integration/Extension/ExtensionManagerIncludeCurrent.php
```diff
@@ -112,4 +104,32 @@ protected function getAssetsFilesystem(): Cloud
...
+
+            $this->extensions->put($current->getId(), $current);
+
+            $this->extensions = $this->extensions->sortBy(function ($extension, $name) {
+                return $extension->composerJsonAttribute('extra.flarum-extension.title');
+            });
```
askvortsov1: Should this be using top-sort instead? I don't remember why this was added here.

php-packages/testing/src/integration/Extension/ExtensionManagerIncludeCurrent.php
```diff
@@ -112,4 +104,32 @@ protected function getAssetsFilesystem(): Cloud
...
+     * Allows symlinking the vendor directory in extensions when running tests on monorepos.
+     */
+    protected function includeMonorepoExtensions($package, string $packagePath)
+    {
+        foreach ($this->subExtensionConfsFromJson($package, $packagePath) ?? [] as $path => $package) {
+            $extension = $this->extensionFromJson($package, $path);
+            $extension->calculateDependencies([], []);
+            $this->extensions->put($extension->getId(), $extension);
+        }
+    }
```
askvortsov1: I don't love these methods as mutations of `this->extensions`; why not return an array instead?

askvortsov1 (commented):
I don't love the complex location of `bootstrap` outside of the extension file. How out-of-date is Flarum CLI? I feel like this would be a decent use case for including a copy of that bootstrap file in every extension's code.

SychO9:
> I don't love the complex location of bootstrap outside of the extension file. How out-of-date is Flarum CLI? I feel like this would be a decent use case for including a copy of that bootstrap file in every extension's code.

the bootstrap file is irrelevant to non-mono-repos though, it only exists under a global `tests` directory to avoid duplicating it across every single extension.

.github/workflows/REUSABLE_backend.yml
SychO9: It doesn't really matter since its removed in the follow up PR 👍🏼

php-packages/testing/src/integration/Extension/ExtensionManagerIncludeCurrent.php
SychO9: Can't recall, let's leave it be for now.

php-packages/testing/src/integration/Extension/ExtensionManagerIncludeCurrent.php
SychO9: imo it's cleaner to mutate, since we have to mutate it after the methods anyway, the constant pass and return collection object reads oddly.

askvortsov1:
> > I don't love the complex location of bootstrap outside of the extension file. How out-of-date is Flarum CLI? I feel like this would be a decent use case for including a copy of that bootstrap file in every extension's code.
> 
> the bootstrap file is irrelevant to non-mono-repos though, it only exists under a global `tests` directory to avoid duplicating it across every single extension.

I don't disagree, but I worry that we might end up with the framework extensions and community extensions having more and more separate structures. I also think there's great value in encouraging other devs/groups to develop in monorepos of their own (e.g. Blomstra, GB, myself, etc), since that really does simplify developer experience a ton. And from that perspective, having a single source of truth for all this boilerplate feels at least worthy of consideration.

SychO9:
I do agree, I tried keeping the extension boilerplate as little customized as possible, but at this time I don't think there is escaping the minimum amount of customization for the monorepo. I considered putting the bootstrap into the testing package and having the monorepo behavior switchable somehow (that might still be a better alterntive actually). But even then requiring the bootstrap file in setup files would still require a custom relative path.

I think that's acceptable, especially since the CLI and docs are keeping the extension boilerplate documented.

imorland (approved):

PR #3815, opened 2023-04-29
chore: unify backend test jobs into one job

SychO9:
Instead of multiple jobs for each extension, we unify them into one `Backend Tests` job so that it's easier to read test results and reduce the number of jobs reducing the wait time for tests to finish.

**Necessity**

**Confirmed**

**Required changes:**


askvortsov1 (approved):
Very nice!

Out of curiosity, what's with the skipped checks that have template vars in their names, e.g. `Backend Tests / run / PHP ${{ matrix.php }} / ${{ matrix.db }} ${{ matrix.prefixStr }} (pull_request) `?

SychO9:
most of those are duplicates from PR (since we get push and PR jobs) which we skip using a conditional cause we don't need dups.

Also this PR is based on #3814

PR #3816, opened 2023-04-30
feat: cli command for enabling or disabling an extension

imorland:
Adds a new CLI command to enable toggling of a given extension.

Usage:

```
 php flarum extension:enable blomstra-turnstile
```

```
 php flarum extension:disable blomstra-turnstile
```

**Screenshot**
image

image

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/framework/core/src/Extension/Console/ToggleExtensionCommand.php
+++ b/framework/core/src/Extension/Console/ToggleExtensionCommand.php
@@ -11,7 +11,6 @@ namespace Flarum\Extension\Console;
 
 use Flarum\Console\AbstractCommand;
 use Flarum\Extension\ExtensionManager;
-use Illuminate\Contracts\Container\Container;
 
 class ToggleExtensionCommand extends AbstractCommand
 {
@@ -26,7 +25,7 @@ class ToggleExtensionCommand extends AbstractCommand
 
         parent::__construct();
     }
-    
+
     /**
      * {@inheritdoc}
      */
--- a/framework/core/src/Extension/Console/ToggleExtensionCommand.php
+++ b/framework/core/src/Extension/Console/ToggleExtensionCommand.php
@@ -43,16 +43,33 @@ class ToggleExtensionCommand extends AbstractCommand
      */
     protected function fire()
     {
-        $name = $this->input->getArgument('extension');
+        $name = $this->input->getArgument('extension-id');
+        $enabling = $this->input->getFirstArgument() === 'extension:enable';
 
-        if ($this->input->getOption('enable')) {
-            $this->info("Enabling $name extension...");
-            $this->extensionManager->enable($name);
-        } elseif ($this->input->getOption('disable')) {
-            $this->info("Disabling $name extension...");
-            $this->extensionManager->disable($name);
-        } else {
-            $this->error('You must specify --enable or --disable.');
+        if ($this->extensionManager->getExtension($name) === null) {
+            $this->error("There are no extensions by the ID of '$name'.");
+            return;
+        }
+
+        switch ($enabling) {
+            case true:
+                if ($this->extensionManager->isEnabled($name)) {
+                    $this->info("The '$name' extension is already enabled.");
+                    return;
+                } else {
+                    $this->info("Enabling '$name' extension...");
+                    $this->extensionManager->enable($name);
+                }
+                break;
+            case false:
+                if (! $this->extensionManager->isEnabled($name)) {
+                    $this->info("The '$name' extension is already disabled.");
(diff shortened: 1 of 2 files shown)


framework/core/src/Extension/Console/ToggleExtensionCommand.php
```diff
@@ -0,0 +1,59 @@
...
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        $this
+            ->setName('extension:toggle')
+            ->setDescription('Enable or disable an extension.')
+            ->addArgument('extension', null, 'The extension to enable or disable.')
+            ->addOption('enable', null, null, 'Enable the extension.')
+            ->addOption('disable', null, null, 'Disable the extension.');
```
SychO9: rather than `extension:toggle --enable` or so, I'd prefer `extension:enable` and `extension:disable`. I'd also like us to be specific about the argument's nature, which is the extension ID.
```suggestion
        $this
            ->setName('extension:enable')
            ->setAliases(['extension:disable'])
            ->setDescription('Enable or disable an extension.')
            ->addArgument('extension-id', null, 'The ID of the extension to enable or disable.');
```

framework/core/src/Extension/Console/ToggleExtensionCommand.php
```diff
@@ -0,0 +1,59 @@
...
+
+        if ($this->input->getOption('enable')) {
+            $this->info("Enabling $name extension...");
+            $this->extensionManager->enable($name);
+        } elseif ($this->input->getOption('disable')) {
+            $this->info("Disabling $name extension...");
+            $this->extensionManager->disable($name);
+        } else {
+            $this->error('You must specify --enable or --disable.');
+        }
```
SychO9: We can then check which command name was used and effectively drop the `else`.
we also have to check if the extension exists before doing anything otherwise we get exceptions.
An additional nicety imo, would be to be explicit about an extension being already enabled or disabled 😸 

```suggestion
        $name = $this->input->getArgument('extension-id');
        $enabling = $this->input->getFirstArgument() === 'extension:enable';

        if ($this->extensionManager->getExtension($name) === null) {
            $this->error("There are no extensions by the ID of '$name'.");
            return;
        }

        switch ($enabling) {
            case true:
                if ($this->extensionManager->isEnabled($name)) {
                    $this->info("The '$name' extension is already enabled.");
                    return;
                } else {
                    $this->info("Enabling '$name' extension...");
                    $this->extensionManager->enable($name);
                }
                break;
            case false:
                if (! $this->extensionManager->isEnabled($name)) {
                    $this->info("The '$name' extension is already disabled.");
                    return;
                } else {
                    $this->info("Disabling '$name' extension...");
                    $this->extensionManager->disable($name);
                }
                break;
        }
```

SychO9 (commented):
This is neat! Thanks! some comments.

SychO9 (approved):

framework/core/src/Extension/Console/ToggleExtensionCommand.php
```diff
@@ -0,0 +1,78 @@
...
+        $enabling = $this->input->getFirstArgument() === 'extension:enable';
+
+        if ($this->extensionManager->getExtension($name) === null) {
+            $this->error("There are no extensions by the ID of '$name'.");
+
+            return;
+        }
+
+        switch ($enabling) {
+            case true:
```
askvortsov1, line 56: I feel like an if/else would suffice here? I don't see how we're going to get any more cases than that.

askvortsov1 (approved):
Are we planning to eventually add some kind of "enable/disable all" to this, or would that (and bisection) be a separate command?

framework/core/src/Extension/Console/ToggleExtensionCommand.php
SychO9: honestly true, I've been using the php8 `match` so often lately that my brain tries to produce something similar 😆

SychO9:
> Are we planning to eventually add some kind of "enable/disable all" to this, or would that (and bisection) be a separate command?

The enabl/disable all could be eventually part of this same command I think, we'd probably use an `all` argument or `--all` option. 

The bisect should be its own command imo

framework/core/src/Extension/Console/ToggleExtensionCommand.php
askvortsov1: Ah yes, pattern matching is indeed wonderful!

*suspiciously drops https://ocaml.org/docs/data-types on the way out*

PR #3818, opened 2023-05-02
fix(testing): always clear cache in integration test's tearDown

n-peugnet:

This prevent integration tests from interacting between each other through the cache. See https://github.com/flarum/framework/issues/3663#issuecomment-1531646474.

**Necessity**

**Confirmed**


SychO9 (approved):
Thanks!

PR #3822, opened 2023-05-08
Sort extensions in console info output

JoshyPHP:
Sort extensions in lexicographic order in the console `info` output. It makes it easier to visually parse when someone reports an issue. Has no other use or impact, does not change the order of the extensions as returned by the ExtensionManager.

**Screenshot**
The output should look like this, with `zlib` at the end of the list rather than lost between `sqlite3` and `ctype`.
```
$ ./flarum info
Flarum core: 1.7.2
PHP version: 8.2.5
MySQL version: 10.6.12-MariaDB-log
Loaded extensions: Core, ctype, curl, date, dba, dom, fileinfo, filter, gd, hash, iconv, intl, json, libxml, mbstring, mysqli, mysqlnd, openssl, pcre, PDO, pdo_mysql, pdo_sqlite, Phar, random, readline, Reflection, session, SimpleXML, sodium, SPL, sqlite3, standard, tokenizer, xml, xmlwriter, xsl, zip, zlib
| flarum-approval      | v1.7.0  |        |
| flarum-bbcode        | v1.7.0  |        |
| flarum-emoji         | v1.7.0  |        |
| flarum-flags         | v1.7.0  |        |
| flarum-lang-english  | v1.7.0  |        |
| flarum-likes         | v1.7.0  |        |
| flarum-lock          | v1.7.0  |        |
| flarum-markdown      | v1.7.0  |        |
| flarum-mentions      | v1.7.0  |        |
| flarum-statistics    | v1.7.0  |        |
| flarum-sticky        | v1.7.0  |        |
| flarum-subscriptions | v1.7.0  |        |
| flarum-suspend       | v1.7.0  |        |
| flarum-tags          | v1.7.1  |        |
```

**QA**

**Necessity**

**Confirmed**

### DIFF
--- a/framework/core/src/Foundation/Console/InfoCommand.php
+++ b/framework/core/src/Foundation/Console/InfoCommand.php
@@ -83,7 +83,9 @@ class InfoCommand extends AbstractCommand
         $this->output->writeln('<info>PHP version:</info> '.$this->appInfo->identifyPHPVersion());
         $this->output->writeln('<info>MySQL version:</info> '.$this->appInfo->identifyDatabaseVersion());
 
-        $phpExtensions = implode(', ', get_loaded_extensions());
+        $phpExtensions = get_loaded_extensions();
+        sort($phpExtensions, SORT_STRING | SORT_FLAG_CASE);
+        $phpExtensions = implode(', ', $phpExtensions);
         $this->output->writeln("<info>Loaded extensions:</info> $phpExtensions");
 
         $this->getExtensionTable()->render();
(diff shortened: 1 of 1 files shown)


framework/core/src/Foundation/Console/InfoCommand.php
```diff
@@ -118,7 +120,15 @@ private function getExtensionTable()
                 (new TableStyle)->setCellHeaderFormat('<info>%s</info>')
             );
 
-        foreach ($this->extensions->getEnabledExtensions() as $extension) {
+        $extensions = [];
+        foreach (array_values($this->extensions->getEnabledExtensions()) as $i => $extension) {
+            // Sort by ID, use their position as tiebreaker to prevent any potential name collision
```
JoshyPHP, line 125: Name collisions are *currently* impossible; This is future proofing.

davwheat:
Not sure how I feel about this. I would imagine, at the moment, it's listed by load order which can be helpful for troubleshooting extension incompatibility in some cases.

JoshyPHP:
For context, I wrote this patch after reading [this post](https://discuss.flarum.org/d/32838-post-discussion-error).

framework/core/src/Foundation/Console/InfoCommand.php
luceos: Name collisions are never possible because we use the shortened package name and if duplicates exists before the Extension Manager identifies the extension, they will simply be discarded at that step.

I think this section of the commit should be reverted.

JoshyPHP:
Not really worth spending any time on this.

framework/core/src/Foundation/Console/InfoCommand.php
JoshyPHP: Please feel free to fork and resubmit.

PR #3829, opened 2023-05-27
chore: recover bundlewatch

SychO9:
Reinstates bundlewatch into workflow checks.

### DIFF
--- a/.bundlewatch.config.json
+++ b/.bundlewatch.config.json
@@ -1,0 +1,13 @@
+{
+    "files": [
+        {
+            "path": "./framework/core/js/dist/*.js",
+            "maxSize": "150KB"
+        },
+        {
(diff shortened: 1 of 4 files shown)


askvortsov1 (approved):
~~Should we also add the extension dist files?~~

EDIT: nvm, didn't read correctly. LGTM!

matteocontrini:
Is the `CI_BRANCH_BASE: ${{ github.event.pull_request.base.ref }}` line actually needed?

Bundlewatch should already read the `GITHUB_BASE_REF` variable automatically as a fallback (through `ci-env`), which is equivalent.

Same for `CI_COMMIT_SHA` probably?

SychO9:
@matteocontrini I think it was trying to pick up master or main as a base, but I'm actually not sure, it doesn't hurt to have anyway. I find docs aren't that clear

PR #3830, opened 2023-05-31
chore: Upgrade dependencies

imorland:
Upgrades:

- Laravel 8.x to 10.x
- Symfony 5.x to 6.3


framework/core/src/Forum/ValidateCustomLess.php
```diff
@@ -68,7 +68,10 @@ function ($settings) use ($event) {
         );
 
         $assetsDir = $this->assets->getAssetsDir();
-        $this->assets->setAssetsDir(new FilesystemAdapter(new Filesystem(new NullAdapter)));
+
+        $adaptor = new InMemoryFilesystemAdapter();
+        $filesystem = new Filesystem($adaptor);
+        $this->assets->setAssetsDir(new FilesystemAdapter($filesystem, $adaptor));
```
davwheat:
```suggestion
        $adapter = new InMemoryFilesystemAdapter();
        $filesystem = new Filesystem($adapter);
        $this->assets->setAssetsDir(new FilesystemAdapter($filesystem, $adapter));
```

php-packages/testing/src/integration/Extension/ExtensionManagerIncludeCurrent.php
```diff
@@ -108,6 +107,9 @@ protected function setEnabledExtensions(array $enabledExtensions): void
      */
     protected function getAssetsFilesystem(): Cloud
     {
-        return new FilesystemAdapter(new FlysystemFilesystem(new Local($this->paths->public.'/assets'), ['url' => resolve('flarum.config')->url().'/assets']));
+        $adaptor = new LocalFilesystemAdapter($this->paths->public.'/assets');
+        $filesystem = new \League\Flysystem\Filesystem($adaptor);
+
+        return new FilesystemAdapter($filesystem, $adaptor);
```
davwheat:
```suggestion
        $adapter = new LocalFilesystemAdapter($this->paths->public.'/assets');
        $filesystem = new \League\Flysystem\Filesystem($adapter);

        return new FilesystemAdapter($filesystem, $adapter);
```

framework/core/tests/integration/api/users/CreateTest.php
```diff
@@ -247,18 +247,18 @@ public function cannot_create_user_with_invalid_avatar_uri()
...
-            'email' => 'test@machine.local',
-            'is_email_confirmed' => 1,
-            'avatar_url' =>  'https://127.0.0.1/image.png'
-        ], []);
+        // $regTokens[] = RegistrationToken::generate('flarum', '1', [
+        //     'username' => 'test',
+        //     'email' => 'test@machine.local',
+        //     'is_email_confirmed' => 1,
+        //     'avatar_url' =>  'https://127.0.0.1/image.png'
+        // ], []);
```
davwheat: Is there a reason we've commented this test out?

framework/core/src/Forum/ValidateCustomLess.php
luceos: Even though the naming has a typo. I would highly recommend to return to not using the temporary variables as they are not being used further down. Extra variables create unnecessary memory usage. It also doesn't really add any value specifying these out.

Ref: https://www.quora.com/Does-declaring-an-unnecessary-variable-in-PHP-consume-memory

framework/core/src/Foundation/Container.php
```diff
@@ -0,0 +1,21 @@
...
+
+class Container extends \Illuminate\Container\Container
+{
+    public function terminating(): void
+    {
+    }
+
+    public function terminate(): void
+    {
+    }
```
luceos, line 20: I don't see these methods in the Illuminate Container, what are they for?

framework/core/src/Http/Middleware/AuthenticateWithHeader.php
```diff
@@ -41,7 +41,7 @@ public function process(Request $request, Handler $handler): Response
                 $request = $request->withAttribute('apiKey', $key);
                 $request = $request->withAttribute('bypassThrottling', true);
             } elseif ($token = AccessToken::findValid($id)) {
-                $token->touch($request);
+                $token->touch(null, $request);
```
luceos: Can't we use:

```suggestion
                $token->touch(request: $request);
```

Or is the php version too low?

framework/core/src/Foundation/Container.php
SychO9: The illuminate container in laravel is the same as the laravel application, so some internal service providers in laravel call app->terminating().

The only way to deal with that is to either re-declare said providers (too much boilerplate) or declare these placeholder methods to that it won't error out.

This is also part of why I believe we should re-couple back into a more laravel like architecture.

framework/core/src/Http/Middleware/AuthenticateWithHeader.php
SychO9: php 8.1 has named parameters, so we can

framework/core/src/User/AvatarUploader.php
```diff
@@ -13,12 +13,13 @@
...
 use Illuminate\Support\Str;
 use Intervention\Image\Image;
+use Intervention\Image\ImageManager;
 
 class AvatarUploader
 {
     protected Filesystem $uploadDir;
 
-    public function __construct(Factory $filesystemFactory)
+    public function __construct(Factory $filesystemFactory, protected ImageManager $imageManager)
```
SychO9: why was this added? seems unused

framework/core/src/Http/Middleware/AuthenticateWithHeader.php
imorland: Yep, I think we can do that from 8.0 actually. Good shout 👍

framework/core/src/User/AvatarUploader.php
imorland: Probably a hangover from where I attempted to upgrade `intervention/image` on this branch 🙈 I'll check it out and revert

framework/core/src/Mail/SmtpDriver.php
```diff
@@ -12,11 +12,16 @@
...
+use Symfony\Component\Mailer\Transport\Dsn;
+use Symfony\Component\Mailer\Transport\Smtp\EsmtpTransportFactory;
+use Symfony\Component\Mailer\Transport\TransportInterface;
 
 class SmtpDriver implements DriverInterface
 {
+    public function __construct(protected EsmtpTransportFactory $factory)
+    {
+    }
+
```
SychO9: Let's stick to one single format for constructors that declare properties 👍🏼 it makes it easier to differantiate right away if the constructor directly declares props or not.

```suggestion
    public function __construct(
        protected EsmtpTransportFactory $factory
    ) {
    }

```

framework/core/src/User/LoginProvider.php
```diff
@@ -23,7 +23,7 @@
  */
 class LoginProvider extends AbstractModel
 {
-    protected $dates = ['created_at', 'last_login_at'];
+    protected $casts = ['created_at' => 'datetime', 'last_login_at' => 'datetime'];
```
SychO9:
```suggestion
    protected $casts = [
        'created_at' => 'datetime',
        'last_login_at' => 'datetime',
    ];
```

framework/core/src/Api/Controller/CreateTokenController.php
```diff
@@ -57,7 +57,7 @@ public function handle(ServerRequestInterface $request): ResponseInterface
         }
 
         // We do a first update here to log the IP/agent of the token creator, even if the token is never used afterwards
-        $token->touch($request);
+        $token->touch(null, $request);
```
SychO9: let's update all instances like these with a names parameter.

framework/core/src/Mail/Job/SendRawEmailJob.php
```diff
@@ -10,7 +10,7 @@
 namespace Flarum\Mail\Job;
 
 use Flarum\Queue\AbstractJob;
-use Illuminate\Contracts\Mail\Mailer;
+use Illuminate\Mail\Mailer;
```
SychO9: why switch out from the contract?

framework/core/src/Mail/FlarumLogTransport.php
```diff
@@ -1,33 +0,0 @@
...
-     * {@inheritdoc}
-     *
-     * @return int
-     */
-    public function send(Swift_Mime_SimpleMessage $message, &$failedRecipients = null)
-    {
-        $this->beforeSendPerformed($message);
-
-        // Overriden to use info, so the log driver works in non-debug mode.
-        $this->logger->info($this->getMimeEntityString($message));
```
SychO9, line 27: do we no loneger need `FlarumLogTransport`?

framework/core/src/Mail/MailServiceProvider.php
```diff
@@ -79,5 +77,7 @@ public function register(): void
 
             return $mailer;
         });
+
+        $this->container->alias('mailer', Mailer::class);
```
SychO9: can we use the contract?

SychO9 (changes requested):

framework/core/src/Mail/Job/SendRawEmailJob.php
imorland: Honestly, I have no idea why I made that change. Reverted

framework/core/tests/integration/api/users/CreateTest.php
imorland: This was one of the tests that fail locally but is fine in CI. Forgot to uncomment 🙈

SychO9 (approved):
This looks good to me. Please do keep track of breaking changes to document later on.

PR #3831, opened 2023-05-31
chore: upgrade to mithril 2.2

SychO9:
Upgrades to mithril 2.2

* Found one breaking change, a vnode with simple text no longer has the text in `vnode.text`, it is now a child node with the `#` tag. Any direct access to `vnode.text` should be replaced in extensions with using the `extractText` utility.

**Necessity**

**Confirmed**

**References**

- https://github.com/MithrilJS/mithril.js/pull/2670


imorland (approved):
Looks all good to me 👍

PR #3832, opened 2023-05-31
fix(mentions): missed post mentions UI changes

iPurpl3x:
**Patches https://github.com/flarum/framework/pull/3780**

- Fix UI to actually use the new `mentionedByCount` instead of checking the length of the `repliers` array.
- Fix the API, so that it serializes the correct value of `mentionedByCount` instead of `0` (else the UI doesn't behave as expected)

The issues are actually fixed, no new performance issue has been introduced.

**QA**

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/extensions/mentions/js/src/forum/addMentionedByList.js
+++ b/extensions/mentions/js/src/forum/addMentionedByList.js
@@ -118,7 +118,7 @@ export default function addMentionedByList() {
         });
 
       const limit = 4;
-      const overLimit = repliers.length > limit;
+      const overLimit = post.mentionedByCount() > limit;
 
       // Create a list of unique users who have replied. So even if a user has
       // replied twice, they will only be in this array once.
@@ -136,7 +136,7 @@ export default function addMentionedByList() {
       // others" name to the end of the list. Clicking on it will display a modal
       // with a full list of names.
       if (overLimit) {
-        const count = repliers.length - names.length;
+        const count = post.mentionedByCount() - names.length;
 
         names.push(app.translator.trans('flarum-mentions.forum.post.others_text', { count }));
       }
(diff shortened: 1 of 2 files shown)


SychO9:
@iPurpl3x can you explain what issue this is trying to fix? the `repliers` which is `post.mentionedBy()` is always the limit, so `repliers.length` is the same as `mentionedByCount()`.

`mentionedByCount()` is serialized to be able to determine if there are more mentions that the limit loaded.

iPurpl3x:
@SychO9 so as I understand the code, `repliers.length` is dependent on `replies.length` and `replies.length` comes from the `post.mentionedBy()` relationship, which usually only contains 4 items as this is the limit for performance. So if a post has 200 replies, `repliers.length == 4` & `post.mentionedByCount() == 200`. Or am I misinterpreting something?... 

Now this PR isn't complete yet, @imorland still needs to fix the API, so that `post.mentionedByCount() == 200` when the discussion is loaded, and not `post.mentionedByCount() == 0` which breaks the UI.

SychO9:
@iPurpl3x oh my bad, I mistakenly assumed it was modifying the dropdown listing logic. Looking closely now I see it is to do with the the replier names logic which I hadn't updated in my previous PR, which now makes it dead code. 

Still not sure how it breaks the Ui apart from not showing the more count though 🤔

imorland:
Actually, no other changes are required here, other than what @iPurpl3x has already pushed. The correct values are returned correctly after a fix to `fof/merge-discussions` was made (it was overriding `ShowDiscussionController`, which in turn was not updated after the recent core changes.


Before this change:
image

After this change:
image

imorland (approved):

SychO9:
I'm 99% confident the limit still applies, so there is no performance impact. Just a missing count to be able to open the modal and see all mentions, which is good cause it makes this into more trivial bug

@iPurpl3x can you please allow maintainers to push to this PR branch?

imorland:
I've pushed the changes suggested by @SychO9 

To clarify for the record, if an extension modifies `ShowDiscussionController` to include `firstPost` (or `lastPost`), the counts are not loaded for them, causing the API to return `0`, which in turn messes up the UI.

iPurpl3x:
@SychO9 I've added you with admin rights to our repo. I didn't find out how to allow all maintainers at once, but as admin you can now add other people.

SychO9 (approved):

PR #3839, opened 2023-06-21
chore: remove unused vars in catch

datlechin:
**Fixes #0000**

Remove unused vars in catch.

As of PHP 8.0.0, the variable name for a caught exception is optional. If not specified, the [catch](https://www.php.net/manual/en/language.exceptions.php) block will still execute but will not have access to the thrown object.

**Screenshot**

**QA**

**Necessity**

**Confirmed**

**Required changes:**


SychO9 (approved):
Thanks!

PR #3840, opened 2023-06-22
chore: corrected typos in routes comments

JamesShaver:
Corrected typos in the comments

**Fixes #0000**

**Screenshot**

**QA**

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/framework/core/src/Api/routes.php
+++ b/framework/core/src/Api/routes.php
@@ -26,21 +26,21 @@ return function (RouteCollection $map, RouteHandlerFactory $route) {
         $route->toController(Controller\ListAccessTokensController::class)
     );
 
-    // List access tokens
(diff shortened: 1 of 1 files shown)


askvortsov1 (approved):
Do we need all these comments? Feels like the code is fairly self explanatory

JamesShaver:
> Do we need all these comments? Feels like the code is fairly self explanatory

The code comments show up in IDEs, so I would say yes each method needs to be commented accurately.

davwheat:
If they were function definitions, sure. I don't oppose to any of these, though.

SychO9 (approved):
Thanks! I do agree with Sasha though, the code is readable enough not to need these comments.

PR #3844, opened 2023-06-23
chore(testing): run flarum/testing packages tests back again

n-peugnet:
These tests were not run on each framework build, I enabled them again. Two of the tests were failing so I disabled them for now, it is more important to have some tests running than none at all.

**Necessity**

**Confirmed**

**Required changes:**


n-peugnet:
Just a comment to make it clear that this PR removes [the CI that was run on each release in `flarum/testing` repo](https://github.com/flarum/testing/actions), as it now useless if the tests are run on each push to master in `flarum/framework`. It was failing anyway.

PR #3848, opened 2023-06-28
feat: make WelcomeHero extensible

imorland:
Construct the `WelcomeHero` component using `ItemList`s so that 3rd party extensions may better customize the component


luceos (changes requested):
I don't think distinguishing between title and subtitle items makes sense. For instance if you want a p on top of a h1 you will incorrectly assign the p to the title items.

I think one items list would be enough? That's why ItemList has priority?

SychO9:
I agree with the above, considering we're just stacking the two lists, one list would be enough and more flexible.

imorland:
Agreed, two is overkill. I will revise

SychO9 (approved):

PR #3855, opened 2023-07-13
fix(regression): Webpack not working on Windows with new export registry

dsevillamartin:
Caused by #3842.

- Use `path.relative` for module path
- Replace system path separators with `/`

Initially, since `.replace()` was hardcoding `/`, the entire absolute path was being added to `flarum.reg.add()`.
Then, building still errored because it was using `\` as path separators, and the following characters weren't valid escape sequences.

```js
ERROR in ./src/common/utils/styleSelectedText.ts 354:32
Module parse failed: Bad character escape sequence (354:32)
File was processed with these loaders:
 * ../../../js-packages/webpack-config/src/autoExportLoader.cjs
 * ../../../node_modules/babel-loader/lib/index.js
You may need an additional loader to handle the result of these loaders.
|   };
| }
> flarum.reg.add('core', 'common\utils\styleSelectedText', styleSelectedText)
```

Tried to use `path.posix`, but it can't normalize win32 paths into unix. There's a package for that ([slash](https://github.com/sindresorhus/slash)), but it basically just replaces `\` with `/` - I don't think this implementation should cause any issues.

**QA**
Build JS. Dist remains the exact same.

**Necessity**

**Confirmed**

### DIFF
--- a/js-packages/webpack-config/src/autoExportLoader.cjs
+++ b/js-packages/webpack-config/src/autoExportLoader.cjs
@@ -179,7 +179,9 @@ module.exports = function autoExportLoader(source) {
   // Get the path of the module to be exported
   // relative to the src directory.
   // Example: src/forum/components/UserCard.js => forum/components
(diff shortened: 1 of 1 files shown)


SychO9 (approved):
Thanks!

PR #3859, opened 2023-07-25
Fixed str_getcsv() on php 8.x

qianmoQ:
**Fixes #0000**

**Screenshot**

**QA**

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/framework/core/src/Search/GambitManager.php
+++ b/framework/core/src/Search/GambitManager.php
@@ -45,7 +45,12 @@ class GambitManager
      */
     protected function explode(string $query): array
     {
-        return str_getcsv($query, ' ');
+        // Fixed Deprecated: str_getcsv(): Passing null to parameter #1 ($string) of type string is deprecated on php 8.x
+        if ($query !== null && is_string($query)) {
+            return str_getcsv($query, ' ');
+        } else {
+            return [];
+        }
     }
 
     protected function applyGambits(SearchState $search, string $query): string


framework/core/src/Search/GambitManager.php
```diff
@@ -45,7 +45,12 @@ public function apply(SearchState $search, string $query): void
      */
     protected function explode(string $query): array
     {
-        return str_getcsv($query, ' ');
+        // Fixed Deprecated: str_getcsv(): Passing null to parameter #1 ($string) of type string is deprecated on php 8.x
```
luceos, line 48: Comments that explain the reason for a PR do not belong in the code. Please remove the comment.

framework/core/src/Search/GambitManager.php
```diff
@@ -45,7 +45,12 @@ public function apply(SearchState $search, string $query): void
...
+        if ($query !== null && is_string($query)) {
+            return str_getcsv($query, ' ');
+        } else {
+            return [];
+        }
```
luceos, line 53: The `is_string` check is unnecessary because the method `explode` already type hints (enforces) `$query` as a `string`.

As you have targeted Flarum 2.x, the minimum php version is 8.1. As such we can simplify your code like this:

```suggestion
        return str_getcsv($query ?? '', ' ');
```

Although I think it's weird that `null` is being passed to `$query` as the type hint disallows that.

luceos (commented):
Hello and thanks for your contribution.

I left a few remarks in your code and I would appreciate it if you would add a reason for this PR in your Pull Request information, not in the code.

I also believe that 2.x is not ready for use and with 2.x we target 8.1 anyway. Solving a deprecation warning now seems highly inefficient.

And for this reason I will be closing this PR. If you want to get rid of the deprecation warning, configure you PHP settings for production (which excludes deprecations).

PR #3862, opened 2023-08-09
chore: merge the app with the container & implement the ApplicationContract

SychO9:
Illuminate components always expect the app to be the container, but also expect the app to be implementing the laravel app contract. This means that very often between minor illuminate updates we get a call to a method on the app that doesn't exist in the Flarum app. This fixes the issue once and for all.

The application contract implementation was separated into a trait to reduce the bloat because all those methods we don't actually need/use, so it can serve as a readable way of telling.

**Necessity**

**Confirmed**


framework/core/src/Foundation/Config.php
```diff
@@ -69,4 +69,9 @@ public function offsetUnset($offset): void
     {
         throw new RuntimeException('The Config is immutable');
     }
+
+    public function environment()
```
askvortsov1: Is the lack of return type intentional?

php-packages/testing/src/integration/TestCase.php
```diff
@@ -46,6 +46,8 @@ protected function tearDown(): void
     protected function app()
     {
         if (is_null($this->app)) {
+            $this->config('env', 'testing');
```
askvortsov1, line 49: Why was this added / why wasn't it needed before?

framework/core/src/Foundation/Concerns/InteractsWithLaravel.php
```diff
@@ -0,0 +1,201 @@
...
+
+use Flarum\Locale\LocaleManager;
+use Illuminate\Support\Arr;
+use Illuminate\Support\Str;
+
+/**
+ * Implementation of the Laravel Application Contract,
+ * for the sake of better integration with Laravel packages/ecosystem.
+ */
+trait InteractsWithLaravel
```
askvortsov1, line 20: A lot of this stuff feels familiar, didn't we have it written somewhere before?

framework/core/src/Foundation/Concerns/InteractsWithLaravel.php
```diff
@@ -0,0 +1,201 @@
...
+        $this->terminatingCallbacks[] = $callback;
+
+        return $this;
+    }
+
+    public function terminate(): void
+    {
+        $index = 0;
+
+        while ($index < count($this->terminatingCallbacks)) {
```
askvortsov1: Why not a regular for loop?

framework/core/src/Foundation/Concerns/InteractsWithLaravel.php
```diff
@@ -0,0 +1,201 @@
...
+    {
+        return $this->joinPaths($this->paths->base, $path);
+    }
+
+    public function publicPath($path = ''): string
+    {
+        return $this->joinPaths($this->paths->public, $path);
+    }
+
+    public function storagePath($path = ''): string
```
askvortsov1, line 48: Don't we have this in `Paths`?

framework/core/src/Foundation/Concerns/InteractsWithLaravel.php
```diff
@@ -0,0 +1,201 @@
...
+        if (count($environments) > 0) {
+            $patterns = is_array($environments[0]) ? $environments[0] : $environments;
+
+            return Str::is($patterns, $this['env']);
+        }
+
+        return $this['env'];
+    }
+
+    public function runningInConsole(): bool
```
askvortsov1, line 119: Do we need to cache this?

framework/core/src/Foundation/Concerns/InteractsWithLaravel.php
```diff
@@ -0,0 +1,201 @@
...
+        return null;
+    }
+
+    public function isDownForMaintenance(): bool
+    {
+        // TODO: Implement isDownForMaintenance() method.
+        return false;
+    }
+
+    public function registerConfiguredProviders()
```
askvortsov1: What's up with the empty ones?

php-packages/testing/src/integration/TestCase.php
SychO9: since we now implement the laravel `ApplicationContract` which has the  `environement` method that can be called internally in illuminate components, we might as well also make sure the value is correct in testing. (that's the gist of the changes in this PR)

framework/core/src/Foundation/Concerns/InteractsWithLaravel.php
SychO9: predates my time in flarum, but it might have implemented the laravel ApplicationContract at some point and then removed.

framework/core/src/Foundation/Concerns/InteractsWithLaravel.php
SychO9: yes, all these methods are implementations of the ApplicationContract, the method itself uses `Paths`

framework/core/src/Foundation/Concerns/InteractsWithLaravel.php
SychO9: it's cached

framework/core/src/Foundation/Concerns/InteractsWithLaravel.php
SychO9: has no actual meaning within our architecture, but have to implement the contract.

framework/core/src/Foundation/Concerns/InteractsWithLaravel.php
askvortsov1: Sorry, poor wording on my end. I mean, is this expensive to compute / is there a reason why we're caching it?

framework/core/src/Foundation/Concerns/InteractsWithLaravel.php
askvortsov1: Are any of the methods / code in this trait used for our needs at all?

askvortsov1: Also, could we mark all these as deprecated to discourage using them in Flarum code?

framework/core/src/Foundation/Concerns/InteractsWithLaravel.php
SychO9: oh, no it's not 🙈

framework/core/src/Foundation/Concerns/InteractsWithLaravel.php
SychO9: > Are any of the methods / code in this trait used for our needs at all?

Indirectly, within illuminate components we use, like `runningInUnitTests` or `runningInConsole`, `terminating`. At any moment an Illuminate component could start calling one of these methods.

PR #3868, opened 2023-08-17
chore: use grid & flexbox for the discussion list item

SychO9:
**Part of flarum/issue-archive#355**

* Refactors the discussion list item component t use flexbox and grid
* Makes its structure more easily extensible and adds item lists.

**Necessity**

**Confirmed**


SychO9:
ref: https://github.com/flarum/framework/pull/3867#issuecomment-1697512660

PR #3870, opened 2023-08-19
chore: Test using MySQL 8.1 also

imorland:
As title, adds MySQL 8.1.0 to the DBs tested.

Q: I think this should probably be run against the `1.x` branch too, what do others think?


luceos:
> Q: I think this should probably be run against the `1.x` branch too, what do others think?

Yeah definitely. We will support 1.x for quite a while still.

SychO9 (approved):

PR #3872, opened 2023-08-19
feat: notification unsubscribe & email overhaul with HTML multipart

imorland:
#### Removals, changes & additions
`SendRawEmailJob` is removed, and replaced with `SendInformationalEmailJob`.

Usage example:
```php
$queue->push(new SendInformationalEmailJob(email: 'user@example.com', subject: 'emailSubject', forumTitle: 'forumTitle', bodyTitle: 'informationHeader'));
```

#### Outgoing email format

image

By setting the `mail_format` to `text`, you're instructing Flarum to only send emails in the plain text format, disregarding the HTML version and multipart option. Adjust this setting according to your needs, but it is recommended to leave it at it's default.

#### Notification templates
Flarum now has two types of notifications/emails. `Notification` and `Informational`. Informational emails are things like, password confirm, test, etc. Basically an email that does not require (or should not have) an unsubscribe link and are classified as "essential" or "service" mail.

For all other emails, this should probably fall into the `Notification` type. These notifications automatically generate an unsubscribe link and this is rendered in the footer of the email (both HTML and plain).

For this reason, when constructing `blade` templates for emails, you should extend one of the base types:

##### HTML
- `@extends('flarum.forum::email.html.notification.base')`
- `@extends('flarum.forum::email.html.information.base')`

When rendering content and/or preview content in a HTML template, it must be wrapped in the utility `$formatter->convert($yourContent)` For example:

```php
{!! $formatter->convert($translator->trans('my_key')) !!}
```

or, for a `Post` object:
```php
{!! $blueprint->formatContent() !!}
```

##### Plain
- `@extends('flarum.forum::email.plain.notification.base')`
- `@extends('flarum.forum::email.plain.information.base')`

Both templates allow for sectioned data to be rendered. It is no longer required to include the user greeting, this is now contained in a translation key `core.email.greeting`, for example "Hey username".

```
@section('content')
{YOUR_CONTENT_HERE}
@endsection
```

```
@section('contentPreview')
{YOUR_CONTENT_HERE}
@endsection
```
 
`contentPreview` is optional, if your blade template does not provide it, nothing will be rendered in it's template area. `content` is always required.

Emails now also have a "signoff" section. Here, the translation key `core.email.signoff` by default will render "The {YOUR_FORUM_NAME} team".

Should you wish to disable either the greeting or signoff, this can be done on a template by template method:

 `@extends('flarum.forum::email.html.notification.base', ['greeting' => false, 'signoff' => false])`

---

- [X] Create a new model `UnsubsribeToken`
- [X] Modify `MailableInterface` so that notification blueprints return both `plain` and `html` views
- [X] Create base blade templates for both `plain` and `html` for all notification emails to extend.
- [X] One click unsubscribe option for the specific notification type
- [x] Redesign the notification content
- [X] Allow admins to specify `multipart`, `plain` or `html` emails from the forum
- [x] Generic multipart template in place of the old `raw` emails (email test, forgot password, etc)

##### Screenshots
Example `informational` email (email test):
image

Example `notification` email (desktop):
image

After clicking to unsubscribe from a notification type, the user is asked to confirm:
image

...

##### Changes:

- Refactored Email Templates: All notification emails now have a consistent base structure, making it easier for extensions to create or modify their notifications.
- HTML Email Support: In addition to the existing plain text email templates, support for HTML email templates has been added. This provides a richer and more visually appealing experience for users. Emails are now sent multipart, with both plain text and html versions.
- Added Unsubscription: Users can now easily unsubscribe from specific types of notifications with a direct link in the email, enhancing user experience and reducing unwanted emails.


framework/core/src/Notification/MailableInterface.php
```diff
@@ -15,11 +15,26 @@ interface MailableInterface
...
+     * Example:
+     * ```php
+     * return [
+     *  'text' => 'flarum-subscriptions::emails.plain.newPost',
+     *  'html' => 'flarum-subscriptions::emails.html.newPost'
+     * ];
+     * ```
      */
-    public function getEmailView(): string|array;
+    public function getEmailView(): array;
```
askvortsov1: Can we use PHPStan's docblock syntax to annotate the returned shape? https://phpstan.org/writing-php-code/phpdoc-types#array-shapes

framework/core/src/Notification/UnsubscribeToken.php
```diff
@@ -0,0 +1,49 @@
...
+     * The table associated with the model.
+     *
+     * @var string
+     */
+    protected $table = 'unsubscribe_tokens';
+
+    /**
+     * The attributes that should be cast to native types.
+     *
+     * @var array
```
askvortsov1: I think we should be able to remove the doc types/comments on these, since they should be inherited from the parent class?

framework/core/src/Forum/Controller/UnsubscribeController.php
```diff
@@ -0,0 +1,60 @@
...
+        $userId = Arr::get($request->getQueryParams(), 'userId');
+        $token = Arr::get($request->getQueryParams(), 'token');
+
+        // Fetch the unsubscribe token record
+        $unsubscribeRecord = UnsubscribeToken::where('user_id', $userId)
+            ->where('token', $token)
+            ->first();
+
+        // If record exists and has not been used before
+        if ($unsubscribeRecord && ! $unsubscribeRecord->unsubscribed_at) {
```
askvortsov1: These probably need to be added to the model via phpdoc comments

framework/core/src/Forum/Controller/UnsubscribeController.php
```diff
@@ -0,0 +1,60 @@
...
+            $unsubscribeRecord->unsubscribed_at = Carbon::now();
+            $unsubscribeRecord->save();
+
+            // Update user preferences
+            /** @var User $user */
+            $user = User::find($userId);
+            $user->setPreference('notify_'.$unsubscribeRecord->email_type.'_email', false);
+            $user->save();
+
+            $message = 'You have successfully unsubscribed from this type of email notification. If you wish to receive it again, please update your settings.';
```
askvortsov1: Should these be translated?

framework/core/src/Forum/Controller/UnsubscribeController.php
```diff
@@ -0,0 +1,60 @@
...
+        // If record exists and has not been used before
+        if ($unsubscribeRecord && ! $unsubscribeRecord->unsubscribed_at) {
+            // Mark as unsubscribed
+            $unsubscribeRecord->unsubscribed_at = Carbon::now();
+            $unsubscribeRecord->save();
+
+            // Update user preferences
+            /** @var User $user */
+            $user = User::find($userId);
+            $user->setPreference('notify_'.$unsubscribeRecord->email_type.'_email', false);
```
askvortsov1: Maybe this should become a method on the User class?

askvortsov1 (commented):
I like this a lot, great job! My only thought: maybe we should have the unsubscribe controller redirect to the settings page, with a payload in the URL that would trigger an alert / popup / etc?

framework/core/src/Notification/MailableInterface.php
imorland: Thanks for the tip, I wasn't aware of this :)

framework/core/src/Forum/Controller/UnsubscribeController.php
imorland: Yes, absolutely. I'm still deciding on exactly what to do with this page/view atm, but I will remove all the hard coded strings and replace with translations soon

imorland:
> I like this a lot, great job! My only thought: maybe we should have the unsubscribe controller redirect to the settings page, with a payload in the URL that would trigger an alert / popup / etc?

I did consider that at one point, however that would also require the user to be either logged in, or be bounced thru an auth flow before bing able to present their settings.

My understanding for `Unsubscribe-List` is that it should ideally be 1-click. Perhaps we can do this by using the one-time token already generated for the unsubscribe to log the user in? 🤔 Then again, that would also be a security risk, as anyone with that email would then be able to access the account (if the token has not already been used).

askvortsov1:
That's a good point. I think "unsubscribe on click + link to settings page" is a good compromise.

Instead of unsubscribing on the GET request, maybe that page should be a form (autosubmitted via JS), that pings a POST endpoint?

imorland:
> Instead of unsubscribing on the GET request, maybe that page should be a form (autosubmitted via JS), that pings a POST endpoint?

And if JS is disabled/not available in the email client webview/whatever?

framework/core/src/Forum/Controller/UnsubscribeController.php
luceos, line 1: This controller seems to imply that if you hit the URL it will immediately unsubscribe. Past experiences with clients like Gmail are that they scan links and as such will hit them ahead of any human interaction. This is also one of the reasons other unsubscribe pages demand you to confirm using a form.

I think a better approach would be to render the form and immediately show all active subscriptions which are all toggleable. This is a practice I've seen other mail list providers take and which would be a huge benefit in terms of UX and GDPR friendliness.

What do you think?

imorland: > This controller seems to imply that if you hit the URL it will immediately unsubscribe. Past experiences with clients like Gmail are that they scan links and as such will hit them ahead of any human interaction. This is also one of the reasons other unsubscribe pages demand you to confirm using a form.

Yes, the controller does immediately unsubscribe from the given notification type. I'm looking at using these headers to negate the problem you describe actually:

```
List-Unsubscribe:<https://dev.flarum.one/notifications/unsubscribe/1/1234>
List-Unsubscribe-Post: List-Unsubscribe=One-Click
```

This will cause the email client to `POST` rather than `GET`, and avoid any link scanning, etc.

> 
> I think a better approach would be to render the form and immediately show all active subscriptions which are all toggleable. This is a practice I've seen other mail list providers take and which would be a huge benefit in terms of UX and GDPR friendliness.
> 
> What do you think?

I'm leaning towards giving users maximum flexibility, but retain ease-of-use. We provide two links in the email footer:

- Unsubscribe from this type (eg postMention)
- Go to notification settings.

The second link currently directs users to `/settings`, but that's no good if they are not currently logged in, so I'm thinking instead of linking there, we present a page as you suggest listing all of their email subs and allow to toggle on/off from there.

I'd prefer to keep the first link "one-click" for friendliness and ease, but I'm open to the idea of tweaking it about, and adding a further link on the confirmation page to manage further email subs..

I'll continue to play about with this, let's see where we end up with this!

askvortsov1:
> > Instead of unsubscribing on the GET request, maybe that page should be a form (autosubmitted via JS), that pings a POST endpoint?
> 
> And if JS is disabled/not available in the email client webview/whatever?

Then show a button. Most unsubscribe links I've experience ask you to click a button to confirm; the annoying ones are where you need to click a bunch of checkboxes. What I'm proposing is pretty much what's discussed in the comment thread started by @luceos

imorland:
Yeah, sorry I was having a "moment" 🙈 

I'm already working on this, should be ready shortly

extensions/mentions/views/emails/html/userMentioned.blade.php
```diff
@@ -0,0 +1,13 @@
...
+@section('notificationContent')
+{!! $html->render($translator->trans('flarum-mentions.email.user_mentioned.html.body', [
+'{mentioner_display_name}' => $blueprint->post->user->display_name,
+'{title}' => $blueprint->post->discussion->title,
+'{url}' => $url->to('forum')->route('discussion', ['id' => $blueprint->post->discussion_id, 'near' => $blueprint->post->number])
+])) !!}
+@endsection
+
+@section('contentPreview')
+    {!! $html->render($blueprint->post->content) !!}
```
SychO9: for actual `posts` can we just call `->formatContent()` instead.

framework/core/src/Formatter/HtmlRenderer.php
```diff
@@ -0,0 +1,25 @@
...
+class HtmlRenderer
+{
+    public function __construct(public Formatter $formatter)
+    {
+    }
+
+    public function render(string $content): string
+    {
+        return $this->formatter->render($this->formatter->parse($content));
+    }
```
SychO9: rather than a brand new class, let's just move this to the formatter class and name the method `convert`.

framework/core/src/Forum/ForumServiceProvider.php
```diff
@@ -147,7 +148,8 @@ public function boot(Container $container, Dispatcher $events, Factory $view): v
 
         $view->share([
             'translator' => $container->make(TranslatorInterface::class),
-            'settings' => $container->make(SettingsRepositoryInterface::class)
+            'settings' => $container->make(SettingsRepositoryInterface::class),
+            'html' => $container->make(HtmlRenderer::class)
```
SychO9: let's rename the `html` var to `formatter` as that's more clear.

framework/core/src/Forum/Controller/UnsubscribeActionController.php
```diff
@@ -0,0 +1,65 @@
...
+    public function handle(Request $request): ResponseInterface
+    {
+        $parsedBody = $request->getParsedBody();
+        $token = Arr::get($parsedBody, 'token');
+        $userId = Arr::get($parsedBody, 'userId');
+
+        // Perhaps we should use a command instead of all the logic being here. WIP...
+        // $this->bus->dispatch(
+        //     new UnsubscribeEmail($user, $token)
+        // );
```
SychO9: leftover wip code

framework/core/src/Forum/Controller/UnsubscribeActionController.php
```diff
@@ -0,0 +1,65 @@
...
+
+            // Update user preferences
+            /** @var User $user */
+            $user = User::find($userId);
+            $user->setNotificationPreference($unsubscribeRecord->email_type, 'email', false);
+            $user->save();
+        }
+
+        // Redirect back to the forum's home page after unsubscribing
+        return new RedirectResponse($this->url->to('forum')->base());
```
SychO9: The code is pretty readable so i'd drop the comments :P 

```suggestion
        /** @var UnsubscribeToken|null $unsubscribeRecord */
        $unsubscribeRecord = UnsubscribeToken::where('user_id', $userId)
            ->where('token', $token)
            ->first();

        if ($unsubscribeRecord && empty($unsubscribeRecord->unsubscribed_at)) {
            $unsubscribeRecord->unsubscribed_at = Carbon::now();
            $unsubscribeRecord->save();

            /** @var User $user */
            $user = User::find($userId);
            $user->setNotificationPreference($unsubscribeRecord->email_type, 'email', false);
            $user->save();
        }

        return new RedirectResponse($this->url->to('forum')->base());
```

framework/core/locale/core.yml
```diff
@@ -737,13 +747,28 @@ core:
...
+
+    # These translations are used by the "notification" email template.
+    notification:
+        default_title: "Notification"
+        footer:
+          main_text: "This email was sent to {email} because you are subscribed to \"{type}\" notifications on {forumTitle}."
+          unsubscribe_text: "If you'd like to stop receiving this type of notification, <a href=\"{unsubscribeLink}\">unsubscribe here</a>."
+          unsubscribe_text_plain: "If you'd like to stop receiving this type of notification, unsubscribe here: {unsubscribeLink}"
+          settings_text: "Manage your notification settings <a href=\"{settingsLink}\">here</a>."
+          settings_text_plain: "Manage your notification settings <a href=\"{settingsLink}\">here</a>."
```
SychO9: we might as well use markdown here too?

framework/core/locale/core.yml
```diff
@@ -727,6 +727,16 @@ core:
...
+    # Translations in this namespace are displayed by the email unsubscribe interface.
+    unsubscribe_email:
+      title: Unsubscribe Confirmation
+      return_to_forum: Back to {forumTitle}
+      confirm_button: Confirm
+      immediate_helptext: You will be immediately unsubscribed once you confirm.
+      confirm_message: "You have requested to unsubscribe from \"{type}\" email notifications from {forumTitle}."
+      success_message: "You have successfully unsubscribed from \"{type}\" notification from {forumTitle}. If you wish to receive them again, please <a href=\"{settingsLink}\">update your settings</a>."
+      invalid_message: "This unsubscribe link is invalid or has already been used. For any changes in your email notifications from {forumTitle}, please <a href=\"{settingsLink}\">check your settings</a>."
+
```
SychO9: we might as well use markdown here too?

framework/core/views/email/html/notification/base.blade.php
```diff
@@ -0,0 +1,18 @@
...
+@section('header')
+    <h2>{{ $title ?? $translator->trans('core.email.notification.default_title') }}</h2>
+@endsection
+
+@section('content')
+    @yield('notificationContent')
+    <div class="content-preview">
+        @yield('contentPreview')
+    </div>
+@endsection
```
SychO9: are we showing both the raw and the preview here?

SychO9 (changes requested):
I think the `email_format` config value should be moved to a setting instead so that we can make it available in the mail settings page for the admin.

framework/core/src/Foundation/Mailer.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
+/*
+ * This file is part of Flarum.
+ *
+ * For detailed copyright and license information, please view the
+ * LICENSE file that was distributed with this source code.
+ */
+
+namespace Flarum\Foundation;
```
SychO9: I think the mailer class belongs better within the `Flarum\Mail` namespace

framework/core/src/Notification/MailableInterface.php
```diff
@@ -14,12 +14,24 @@
...
 
     /**
      * Get the subject line for a notification email.
      */
     public function getEmailSubject(TranslatorInterface $translator): string;
+
+    /**
+     * Get the serialized type of this activity.
+     */
+    public static function getType(): string;
```
SychO9: looking at the code, i think it would be better to replace the `MailableInterface` typing in the send email job and notification mailer classes with `MailableInterface&BlueprintInterface`

SychO9 (changes requested):

extensions/mentions/views/emails/html/userMentioned.blade.php
imorland: I was aiming for a consistent formatting call, but have replaced this with your suggestion instead.

framework/core/views/email/html/notification/base.blade.php
imorland: No, `notificationContent` is the initial action description, ie `User mentioned you`, the `contentPreview` is the actual preview, if it is supplied (optional)

framework/core/views/email/html/base.blade.php
```diff
@@ -0,0 +1,80 @@
...
+    <div class="greeting">
+        <p>{!! $translator->trans('core.email.greeting', ['displayName' => $username]) !!}</p>
+    </div>
+    <div class="main-content">
+        @yield('content')
+    </div>
+    <div class="signoff">
+        <p>{!! $translator->trans('core.email.signoff', ['forumTitle' => $settings->get('forum_title')]) !!}</p>
+    </div>
+</div>
```
SychO9: the separation of the greeting from each mail (and future ones) doesn't seem right, isn't it basically enforced on every email now? what if I don't want my custom email to have that, or what if i'm going for something different?

framework/core/src/Settings/SettingsServiceProvider.php
```diff
@@ -24,6 +24,7 @@ public function register(): void
             return new Collection([
                 'theme_primary_color' => '#4D698E',
                 'theme_secondary_color' => '#4D698E',
+                'mail_format' => 'multipart',
```
SychO9, line 27: let's also add this to the installation process defaults `WriteSettings`

framework/core/src/Notification/NotificationMailer.php
```diff
@@ -9,34 +9,70 @@
...
+     *     html: string
+     * }
+     */
+    protected function getEmailViews(MailableInterface&BlueprintInterface $blueprint): array
+    {
+        $views = $blueprint->getEmailViews();
+
+        // check that both text and html views are provided
+        if (! Arr::has($views, ['text', 'html'])) {
+            throw new \InvalidArgumentException('Both text and html views must be provided to send an email notification of type'.$blueprint::getType().'.');
```
SychO9, line 73: should `text` here also be changed to `plain` based on recent commits? consistency would clear future confusion.

framework/core/views/email/html/base.blade.php
imorland: Extracting the greeting out, means that it's now easier to translate (just the once).

Previously, the greeting had to be implemented every time, and sometimes not consistently, this way, it is consistent.

framework/core/src/Notification/NotificationMailer.php
imorland: The Symfony mailer expects the keys `text` and `html`

framework/core/views/email/html/base.blade.php
SychO9: still, we are now enforcing it on every email without the ability to remove it. imo, if we had to choose between translating it once and allowing it to be editable/custom per email, the latter makes more sense to me.

imorland: I'm afraid I disagree here.

Having the greeting uniform over all emails, presents a consistent approach, which users can use to help in part to identify the email as genuine, as it will always greet them in the same way and their current display name.

Having the possibility of seperating this out for each email, whilst it does allow for maximum freedom, just doesn't outweigh the consistency aspect, in my mind at least

SychO9: I can imagine quite a few cases where an extension will want a mail layout without the enforced greeting. This can be solved by adding a check along the lines of `@if(isset($greeting) && !$greeting)` to this base template.

luceos: > I can imagine quite a few cases where an extension will want a mail layout without the enforced greeting. This can be solved by adding a check along the lines of `@if(isset($greeting) && !$greeting)` to this base template.

I was about to say the same thing. And I'd highly recommend against enforcing certain patterns, so this also includes the signoff.

Remember we are building software that is extensible, that means we are not to judge whether something is necessary. Especially with rewrites we need to be careful in allowing more extensibility and customizability instead of increasing opinionatedness.

@imorland remember that client that wanted a community where anonymity is key; it would be interesting for them to be able to disable greeting and signoff.

SychO9 (approved):

PR #3877, opened 2023-08-30
chore: normalize form semantics

SychO9:
**Fixes https://github.com/flarum/issue-archive/issues/106**

* Normalizes the semantics of the forms we use (`Form` `Form-group` `FieldSet`) and uses flexbox/grid to organize elements.

* There are no visual changes from before. These changes are more convenient when customizing the styling and building more similar structures.

**Necessity**

**Confirmed**


SychO9:

PR #3878, opened 2023-09-01
fix: styling regressions

SychO9:
**Fixes regressions in styling from the previous PRs**

* Post avatar is rendered twice, one to display on mobile and one on tablet-up. This removes the need for hacky code to display the avatar where wanted.
* Fixes the spacing of the post element, so that styling introduced by flags works as expected.

**Necessity**

**Confirmed**


SychO9:

PR #3880, opened 2023-09-02
feat: extensible global notices

SychO9:
**Fixes https://github.com/flarum/issue-archive/issues/77**

* Makes the top global alerts flexible and easy to extend rather than a hardcoded implementation.

**Screenshot**
image

**Necessity**

**Confirmed**


SychO9:

Issue #3884, opened 2023-09-17
Search drivers


SychO9:
This is a breakdown of the current implementation of searching/filtering, and a plan to implement search drivers. The work done prior to stable 1.0 presents a good base to build on top. But there is some refactoring to be done first.

## Current Implementation

The current system in place separates the concept of searching & filtering. So we have Gambits and Filters, which are in practice both filtering mechanisms that do the the same thing. This adds some complexity on the backend side as well as the frontend side where applying a filter (like when clicking on the following sidebar link to show subscribed to discussions) requires doing it in two formats by checking if a search is active or not.

The choice between filtering and searching rests on whether a `q` filter is active. From there we either apply basic database filtering, or searching.

The current `AbstractSearcher`s are a hardcoded database search driver implementation. 

#### What should not change:
* Driver per model. Each model searcher is driver-specific, the current default ones fall under the default database driver.
* Gambits(essentially filters) and mutators per driver model searcher.

#### Current search extensions
* Clark's [scout extension](https://github.com/clarkwinkelmann/flarum-ext-scout) overrides the Fulltext gambits, retrieves relevant result IDs from the search driver. Then applies a database query to grab those IDs on top of the visibility scoping and applied gambits if any by the hardcoded database search driver.
* Blomstra's [elastic search](https://github.com/blomstra/flarum-ext-search) applies access control and filters on its own.

![search-drivers-2--1](https://github.com/flarum/framework/assets/20267363/467d1b81-ef00-4d99-a117-a8a36dbbb5cd)

#### Gambits Vs Filters
Dropping gambits on the backend should be the first step. Gambits are just filters apply in a different format. We can transform gambits to API request filters from the frontend. The additional advantage of that is the frontend becomes aware of gambits which allows us to improve the search UI/UX after the work here is done.

#### Filterers VS Searchers
Removing gambits still means we need to have filters per searcher. We could simply replace the idea of gambits in searchers with filters, but that leaves some unnecessary duplication between Filterers Vs Searchers. 

It seems clear that we can simply do without the separate concept of a Filterer. If a search driver implements its own filtering (like Blomstra's elastic implementation) then it can do the filtering. If a driver does not (like Clark's scout) then it can fallback to the database driver to apply filters (like Clark's scout impl). The difference lies in what the driver does. So it needs to be able to fallback to the database driver after it has the results it needs (the same line of thinking applies to visibility scoping).

[...]

PR #3885, opened 2023-09-17
refactor: move gambits to frontend

SychO9:
**Part of #3884**

* Moves gambits to a frontend implementation where they are converted to filters.
* This only works when using `app.store.find` to search resources.
* There is still a lot of gambit concept code on the backend which will be dropped when following through other changes planned for the backend.

**Necessity**

**Confirmed**


SychO9:
I'll merge any and all search-related PRs into a separate https://github.com/flarum/framework/pull/3893 branch so that they can still be reviewed before merging this new branch into 2.x (I need to continue building on top of all the work done so far).

PR #3890, opened 2023-09-20
fix(1.x,suspend): suspended users can abuse avatar upload

SychO9:
Suspended users can abuse the avatar upload functionality. This prevents that y denying access.

**Necessity**

**Confirmed**


imorland (approved):
Not tested locally, but this seems sound 👍 

Would be nice to have a test for this too?

matteocontrini:
The corresponding issue should be closed: https://github.com/flarum/issue-archive/issues/214

PR #3892, opened 2023-09-26
refactor: merge filterer and searcher concepts

SychO9:
**Part of #3884**

* Completely removes the concept of gambits from the backend.
* Merges the filterer and searcher systems into a single searcher system.
* Shared Query namespace is dropped.
* Searcher now have Filters instead of Gambits (and Fulltext Filters).
* Forum query parameters: it used to be when you search, the URL is changed to the following `?q=lorem is:locked`. Since this doesn't work when the backend has to process that (not aware of gambits anymore). The URL is applied with filters auto transformed) into `?q=lorem&filter[locked]=true`. We now also translates a set of filters into gambits (to fill the search input).
* Extensions such as subscriptions no longer has to append to `q` to filter the discussion list, it only adds an actual filter.

**Necessity**

**Confirmed**


SychO9:
I'll merge any and all search-related PRs into a separate https://github.com/flarum/framework/pull/3893 branch so that they can still be reviewed before merging this new branch into 2.x (I need to continue building on top of all the work done so far).

PR #3893, opened 2023-09-26
feat: revamp search

### LINKED ISSUE #3884: Search drivers

SychO9:
**Closes #3884**

Part of the roadmap, each commit has a separate PR. Changes are made one at a time for easier review/readability.

***feel free to review within the related PR even if it's marked as merged.***

**Necessity**

**Confirmed**

**Required changes:**

- [x] Related documentation PR: https://github.com/flarum/docs/pull/467


SychO9:
Feel free to leave separate reviews in each PR for easier back n fourth if necessary.

SychO9:
Also, proof of concept search driver here: https://github.com/flarum/framework/pull/3902#issue-1949872262

imorland (approved):
Aside from it likely taking me a while to get used to, I _really_ do like this alot. Great job @SychO9 :)

imorland:
Just need to resolve the conflict, then good to marge. Are you able to tackle that @SychO9 ?

PR #3898, opened 2023-10-11
[1.x] Conditional extender instantiation

### LINKED ISSUE #3897: [1.x] `Conditional` extender instantiates the extenders array even when the conditional is false
### Current Behavior

The `Conditional` extender instantiates provided extenders even if the associated condition is `false`. This can lead to unintended side effects, especially if the constructors of these extenders perform actions. More critically, if the extender class doesn't even exist, it will result in a fatal error due to class not being found, regardless of the condition's value.

### Steps to Reproduce

Reference a nonexistent extender class within the Conditional extender.

```php
(new Extend\Conditional())
    ->when(false, [new NonexistentExtender()]);
```

Upon execution, PHP will throw a fatal error stating that the `NonexistentExtender` class was not found, even though the condition was set to false.

### Expected Behavior

The expected behavior is that the `Conditional` extender should only attempt to instantiate and use provided extenders when the associated condition evaluates to `true`, and should completely ignore them when the condition is `false`.

### Screenshots

_No response_

### Environment

- Flarum version: x.y.z
- PHP version: x.y.z

```
```

### Possible Solution

Instead of directly providing an array of extenders, wrap them in a callback, ie invokable class

### Additional Context

_No response_

imorland:
Addresses #3897

### DIFF
--- a/framework/core/src/Extend/Conditional.php
+++ b/framework/core/src/Extend/Conditional.php
@@ -29,7 +29,7 @@ class Conditional implements ExtenderInterface
      * - 'condition': a boolean or callable that should return a boolean.
      * - 'extenders': an array of extenders, a callable returning an array of extenders, or an invokable class string.
      *
-     * @var array<array{condition: bool|callable, extenders: ExtenderInterface[]}>
+     * @var array<array{condition: bool|callable|string, extenders: ExtenderInterface[]}>
      */
     protected $conditions = [];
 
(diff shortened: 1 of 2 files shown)


framework/core/src/Extend/Conditional.php
```diff
@@ -52,7 +80,15 @@ public function extend(Container $container, Extension $extension = null)
...
 
             if ($condition['condition']) {
-                foreach ($condition['extenders'] as $extender) {
+                $extenders = $condition['extenders'];
+
+                if (is_string($extenders)) {
+                    $extenders = $container->call($extenders);
+                } elseif (is_callable($extenders)) {
+                    $extenders = $container->call($extenders);
+                }
```
SychO9: Lets merge these
```suggestion
                if (is_string($extenders) || is_callable($extenders)) {
                    $extenders = $container->call($extenders);
                }
```

SychO9 (approved):

PR #3904, opened 2023-10-18
chore: remove ExtenderInterface[] as a conditional option, only support callable or ::class invoke

imorland:
Similar to #3900 , but removes passing an array from `extend.php`, thus preventing possible early/unwanted instatiation


extensions/mentions/extend.php
```diff
@@ -126,23 +124,5 @@
...
-
-            (new Extend\ApiController(Controller\ListDiscussionsController::class))
-                ->load([
-                    'firstPost.mentionsTags', 'lastPost.mentionsTags',
-                ]),
-
-            (new Extend\ApiController(Controller\ListPostsController::class))
-                ->load(['mentionsTags']),
-        ]),
+        ->whenExtensionEnabled('flarum-tags', TagExtender::class),
```
SychO9: please keep this inside the `extend.php` file as it's easier to read 🙏. with a simple `fn () => [...]`

SychO9 (approved):

PR #3908, opened 2023-10-25
feat(emoji): allow the user to set the cdn address

imdong:
* add cdn mirror address configuration items

Modified extension emoji, the cdn mirror address configuration item is added to the extended emoji settings page of the management interface.
This option affects the Preview address of emoji in the Post, Reply, and Reply edit box Autocomplete.

Some cdn only provide version 14.0.2 and do not provide a short address of 14, and if the user enters the wrong address, it cannot be displayed.
In addition, non-native English speakers, such as translation documents or comments unreasonable please understand

**Screenshot**

image

**QA**

**Necessity**

**Confirmed**


extensions/emoji/js/src/admin/index.js
```diff
@@ -0,0 +1,19 @@
+import app from 'flarum/admin/app';
+import {version} from "../forum/cdn";
```
SychO9: since this is needed from both `forum` and `admin` could you please move it to `common` and update import paths?

extensions/emoji/extend.php
```diff
@@ -30,4 +33,7 @@
         }),
 
     new Extend\Locales(__DIR__.'/locale'),
+
+    (new Extend\Settings)
+        ->serializeToForum('flarum-emoji.cdn', 'flarum-emoji.cdn'),
```
SychO9: let's drop the `cdn` const from `cdn.js` and instead add a default value here with that value `->default(key, value)`

imdong: Ok,I will process it a few days later, have a nice weekend^_^

imdong:
The cdn.js file has been moved to common , which should be done because it is common.

And the default address has been set to the default value of the configuration item, and the default setting has been removed from the js code.

But this change may have a potential problem:

The user sets up a cdn and then clears the settings to try to restore to the default configuration.

Then at this time, the page may report an error because it obtains a blank string. Of course, there is a default value from the beginning and there is no description of blank items.
However, the best solution here may be to add a' reset' button to the input box with a' default' configuration, which may be the logic of core.

SychO9 (approved):
Sorry for the delay! and thanks for the PR!

@imdong that's a fair point, however it's a problem with not just this setting, but a lot of settings in both core and 2nd/3rd part extensions, which hopefully can be fixed through https://github.com/flarum/framework/pull/3935.

PR #3925, opened 2023-11-06
fix(1.x,approval): correct `PostWasApproved` event trigger condition

### LINKED ISSUE #3921: Akismet False Positives Triggered by Flag Removal
### Current Behavior

I've noticed an increased number of false-positives in Akismet. It appears that when a moderator marks a user as a spammer using the FriendsOfFlarum Spamblock extension, the spam is reported as a false-positive. This happens because when the flag is removed, the `PostWasApproved` event from the `flarum/approval` extension is triggered, which leads to the Ham submission to Akismet. I have an extension that logs all events on the forum, and it confirms my observations.

https://github.com/FriendsOfFlarum/spamblock/blob/5f5aa0ce42068342d3e78612bd75e87eebe0cb49/src/Controllers/MarkAsSpammerController.php#L97

### Steps to Reproduce

1. Wait for Akismet to identify a post as spam.
2. Hide the identified spam post.
3. Remove the flag from the post using `DeleteFlags` command from `flarum/flags`.

### Expected Behavior

The Approval extension shouldn't trigger the `PostWasApproved` event when a flag is removed but only when the post is genuinely approved.

### Screenshots

image

### Environment

- Flarum version: 1.8.3
- PHP version: 8.2

### Possible Solution

I can create a PR for 1.x. However, I am uncertain whether I should fix this in `flarum/approval` or `flarum/akismet`?

https://github.com/flarum/framework/blob/f07336e204aef7876170174865e0b04c0ee7d16d/extensions/akismet/src/Listener/SubmitHam.php#L1-L45

rafaucau:
**Fixes #3921**

Ensure the `PostWasApproved` event is only raised for posts that are not hidden and introduce the `PostWasUnapproved` event to handle scenarios where a post is unapproved, particularly when it's the first post and the discussion needs to be hidden as a result.

**QA**

**Necessity**

**Confirmed**

- [x] Backend changes.


rafaucau:
I have noticed a problem (not introduced in this PR) with hiding discussions. When a post is hidden, the discussion is left empty. I will add a fix to this PR.

SychO9 (commented):
Thanks for the PR!

I have dived into the issue to understand the flow, and the main problem is that the `PostWasApproved` event is triggered always when hiding or restoring a post, regardless of actual approval status/action. Which does not make sense. If I am listening to the post approval even, I expect that there was a state change from `approved` to `unapproved`. But right now even if you hide any approved post, the event is triggered.

The changes here seem close but based on testing I feel this is closer to what we would want: https://github.com/flarum/framework/commit/bed386f8e4f1fa29e3696784ec325c84b2020bbc

What do you think?

I don't understand the second commit though:
- When the first post is hidden and the discussion is empty, the entire discussion is hidden away from other users so I don't think we need to change anything about it.
- The PostWasUnpproved even is identical to the Hidden event in terms of when it is triggered so a bit redundant.

rafaucau:
> When the first post is hidden and the discussion is empty, the entire discussion is hidden away from other users so I don't think we need to change anything about it.

I thought everyone could see the discussion without any post:

https://github.com/flarum/framework/assets/25438601/d9ab6552-c616-4024-963b-eaf6accb0494

Now I see that other users except the admin and the author don't see it.

---

> The PostWasUnpproved even is identical to the Hidden event in terms of when it is triggered so a bit redundant.

Maybe so, but it is more specific and can be useful in some cases such as here: https://github.com/flarum/framework/pull/3925/files#diff-dd866324feaa8087e4ee73330d8affaf8352fd52dde6ca007a837e49179957e7R27

SychO9:
I'm afraid it only seems to add more confusion, because the reason the `PostWasUnapproved` would get triggered is because `isHidden` was set true, which is what the Hidden event is for.

Also hiding != unapproving thought the two are already confusing right now

rafaucau:
hiding != unapproving but unapproving = hiding so it's more specific and can be useful in cases where someone wants to listen for a post unapproval event only

SychO9:
there is no unapproving. You can only change  `is_approved` to true. The endpoint will not change it to false even with `isApproved: false`. Triggering PostWasUnapproved when the state change is from `hidden=false;approved=any` to `hidden=true;approved=unchanged` is misleading. It only means what you actually intend to use is the `Hidden` event

rafaucau:
I have reverted the commit that added the `PostWasUnapproved` event and incorporated your changes.

rafaucau:
@SychO9 Can this be merged and released as v1.8.4? By sending real spam as false-positive it messes up Akismet's spam detection a bit (#3921), so it's worth releasing.

SychO9:
@rafaucau that's the plan, I should be able to get around to it today

SychO9 (approved):

PR #3927, opened 2023-11-10
chore: extract `buildSettingComponent` method into a `FormGroup` component

SychO9:
* Extracts what the method does into a component reusable across all the admin panel. I pondered moving this to common to allow using on the forum frontend as well, but not sure if the use case for it exists, are there extensions that add forms as in the admin side that could benefit from this?

**Necessity**

**Confirmed**


luceos (commented):
I remember having tried using this in the forum frontend at some point. So moving it to common would get an upvote from me.

askvortsov1 (approved):
+1 on putting this in common if possible, this is way preferable to writing custom forms by hand.

PR #3928, opened 2023-11-10
feat: ItemList component

SychO9:
* Introduces an `ItemList` component to simplify the way we define itemlists while maintaining a more readable structure.
* If we decide to replace the component item lists in he code-base with this, it would be a major breaking change.
* We do have an auto upgrade tool planned so it won't be as bad if we maintain the same references to the item lists, namely `Component.methodName` as item list key.

* Would we be alright with auto injecting `context={this}` using the webpack configuration?

**Necessity**

**Confirmed**

### DIFF
--- a/framework/core/less/forum/DiscussionPage.less
+++ b/framework/core/less/forum/DiscussionPage.less
@@ -40,6 +40,10 @@
 
       &-sidebar {
         margin-top: 0;
+
+        &-main {
+          height: 100%;
+        }
       }
     }
   }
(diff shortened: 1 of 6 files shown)


SychO9:
@clarkwinkelmann would love to get your opinion on this one if and when you get the chance 🙏🏼

framework/core/less/forum/DiscussionPage.less
```diff
@@ -40,6 +40,10 @@
 
       &-sidebar {
         margin-top: 0;
+
+        &-main {
+          height: 100%;
+        }
```
luceos, line 46: Are these changes necessary in this pr? Seems unrelated?

luceos (approved):
Left one remark. Seems good to go.

PR #3929, opened 2023-11-16
Add fallback timeout to reset `modalClosing` flag in `ModalManager`

iPurpl3x:
- Added a fallback timeout to the `animateHide` function to ensure the `modalClosing` flag is reset. This change aims to prevent potential issues where a modal could become unresponsive if the `modalClosing` flag remains set to `true`.

- The implementation of the timeout in the `animateHide` function. 
- Ensuring that the addition of the timeout and the clearTimeout call do not introduce any regressions or unintended side effects.

**QA**

**Necessity**

**Confirmed**

### DIFF
--- a/framework/core/js/src/common/components/ModalManager.tsx
+++ b/framework/core/js/src/common/components/ModalManager.tsx
@@ -173,7 +173,21 @@ export default class ModalManager extends Component<IModalManagerAttrs> {
       closedCallback();
     };
 
-    this.activeDialogElement.addEventListener('transitionend', afterModalClosedCallback, { once: true });
+    // Set a fallback timeout to ensure the modalClosing flag is reset
+    // If not reset, there could be another modal that cannot be closed anymore
+    // because the modalClosing flag is still set to true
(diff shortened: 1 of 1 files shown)


iPurpl3x:
@SychO9 can you explain why this got closed?

imorland:
@iPurpl3x I would guess because it's targetting `1.x`? Other than fixes, `1.x` is more or less in feature freeze. I'm sure if you could retarget for `2.x` we could accept this change 🤔 

I'll wait to see what @SychO9 thinks also...

SychO9:
I deleted the `1.x` branch to replace it with another to cleanup a few commits when working on the extension manager which is what seems to have auto closed this *on delete*.

reopening but we are pretty much only patching critical issues for 1.x to focus on 2.x development. We could include this in a patch should anything else (or more fixes) turn up for core as well but we unfortunately don't intend to make a new patch for every 1.x PR 🙏🏼, I hope you understand that just wouldn't be feasible.

PR #3931, opened 2023-11-22
perf(flags): add pagination to flags list

### LINKED ISSUE #3787: flarum/flags should use pagination
### Current Behavior

Currently in one of the forums I manage I have about 450 flagged posts that are somewhat stagnant, the thing is that when you click on the dropdown it takes some time to display the list (even if I remember correctly after 150 you can start to notice a slowdown), I think it would be good to introduce pagination to avoid this, similar to the one used in the notifications dropdown.

### Steps to Reproduce

1. Flag several hundred posts
2. Open the flags dropdown

### Screenshots

_No response_

### Environment

- Flarum version: 1.7.2
- PHP version: 8.2.4

### Possible Solution

Make use pagination to improve performance and avoid lots of requests.

OrdinaryJellyfish:
**Fixes #3787**

* Adds pagination to flags list

* If the pagination works properly
* If flags are ordered properly

**Screenshot**

**QA**

**Necessity**

**Confirmed**

### DIFF
--- a/extensions/flags/js/src/forum/components/FlagList.tsx
+++ b/extensions/flags/js/src/forum/components/FlagList.tsx
@@ -8,6 +8,9 @@ import HeaderListItem from 'flarum/forum/components/HeaderListItem';
 import type Mithril from 'mithril';
 import type Post from 'flarum/common/models/Post';
 import type FlagListState from '../states/FlagListState';
+import type Flag from '../models/Flag';
+import { Page } from 'flarum/common/states/PaginatedListState';
+
 
 export interface IFlagListAttrs extends ComponentAttrs {
   state: FlagListState;
@@ -35,10 +38,10 @@ export default class FlagList<CustomAttrs extends IFlagListAttrs = IFlagListAttr
     );
   }
 
-  content(state) {
+  content(state: FlagListState) {
     if (!state.isLoading() && state.hasItems()) {
-      return state.getPages().map((page) => {
-        return page.items.map((flag) => {
+      return state.getPages().map((page: Page<Flag>) => {
+        return page.items.map((flag: Flag) => {
           const post = flag.post() as Post;
 
           return (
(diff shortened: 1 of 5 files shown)


extensions/flags/js/src/forum/components/FlagList.tsx
```diff
@@ -16,49 +18,53 @@ export interface IFlagListAttrs extends ComponentAttrs {
...
         className="FlagList"
         title={app.translator.trans('flarum-flags.forum.flagged_posts.title')}
-        hasItems={flags.length}
-        loading={this.state.loading}
+        hasItems={state.hasItems()}
+        loading={state.isLoading()}
         emptyText={app.translator.trans('flarum-flags.forum.flagged_posts.empty_text')}
+        loadMore={() => state.hasNext() && !state.isLoadingNext() && state.loadNext()}
       >
-        <ul className="HeaderListGroup-content">
```
SychO9, line 33: any reason why the `ul` with `li` items is gone now?

extensions/flags/src/Api/Controller/ListFlagsController.php
```diff
@@ -28,26 +29,57 @@ class ListFlagsController extends AbstractListController
...
+            ->groupBy('post_id')
+            ->orderBy('created_at', 'DESC')
+            ->skip($offset)
+            ->take($limit + 1);
+
+        $flags = Flag::whereVisibleTo($actor)
+            ->select('flags.*')
+            ->joinSub($primaries, 'p', 'flags.id', '=', 'p.id')
+            ->latest()
+            ->get();
```
SychO9: I don't understand the two queries, what does it do and why not a simple paginated query?

SychO9 (commented):
Thanks for the PR! as a side not, you usually want to avoid committing anything in dist or dist-typings

extensions/flags/js/src/forum/components/FlagList.tsx
OrdinaryJellyfish: That was an accident! I will add it back.

extensions/flags/src/Api/Controller/ListFlagsController.php
OrdinaryJellyfish: I just copied how it was done from `NotificationRepository`. When I tried a single query it wouldn't load more past the first 20.

SychO9: Oh I see. Yea this querying logic is unique to notifications, so what you want to do is recover the old simple flag query and in addition apply a ->limit($limit + 1)->offset($offset) and it should be good to go

OrdinaryJellyfish: Alrighty!

SychO9 (approved):
Looks good to me!

PR #3935, opened 2023-11-23
feat: allow resetting settings to default

### LINKED ISSUE #3934: Empty settings don't return default values
## Bug Report

This has been an annoyance forever, as an extention developer.

For example, we have a setting, a string let's say. We can provide a default value for that, great. However, if that setting is saved as simply an empty string '', I would expect the provided default to be returned. That's not currently the case.

At present, the logic for deciding if the default should be returned is based on is_null, which only works if that setting key does not exist on the table. 

**Expected Behavior**
In the case where a setting key exists, but it has no value, the default (if provided) should be returned.

**Environment**
- Flarum version: any
- Website URL: n/a
- Webserver: any
- Hosting environment: any
- PHP version: any
- Browser: any

**Possible Solution**
I initially thought using `empty()`, but as @SychO9 pointed out, a string value of `"0"` would return `true`

OrdinaryJellyfish:
**Fixes #3934**

* Adds a forget method to settings extender. This method allows a callback returning a bool, which if returns true, deletes the setting so the default is used.

* Whether this new extender was written properly
* If the forget method properly deletes a setting when callback returns true

**Screenshot**

**QA**

**Necessity**

**Confirmed**

**Required changes:**


SychO9 (commented):
Close enough but not quite ;)

The problem we are trying to solve is that, when the admin saves a value (empty string for example) that will be used as the setting value, even if an extension developer adds a setting that must be filled.

So the idea is that the callback will be used in the `SetSettingsController` class to determine which values to save and which to forget, because when an admin saves settings from the admin UI that's the controller that's used.

SychO9 (commented):
Haven't tested yet, But this looks good!

Now all we have to do (as per the conversation with Sasha on the issue) is to inverse the intent (and the name) of the `forget` extender. So instead of saying *forget this setting when the value equals this*. It would become *store this setting when the value is this*.

For naming, maybe `filter` would make sense, what do you think?

OrdinaryJellyfish:
Sounds good to me!

SychO9 (approved):
Thanks! Sorry for the delay.

Documentation added: https://github.com/flarum/docs/pull/478

PR #3945, opened 2023-12-14
fix: Search in Chinese

WaWaEMu:
Fixed the error that when searching in Chinese, the string cannot be searched if there is no exact match between characters.

**Fixes #0000**

**Screenshot**
Screenshot from 2023-12-14 11-11-00

**QA**

**Necessity**

**Confirmed**

**Required changes:**


SychO9 (commented):
Thanks for the PR!

I understand the intent behind replacing fulltext search with the simple `like` operator for CJK languages, as those have proved to require some config changes in mysql. But half dropping fulltext search here for all languages is likely not what we would want to do.

I think the best thing to do would be a second simple search driver which only utilizes the `LIKE` operator, or some replacement of the sort, 2.0 will allow drivers to be added more easily. We had planned to have a basic in-house search implementation as well, but that's never been done.

I would also recommend checking: https://discuss.flarum.org/d/22321-improving-search-for-chinese-characters

PR #3948, opened 2024-01-04
[1.x] fix: Logout controller allows open redirects

imorland:
Prevents open redirects on the `LogoutController`

By default, only return URL's on the forum host are permitted. Additional domains may be whitelisted using `config.php`:

```php
...
  'redirectDomains' =>
  array (
    'trusted.com`,
    'another-trusted.org'
  )
```

### DIFF
--- a/framework/core/src/Forum/Controller/LogOutController.php
+++ b/framework/core/src/Forum/Controller/LogOutController.php
@@ -124,10 +124,11 @@ class LogOutController implements RequestHandlerInterface
         return $this->rememberer->forget($response);
     }
 
-    protected function sanitizeReturnUrl(string $url, string $base): string
+    protected function sanitizeReturnUrl(string $url, string $base): Uri
     {
+        $parsedBase = new Uri($base);
         if (empty($url)) {
-            return $base; // Return base URL for empty return URL
+            return $parsedBase; // Return base URL for empty return URL
         }
         
         $parsed = new Uri($url);
@@ -135,10 +136,10 @@ class LogOutController implements RequestHandlerInterface
         $host = $parsed->getHost();
 
         if (in_array($host, $this->getWhitelistedRedirectDomains())) {
-            return $url;
+            return $parsed;
         }
 
-        return $base; // Return base url for non-whitelisted domains
+        return $parsedBase; // Return base url for non-whitelisted domains
     }
 
     protected function getWhitelistedRedirectDomains(): array
--- a/framework/core/src/Forum/Controller/LogOutController.php
+++ b/framework/core/src/Forum/Controller/LogOutController.php
@@ -93,8 +93,8 @@ class LogOutController implements RequestHandlerInterface
         $actor = RequestUtil::getActor($request);
         $base = $this->url->to('forum')->base();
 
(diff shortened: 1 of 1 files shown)


framework/core/src/Forum/Controller/LogOutController.php
```diff
@@ -81,29 +91,29 @@ public function handle(Request $request): ResponseInterface
     {
         $session = $request->getAttribute('session');
         $actor = RequestUtil::getActor($request);
+        $base = $this->url->to('forum')->base();
 
-        $url = Arr::get($request->getQueryParams(), 'return', $this->url->to('forum')->base());
+        $rurl = Arr::get($request->getQueryParams(), 'return');
```
luceos: Please use a sensible variable name like returnUrl here. Short variables will bring confusion in the long run.

framework/core/src/Forum/Controller/LogOutController.php
```diff
@@ -113,4 +123,33 @@ public function handle(Request $request): ResponseInterface
...
+        }
+
+        if (in_array($parsedUrl->getHost(), $this->getWhitelistedRedirectDomains())) {
+            return $parsedUrl;
+        }
+
+        return new Uri($base);
+    }
+
+    protected function getWhitelistedRedirectDomains(): array
```
luceos: Though not going into the politics of whitelisting. Can we use getAllowedRedirectDomains here?

askvortsov1 (approved):
Would be nice to see an integration test, but otherwise the code makes sense!

framework/core/src/Forum/Controller/LogOutController.php
```diff
@@ -113,4 +123,33 @@ public function handle(Request $request): ResponseInterface
...
+    {
+        $forumUri = new Uri($this->config->url());
```
luceos: Wasn't the config URL not already a Uri instance? This seems superfluous.

luceos (changes requested):
Haven't tested this, but left some comments for improvement.

framework/core/src/Forum/Controller/LogOutController.php
imorland: 🤦‍♂️ yes, you're absolutely right. Updated.

imorland:
> Would be nice to see an integration test, but otherwise the code makes sense!

I did try, however we need to make some changes to `flarum/testing` in order to pass the query string, as it looks like this is not currently supported.

For example:
```php
    public function logout_with_forum_redirect(string $returnUrl)
    {
        $encodedReturnUrl = urlencode($returnUrl);

        $response = $this->send(
            $this->request('GET', '/logout?return=' . $encodedReturnUrl)
        );

        $this->assertEquals(302, $response->getStatusCode());
        $this->assertEquals($returnUrl, $response->getHeaderLine('location'));
    }
```

This is a test I attempted to write for this, but we currently don't have the ability to pass the query, due to https://github.com/flarum/framework/blob/45a8b572e3bbedc80296e07d3125ba2f3c63f077/php-packages/testing/src/integration/TestCase.php#L269

luceos (approved):

PR #3953, opened 2024-01-09
[1.x] fix(Mentions): allow renderer to be used without context

DavideIadeluca:
**Changes proposed in this pull request:**
Allow post mentions to be used without context. As far as I understand, the same thing has already been done for user mentions in the past:

- https://github.com/flarum/mentions/commit/c9d70a844c0a9eb8e9c5d63eb2f7fee3185ac370
- https://github.com/flarum/mentions/commit/344abc0f87a1d1514f650b247f9d7511b5245088

**Reviewers should focus on:**
Focus if this PR introduces regression in other parts of the code. I'll also have a quick look and see if I can quickly create a test which tests this.

**Screenshot**
_Not applicable_

**QA**

**Necessity**

- [ ] Has the problem that is being solved here been clearly explained?
- [ ] If applicable, have various options for solving this problem been considered?
- [ ] For core PRs, does this need to be in core, or could it be in an extension?
- [ ] Are we willing to maintain this for years / potentially forever?

**Confirmed**

- [ ] Frontend changes: tested on a local Flarum installation.
- [ ] Backend changes: tests are green (run `composer test`).
- [ ] Core developer confirmed locally this works as intended.
- [ ] Tests have been added, or are not appropriate here.

**Required changes:**

- [ ] Related documentation PR: (Remove if irrelevant)
- [ ] Related core extension PRs: (Remove if irrelevant)

PR #3954, opened 2024-01-09
[2.x] fix(Mentions): allow renderer to be used without context

DavideIadeluca:
Related to [#3953](https://github.com/flarum/framework/pull/3953). See the description there for why this PR was opened. Also, please note that I have not actually tested it locally because I have not yet setup 2.x for development; therefore I don't know if the tests are actually passing.

I'll leave this PR as a Draft until I've confirmed this code actually works when I have 2.x setup or you may do it yourself.


imorland (approved):

imorland:
Confirmed as working as intended on my local 2.x instance

PR #3965, opened 2024-02-21
fix(tags): Edit tags icon fontawesome url to v5

dian-kaagman:
**Fixes #0000** (no issue)

The Fontawesome url referred to the newest icons overview, since the tags extension only supports v5 icons it's confusing for admin users why some icons don't work. The url for supported icons is: https://fontawesome.com/v5/icons?m=free.

Scherm­afbeelding 2024-02-21 om 13 48 20

-

**Screenshot**
-

**QA**
-

**Necessity**

**Confirmed**

### DIFF
--- a/extensions/tags/js/src/admin/components/EditTagModal.tsx
+++ b/extensions/tags/js/src/admin/components/EditTagModal.tsx
@@ -117,7 +117,7 @@ export default class EditTagModal extends FormModal<EditTagModalAttrs> {
       <div className="Form-group">
         <label>{app.translator.trans('flarum-tags.admin.edit_tag.icon_label')}</label>
         <div className="helpText">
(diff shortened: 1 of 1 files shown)


davwheat (approved):

askvortsov1 (approved):

luceos:
Although this PR makes a lot of sense, I think we need to aim to include FA 6 with 2.x. And this link, with the PR targeting 2.x, would as such then be incorrect. Maybe this change is best made to `main` branch instead.

My apologies for the nitpick @dian-kaagman ; je bijdrage is alsnog erg gewaardeerd 👍

SychO9:
Thank you for the PR and sorry for the delayed response! 2.x now supports FA v6 #4020. But the link could still use updating though to: 

https://fontawesome.com/search?o=r&m=free so that the page is actually the search for free icons one.

@dian-kaagman would you like to make that change?

imorland:
Implemented via https://github.com/flarum/framework/pull/4038, closing this, but thank you for your efforts here

PR #3970, opened 2024-03-04
fix: set parameter to string in explode function

pplulee:
**Fixes FriendsOfFlarum/user-directory#109**

The issue can be reproduced in PHP8.x environment when using the extension [FriendsOfFlarum/user-directory](https://github.com/FriendsOfFlarum/user-directory). Issue is caused by passing null parameter into function `str_getcsv()` in Flarum Core, where PHP8.x doesn't allow that.

Issue is solved by set the parameter to string.

GambitManager.php

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/framework/core/src/Search/GambitManager.php
+++ b/framework/core/src/Search/GambitManager.php
@@ -64,7 +64,7 @@ class GambitManager
      */
     protected function explode($query)
     {
-        return str_getcsv($query, ' ');
+        return str_getcsv((string)$query, ' ');
     }
 
     /**


SychO9:
Hello @pplulee thanks for the PR!

The 1.x branch is only currently receiving critical patches. This is a very minor issue and involves improving your web server configuration by disabling deprecation warnings in production to avoid the likely numerous similar warnings.

I would redirect your PR to 2.x, but the 2.x code has changed a lot and likely this is solved there.

Thanks again! Feel free to send more PRs in the future.

luceos:
I've confirmed this is not currently an issue in 2.x.

Thank you for your contribution, please suppress deprecations and warnings through your php.ini or using `ini_set` inside the config.php to go around this issue.

PR #3971, opened 2024-03-08
refactor: JSON:API

### LINKED ISSUE #3964: JSON:API Layer Refactor
We currently use [tobyzerner/json-api-php](https://github.com/tobyzerner/json-api-php) for our JSON:API layer implementation. Our implementation which is based on this package requires quite a lot of boilerplate to read and write resources. The primary goal is to reduce the boilerplate needed.

Additionally the package has been abandoned for a long time, though that should not actually matter, it still does work and it is very likely that regardless of the new chosen library, we will need to have a fork of our own where we can implement our own custom changes and unique use cases unsupported by said library (and sync back to the original what seems valid).

Crucial aspects needed in the revamp:
* Reduced boilerplate for serializing and writing resources.
* Maintained extensibility  of existing resources (read & write)
* Potentially if applicable, as little changes as possible to the related Extension API.

---

## Current API

Using the current implementation, a developer has to create the following boilerplate for full CRUD of a resource.

```
Api\Controller
 - CreateGroupController.php
 - ListGroupsController.php
 - ShowGroupController.php
 - UpdateGroupController.php
 - DeleteGroupController.php

Api\Serializer
 - GroupSerializer.php

Api\routes.php
 - ->get('/')
 - ->get('/{id}')
 - ->post('/')
 - ->patch('/{id}')
 - ->delete('/{id}')

Group\Command
 - CreateGroup.php
 - CreateGroupHandler.php
 - DeleteGroup.php
 - DeleteGroupHandler.php
 - EditGroup.php
 - EditGroupHandler.php
```

*The domain commands are optional, but still used a lot for re-usability, useful to at least keep in mind. as we do need to maintain that re-usability from within.

The implementations of most CRUD controllers are almost always identical. They follow the same usual steps. For example, the creation of a resource usually goes as:

```php
        // Access check
        $actor->assertRegistered();
        $actor->assertCan('createGroup');

        // Making of the resource
        $group = Group::build(
            Arr::get($data, 'attributes.nameSingular'),
            Arr::get($data, 'attributes.namePlural'),
            Arr::get($data, 'attributes.color'),
            Arr::get($data, 'attributes.icon'),
            Arr::get($data, 'attributes.isHidden', false)
        );

        // Pre-saving events
        $this->events->dispatch(
            new Saving($group, $actor, $data)
        );

        // Validation
        $this->validator->assertValid($group->getAttributes());

        // Saving
        $group->save();

        // Post-saving events
        $this->dispatchEventsFor($group, $actor);

        return $group;
```

Or listing:

```php
        // Request inputs
        $filters = $this->extractFilter($request);
        $sort = $this->extractSort($request);
        $sortIsDefault = $this->sortIsDefault($request);
        $limit = $this->extractLimit($request);

        // Querying/(Optional Searching)
        $queryResults = $this->search->query(
            Group::class,
            new SearchCriteria($actor, $filters, $limit, $offset, $sort, $sortIsDefault)
        );

        // Pagination links
        $document->addPaginationLinks(
            $this->url->to('api')->route('groups.index'),
            $request->getQueryParams(),
            $offset,
            $limit,
            $queryResults->areMoreResults() ? null : 0
        );

        $results = $queryResults->getResults();

        // Eager loading
        $this->loadRelations($results, [], $request);

        return $results;
```

## Alternative JSON:API packages

Most of the implementations of the json:api spec (https://jsonapi.org/implementations/#server-libraries-php) have not been updated in some years. Of the rest, only a few are framework agnostic, where the rest are specific to Laravel or Symfony frameworks. The Laravel ones are tightly coupled to Laravel so they expect Laravel's `Request` and `Routing` so those are not an option.

I believe the most appropriate package for this to be https://github.com/tobyzerner/json-api-server. From its documentation, the package simplifies the boilerplate necessary tremendously, its API is very readable and well put (and the code behind it is simple).

```php
class UsersResource extends EloquentResource
{
    public function type(): string
    {
        return 'users';
    }

    public function newModel(Context $context): object
    {
        return new User();
    }

    public function endpoints(): array
    {
        return [
            Endpoint\Show::make(),
            Endpoint\Index::make()->paginate(),
            Endpoint\Create::make()->visible(Laravel\can('create')),
            Endpoint\Update::make()->visible(Laravel\can('update')),
            Endpoint\Delete::make()->visible(Laravel\can('delete')),
        ];
    }

    public function fields(): array
    {
        return [
            Field\Attribute::make('name')
                ->type(Type\Str::make())
                ->writable()
                ->required(),

            Field\ToOne::make('address')->includable(),

            Field\ToMany::make('friends')
                ->type('users')
                ->includable(),
        ];
    }

    public function filters(): array
    {
        return [Filter\Where::make('id'), Filter\Where::make('name')];
    }
}
```

If we create an abstraction layer on top, we can have this integrated into Flarum to follow our internal needs and deal with changes from the underlying package without breaking our own API.

That would also enable us to enforce using our own filtering and searching systems, and preserve the same similar behavior of event dispatching like outlined in the previous section.

We will not escape having to fork and adapt any package of choice however as they are bound to have certain behavior we need changed. For example:
* Validation of attributes is scoped to the attribute's value only. Meaning validation rules such as `required_with` will not work.
* Once we pass control of the `/api` endpoint we would no longer be able to create simple controller endpoints in `/api` such as `/api/notifications/readAll`. So we need to change how routes lead to each resource endpoints.

## Propositions

There are two paths forward here:
* Although abandoned, we could keep the current `json-api` package and adopt it under `flarum` to continue maintaining it ourselves. We would then built the abstraction layer we require on top of it, or refactor it as needed, this would be a lot more work of course.
* We can go with the suggested package in the previous section. But still fork it to adapt it where necessary, and then add an abstraction layer either within core or the forked package (preferably core if not too much code).

The decision between the two will come down to some experimentation, but I feel it likely that the second option is the best.

---

***--> going with option 2***

#### BC Layer

Instead of directly using classes from the library, which is currently in beta and will have breaking changes. We need a layer on top to facilitate updating or switching to a different library. So for example, instead of directly extending `\Tobyz\JsonApiServer\Resource\AbstractResource` we create our own `AbstractResource` to extend from.

#### Routing

Instead of giving control of the entire `/api` endpoint over to the library, and therefore only resources. We need to selectively create the endpoint routes of each resource into our router (`/api/resource`). And invoke the library on each individual resource route. This maintains the ability to create `/api` routes other than resources.

#### Saving Flow

Flarum's flow of saving is:
* authorization -> fill data -> saving hook(event) -> validation -> save -> post-save-hooks(events).

The library's is:
* authorization -> validation -> fill data -> saving hook -> save -> post-save-hooks.

To maintain BC for the various model `Saving` events, we need it dispatched before validation, though we will very likely just break BC for this one.

#### Validation

The way validation works in the library is that each field is given its value from the request, and the field applies whatever validation, in isolation from other request data. This will not work for us as we need to be able to validate for example, that a field is required when another was not provided.

The validation relies on manual callbacks to be provided with the ability of passing Laravel rules using a helper:

```php
->validate(function ($value, $fail, $context) {
	// ..
})

->validate(Laravel\rules([...]))
```

Since we entirely rely on `illuminate\validation` we need to change things to allow directly passing illuminate rules, then gather all of that, and in the validation process use a single validator with all the data.

```php
Schema\Str::make('username') 
    ->rule('max:30')
    ->rule('min:3')
    ->rules(['regex/^[a-z0-9_-]+$/i'])
```

Which would also allow us to more easily create helper validation methods:

```php
Schema\Str::make('username')  
    ->requiredOnCreate()  
    ->unique('users', 'username', true)  
    ->regex('/^[a-z0-9_-]+$/i')  
    ->validationMessages([  
        'username.regex' => '...'
    ])  
    ->minLength(3)  
    ->maxLength(30)
```
#### Command Bus (domain commands, not console)

Flarum was built with the concept of command bus (not the console) where a domain operation takes place without any strong coupling to the controller. This makes calling these operations in different other contexts easily possible:

```php
new CreateGroup(User::find(1), ['attributes' => [  
    'nameSingular' => 'test group',  
    'namePlural' => 'test groups',  
    'color' => '#000000',  
    'icon' => 'fas fa-crown',  
]])
```

We need to maintain the ability to re-use the endpoint logic. The endpoint itself is pretty well isolated. It only needs a context object with a request.

With custom changes, we will have something along the lines of:

```php
$group = $api->forResource(GroupResource::class)  
    ->forEndpoint(Create::class)  
    ->execute([  
        'attributes' => [  
            'nameSingular' => 'test group',  
            'namePlural' => 'test groups',  
            'color' => '#000000',  
            'icon' => 'fas fa-crown',  
        ]  
    ]);
```

#### Relationship linkage vs Inclusion

Flarum has the unusual case of the show discussion endpoint, which adds the linkage of all the discussion's posts, but only includes a subset (This behavior is the basis for the post stream scrubber feature on the frontend). 

The library does not separate between the values used for linkage and inclusion. So at the moment we have to override its `Serializer` class to allow this distinction. Though preferably we need to look into a different solution for the post scrubber feature that doesn't require this behavior. I avoided that here to not go out of scope.

###### Problem 1

The library handles eager loading of included relationships by deferring getting the value of each relation and using a buffer of model relations to eager load.

Technically, eager loading does not *just* apply to included relationships however, all it takes is for a field visibility callback or getter callback to access a relation for that relation to require eager loading, so we need a manual API to select which relations to eager load on X endpoint.

We can't just port the exact logic we already use in our current implementation however:
1. We want to be able to manually eager load needed relations.
2. We don't want to replace the library's eager loading of included relations because that takes care of scoping those relations.

We need a hybrid process of:
1. before serialization, eager load R list of relations that will only be accessed internally and not included in the response.
2. while serializing eager load S list of included relations with scoping and with R' list of relations we need internally (subset of R).

This is still not perfect however.

###### Problem 2

The serialization flow of the library is:

- `addToPrimary`: Fields are added to a map after a visibility check and resolved immediately by default (unless a callback is used for the getter).
- `serialize` : Deferred fields are resolved and added to the map, then the rest of the serialization process happens.

***Problem:*** N+1 Queries are easily produced in this flow, especially coming from the visibility checker, and regardless of the eager loading API we're adding from Problem 1.

***Example***: Resource has one attribute A and `ToOne` relationship B, attribute A is only visible when B->col is true. `->visible(fn (object $a) => $a->B->col)` Even though the library does its own eager loading of included relationship B, the visibility callback is accessed earlier than that.

***Solution 1***: Eager load relation B twice. The first one for internal use (visibility checking), the second one for inclusion in the response (scoped).

***Solution 2***: Change the serialization flow. Defer all the fields, then prioritize resolving the relationship fields first. By the time visibility checking and getters of the attributes is accessed, relationship B will have been loaded. However if internally a relation is accessed which is not meant to be visible publicly, it'll leads to *some* N+1 queries.

--> Going for solution 1 as it's the least complicated and ensures internally accessed relations are always eager loaded while response included relations are always scoped. While this means *some* duplicate queries, it is still an improvement over the n+1 issue.
## Additions

The excellent implementation of the library allows us to add some Flarum specific features to it (in the goal of reducing boilerplate). This includes both features already existent in the current API implementation (like default includes) or new features.

#### Default included relationships

Unfortunately the current library has no way of specifying relationships that are by default included in the response document. This needs to be added as it is a thing in 1.x, we can do so by adding a `defaultInclude` method on endpoints:

```php
Endpoint\Show::make()
    ->defaultInclude(['groups', 'actor'])
```

However, I would prefer we move away from including relations by default, and instead, specifying on the frontend side on each request, what relationships to include. That way endpoints that require more data than others, don't affect the response size elsewhere. And extensions have taken the habit of just always using default includes which contributes to bad performance.

#### Custom Endpoints

This is a necessary addition. We have routes such as `DeleteAvatarController` which return a serialized user model, so it would better to have the ability to add custom endpoints within the resources, rather than try to re-use the JsonApi inside each similar custom route.
#### Custom request params extraction

Some of our endpoints -like the list posts endpoint- have cusom logic for extracting some of the request parameters -like the offset-. We need to support custom callbacks on at least the Index endpoint (potentially the Show endpoint as well) to allow customizing the logic for the extraction per resource.

```php
Endpoint\Index::make()  
    ->extractOffset(function (Context $context, array $defaultExtracts): int {  
        $queryParams = $context->request->getQueryParams();  
  
        if (($near = Arr::get($queryParams, 'page.near')) > 1) {  
            $sort = $defaultExtracts['sort'];  
            $filter = $defaultExtracts['filter'];  
  
            if (count($filter) > 1 || ! isset($filter['discussion']) || $sort) {  
                throw new BadRequestException(  
                    'You can only use page[near] with filter[discussion] and the default sort order'  
                );  
            }  
  
            $limit = $defaultExtracts['limit'];  
            $offset = resolve(PostRepository::class)->getIndexForNumber((int) $filter['discussion'], $near, $context->getActor());  
  
            return max(0, $offset - $limit / 2);  
        }  
  
        return $defaultExtracts['offset'];  
    })  
    ->paginate()
```

#### Endpoint Visibility

The library allows specifying when an endpoint is visible:

```php
Endpoint\Create::make()  
    ->visible(function (Context $context) {  
        $actor = RequestUtil::getActor($context->request);  
        return ! $actor->isGuest() && $actor->can('createGroup');  
    }),
```

We can add `authenticated` and `can` methods to simplify this:

```php
Endpoint\Create::make()  
    ->authenticated()
    ->can('createGroup'),
```

SychO9:
**Closes #3964**

This is a difficult review, I couldn't really find a way to split this since it can only be all done together,  I would recommend looking at some specific resources, for example, comparing the main branch Discussion model JSON API with this one. The controllers and serializer class and command handlers vs the DiscussionResource class.

Reading the issue itself will help you a lot in getting context about the changes here. Other than that, some key things to think about:
* Do we want to add out fork of the library into the monorepo? I'm leaning towards yes.
* Any specific concerns about the new API? everything that an extension dev could do before should be able to do now as well, only differently. Once we start updating FoF extensions we will be able to see and make any necessary changes.

**Some notable changes**
* The `validation.yml` file has been updated.
* The discussions sortmap is no longer bound to the container, instead the `SortColumn` on the DiscussionResource can receive an `ascendingAlias` and `descendngAlias` which can them easily be accessed in the content classes, this also means extensions can add to the sortmap by just the sorts they add themselves in the same manner.

**Todo**
* [x] Tag a 1.0 of the forked library: https://github.com/flarum/json-api-server
* [x] Extensively document the changes from 1.0 to extension devs. (https://github.com/flarum/docs/pull/472)


SychO9:
still a couple test failures to fix as well, and phpstan..

framework/core/src/Api/Resource/ForumResource.php
```diff
@@ -0,0 +1,168 @@
...
+
+    public function getId(object $model, \Tobyz\JsonApiServer\Context $context): string
+    {
+        return '1';
+    }
+
+    public function id(\Tobyz\JsonApiServer\Context $context): ?string
+    {
+        return '1';
+    }
```
luceos, line 58: For multi tenant communities, it would be helpful if this value could actually be set. So that if any json is visible to the world, it provides a human readable identifier. An option here would be to take that ID from the `config.php` if set, or fallback to 1. Unless this is extensible, it feels like a hardcoded value that might byte us in the future..

framework/core/src/Api/Resource/MailSettingResource.php
```diff
@@ -0,0 +1,88 @@
...
+    public function find(string $id, Context $context): ?object
+    {
+        return new stdClass();
+    }
+
+    public function endpoints(): array
+    {
+        return [
+            Endpoint\Show::make()
+                ->route('GET', '/')
```
luceos, line 58: Would defining a route using `get` be the same as `GET` for this library? To simplify this definition we might as well define an ENUM to be used to reduce the chance of typo's?

luceos (commented):
This is based off an initial reading from core changes, no extensions or implementation has been done.

For my understanding:

- serializers have been replaced with resources, and resources contain not just endpoint instruction, but also field instructions; I really like that
- many of the controllers have been simplified or removed entirely because of this refactor, which is also nice. I do think it will need some getting used to, to easily find where what is declared when looking things up.
- Can we somehow get rid of all the Command/Handlers and move them to the respective controllers where possible or would that better be done in a follow up PR?

Overall this looks good, I will start doing some implementation attempts for the federation extension so that I can fully understand how things work etc.

framework/core/src/Api/Resource/ForumResource.php
SychO9: That's a good idea, but not really tied to this PR. It was hardcoded as `1` before this aswell in the ForumSerializer.

Let's do that in a separate issue to not go out of scope 👍🏼

framework/core/src/Api/Resource/MailSettingResource.php
SychO9: Good shout, there are constants and enums available from the http packages we use so we can use those, but other than that, the underlying package needs to be changed to not be case sensitive. Will look into that.

SychO9:
> serializers have been replaced with resources, and resources contain not just endpoint instruction, but also field instructions

Yes, and columns. Filters have been disabled since we have our own driver-based filter implementation (which we can always improve in terms of boilerplate as well).

> many of the controllers have been simplified or removed entirely because of this refactor, which is also nice. I do think it will need some getting used to, to easily find where what is declared when looking things up.

Keep in mind, you can still create a separate controller the old fashion way, only there are non of the previous parent serializer classes. So the result of it cannot be the serialization of a resource. You can also just use a custom endpoint (see user resource delete avatar).

> Can we somehow get rid of all the Command/Handlers and move them to the respective controllers where possible or would that better be done in a follow up PR?

They are still useful reusable domain logic, for example check the delete avatar custom endpoint in the user resource class. Now that most controllers have been removed, command handlers aren't actually bad to have in terms of boilerplate. An additional improvement to them would be to make them one class rather than requiring two classes imo.

SychO9:
documentation added: https://github.com/flarum/docs/pull/472

framework/core/src/Api/Resource/MailSettingResource.php
SychO9: package has been updated to be case-insensitive 👍🏼

luceos:
There's two things I'd love to be able to do:

- allow custom paths that aren't prefixed with `/api`, eg `/.well-known/webfinger`
- allow a callback to be executed before payload is send to the user, so I can flatten the response or filter some values. I considered a middleware, but I'm not sure this feature could be applied for other purposes than mine (flattening the json payload by pushing `attributes` one level up). And yes this is a hack, but extensibility is about flexibility too.

Overall using this is very nice, it reminds me a lot of the filament resources logic. I have had no issue declaring the resources, their endpoints and their schema. So 👍

SychO9:
> allow custom paths that aren't prefixed with /api, eg /.well-known/webfinger

That would mean adding a route to the forum rather than the API, it's quite messy and would require quite a lot of changes.

This is an edge case where I would rather you use a normal controller. Does this endpoint require the api resourceat all? if it returns custom json structure that isn't actually of the JSON:API spec for resources, all the more reason to use a normal controller instead.

If it does return valid json for a resource, then checkout the `ShowForumController` controller and how you can call an api resource's endpoint.

> allow a callback to be executed before payload is send to the user, so I can flatten the response or filter some values.

You can use the `response` method on the endpoint. Though again, if you are returning non json api spec json, what you might be needing to do instead is using custom controllers.

extensions/approval/extend.php
```diff
@@ -36,17 +37,11 @@
         ->default('is_approved', true)
         ->cast('is_approved', 'bool'),
 
-    (new Extend\ApiSerializer(BasicDiscussionSerializer::class))
-        ->attribute('isApproved', function (BasicDiscussionSerializer $serializer, Discussion $discussion): bool {
-            return $discussion->is_approved;
-        }),
+    (new Extend\ApiResource(Resource\DiscussionResource::class))
+        ->fields(DiscussionResourceFields::class),
```
askvortsov1: Could these be inlined? If we don't need to resolve anything in the container, maybe these could be first class functions? Or maybe this API could accept a list directly? I feel like this would result in easier readability, and less boilerplate for creaitng classes.

extensions/flags/extend.php
```diff
@@ -51,27 +41,26 @@
...
 
-    (new Extend\ApiController(ShowDiscussionController::class))
-        ->addInclude(['posts.flags', 'posts.flags.user']),
+    (new Extend\ApiResource(Resource\ForumResource::class))
+        ->fields(ForumResourceFields::class),
 
-    (new Extend\ApiController(ListPostsController::class))
-        ->addInclude(['flags', 'flags.user']),
+    (new Extend\ApiResource(Resource\DiscussionResource::class))
+        ->endpoint(Endpoint\Show::class, function (Endpoint\Show $endpoint) {
```
askvortsov1, line 56: Maybe in the docs it would be nice to have some guidelines for when to use default includes vs not.

extensions/flags/src/Api/ForumResourceFields.php
```diff
@@ -0,0 +1,38 @@
...
+
+use Flarum\Api\Context;
+use Flarum\Api\Schema;
+use Flarum\Flags\Flag;
+use Flarum\Settings\SettingsRepositoryInterface;
+
+class ForumResourceFields
+{
+    public function __construct(
+        protected SettingsRepositoryInterface $settings
```
askvortsov1: Doesn't look like we need settings here

extensions/flags/src/Api/Resource/FlagResource.php
```diff
@@ -0,0 +1,165 @@
...
+    {
+        return 'flags';
+    }
+
+    public function model(): string
+    {
+        return Flag::class;
+    }
+
+    public function query(Context $context): object
```
askvortsov1, line 53: We need to override this to add the group by?

extensions/subscriptions/src/Api/UserResourceFields.php
```diff
@@ -0,0 +1,37 @@
...
+use Flarum\Api\Schema;
+use Flarum\Discussion\Discussion;
+
+class UserResourceFields
+{
+    public function __invoke(): array
+    {
+        return [
+            Schema\Str::make('subscription')
+                ->writable(fn (Discussion $discussion, Context $context) => $context->updating() && ! $context->getActor()->isGuest())
```
askvortsov1: Should any fields be writable by default for guests? I suppose most checks will happen at the endpoints level, but still. Might be a performance overhead though, so not sure that something more "explicit" is worth it.

extensions/suspend/src/Api/UserResourceFields.php
```diff
@@ -0,0 +1,35 @@
...
+{
+    public function __invoke(): array
+    {
+        return [
+            Schema\Boolean::make('canSuspend')
+                ->get($canSuspend = fn (User $user, Context $context) => $context->getActor()->can('suspend', $user)),
+            Schema\Str::make('suspendReason')
+                ->writable($canSuspend)
+                ->visible($canSuspend),
+            Schema\Str::make('suspendMessage')
```
askvortsov1, line 26: Can we require for `suspendedUntil` that reason and message are also provided?

extensions/tags/src/Api/DiscussionResourceFields.php
```diff
@@ -0,0 +1,114 @@
...
+        protected SettingsRepositoryInterface $settings,
+        protected Factory $validator,
+        protected TranslatorInterface $translator
+    ) {
+    }
+
+    public function __invoke(): array
+    {
+        return [
+            Schema\Boolean::make('canTag')
```
askvortsov1, line 35: I wonder if it would make sense to have a `Schema\Permission`

extensions/tags/src/Api/Resource/TagResource.php
```diff
@@ -0,0 +1,167 @@
...
+    public function model(): string
+    {
+        return Tag::class;
+    }
+
+    public function scope(Builder $query, Context $context): void
+    {
+        $query->whereVisibleTo($context->getActor());
+
+        if ($context->listing(self::class) || $context->showing(self::class)) {
```
askvortsov1, line 49: Is scoping the right place for this? Feels like it should go into some hook.

extensions/tags/src/Api/Resource/TagResource.php
```diff
@@ -0,0 +1,167 @@
...
+            $query->withStateFor($context->getActor());
+        }
+    }
+
+    public function find(string $id, Context $context): ?object
```
askvortsov1, line 54: Is this something that any resource with a slug manager will have to implement?

framework/core/js/src/common/states/PaginatedListState.ts
```diff
@@ -121,6 +121,10 @@ export default abstract class PaginatedListState<T extends Model, P extends Pagi
       include,
     };
 
+    if (!params.include) {
+      delete params.include;
```
askvortsov1, line 125: Why is this necessary? If a falsy value is provided, should we just ignore it on the server?

framework/core/locale/validation.yml
```diff
@@ -1,107 +1,139 @@
...
+  array: "The :attribute field must be an array."
+  ascii: "The :attribute field must only contain single-byte alphanumeric characters and symbols."
+  before: "The :attribute field must be a date before :date."
+  before_or_equal: "The :attribute field must be a date before or equal to :date."
   between:
-    numeric: "The :attribute must be between :min and :max."
-    file: "The :attribute must be between :min and :max kilobytes."
-    string: "The :attribute must be between :min and :max characters."
-    array: "The :attribute must have between :min and :max items."
+    array: "The :attribute field must have between :min and :max items."
```
askvortsov1:
```suggestion
    array: "The :attribute field must contain between :min and :max items."
```

askvortsov1 (commented):
Overall this looks really nice! I've looked through all the extension changes, will try to do core next time.

extensions/flags/extend.php
SychO9: docs actually recommend avoiding default includes whenever possible. Otherwise endpoints become unnecessarily bloated which easily harms performance. I have noticed for example some extensions adding default includes to endpoint A in order to display a relation in frontend page X, only for frontend page B to now unnecessarily have an api call that includes a relation it does not need (which often leads to more issues when you consider lack of eager loading).

With multiple extensions, it all builds up to a pretty bad effect.

extensions/tags/src/Api/DiscussionResourceFields.php
SychO9: Why not, the best part of this library is how much we can add on top that adds to the developer experience.

extensions/tags/src/Api/Resource/TagResource.php
SychO9: yup

extensions/approval/extend.php
SychO9: changed to 

```php
    (new Extend\ApiResource(Resource\DiscussionResource::class))
        ->fields(fn () => [
            Schema\Boolean::make('isApproved'),
        ])
```

extensions/flags/src/Api/Resource/FlagResource.php
SychO9: yes, would be better to have something like this wouldn't it

```php
Endpoint\Index::make()
    ->constrain(fn ($query) => $query->groupBy('post_id')),
```

extensions/subscriptions/src/Api/UserResourceFields.php
SychO9: > Should any fields be writable by default for guests?

the checking here is actually unnecessary since the user update endpoint is for registered users only. But yes if an endpoint is open to guests but a specific field shouldn't be then writability has to be scoped.

there probably won't be many instances of guest write access to an endpoint. Like the User::Create is a unique case for registration.

extensions/suspend/src/Api/UserResourceFields.php
SychO9: theoretically, yes `->requiredWith` `->requiredOnUpdateWith`

though that is not the behavior prior to this PR so out of scope (main goal is to avoid regressions) https://github.com/flarum/framework/blob/2.x/extensions/suspend/src/Listener/SaveSuspensionToDatabase.php

extensions/tags/src/Api/Resource/TagResource.php
SychO9: this is actually the perfect spot, scope is used by both the listing query, and the relationship eager loading query, which is what we want. It's also the author's recommended place for this.

framework/core/js/src/common/states/PaginatedListState.ts
SychO9: good catch, `include=` should not be ignored, but this is meant to prevent the case where no import is provided in the fontend code, since that would still send an empty `include` to the server and the server interprets it as no includes. when in reality the frontend just didn't have any developer-specified incudes which isn't the same as empty includes

changed to
```js
    if (typeof params.include === 'undefined') {
      delete params.include;
    }
```

extensions/flags/extend.php
askvortsov1: Yeah this makes sense. I meant more like "default includes are almost always bad, here's why, here's what might qualify as an exception but seriously please don't"

extensions/tags/src/Api/Resource/TagResource.php
askvortsov1: Why couldn't it be part of the generic find? Is the concern that some models might have non-numerical IDs?

framework/core/src/Api/ApiServiceProvider.php
```diff
@@ -130,27 +155,47 @@ public function register(): void
...
+             * None of the injected dependencies should be directly used within
+             *   the `endpoints` method. Encourage using callbacks.
+             *
+             * @var array<Endpoint&EndpointInterface> $endpoints
+             */
+            $endpoints = $resource->resolveEndpoints(true);
+
+            foreach ($endpoints as $endpoint) {
+                $method = $endpoint->method;
+                $path = rtrim("/$type$endpoint->path", '/');
```
askvortsov1, line 208: Should there be a slash between type and endpoint?

framework/core/src/Api/Context.php
```diff
@@ -0,0 +1,99 @@
...
+{
+    protected ?SearchResults $search = null;
+
+    /**
+     * Data passed internally when reusing resource endpoint logic.
+     */
+    protected array $internal = [];
+
+    /**
+     * Parameters mutated on the current instance.
```
askvortsov1, line 27: The current instance of what?

framework/core/src/Api/Context.php
```diff
@@ -0,0 +1,99 @@
...
+    {
+        return $this->endpoint instanceof Endpoint\Create && (! $resource || is_a($this->collection, $resource));
+    }
+
+    public function updating(string|null $resource = null): bool
+    {
+        return $this->endpoint instanceof Endpoint\Update && (! $resource || is_a($this->collection, $resource));
+    }
+
+    public function deleting(string|null $resource = null): bool
```
askvortsov1, line 85: Shouldn't deleting/updating/etc require that the instance exists and isn't a collection?

framework/core/src/Api/Endpoint/Concerns/HasAuthorization.php
```diff
@@ -0,0 +1,91 @@
...
+
+trait HasAuthorization
+{
+    protected bool|Closure $authenticated = false;
+
+    protected null|string|Closure $ability = null;
+
+    protected bool $admin = false;
+
+    public function authenticated(bool|Closure $condition = true): self
```
askvortsov1, line 26: Nit: should these be prefixed with set? The types should be a good guardrail so I don't think we absolutely need it.

framework/core/src/Api/Endpoint/EndpointInterface.php
```diff
@@ -0,0 +1,18 @@
...
+ * This file is part of Flarum.
+ *
+ * For detailed copyright and license information, please view the
+ * LICENSE file that was distributed with this source code.
+ */
+
+namespace Flarum\Api\Endpoint;
+
+/**
+ * @mixin \Tobyz\JsonApiServer\Endpoint\Endpoint
```
askvortsov1, line 13: Is this doc type sufficient for editors / php runtime?

framework/core/src/Api/JsonApi.php
```diff
@@ -0,0 +1,165 @@
...
+    {
+        $this->endpointName = $endpointName;
+
+        return $this;
+    }
+
+    protected function makeContext(Request $request): Context
+    {
+        if (! $this->endpointName || ! $this->resourceClass || ! class_exists($this->resourceClass)) {
+            throw new BadRequestException('No resource or endpoint specified');
```
askvortsov1, line 50: Should this be split into separate errors, so it's clearer which of the pieces went wrong?

framework/core/src/Api/JsonApi.php
```diff
@@ -0,0 +1,165 @@
...
+
+        return (new Context($this, $request))
+            ->withCollection($collection)
+            ->withEndpoint($this->findEndpoint($collection));
+    }
+
+    protected function findEndpoint(?Collection $collection): Endpoint&EndpointInterface
+    {
+        /** @var Endpoint&EndpointInterface $endpoint */
+        foreach ($collection->resolveEndpoints() as $endpoint) {
```
askvortsov1, line 63: Should we cache the resolved endpoints?

framework/core/src/Api/JsonApi.php
```diff
@@ -0,0 +1,165 @@
...
+
+        $request = $request
+            ->withParsedBody([
+                ...$body,
+                'data' => [
+                    ...($request->getParsedBody()['data'] ?? []),
+                    ...($body['data'] ?? []),
+                    'type' => $resource instanceof Resource
+                        ? $resource->type()
+                        : $resource->name(),
```
askvortsov1, line 104: Is the alternative a Database resource?

framework/core/src/Api/JsonApi.php
```diff
@@ -0,0 +1,165 @@
...
+        );
+
+        return $context->endpoint->process($context);
+    }
+
+    public function validateQueryParameters(Request $request): void
+    {
+        foreach ($request->getQueryParams() as $key => $value) {
+            if (
+                ! preg_match('/[^a-z]/', $key) &&
```
askvortsov1, line 128: If it doesn't contain any non-lowercase characters, we assert that it has to be one of the filter/search parameters? This is to enforce the OpenAPI spec? If so, can the comment say this more explicitly? "Invalid query parameter" doesn't say why it's not allowed.

I feel like this could be more explicit in which patterns it does allow.

framework/core/src/Api/Resource/AbstractDatabaseResource.php
```diff
@@ -0,0 +1,205 @@
...
+
+    abstract public function model(): string;
+
+    /** @inheritDoc */
+    public function newModel(Context $context): object
+    {
+        return new ($this->model());
+    }
+
+    public function resource(object $model, Context $context): ?string
```
askvortsov1, line 44: Why not resourceType? We're not instantiating a new Resource object

framework/core/src/Api/Resource/AbstractDatabaseResource.php
```diff
@@ -0,0 +1,205 @@
...
+        $model = parent::updateAction($model, $context);
+
+        $this->dispatchEventsFor($model, $context->getActor());
+
+        return $model;
+    }
+
+    public function deleteAction(object $model, Context $context): void
+    {
+        $this->deleting($model, $context);
```
askvortsov1, line 80: Should these not have something in between them?

framework/core/src/Api/Resource/AbstractDatabaseResource.php
```diff
@@ -0,0 +1,205 @@
...
+     * @param FlarumContext $context
+     */
+    public function deleted(object $model, Context $context): void
+    {
+        //
+    }
+
+    public function dispatchEventsFor(mixed $entity, User $actor = null): void
+    {
+        if (method_exists($entity, 'releaseEvents')) {
```
askvortsov1, line 169: This method comes from eloquent? If so, can we add a comment with a link to docs?

framework/core/src/Api/Resource/AbstractDatabaseResource.php
```diff
@@ -0,0 +1,205 @@
...
+            $this->traitDispatchEventsFor($entity, $actor);
+        }
+    }
+
+    /**
+     * @param FlarumContext $context
+     */
+    public function mutateDataBeforeValidation(Context $context, array $data): array
```
askvortsov1, line 177: I assume this is intended to be overridden. Is there a phpdoc to annotate this?

framework/core/src/Api/Resource/AccessTokenResource.php
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    public function scope(Builder $query, \Tobyz\JsonApiServer\Context $context): void
+    {
+        $query->whereVisibleTo($context->getActor());
+    }
+
+    public function newModel(\Tobyz\JsonApiServer\Context $context): object
+    {
+        if ($context->creating(self::class)) {
```
askvortsov1, line 54: I'm not sure myself, but I wonder if `isCreateEndpoint` might be clearer than `creating`

framework/core/src/Api/Resource/AccessTokenResource.php
```diff
@@ -0,0 +1,143 @@
...
+            $token = DeveloperAccessToken::make($context->getActor()->id);
+            $token->last_activity_at = null;
+
+            return $token;
+        }
+
+        return parent::newModel($context);
```
askvortsov1, line 61: What would this do?

framework/core/src/Api/Resource/AccessTokenResource.php
askvortsov1, line 54: Also, if we don't have an update endpoint, aren't we always creating?

framework/core/src/Api/Resource/Concerns/Extendable.php
```diff
@@ -0,0 +1,93 @@
...
+    public function resolveEndpoints(bool $earlyResolution = false): array
+    {
+        if (! is_null($this->cachedEndpoints) && ! $earlyResolution) {
+            return $this->cachedEndpoints;
+        }
+
+        $endpoints = $this->endpoints();
+
+        foreach (array_reverse(array_merge([static::class], class_parents($this))) as $class) {
+            if (isset(static::$endpointModifiers[$class])) {
```
askvortsov1, line 46: Nit: could this be a `foreach (a ?? [])`

framework/core/src/Api/Resource/DiscussionResource.php
```diff
@@ -0,0 +1,368 @@
...
+            SortColumn::make('commentCount')
+                ->descendingAlias('top'),
+            SortColumn::make('createdAt')
+                ->ascendingAlias('oldest')
+                ->descendingAlias('newest'),
+        ];
+    }
+
+    /** @param Discussion $model */
+    public function creating(object $model, \Tobyz\JsonApiServer\Context $context): ?object
```
askvortsov1, line 278: Why doesn't it get these from the fields? Is it because we need to dispatch a saving before we actually save the model?

framework/core/src/Api/Resource/ForumResource.php
```diff
@@ -0,0 +1,168 @@
...
+            Schema\Str::make('assetsBaseUrl')
+                ->get(fn () => rtrim($this->assetsFilesystem->url(''), '/')),
+            Schema\Str::make('jsChunksBaseUrl')
+                ->get(fn () => $this->assetsFilesystem->url('js')),
+
+            Schema\Str::make('adminUrl')
+                ->visible(fn ($model, Context $context) => $context->getActor()->can('administrate'))
+                ->get(fn () => $this->url->to('admin')->base()),
+            Schema\Str::make('version')
+                ->visible(fn ($model, Context $context) => $context->getActor()->can('administrate'))
```
askvortsov1, line 137: Don't we have an `isAdmin` method somewhere?

framework/core/src/Api/Resource/UserResource.php
```diff
@@ -0,0 +1,456 @@
...
+                ->property('joined_at'),
+            Schema\Integer::make('discussionCount'),
+            Schema\Integer::make('commentCount'),
+            Schema\DateTime::make('lastSeenAt')
+                ->visible(function (User $user, Context $context) {
+                    return $user->getPreference('discloseOnline') || $context->getActor()->can('viewLastSeenAt', $user);
+                }),
+
+            Schema\DateTime::make('markedAllAsReadAt')
+                ->visible(fn (User $user, Context $context) => ($context->collection instanceof self || $context->collection instanceof ForumResource) && $context->getActor()->id === $user->id)
```
askvortsov1, line 256: Can we factor this logic out into a helper function?

framework/core/src/Api/Sort/SortColumn.php
```diff
@@ -0,0 +1,48 @@
...
+use Tobyz\JsonApiServer\Laravel\Sort\SortColumn as BaseSortColumn;
+
+class SortColumn extends BaseSortColumn
+{
+    protected array $alias = [
+        'asc' => null,
+        'desc' => null,
+    ];
+
+    public function ascendingAlias(?string $alias): static
```
askvortsov1, line 21: Where do we use aliases for this?

framework/core/src/Http/Middleware/PopulateWithActor.php
```diff
@@ -0,0 +1,29 @@
...
+use Flarum\Discussion\Discussion;
+use Flarum\Http\RequestUtil;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\MiddlewareInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+
+class PopulateWithActor implements MiddlewareInterface
+{
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
```
askvortsov1, line 21: This (and iirc translation locales) are one of the few global bits of per-request state that make using octane difficult. Maybe in another pr, it would make sense to store all them in some centralized place so it's easier to keep track of and eventually eliminate?

framework/core/src/Http/RequestUtil.php
```diff
@@ -32,4 +33,142 @@ public static function withActor(Request $request, User $actor): Request
...
+
+    public static function extractOffsetFromNumber(Request $request, int $limit): int
+    {
+        $page = (int) ($request->getQueryParams()['page']['number'] ?? 1);
+
+        if ($page < 1) {
+            throw new BadRequestException('page[number] must be at least 1');
+        }
+
+        return ($page - 1) * $limit;
```
askvortsov1, line 125: Is the page zero-indexed or one-indexed?

askvortsov1 (commented):
Amazing job, this new API is such a great transformation!

Left some more comments. Some of these are clarifications / stuff I've gotten a bit rusty one, and some more might be better noted and saved for a future feature.

Thanks for putting this together, it's really exciting to see!

extensions/tags/src/Api/Resource/TagResource.php
SychO9: I don't really follow. what generic find? The eloquent find? I don't like the idea of overriding the eloquent find. Better practice is to create a custom `findByX` method. But that should be entirely up to the extension dev. Either way they will have to override the API resource find, as the slug finding isn't a global process but only needed in endpoints.

framework/core/src/Api/ApiServiceProvider.php
SychO9: The endpoint path is always slashed, so nope.

framework/core/src/Api/Context.php
SychO9: This context.

framework/core/src/Api/Context.php
SychO9: no, at this stage the model definitely exists, that's checked a lot earlier. Only purpose here is to check we are on a Delete endpoint of resource X

framework/core/src/Api/Endpoint/Concerns/HasAuthorization.php
SychO9: I don't believe it necessary.

framework/core/src/Api/Endpoint/EndpointInterface.php
SychO9: Yes, we'll probably drop this anyway when we refactor to rely on the original package instead of the fork.

framework/core/src/Api/JsonApi.php
SychO9: This is only needed when problematically using an endpoint. Where you have to call `->forResource(...)->forEndpoint(...)` so it should be pretty clear what you missed providing based on just this.

framework/core/src/Api/JsonApi.php
SychO9: already cached inside `resolveEndpoints`

framework/core/src/Api/JsonApi.php
SychO9: No, regardless of Database or not, $resource is either a `Resource` or a `Collection` of resources.

framework/core/src/Api/Resource/AbstractDatabaseResource.php
SychO9: The interface method of the package is `resource`

framework/core/src/Api/Resource/AbstractDatabaseResource.php
SychO9: like what?

framework/core/src/Api/Resource/AbstractDatabaseResource.php
SychO9: no, it's internal flarum

framework/core/src/Api/Resource/AccessTokenResource.php
SychO9: Unnecessarily verbose imo.

framework/core/src/Api/Resource/AccessTokenResource.php
SychO9: fallback to the default. While `newModel` is only called in the Create endpoint right now. It doesn't mean that will always be the case.

framework/core/src/Api/Resource/AccessTokenResource.php
SychO9: Best not to make such assumptions in an extendable env like Flarum. An extension can easily call newModel in its custom endpoint.

framework/core/src/Api/Resource/DiscussionResource.php
SychO9: Sorry I don't follow. What are you referring to that it can't get?

framework/core/src/Api/Resource/ForumResource.php
SychO9: That's for endpoints. We can always add more stuff to fields/endpoints in the future

framework/core/src/Api/Sort/SortColumn.php
SychO9: Aliases are used to generate the sortmap, so no more binding discussion sortmap to container for extensibility.

framework/core/src/Http/RequestUtil.php
SychO9: 1 indexed

framework/core/src/Api/Resource/AbstractDatabaseResource.php
askvortsov1: Misinterpreted $this->delete, my bad

framework/core/src/Api/Resource/DiscussionResource.php
askvortsov1: Created at, and the user.

SychO9: Created at and the actor aren't writable fields so they're never used for filling the values. We must fill these manually like we do here.

askvortsov1 (approved):
https://youtu.be/6QCWZJJr3BM?si=naU5m8mDBw0hqxO2

SychO9:
Will merge after the DB PRs are merged.

luceos:
@SychO9 this one could/should be merged asap. Let's work off of this codebase moving forward before any other branches change dramatically.

PR #3975, opened 2024-04-04
feat: low maintenance mode (maintenance with admin access)

SychO9:
**Part of #3976**
**Fixes https://github.com/flarum/issue-archive/issues/255**

**Changes proposed in this pull request:**

**Reviewers should focus on:**

**Screenshot**

image
image
image
image
image

**Required changes:**

- [ ] Related documentation PR: (Remove if irrelevant)
- [ ] Related core extension PRs: (Remove if irrelevant)

PR #3977, opened 2024-04-05
feat: advanced maintenance modes

### LINKED ISSUE #3976: Advanced Maintenance
Currently Flarum has an `offline` mode that completely locks down the forum to everyone.

This can be enabled through the `'offline' => true` key in `config.php`

**Goals**

* A lower maintenance mode that allows only admins to access the forum & therefore the admin section. (https://github.com/flarum/issue-archive/issues/255)
* A maintenance mode that disables all extensions with admin only access. (https://github.com/flarum/framework/pull/3073)

---

**Plan**

There are 3 maintenance modes needed in total:
* High maintenance mode: complete lock down.
* Low maintenance mode: allowed admin access.
* Safe mode: Low maintenance mode + all extensions disabled.

So we can allow an integer to be passed to the `offline` configuration key with the following possible values:

* `0`: no maintenance mode.
* `1`: high maintenance mode (complete lock down).
* `2`: low maintenance mode (admin access/admin login allowed)
* `3`: safe mode: low maintenance mode + no extensions booting.

We do not have to break backwards compatibility, so a `'offline' => true` can be interpreted as high maintenance mode (`1`).

**Why include safe mode into low maintenance mode?**

Preventing all extensions from booting should be limited to admin access only in case of unintended actions by users or un-scoped visibility of information.

That said, we could rename `offline` to `maintenance` and retain BC for `offline=true` meaning `maintenance=1`

**Admin Login**

For low maintenance, we need to allow admin login through a basic NoJs auth form. which means excluding a couple of routes from maintenance mode.

**Alert**

Must clearly display that the forum is in low maintenance mode.

**Potentially**

* Should this be configured from the admin settings? it would be doable for modes `2` & `3` but not `1`.

---

**Pull Requests**

This will be tackled in two PRs:
1. First will be the low maintenance mode (https://github.com/flarum/framework/pull/3975)
2. Then the safe mode.

SychO9:
**Closes #3976**

**Separate PRs**
* #3975 
* https://github.com/flarum/framework/pull/3978


framework/core/js/src/admin/components/AdminPage.tsx
```diff
@@ -137,9 +141,34 @@ export default abstract class AdminPage<CustomAttrs extends IPageAttrs = IPageAt
...
     }
 
-    const { setting, ...attrs } = entry;
+    const { setting, json, refreshAfterSaving, ...attrs } = entry;
 
-    return <FormGroup bidi={this.setting(setting)} {...attrs} />;
+    const originalBidi: (value?: string) => any = this.setting(setting);
+    let bidi: (value?: string) => any;
+
+    if (json) {
```
askvortsov1, line 149: Why do this on the frontend and not the server?

framework/core/js/src/admin/components/AdvancedPage.tsx
```diff
@@ -96,23 +96,49 @@ export default class AdvancedPage<CustomAttrs extends IPageAttrs = IPageAttrs> e
...
         <Form>
           {this.buildSettingComponent({
             type: 'select',
             help: app.translator.trans('core.admin.advanced.maintenance.help'),
             setting: 'maintenance_mode',
+            refreshAfterSaving: true,
             options: {
-              0: app.translator.trans('core.admin.advanced.maintenance.options.0'),
-              1: {
+              [MaintenanceMode.NO_MAINTENANCE]: app.translator.trans('core.admin.advanced.maintenance.options.0'),
```
askvortsov1: Could we use human readable names for the translation keys?

framework/core/js/src/admin/components/AdvancedPage.tsx
```diff
@@ -96,23 +96,49 @@ export default class AdvancedPage<CustomAttrs extends IPageAttrs = IPageAttrs> e
...
+              [MaintenanceMode.HIGH_MAINTENANCE]: {
                 label: app.translator.trans('core.admin.advanced.maintenance.options.1'),
                 disabled: true,
               },
-              2: app.translator.trans('core.admin.advanced.maintenance.options.2'),
+              [MaintenanceMode.LOW_MAINTENANCE]: app.translator.trans('core.admin.advanced.maintenance.options.2'),
+              [MaintenanceMode.SAFE_MODE]: app.translator.trans('core.admin.advanced.maintenance.options.3'),
             },
             default: 0,
```
askvortsov1: Can this default use the constants?

framework/core/js/src/admin/components/ExtensionPage.tsx
```diff
@@ -61,17 +63,31 @@ export default class ExtensionPage<Attrs extends ExtensionPageAttrs = ExtensionP
...
         ) : (
-          <div className="ExtensionPage-body">{this.sections(vnode).toArray()}</div>
+          this.body(vnode)
         )}
       </div>
     );
   }
 
+  body(vnode: Mithril.VnodeDOM<Attrs, this>) {
+    return !this.isEnabled() ? (
```
askvortsov1: why not put the true case first?

framework/core/locale/core.yml
```diff
@@ -700,13 +704,18 @@ core:
...
+      placeholder: Select multiple options
+
     # These translations are used in the navigation header.
     nav:
       drawer_button: Open Navigation Drawer
 
     # These translations are used in forum & admin notices.
     notices:
-      maintenance_mode_2: Down for maintenance. Only administrators may access the forum.
+      maintenance_mode_2: Down for maintenance. Only administrators can access the forum.
```
askvortsov1: similar here, why not use the human readable names instead of the numbers? Then we'd only have to sync numbers in the backend + frontend enums.

framework/core/src/Admin/AdminServiceProvider.php
```diff
@@ -104,6 +107,10 @@ public function register(): void
             return $assets;
         });
 
+//        $this->container->afterResolving(AssetManager::class, function (AssetManager $assets) {
```
askvortsov1: Should this be here?

framework/core/src/Extension/ExtensionManager.php
```diff
@@ -302,7 +304,17 @@ public function getEnabledExtensions(): array
      */
     public function extend(Container $container): void
     {
-        foreach ($this->getEnabledExtensions() as $extension) {
+        $extensions = $this->getEnabledExtensions();
+
+        if ($this->maintenance->isSafeMode()) {
```
askvortsov1: Should this run before dependency resolution? Seems possible that this could be used to enable a dependent of something disabled.

askvortsov1 (changes requested):

askvortsov1:
Why not use a string enum for maintenance modes instead of integers?

framework/core/js/src/admin/components/AdminPage.tsx
SychO9: because if we do something on the server about setting types, I would prefer a more solid implementation (I really want to go for something like: https://github.com/spatie/laravel-settings)

Don't think this'll be a 2.x thing though.

framework/core/js/src/admin/components/AdvancedPage.tsx
SychO9: Sorry, you probably reviewed a specific PR commit, I did push a third separate commit for that: https://github.com/flarum/framework/pull/3977/commits/7c2a1319ea1cc106f7085e594d2b10b8dd69fb10

SychO9:
Sorry, you probably reviewed a specific PR commit, I did push a third separate commit for that: https://github.com/flarum/framework/pull/3977/commits/7c2a1319ea1cc106f7085e594d2b10b8dd69fb10

framework/core/src/Extension/ExtensionManager.php
SychO9: nice catch 🤔

SychO9: image
image

askvortsov1: Good feature, but I think some help text / an icon tooltip / something is needed to explain why the disabled buttons are disabled.

askvortsov1 (approved):
LGTM, other than the note about adding help text to disabled extensions because of missing dependencies.

SychO9:
image

PR #3978, opened 2024-04-05
feat: safe mode

SychO9:
**Part of #3976**
**https://github.com/flarum/issue-archive/issues/85**

* Safe mode that prevents extensions from booting while in low maintenance mode.
* Configurable from settings, config.php takes priority however.
* Ability to set a list of extensions to enforce in safe mode (config `safe_mode_extensions` key, falls back to setting with same key).
* Introduces a `MultiSelect` component (needed to select extensions to enforce).
* Adds a `json` field to admin settings definition. Allows making an admin setting `json` (parses json when reading the setting value, and stringifies when writing).
* Adds a `refreshAfterSaving` field to admin settings definition, which refreshes the admin page when said setting is changed.

**Screenshot**
image
image
image
image
image


askvortsov1:
Is the plan here to review the maintenance modes branch into main first, and then this branch?

SychO9:
@askvortsov1 the maintenance modes branch is the final. Will be undrafted after this is merged into it.

Though there is no requirement to review there. Comments may be added wherever.

PR #3979, opened 2024-04-10
fix: change return type of hasOne and belongsTo to nullable object type.

novacuum:
Maybe there is a different sollution than with UnionType, but I found no hints about setting the ObjectType as nullable one.

**Fixes: no issues **

Change return type of hasOne and belongsTo to nullable object type in the RelationProperty Phpstan Reflection Class.

Is this Valid in every case, is there a case where nullable is not true?

**Screenshot**

**Necessity**

**Confirmed**


SychO9 (commented):
@novacuum Thanks for the PR! could you base this on the 2.x branch instead?

PR #3984, opened 2024-05-03
feat: add support for `SQLite`

SychO9:
**Part of the roadmap**

Adds support for installing and using Flarum with SQLite 3.8.8+

While it is trivial to support it from core (apart from a couple exceptions where queries needed to be adapted), whether third party extensions will work with sqlite is a different matter. So I've added a warning about that on the installation page.

**Screenshot**
image
image

**Necessity**

**Confirmed**

**Required changes:**


framework/core/src/Database/Migrator.php
```diff
@@ -31,12 +31,14 @@ public function __construct(
         protected ConnectionInterface $connection,
         protected Filesystem $files
     ) {
-        if (! ($connection instanceof MySqlConnection)) {
-            throw new InvalidArgumentException('Only MySQL connections are supported');
+        $doctrine = $connection->getDoctrineConnection()->getDatabasePlatform();
+
+        if (! Type::hasType('timestamp')) {
```
askvortsov1, line 36: What does this do?

framework/core/src/Install/Steps/ConnectToDatabase.php
```diff
@@ -33,17 +37,27 @@ public function getMessage(): string
...
+    private function mysql(array $config): void
+    {
         $pdo = (new MySqlConnector)->connect($config);
 
         $version = $pdo->query('SELECT VERSION()')->fetchColumn();
 
         if (Str::contains($version, 'MariaDB')) {
-            if (version_compare($version, '10.0.5', '<')) {
+            if (version_compare($version, '10.10.0', '<')) {
                 throw new RangeException('MariaDB version too low. You need at least MariaDB 10.0.5');
```
askvortsov1, line 56: Should this message be adjusted to match the code?

askvortsov1 (changes requested):
I'm still a bit worried that people using sqlite will disregard the warning, and worry why their queries are erroring. Is it possible to catch syntax errors, and in the error message, flag use of sqllite as a possible culprit?

framework/core/src/Database/Migrator.php
SychO9: fixes timestamp columns on sqlite migrations. (fails with an error otherwise)

framework/core/src/Install/Steps/ConnectToDatabase.php
SychO9: done in the pgsql branch

SychO9:
> is it possible to catch syntax errors, and in the error message, flag use of sqllite as a possible culprit?

Done in the pgsql branch

image

jaggy (commented):
Amazing work on the PR!

In regards to extensions, I wonder if we can add some sort of support for `database` support in the composer.json in `extra.flarum-extension` or some config so it's easier to show extension compatibility. 🤔

SychO9:
great idea @jaggy! will add that in the PgSQL PR which builds on top of this branch: https://github.com/flarum/framework/pull/3985

SychO9:
added in https://github.com/flarum/framework/pull/3985#issuecomment-2152822982

askvortsov1 (approved):

PR #3985, opened 2024-05-05
feat: add support for PgSQL

SychO9:
**Part of the roadmap**

Postgres was trickier than SQLite, but that was mostly due to having some code that dates back years ago when eloquent/ the laravel schema builder was less flexible.

In most cases there should be no need to add any custum handling for the database driver when building queries. But that does become a necessity when writing more complex / larger queries. Example: fulltext search and statistics.

The easiest instance extension devs will run into when wanting to support PgSQL is the loose group by.

```php
// Allowed in mysql
->select('col', 'col2', ..)
->groupBy('col')
->orderBy(..)

// must be done this way in pgsql
->select('col', 'col2', ..)
->distinct('col')
->orderBy('col')
->orderBy(..)

// can be simplified to
->select('col', 'col2', ..)
->whenMySql(fn (Builder $query) => $query->groupBy('col'))
->whenPgSql(fn (Builder $query) => $query->distinct('col')->orderBy('col'))
->orderBy(..)
```

**Screenshot**
image
image

**Necessity**

**Confirmed**

**Required changes:**


SychO9:
Added support for defining supported databases

image
image

extensions/flags/src/Api/Controller/ListFlagsController.php
```diff
@@ -51,11 +52,15 @@ protected function data(ServerRequestInterface $request, Document $document): it
...
         $flags = Flag::whereVisibleTo($actor)
-            ->latest('flags.created_at')
-            ->groupBy('post_id')
             ->limit($limit + 1)
             ->offset($offset)
+            ->whenPgSql(
+                fn (Builder $query) => $query->distinct('post_id')->orderBy('post_id'),
+                else: fn (Builder $query) => $query->groupBy('post_id')
+            )
+            ->latest('flags.id')
```
askvortsov1: Why did we chance the "latest" field? I guess the result will be the same, since things are created with ascending IDs, but is the goal to sort more performantly? I worry that we lose some explicitness here, and if we manually change "created at", sorting would be incorrect.

extensions/flags/tests/integration/api/flags/ListTest.php
```diff
@@ -77,12 +78,14 @@ public function admin_can_see_one_flag_per_post()
...
 
-        $this->assertEquals(200, $response->getStatusCode());
+        $body = $response->getBody()->getContents();
 
-        $data = json_decode($response->getBody()->getContents(), true)['data'];
+        $this->assertEquals(200, $response->getStatusCode(), $body);
+
+        $data = json_decode($body, true)['data'];
 
         $ids = Arr::pluck($data, 'id');
```
askvortsov1, line 90: This variable is now unused. Is the problem that sort order is no longer deterministic?

extensions/mentions/tests/integration/api/ListPostsTest.php
```diff
@@ -116,35 +116,35 @@ protected function prepareMentionedByData(): void
     {
         $this->prepareDatabase([
             Discussion::class => [
-                ['id' => 100, 'title' => __CLASS__, 'created_at' => Carbon::now(), 'user_id' => 1, 'first_post_id' => 101, 'comment_count' => 12],
+                ['id' => 100, 'title' => __CLASS__, 'created_at' => Carbon::parse('2024-05-04'), 'user_id' => 1, 'first_post_id' => 101, 'comment_count' => 12],
```
askvortsov1, line 119: Should we extract this variable so we don't need to repeat it everywhere? Maybe even into the central testing library?

extensions/mentions/tests/integration/api/UserMentionsTest.php
```diff
@@ -488,9 +489,11 @@ public function editing_a_post_with_deleted_author_that_has_a_mention_works()
             ])
         );
 
-        $this->assertEquals(200, $response->getStatusCode());
+        $body = $response->getBody()->getContents();
```
askvortsov1, line 504: Maybe we should make a helper function for this/

extensions/statistics/src/Api/Controller/ShowStatisticsData.php
```diff
@@ -130,11 +130,18 @@ private function getTimedCounts(Builder $query, string $column, ?DateTime $start
...
         // if within the last 24 hours, group by hour
-        $format = 'CASE WHEN '.$column.' > ? THEN \'%Y-%m-%d %H:00:00\' ELSE \'%Y-%m-%d\' END';
+        $format = "CASE WHEN $column > ? THEN '$formats[0]' ELSE '$formats[1]' END";
+
         $dbFormattedDatetime = match ($query->getConnection()->getDriverName()) {
-            'sqlite' => 'strftime('.$format.', '.$column.')',
-            default => 'DATE_FORMAT('.$column.', '.$format.')',
+            'sqlite' => "strftime($format, $column)",
+            'pgsql' => "TO_CHAR($column, $format)",
+            default => "DATE_FORMAT($column, $format)",
```
askvortsov1: Why not be explicit about this being `mysql`, and have default raise?

extensions/sticky/src/PinStickiedDiscussionsToTop.php
```diff
@@ -58,7 +58,7 @@ public function __invoke(DatabaseSearchState $state, SearchCriteria $criteria):
                 // Add the bindings manually (rather than as the second
                 // argument in orderByRaw) for now due to a bug in Laravel which
                 // would add the bindings in the wrong order.
-                $q->selectRaw('(is_sticky and not exists ('.$read->toSql().') and last_posted_at > ?) as is_unread_sticky', array_merge($read->getBindings(), [$state->getActor()->marked_all_as_read_at ?: 0]));
+                $q->selectRaw('(is_sticky and not exists ('.$read->toSql().') and last_posted_at > ?) as is_unread_sticky', array_merge($read->getBindings(), [$state->getActor()->marked_all_as_read_at ?: '1970-01-01 00:00:00']));
```
askvortsov1: Is there no constant somewhere for epoch time?

framework/core/js/src/admin/AdminApplication.tsx
```diff
@@ -32,6 +32,7 @@ export type Extension = {
   extra: {
     'flarum-extension': {
       title: string;
+      'database-support': undefined | string[];
```
askvortsov1, line 35: Should this be a union type instead of a generic string?

framework/core/locale/core.yml
```diff
@@ -661,6 +661,7 @@ core:
     # These translations are displayed as error messages.
     error:
       circular_dependencies_message: "Circular dependencies detected: {extensions}. Aborting. Please disable one of the extensions and try again."
+      db_error_message: "An error occurred while communicating with the database. This could mean an incompatibility with your database driver."
```
askvortsov1:
```suggestion
      db_error_message: "Database query failed. This may be caused by an incompatibility between an extension and your database driver."
```

framework/core/js/src/common/Application.tsx
```diff
@@ -550,7 +550,11 @@ export default class Application {
         break;
 
       default:
-        if (this.requestWasCrossOrigin(error)) {
+        const code = error.response?.errors?.[0]?.code;
+
+        if (code === 'db_error' && app.session.user?.isAdmin()) {
```
askvortsov1, line 564: Should we print the current db driver?

framework/core/migrations/2018_01_11_093900_change_access_tokens_columns.php
```diff
@@ -26,11 +26,22 @@
             $table->integer('user_id')->unsigned()->change();
         });
 
-        // Use a separate schema instance because this column gets renamed
-        // in the previous one.
-        $schema->table('access_tokens', function (Blueprint $table) {
-            $table->dateTime('last_activity_at')->change();
-        });
+        if ($schema->getConnection()->getDriverName() !== 'pgsql') {
```
askvortsov1: Can we put the pgsql case first, avoiding `if a != b`?

framework/core/src/Discussion/Search/FulltextFilter.php
```diff
@@ -84,9 +103,69 @@ public function search(SearchState $state, string $value): void
...
     }
+
+    protected function pgsql(SearchState $state, string $value): void
+    {
+        /** @var Builder $query */
+        $query = $state->getQuery();
+
+        $grammar = $query->getGrammar();
+
+        $matchCondition = "to_tsvector('english', ".$grammar->wrap('posts.content').") @@ plainto_tsquery('english', ?)";
```
askvortsov1: Does pgsql support language search for other languages? If so, should we make this configurable in settings?

framework/core/src/Install/Console/UserDataProvider.php
```diff
@@ -42,20 +42,31 @@ public function configure(Installation $installation): Installation
 
     private function getDatabaseConfiguration(): DatabaseConfig
     {
-        $host = $this->ask('Database host (required):');
-        $port = 3306;
+        $driver = $this->ask('Database driver (mysql, sqlite, pgsql) (Default: mysql):', 'mysql');
+        $port = match ($driver) {
```
askvortsov1, line 46: Do we surface these default ports in the UI?

framework/core/src/Install/Steps/EnableBundledExtensions.php
```diff
@@ -24,22 +24,7 @@
...
-        'flarum-lock',
-        'flarum-markdown',
-        'flarum-mentions',
-        'flarum-statistics',
-        'flarum-sticky',
-        'flarum-subscriptions',
-        'flarum-suspend',
-        'flarum-tags',
-    ];
+    public const EXTENSION_WHITELIST = [];
```
askvortsov1: Unrelated, but maybe we should rename this to `EXTENSION_ALLOWLIST` or `DEFAULT_ENABLED_EXTENSIONS`

askvortsov1 (changes requested):
Awesome work! Should we show a warning in the extension UI and/or when enabling extensions if they don't specify compatibility with the current db driver?

extensions/flags/src/Api/Controller/ListFlagsController.php
SychO9: I do not recall, but I think this wasn't supposed to make it to the commit (was battling with tests)

reverting

extensions/flags/tests/integration/api/flags/ListTest.php
SychO9: When grouping by, PgSQL acts differently than sqlite and mysql. PgSQL picks the first element of a group, while the others pick the last. That shouldn't be a problem since all we need is that it picks one flag only from each post regardless of which it is.

I've changed this to:

```php
        if ($this->database() instanceof PostgresConnection) {
            $this->assertEqualsCanonicalizing(['3', '4', '5'], $ids);
        } else {
            $this->assertEqualsCanonicalizing(['1', '4', '5'], $ids);
        }
```

extensions/mentions/tests/integration/api/ListPostsTest.php
SychO9: I think we can probably just do something in the post factory, though not a priority right now tbh.

extensions/mentions/tests/integration/api/UserMentionsTest.php
SychO9: I actually plan to make a separate PR that improves the testing experience, rather than a helper function I'm thinking of creating a `TestReponse` that is essentially a wrapper around the response, and allows similar laravel useful methods like

```php
$response->assertOk()
$response->assertStatus(201);
$response->body()
...
```

It would make a huge difference from my experience writing tests with Laravel VS Flarum.

extensions/sticky/src/PinStickiedDiscussionsToTop.php
SychO9: changed to:

```php
$epochTime = (new DateTime('@0'))->format('Y-m-d H:i:s');
```

framework/core/js/src/admin/AdminApplication.tsx
SychO9: We can't really ensure people will actually use the exact strings mysql, pgsql, sqlite

It's very likely someone will write something different, cause that's just how user input is.

framework/core/js/src/common/Application.tsx
SychO9: The admin is able to see their db driver in the admin page so I don't think it's necessary.

framework/core/src/Install/Console/UserDataProvider.php
SychO9: The UI allows passing a port into the host by adding `:3306` for example. Otherwise no, there is no port input with the default port displayed.

SychO9:
> Awesome work! Should we show a warning in the extension UI and/or when enabling extensions if they don't specify compatibility with the current db driver?

image

SychO9:
@askvortsov1 

https://github.com/flarum/framework/pull/3985/commits/ff96395941ea36e40672ff484bd79f8b8bf12a06

image

framework/core/src/Discussion/Search/FulltextFilter.php
SychO9: see comment below

framework/core/js/src/admin/AdminApplication.tsx
askvortsov1: Fair enough, we could try to validate, but it's such a low priority, I'm not sure it's worth it.

askvortsov1 (approved):

PR #3986, opened 2024-05-10
chore: drop the need for a `json-api-server` fork

SychO9:
**Part of #3964**

With the recent changes to the json-api-server package made by Toby, it is easier to not rely on the package, though we still have to apply many custom changes in core itself, it is better than maintaining a fork and the custom changes are likely to decrease as the package becomes even more flexible in time.

Also the fork technicaly sill exists, but is 1:1 to the original package, until the original package is tagged with the latest changes Or we might always keep it manually 1:1 synced in case of any urgent changes.

**Screenshot**

**Necessity**

**Confirmed**

**Required changes:**


luceos:
If we can drop our own fork in between I'd highly recommend doing that for now. We can see if that remains a valid strategy up to the beta builds of 2.0.

PS I have implemented this package on flarum.org now too and it seems to be very reliable and stable so far 👏

PR #3992, opened 2024-05-18
fix: setting key `safe_mode_extensions` not exists

datlechin:
Fix setting key `safe_mode_extensions` not exists

**Fixes #0000**

**Screenshot**

**QA**

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/framework/core/src/Admin/WhenSavingSettings.php
+++ b/framework/core/src/Admin/WhenSavingSettings.php
@@ -36,14 +36,18 @@ class WhenSavingSettings
     {
         if (array_key_exists('safe_mode_extensions', $event->settings)) {
             $safeModeExtensions = json_decode($event->settings['safe_mode_extensions'] ?? '[]', true);
+            $sorted = [];
 
-            $extensions = $this->extensions->getExtensions()->filter(function ($extension) use ($safeModeExtensions) {
-                return in_array($extension->getId(), $safeModeExtensions);
-            });
+            if ($safeModeExtensions) {
+                $extensions = $this->extensions->getExtensions()->filter(function ($extension) use ($safeModeExtensions) {
+                    return in_array($extension->getId(), $safeModeExtensions);
+                });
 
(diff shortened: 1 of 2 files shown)


framework/core/src/Admin/WhenSavingSettings.php
```diff
@@ -36,14 +36,18 @@ public function beforeSave(Saving $event): void
...
+                });
 
-            $sorted = array_map(fn (Extension $e) => $e->getId(), $this->extensions->sortDependencies($extensions->all()));
+                $sorted = array_map(fn (Extension $e) => $e->getId(), $this->extensions->sortDependencies($extensions->all()));
+                $sorted = array_values($sorted);
+            }
 
-            $event->settings['safe_mode_extensions'] = json_encode(array_values($sorted));
+            $event->settings['safe_mode_extensions'] = json_encode($sorted);
         }
```
SychO9, line 51: does this part throw an error like the config one? I would assume it wouldn't since we do `$event->settings['safe_mode_extensions'] ?? '[]'` here 🤔

SychO9 (commented):
Thanks for the PR!

framework/core/src/Admin/WhenSavingSettings.php
datlechin: `safe_mode_extensions` always includes in Saving event with empty value (`safe_mode_extensions
: 
""` in the Advanced page, so it doesn't cause errors here

datlechin: It may looks like this

```php
// $event->settings['safe_mode_extensions'] is '' if no extensions selected
$safeModeExtensions = ! empty($event->settings['safe_mode_extensions']) ? $event->settings['safe_mode_extensions'] : '[]';
$safeModeExtensions = json_decode($event->settings['safe_mode_extensions'], true);
```

SychO9: Ah that makes sense!

SychO9 (approved):

PR #3994, opened 2024-05-28
feat: add option for filtering read stickied on all discussions page

adrianmccay:
Firstly, apologies if you are not taking PR's for 1.x or if this is just not something that would be considered. I can create a third party extension if that is a better option.

As per discussion on Flarum Discuss - https://discuss.flarum.org/d/34607-flarum-do-perform-some-automated-update
For some, the stickied posts dropping down the order on "All Discusisons" like a non-stickied post when the discussion has been read is not an expected behaviour. I have added a simple on/off option which defaults to the existing behaviour. With the switch on, the read discussions on "All Discussions" are not treated as stickied. With the switch off, all stickied posts remain stickied on "All Discussions" regardless of their read status.

n/a

**Screenshot**
Settings
Settings page

Default/turned on
Settings page

Turned off
Settings page

**QA**
1. On a forum with a heap of discussions, sticky a few random ones
2. Mark some at the top as "read" - refresh "All Discussions" page to see those discussions drop down the list
3. Switch the option `Filter read from stickied on All Discussions` page off
4. See all stickied discussions at the top of "All Discussions" page

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/extensions/sticky/extend.php
+++ b/extensions/sticky/extend.php
@@ -60,4 +60,8 @@ return [
 
     (new Extend\SimpleFlarumSearch(DiscussionSearcher::class))
         ->addGambit(StickyFilterGambit::class),
+
+    (new Extend\Settings())
+        ->default('flarum-sticky.filter_read_from_stickied', true)
+        ->serializeToForum('filterReadFromStickied', 'flarum-sticky.filter_read_from_stickied', 'boolval'),
 ];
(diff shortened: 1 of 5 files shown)


SychO9 (commented):
Thanks for the PR! 1.x is on feature freeze and only receives critical patches, could you please recreate this for 2.x 🙏🏼

adrianmccay:
@SychO9 yep, no worries!

PR #3995, opened 2024-05-29
feat: add option for filtering read stickied on all discussions page

adrianmccay:
This is an updated PR from the previous one that was pointing to 1.x. I've set up 2.x (as much as I could) locally and have tested this one - I have a bit less test data this time as I was unable to get the fake-data extension to install on 2.x.

As per discussion on Flarum Discuss - https://discuss.flarum.org/d/34607-flarum-do-perform-some-automated-update
For some, the stickied posts dropping down the order on "All Discusisons" like a non-stickied post when the discussion has been read is not an expected behaviour. I have added a simple on/off option which defaults to the existing behaviour. With the switch on, the read discussions on "All Discussions" are not treated as stickied. With the switch off, all stickied posts remain stickied on "All Discussions" regardless of their read status.

n/a

**Screenshot**
Settings
Settings page

Default/turned on
Settings page

Turned off
Settings page

**QA**
1. On a forum with a heap of discussions, sticky a few random ones
2. Mark some at the top as "read" - refresh "All Discussions" page to see those discussions drop down the list
3. Switch the option `Filter read from stickied on All Discussions` page off
4. See all stickied discussions at the top of "All Discussions" page

**Necessity**

**Confirmed**

- [x] Tests have been added, ~~or are not appropriate here.~~
 
**Required changes:**

### DIFF
--- a/extensions/sticky/locale/en.yml
+++ b/extensions/sticky/locale/en.yml
@@ -7,8 +7,8 @@ flarum-sticky:
   # Translations in this namespace are used by the admin interface.
   admin:
     settings:
-      filter_read_from_stickied_label: Filter read stickied (All Discussions)
-      filter_read_from_stickied_help: If a discussion is read, it will not be considered stickied on the All Discussions page
+      only_sticky_unread_discussions_label: Only sticky unread discussions
+      only_sticky_unread_discussions_help: On the All Discussions page, unread sticky discussions pin to the top, while read sticky discussions follow the regular order.
 
     # These translations are used in the Permissions page of the admin interface.
     permissions:
--- a/extensions/sticky/locale/en.yml
+++ b/extensions/sticky/locale/en.yml
@@ -6,6 +6,9 @@ flarum-sticky:
 
   # Translations in this namespace are used by the admin interface.
   admin:
+    settings:
+      filter_read_from_stickied_label: Filter read stickied (All Discussions)
+      filter_read_from_stickied_help: If a discussion is read, it will not be considered stickied on the All Discussions page
 
     # These translations are used in the Permissions page of the admin interface.
     permissions:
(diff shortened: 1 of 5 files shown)


extensions/sticky/src/PinStickiedDiscussionsToTop.php
```diff
@@ -11,26 +11,39 @@
...
 
 class PinStickiedDiscussionsToTop
 {
+    protected SettingsRepositoryInterface $settings;
+
+    public function __construct(SettingsRepositoryInterface $settings)
+    {
+        $this->settings = $settings;
+    }
+
```
SychO9:
```suggestion
    public function __construct(
        protected SettingsRepositoryInterface $settings
    ) {
    }

```

extensions/sticky/locale/en.yml
```diff
@@ -6,6 +6,9 @@ flarum-sticky:
 
   # Translations in this namespace are used by the admin interface.
   admin:
+    settings:
+      filter_read_from_stickied_label: Filter read stickied (All Discussions)
```
SychO9: I think it would be more understandable to say

```suggestion
      filter_read_from_stickied_label: Only sticky unread discussions
```

what do you think?

SychO9 (changes requested):

extensions/sticky/locale/en.yml
adrianmccay: Much better! I'll update the label and variables so they make sense. Thanks :)

adrianmccay:
@SychO9 I've made changes to the PR based on feedback - I also updated the help text so hopefully that isn't too horrible!

askvortsov1 (approved):

SychO9:
Thanks for the PR!

sorry for the delayed response @adrianmccay, could you please check the box to allow maintainers to make edits? it should be on the right sidebar of this page. So that I may fix the conflicts.

adrianmccay:
@SychO9 no worries - I couldnt find the option so tried closing this one to create a new one but still cant see the option. Maybe I set my fork up wrong.

SychO9:
@adrianmccay this is what it looks like:

image

If you still can't find it it's ok I can pull your commits into another branch and fix things there. Let me know.

SychO9:
https://github.com/flarum/framework/pull/4073

PR #3996, opened 2024-06-08
feat: vanilla CSS color scheme changes

SychO9:
**Fixes #0000**

Changes the LESS/CSS code to be able to switch color schemes using Vanilla CSS (browser level) rather than having to recompile all of the LESS code based on configuration values.

This also means the ability to easily have per-user color scheme preferences, so this PR also adds the ability to select a color scheme per user while checking the user system's color scheme preference as default.

This also removes the `@config-dark-mode` and `@config-colored-header` less variables as they are no longer used and should no longer be used in extensions in 2.0.

Instead one can do:

```css
[data-theme=dark] {
  .my-element { ... }
}

[data-colored-header=true] {
  .my-element { ... }
}
```

Additionally, this will allow easily adding high contrast color schemes and the like for a11y (accessibility).

**Screenshot**
image
image

**Necessity**

**Confirmed**

**Required changes:**


SychO9:
Added High contrast schemes for a11y

image
image

framework/core/js/src/common/Application.tsx
```diff
@@ -356,9 +358,61 @@ export default class Application {
...
+
+    // Listen for browser color scheme changes and update the theme accordingly
+    if (this.allowUserColorScheme) {
+      this.watchSystemColorSchemePreference(() => {
+        this.initColorScheme(forumDefault);
+      });
+    }
+  }
+
+  getSystemColorSchemePreference(): string {
```
askvortsov1: Why not type this as a string union?

askvortsov1 (approved):

framework/core/js/src/common/Application.tsx
SychO9: changed to `ColorScheme | string` but kept string because extensions may add their own color schemes and extend these methods.

PR #3998, opened 2024-06-16
[suspend][core] [1.x] fix: suspended users can remove avatar

imorland:
Reported https://discuss.flarum.org/d/32899-flarum-v180-released/83 and following on from https://github.com/flarum/framework/pull/3890

Dispatch `AvatarDeleting` before beginning the avatar deletion process, add a listener to `flarum/suspend` to prevent a suspended user from removing their avatar. Admins and other users with permission may still remove the avatar of the suspended user if required.

New tests for this scenario are included. Happy to make the same change against the `2.x` branch if required, please let me know.

**Screenshot**

**QA**

**Necessity**

**Confirmed**

### DIFF
--- a/extensions/suspend/extend.php
+++ b/extensions/suspend/extend.php
@@ -19,6 +19,7 @@ use Flarum\Suspend\Notification\UserSuspendedBlueprint;
 use Flarum\Suspend\Notification\UserUnsuspendedBlueprint;
 use Flarum\Suspend\Query\SuspendedFilterGambit;
 use Flarum\Suspend\RevokeAccessFromSuspendedUsers;
+use Flarum\User\Event\AvatarDeleting;
 use Flarum\User\Event\Saving;
 use Flarum\User\Filter\UserFilterer;
 use Flarum\User\Search\UserSearcher;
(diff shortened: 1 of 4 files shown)


imorland:
Ugh, damn StyleCI. See https://github.com/flarum/framework/pull/3998/commits/f7c0a3b7e95c1fecae9c1fcfae39fedd0149c3b0 for my changes here.

luceos (dismissed):
Can you please do the reformatting as a direct commit to main/master and then apply the change for the subject as a or?

Thanks.

luceos (commented):
I am a bit confused. It seems the abuse prevention @SychO9 implemented with https://github.com/flarum/framework/pull/3890/files#diff-d6304eb2471ddda50b8b56050c4e6ed1f6f0ac4b2511af0de4f363cddeca3bf8 uses the Gate. Why can't we either a) implement another policy ability for avatar removal or b) rename the implemented one to avatar changing. The intention of both PR's seem very similar, yet the implementations are entirely different with one using a Gate and the other Events.

imorland:
I agree that what you're saying would make sense for `2.x`, but I feel that if that was implemented in `1.x` we run the risk of introducing a breaking change by adding a `removeAvatar` gate.

How would we feel about the change proposed here for `1.x` as a "quick fix/patchup", and a solution based on a gate for `2.x`?

PR #4001, opened 2024-06-17
fix: cannot install without db prefix

luceos:
Database prefix and database password can be empty.

Caused by the rewrite in https://github.com/flarum/framework/commit/6f11e044a7c18c14793c7d98e99a83a6ade5ae97, as a consequence some settings of the database config are now incorrectly enforced/required.

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/framework/core/src/Database/Migrator.php
+++ b/framework/core/src/Database/Migrator.php
@@ -262,7 +262,7 @@ class Migrator
 
             $statement = str_replace(
                 'db_prefix_',
-                $this->connection->getTablePrefix(),
+                $this->connection->getTablePrefix() ?? '',
                 $statement
             );
             $this->connection->statement($statement);
--- a/framework/core/src/Install/DatabaseConfig.php
+++ b/framework/core/src/Install/DatabaseConfig.php
@@ -21,7 +21,7 @@ class DatabaseConfig implements Arrayable
         private string $database,
         private readonly ?string $username,
         private readonly ?string $password,
-        private readonly string $prefix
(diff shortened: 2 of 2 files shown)


askvortsov1 (commented):
Allowing no prefix makes sense, are we allowing no password for socket connections?

luceos:
> Allowing no prefix makes sense, are we allowing no password for socket connections?

You can use no password, on development and production. Whether it's advisable has to be seen. Regardless of that decision, a user will receive errors based on their credentials and the credentials the database server expects anyway.. But blatantly demanding a password would potentially block developers with a local dev environment where root has no password.

Ref: `MYSQL_ALLOW_EMPTY_PASSWORD` on https://hub.docker.com/_/mysql for instance.

askvortsov1 (approved):
Thanks!

SychO9:
Don't merge yet please. Till the big db ones are merged

luceos:
Ref:

- https://github.com/flarum/framework/pull/3985
- https://github.com/flarum/framework/pull/3984

Both merged, will rebase and see if the issue persisted.

luceos:
@SychO9 seems one of your PR's made password already optional, I assume sqlite because that's completely file based. So the only thing this PR now fixes is:

a) that prefixes are optional
b) that without a prefix no "preg_replace cannot use null as argument"-error is thrown

PS force pushed because I just made the changes on top of the merged/updated 2.x branch instead of rebasing.

SychO9 (approved):

PR #4004, opened 2024-06-28
feat(nicknames): allow OAuth to provide a nickname

blt950:

**Scenario**: User logins with OAuth2, the OAuth server provides and numeric `id` and a string `nickname` such as "Robert Lan". This nickname is not valid as username in Flarum and will fail the regex validation. Hence I wish to set the user's name as nickname as it can't be a username.

**Purpose of change:** Today the Nickname signup modal doesn't respect code to provide nickname when username is also supplied.
```
Example from OAuth call
->provide('username', $user->getId())
->provide('nickname', $user->getName())
``` 
This ends up with both nickname and username fields just replacing the username. This is because of this line https://github.com/flarum/framework/blob/873db6e00be8db86835e20d7244c14514a6c4b5a/extensions/nicknames/js/src/forum/index.js#L60 which only fetches a copy of the username in the nickname box. This is an issue because if you provide (instead of suggesting), the user should not be able to edit this field, though it's even more important that the provided value is displayed.

**What has changed**
I've just added a check for `attr.nickname` before it falls back to `username` or blank.  This seems to solve the issue.

**Necessity**

**Confirmed**

**Required changes:**


SychO9 (approved):
Thanks!

PR #4006, opened 2024-07-11
Export all missing modules in compat

DavideIadeluca:
**Fixes #0000**

Export all missing modules with the Compat API and therefore improve the extensibility of the framework.

We need to make sure that those exports are adopted in the 2.x equivalent (`forum.ts`). I suggest that the export registry upgrade step is used just before the first 2.0 beta is released.

**Screenshot**

**QA**
1. Try to extend/override a newly added component from core
2. Try to extend/override a newly added component from an extension
3. Try to use a newly added util which is the default export (i.e. `fireDebugWarning`)
4. Try to use a newly added util which is a non default export (i.e. `fireDeprecationWarning`)

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/extensions/emoji/js/src/forum/compat.ts
+++ b/extensions/emoji/js/src/forum/compat.ts
@@ -1,0 +1,7 @@
+import AutocompleteDropdown from './fragments/AutocompleteDropdown';
+import getEmojiIconCode from './helpers/getEmojiIconCode';
+
+export default {
+  'emoji/fragments/AutocompleteDropdown': AutocompleteDropdown,
+  'emoji/helpers/getEmojiIconCode': getEmojiIconCode,
+};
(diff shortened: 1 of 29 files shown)


imorland:
@DavideIadeluca what's the progress on this one, are you able to complete this soon-ish?

DavideIadeluca:
> @DavideIadeluca what's the progress on this one, are you able to complete this soon-ish?

I just did a round of testing and everything seems to be in order.  The CI is failing, but it doesn't appear to be related to the changes I made.

PR is ready for review from my side now.

imorland (approved):
Thank you for this, I'll include it in the upcoming `1.8.6` release

imorland:
Ps, would you mind doing the same for `2.x` when you have the chance, please?

DavideIadeluca:
> Ps, would you mind doing the same for `2.x` when you have the chance, please?

Yes, sure I can. I suggested doing it just before the first beta is released in case new components/utils are added/removed etc. and forgotten to be added in `forum.ts`. But I'm happy to do it anyway in the next coming days.

imorland:
I am reverting this until we can resolve the js typings errors in the CI..

@DavideIadeluca can you possibly take a look at this when you can?

Issue #4007, opened 2024-07-11
Tag state loaded for first user in `tag_user` table instead of logged-in user in `ShowTagController`


rafaucau:
### Current Behavior

The tag state is being loaded for the first user in the database instead of the currently logged-in user. This issue affects extensions like Follow Tags that rely on eager loading of the tag state, causing them to not function correctly.
See: https://github.com/FriendsOfFlarum/follow-tags/issues/72#issuecomment-2222283587

### Steps to Reproduce

1. Log in as a user other than the first user in the database.
2. Observe the subscription state of tag on tag page.
3. The tag state reflects the subscriptions of the first user in the database rather than the currently logged-in user.

### Expected Behavior

The subscription state of tags should be loaded for the currently logged-in user, ensuring that extensions like Follow Tags work as expected.

### Screenshots
Modified [vendor/flarum/tags/src/Api/Controller/ShowTagController.php](https://github.com/flarum/framework/blob/2a693db1b65cf7b0f7e76cbaa695a5e65bd947f6/extensions/tags/src/Api/Controller/ShowTagController.php) to display some info:
![image](https://github.com/flarum/framework/assets/25438601/8a4cec07-de8e-43ea-88b4-dc3785a74975)

### Environment

- Flarum version: 1.8.5
- PHP version: 8.3

### Additional Context

This issue was initially identified while troubleshooting an issue with the Follow Tags (v1.2.3) extension.
The problem lies in this controller: [ShowTagController](https://github.com/flarum/framework/blob/2a693db1b65cf7b0f7e76cbaa695a5e65bd947f6/extensions/tags/src/Api/Controller/ShowTagController.php)

https://github.com/flarum/framework/blob/2a693db1b65cf7b0f7e76cbaa695a5e65bd947f6/extensions/tags/src/Api/Controller/ShowTagController.php#L72

dsevillamartin:
That seems right regarding the source - this is what Clockwork shows as the query coming with no user constraint

![image](https://github.com/user-attachments/assets/c3460f68-4b02-424a-8e8d-3bb7837d5541)
![image](https://github.com/user-attachments/assets/b2164e48-22da-4759-8db8-13b638066a25)

luceos:
Is either of you able to provide a patch to 1.x. I'm not sure this applies to 2.x, I'd have to check.

rafaucau:
[...]

PR #4009, opened 2024-07-29
[2.x] perf(tags): prevent loading tag state if loaded previously

rafaucau:
This PR is similar to #4008, but targets the 2.x branch.
I have not been able to verify if the issue described in #4007 occurs in 2.x, as the Clockwork extension doesn't work with this version, preventing me from observing database queries. Nevertheless, implementing these changes is a precautionary step that can potentially prevent similar issues in the 2.x branch.

**Necessity**

### DIFF
--- a/extensions/tags/src/Tag.php
+++ b/extensions/tags/src/Tag.php
@@ -43,7 +43,7 @@ use Illuminate\Database\Query\Builder as QueryBuilder;
  * @property int $last_posted_user_id
  * @property string $icon
  *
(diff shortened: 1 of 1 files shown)


luceos (approved):
The logic looks sound. I thought the conditional could be simplified, but then we would remove the check for the state actually being null. So 👍 all good.

I would have loved to see the benefit of this on the queries though.

SychO9 (approved):

PR #4019, opened 2024-09-10
feat: post search adapted with global search

SychO9:
**Part of the 2.0 milestone**

* Adds searching of posts
  * A new search source in the search modal
  * A new page dedicated to post search results (which could technically also be viewed without search when directly going to the link /posts)
  * When a discussion, or user posts, or user liked posts, or user mentioned posts page is currently open, the search modal automatically selects the post search tab and automatically selects the correct filter in context to the current page.

**Screenshot**
image
---
image
image
image
image
image

**Necessity**

**Confirmed**

**Required changes:**


luceos:
Why is the implementation of author different than the other related to users? I feel like their use should be identical or at least similar in such a way that instead of relying on users to know IDs, we use usernames.

SychO9:
That's how the filter was implemented prior to this. I think these user filters should all support username and ID, and in the UI default to use the username, in the future we will likely add something to allow hiding usernames which is why ID still matters.

will modify them as such.

SychO9:
updated

image
image

luceos:
If this works, I'd say start merging everything in whenever it suits you and we'll start testing end-to-end when you give the sign.

💪 Excellent work, also from @iamdarkle. Really looks awesome.

PR #4023, opened 2024-09-25
chore: Bisect should use `CacheClearCommand` in core instead

datlechin:
Currently, the Flarum framework (https://github.com/flarum/framework/blob/2.x/framework/core/src/Extension/Bisect.php#L35) uses the `Composer\Command\ClearCacheCommand` class, but the core `composer.json` does not require the `composer/composer` package. This causes an error when running the `php flarum` command.

To resolve this, there are two possible approaches:

1. Move the `composer/composer` package to the core and remove it from the `extension-manager`.
2. Create a new command in the core to handle clearing the Composer cache by forking the necessary code from the `composer/composer` package.

In this PR, I've opted for the first solution by moving the `composer/composer` package to the core composer.json and removing it from the `extension-manager`.

**Fixes #0000**

**Screenshot**

**QA**

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/extensions/package-manager/composer.json
+++ b/extensions/package-manager/composer.json
@@ -22,7 +22,8 @@
         "source": "https://github.com/flarum/extension-manager"
     },
     "require": {
-        "flarum/core": "^2.0"
+        "flarum/core": "^2.0",
+        "composer/composer": "^2.7"
     },
     "require-dev": {
         "flarum/testing": "^2.0",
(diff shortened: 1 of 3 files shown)


SychO9:
Thanks for the PR!

The mistake is actually that the bisect class should be using the core `CleaeCacheCommand` instead. We would definitely not want to require composer in core otherwise.

Could you please make that change?

SychO9 (approved):

PR #4033, opened 2024-09-28
fix: respect user theme preference

datlechin:
This PR fixes the issue where `data-theme` defaulted to `auto`, causing a flash of light mode before switching to dark. It ensures the theme respects the user's preference for a smoother experience.

**Fixes #0000**

**Screenshot**

**QA**

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/framework/core/js/src/common/Application.tsx
+++ b/framework/core/js/src/common/Application.tsx
@@ -377,7 +377,7 @@ export default class Application {
   }
 
   private initColorScheme(forumDefault: string | null = null): void {
-    forumDefault ??= document.documentElement.getAttribute('data-theme') ?? 'auto';
+    forumDefault ??= app.forum.attribute('colorSchema') ?? 'auto';
     this.allowUserColorScheme = forumDefault === 'auto';
     const userConfiguredPreference = this.session.user?.preferences()?.colorScheme;
 
--- a/framework/core/js/src/common/Application.tsx
+++ b/framework/core/js/src/common/Application.tsx
@@ -377,7 +377,7 @@ export default class Application {
   }
 
   private initColorScheme(forumDefault: string | null = null): void {
-    forumDefault ??= app.forum.attribute('colorSchema') ?? 'auto';
+    forumDefault ??= app.forum.attribute('colorScheme') ?? 'auto';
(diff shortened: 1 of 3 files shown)


SychO9 (commented):
Thanks for the PR!

The reason auto is used is so that the JavaScript code can determine whether to use the user-preference or user browser color scheme.

If `data-theme` can no longer be `auto` it would undo that logic (see the `setColorScheme` method).

datlechin:
@SychO9 The logic behind auto is still preserved. If `color_scheme` is set to auto, it falls back to the user's browser preference via `RequestUtil::getActor()->getPreference('colorScheme')`. This ensures the same functionality while preventing the flash of light mode.

datlechin:
### Before:

https://github.com/user-attachments/assets/e197bf84-41de-419d-b0a5-43f089089d94

___

## After:

https://github.com/user-attachments/assets/3c812b68-d6ae-41d6-b41c-c37f1aa050b9

framework/core/src/Api/Resource/ForumResource.php
```diff
@@ -101,6 +101,8 @@ public function fields(): array
                 ->get(fn () => $this->settings->get('theme_primary_color')),
             Schema\Str::make('themeSecondaryColor')
                 ->get(fn () => $this->settings->get('theme_secondary_color')),
+            Schema\Str::make('colorSchema')
```
SychO9:
```suggestion
            Schema\Str::make('colorScheme')
```

framework/core/js/src/common/Application.tsx
```diff
@@ -377,7 +377,7 @@ export default class Application {
   }
 
   private initColorScheme(forumDefault: string | null = null): void {
-    forumDefault ??= document.documentElement.getAttribute('data-theme') ?? 'auto';
+    forumDefault ??= app.forum.attribute('colorSchema') ?? 'auto';
```
SychO9:
```suggestion
    forumDefault ??= app.forum.attribute('colorScheme') ?? 'auto';
```

SychO9 (changes requested):
Oh sorry, you're absolutely right!

SychO9 (approved):

PR #4036, opened 2024-09-28
chore: update FontAwesome v6 icon search link

datlechin:
**Fixes #0000**

**Screenshot**

**QA**

**Necessity**

**Confirmed**

**Required changes:**


SychO9 (changes requested):
Could you please do the same for any other fonteawesome links? like the ones in tags for example, there might be others.

I would suggest creating a new property in the common/Application class, maybe named `refs`.

and then we could use that (`app.refs.fontawesome`) every time we need to link to fontawesome search.

Also please use `https://fontawesome.com/v6/icons?o=r&m=free`

datlechin:
@SychO9, I have added a new property in the common Application, following your suggestion, to reuse `app.refs.fontawesome`. Please let me know if any changes are needed

framework/core/js/src/common/Application.tsx
```diff
@@ -248,6 +249,8 @@ export default class Application {
 
   allowUserColorScheme!: boolean;
 
+  refs: Record<string, unknown> = {};
```
SychO9: Let's directly add the value here instead of calling a function. The function isn't necessary at this time 👍

SychO9 (changes requested):

framework/core/js/src/common/Application.tsx
```diff
@@ -248,6 +248,10 @@ export default class Application {
 
   allowUserColorScheme!: boolean;
 
+  refs: Record<string, unknown> = {
```
SychO9: I think we can probably always expect the values to be strings for this one 🤔 

```suggestion
  refs: Record<string, string> = {
```

SychO9 (approved):
Thanks!

thatdh:
Seems `/icons?o=r&m=free` is broken (redirects to all icons)

main `https://fontawesome.com/v5/search?ic=free`
2.x `https://fontawesome.com/v6/search?ic=free`

PR #4044, opened 2024-09-29
[1.x] [extensibility] Export all missing modules in compat

DavideIadeluca:
**Fixes https://github.com/flarum/framework/pull/4006#issuecomment-2380559537**

- Export all missing modules with the Compat API and therefore improve the extensibility of the framework.
- Fix tsconfig

**Screenshot**

**QA**
1. Try to extend/override a newly added component from core
2. Try to extend/override a newly added component from an extension
3. Try to use a newly added util which is the default export (i.e. `fireDebugWarning`)
4. Try to use a newly added util which is a non default export (i.e. `fireDeprecationWarning`)

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/extensions/akismet/js/tsconfig.json
+++ b/extensions/akismet/js/tsconfig.json
@@ -15,6 +15,7 @@
     "declarationDir": "./dist-typings",
     "paths": {
       "flarum/*": ["../../../framework/core/js/dist-typings/*"],
+      "@flarum/core/*": ["../../../framework/core/js/dist-typings/*"],
       "flarum/flags/*": ["../../flags/js/dist-typings/*"]
     }
   }
--- a/extensions/emoji/js/src/forum/compat.ts
+++ b/extensions/emoji/js/src/forum/compat.ts
@@ -1,0 +1,7 @@
+import AutocompleteDropdown from './fragments/AutocompleteDropdown';
+import getEmojiIconCode from './helpers/getEmojiIconCode';
+
+export default {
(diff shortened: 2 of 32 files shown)


SychO9:
@DavideIadeluca for 2.x you don't actually need to do anything 👍 

In 2.x every module used by the code (core and extensions) is accessible by other extensions. Unless the module is declared but not used (which is not the case currently).

DavideIadeluca:
> @DavideIadeluca for 2.x you don't actually need to do anything 👍
> 
> In 2.x every module used by the code (core and extensions) is accessible by other extensions. Unless the module is declared but not used (which is not the case currently).

@SychO9 Yeah I was also of the same understanding per the diary you shared on discuss. But I'm kind of confused now what the usage is for the newly added `forum.ts` files in core and extensions?

https://github.com/flarum/framework/blob/da441c77caae7688a39099a2b9c39b1c40cc9604/framework/core/js/src/forum/forum.ts#L1-L53

https://github.com/flarum/framework/blob/da441c77caae7688a39099a2b9c39b1c40cc9604/extensions/flags/js/src/forum/forum.ts#L1-L10

SychO9:
@DavideIadeluca those are mostly the result of automatically converting the old compat files. A lot of the imports there are redundant and add nothing. However because we now lazy load certain components, those files serve as a way to ensure certain components are not lazy loaded, without having to constantly check where they are used.

We'll probably remove those (forum.ts, common.ts, admin.ts) files for the bundled extensions before 2.0 is out.

imorland (approved):

PR #4055, opened 2024-10-02
[1.x] [tags] feat: option to display tags in DiscussionsSearchSource

imorland:
Introduces an option to display the tag label(s) above the discussion title in search results (`DiscussionsSearchSource`). Disabled by default to maintain current behaviour.

@flarum/core I feel this is something to also include in `2.x`?

**Screenshot**
Search:
search

Tags settings:
tags settings

**QA**

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/extensions/tags/extend.php
+++ b/extensions/tags/extend.php
@@ -100,6 +100,8 @@ return [
         ->prepareDataForSerialization(LoadForumTagsRelationship::class),
 
     (new Extend\Settings())
+        ->default('flarum-tags.show_tags_in_discussion_search_results', false)
+        ->serializeToForum('showTagsInDiscussionSearchResults', 'flarum-tags.show_tags_in_discussion_search_results', 'boolVal')
         ->serializeToForum('minPrimaryTags', 'flarum-tags.min_primary_tags')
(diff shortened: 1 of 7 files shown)


SychO9:
Why not just make a custom extension? especially since this is not something that can be ported to 2.x as 2.x search UI is highly different and has tags by default

imorland:
> Why not just make a custom extension? especially since this is not something that can be ported to 2.x as 2.x search UI is highly different and has tags by default

Fair comment, I've not seen 2.x for some time, so didn't know about that!

Yes a custom extension can absolutely be created now that `DiscussionsSearchSource` and `DiscussionsSearchItem` can be extended. The feeling was to push this functionality back to the root so that it can potentially benefit the whole 1.x ecosystem :)

SychO9:
imho, a custom extension released to the wider community is just as beneficial for those looking for such a feature.

imorland:
> imho, a custom extension released to the wider community is just as beneficial for those looking for such a feature.

Fair enough, I'm having issues with typings (again!) on this one anyway!. Will close this and consider a standalone extension :)

PR #4074, opened 2024-10-17
fix: usage of aria-hidden=false might cause inconsistent behavior

### LINKED ISSUE #4014: aria-hidden="false" is unreliable for accessibility
### Current Behavior

There's an `aria-hidden="false"` on the main #app div, where most of the page content is loaded. This doesn't work reliably in assistive technology - it's just as likely to be interpreted as `aria-hidden="true"` as `aria-hidden="false"`, so in spite of your consideration, some screen reader users (or voice input users, who also rely on the accessibility tree) still may not be able to see or interact with the main content at all 😅 

### Steps to Reproduce

1. go to pretty much any page
2. observe the aria-hidden attribute in the #app div (class and style attributes removed for clarity): `<div id="app" aria-hidden="false">`

### Expected Behavior

no potential for unreliability!

### Screenshots

_No response_

### Environment

- Flarum version: 1.8.5

```
```

### Possible Solution

The best way to treat `aria-hidden` is as if it were a boolean attribute instead, so when it's true, it says `<div id="app" aria-hidden="true">`, and when it's false it just says `<div id="app">`.

I'm not yet confident enough with Flarum to try submitting a pull request, but I'm hoping this is a simple enough fix that I'll be able to use it as a starter demonstration if someone is able to pick it up! 

### Additional Context

See the second warning note on the ARIA 2.1 spec for the `aria-hidden` state here (copied below): https://www.w3.org/TR/wai-aria-1.2/#aria-hidden

> At the time of this writing, [aria-hidden](https://www.w3.org/TR/wai-aria-1.2/#aria-hidden)="false" is known to work inconsistently in browsers. As future implementations improve, use caution and test thoroughly before relying on this approach.

YUCLing:
**Fixes #4014**

Replace all usage of `aria-hidden=false` with removing the attribute instead.

**Screenshot**

**QA**
Check if aria-hidden=false is still used in DOM tree.

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/framework/core/js/src/admin/components/UserListPage.tsx
+++ b/framework/core/js/src/admin/components/UserListPage.tsx
@@ -318,7 +318,7 @@ export default class UserListPage extends AdminPage {
               )
             );
 
-            emailAddress.attr('aria-hidden', visible ? 'false' : 'true');
+            emailAddress.attr('aria-hidden', visible ? null : 'true');
 
             if (visible) {
               emailToggleButtonIcon.addClass('fa-eye');
--- a/framework/core/js/src/common/components/ModalManager.tsx
+++ b/framework/core/js/src/common/components/ModalManager.tsx
@@ -98,7 +98,7 @@ export default class ModalManager extends Component<IModalManagerAttrs> {
         // See: http://web-accessibility.carnegiemuseums.org/code/dialogs/
 
         if (!this.attrs.state.isModalOpen()) {
-          document.getElementById('app')?.setAttribute('aria-hidden', 'false');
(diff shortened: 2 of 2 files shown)


SychO9 (commented):
Thanks for the PR!

could you check if ModalManager line 47 is also one of them? seems like it can end up being false too

YUCLing:
> seems like it can end up being false too

I just confirmed it, using `false` boolean value seems to remove the whole attribute in Mithril.

SychO9 (approved):
Ah true. Makes sense.

hollsk:
Thanks @YUCLing (and @SychO9)! 😄

PR #4075, opened 2024-10-17
[1.x] [likes] fix: Ignore isLiked when post is not comment post.

zxy19:

When someone set `isLiked` attribute on a post that's not a `comment` (e.g. type `discussionTagged`) through api, it will be saved and send the notification to user. Then in `PostLikedNotification` component, the method `excerpt` is trying to get the `plainContent` to truncate, which will lead to a frontend error.

According to `addLikeAction.js`, user should only be able to like a comment post, so it should be ignore if target post is not a comment post.

**Necessity**

**Confirmed**

### DIFF
--- a/extensions/likes/migrations/2024_10_18_000000_detach_likes_from_non_comment_posts.php
+++ b/extensions/likes/migrations/2024_10_18_000000_detach_likes_from_non_comment_posts.php
@@ -1,0 +1,24 @@
+<?php
+
+/*
+ * This file is part of Flarum.
+ *
+ * For detailed copyright and license information, please view the
+ * LICENSE file that was distributed with this source code.
+ */
+
+use Illuminate\Database\Schema\Builder;
+
+return [
+    'up' => function (Builder $schema) {
+        // Detach likes on non-comment posts
+        $schema->getConnection()
+            ->table('post_likes')
+            ->whereNotExists(function ($query) {
+                $query->selectRaw(1)->from('posts')->whereColumn('id', 'post_id')->where('type', 'comment');
+            })
+            ->delete();
+    },
+    'down' => function (Builder $schema) {
+    }
+];
--- a/extensions/likes/src/Listener/SaveLikesToDatabase.php
+++ b/extensions/likes/src/Listener/SaveLikesToDatabase.php
@@ -34,7 +34,7 @@ class SaveLikesToDatabase
         $post = $event->post;
         $data = $event->data;
 
-        if ($post->exists && isset($data['attributes']['isLiked'])) {
+        if ($post->exists && isset($data['attributes']['isLiked']) && $post->type === 'comment') {
             $actor = $event->actor;
             $liked = (bool) $data['attributes']['isLiked'];
 


YUCLing:
In that case, you'll be better off detaching all the likes from the posts. Since the like and unlike are both forbidden now, there's no way to achieve it except by modifying the database.

zxy19:
> In that case, you'll be better off detaching all the likes from the posts. Since the like and unlike are both forbidden now, there's no way to achieve it except by modifying the database.

You are right. A new migration file was add to remove those records.

YUCLing:
Hang on... I think this fix will reduce the extensibility of the Likes extension. Another way of fixing this is preferred. 🤔

extensions/likes/src/Listener/SaveLikesToDatabase.php
```diff
@@ -34,7 +34,7 @@ public function whenPostIsSaving(Saving $event)
         $post = $event->post;
         $data = $event->data;
 
-        if ($post->exists && isset($data['attributes']['isLiked'])) {
+        if ($post->exists && isset($data['attributes']['isLiked']) && $post->type === 'comment') {
```
ExerciseBook, line 37: I think this is not a common usage.

If we do want to LIKE a main discussion?

extensions/likes/migrations/2024_10_18_000000_detach_likes_from_non_comment_posts.php
```diff
@@ -0,0 +1,24 @@
...
+
+return [
+    'up' => function (Builder $schema) {
+        // Detach likes on non-comment posts
+        $schema->getConnection()
+            ->table('post_likes')
+            ->whereNotExists(function ($query) {
+                $query->selectRaw(1)->from('posts')->whereColumn('id', 'post_id')->where('type', 'comment');
+            })
+            ->delete();
```
ExerciseBook, line 20: is it no need to delete?

just keep it in database for rolling back.

ExerciseBook (changes requested):

ExerciseBook:
If you do consider it as a big due, just filter this logic in client side.

Keep connection IDL simple and do things on client side maybe a better choice.

extensions/likes/src/Listener/SaveLikesToDatabase.php
ExerciseBook, line 37: What we can do if we do want to LIKE a main discussion?

extensions/likes/src/Listener/SaveLikesToDatabase.php
```diff
@@ -34,7 +34,7 @@ public function whenPostIsSaving(Saving $event)
...
             $actor = $event->actor;
             $liked = (bool) $data['attributes']['isLiked'];
```
ExerciseBook, line 40: adding it in `$actor->assertCan('like', $post);` may be better?

ExerciseBook (changes requested):

luceos:
So the issue here is:

Liking any other Post type than CommentPost is causing issues with notifications.

The PR #4095 relates to this exact issue as well. Reference: https://github.com/flarum/framework/pull/4095#discussion_r1839725082

What I think should happen is create an extensible layer by adding a property to the implementation of every `Post` type, eg `CommentPost` that indicates whether notifications should be sent.

I'm closing this PR and the other one and will create an issue to explain further.

PS deleting all previous likes to start with a clean slate will not be appreciated by our users.

PR #4087, opened 2024-10-22
Fix Alert component layout

YUCLing:
**Fixes #0000**

Adds `gap` for `.Alert-container` in admin styles.

This change improves the readability of the alert.

**Screenshot**
Before
image
After
image

**QA**

**Necessity**

**Confirmed**

**Required changes:**


SychO9 (changes requested):
Thanks for the PR!

This is what it actually should look like:
image

 looks like I broke by adding flex display on the alert container while trying to fix this:
image

Best way to fix this is to wrap Alert-title and Alert-body divs into a Alert-content div.

YUCLing:
> Best way to fix this is to wrap Alert-title and Alert-body divs into a Alert-content div.

Fixed it. See if it looks good to you.
image
image

SychO9 (approved):
Thanks!

PR #4088, opened 2024-10-23
chore: switch formatter to `format-message`

### LINKED ISSUE #3685: Javascript translator may return HTML entities when using plural rules
### Current Behavior

This is a high level description of the issue. The problem has already been identified and lies in the internals of the `rich-icu-message-formatter` library.

Detailed issue https://github.com/askvortsov1/rich-icu-message-formatter/issues/3

Calling `app.translator.trans('core.lib.meta_titles.without_page_title', {forumName: 'A & B', pageNumber: 1})` returns `['A &amp; B']` instead of the expected `['A & B']`

The translation is defined as such:

```yaml
without_page_title: "{pageNumber, plural, =1 {{forumName}} other {Page # - {forumName}}}"
```

The problem doesn't exist with translations that only have a flat message without any pluralization switch.

The problem happens whenever the `{placeholder}` isn't in the same place in the sentence in the plural variant. If both placeholders are at the same index, the entities are correctly decoded.

The issue is most noticeable when using the "1" or "0" rule of a pluralization message since "other" is most likely to be selected at random by the ICU library. But since this is partly based on a randomly generated value, the behavior can be different once in a while.

### Steps to Reproduce

.

### Expected Behavior

.

### Screenshots

_No response_

### Environment

- Flarum version: test on dev-main, applies at least back to 1.5
- Hosting environment: local
- PHP version: irrelevant
- Browser: irrelevant, can reproduce in jest unit tests

```
```

### Possible Solution

Described in the ICU issue. Pluralization parameters need to be isolated from other placeholders so that the entire logic runs with the same pluralization instead of using singular for part of the code and plural for the other half.

### Additional Context

This is the underlying cause behind these previous issues and fixes for the page title. But the underlying issue applies to a lot more than just the title. It just happens that Flarum title uses a pluralization rule.

https://github.com/flarum/framework/issues/3514
https://github.com/flarum/framework/pull/3542

### LINKED ISSUE #3702: ICU message formatter used in JS incorrectly handles `'{placeholder}'` syntax
### Current Behavior

Placeholders wrapped between `''` should be treated as literal string, so `User: '{user}'` should be always translated to `User: '{user}'`. Currently it works correctly if PHP translator is used, but JS translator ignores quoting and replaces placeholders in every case, and produces `User: 'rob006'`. As a result translators needs to quote differently depending on context - we need `User: '{user}'` for JS and `User: ''{user}''` for PHP. There is no convenient way to know where specific string is used, so this is super inconvenient for languages that use single quote by default. Currently there is \~40 such translations on Weblate: https://weblate.rob006.net/search/flarum/?offset=1&q=target%3Ar%22%27%5C%7B%22&sort_by=context&checksum= and it is hard to say which translations are correct and which should be fixed by replacing `'` with `''`.

### Steps to Reproduce

1. Create `User: ''{user}''` translation.

### Expected Behavior

Results for `$translator->trans('my-key', ['user' => 'rob006'])` and ` app.translator.trans('my-key', { user: 'rob006' })` should be the same: `User: 'rob006'`.

### Screenshots

_No response_

### Environment

- Flarum version: 1.6.2
- PHP version: 7.4.33

_No response_

### Possible Solution

_No response_

### Additional Context

_No response_

### LINKED ISSUE #3734: Issue of encoding apostrophes when using a plural rule in translation
### Current Behavior

I have an issue that I can't solve. I'm the author of the French translation of Flarum and its extensions. 

This issue concerns the display of the apostrophe which is badly encoded. Instead of displaying `'`, it displays `&#039;`. The issue seems to appear only when there is a singular or plural rule.

You can see [on line 69 of the translation file](https://github.com/flarum-lang/french/blob/master/locale/flarum-tags.yml): 
`      tags_text: '{count, plural, one {l''étiquette {tags}} other {les étiquettes {tags}}}'`

This should work, but it displays `l&#039;étiquette`, not `l'étiquette`.

Again, the display of apostrophes works everywhere else in the translation, it's really related to the plural rule. I also checked, the file is encoded in UTF8 without BOM so that's okay. Do you have any idea how to solve this issue? Thank you.

This issue have been reported by a user of the translation: https://discuss.flarum.org/d/615-french-language-pack/314

### Steps to Reproduce

Install the French language pack and see the issue.

You can see the issue on: https://lepointdarret.com/public/d/32-avatar-autre-que-gravatar

### Expected Behavior

Display `'` instead of  `&#039;`.

### Screenshots

_No response_

### Environment

- Flarum version: 1.6.3

```
```

### Possible Solution

_No response_

### Additional Context

_No response_

SychO9:
**Fixes #3702**
**Fixes #3734**
**Fixes #3685**
**Closes #3735**

* Switches to the more mature `format-message` package instead of the less capable one we currently use.

**Necessity**

**Confirmed**


framework/core/js/tests/unit/common/utils/Translator.test.ts
```diff
@@ -0,0 +1,59 @@
...
+});
+
+test('missing placeholders', () => {
+  const translator = new Translator();
+  translator.addTranslations({
+    test1: 'test1 {placeholder} test1',
+  });
+
+  expect(extractText(translator.trans('test1', {}))).toBe('test1 {placeholder} test1');
+});
```
SychO9: @rob006 the only failing test is this one, isn't the test actually incorrect since the placeholder is neither escaped nor provided? just making sure, you're more familiar with this.

image

rob006: Unfortunately, I don't remember the details. I think it was defined in this way, because PHP translator works like that, so the goal was to make JS and PHP implementations consistent. But I don't think this scenario is important and some discrepancy may be accepted in edge cases like that. I would just leave commented version of this test with note that this case works differently in PHP.

rob006: Also, IIRC some JS implementations throws exception in case of missing params, so clearly there is no consensus what to do in that scenario. IMO leaving params as is is the golden mean - it does not break app completely, but it is clearly visible that something is missing.

SychO9: Makes sense. The only alternative thing we can do is leave it as undefined (I made a change to empty it) that does communicate that there should be something there, wdyt?

Screenshot from 2024-10-23 18-14-44

rob006: I'm not sure, it is the 4th way of handling this and it may create confusing results. :D But it also kinda make sense in JS world.

rob006: Maybe use `{undefined}` as placeholder in that case? `undefined` may sound like actual value in many cases and using it may make it even harder to realize that param is missing.

SychO9:
**forum side:** +3.11KB
**admin side:** +2.85KB

luceos (approved):

PR #4095, opened 2024-10-25
[likes] Don't send liked notification for posts without plain content.

YUCLing:
**Fixes #0000**

Don't send liked notifications for posts that don't have plain content.

In this approach, the extensibility of the Likes extension won't be reduced. Since it allows all other post types to be liked, it also keeps the original notification ability from the Likes extension for all other post types that contain valid plain content.

**Screenshot**

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/extensions/likes/src/Listener/SendNotificationWhenPostIsLiked.php
+++ b/extensions/likes/src/Listener/SendNotificationWhenPostIsLiked.php
@@ -22,7 +22,7 @@ class SendNotificationWhenPostIsLiked
 
     public function handle(PostWasLiked $event): void
     {
-        if ($event->post->user && $event->post->user->id != $event->user->id) {
+        if (gettype($event->post->content) == "string" && $event->post->user && $event->post->user->id != $event->user->id) {
             $this->notifications->sync(
                 new PostLikedBlueprint($event->post, $event->user),
                 [$event->post->user]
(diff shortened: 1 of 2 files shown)


extensions/likes/src/Listener/SendNotificationWhenPostIsLiked.php
```diff
@@ -22,7 +22,7 @@ public function __construct(
 
     public function handle(PostWasLiked $event): void
     {
-        if ($event->post->user && $event->post->user->id != $event->user->id) {
+        if (gettype($event->post->content) == 'string' && $event->post->user && $event->post->user->id != $event->user->id) {
```
luceos, line 25: - It would be better to do an exact `===` comparison while using comparisons like these
- Instead of using `gettype`, just using the php `is_string()` method would probably be sufficient, it also reads better instead of seeing a `'string'` for something that is a native php type

My biggest gripe with this is however that there's no guarantee whatsoever that `content` is going to be just post body; what I would probably do is:

```suggestion
        if ($event->post->content instanceof CommentPost && $event->post->user && $event->post->user->id != $event->user->id) {
```

Now this gripe is based on the idea that notifications for likes would only be sent for posts that have readable content, but it would probably be better to completely separate the logic into a "Post Types that send Notifications" logic, so that it becomes flexible enough for extending; what do you think @flarum/core ?

luceos (changes requested):

luceos:
So the issue here is:

Liking any other Post type than CommentPost is causing issues with notifications.

The PR #4075 relates to this exact issue as well.

What I think should happen is create an extensible layer by adding a property to the implementation of every `Post` type, eg `CommentPost` that indicates whether notifications should be sent.

I'm closing this PR and the other one and will create an issue to explain further.

PR #4101, opened 2024-10-30
chore: remove frontend scoping in `Modal`

DavideIadeluca:
**Fixes #0000**
TS will throw errors of various kinds if a third-party extension tries to customize the behaviour of existing Modals. For example, a third-party implementation might want to remove the ability to close the `SignUpModal` via the close button:

```tsx
import SignUpModal from 'flarum/forum/components/SignUpModal';

export default function () {
    SignUpModal.isDismissibleViaCloseButton = false;
  }
```

This will result in the following error: 

`Cannot assign to 'isDismissibleViaCloseButton' because it is a read-only property.ts(2540)`

- Remove explicit scoping of properties, methods and getter in the `Modal` class and remove `readonly` modifier.

This change got me thinking... Do we really need to constrain scopes in the JS of the framework, like at all? I think in the spirit of Flarum it would be a welcome change for extension developers if such scoping is removed and TS workarounds related to this were no longer required.

I would be happy to amend this PR do the change across the whole codebase 

**Screenshot**

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/framework/core/js/src/common/components/Modal.tsx
+++ b/framework/core/js/src/common/components/Modal.tsx
@@ -34,15 +34,15 @@ export default abstract class Modal<ModalAttrs extends IInternalModalAttrs = IIn
    *
    * If `false`, no close button is shown.
    */
-  protected static readonly isDismissibleViaCloseButton: boolean = true;
+  static isDismissibleViaCloseButton: boolean = true;
   /**
    * Can the modal be dismissed by pressing the Esc key on a keyboard?
    */
-  protected static readonly isDismissibleViaEscKey: boolean = true;
+  static isDismissibleViaEscKey: boolean = true;
   /**
    * Can the modal be dismissed via a click on the backdrop.
    */
-  protected static readonly isDismissibleViaBackdropClick: boolean = true;
+  static isDismissibleViaBackdropClick: boolean = true;
 
   static get dismissibleOptions(): IDismissibleOptions {
     return {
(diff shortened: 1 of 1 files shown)


SychO9 (changes requested):
Thanks for the PR!

Generally, it depends on the specific case. We mark certain things as protected or readonly when we believe something might change in a minor release, or that there is a better way to make the modification, that way we are still able to make an amount of changes that are not considered breaking. It also allows the extension developer to know that they are modifying something that they should expect might change soon and is therefore not considered public API.

The alternative is that more changes to the core frontend are more likely to be breaking changes, which makes maintenance more difficult.

In this case things should remain protected imo. The reason is that doing `SignUpModal.isDismissibleViaCloseButton = false;` will not work in 2.0

The SignUpModal and most modals are now lazy loaded, so we encourage you to use the `extend` or `override` helper instead on a method, and make the modification there, where protected vs public wouldn't matter as it is would be a bound method.

I would suggest keeping the properties readonly, and opening a single point of extension, which would be making the dismissibleOptions instance getter a normal method instead and also public, so that you can do:

```ts
override('flarum/forum/components/SignUpModal', 'dismissibleOptions', function () {
  return {
    viaCloseButton: false,
    viaEscKey: this.isDismissibleViaEscKey,
    viaBackdropClick: false,
  };
});
```

This would both allow customization without creating multiple points of entry 👍🏼

DavideIadeluca:
Thanks for the detailed insights! Much appreciated.

Closing this one and will reevaluate again once I have a proper 2.x dev environment locally.

PR #4103, opened 2024-11-07
test: php 8.4

SychO9:
Adds PHP 8.4 Compatibility (fixes deprecations)

Primarily, PHP 8.4 deprecates arguments with a null default value but no nullable type.

### DIFF
--- a/framework/core/src/Database/Exception/MigrationKeyMissing.php
+++ b/framework/core/src/Database/Exception/MigrationKeyMissing.php
@@ -13,13 +13,13 @@ use Exception;
 
 class MigrationKeyMissing extends Exception
 {
-    public function __construct(protected string $direction, string $file = null)
+    public function __construct(protected string $direction, ?string $file = null)
     {
         $fileNameWithSpace = $file ? ' '.realpath($file) : '';
         parent::__construct("Migration file $fileNameWithSpace should contain an array with up/down (looking for $direction)");
     }
 
-    public function withFile(string $file = null): self
+    public function withFile(?string $file = null): self
     {
         return new self($this->direction, $file);
     }
--- a/framework/core/src/Database/Exception/MigrationKeyMissing.php
+++ b/framework/core/src/Database/Exception/MigrationKeyMissing.php
@@ -19,7 +19,7 @@ class MigrationKeyMissing extends Exception
         parent::__construct("Migration file $fileNameWithSpace should contain an array with up/down (looking for $direction)");
     }
 
-    public function withFile(?string $file = null): self
+    public function withFile(string $file): self
     {
         return new self($this->direction, $file);
     }
--- a/framework/core/src/Discussion/Discussion.php
+++ b/framework/core/src/Discussion/Discussion.php
@@ -117,7 +117,7 @@ class Discussion extends AbstractModel
     /**
      * Start a new discussion. Raises the DiscussionWasStarted event.
      */
-    public static function start(?string $title, User $user, self $model = null): self
+    public static function start(?string $title, User $user, ?self $model = null): self
     {
         $discussion = $model ?? new static;
 
--- a/framework/core/src/Mail/Mailer.php
+++ b/framework/core/src/Mail/Mailer.php
@@ -21,7 +21,7 @@ class Mailer extends IlluminateMailer
         string $name,
         Factory $views,
         TransportInterface $transport,
-        Dispatcher $events = null,
+        ?Dispatcher $events,
         protected SettingsRepositoryInterface $settings
     ) {
         parent::__construct($name, $views, $transport, $events);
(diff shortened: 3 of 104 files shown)


framework/core/src/Database/Exception/MigrationKeyMissing.php
```diff
@@ -13,13 +13,13 @@
...
 {
-    public function __construct(protected string $direction, string $file = null)
+    public function __construct(protected string $direction, ?string $file = null)
     {
         $fileNameWithSpace = $file ? ' '.realpath($file) : '';
         parent::__construct("Migration file $fileNameWithSpace should contain an array with up/down (looking for $direction)");
     }
 
-    public function withFile(string $file = null): self
+    public function withFile(?string $file = null): self
```
luceos: That file being optional seems to be counter intuitive? One could just call construct without the 2nd argument or we could introduce a `withoutFile` method instead?

framework/core/src/Discussion/Discussion.php
```diff
@@ -117,7 +117,7 @@ public static function boot()
     /**
      * Start a new discussion. Raises the DiscussionWasStarted event.
      */
-    public static function start(?string $title, User $user, self $model = null): self
+    public static function start(?string $title, User $user, ?self $model = null): self
```
luceos, line 120: Optional arguments should always be listed after required ones. I know this is going to be non-backward compatible though. Not sure if we want to take care of that now, but it should be documented at least.

But considering this PR is going to be breaking every extension that introduce their own extenders, we might as well tackle this in one swoop too?

framework/core/src/Mail/Mailer.php
```diff
@@ -21,7 +21,7 @@ public function __construct(
         string $name,
         Factory $views,
         TransportInterface $transport,
-        Dispatcher $events = null,
+        ?Dispatcher $events,
```
luceos, line 24: Removing the default `null` value will increase the required code outside of this class, is there a specific reason to do this?

framework/core/src/Mail/MutateEmail.php
```diff
@@ -15,10 +15,10 @@ class MutateEmail
 {
     public function handle(MessageSending $event): bool
     {
-        if (! empty($link = $event->data['unsubscribeLink'])) {
+        if (! empty($event->data['unsubscribeLink'])) {
```
luceos:
```suggestion
        $unsubscribeLink = Arr::get($event->data, 'unsubscribeLink');
        if (! $unsubscribeLink) {
```

Can't we use something like this for better code readability?

Imports not loaded nor variable added in line 21.

luceos (changes requested):
Wow, just wow. So many instances where nullable variables weren't properly type hinted 🙈 

I left a couple of questions, hope you don't mind.

framework/core/src/Database/Exception/MigrationKeyMissing.php
SychO9: @luceos there is no difference in behavior, it was optional before, it is still optional now, the typing is just more clear that it can be null

before you could call ->withFile() with no parameter, but couldn't call ->withFile(null), now you can call ->withFile(null) because it should be allowed if you can call ->withFile() and the arg is assumed null

this applies to all changes in this pr 👍🏼

framework/core/src/Discussion/Discussion.php
SychO9: @luceos this is not a breaking change, model is the only optional argument and it is provided last, a nullable argument is not an optional argument 👍🏼

framework/core/src/Mail/Mailer.php
SychO9: @luceos dependency injection fills the value

SychO9: there is a non optional argument after it, so it being optional or not makes no difference in this case. it can't be optional, only nullable

framework/core/src/Database/Exception/MigrationKeyMissing.php
luceos: Yeah, I completely understand the logic the method offers. I'm talking about the method being named "with file" and the method allowing a null value, implying there is no file. Semantically, the logic of the method goes against the name of the method.

framework/core/src/Discussion/Discussion.php
luceos: > @luceos this is not a breaking change, model is the only optional argument and it is provided last, a nullable argument is not an optional argument 👍🏼

I missed that, good call.

framework/core/src/Database/Exception/MigrationKeyMissing.php
SychO9: @luceos I understand, but that's not really a semantic added by this PR anyway, it's there before and after this PR, so it's out of scope for the purpose of the changes here, it would be best to open an issue for new stuff, just we can stay focused on one thing at a time imo 🫡

SychO9: I've pushed a commit to make the argument required since it's only used in one spot that provides a string

PR #4107, opened 2024-11-09
[1.x] feat: conditional extend whenExtensionDisabled

imorland:
**Fixes #0000**

Counterpart to #3847 for `1.x`

**Screenshot**

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/framework/core/src/Extend/Conditional.php
+++ b/framework/core/src/Extend/Conditional.php
@@ -48,6 +48,20 @@ class Conditional implements ExtenderInterface
     }
 
     /**
+     * Apply extenders only if a specific extension is disabled/not installed.
+     *
+     * @param string $extensionId The ID of the extension.
+     * @param ExtenderInterface[]|callable|string $extenders A callable returning an array of extenders, or an invokable class string.
+     * @return self
+     */
+    public function whenExtensionDisabled(string $extensionId, $extenders): self
+    {
+        return $this->when(function (ExtensionManager $extensions) use ($extensionId) {
+            return ! $extensions->isEnabled($extensionId);
+        }, $extenders);
+    }
+
+    /**
      * Apply extenders based on a condition.
      *
      * @param bool|callable $condition A boolean or callable that should return a boolean.
--- a/framework/core/tests/integration/extenders/ConditionalTest.php
+++ b/framework/core/tests/integration/extenders/ConditionalTest.php
@@ -15,6 +15,7 @@ use Flarum\Extend;
 use Flarum\Extension\ExtensionManager;
 use Flarum\Testing\integration\RetrievesAuthorizedUsers;
 use Flarum\Testing\integration\TestCase;
+use s9e\TextFormatter\Configurator\TemplateNormalizations\Custom;
 
 class ConditionalTest extends TestCase
 {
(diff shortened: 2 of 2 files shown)


framework/core/src/Extend/Conditional.php
```diff
@@ -47,6 +47,20 @@ public function whenExtensionEnabled(string $extensionId, $extenders): self
...
+     * @param string $extensionId The ID of the extension.
+     * @param ExtenderInterface[]|callable|string $extenders A callable returning an array of extenders, or an invokable class string.
+     * @return self
+     */
+    public function whenExtensionDisabled(string $extensionId, $extenders): self
+    {
+        return $this->when(function (ExtensionManager $extensions) use ($extensionId) {
+            return ! $extensions->isEnabled($extensionId);
+        }, $extenders);
+    }
```
luceos, line 62: The function method, to me, implies the extension is installed, not enabled. Yet the logic here makes no difference with extensions that are installed and disabled versus extensions that aren't installed at all.

I fear that this confusion will increase the longer this logic remains as is. I would highly recommend adding the missing methods `whenExtensionInstalled` and `whenExtensionMissing` or something similar to create clear usage examples and reduced confusion.

luceos (changes requested):

framework/core/src/Extend/Conditional.php
SychO9: Understandable line of thinking, but is there an actual use case for that though?

I can't think of a scenario where those would make a difference 🤔

Why add extenders just for when an extension is specifically missing rather than "disabled or missing"

Or why specifically when installed regardless of enabled/disabled

luceos: Extensibility is never about what scenario's we can imagine to warrant certain features. 

But regardless, an example would be that an extension would re-use the code from another extension, even if it is not enabled, as PHP code can be re-used from another extension in whatever state it is in.

And even if there wouldn't be a use case for this, code that adds confusion should always be prevented. I believe the current implementation is not without confusion, so why not try to aim for a better solution?

SychO9: > Extensibility is never about what scenario's we can imagine to warrant certain features.

I would disagree with that. We've always led by use case in order not to bloat core with unnecessary additions.

> an example would be that an extension would re-use the code from another extension,

True, that's a valid use case for installed package check where enabled state doesn't matter.

> code that adds confusion should always be prevented. I believe the current implementation is not without confusion

I personally don't think whenDisabled is confusing as the use case for it is to apply extenders when an extension is not running (missing or not doesn't matter here imo)

But I do see now how you'd want to tell installation state or lack of it 👍

imorland: So how should I proceed here?

SychO9: sorry, not objecting to the proposed additions

luceos: My suggestion would be to close this comment on the `1.x` branch, but make amends on the `2.x` branch to ensure we will not bring this into the future as is. At the time of my comment I wasn't aware this was a copy of the changes to 2.x.

Unless we really want to try to keep 1.x as identical as possible.

PR #4117, opened 2024-11-18
[1.x] fix: change length of email field

DavideIadeluca:
`1.x` pendent to #4118

**Fixes #0000**
It appears as if the email validator allows 254 chars for the entire email, while the DB field only allows 150. This results in email addresses simply being truncated without any error message when creating users with longer email addresses.

Increase length of email field to 254 chars to match the validator

See https://www.rfc-editor.org/errata/eid1690

**Screenshot**

**Necessity**

**Confirmed**

**Required changes:**


imorland (approved):
Looks all good to me.

I think for the tests, we only really need test the longest valid and assert that longer fails. Testing the local and domain parts feels like it's unnecessary and we're just testing the validator, which should already be done in Laravel?

Anyhow, I'm happy overall here, good job!

PR #4118, opened 2024-11-18
fix: change length of email field

DavideIadeluca:
`2.x` pendent to #4117 

**Fixes #0000**
It appears as if the email validator allows 254 chars for the entire email, while the DB field only allows 150. This results in email addresses simply being truncated without any error message when creating users with longer email addresses.

Increase length of email field to 254 chars to match the validator

See https://www.rfc-editor.org/errata/eid1690

**Screenshot**

**Necessity**

**Confirmed**

**Required changes:**


SychO9 (changes requested):
Looks like the reason it's failing is that the email tokens table also needs its email column to be increased in length, could you please add a migration for that as well? MySQL probably just truncates without failing whereas PgSQL is more strict so it fails. Which is likely why tests succeeded in 1.x

but that also means email tokens (needed for account confirmations) won't work properly for emails longer than 150 in 1.x, it looks like confirming the account in 1.x will trigger the app to change the email back to the truncated value, assuming that's what's happening (https://github.com/flarum/framework/blob/1.x/framework/core/src/User/Command/ConfirmEmailHandler.php#L45)

DavideIadeluca:
Yep, seems like it was exactly what you mentioned – Thanks for looking into this. Will PR `1.x` with the same fix

SychO9 (approved):
Thanks!

PR #4127, opened 2024-11-23
[tags] Wrong tag input width if contains CJK characters

YUCLing:
**Fixes same issue mentioned in #4094**

Same as #4094, but in a much better and lighter approach.

**Screenshot**

**Necessity**

**Confirmed**

**Required changes:**


SychO9 (approved):
Thanks!

Issue #3533, opened 2020-10-02
Convert core JS to Typescript


askvortsov1:
This should be done in small, maximally isolated PRs: no more than a few files updated per PR so that we can slowly review and phase these in over time. Community contributions appreciated!

For overall progress, see the GitHub Project board: https://github.com/orgs/flarum/projects/21

sweetrush:
Hi Guys and Girls , just join this wonderful collab , is there a specific channel for this on discord , just would like to know which files everyone might be working on so not so double do ..

luceos:
I saw your message on discord @sweetrush. Internals would be the right channel 👍

Niveditha-K17:
Hello maintainers!
I'm new to Typescript but I'm eager to learn and would love to contribute.
Is there any issue I can take up?

afunctional:
Hi everyone! I'm in the same @Niveditha-K17 situation, i want to contribute, it would be a fantastic way to improve my Typescript skills. Still need PRs?

luceos:
> Hi everyone! I'm in the same @Niveditha-K17 situation, i want to contribute, it would be a fantastic way to improve my Typescript skills. Still need PRs?

Yeah, there's still plenty to convert. See for instance this path: https://github.com/flarum/framework/tree/2.x/framework/core/js/src/forum/components

afunctional:
Ok, thank you. Can I take files freely or do I have to warn/agree with someone? I read that there was talk of agreeing on Discord but not being on the server I don't know

dk5488:
Hey everyone! Is there anything left to convert?? would love to help

produdez:
Hi guys, i'm also interested in helping if there's any TS conversion work needed? Couldn't access the project board link  https://github.com/orgs/flarum/projects/21.

PR #4130, opened 2024-11-24
feat: add support for customizable user avatar drivers

datlechin:
This pull request introduces a new avatar driver system. The changes include adding support for different avatar drivers, updating the user model to accommodate these changes, and modifying relevant components to utilize the new avatar driver system.

**Fixes #0000**

**Screenshot**

**Necessity**

**Confirmed**

**Required changes:**

- [ ] Related documentation PR: https://github.com/flarum/docs/pull/498


imorland:
Hey @datlechin sorry this has been dormant for so long. Do you have any interest in continuing user avatar drivers? If so, please would you rebase and resolve the conflicts here?

I'd really like to include this before the 2.0 stable release if possible 🙏

datlechin:
@imorland, I’ve completed the work on this PR. Should I include the Gravatar driver along with the default driver in the core?

imorland (commented):
Nice one @datlechin this looks good to me.

Yes, let's include gravatar support in core. I think it would be used by some communities, and admins can always choose to switch from 'classic' to gravatar as they wish. I don't think the overhead would be too much here.

I think it best to add that as a seperate PR tho, rather than in this one.

Before I merge this, would you please create an counterpart PR to https://github.com/flarum/docs so that developers can see how to use it?

datlechin:
@imorland I’ve opened the pull request in [flarum](https://github.com/flarum/docs/pull/498). I’ll also open another pull request to add Gravatar support to the core when this PR was merged

DavideIadeluca (commented):
Just jumping in here. Generally, I really like the idea of custom avatar drivers and would like to see it for 2.x. I implemented a Gravatar driver for testing purposes locally, and everything is working as expected.

One thing that stood out to me is that two values related to the avatar URL are now serialized (`avatarUrl` and `originalAvatarUrl`). I haven’t dug into the details yet as to why this was done, but I generally think we’d be better off serializing just one value to the frontend and letting the backend handle the rest.

Besides this, lgtm

framework/core/src/User/User.php
```diff
@@ -111,7 +112,12 @@ class User extends AbstractModel
     /**
      * A driver for getting display names.
      */
-    protected static DriverInterface $displayNameDriver;
+    protected static DisplayNameDriver $displayNameDriver;
+
```
imorland, line 116: `$this->attributes['avatar_url']` will throw an `ErrorException: Undefined array key "avatar_url"` if the model was loaded without that column (e.g. `User::select('id')->find(1)`). The Eloquent accessor parameter (`?string $value = null`) handles this safely. Use:

```php
return $this->attributes['avatar_url'] ?? null;
```

framework/core/src/Api/Resource/UserResource.php
```diff
@@ -243,6 +243,7 @@ public function fields(): array
                 ->save(fn () => null),
             Schema\Str::make('displayName'),
             Schema\Str::make('avatarUrl'),
+            Schema\Str::make('originalAvatarUrl'),
```
imorland: `originalAvatarUrl` is exposed to all API consumers with no visibility constraint. This reveals whether a user has a real uploaded avatar vs a driver-generated one — an internal implementation detail. Consider whether this field is necessary in the public API, or whether a boolean `hasCustomAvatar` would serve the `AvatarEditor` use case with a smaller surface area. If kept, it should at minimum not be serialised for guest/other-user requests where the raw storage path has no value.

framework/core/js/src/admin/components/BasicsPage.tsx
```diff
@@ -58,6 +59,9 @@ export default class BasicsPage<CustomAttrs extends IPageAttrs = IPageAttrs> ext
       display_name: {
         username: extractText(app.translator.trans('core.admin.basics.display_name_driver_options.username')),
       },
+      avatar: {
+        default: extractText(app.translator.trans('core.admin.basics.avatar_driver_options.default')),
+      },
```
imorland, line 64: `driverLocale()` is a static method that only knows about the built-in `'default'` label. A third-party extension registering an avatar driver via `(new Extend\User)->avatarDriver('gravatar', ...)` has no way to register a human-readable label — it falls back to the raw identifier string. The `displayNameDriver` system has the same gap, but it would be good to at least document this so extension authors know to expect it.

framework/core/src/User/UserServiceProvider.php
```diff
@@ -84,7 +87,30 @@ protected function registerDisplayNameDrivers(): void
                 : $container->make(UsernameDriver::class);
         });
 
-        $this->container->alias('flarum.user.display_name.driver', DriverInterface::class);
+        $this->container->alias('flarum.user.display_name.driver', DisplayNameDriverInterface::class);
+    }
+
```
imorland, line 92: The fallback `$driverName = $settings->get('avatar_driver', '')` uses an empty string as the default. An empty string is not the same as `'default'`, and `Arr::get($drivers, '')` will always return `null`, falling through to `AvatarDefaultDriver`. This works, but it would be clearer and more self-documenting to use `$settings->get('avatar_driver', 'default')` so the intent is explicit.

imorland (changes requested):
This is a well-structured feature that follows established Flarum patterns (mirrors `displayNameDriver` closely). The architecture is sound and it would enable Gravatar, identicons, and similar drivers as first-class extensions. Several things need to be addressed before merge.

**Design question: does `originalAvatarUrl` belong in the public API?**
Exposing `originalAvatarUrl` as a serialised API field is a significant API surface addition. Its only purpose in this PR is to let `AvatarEditor` know whether the user has a real uploaded avatar (to show the pencil vs plus-circle icon). This is an implementation detail that leaks through to every API consumer. Consider instead returning a boolean field (e.g. `hasCustomAvatar`) or resolving this client-side by comparing `avatarUrl` to what a driver would produce. If `originalAvatarUrl` must be exposed, it should have the same visibility constraints as the existing `avatarUrl` field.

**`AvatarEditor.js` should be `.tsx`**
The three changed lines in `AvatarEditor.js` are the only changes to that file. Since this is a JS → TS migration project (#3533), this file could be converted at the same time. Not a blocker, but a missed opportunity.

**`driverLocale()` is not extensible for third-party driver labels**
`BasicsPage.driverLocale()` is a `static` method that hardcodes the known driver labels. When a Gravatar extension registers an `avatarDriver('gravatar', GravatarDriver::class)`, it has no way to register a human-readable label — it will fall back to the raw identifier string (`'gravatar'`). The existing `displayNameDriver` pattern has the same gap, but this PR makes it worse by adding another driver type. Consider documenting this limitation or providing an extension point (e.g. a static `avatarDriverLocale` map that extensions can populate).

**`getOriginalAvatarUrlAttribute` array access**
`$this->attributes['avatar_url']` will throw an `ErrorException` (undefined array key) if the key is not present in the attributes array, e.g. when the model is freshly instantiated without all columns loaded. The existing `getAvatarUrlAttribute` receives `?string $value = null` from Eloquent (which handles absent keys safely). Consider using `$this->attributes['avatar_url'] ?? null` instead.

**Fallback when `avatar_driver` setting holds an unregistered identifier**
In `UserServiceProvider`, when `Arr::get($drivers, $driverName)` returns `null` (unknown driver name in settings), the code falls back to `AvatarDefaultDriver`. That's reasonable, but it silently ignores a misconfiguration. This matches `displayNameDriver` behaviour, so acceptable — just worth noting.

**No default value registered for `avatar_driver` setting**
The `Settings` extender is not used to register a default for `avatar_driver`. Calling `$settings->get('avatar_driver', '')` with an empty-string default works, but it's inconsistent — an empty string is not the same as `'default'`. If a driver named `''` were ever registered, this would resolve incorrectly. Using `'default'` as the fallback default would be clearer.

imorland (approved):

PR #4132, opened 2024-11-29
feat: separate mariadb driver

SychO9:
**Fixes #0000**

* Fixes checking of MariaDB in the migrations to using `->isMaria()`
* Adds `mariadb` as a separate database driver, laravel 11 has [done the same](https://github.com/laravel/framework/pull/50146) in order to more easily handle the growing differences between mysql and mariadb.
* This will not break things for instances using mariadb under the mysql driver. But we need to document switching to mariadb driver nonetheless.

**Screenshot**
image

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/.github/workflows/REUSABLE_backend.yml
+++ b/.github/workflows/REUSABLE_backend.yml
@@ -94,7 +94,7 @@ jobs:
             driver: mysql
           - service: mariadb
             db: MariaDB
-            driver: mysql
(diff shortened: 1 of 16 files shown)


luceos:
@SychO9 feel free to merge. I will do a test run early this week on Nightly (if it doesn't already auto fix with this).

PR #4111, opened 2024-11-13
Remove jQuery & Bootstrap

YUCLing:
_This is still an ongoing work and brings huge breaking changes. Quality check and feedback are required before final merge._

**Completes https://github.com/orgs/flarum/projects/22?pane=issue&itemId=5300309**

**Changes proposed in this pull request:**
This PR removes jQuery and Bootstrap completely from Flarum, which can greatly reduce the frontend size.

**Reviewers should focus on:**

- [ ] Are all jQuery usages removed?
- [ ] Are all Bootstrap usages removed?
- [ ] Do vanilla implementations match the original jQuery/Bootstrap behaviors?
  - nullable element variable and zero-length jQuery objects
  - select multiple elements or single element
  - jQuery `css` returns **computed** style value

**Necessity**

- [ ] Has the problem that is being solved here been clearly explained?
- [ ] If applicable, have various options for solving this problem been considered?
- [ ] For core PRs, does this need to be in core, or could it be in an extension?
- [ ] Are we willing to maintain this for years / potentially forever?

**Components using jQuery**

<details>
<summary>Core</summary>

_Common_
- [x] `components/AutocompleteDropdown`
- [x] `components/ConfirmDocumentUnload`
- [x] `components/EditUserModal`
- [x] `components/FormModal`
- [x] `components/SearchModal`
- [x] `components/TextEditor`
- [x] `components/UploadImageButton`
- [x] `helpers/highlight`
- [x] `utils/anchorScroll`
- [x] `utils/GambitsAutocomplete`
- [x] `utils/KeyboardNavigatable`

_Forum_
- [x] `ForumApplication`
- [x] `components/AbstractPost`
- [x] `components/AffixedSidebar`
- [x] `components/AvatarEditor`
- [x] `components/CommentPost`
- [x] `components/Composer`
- [x] `components/HeaderList`
- [x] `components/LogInButton`
- [x] `components/NotificationGrid`
- [x] `components/PostMeta`
- [x] `components/PostsPage`
- [x] `components/PostStream`
- [x] `components/PostStreamScrubber`
- [x] `components/ReplyPlaceholder`
- [x] `components/Search`
- [x] `components/SignUpModal`
- [x] `components/WelcomeHero`
- [x] `states/ComposerState`

_Admin_
- [x] `components/AdminNav`
- [x] `components/CreateUserModal`
- [x] `components/UserListPage`

</details>

<details>
<summary>Extensions</summary>

_embed_

- forum
  - [x] `index`

_emoji_

- forum
[...]

PR #4142, opened 2024-12-17
Fix missing affixed sidebar offset

### LINKED ISSUE #4122: [2.x] Side nav positioning offset missing
### Current Behavior

With AffixedSidebar, the side nav uses `position: fixed` for side nav's `ul` to make it fixed on viewport. However, in #3876, the `margin-top` of `ul` is moved to the AffixedSidebar's parent element `Page-sidebar`.

From the `top` value used below, we can see it only includes the height of header, so the side nav is at a taller position than 1.x.

https://github.com/flarum/framework/blob/4feb4a3820a6c6e3c13fd7034ae583a628d86b30/framework/core/less/common/sideNav.less#L66

### Steps to Reproduce

1. Find a page with a side nav using AffixedSidebar
2. Scroll down

### Expected Behavior

The missing offset provided by `margin-top` should be still there.

### Screenshots

image

### Environment

- Flarum version: 2.0-dev

_No response_

### Possible Solution

Backport https://github.com/flarum/framework/pull/4111/commits/11fc74db3f6e76dfcaa2fabdb3213e809e31ddb5 and https://github.com/flarum/framework/pull/4111/commits/ea21ef4c23ac8c7f9541b69905f1af3b69bc8619, which uses `position: sticky` and adds the missing offset.

### Additional Context

_No response_

YUCLing:
**Fixes #4122**

Switched to sticky positioning and fixed the missing offset by backporting #4111. Bootstrap's affix plugin is kept for compatibility.

**Screenshot**
a27d080cab47a1f7239403c954a823fe

**Necessity**

**Confirmed**

**Required changes:**


SychO9 (commented):
Hi @YUCLing thanks for the PR!

Looking at the issue, it seems easier to just add the needed margin when the nav is fixed in place: https://github.com/flarum/framework/pull/4147

YUCLing:
> Looking at the issue, it seems easier to just add the needed margin when the nav is fixed in place: #4147

Well, this also switches to a more modern way to make the nav fixed. Still, it's up to you to decide which way to take.

SychO9:
I'd definitely like to see us replacing jquery/bootstrap, but since that won't happen in the 2.x line anyway, my preference is simpler changes especially as we're looking to stabilize 2.0 right now.

Still, this will be a great reference for v3 when we get to dropping jquery and bootstrap!

Closing for now. Thanks again!

PR #4145, opened 2024-12-31
wip

luceos:
- [ ] create different renderers so we can render Info output in CLI and web
- [x] add middleware in admin to understand web versus cli user
- [x] add middleware in admin to understand opcache status as this is different (always disabled) in cli
- [x] removed translation on scheduler status


imorland:
Thanks @luceos — revisiting this one.

The core goals here landed in 2.x through `ApplicationInfoProvider`, now shared by both the CLI `InfoCommand` and the admin dashboard (`AdminPayload` surfaces PHP version, DB driver/version + driver mismatch, DB options, debug state and scheduler status on the web side). Web-vs-CLI differences are handled on the provider itself (`identifySessionDriver($forWeb)`, the web-friendly session driver name from #3690), so we get CLI + web output from one source without the separate renderer/section abstraction proposed here.

Given that, and that this branch is a WIP that's drifted well behind `2.x`, I'll close it as superseded.

The only piece not carried over is opcache **status reporting** in the info output (real status on web vs. always-disabled under CLI). Note this is distinct from #4569, which covers opcache *flushing* after package operations — different concern. If accurate opcache-status reporting is still wanted, it'd be a small standalone PR against the current `ApplicationInfoProvider`. Thanks for the groundwork here!

PR #4161, opened 2025-01-17
[2.x] Codebase cleanup

xHeaven:
This PR aims to cleanup the codebase as well as utilize the latest technologies provided by PHP up to 8.2.

It's a pretty big one, but I've cut everything up into lots of commits, so if you don't like something, it's easily revertable.

There are no breaking changes here. Code style will have to be updated by StyleCI(?) - though, I wish there was php-cs-fixer or something like that included in the project, so I could properly adhere to the coding styles without thinking about it too much.


MrSuddenJoy:
This code looks good to me. Well done 👍🏻

MrSuddenJoy (approved):
This code looks good to me. Well done 👍🏻

framework/core/src/Api/Middleware/ThrottleApi.php
```diff
@@ -42,7 +42,9 @@ public function throttle(Request $request): bool
             // Anything else is ignored.
             if ($result === false) {
                 return false;
-            } elseif ($result === true) {
+            }
+
+            if ($result === true) {
                 $throttle = true;
             }
```
luceos, line 49: A `match($result)` would probably be cleaner.

framework/core/src/Api/Resource/AccessTokenResource.php
```diff
@@ -79,18 +79,18 @@ public function fields(): array
     {
         return [
             Schema\Str::make('token')
-                ->visible(function (AccessToken $token, Context $context) {
+                ->visible(function (#[\SensitiveParameter] AccessToken $token, Context $context) {
```
luceos, line 82: I learned something today, good use of attributes 🙏 

I don't see a way to mark the whole class as sensitive, which - in terms of code clarity - would have been much better, especially when extensions usually don't adhere to best practices.

framework/core/src/Discussion/Search/FulltextFilter.php
```diff
@@ -42,7 +42,6 @@ public function search(SearchState $state, string $value): void
 
     protected function sqlite(DatabaseSearchState $state, string $value): void
     {
-        /** @var Builder $query */
```
luceos, line 45: These are actually IDE helpers, so that we know exactly which type of Builder (eloquent versus query) is given. Is this now implied from the `$state->getQuery()` instead?

framework/core/src/Forum/Controller/RegisterController.php
```diff
@@ -32,7 +32,7 @@ public function handle(Request $request): ResponseInterface
 
         $response = $this->api->withParentRequest($request)->withBody($params)->post('/users');
 
-        $body = json_decode($response->getBody());
+        $body = json_decode($response->getBody(), false);
```
luceos: I would prefer seeing named arguments here as well, if you add it elsewhere.

framework/core/src/Frontend/FrontendServiceProvider.php
```diff
@@ -141,7 +141,7 @@ function (Container $container) {
         $this->container->singleton(
             'flarum.frontend.custom_less_functions',
             function (Container $container) {
-                $extensionsEnabled = json_decode($container->make(SettingsRepositoryInterface::class)->get('extensions_enabled'));
+                $extensionsEnabled = json_decode($container->make(SettingsRepositoryInterface::class)->get('extensions_enabled'), true);
```
luceos: Named arguments 🙏

luceos (changes requested):
That's is fantastic work, wow. Gigantic effort to go over all these files, did you do all of this manually or use some kind of refactoring utility?

I left a few remarks here and there, but overall I could not find any area where changes would negatively impact the application.

framework/core/src/Api/Middleware/ThrottleApi.php
xHeaven: Something like this?
```php
$throttle = match ($result) {
    false => false,
    true => true,
    default => $throttle,
};
```

Note that, currently we instantly return on the first `false` encounter, but if we get `true` (or anything other than `true`/`false`), we continue the loop. We can't return in a match expression, though, so we would be forced to continue the loop even on `false` result.

framework/core/src/Api/Resource/AccessTokenResource.php
xHeaven: Unfortunately, there is no way currently to mark whole classes as sensitive - in this case, I agree it'd be a lot cleaner, however, it's not possible right now.

framework/core/src/Discussion/Search/FulltextFilter.php
xHeaven: That's right, `DatabaseSearchState`'s `getQuery()` method [has a clear return type](https://github.com/flarum/framework/blob/2.x/framework/core/src/Search/Database/DatabaseSearchState.php#L24) of `Illuminate\Database\Eloquent\Builder` (therefore, it can't be anything else), which is inferred just fine by IDEs, so there is no need to specify it explicitly anymore.

xHeaven:
> That's is fantastic work, wow. Gigantic effort to go over all these files, did you do all of this manually or use some kind of refactoring utility?
> 
> I left a few remarks here and there, but overall I could not find any area where changes would negatively impact the application.

It was a completely manual effort, except for a few small things that were reported by my IDE itself. It certainly took me a bit of time, though, due to the size of the project, but I hope it was worth it. ✨

luceos (approved):

PR #4180, opened 2025-02-08
feat(pm): delete own messages

SychO9:
**Fixes #0000**

* Adds the ability/permission to delete one's own messages.


GeorPe:
If you delete the first message in the thread, the remaining messages will always remain loading.

flarum (2)

flarum

luceos:
> If you delete the first message in the thread, the remaining messages will always remain loading.
> flarum (2) flarum

Please, if you can isolate this issue from the logs and post them on discuss we can take action: https://discuss.flarum.org/t/support, thank you.

Issue #4060, opened 2024-10-08
Port 1.8.x patch changes to 2.x


imorland:
See https://github.com/flarum/framework/milestone/41?closed=1

Or: https://github.com/flarum/framework/commits/1.x/?since=2024-09-01&until=2024-12-06

PR #4186, opened 2025-02-10
feat: make it easier to add content after the first post

DavideIadeluca:
2.x port of https://github.com/flarum/framework/pull/4050

**Progresses #4060**

**Screenshot**

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/framework/core/js/src/forum/components/PostStream.js
+++ b/framework/core/js/src/forum/components/PostStream.js
@@ -84,7 +84,7 @@ export default class PostStream extends Component {
         </div>
       );
 
-      if (post && post.number() === 1 && this.afterFirstPostItems().toArray().length > 0) {
+      if (post && post.id() === this.discussion.data.relationships.firstPost?.data.id && this.afterFirstPostItems().toArray().length > 0) {
         return m.fragment({ ...attrs }, [
           postStreamElement,
           <div className="PostStream-item PostStream-afterFirstPost" key="afterFirstPost">
--- a/framework/core/js/src/forum/components/PostStream.js
+++ b/framework/core/js/src/forum/components/PostStream.js
@@ -84,7 +84,7 @@ export default class PostStream extends Component {
         </div>
       );
 
-      if (i === 0 && this.afterFirstPostItems().toArray().length > 0) {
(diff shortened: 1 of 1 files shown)


framework/core/js/src/forum/components/PostStream.js
```diff
@@ -78,11 +78,22 @@ export default class PostStream extends Component {
...
       }
 
-      return (
+      const postStreamElement = (
         <div className="PostStream-item" {...attrs}>
           {content}
         </div>
       );
+
+      if (i === 0 && this.afterFirstPostItems().toArray().length > 0) {
```
SychO9: is the intent after the first post of the discussion, or after the first of the visible posts? because i === 0 will always be the first of the currently visible posts, not necessarily the first post of the discussion.

For example, if you are viewing posts 200 - 220 of a discussion, the 200th is the first (i === 0)

DavideIadeluca: Yes that was most certainly the intention. Pushed a change which works much better now

framework/core/js/src/forum/components/PostStream.js
```diff
@@ -78,11 +78,22 @@ export default class PostStream extends Component {
...
       }
 
-      return (
+      const postStreamElement = (
         <div className="PostStream-item" {...attrs}>
           {content}
         </div>
       );
+
+      if (post && post.number() === 1 && this.afterFirstPostItems().toArray().length > 0) {
```
SychO9: I don't trust the first post number to always be 1, a more certain check would be using the discussion's first_post_id value, can we serialize that then use it hear?

SychO9: Actually in 2.0 the relationship is always "linked" so no need for extra serialization, we can do:

```suggestion
      if (post && post.id() === this.discussion.data.relationships.firstPost?.data.id && this.afterFirstPostItems().toArray().length > 0) {
```

DavideIadeluca:
@SychO9 seems like `phpstan/phpstan` v1.12.18 (released yesterday) has broken CI here. Might you be able to look into this? Don't think it's a GitHub quirk this time around

SychO9 (approved):
Thanks!

---

Don't worry about the phpstan failure, will look into that separately.

PR #4199, opened 2025-02-17
feat: improve extensibility of `WelcomeHero`

DavideIadeluca:
2.x port of https://github.com/flarum/framework/pull/4039

**Progresses #4060**

**Screenshot**

**Necessity**

**Confirmed**

**Required changes:**


SychO9 (commented):
I wonder if introducing a reusable hero component would be even better for extensibility/theming

it could be composition based: <Hero title={welcomeTitle} content={...} dismissible={true} />

or something like that? doesn't necessarily have to be done in this PR, but would love the feedback on that before 2.0 goes stable

DavideIadeluca:
Yeah, I think it would certainly be worth it to have a go and see how it goes. Would prefer if this is done in a separate PR though

SychO9 (approved):

PR #4200, opened 2025-02-18
refactor(core): improve extensibility of Admin Pages

DavideIadeluca:
2.x port of #4037

**Progresses #4060**

**Screenshot**

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/framework/core/js/src/admin/components/AppearancePage.tsx
+++ b/framework/core/js/src/admin/components/AppearancePage.tsx
@@ -22,57 +22,91 @@ export default class AppearancePage extends AdminPage {
   }
 
   content() {
-    return (
-      <>
-        <Form>
-          <FieldSet
-            className="AppearancePage-colors"
-            label={app.translator.trans('core.admin.appearance.colors_heading')}
-            description={app.translator.trans('core.admin.appearance.colors_text')}
-          >
-            {this.colorItems().toArray()}
-          </FieldSet>
-        </Form>
+    return this.contentItems().toArray();
+  }
 
-        <Form>
-          <div className="Form-group">
-            <label>{app.translator.trans('core.admin.appearance.logo_heading')}</label>
-            <div className="helpText">{app.translator.trans('core.admin.appearance.logo_text')}</div>
-            <UploadImageButton name="logo" routePath="logo" value={app.data.settings.logo_path} url={app.forum.attribute('logoUrl')} />
-          </div>
+  contentItems(): ItemList<Mithril.Children> {
+    const items = new ItemList<Mithril.Children>();
 
-          <div className="Form-group">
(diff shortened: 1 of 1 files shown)


DavideIadeluca:
@SychO9 Before continuing with this PR I need some clarification on how to proceed here. I already refactored the `AppearancePage` to use multiple ItemLists. I now saw that there was a refactoring done into how settings are defined (#4022).

Mainly the `BasicsPage` has been refactored to leverage the `AdminRegistry`. Wrapping those settings in a ItemList seem superfluous to me as the settings are already an `ItemList` behind the scenes.

Stepping back for a bit – here are two use cases why those extensibility improvements in the Admin Frontend are useful and how they are used in 1.x:

1. In one of our proprietary extensions we allow making the forum title translatable. We replace the original input field and add a button instead with the functionality to add translations for each language. For 1.x this is done like so:
    ```tsx
    extend(BasicsPage.prototype, 'contentItems', function (items) {
		items.setContent(
			'forum-title',
			<div className='Form-group'>
				<label>{app.translator.trans('core.admin.basics.forum_title_heading')}</label>
				<TranslationKey translation={...} />
			</div>,
		);
	});
    ```
2. In a community, the Welcome Hero is not used at all. To prevent confusion, we removed the Welcome Banner setting:
    ```tsx
    extend(BasicsPage.prototype, 'contentItems', function (items) {
		items.remove('welcome-banner');
	});
    ```

Right now, I have 2 questions:

1. How should we cover those use cases of customizing and removing core settings in a developer-friendly way? Perhaps we introduce new methods in the `AdminRegistry`
2. Should/will the `AppearancePage` eventually be refactored to leverage the app registry (mainly thinking of the `theme_primary_color` and `'theme_secondary_color'` here.

Thoughts?

SychO9:
> How should we cover those use cases of customizing and removing core settings in a developer-friendly way? Perhaps we introduce new methods in the AdminRegistry

I think adding `setSetting` and `removeSetting` (and same for permissions perhaps) would make sense.

> Should/will the AppearancePage eventually be refactored to leverage the app registry (mainly thinking of the theme_primary_color and 'theme_secondary_color' here.

Maybe, maybe not, the reason I left it the way it is, is because of how custom the structure of the appearence settings are (color items, theme modes, ..etc) unlike other pages where settings are directly listed. It'll probably stay the way it is i think 🤷🏼

SychO9:
> I think adding setSetting and removeSetting (and same for permissions perhaps) would make sense.

Addressed in https://github.com/flarum/framework/pull/4209

This PR should be good to go

DavideIadeluca:
Thanks for taking care of this!

Also good to go from my side

SychO9 (approved):
Thanks!

PR #4211, opened 2025-04-18
chore(a11y): misc a11y improvements

SychO9:
**Part of the roadmap**

Attempts to address as much of the a11y audit results as possible.

### DIFF
--- a/framework/core/js/src/common/utils/a11y.ts
+++ b/framework/core/js/src/common/utils/a11y.ts
@@ -6,7 +6,6 @@ export function prepareSkipLinks() {
   document.querySelectorAll('.sr-only-focusable-custom:not([data-prepared])').forEach((el) => {
     el.addEventListener('click', function (e) {
       e.preventDefault();
-      console.log('intercepted');
       const target = el.getAttribute('href')!;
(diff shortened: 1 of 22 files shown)


framework/core/js/src/common/utils/a11y.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * Fix a11y skip links by manually focusing on the href target element.
+ * This prevents unwanted/unexpected reloads of the page.
+ */
+export function prepareSkipLinks() {
+  document.querySelectorAll('.sr-only-focusable-custom:not([data-prepared])').forEach((el) => {
+    el.addEventListener('click', function (e) {
+      e.preventDefault();
+      console.log('intercepted');
```
luceos: The console.log needs to be removed.

luceos (changes requested):

framework/core/js/src/common/utils/a11y.ts
SychO9: nice catch 🙈

SychO9:
```suggestion
```

Issue #3692, opened 2022-11-24
A problem with tag permission for user.


MIKU-N:
### Current Behavior

I get a problem with some permission about tag.
When I set permission by tag,although the global permission is set for Indefinitely,but tag permission is null,and I can't edit this permission.User also can't get this permission.like rename post himself,delete post himself on this tag.

### Steps to Reproduce

1.Create a Tag
2.Create a permission rule by this tag,set global permission `Allow renaming`  to  `Indefinitely`
3.Login with a default users,and create a post with this tag.
4.Now,the user can't edit himself post title,and he can't delete this post.

### Expected Behavior

I think all tag permission maybe can edit,If tag permissions not set ,The tag permissions must inherit global permissions.
This rule is also same in `Restrict by Tag`.When admin set the tag permission,the permissions of 'null' must inherit global permissions.

### Environment

- Flarum version: 1.6.1,Core Version 1.6.2
- Website URL:  All flarum have this problem
- Webserver: Nginx
- Hosting environment: 
- PHP version: 8.1.7

### Output of `php flarum info`

```
Output of "php flarum info", run this in terminal in your Flarum directory.
```
Flarum core 1.6.2
PHP version: 8.1.7
MySQL version: 8.0.24
Loaded extensions: Core, date, libxml, openssl, pcre, sqlite3, zlib, bcmath, ctype, curl, dom, filter, ftp, gd, gettext, hash, iconv, intl, json, mbstring, SPL, session, pcntl, standard, mysqlnd, PDO, pdo_mysql, pdo_sqlite, Phar, posix, Reflection, mysqli, shmop, SimpleXML, soap, sockets, sodium, sysvsem, tokenizer, xml, xmlreader, xmlwriter, zip, fileinfo, redis, exif, imagick, gmp, memcached, Zend OPcache
+-------------------------------------+--------------+--------+
| Flarum Extensions                   |              |        |
+-------------------------------------+--------------+--------+
| ID                                  | Version      | Commit |
+-------------------------------------+--------------+--------+
| flarum-flags                        | v1.6.1       |        |
| flarum-approval                     | v1.6.1       |        |
| flarum-tags                         | v1.6.1       |        |
| flarum-suspend                      | v1.6.1       |        |
| askvortsov-auto-moderator           | dev-master   |        |
| flarum-sticky                       | v1.6.0       |        |
| flarum-nicknames                    | v1.6.0       |        |
| flarum-lock                         | v1.6.0       |        |
[...]

PR #4214, opened 2025-05-16
Fixed issue #3692

edgeinfinity1:
**Fixes #0000**

Fixing an issue causing users unable to rename their posts, which has been confirmed at https://github.com/flarum/framework/issues/3692

This is a rather brutal fix, since I'm not too into flarum. Please consider rewriting this in an elegant way.
At least I have confirmed it working on my own deployment at https://forum.monika.love

**Screenshot**

**Necessity**

**Confirmed**


edgeinfinity1:
- Fixed another issue where users cannot delete their own threads while tags enabled. It's the same cause.

imorland:
Hi @edgeinfinity1 sorry it's taken us so long here 🙈 

I've taken your work here as a starting point and tried to implement a little more egegantly as you suggest in #4379 

I'll close this PR now and focus on the new one. Thanks for this, really appreciate it!

PR #4215, opened 2025-05-21
[2.x] feat: create reusable `Hero` component

### LINKED ISSUE #4202: Create Reusable `Hero` Component
Follow up of https://github.com/flarum/framework/pull/4199#pullrequestreview-2631069300

### Current Behavior

There are a multitude of `Hero` like components & Elements across the codebase. Examples:

- `WelcomeHero`
- `DiscussionHero`
- `TagHero`
- `MessagesPage` hero

### Steps to Reproduce

N/A

### Expected Behavior

For extensibility purposes, it would be worth it to try to create a single reusable `Hero `component which can be extended

### Screenshots

_No response_

### Environment

N/A

_No response_

### Possible Solution

_No response_

### Additional Context

https://github.com/flarum/framework/pull/4199#pullrequestreview-2631069300

DavideIadeluca:
**Fixes #4202**

- Create new reusable hero component to streamline the implementation of each variation

- Tried to keep the core abstract implementation as lightweight as possible
- Perhaps some methods or the new class itself could be named better
- Does this change need to be documented?
- Are there any other Hero implementations i've missed?

**Screenshot**
_DOM / UI should stay the same_

**Necessity**

**Confirmed**

**Required changes:**


DavideIadeluca:
FYI @imorland I just made a minor change I've had in mind. This PR is now good to go from my side

framework/core/js/src/forum/components/Hero.tsx
```diff
@@ -0,0 +1,70 @@
...
+   *     backgroundColor: '#e7672e',
+   *   };
+   * }
+   * ```
+   */
+  style(): Record<string, string> | undefined {
+    return undefined;
+  }
+
+  view(): Mithril.Vnode | null {
```
imorland: The base `view()` never returns `null` — that only happens in `WelcomeHero`'s override. Typing the base as `Mithril.Vnode | null` is misleading. Please use `JSX.Element` to be consistent with other view methods in core (e.g. `HeaderPrimary.view(): JSX.Element`).

framework/core/js/src/forum/components/Hero.tsx
```diff
@@ -0,0 +1,70 @@
...
+    return (
+      <header className={classList('Hero', this.className())} style={this.style() ?? undefined}>
```
imorland: `style()` already returns `Record<string, string> | undefined` — it can never return `null`, so `?? undefined` is a no-op. Simplify to:

```tsx
<header className={classList('Hero', this.className())} style={this.style()}>
```

framework/core/js/src/forum/components/Hero.tsx
```diff
@@ -0,0 +1,70 @@
...
+        {this.viewItems().toArray()}
+      </header>
+    );
+  }
+
+  viewItems(): ItemList<Mithril.Children> {
```
imorland, line 83: `viewItems()` is the primary structural extension point — the method third-party extensions would `extend()` to wrap or replace the `.container` element (e.g. adding a background image or overlay layer outside the container). It has no JSDoc while `className()` and `bodyItems()` both have thorough documentation with examples. Please add equivalent documentation here.

framework/core/js/src/forum/components/WelcomeHero.tsx
```diff
@@ -1,37 +1,29 @@
 import app from '../app';
-import Component from '../../common/Component';
+import Hero, { IHeroAttrs } from './Hero';
 import Button from '../../common/components/Button';
 import type Mithril from 'mithril';
 import ItemList from '../../common/utils/ItemList';
 
-export interface IWelcomeHeroAttrs {}
+export interface IWelcomeHeroAttrs extends IHeroAttrs {}
```
imorland, line 8: **Breaking change notice needed.** The current published `dist-typings/forum/components/WelcomeHero.d.ts` exposes `viewItems()` as a public method. Any third-party extension using `extend(WelcomeHero.prototype, 'viewItems', ...)` will silently stop firing after this rename.

Same applies to `TagHero` — its dist-typings also expose `viewItems()`.

For a 2.x pre-release this is acceptable, but please:
1. Add an explicit note to the PR description calling this out
2. Ensure it is captured in the upgrade guide

extensions/tags/js/src/forum/components/TagHero.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+import type Tag from '../../common/models/Tag';
+import type Mithril from 'mithril';
+
+export interface ITagHeroAttrs extends IHeroAttrs {
+  model: Tag;
+}
+
+export default class TagHero<CustomAttrs extends ITagHeroAttrs = ITagHeroAttrs> extends Hero<CustomAttrs> {
+  className(): string {
```
imorland, line 15: `className()` returns a space-separated string like `'TagHero TagHero--colored dark-on-dark'`, which is then passed into `classList('Hero', this.className())` in `Hero.view()`.

This works today because Flarum uses clsx `^1.1.1`, and clsx 1.x splits space-separated strings. However, clsx 2.x (a potential future upgrade) does **not** handle space-separated strings, so this would silently break.

A safer approach is to override `view()` in `TagHero` directly:

```tsx
className(): string {
  return 'TagHero';
}

view() {
  const color = this.attrs.model.color();
  return (
    <header
      className={classList('Hero', 'TagHero', { 'TagHero--colored': color, [textContrastClass(color)]: color })}
      style={this.style()}
    >
      {this.viewItems().toArray()}
    </header>
  );
}
```

This makes the conditional class logic explicit and safe across clsx versions.

extensions/messages/js/src/forum/components/MessagesPageHero.tsx
```diff
@@ -0,0 +1,38 @@
+import app from 'flarum/forum/app';
+import Hero, { IHeroAttrs } from 'flarum/forum/components/Hero';
+import Icon from 'flarum/common/components/Icon';
+import ItemList from 'flarum/common/utils/ItemList';
+
+import type Mithril from 'mithril';
+
+export interface IMessagesPageHeroAttrs extends IHeroAttrs {}
```
imorland, line 8: Empty interface extending another is fine for extensibility, but a brief comment would help clarify the intent to readers:

```tsx
// Intentionally empty — extend this interface to add attrs for extensions.
export interface IMessagesPageHeroAttrs extends IHeroAttrs {}
```

imorland (changes requested):
Thanks for this — the design is solid and consistent with the existing `Modal` pattern. This PR has been open for a while but I checked: the Laravel upgrade, LESS 5.x upgrade, and webpack changes that landed since opening do not affect any of this, so it is not stale. A few things to address before merge.

**Summary of required changes:**

- Add `import './components/Hero'` to `forum.ts` — explicit registration is the established convention and important for discoverability as a first-class extension point
- Add JSDoc to `viewItems()` — it is the primary structural extension point but has no documentation
- Remove `?? undefined` from `style={}` in `Hero.view()` — redundant since `style()` already returns `undefined`, not `null`
- Fix `Hero.view()` return type from `Mithril.Vnode | null` to `JSX.Element` — the base never returns `null`
- Document the `viewItems()` → `bodyItems()` rename as a breaking change in the PR description and upgrade guide
- Add a note about `UserPage`'s `UserHero` (via `UserCard`) being intentionally excluded

**Also worth considering:**
- The `TagHero.className()` approach of returning a space-separated string passed to `classList()` works with clsx 1.x but would silently break on a future upgrade to clsx 2.x, which does not split strings. See inline comment.
- Tests: the checklist item is unchecked. Snapshot/render tests confirming HTML structure is preserved would be valuable given the inheritance chain changes.

extensions/messages/js/src/forum/components/MessagesPageHero.tsx
DavideIadeluca: It's pretty common across the codebase to have empty interfaces like this — none of which have comments like this. If you would like we can follow up in a separate PR to streamline & create JSDoc's for interfaces

framework/core/js/src/forum/components/Hero.tsx
DavideIadeluca: Yeah makes sense. Simplified in 6f2e1598f9a34f95c394d4f895424de47ee41749

framework/core/js/src/forum/components/Hero.tsx
DavideIadeluca: You have explained the reasoning behind `null` yourselves :) Some implementations might not return anything in the view in some cases, like the `WelcomeHero`. Typing the `view()` of the `WelcomeHero` as `JSX.Element | null` is not compatible with the base.

Switching from `Mithril.Vnode` to JSX.Element makes sense though for better consistency. Changed in 49c820c1250e149059215cb13b85e5a9840e30f1

framework/core/js/src/forum/components/WelcomeHero.tsx
DavideIadeluca: `viewItems()` is moved to the abstract class and can technically continue to be consumed the same way. In the grand scheme of things, this is a minuscule, non breaking change. Created a documentation PR for this anyways at https://github.com/flarum/docs/pull/500

framework/core/js/src/forum/components/Hero.tsx
DavideIadeluca: Added in 721bb3482cdfea63620fd343e1970b8a41c367e7

extensions/tags/js/src/forum/components/TagHero.tsx
DavideIadeluca: Overriding the entire `view()` method just for changing the inline styles is an anti pattern in Flarums philosophy. Past reviews have called such a approach out as being very much undesired. 

About clsx... this is incorrect. clsx 2.0 continues to pretty much behave exactly the same as the 1.0 version. See https://github.com/lukeed/clsx/releases/tag/v2.0.0.

imorland (approved):

PR #4221, opened 2025-05-30
fix: prevent <button>'s from becoming form submit when they shouldn't

dsevillamartin:
Adds `type=button` to most `<button>` usage (except the `AdminPage#submitButton`). The first button of type `submit` (which is the default in HTML if undeclared) becomes what the enter keybind presses in a form. This makes the behavior with these components make more sense.

I found this behavior while trying to make an extension page use a <form>, and enter kept opening a dropdown way down the page. Even though I could've rectified this by implementing `onsubmit`, the submit behavior was already declared in the button (using `AdminPage#submitButton`) so there was no reason to if I could stop the dropdown from becoming the form's submit button.

I believe all the buttons that were missing a `type` declaration (except the one in AdminPage, which I added just for clarity)  should all be regular buttons. They can still be clicked by pressing enter when focused, but they won't override the submit button of a `<form>` if they're the first submit button anymore.

**Necessity**

**Confirmed**

  - Did not test extensions, but just adding `type="submit"` and having the code compile should be fine.

**Required changes:**

### DIFF
--- a/extensions/emoji/js/src/forum/addComposerAutocomplete.js
+++ b/extensions/emoji/js/src/forum/addComposerAutocomplete.js
@@ -70,6 +70,7 @@ export default function addComposerAutocomplete() {
           return (
             <Tooltip text={name}>
               <button
+                type="button"
                 key={emoji}
                 onclick={() => applySuggestion(emoji)}
                 onmouseenter={function () {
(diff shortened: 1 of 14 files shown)


SychO9 (approved):
Thanks!

PR #4241, opened 2025-09-27
feat: Add support for pgsql's search_path

### LINKED ISSUE #4240: [2.x] Add support for pgsql's search_path
### Current Behavior

**Note: This only applies to the pgsql driver!**

Can we, please, add support for specifying and using a database schema other than the default `"public"`?

In the back of this issue, I am submitting a PR with some changes I propose to achieve that.

### Steps to Reproduce

1. Start a new install.
2. Select the PostgreSQL driver.
3. Observe:
    - There is no field to specify the schema to be used in the database (it will use the "public" schema by default).
4. Fill the form.
5. Install.
6. Observe:
    - It used the "public" schema by default.

### Expected Behavior

The user should be able to specify and use a database schema other than the default `"public"`.

### Screenshots

**(Proposed change)**

Image

### Environment

- Flarum version: 2.0.0-beta.3
- Website URL: N/A
- Webserver: Nginx
- Hosting environment: WSL
- PHP version: 8.4.10
- Browser: Edge 140

```
johann@DESKTOP1:~/websites/REDACTED/flarum$ php flarum info
Flarum core: 2.0.0-beta.3
PHP version: 8.4.10
PostgreSQL version: PostgreSQL 17.5 (Debian 17.5-1.pgdg120+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 12.2.0-14) 12.2.0, 64-bit
Loaded extensions: Core, date, libxml, openssl, pcre, zlib, filter, hash, json, pcntl, random, Reflection, SPL, session, standard, sodium, mysqlnd, PDO, xml, calendar, ctype, curl, dom, mbstring, FFI, fileinfo, ftp, gd, gettext, iconv, intl, ldap, exif, mysqli, pdo_mysql, pdo_pgsql, pdo_sqlite, pgsql, Phar, posix, readline, shmop, SimpleXML, sockets, sqlite3, sysvmsg, sysvsem, sysvshm, tokenizer, xmlreader, xmlwriter, xsl, zip, Zend OPcache, xdebug
| flarum-flags         | 2.x-dev |        |
| flarum-approval      | 2.x-dev |        |
| flarum-tags          | 2.x-dev |        |
| flarum-suspend       | 2.x-dev |        |
| flarum-subscriptions | 2.x-dev |        |
| flarum-sticky        | 2.x-dev |        |
| flarum-statistics    | 2.x-dev |        |
| flarum-mentions      | 2.x-dev |        |
| flarum-markdown      | 2.x-dev |        |
| flarum-lock          | 2.x-dev |        |
| flarum-likes         | 2.x-dev |        |
| flarum-lang-english  | 2.x-dev |        |
| flarum-emoji         | 2.x-dev |        |
| flarum-bbcode        | 2.x-dev |        |
Base URL: http://REDACTED
Installation path: /var/www/REDACTED/flarum
Queue driver: sync
Session driver: file
Mail driver: mail
Debug mode: off
```

### Possible Solution

In the back of this issue, I am submitting a PR with some changes I propose to achieve that.

### Additional Context

_No response_

johannsa:
**Fixes #4240**

Added support for specifying and using pgsql's search_path

- Some changes to existing classes and method signatures.
- Some changes to UI:
    - Addition of a "Schema" field.
    - Triggering the update of the UI upon DOM loading, so that the field is only displayed for the relevant driver.

**Screenshot**

image

**Necessity**

    - Not acceptable: Install in `"public"` schema, then move tables to other schema and finally change `config.php`.
    - Needs to be in core as the issue/limitation exists since install.

**Confirmed**

    - There is no such command on my `composer.json`s

**Required changes:**

### DIFF
--- a/framework/core/src/Database/Migrator.php
+++ b/framework/core/src/Database/Migrator.php
@@ -11,6 +11,7 @@ namespace Flarum\Database;
 
 use Flarum\Database\Exception\MigrationKeyMissing;
 use Flarum\Extension\Extension;
+use Flarum\Install\DatabaseConfig;
 use Illuminate\Contracts\Filesystem\FileNotFoundException;
 use Illuminate\Database\ConnectionInterface;
 use Illuminate\Filesystem\Filesystem;
@@ -224,8 +225,11 @@ class Migrator
      *
      * @param string $path to the directory containing the dump.
      */
-    public function installFromSchema(string $path, string $driver): bool
+    public function installFromSchema(string $path, DatabaseConfig $dbConfig): bool
     {
+        $driver = $dbConfig->toArray()['driver'];
+        $pgSearchPath = $dbConfig->toArray()['search_path'] ?? 'public';
+
         $schemaPath = "$path/$driver-install.dump";
 
         if (! file_exists($schemaPath)) {
@@ -252,11 +256,16 @@ class Migrator
                 $this->connection->getTablePrefix() ?? '',
                 $statement
             );
+
+            if ($driver === 'pgsql' && $pgSearchPath != 'public') {
+                $statement = preg_replace('/(public)([\s.;])/', "$pgSearchPath\$2", $statement);
+            }
+
             $this->connection->statement($statement);
         }
 
         if ($driver === 'pgsql') {
-            $this->connection->statement('SELECT pg_catalog.set_config(\'search_path\', \'public\', false)');
+            $this->connection->statement("SELECT pg_catalog.set_config('search_path', '$pgSearchPath', false)");
         }
 
         $this->connection->getSchemaBuilder()->enableForeignKeyConstraints();
--- a/framework/core/src/Install/Console/FileDataProvider.php
+++ b/framework/core/src/Install/Console/FileDataProvider.php
@@ -73,6 +73,7 @@ class FileDataProvider implements DataProviderInterface
             $this->databaseConfiguration['host'] ?? 'localhost',
             $this->databaseConfiguration['port'] ?? 3306,
             $this->databaseConfiguration['database'] ?? 'flarum',
+            $this->databaseConfiguration['search_path'] ?? 'public',
             $this->databaseConfiguration['username'] ?? 'root',
             $this->databaseConfiguration['password'] ?? '',
             $this->databaseConfiguration['prefix'] ?? ''
--- a/framework/core/src/Install/Console/UserDataProvider.php
+++ b/framework/core/src/Install/Console/UserDataProvider.php
@@ -56,19 +56,30 @@ class UserDataProvider implements DataProviderInterface
             if (Str::contains($host, ':')) {
                 [$host, $port] = explode(':', $host, 2);
             }
+        }
 
+        $database = $this->ask('Database name (required):', required: true);
+
+        if ($driver === 'pgsql') {
+            $schema = $this->ask('Schema (Default: public):', 'public');
+        }
+
+        if (in_array($driver, ['mysql', 'mariadb', 'pgsql'])) {
             $user = $this->ask('Database user (required):', required: true);
             $password = $this->secret('Database password:');
         }
+
+        $prefix = $this->ask('Prefix:');
 
         return new DatabaseConfig(
             $driver,
             $host ?? null,
             intval($port),
-            $this->ask('Database name (required):', required: true),
+            $database,
+            $schema ?? null,
             $user ?? null,
             $password ?? null,
-            $this->ask('Prefix:')
+            $prefix ?? null
         );
     }
 
--- a/framework/core/src/Install/Controller/InstallController.php
+++ b/framework/core/src/Install/Controller/InstallController.php
@@ -92,6 +92,7 @@ class InstallController implements RequestHandlerInterface
             $driver,
             $host,
             intval($port),
+            Arr::get($input, 'dbSchema'),
             Arr::get($input, 'dbName'),
             Arr::get($input, 'dbUsername'),
             Arr::get($input, 'dbPassword'),
--- a/framework/core/src/Install/Controller/InstallController.php
+++ b/framework/core/src/Install/Controller/InstallController.php
@@ -92,8 +92,8 @@ class InstallController implements RequestHandlerInterface
             $driver,
             $host,
             intval($port),
-            Arr::get($input, 'dbSchema'),
             Arr::get($input, 'dbName'),
+            Arr::get($input, 'dbSchema'),
             Arr::get($input, 'dbUsername'),
             Arr::get($input, 'dbPassword'),
             Arr::get($input, 'tablePrefix')
--- a/framework/core/src/Install/DatabaseConfig.php
+++ b/framework/core/src/Install/DatabaseConfig.php
@@ -18,8 +18,8 @@ class DatabaseConfig implements Arrayable
         private readonly string $driver,
         private readonly ?string $host,
         private readonly int $port,
-        private string $schema,
         private string $database,
+        private readonly ?string $schema,
         private readonly ?string $username,
         #[\SensitiveParameter] private readonly ?string $password,
         private readonly ?string $prefix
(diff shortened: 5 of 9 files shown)


johannsa:
~~I am working on those failed test~~ **✅ Done!**

imorland (changes requested):
Hello @johannsa and thanks for the PR!

I must admit I'm not overly familiar with pgsql, so I relied on AI to help with reviewing this. Output is below:

---

In PostgreSQL, `search_path` is similar to namespaces in other languages. It determines which schema(s) PostgreSQL searches when you reference a table without specifying the schema explicitly. By default, PostgreSQL uses the `public` schema, but in enterprise environments or multi-tenant applications, you might want to use different schemas to organize or isolate data.

For example:
- `public.users` - table in the public schema
- `myapp.users` - table in the myapp schema

The `search_path` setting tells PostgreSQL which schema to use by default.

---

## **Issues Found in the PR:**

### 🔴 **Critical Issue #1: Schema Validation Logic is Wrong**

In the PR diff for `DatabaseConfig.php`, lines 62-64 add problematic validation:

```php
if (empty($this->schema) && $this->driver == 'pgsql') {
    throw new ValidationFailed('Please specify the schema name.');
}
```

**Problem**: This makes the schema **required** for PostgreSQL, but based on the PR, the default should be `'public'`. This validation will fail when `$this->schema` is `null` even though the code elsewhere treats `null` as defaulting to `'public'`. What should happen:
- Either remove this validation entirely (since there's a default)
- Or change it to validate the schema name format if provided, but allow it to be empty/null

### 🟡 Issue #2: Inconsistent Default Handling

The PR has inconsistent default value handling:

1. In `Migrator.php:230`: `$pgSearchPath = $dbConfig->toArray()['search_path'] ?? 'public';`
2. In `UserDataProvider.php:63`: Default is `'public'` when asking user
3. In `FileDataProvider.php:76`: `$this->databaseConfiguration['search_path'] ?? 'public'`
4. In `install.php` (from PR): `<input class="FormControl" name="dbSchema" value="public">`

**Problem**: The DatabaseConfig constructor receives this value but the `driverOptions()` method at line 105 (from PR) has:

`'search_path' => $this->schema`,

If `$this->schema` is `null`, this will set `search_path` to `null` instead of `'public'`. 

**Solution**: In `DatabaseConfig`, either:

- Make the default `'public'` in the constructor parameter: `private readonly string $schema = 'public'`
- Or handle the `null` case in `driverOptions()`: `'search_path' => $this->schema ?? 'public'`

### 🟡 Issue #3: Regex Replacement Could Be Fragile

In the PR's `Migrator.php` changes:

```php
if ($driver === 'pgsql' && $pgSearchPath != 'public') {
    $statement = preg_replace('/(public)([\s.;])/', "$pgSearchPath\$2", $statement);
}
```

**Concerns**:

1. This regex will replace ANY occurrence of the word "public" followed by whitespace, dot, or semicolon - even in comments, string literals, or column names
2. The pgsql-install.dump has `SELECT pg_catalog.set_config('search_path', '', false);`, `CREATE TABLE public.db_prefix_access_tokens`, and `COMMENT ON SCHEMA public IS ''`

**Better approach**:

- Use word boundaries for safety: `/\bpublic\b([\s.;])/`
- This ensures you don't accidentally replace "public" within other words

However, since tests are passing with PostgreSQL 10, this might work in practice. Just be aware it's a bit fragile.

### 🟢 Good Things:

1. **UI changes look good** - The schema field only shows for PostgreSQL with the `data-group="pgsql"` attribute and the JavaScript properly triggers on page load via `dbDriver.dispatchEvent(new Event('change', { bubbles: true }));`
2. **Tests all pass** - Including PostgreSQL 10 tests with prefix
3. **Signature changes are consistent** - All places that create `DatabaseConfig` have been updated to include the schema parameter
4. **The feature is genuinely useful** - Many organizations require non-public schemas for security/organization reasons

### Recommendations

#### Must Fix:
1. **Fix the validation logic** - Either remove the required check for schema or handle the null/default properly
2. **Fix default value handling** - Ensure `$this->schema` defaults to `'public'` consistently

#### Nice to Have:
1. **Improve the regex** - Make it more robust with word boundaries: `/\bpublic\b([\s.;])/`
2. **Add a comment** explaining why the regex replacement is needed (for developers reading the code later)

johannsa:
Thanks for your comments, @imorland!

> ## **Issues Found in the PR:**
> ### 🔴 **Critical Issue #1: Schema Validation Logic is Wrong**
> In the PR diff for `DatabaseConfig.php`, lines 62-64 add problematic validation:
> 
> ```
> if (empty($this->schema) && $this->driver == 'pgsql') {
>     throw new ValidationFailed('Please specify the schema name.');
> }
> ```
> 
> **Problem**: This makes the schema **required** for PostgreSQL, but based on the PR, the default should be `'public'`. This validation will fail when `$this->schema` is `null` even though the code elsewhere treats `null` as defaulting to `'public'`. What should happen:
> 
> * Either remove this validation entirely (since there's a default)
> * Or change it to validate the schema name format if provided, but allow it to be empty/null
> 

I believe that at the driver level, we should avoid assumptions and rigid defaults (especially when these can be changed). This is consistent with the same way that the driver is not assuming the ports for each drive (e.g. 3306 for MySQL, 5432 for PostgreSQL, etc.).

At some point, for the sake of flexibility, we might want to re-evaluate other hardcoded assumptions and rigid defaults that are still in the `DatabaseConfig.php` such as the character sets and collations.

> ### 🟡 Issue #2: Inconsistent Default Handling
> The PR has inconsistent default value handling:
> 
> 1. In `Migrator.php:230`: `$pgSearchPath = $dbConfig->toArray()['search_path'] ?? 'public';`

I have not tested it, but perhaps we could remove that default value as it should never be `null` since validation to ensure that this is not the case is enforced in the constructor (see [here](https://github.com/flarum/framework/blob/8241635b7cb92d708bdb71f1b69e0caf7fe7a291/framework/core/src/Install/DatabaseConfig.php#L27)).

> 2. In `UserDataProvider.php:63`: Default is `'public'` when asking user

This is a sane default in user input that should cover the majority of cases, reduce the amount of `null`s sent to `DatabaseConfig`  and help with UX.

> 3. In `FileDataProvider.php:76`: `$this->databaseConfiguration['search_path'] ?? 'public'`

This is avoid breaking existing automations and scripts depending on the current behaviour (backwards compatibility).

> 4. In `install.php` (from PR): `<input class="FormControl" name="dbSchema" value="public">`

[Same as 2] This is a sane default in user input that should cover the majority of cases, reduce the amount of `null`s sent to `DatabaseConfig`  and help with UX.

> 
> **Problem**: The DatabaseConfig constructor receives this value but the `driverOptions()` method at line 105 (from PR) has:
> 
> `'search_path' => $this->schema`,
> 
> If `$this->schema` is `null`, this will set `search_path` to `null` instead of `'public'`.
> 
> **Solution**: In `DatabaseConfig`, either:
> 
> * Make the default `'public'` in the constructor parameter: `private readonly string $schema = 'public'`
> * Or handle the `null` case in `driverOptions()`: `'search_path' => $this->schema ?? 'public'`
>

This should not be an issue since validation to ensure that this is not the case is enforced in the constructor (see [here](https://github.com/flarum/framework/blob/8241635b7cb92d708bdb71f1b69e0caf7fe7a291/framework/core/src/Install/DatabaseConfig.php#L27)) and we should reasonably expect that any code calling the constructor is doing so with the required parameters, would handle exceptions during validation, etc... And also for host, port, etc.!

> ### 🟡 Issue #3: Regex Replacement Could Be Fragile
> In the PR's `Migrator.php` changes:
> 
> ```
> if ($driver === 'pgsql' && $pgSearchPath != 'public') {
>     $statement = preg_replace('/(public)([\s.;])/', "$pgSearchPath\$2", $statement);
> }
> ```
> 
> **Concerns**:
> 
> 1. This regex will replace ANY occurrence of the word "public" followed by whitespace, dot, or semicolon - even in comments, string literals, or column names

This is to deal with the high possibility that the schema dump will possibly be with schema `"public"` kind of "hardcoded" and that we need to change that before we install in a separate schema.

> 2. The pgsql-install.dump has `SELECT pg_catalog.set_config('search_path', '', false);`, `CREATE TABLE public.db_prefix_access_tokens`, and `COMMENT ON SCHEMA public IS ''`

See above.

> Better approach:
>
>Use word boundaries for safety: /\bpublic\b([\s.;])/
>This ensures you don't accidentally replace "public" within other words

Will test that suggestion and, if it works, add it to my PR to avoid any tech debt.

In the meantime, please, let me know what you think about my reply to your comments in Issue 1 and Issue 3.

Thanks beforehand!

imorland:
So this does not get delayed any further, I will merge this now, and see how we go with it from being out in the wild with some early adopters etc. We can then re-visit down the line if neccessary.

Thanks for your work on this @johannsa

imorland (approved):

PR #4242, opened 2025-10-02
[1.x] fix(a11y): missing default alt text for avatar

imorland:
This PR improves accessibility of the avatar helper by ensuring that  avatars include meaningful alt text by default.
	•	If a user is present and no alt attribute is provided, it falls back to the user’s display name.
	•	If explicitly set to null/false, the alt attribute is omitted (treated as decorative).
	•	For cases where no user is passed, the alt defaults to an empty string.

This change aligns avatars with [WCAG guidelines](https://www.w3.org/WAI/WCAG21/Understanding/non-text-content.html) by providing screen readers with descriptive text, while still allowing developers to override or suppress alt when needed.

A follow-up PR for `2.x` will also be made

**Fixes #0000**

**Screenshot**

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/framework/core/js/src/common/helpers/avatar.tsx
+++ b/framework/core/js/src/common/helpers/avatar.tsx
@@ -22,12 +22,6 @@ export default function avatar(user: User | null, attrs: ComponentAttrs = {}): M
   const hasTitle: boolean | string = attrs.title === 'undefined' || attrs.title;
   if (!hasTitle) delete attrs.title;
 
-  // If the `alt` attribute is set to null or false, we don't want to give the
-  // avatar an alt description. If it hasn't been set, we can default it later
-  // to the user's display name for accessibility.
-  const hasAlt: boolean | string = attrs.alt === 'undefined' || attrs.alt;
-  if (!hasAlt) delete attrs.alt;
-
   // If a user has been passed, then we will set up an avatar using their
   // uploaded image, or the first letter of their username if they haven't
   // uploaded one.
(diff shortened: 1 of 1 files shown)


framework/core/js/src/common/helpers/avatar.tsx
```diff
@@ -31,12 +37,24 @@ export default function avatar(user: User | null, attrs: ComponentAttrs = {}): M
...
+      attrs.alt = username;
+    }
+
     if (avatarUrl) {
-      return <img {...attrs} src={avatarUrl} alt="" />;
+      return <img {...attrs} src={avatarUrl} />;
     }
 
     content = username.charAt(0).toUpperCase();
     attrs.style = { '--avatar-bg': user.color() };
```
DavideIadeluca, line 51: _Don't commit the following thru GH, the diff kind of got messed up_

```suggestion
content = username.charAt(0).toUpperCase();
attrs.style = { '--avatar-bg': user.color() };

delete attrs.alt;
delete attrs.loading;
attrs.role = 'img';
attrs['aria-label'] = username;
```

When no avatarUrl exists, this extra logic removes the alt (and for good measures the `loading="lazy"`  property) and instead sets the aria `role` to `img` and the `aria-label` to username

DavideIadeluca (changes requested):
Tested locally — working as expected, though I have some improvements for the case where no `avatarUrl` exists

The alt property should be omitted in case of no user avatar (`alt` on `span` is invalid)

framework/core/js/src/common/helpers/avatar.tsx
imorland: I did that in #4243 not sure why I didn't do it here! 🙈

DavideIadeluca (approved):
tested locally, looks good to me now

PR #4246, opened 2025-10-03
[1.x] chore(core): improve `a11y` of avatar with no `avatarUrl`

DavideIadeluca:
**Fixes #0000**

* Removes the unnecessary `loading="lazy"` attribute for avatars with no `avatarUrl`
* Sets the aria role to `img` and adds an `aria-label`

**Screenshot**
Output before this change

```html
<span class="Avatar PostUser-avatar" loading="lazy" style="--avatar-bg: #a0e5b6;">D</span>
```

Output after this change:

```html
<span class="Avatar PostUser-avatar" role="img" aria-label="davetodave178" style="--avatar-bg: #a0e5b6;">D</span>
```

**Necessity**

**Confirmed**

**Required changes:**


imorland (approved):
Thanks Dave!

PR #4254, opened 2025-10-06
[1.x] refactor(core): improve extensibility of `NotificationList`

DavideIadeluca:
**Fixes #0000**

This PR drastically improves the extensibility of the NotificationList component

While the group parameter is unused in this implementation, it allows third party extensions to differentiate between notification types and then customize this value only when needed

* This change must obviously be 100% backwards compatible, existing appearance  / behaviour must be the same

Example of how this group title _can_ be customized:

```jsx
extend(NotificationList.prototype, 'groupHeaderItems', (items, group) => {
    if (group.discussion) return;

    const hasNotificationType = (group.notifications || []).some((notification) => notification.contentType?.() === 'extensionDisabled');

    if (!hasNotificationType) return;

    items.setContent('neutral', <div className="NotificationGroup-header">Custom Title</div>);
  });
```

**Screenshot**

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/framework/core/js/src/forum/components/NotificationList.js
+++ b/framework/core/js/src/forum/components/NotificationList.js
@@ -14,19 +14,27 @@ import Tooltip from '../../common/components/Tooltip';
  */
 export default class NotificationList extends Component {
   view() {
+    return <div className="NotificationList">{this.viewItems().toArray()}</div>;
+  }
+
+  viewItems() {
     const state = this.attrs.state;
+    const items = new ItemList();
 
-    return (
-      <div className="NotificationList">
-        <div className="NotificationList-header">
-          <h4 className="App-titleControl App-titleControl--text">{app.translator.trans('core.forum.notifications.title')}</h4>
+    items.add('header', <div className="NotificationList-header">{this.headerItems().toArray()}</div>, 100);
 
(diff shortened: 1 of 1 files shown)


imorland:
Thanks for this @DavideIadeluca :)

Whilst I can see that this will solve the immediate problem of not being able to directly set the title when no group is defined, I feel that as a framework, we should improve the extensibility for all of this component.

Could we perhaps take a similar approach as in #4249 here for maximum flexibility? This would also mean we can `extend` rather than `override` this component :)

DavideIadeluca:
Will reconsider and get back to you 👍

DavideIadeluca:
@imorland I’ve made the requested changes and updated the example showing how a group title can be customized. For reviewing, I recommend using the old review UI in Split View.

For testing, please log in with a user that has a variety of notifications, then open two tabs — one with the old JS dist and one with the dev server running, and then compare if everything looks and works the same.

imorland (approved):
Looks good to me @DavideIadeluca - thanks for putting this together!

I'll include this for `1.8.11`, would you have the time at some point to create a companion PR targetting `2.x` as well, so that these changes carry forward? 🙏

DavideIadeluca:
Sure can do! :)

PR #4259, opened 2025-10-11
[2.x] feat: Include `DatabaseQueue` in core

imorland:
**Fixes #0000**

**Screenshot**

**Necessity**

**Confirmed**

**Required changes:**


luceos:
I honestly don't think db queue should be set to the default from the get go. It makes the hurdle to "go live" even more difficult to achieve. What we should look for in 2.x is the ability to understand community scale, then offer to enable the db queue with sufficient documentation on what to do to make it work. And yes, I do believe the db queue should be part of core, just not the default.

Flarum has set out to be as simple as possible, without much configuration. This change steers away from that.

imorland:
I came to a similar conclusion here, which is why this has not seen any progress..

The database queue is actually not difficult to run or maintain, it's only requirement is the scheduler being active. Therefore my thinking for when I come back to this is as follows:

- If the scheduler is not running, default to the `sync` queue (current `1.x` and `2.x` behaviour)
- If the scheduler is running, switch to the `database` queue
- Provide a mechanism via `config.php` to pin the queue to a given driver, for maximum flexibility

Additional things to consider:

- What happens when another queue driver is introduced (redis, messageMQ, etc)

I think step 1 (sync vs db) is fairly clear and straightforward to implement and document. Step 2 (other queues) still needs some thought...

PR #4260, opened 2025-10-12
[2.x] Fix condition for displaying not sending message in MailPage

ghost:
Correct the logic to display the not sending message in the MailPage based on the sending status.


imorland (approved):
Thank you @notanothervibecoder

PR #4333, opened 2026-01-03
[2.x] Add option to reuse default request error handler to handle extension toggles

### LINKED ISSUE #4332: [2.x] 500s when enabling extensions aren't handled properly in UI
### Current Behavior

If I enable an extension through admin and it fails with a 500, the loading modal closes and nothing else appears except for an Uncaught exception in console. This is partly because it has a custom error handler.

### Steps to Reproduce

1. Somehow cause a 500 when enabling extensions (mine was due to a migration that no longer works with Laravel update)
2. Try to enable extension

### Expected Behavior

I should get an error message outside the console.

### Screenshots

Image

### Environment

- Flarum version: ee2a79f077978fd4a8d569e37549250b6b21c0a5
- Webserver: nginx
- Hosting environment: local
- PHP version: 8.3.16
- Browser: Firefox 146

### Possible Solution

Reuse default error handler (seems like this is desired behavior due to the throw - not sure if it worked like this in 1.x). I am creating a PR to address it.

### Additional Context

_No response_

dsevillamartin:
**Fixes #4332**

- Add option to fallback to default error handler if custom one re-throws the request error
  - If a non-RequestError is thrown, it will just be re-thrown by the default error handler so nothing happens
- Update error handling in ExtensionPage 

Should this be an option at all? Should it be true by default? I didn't dive deep into what the previous functionality was -- not sure what would be best to have either. I'd say more error handling is better than less though, but it could have sideffects elsewhere.

**Screenshot**
https://github.com/user-attachments/assets/e598272f-a943-42eb-a2cf-13a4caeb9414

**Necessity**

**Confirmed**

### DIFF
--- a/framework/core/js/src/admin/components/ExtensionPage.tsx
+++ b/framework/core/js/src/admin/components/ExtensionPage.tsx
@@ -294,7 +294,6 @@ export default class ExtensionPage<Attrs extends ExtensionPageAttrs = ExtensionP
         method: 'PATCH',
         body: { enabled: !enabled },
         errorHandler: this.onerror.bind(this),
-        fallbackToDefaultErrorHandler: true,
       })
       .then(() => {
         if (!enabled) localStorage.setItem('enabledExtension', this.extension.id);
@@ -309,14 +308,14 @@ export default class ExtensionPage<Attrs extends ExtensionPageAttrs = ExtensionP
     return isExtensionEnabled(this.extension.id);
   }
 
-  onerror(e: RequestError) {
+  onerror(e: RequestError): void | false {
     app.modal.close();
 
     this.changingState = false;
 
-    // If it's not a conflict error, rethrow it to be handled by the default handler.
+    // If it's not a conflict error, use the default request error handler.
     if (e.status !== 409) {
-      throw e;
+      return false;
     }
 
     const error = e.response?.errors?.[0];
--- a/framework/core/js/src/admin/components/ExtensionPage.tsx
+++ b/framework/core/js/src/admin/components/ExtensionPage.tsx
@@ -294,11 +294,13 @@ export default class ExtensionPage<Attrs extends ExtensionPageAttrs = ExtensionP
         method: 'PATCH',
         body: { enabled: !enabled },
         errorHandler: this.onerror.bind(this),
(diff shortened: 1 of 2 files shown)


imorland:
I'm not sure I fully understand this. Is this not expected behaviour when debug mode is active and the backend throws a 500?

dsevillamartin:
@imorland The error currently is that _nothing_ happens if enabling/disabling the extension throws a non-409 error. Debug mode does not matter. There's more info in the issue description of #4332 about the problem that made me look into this.
The video in this PR is the *fixed* state after this PR. Currently, there is only an uncaught exception (both with  debug mode on or off).

The aim here is to allow for falling back to the default error handler if an HTTP error is thrown from the error handler. Looking at v1 code it seems like it was maybe an issue there as well ([ExtensionPage onerror throws](https://github.com/flarum/framework/blob/601e0f5a250203913e64d1a822f72c103c544ed5/framework/core/js/src/admin/components/ExtensionPage.tsx#L271-L273), which is [not caught by the error handler](https://github.com/flarum/framework/blob/601e0f5a250203913e64d1a822f72c103c544ed5/framework/core/js/src/common/Application.tsx#L579-L581) and thus no alert is shown).

This can also be useful in other scenarios where one only wants to have a custom error handler for a specific error, and use the default one if the condition doesn't match. I believe this should likely be default behavior but I didn't look into all the custom error handling that may occur in the codebase, so I made it optional.

---

Looking at my new code again, I do see that the catch just assumes the re-thrown error was the original, which is not good (silences new error since it assumes it was the original HTTP error). I'll turn this into draft just to not merge right now.

dsevillamartin:
I've simplified it to just return `false` explicitly from the error handler. That way the error does not need to be checked against, and the behavior is intentional. I also updated the comments to explain why one would want to use it and when it is better to use `.catch` instead. The `.catch` addition in ExtensionPage is to avoid the uncaught rejection error message, since the error is handled (also done elsewhere in codebase already!)

Existing uses of `errorHandler` in core & core exts don't conflict with this implementation (it is mostly just the FormModal onerror that is used).

imorland (approved):

PR #4338, opened 2026-01-11
[1.x] Fix running phpstan with newer PHP versions (e.g. 8.4)

n-peugnet:

phpstan 1.8 cannot run properly with PHP 8.4, but newer versions in the 1.x branch do, thus we change the requirement to phpstan ^1.9, to allow version up to 2.0 (excluding).

Newer phpstan versions now detect common errors in regexes:

    Ignored error #Template type T of function resolve[()]{2} is not referenced in a parameter.# has an unescaped '()' which leads to ignoring all errors. Use '\(\)' instead.
    Ignored error #^Unsafe usage of new static[()]{2}.$# has an unescaped '()' which leads to ignoring all errors. Use '\(\)' instead.

So these regexes are also fixed by using [non-evaluated string literals](https://doc.nette.org/en/neon/format#toc-strings) and escaping the parenthesis as recommended instead of using a weird trick with the brackets.

???

**Screenshot**

Before:

2026-01-11-170241_1080x639_scrot

After:
2026-01-11-165928_1081x610_scrot

**Necessity**

**Confirmed**

### DIFF
--- a/php-packages/phpstan/composer.json
+++ b/php-packages/phpstan/composer.json
@@ -4,7 +4,7 @@
     "minimum-stability": "dev",
     "license": "MIT",
     "require": {
-        "phpstan/phpstan": ">=1.8.11 < 1.9.0",
+        "phpstan/phpstan": "^1.9",
         "nunomaduro/larastan": "^1.0"
     },
     "autoload": {
--- a/php-packages/phpstan/phpstan-baseline.neon
+++ b/php-packages/phpstan/phpstan-baseline.neon
@@ -9,12 +9,12 @@ parameters:
 		  reportUnmatched: false
 
 		#	We ignore this because resolve can either take a class name as the generic return type or just a binding name.
-		- message: "#Template type T of function resolve[()]{2} is not referenced in a parameter.#"
+		- message: '#Template type T of function resolve\(\) is not referenced in a parameter.#'
 		  reportUnmatched: false
 
 		#	We ignore new static errors because we want extensibility.
 		#	@TODO: needs discussion.
-		- message: "#^Unsafe usage of new static[()]{2}.$#"
+		- message: '#^Unsafe usage of new static\(\).$#'
 		  reportUnmatched: false
 
 		# ConnectionInterface lacks methods that exist in the implementation,


imorland (changes requested):
Thank you for this @n-peugnet. I've not ran this locally yet, but yes I think this will probably work out!

Can you please rebase this, and target the `1.x` branch 🙏

n-peugnet:
@imorland: just rebased, sorry I didn't find this `1.x` branch and thought that `main` was the version 1 branch.

imorland:
Thank you for this PR @n-peugnet! I've tested the changes locally with PHP 8.4 and unfortunately discovered a compatibility issue that prevents us from merging this.

## The Problem

While PHPStan 1.9+ works perfectly with PHP 8.4, it's incompatible with the version of Larastan we're using. Specifically:

- Flarum 1.x uses Laravel 8 (illuminate packages ^8.0)
- Laravel 8 requires Larastan 1.x (both `nunomaduro/larastan` and `larastan/larastan`)
- Larastan 1.x only supports PHPStan `^1.0 <1.9` due to breaking API changes in PHPStan 1.9

When running PHPStan 1.9+ with Larastan 1.x, we get this error:
```
Internal error: Too few arguments to function PHPStan\Analyser\MutatingScope::assignVariable(), 
2 passed in vendor/nunomaduro/larastan/src/ReturnTypes/CollectionFilterDynamicReturnTypeExtension.php 
on line 96 and at least 3 expected
```

## Testing

I confirmed that:
- ✅ PHPStan 1.12.32 runs successfully on PHP 8.4 when Larastan is disabled
- ❌ PHPStan 1.9+ crashes when Larastan is enabled due to API incompatibility

## Path Forward

For Flarum 1.x, we'll need to keep PHPStan pinned to `>=1.8.11 < 1.9.0`. Static analysis can continue running on PHP ≤8.3 while production code runs on PHP 8.4+.

This should be revisited for Flarum 2.x when we upgrade to Laravel 10+, which will allow us to use Larastan 2.x with PHPStan 1.9+.

I think the only option is to close this PR for now, but I really appreciate you taking the time to investigate and submit this fix! 🙏

n-peugnet:
I understand, then what about setting `>=1.8.11 < 2.0` instead and let users of larastan use the [version 1.0.4](https://packagist.org/packages/larastan/larastan#1.0.4) that specifies `phpstan ^1.0 <1.9`?

This would allow extension developers that do not use larastan to still use phpstan with PHP >= 8.4, and users of larastan to still be able to use it with PHP < 8.4.

imorland:
Thanks again for digging into this @n-peugnet — genuinely appreciated.

Unfortunately this can't land on 1.x. `flarum/phpstan` hard-requires `nunomaduro/larastan: ^1.0`, and Larastan 1.x only supports PHPStan `^1.0 <1.9`. So even with `>=1.8.11 <2.0`, Composer will always resolve PHPStan back down to <1.9 to satisfy Larastan — which means PHP 8.4 still ends up on a PHPStan 1.8.x that can't run there. The only way around it would be decoupling Larastan from `flarum/phpstan`, which isn't something we want to take on for 1.x (security-maintenance only).

The good news: this is already resolved on **2.x**, which uses `phpstan/phpstan: ^2.1` + `larastan/larastan: ^3.8` and runs fine on PHP 8.4.

Closing as `wontfix` for 1.x. Thanks again for the thorough investigation!

PR #4350, opened 2026-01-24
[2.x] feat(core, mentions): implement ability to order groups

DavideIadeluca:
**Fixes #0000**

* Implement ability to change ordering of Groups, similar to how it works for `flarum/tags`
* Implement reusable `GroupBar` component which contains the `sortablejs` integration. Extensions like https://github.com/imorland/flarum-ext-twofactor/blob/2.x/js/src/admin/components/ExtractedGroupBar.tsx could consume this new component

* Backend, how to handle the Guest and Member Group here? The DB migration does set a position, but the first time the order is changed in the frontend those groups' position would be set back to null. From a usability point of view it doesn't matter, but it could be argued it's a quirk we don't wanna have
* Also the changes in the webpack config.. it appears like that the current minification step from Terser fails to parse the esm build of sortable js due to unsupported syntax. The alias I added forces webpack to use sortables UDM build which appears to fix the issues. Really unsure how we should address this issue and if maybe changes in flarums webpack config itself are needed.

This was the error message (reproducible both in CI and when running `yarn build` locally):
```
ERROR in admin.js
  admin.js from Terser plugin
  Unexpected token: punc ({) [webpack://../node_modules/sortablejs/modular/sortable.esm.js:1212,31][admin.js:22591,37]
      at js_error (/home/runner/work/framework/framework/node_modules/terser/dist/bundle.min.js:536:11)
      at croak (/home/runner/work/framework/framework/node_modules/terser/dist/bundle.min.js:1264:9)
      at token_error (/home/runner/work/framework/framework/node_modules/terser/dist/bundle.min.js:1272:9)
      at unexpected (/home/runner/work/framework/framework/node_modules/terser/dist/bundle.min.js:1278:9)
      at semicolon (/home/runner/work/framework/framework/node_modules/terser/dist/bundle.min.js:1316:56)
      at simple_statement (/home/runner/work/framework/framework/node_modules/terser/dist/bundle.min.js:1577:73)
      at statement (/home/runner/work/framework/framework/node_modules/terser/dist/bundle.min.js:1390:19)
      at _embed_tokens_wrapper (/home/runner/work/framework/framework/node_modules/terser/dist/bundle.min.js:1329:26)
      at block_ (/home/runner/work/framework/framework/node_modules/terser/dist/bundle.min.js:2168:20)
      at _function_body (/home/runner/work/framework/framework/node_modules/terser/dist/bundle.min.js:2080:21)
```
**Screenshot**
https://github.com/user-attachments/assets/8dcb047b-bd52-435e-a5b1-a5da46673664

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/framework/core/js/src/admin/components/GroupBar.tsx
+++ b/framework/core/js/src/admin/components/GroupBar.tsx
@@ -14,7 +14,7 @@ export interface IGroupBarAttrs extends ComponentAttrs {
   groups: Group[];
 }
 
-export default abstract class GroupBar<CustomAttrs extends IGroupBarAttrs = IGroupBarAttrs> extends Component<CustomAttrs> {
+export default class GroupBar<CustomAttrs extends IGroupBarAttrs = IGroupBarAttrs> extends Component<CustomAttrs> {
   groups: Group[] = [];
 
   oninit(vnode: Mithril.Vnode<CustomAttrs, this>) {
--- a/framework/core/js/src/admin/components/GroupBar.tsx
+++ b/framework/core/js/src/admin/components/GroupBar.tsx
@@ -14,6 +14,21 @@ export interface IGroupBarAttrs extends ComponentAttrs {
   groups: Group[];
 }
 
+/**
+ * A reusable component for displaying a list of groups in the admin interface,
+ * with the ability to edit and reorder them.
+ *
+ * @property {Group[]} groups Required. Groups to display.
+ *
+ * @example
+ * ```ts
+ * const availableGroups = app.store
+ *   .all<Group>('groups')
+ *   .filter((group) => [Group.GUEST_ID, Group.MEMBER_ID].indexOf(group.id()!) === -1);
+ *
+ * <GroupBar groups={availableGroups} />
+ * ```
+ */
 export default class GroupBar<CustomAttrs extends IGroupBarAttrs = IGroupBarAttrs> extends Component<CustomAttrs> {
   groups: Group[] = [];
 
--- a/framework/core/js/src/admin/components/GroupBar.tsx
+++ b/framework/core/js/src/admin/components/GroupBar.tsx
@@ -1,0 +1,80 @@
+import sortable from 'sortablejs';
+
+import app from '../../admin/app';
+import Component, { ComponentAttrs } from '../../common/Component';
+import GroupBadge from '../../common/components/GroupBadge';
+import Icon from '../../common/components/Icon';
+import EditGroupModal from './EditGroupModal';
+import sortGroups from '../../common/utils/sortGroups';
+
+import type Group from '../../common/models/Group';
+import type Mithril from 'mithril';
+
+export interface IGroupBarAttrs extends ComponentAttrs {
+  groups: Group[];
+}
+
+export default abstract class GroupBar<CustomAttrs extends IGroupBarAttrs = IGroupBarAttrs> extends Component<CustomAttrs> {
+  groups: Group[] = [];
+
+  oninit(vnode: Mithril.Vnode<CustomAttrs, this>) {
+    super.oninit(vnode);
+
+    this.groups = sortGroups(this.attrs.groups);
+  }
+
+  view(): JSX.Element {
+    return (
+      <div className="GroupBar" oncreate={this.onGroupBarCreate.bind(this)}>
+        {this.groups.map((group) => (
+          <button className="Button Group" type="button" data-id={group.id()} onclick={() => app.modal.show(EditGroupModal, { group })}>
+            <GroupBadge group={group} className="Group-icon" label={null} />
+            <span className="Group-name">{group.namePlural()}</span>
+          </button>
+        ))}
+        <button className="Button Group Group--add" type="button" onclick={() => app.modal.show(EditGroupModal)}>
+          <Icon name="fas fa-plus" className="Group-icon" />
+          <span className="Group-name">{app.translator.trans('core.admin.permissions.new_group_button')}</span>
+        </button>
+      </div>
+    );
+  }
+
+  onGroupBarCreate(vnode: Mithril.VnodeDOM) {
+    sortable.create(vnode.dom as HTMLElement, {
+      group: 'groups',
+      delay: 50,
+      delayOnTouchOnly: true,
+      touchStartThreshold: 5,
+      animation: 150,
+      swapThreshold: 0.65,
+      dragClass: 'Group-Sortable-Dragging',
+      ghostClass: 'Group-Sortable-Placeholder',
+
+      filter: '.Group--add',
+      onMove: (evt) => !evt.related.classList.contains('Group--add'),
+
+      onSort: (e) => this.onSortUpdate(),
+    });
+  }
+
+  onSortUpdate() {
+    const order = this.$('.Group:not(.Group--add)')
+      .map(function () {
+        return $(this).data('id');
+      })
+      .get();
+
+    order.forEach((id, i) => {
+      app.store.getById<Group>('groups', id)?.pushData({
+        attributes: { position: i },
+      });
+    });
+
+    app.request({
+      url: app.forum.attribute('apiUrl') + '/groups/order',
+      method: 'POST',
+      body: { order },
+    });
+  }
+}
--- a/framework/core/js/src/admin/components/GroupBar.tsx
+++ b/framework/core/js/src/admin/components/GroupBar.tsx
@@ -38,6 +38,11 @@ export default class GroupBar<CustomAttrs extends IGroupBarAttrs = IGroupBarAttr
     this.groups = sortGroups(this.attrs.groups);
   }
 
+  onupdate(vnode: Mithril.VnodeDOM<CustomAttrs, this>) {
+    super.onupdate(vnode);
+    this.groups = sortGroups(this.attrs.groups);
+  }
+
   view(): JSX.Element {
     return (
       <div className="GroupBar" oncreate={this.onGroupBarCreate.bind(this)}>
--- a/framework/core/js/src/admin/components/GroupBar.tsx
+++ b/framework/core/js/src/admin/components/GroupBar.tsx
@@ -69,7 +69,7 @@ export default class GroupBar<CustomAttrs extends IGroupBarAttrs = IGroupBarAttr
       filter: '.Group--add',
       onMove: (evt) => !evt.related.classList.contains('Group--add'),
 
-      onSort: (e) => this.onSortUpdate(),
+      onSort: () => this.onSortUpdate(),
     });
   }
 
--- a/framework/core/js/src/common/utils/sortGroups.ts
+++ b/framework/core/js/src/common/utils/sortGroups.ts
@@ -1,5 +1,14 @@
 import type Group from '../models/Group';
 
 export default function sortGroups(groups: Group[]) {
-  return groups.slice().sort((a, b) => a.position() - b.position());
+  return groups.slice().sort((a, b) => {
+    const aPos = a.position();
+    const bPos = b.position();
+
+    if (aPos === null && bPos === null) return 0;
+    if (aPos === null) return 1;
+    if (bPos === null) return -1;
+
+    return aPos - bPos;
+  });
 }
--- a/framework/core/js/src/common/utils/sortGroups.ts
+++ b/framework/core/js/src/common/utils/sortGroups.ts
@@ -1,0 +1,5 @@
+import type Group from '../models/Group';
+
+export default function sortGroups(groups: Group[]) {
+  return groups.slice().sort((a, b) => a.position() - b.position());
+}
--- a/framework/core/js/webpack.config.cjs
+++ b/framework/core/js/webpack.config.cjs
@@ -5,4 +5,9 @@ module.exports = merge(config(), {
   output: {
     library: 'flarum.core',
   },
+  resolve: {
+    alias: {
+      sortablejs: require.resolve('sortablejs/Sortable.js'),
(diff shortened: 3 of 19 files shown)


framework/core/js/src/common/utils/sortGroups.ts
```diff
@@ -0,0 +1,5 @@
+import type Group from '../models/Group';
+
+export default function sortGroups(groups: Group[]) {
+  return groups.slice().sort((a, b) => a.position() - b.position());
```
imorland: This will behave incorrectly when `position()` returns `null` — `null - null` coerces to `0` in JS, and `null - 5` coerces to `-5`, so groups with a null position sort as if they have position `0`, jumping to the front. This happens for new groups created after an ordering, and for any group not included in the `order` payload.

Please handle null positions explicitly, following the same pattern as `sortTags.tsx`:

```typescript
export default function sortGroups(groups: Group[]) {
  return groups.slice().sort((a, b) => {
    const aPos = a.position();
    const bPos = b.position();

    if (aPos === null && bPos === null) return 0;
    if (aPos === null) return 1;
    if (bPos === null) return -1;

    return aPos - bPos;
  });
}
```

framework/core/js/src/admin/components/GroupBar.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+import type Group from '../../common/models/Group';
+import type Mithril from 'mithril';
+
+export interface IGroupBarAttrs extends ComponentAttrs {
+  groups: Group[];
+}
+
+export default abstract class GroupBar<CustomAttrs extends IGroupBarAttrs = IGroupBarAttrs> extends Component<CustomAttrs> {
+  groups: Group[] = [];
```
imorland, line 33: `GroupBar` is declared `abstract` but `PermissionsPage` uses it directly via `<GroupBar groups={availableGroups} />`. TypeScript will compile this, but Mithril instantiates components at runtime — instantiating an abstract class throws in JS.

Either:
1. Remove `abstract` if this is meant to be directly usable as-is, or
2. Keep it abstract and introduce a concrete subclass (`class DefaultGroupBar extends GroupBar {}`) for `PermissionsPage` to use — which would also give extensions a clean extension point without having to extend an abstract class directly.

framework/core/js/src/admin/components/GroupBar.tsx
```diff
@@ -0,0 +1,80 @@
...
+      delay: 50,
+      delayOnTouchOnly: true,
+      touchStartThreshold: 5,
+      animation: 150,
+      swapThreshold: 0.65,
+      dragClass: 'Group-Sortable-Dragging',
+      ghostClass: 'Group-Sortable-Placeholder',
+
+      filter: '.Group--add',
+      onMove: (evt) => !evt.related.classList.contains('Group--add'),
```
imorland, line 75: `onSort` receives an event `e` that is declared but never used. Either use it or remove the parameter:

```typescript
onSort: () => this.onSortUpdate(),
```

framework/core/js/src/admin/components/GroupBar.tsx
```diff
@@ -0,0 +1,80 @@
...
+      .map(function () {
+        return $(this).data('id');
+      })
+      .get();
+
+    order.forEach((id, i) => {
+      app.store.getById<Group>('groups', id)?.pushData({
+        attributes: { position: i },
+      });
+    });
```
imorland, line 92: `app.request` is fire-and-forget with no error handling. If the request fails, SortableJS has already reordered the DOM and the store has been updated via `pushData`, leaving the UI and server in an inconsistent state.

At minimum, revert the store positions on failure:

```typescript
const previousPositions = order.map((id, i) => ({
  id,
  position: app.store.getById<Group>("groups", id)?.position() ?? null,
}));

app.request({ ... }).catch(() => {
  previousPositions.forEach(({ id, position }) => {
    app.store.getById<Group>("groups", id)?.pushData({ attributes: { position } });
  });
  m.redraw();
});
```

framework/core/js/webpack.config.cjs
```diff
@@ -5,4 +5,9 @@ module.exports = merge(config(), {
   output: {
     library: 'flarum.core',
   },
+  resolve: {
```
imorland, line 8: This alias fixes the Terser/ESM issue for core, but any extension that bundles sortablejs directly will hit the same error. The fix should live in `flarum-webpack-config` so it applies everywhere, or sortablejs should be exposed as an external from the core bundle (similar to how mithril/jQuery are handled), so extensions can reference it without bundling their own copy.

Could you look into moving this to `flarum-webpack-config`? Or would you prefer we handle that separately?

framework/core/src/Api/Resource/GroupResource.php
```diff
@@ -93,6 +93,7 @@ public function fields(): array
                 ->writable(),
             Schema\Boolean::make('isHidden')
                 ->writable(),
+            Schema\Integer::make('position')
```
imorland, line 96: `position` should be `->readonly()`. Without it, the field is writable via `PATCH /api/groups/{id}`, bypassing the ordering endpoint and its admin check entirely. Position should only be settable through `POST /api/groups/order`.

```php
Schema\Integer::make(position)
    ->nullable()
    ->readonly(),
```

framework/core/src/Api/Controller/OrderGroupsController.php
```diff
@@ -0,0 +1,40 @@
...
+    public function handle(ServerRequestInterface $request): ResponseInterface
+    {
+        RequestUtil::getActor($request)->assertAdmin();
+
+        $order = Arr::get($request->getParsedBody(), 'order');
+
+        if ($order === null) {
+            return new EmptyResponse(422);
+        }
+
```
imorland, line 31: A few input validation gaps here:

1. No type check — `$order` could be a non-array (e.g. a string), which would make `foreach` throw.
2. No validation that the IDs in `$order` are real group IDs — an invalid ID silently does nothing but the reset at line 33 (`update position = null` for all) still runs.
3. Guest (ID 2) and Member (ID 3) groups could be included in the payload, which would set positions on groups the frontend intentionally excludes. Consider ignoring or rejecting those IDs.

Suggested additions:
```php
if (\!is_array($order)) {
    return new EmptyResponse(422);
}
```

framework/core/src/Group/Group.php
```diff
@@ -27,6 +27,7 @@
  * @property string|null $color
  * @property string|null $icon
  * @property bool $is_hidden
+ * @property int $position
```
imorland: The migration adds `->nullable()` and `GroupFactory` defaults to `null`, so this should be `@property int|null $position`.

framework/core/less/admin/PermissionsPage.less
```diff
@@ -1,10 +1,15 @@
...
+  display: flex;
+  gap: 10px;
+  flex-wrap: wrap;
   padding: 10px;
+  border-radius: var(--border-radius);
 }
+.Group-Sortable-Placeholder {
+    outline: 2px dashed var(--body-bg);
+    border-radius: var(--border-radius);
+  }
```
imorland, line 12: Inconsistent indentation — the rest of the file uses 2-space indent but `.Group-Sortable-Placeholder` uses 4-space. Please normalise.

imorland (changes requested):
Thanks for this PR — group ordering is a welcome addition and the overall structure is good. A few issues need addressing before this can merge, from correctness problems down to minor polish. See inline comments for specifics.

**Summary of required changes:**

- `sortGroups`: handle `null` positions (same pattern as `sortTags`)
- `GroupBar`: remove `abstract` or use a concrete subclass in `PermissionsPage` — cannot instantiate an abstract class directly
- `GroupResource`: add `->readonly()` to the `position` field
- `OrderGroupsController`: add input validation (type-check, verify IDs are real groups, protect Guest/Member)
- Webpack alias: this fix belongs in `flarum-webpack-config`, not just core — otherwise extension authors who bundle sortablejs will hit the same Terser error
- Clean up the two `chore: debugging` commits before merge (squash or rebase)
- Error handling on the `app.request` in `onSortUpdate` — revert on failure

**Questions for you:**
- Is `GroupBar` intended to be a base class that extensions subclass (as the twofactor example suggests), or should it be directly usable? If the former, `PermissionsPage` needs a concrete `DefaultGroupBar extends GroupBar`.
- For the webpack fix — do you want to explore moving sortablejs to a Flarum external (exposed from core bundle) rather than an alias? That would be cleaner long-term.

framework/core/js/src/admin/components/GroupBar.tsx
DavideIadeluca: Weird that the direct usage of the abstract GroupBar worked before.. but yeah the intention is that the Class can be used as is without a subclass. Changed in b64a1123f19a5b37eddb29c2568ab1924dd5a9f0. Also added JSDoc in d9c7e4a4c06aaeda0f11b135d7035d674354dfbf

framework/core/js/src/admin/components/GroupBar.tsx
DavideIadeluca: > app.request is fire-and-forget with no error handling.

This is not quite correct. `app.request()` actually has a ton of error handling. But yes it's true that, should the request fail, that the UI and Server are in an inconsistent state.

This is the same behaviour as for when ordering Tags or other implementations of `app.request()` like dismissing flags or changing permissions. Should those implementations also be adapted to make sure that the UI state is correctly reflected when the server returns an error? If yes let's create a follow up issue for this, If no I'd say the implementation in this PR just follows existing standards

framework/core/js/src/admin/components/GroupBar.tsx
DavideIadeluca: Yep good catch. Removed in ca59aca50798611632bc881a3abb2a5a2d5fab79

framework/core/src/Api/Resource/GroupResource.php
DavideIadeluca: `FlarumField::readonly` doesn't actually exist. From what I see in other implementations, fields are readonly by default and need to me made writable thru `SetsValue::writable`. Even if it was writable, the permissions checks wouldn't be bypassed as the field would be behind the same existing permissions checks `HasAuthorization::authenticated` and `HasAuthorization::can` in the `endpoints` method.

But adding `FlarumField::nullable` seems logical to me (new groups by default have no position and can exist without it with no further issues). Added in 30197e4141a2bd0155d6becae21b5fba2d89008c.

framework/core/src/Group/Group.php
DavideIadeluca: Yep you're right. Fixed in 76afa29f906edb0a610cd42e1671af67925bdc8d

framework/core/less/admin/PermissionsPage.less
DavideIadeluca: Changed in c3c5cda88054a1dcc1f946dfa2790620861e9f5b

framework/core/src/Api/Controller/OrderGroupsController.php
DavideIadeluca: > No type check — $order could be a non-array (e.g. a string), which would make foreach throw.

Changed and added a test in 91de3579dd174e4d0a56c4d7d2083d3d182e0cf6

> No validation that the IDs in $order are real group IDs — an invalid ID silently does nothing but the reset at line 33 (update position = null for all) still runs.

This is related to the quirk I described in the PR description. The DB Migration does set a position for all groups, including Member and Guest: so this behaviour is kind of desired with the above quirk. Essentially in the DB migration doesn't know for a fact that ID's 2 and 3 are indeed the Guest and Member Group without some runtime check which validates the class constants of `Group`. 

> Guest (ID 2) and Member (ID 3) groups could be included in the payload, which would set positions on groups the frontend intentionally excludes. Consider ignoring or rejecting those IDs.

Idk if this is a bug or a feature. The implementation in PermissionsPage does ignore those Groups, but the new component does technically support them as groups are passed into the component. There might be some cases where third party customizations want control over this. Not sure if it's worth it to restrict this

framework/core/js/src/common/utils/sortGroups.ts
DavideIadeluca: Yeah I guess why not. Changed in 11fb48efd855b90ef1c509f1adc4d4687c0f6268

framework/core/js/webpack.config.cjs
DavideIadeluca: Yeah I'd prefer if we handle this separately. Open to moving this PR back to draft until that is done or proceed with this PR and make the webpack fix later. Can create a follow up issue if you woud like

imorland (changes requested):
**`GroupBar` has a stale-state bug after group create/delete**

`GroupBar.groups` is assigned once in `oninit` and never updated. After `EditGroupModal` creates or deletes a group, the parent `PermissionsPage` re-renders but `GroupBar`'s internal array and sortable instance remain stale until a page reload.

An `onupdate` hook is needed to detect when `attrs.groups` changes and reinitialise:

```ts
onupdate(vnode: Mithril.VnodeDOM<CustomAttrs, this>) {
  super.onupdate(vnode);
  this.groups = sortGroups(this.attrs.groups);
  // reinitialise sortable on the updated DOM
}
```

DavideIadeluca:
> GroupBar has a stale-state bug after group create/delete

Good catch! fixed in https://github.com/flarum/framework/pull/4350/commits/7138aa6b4a1801d02db74106c456ea1106a17387

imorland (approved):

PR #4351, opened 2026-01-26
Convert punctuateSeries to TypeScript

vtejasri1511-sudo:
**Partially addresses #3533**

- Converted `framework/core/js/src/common/helpers/punctuateSeries` from JavaScript to TypeScript.
- No behavior or API changes intended; this is a small isolated step in the ongoing core JS → TS migration.

- Confirm logic is unchanged (only typing / file extension change).
- Confirm TS types (`mithril` `Children`) are appropriate and build tooling accepts the new `.ts` file.

**Screenshot**
N/A (no user-facing change)


framework/core/js/src/common/helpers/punctuateSeries.ts
```diff
@@ -0,0 +1,34 @@
+import type { Children } from 'mithril';
+import app from '../../common/app';
+
+/**
+ * Formats a list of strings/nodes to read fluently in the locale.
+ */
+export default function punctuateSeries(items: Children[] = []): Children {
```
imorland, line 7: The function-level JSDoc describing the purpose and example has been removed. Since `dist-typings` is generated from this file, that documentation will be lost for extension authors browsing the `.d.ts` declarations. Restore the description and `@example` from the original `.js` file:

```ts
/**
 * The `punctuateSeries` helper formats a list of strings (e.g. names) to read
 * fluently in the application's locale.
 *
 * ```js
 * punctuateSeries(['Toby', 'Franz', 'Dominion']) // Toby, Franz, and Dominion
 * ```
 */
```

imorland, line 7: The `items` parameter has a default of `[]`. The original function required an argument. While defensive, this is a subtle API change — callers passing `undefined` will now silently get `[]` back. Since all current call sites pass the array explicitly, consider dropping the default to keep the types tight:

```ts
export default function punctuateSeries(items: Children[]): Children {
```

framework/core/js/src/common/helpers/punctuateSeries.ts
```diff
@@ -0,0 +1,34 @@
...
+  if (items.length === 2) {
+    return app.translator.trans('core.lib.series.two_text', {
+      first: items[0],
+      second: items[1],
+    });
+  }
+
+  if (items.length >= 3) {
+    const glue = app.translator.trans('core.lib.series.glue_text') as Children;
+
```
imorland, line 17: `trans()` with parameters returns `any[]` (see `Translator.tsx` overloads), not `Children`. The `as Children` assertion is masking an underlying type mismatch rather than resolving it.

The simplest correct approach is to type `glue` as `any[]`:

```ts
const glue = app.translator.trans('core.lib.series.glue_text') as any[];
```

Or, since the translator returns the rendered translation here, you can just use `any` and let the array push accept it — or keep `as Children` with a comment explaining why. Either way, a note clarifying the intentional cast would help reviewers.

imorland (commented):
Good first contribution — the migration is on the right track and the logic is correct. A few things need addressing before merge.

**Logic equivalence ✅**
The new `for` loop produces identical output to the original `.reduce().slice(0,-1)` chain. Both correctly build `[item, glue, item, ...]` without a trailing glue. The refactor is actually clearer.

**Return type concern**
`trans()` with parameters is typed as `any[]`, not `Children`. The cast `as Children` on the `glue` variable silences a type error rather than reflecting the actual types involved. The return type `Children` on the function itself is fine (it matches the existing `.d.ts`), but the `as Children` assertion is a sign the types aren't quite right.

**Default parameter**
Adding `= []` as a default is a minor API change. The existing call sites always pass the array, but it does change the observable signature. Not a blocker, but worth noting.

framework/core/js/src/common/helpers/punctuateSeries.ts
imorland, line 7: The function-level JSDoc describing the purpose and example has been removed. Since `dist-typings` is generated from this file, that documentation will be lost for extension authors browsing the `.d.ts` declarations. Restore the description and `@example` from the original `.js` file:

```ts
/**
 * The `punctuateSeries` helper formats a list of strings (e.g. names) to read
 * fluently in the application's locale.
 *
 * ```js
 * punctuateSeries(['Toby', 'Franz', 'Dominion']) // Toby, Franz, and Dominion
 * ```
 */
```

framework/core/js/src/common/helpers/punctuateSeries.ts
imorland, line 17: `trans()` with parameters is typed as `any[]` (see `Translator.tsx` overloads), not `Children`. The cast `as Children` is masking an underlying type mismatch rather than resolving it.

The simplest correct approach is to type `glue` as `any[]`:

```ts
const glue = app.translator.trans('core.lib.series.glue_text') as any[];
```

Or keep `as Children` but add a comment explaining the intentional cast so reviewers understand it's deliberate.

imorland (changes requested):
Good first contribution — the migration is on the right track and the logic is correct. A few things need addressing before merge.

**Logic equivalence ✅**
The new `for` loop produces identical output to the original `.reduce().slice(0,-1)` chain. Both correctly build `[item, glue, item, ...]` without a trailing glue. The refactor is actually clearer.

**Return type concern**
`trans()` with parameters is typed as `any[]`, not `Children`. The cast `as Children` on the `glue` variable silences a type error rather than reflecting the actual types involved. The return type `Children` on the function itself is fine (it matches the existing `.d.ts`), but the `as Children` assertion is a sign the types aren't quite right.

**Default parameter**
Adding `= []` as a default is a minor API change. The existing call sites always pass the array, but it does change the observable signature. Not a blocker, but worth noting.

PR #4361, opened 2026-02-13
fix(forum): prevent scroll jump when loading discussion to specific post (#4346)

### LINKED ISSUE #4346: [2.x] Discussion rendering does not allow jumping to the right position
### Current Behavior

Since Flarum 1.x Flarum had the weak-point of not being able to calculate sizes of certain elements in posts (such as images). This leads to the effect that if you jump to a certain post (like the last new one) you often are not at the right position as content is loaded.

Comment from @clarkwinkelmann:
> The page jumping around on discussion load has become much much worse in Flarum v2. Previously this was only an issue if the posts above the viewport contained images, but now it seems almost every formatting feature (mentions, quotes, GitHub links, ...) causes a shift in the layout after the page was initially loaded.
> 
> If I find some time I'll investigate further. As things currently stand this is very annoying and I don't think this should be shipped like this. Either we need to revert back to the v1 behavior, of finally implement proper layout observer and resize so that the scroll never jumps when a post above the viewport changes height.

see also: https://discuss.flarum.org/d/38539-discuss-now-running-on-flarum-v2/68

### Steps to Reproduce

Create a discussion with multiple posts that contains images for example and jump to a post below the content.

### Expected Behavior

Jumping to the post always shows the right discussion

### Screenshots

_No response_

### Environment

- Flarum version: x.y.z
- PHP version: x.y.z

```
```

### Possible Solution

_No response_

### Additional Context

_No response_

iPurpl3x:
## Description

Prevents the viewport from jumping when opening a discussion with a post number in the URL (e.g. `/d/9-slug/63`). Asynchronous content (images, embeds) loading above the viewport was shifting the target post out of view after the initial scroll.

## Solution

- **Settling guard**: After programmatic scroll to a target post, we enter a short "settling" phase. A `ResizeObserver` watches `.PostStream-item` elements; on any resize we re-scroll to the target post **synchronously** (in the same callback, before paint), so the user never sees the shifted layout.
- Settling ends when: the user interacts (wheel/touch/key), no resizes occur for 3 seconds, or a 30s hard timeout. We then call `calculatePosition()` to update the URL and scrubber.
- During settling, scroll events do not update the URL so we don't overwrite the intended post number with a wrong one.

## Changes

- **PostStream.js**: `setupScrollAnchoring()`, `reScrollToTarget()`, settling lifecycle (start/stop listeners, stability timer, `endSettling()`), `observeNewPostItems()`, `cleanupScrollAnchoring()`; `onscroll` skips position/URL updates while settling; `scrollToItem()` enters settling after the initial scroll and defers `calculatePosition()` until settling ends.
- **PostStream.test.ts**: ResizeObserver stub for JSDOM; tests for setup/cleanup, observing new items, re-scroll during settling, no re-scroll when not settling, user interaction and stability timeout ending settling.

## Testing

- Open a discussion with a post number in the URL (e.g. `/d/9-slug/63`). Reload multiple times: the viewport should stay on the target post as images load, without jump or flicker.
- Existing PostStream integration tests updated; new tests cover scroll-anchoring behaviour.

Closes #4346.

### DIFF
--- a/framework/core/js/src/forum/components/PostStream.js
+++ b/framework/core/js/src/forum/components/PostStream.js
@@ -152,12 +152,14 @@ export default class PostStream extends Component {
     super.onupdate(vnode);
 
     this.triggerScroll();
+    this.observeNewPostItems();
   }
 
   oncreate(vnode) {
     super.oncreate(vnode);
 
     this.triggerScroll();
+    this.setupScrollAnchoring();
 
     // This is wrapped in setTimeout due to the following Mithril issue:
     // https://github.com/lhorie/mithril.js/issues/637
@@ -168,6 +170,7 @@ export default class PostStream extends Component {
     super.onremove(vnode);
 
     this.scrollListener.stop();
+    this.cleanupScrollAnchoring();
     clearTimeout(this.calculatePositionTimeout);
   }
 
@@ -197,6 +200,11 @@ export default class PostStream extends Component {
     this.updateScrubber(top);
 
     this.loadPostsIfNeeded(top);
+
+    // While the page is settling after a programmatic scroll, layout shifts
+    // are still happening (images loading, etc). Skip position/URL updates
+    // until settling completes -- endSettling() will call calculatePosition().
+    if (this.settling) return;
 
     // Throttle calculation of our position (start/end numbers of posts in the
     // viewport) to 100ms.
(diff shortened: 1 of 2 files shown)


imorland (approved):
Thank you @iPurpl3x this is a very elegant solution to the problem! This will form part of the upcoming `2.0.0-beta.7` release

PR #4371, opened 2026-02-21
fix(forum): sync discussion URL immediately on programmatic scroll (#4346)

### LINKED ISSUE #4346: [2.x] Discussion rendering does not allow jumping to the right position

iPurpl3x:
## Description

Two bugs introduced by the scroll-anchoring settling guard (#4361):

1. Clicking or dragging the scrubber delayed the browser URL update by ~3s. The settling guard suppresses `calculatePosition()` during settling, so the URL was not updated until the stability timer fired.
2. If a user opened a discussion and navigated away within 3s, the browser URL would remain on the discussion URL. On unmount, the still-running settling guard called `calculatePosition()` → `positionChanged()` → `window.history.replaceState()`, overwriting the destination page's URL with the old discussion URL.

## Solution

- **Immediate URL emit**: `scrollToItem()` calls `calculatePosition({ fallbackToTarget: true })` right after the scroll lands and before entering settling. This updates the URL without waiting for the stability timer, while the scroll-anchoring behavior introduced in #4361 is fully preserved.
- **Unmount safety**: `cleanupScrollAnchoring()` now calls `endSettling({ updatePosition: false })`, so the settling teardown on unmount never writes to the browser history.
- **Duplicate write guard**: `positionChanged()` skips `replaceState` and `app.history.push` when `startNumber` hasn't changed, preventing a double write from the immediate emit and the settle-end reconciliation resolving to the same post.
- **Consecutive jump cleanup**: `beginSettling(target)` replaces the inline settling setup in `scrollToItem()` and fully tears down any in-progress settling (listeners, timers) before arming new ones, preventing event listener and timer leaks on rapid scrubber jumps.

## Changes

- **`PostStream.js`**: Extract `computeVisiblePosition(top)` (pure DOM scan), `resolveTargetStartNumber()` (target-based fallback when visible items are loading placeholders), `emitPosition()` (single call site for `onPositionChange`); refactor `calculatePosition()` to compose these with an optional `fallbackToTarget` param; add `beginSettling(target)`; add `updatePosition` param to `endSettling()`; update `cleanupScrollAnchoring()` to pass `updatePosition: false`; call `calculatePosition({ fallbackToTarget: true })` and `beginSettling()` in `scrollToItem()`; replace deprecated `window.pageYOffset` with `window.scrollY` throughout.
- **`DiscussionPage.tsx`**: Gate `replaceState` and `app.history.push` in `positionChanged()` on `startNumber !== this.near`; read-state save is unaffected.
- **`PostStream.test.ts`**: 12 new tests covering immediate emit, settling suppression via `onscroll`, `endSettling` reconciliation, unmount no-op, consecutive-jump listener leak, placeholder fallback, reply-target no-emit, and all `resolveTargetStartNumber` target shapes.

## Testing

- Scrubber click/drag: URL updates immediately, no ~3s wait.
- Open a discussion and navigate away within 3s: URL must follow the navigation, not remain on the discussion.
- Media-heavy discussion: scroll-anchoring (no viewport jump) must still work as in #4361.
- Existing PostStream integration tests pass; 12 new tests added.

Closes #4346.

### DIFF
--- a/framework/core/js/src/forum/components/DiscussionPage.tsx
+++ b/framework/core/js/src/forum/components/DiscussionPage.tsx
@@ -214,18 +214,28 @@ export default class DiscussionPage<CustomAttrs extends IDiscussionPageAttrs = I
   /**
    * When the posts that are visible in the post stream change (i.e. the user
    * scrolls up or down), then we update the URL and mark the posts as read.
+   *
+   * URL and history writes are skipped when startNumber hasn't changed from the
+   * last known position — this prevents double history churn from the immediate
+   * post-scroll emit and the subsequent settle-end reconciliation both resolving
+   * to the same post. Read-state saves are intentionally independent of this
+   * guard and use their own endNumber condition.
    */
   positionChanged(startNumber: number, endNumber: number): void {
     const discussion = this.discussion;
 
     if (!discussion) return;
 
-    // Construct a URL to this discussion with the updated position, then
-    // replace it into the window's history and our own history stack.
-    const url = app.route.discussion(discussion, (this.near = startNumber));
+    // Only write history when the visible start post actually changed.
+    if (startNumber !== this.near) {
+      // Construct a URL to this discussion with the updated position, then
+      // replace it into the window's history and our own history stack.
+      const url = app.route.discussion(discussion, startNumber);
+      this.near = startNumber;
 
-    window.history.replaceState(null, document.title, url);
-    app.history.push('discussion', discussion.title());
+      window.history.replaceState(null, document.title, url);
+      app.history.push('discussion', discussion.title());
+    }
 
     // If the user hasn't read past here before, then we'll update their read
     // state and redraw.
--- a/framework/core/js/src/forum/components/PostStream.js
+++ b/framework/core/js/src/forum/components/PostStream.js
@@ -194,7 +194,7 @@ export default class PostStream extends Component {
    *
    * @param {number} top
    */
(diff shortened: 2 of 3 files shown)


imorland (approved):
Looks good to me, thank you so much @iPurpl3x 

I'll merge this now for futher testing

PR #4393, opened 2026-02-27
[2.x] Fixed tags mutation shows [deleted] in some cases, together with improvements

edgeinfinity1:
**Fixes #0000**

As title. If a tag was deleted from a discussion, it will no longer be included in its relationship, causing the DiscussionTaggedPost shows [deleted].
I modernized the component, added loading indicator, and now it will fetch tags list if any uncached tag is mentioned in mutation.

Maybe adding deleted tags in discussions' relationship is a better solution? Though I don't think it's necessary.

**Screenshot**

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/extensions/tags/js/src/forum/components/DiscussionTaggedPost.js
+++ b/extensions/tags/js/src/forum/components/DiscussionTaggedPost.js
@@ -2,18 +2,43 @@ import EventPost from 'flarum/forum/components/EventPost';
 import tagsLabel from '../../common/helpers/tagsLabel';
 
 export default class DiscussionTaggedPost extends EventPost {
-  static initAttrs(attrs) {
-    super.initAttrs(attrs);
+  oninit(vnode) {
+    super.oninit(vnode);
 
-    const oldTags = attrs.post.content()[0];
-    const newTags = attrs.post.content()[1];
+    const oldTags = this.attrs.post.content()[0];
+    const newTags = this.attrs.post.content()[1];
 
-    function diffTags(tags1, tags2) {
-      return tags1.filter((tag) => tags2.indexOf(tag) === -1).map((id) => app.store.getById('tags', id));
+    this.tagsAdded = [];
+    this.tagsRemoved = [];
+
+    this.fetchRequired = false;
+    this.loading = true;
+
+    const tagsMutation = [...this.diffTags(newTags, oldTags), ...this.diffTags(oldTags, newTags)];
+    if (tagsMutation.includes(undefined)) {
+      this.fetchRequired = true;
     }
 
-    attrs.tagsAdded = diffTags(newTags, oldTags);
-    attrs.tagsRemoved = diffTags(oldTags, newTags);
+    const afterFetch = () => {
+      this.tagsAdded = this.diffTags(newTags, oldTags);
+      this.tagsRemoved = this.diffTags(oldTags, newTags);
+
+      this.loading = false;
+      m.redraw();
+    };
+
+    if (this.fetchRequired) {
+      app.store.find('tags').then(afterFetch);
+    } else {
+      afterFetch();
+    }
+  }
+
+  diffTags(tags1, tags2) {
+    return tags1
+      .filter((tag) => tags2.indexOf(tag) === -1)
+      .map((id) => app.store.getById('tags', id))
+      // .filter(Boolean);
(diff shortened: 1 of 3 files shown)


imorland (changes requested):
Thanks for the PR! The core approach is sound — detecting uncached tags and re-fetching before rendering is the right direction. A few issues need addressing before this can merge:

### Truly deleted tags still produce `undefined`

The commented-out `.filter(Boolean)` reveals an unresolved decision:

```js
.map((id) => app.store.getById('tags', id))
// .filter(Boolean);
```

After the fetch, a tag that has been genuinely deleted from the database will still return `undefined` from `getById` — it simply won't be in the store. So the `[deleted]` rendering can still occur, just less frequently (only for truly deleted tags rather than uncached ones). 

Please either:
- Uncomment `.filter(Boolean)` to silently drop deleted tags from the display, or
- Explicitly handle `undefined` entries in `tagsLabel` / the description text

### No error handling on the fetch

If `app.store.find('tags')` rejects (network error, server error), `this.loading` stays `true` forever and the component shows a spinner indefinitely. Please add a `.catch()`:

```js
app.store.find('tags').then(afterFetch).catch(() => {
  this.loading = false;
  m.redraw();
});
```

### Minor: hardcoded `height: 62px`

```less
.EventPost.DiscussionTaggedPost.Post.Post--loading {
    height: 62px;
}
```

Magic number — fragile if the post layout changes. Consider using `min-height` or deriving from a variable instead.

imorland (changes requested):
Thanks for the PR! The core approach is sound — detecting uncached tags and re-fetching before rendering is the right direction. A few issues need addressing before this can merge:

### Truly deleted tags still produce `undefined`

The commented-out `.filter(Boolean)` reveals an unresolved decision:

```js
.map((id) => app.store.getById('tags', id))
// .filter(Boolean);
```

After the fetch, a tag that has been genuinely deleted from the database will still return `undefined` from `getById` — it simply won't be in the store. So the `[deleted]` rendering can still occur, just less frequently (only for truly deleted tags rather than uncached ones).

Please either:
- Uncomment `.filter(Boolean)` to silently drop deleted tags from the display, or
- Explicitly handle `undefined` entries in `tagsLabel` / the description text

### No error handling on the fetch

If `app.store.find('tags')` rejects (network error, server error), `this.loading` stays `true` forever and the component shows a spinner indefinitely. Please add a `.catch()`:

```js
app.store.find('tags').then(afterFetch).catch(() => {
  this.loading = false;
  m.redraw();
});
```

### Minor: hardcoded `height: 62px`

```less
.EventPost.DiscussionTaggedPost.Post.Post--loading {
    height: 62px;
}
```

Magic number — fragile if the post layout changes. Consider using `min-height` or deriving from a variable instead.

edgeinfinity1:
I appreciate your advice a lot and added some patches, but I still have a few questions.

> So the [deleted] rendering can still occur, just less frequently (only for truly deleted tags rather than uncached ones).

Returning `undefined` by default renders [deleted] on frontend, but this could only happen if the tag **actually no longer exists**. Isn't that the expected behavior already? I think it might be better to inform users that some tags are included in mutation history, even if they no longer exist.

> No error handling on the fetch

I patched that according to your guidance.

> Minor: hardcoded height: 62px

I added this style to fix loading indicator being too tall, since it's around 100px by default. When I was looking further into the size calculation according to your advice, I found that it was pretty hard-coded already:
image
image
These means that the height of this component could only be exactly 22 + 20 * 2 = 62px if window width >=768px, and exactly 21 + 20 * 2 = 61px if window width <=767.98px. Making it `max-height` or `min-height` neither works nor makes sense. Using calculations or variables are also seemingly unnecessary.
This height only affects the loading indicators' visual position, not any layout. Should I add the 62/61 media switch, or just leave it be? Or if there's a better approach, please tell me.

All the testing and observation were done on `1.8.x` stable. Please tell me if anything has changed in `2.x`.

I'm looking forward to your reply.

imorland (approved):
Thanks for the follow-up and the explanations. I agree with your reasoning on both points — showing `[deleted]` for genuinely deleted tags is the correct behaviour, and the hardcoded `62px` is justified given the fixed layout math. The `.catch()` is in place. Approving.

PR #4397, opened 2026-03-03
[2.x] fix(webpack): prevent `Uncaught ReferenceError` for async default exports

rafaucau:
image

* **`autoExportLoader.cjs`**: Fixed a regex bug that incorrectly captured the `async` keyword as an export name when using anonymous arrow functions (e.g., `export default async () => {}`). This prevents the `Uncaught ReferenceError: async is not defined` runtime error in extensions.
* **`babel.config.cjs`**: Dynamically set `useESModules` based on `process.env.NODE_ENV === 'test'`. This fixes the Jest test suite which was failing due to ESM imports being injected into the CommonJS test environment.
* Added tests for both named and anonymous async exports.

**Necessity**

**Confirmed**

### DIFF
--- a/js-packages/webpack-config/babel.config.cjs
+++ b/js-packages/webpack-config/babel.config.cjs
@@ -1,3 +1,5 @@
+const isTest = process.env.NODE_ENV === 'test';
+
 module.exports = {
   assumptions: {
     // Defines assumptions Babel can make about our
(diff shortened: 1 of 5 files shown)


imorland (approved):
Thank you @rafaucau this is a much needed fix!

PR #4407, opened 2026-03-06
[2.x] fix(mentions): wait for async TextEditor loading before injecting composer mentions/quotes

### LINKED ISSUE #4401: [2.x] Composer loses Quote/Reply context if triggered before fully loaded
### Current Behavior

When connection are slow, clicking "Reply" or "Quote" to open the composer will not inject the intended quote or mention.
Reproducible on both: https://discuss.flarum.org and https://nightly.flarum.site
Seems to be due to the lazy loading of composer in flarum 2.x

### Steps to Reproduce

1. Open a discussion in Flarum.
2. Open browser Developer Tools and set Network Throttling to "Fast 4G" (or even slower if needed).
3. Hard refresh the page.
4. Click the Reply button on a specific post. (Or Quote from the text in the post)
5. Observe that the composer eventually opens, but it fails to include the proper @ mention or quote context. Closing it and clicking "Reply" a second time works correctly.

### Expected Behavior

The quote or mention should appear in the composer when it opens.

### Screenshots

https://github.com/user-attachments/assets/37597210-0329-4e1e-96be-c59d24885917

### Environment

- Flarum version: x.y.z
- PHP version: x.y.z

```
```

### Possible Solution

_No response_

### Additional Context

_No response_

huoxin233:
**Fixes #4401**

Waits for composer editor to be ready before interaction

Resolves a race condition where actions like inserting mentions or quotes could attempt to modify the composer's text editor before it has been fully initialized and assigned.

Introduces an `editorReady` promise in `ComposerState` that resolves once the editor is available. Updates the `insertMention` utility to await this promise, ensuring the editor is ready for content manipulation. A timeout is included as a fallback for robustness.

In `reply.js`, I wrapped the `editorReady()` call in a `Promise.race()` with a 3000ms timeout:
https://github.com/huoxin233/flarum-framework/blob/14d9f283f5a92fbd516ba189bddcef979c8ae36a/extensions/mentions/js/src/forum/utils/reply.js#L28-L38

I added it for the purpose of: If a severe error from another extension completely breaks the `TextEditor` and prevents the editor from ever loading, `editorReady()` would ordinarily remain pending forever. By forcing it to try to insert after 3000ms, the code aggressively throws the native `TypeError: Cannot read properties of null exception`, preserving the stack trace in the browser console for easier debugging.

But, I am not quite sure how much timeout is needed, would 3000ms be enough?

**Screenshot**

Before:

https://github.com/user-attachments/assets/0d42dcb2-7bd6-4f93-b160-dc42f6aef153

After:

https://github.com/user-attachments/assets/cbe8a723-416d-4418-8bb1-2d63211bc728

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/extensions/mentions/js/src/forum/utils/reply.js
+++ b/extensions/mentions/js/src/forum/utils/reply.js
@@ -3,25 +3,42 @@ import DiscussionControls from 'flarum/forum/utils/DiscussionControls';
 
 export function insertMention(post, composer, quote) {
   return new Promise((resolve) => {
-    const mention = app.mentionFormats.mentionable('post').replacement(post) + ' ';
+    const doInsert = () => {
+      const mention = app.mentionFormats.mentionable('post').replacement(post) + ' ';
 
-    // If the composer is empty, then assume we're starting a new reply.
-    // In which case we don't want the user to have to confirm if they
-    // close the composer straight away.
-    if (!composer.fields.content()) {
-      composer.body.attrs.originalContent = mention;
+      // If the composer is empty, then assume we're starting a new reply.
+      // In which case we don't want the user to have to confirm if they
+      // close the composer straight away.
+      if (!composer.fields.content()) {
+        composer.body.attrs.originalContent = mention;
+      }
+
+      const cursorPosition = composer.editor.getSelectionRange()[0];
+      const preceding = composer.fields.content().slice(0, cursorPosition);
+      const precedingNewlines = preceding.length == 0 ? 0 : 3 - preceding.match(/(\n{0,2})$/)[0].length;
+
+      composer.editor.insertAtCursor(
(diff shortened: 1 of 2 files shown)


imorland:
Thank you so much for digging into this and putting together a fix — we really appreciate the effort! You correctly identified the root cause (the async `TextEditor` lifecycle) and the promise-based approach is exactly the right direction.

We've gone ahead and opened #4411 as an alternative that builds on your idea but avoids the hard timeout entirely. The key insight is that `TextEditor` already fires an `onTextEditorBuilt` callback at exactly the right moment — so we can wire the `editorReady()` promise to resolve from there, rather than racing against a timer. This means `insertMention` will always wait for the real signal, with no arbitrary fallback period.

Closing this in favour of #4411 — thanks again!

PR #4409, opened 2026-03-07
fix(less): fix FA Kit icon alignment and spurious icon regressions

imorland:
## Summary

Fixes several icon layout issues introduced by FA Kit (SVG+JS mode) support and a regression from the `.Button--icon` LESS mixin conversion.

### Root causes

1. **FA Kit injects a stylesheet dynamically** (after Flarum's compiled CSS) that sets `display: block` via `--fa-display` on injected SVG elements. The previous fix used a broad `display: inline-block !important` on `.Button-icon`/`.Button-caret` globally, which then required counter-`!important` rules in `Dropdown.less` to hide elements — creating a cascade of overrides.

2. **`d2129e63` converted `.Button--icon` to a LESS-only parametric mixin**, removing its class-level CSS output. DOM elements with `class="Button--icon"` no longer had the caret/label hidden, causing the caret to appear in icon-only buttons (e.g. post controls `···` dropdown).

3. **UserCard controls dropdown** was passing `icon="fas fa-ellipsis-v"` unconditionally. With 2.x's `inline-flex` Button layout the icon rendered visibly alongside the "Controls" label on the user profile page.

### Fixes

- **Button.less**: Scope the Kit display fix to `.Button .Button-icon, .Button .Button-caret` (no `!important`), placed *before* `.Button--icon` so hiding rules that follow override it by cascade order. Restore `.Button--icon` as a real CSS class selector.
- **Dropdown.less**: Remove `!important` from split-dropdown hide/show rules (now handled by specificity/cascade). Add `width: 16px !important` on dropdown menu `.Button-icon` to prevent the Kit stylesheet overriding the fixed column width.
- **Input.less**: Add `width: 36px !important` on `.Input-prefix-icon` to prevent the Kit stylesheet shifting the search icon left.
- **UserCard.js**: Only pass the `icon` to the controls Dropdown when `controlsButtonClassName` includes `Button--icon` (icon-only context).

## Test plan

- [ ] Local font source: all buttons render correctly, no caret in `···` post controls, no ellipsis in UserPage "Controls" button, split dropdown shows caret only
- [ ] CDN source: same as above
- [ ] FA Kit source: same as above, plus search icon stays centred, split dropdown shows caret only (not ellipsis + caret)

### DIFF
--- a/framework/core/js/src/forum/components/UserCard.js
+++ b/framework/core/js/src/forum/components/UserCard.js
@@ -129,7 +129,7 @@ export default class UserCard extends Component {
           buttonClassName={this.attrs.controlsButtonClassName}
           label={app.translator.trans('core.forum.user_controls.button')}
           accessibleToggleLabel={app.translator.trans('core.forum.user_controls.toggle_dropdown_accessible_label')}
-          icon="fas fa-ellipsis-v"
+          icon={this.attrs.controlsButtonClassName?.includes('Button--icon') ? 'fas fa-ellipsis-v' : undefined}
         >
           {controls}
         </Dropdown>,
--- a/framework/core/less/common/Button.less
+++ b/framework/core/less/common/Button.less
@@ -232,13 +232,23 @@
     margin-top: 5px;
   }
 }
+// FontAwesome Kit dynamically injects a stylesheet that sets `display: block` (via --fa-display)
+// on SVG elements it injects, breaking the button's inline-flex layout.
+// Scoped to inside .Button so hiding rules that follow (.Button--icon, .Dropdown--split)
+// can override this by coming later in the cascade at equal specificity.
+.Button .Button-icon,
+.Button .Button-caret {
+  display: inline-block;
+}
+
 // Little round icon buttons
 .Button--icon() {
   width: 36px;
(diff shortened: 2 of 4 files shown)


framework/core/js/src/forum/components/UserCard.js
```diff
@@ -129,7 +129,7 @@ export default class UserCard extends Component {
           buttonClassName={this.attrs.controlsButtonClassName}
           label={app.translator.trans('core.forum.user_controls.button')}
           accessibleToggleLabel={app.translator.trans('core.forum.user_controls.toggle_dropdown_accessible_label')}
-          icon="fas fa-ellipsis-v"
+          icon={this.attrs.controlsButtonClassName?.includes('Button--icon') ? 'fas fa-ellipsis-v' : undefined}
```
dsevillamartin, line 132: This seems to have caused the ellipsis to not show up at all for eg. user actions on mobile. Icon is no longer present on discuss (the caret one is, but hidden) and instead has a hidden "Controls" button label.



https://discuss.flarum.org/d/38800-flarum-200-beta7-released-rc-is-on-the-horizon/115

PR #4433, opened 2026-03-12
[2.x] Implement `CloseWatcher` for modals and dropdowns

rafaucau:
On mobile devices users intuitively use the "back" gesture/button to dismiss open modals. Currently, this triggers a browser history back navigation, leading to frustrating unintended page departures.

https://developer.mozilla.org/en-US/docs/Web/API/CloseWatcher

Components updated:
- [x] Drawer
- [x] ModalManager
- [x] Dropdowns
- [x] PostStreamScrubber

On browsers without `CloseWatcher` support, the affected components behave as before.

**Screenshot**

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/896f8d46-67b9-4935-a500-527882c0157a"></video> | <video src="https://github.com/user-attachments/assets/2cf3bf1e-b894-4902-9b59-e95f523ed869"></video> |

**Necessity**

**Confirmed**

### DIFF
--- a/framework/core/js/package.json
+++ b/framework/core/js/package.json
@@ -25,6 +25,7 @@
         "@flarum/jest-config": "^2.0.0",
         "@flarum/prettier-config": "^1.0.0",
         "@types/body-scroll-lock": "^3.1.0",
+        "@types/dom-close-watcher": "^1.0.0",
         "@types/jquery": "^3.5.10",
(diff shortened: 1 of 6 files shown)


imorland (approved):
Looks great, thank you @rafaucau

PR #4448, opened 2026-03-14
[2.x] feat: change logout from GET to POST

### LINKED ISSUE #4427: [2.0] Change HTTP Request Method for Logout Route
### Current Behavior

Users logout of Flarum through the `/logout` route. If no `token` query param is provided, users just get a static page with a logout button, when hitting the logout button a `token` query param is added which triggers the actual logout.

The `/logout` route is a GET request, which is undesired for the following reasons:
* `GET` requests should be idempotent
* A `GET` for logout is not semantically correct
* Browsers or prefetchers might trigger it accidentally
* CSRF tokens in a query param might get leaked accidentally 

### Steps to Reproduce

Logging out of Farum happens through a `GET` Request

### Expected Behavior

Actually logging out of the Forum should not be a `GET` request, but rather preferably a `POST` request (`DELETE` can maybe also be considered).

### Screenshots

_No response_

### Environment

Flarum 1.x & 2.x

_No response_

### Possible Solution

Keep a dedicated static logout page as a GET request, but introduce a new `POST` endpoint

### Additional Context

At the same time we might want to think about removing the token from the query param and instead pass it thru the request body, which mitigates potential issues with analytics / tracking scripts picking up the token accidentally.

imorland:
## Summary

- `GET /logout` now only shows a no-JS confirmation page (`LogOutViewController`) — it never performs a logout
- `POST /logout` is the new action endpoint (`LogOutController`) — reads the CSRF token from the request body and performs the logout
- `Session.ts` now submits a hidden form via POST instead of navigating via `window.location.href`
- The no-JS confirmation page (`log-out.blade.php`) now renders a `<form method="POST">` with a hidden token field instead of an `<a>` link

**Why:**
- `GET` requests should be side-effect free; session destruction is a state mutation
- The CSRF token was previously in the query string, leaking to server logs, browser history, and `Referer` headers
- Prefetchers / link crawlers can follow GET links without user intent

Closes #4427

## Test plan

- [x] 9 new integration tests in `LogoutTest.php` covering: guest POST redirects, wrong/missing token falls back to confirmation page, valid token logs out and deletes the access token, safe return URL is honoured, external return URL is blocked, GET shows confirmation page with POST form, GET does not destroy any tokens
- [ ] Manual: click Log Out in the UI, verify the session is destroyed and you're redirected
- [ ] Manual: visit `/logout` directly (no JS), verify the confirmation page shows a button that works


imorland:
Docs: flarum/docs#510

PR #4457, opened 2026-03-15
[2.x] feat: add dark mode logo support

rafaucau:

Not every logo works well in both light and dark mode. For example:

| Light mode | Dark mode |
| --- | --- |
| light-mode | dark-mode |

This PR adds an option to set an alternative logo for dark mode.
If no dark mode logo is set, the regular logo is used in dark mode.

**Screenshot**
| Light mode | Dark mode |
| --- | --- |
| light-mode-after | dark-mode-after |

**Necessity**

**Confirmed**

### DIFF
--- a/framework/core/js/src/admin/components/AppearancePage.tsx
+++ b/framework/core/js/src/admin/components/AppearancePage.tsx
@@ -61,6 +61,21 @@ export default class AppearancePage extends AdminPage {
     );
 
     items.add(
+      'logo-dark-mode',
+      <div className="Form-group">
+        <label>{app.translator.trans('core.admin.appearance.logo_dark_mode_heading')}</label>
+        <div className="helpText">{app.translator.trans('core.admin.appearance.logo_dark_mode_text')}</div>
+        <UploadImageButton
+          name="logo-dark-mode"
+          routePath="logo-dark-mode"
+          value={app.data.settings.logo_dark_mode_path}
+          url={app.forum.attribute('logoDarkModeUrl')}
+        />
+      </div>,
+      99
+    );
+
+    items.add(
       'favicon',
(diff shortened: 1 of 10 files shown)


imorland (commented):
Nice, will look over this!

imorland (changes requested):
Nice feature, clean implementation. A few notes:

**Change request:**

The dark logo `` has no base `display: none` rule, so browsers without `:has()` support (notably Firefox < 121) will render both logos simultaneously. Adding:

```less
.Header-logo--dark-mode {
  display: none;
}
```

as a base rule would give correct graceful degradation — those browsers would simply always show the regular logo.

**Notes (non-blocking):**

- The dark logo is not used in HTML mail templates — email dark mode support is too inconsistent for this to be practical, just worth documenting as a known limitation.
- `[data-theme^='dark']` correctly covers `dark-hc` (dark high contrast), so high contrast users get the dark logo. No separate variant is possible, but that's a reasonable limitation for now.
- No tests, consistent with the existing light logo situation — not a blocker.

rafaucau:
@imorland `.Header-logo--dark-mode` is already hidden by default.

https://github.com/rafaucau/flarum-framework/blob/9498eb2101e01b1323ec77de4c2370eae00ac1b5/framework/core/less/common/App.less#L187-L189

So in browsers without `:has()` support, this will gracefully fall back to the current behavior and continue showing the regular logo.

imorland (approved):
Apologies for the confusion — you're right, the base  rule is already there. Change request was incorrect. LGTM\!

imorland (approved):
Apologies for the confusion — you're right, the base display:none rule is already there. Change request was incorrect. LGTM\!

PR #4464, opened 2026-03-16
[2.x] feat(subscriptions): add support for auto follow discussions on create

huoxin233:

This PR introduces a new setting in the `flarum/subscriptions` extension that automatically subscribes users to discussions they created.

The original `followAfterReply` does cover discussion creation also, but not all users will want to follow **every** discussion they reply to, thats why a `followAfterCreate` is needed.

The default preference for `followAfterCreate` is set to true as discussion creators usually want to know when someone replied to their discussion.

Would you consider including this in the core extension? Or should it be extended using another extension, as shown in my extension: https://discuss.flarum.org/d/34680-auto-follow-discussion

There have also been quite a few discussions on this behavior over the years:
  - https://discuss.flarum.org/d/38423-how-to-auto-follow-created-discussions-only-not-replied-ones
  - https://discuss.flarum.org/d/38851-op-has-to-manually-sub-his-own-discussion-to-get-notification-of-replies

Personally, I would definitely like to see this in core as this is something that should come out of the box.

**Screenshot**

image

**Necessity**

**Confirmed**

**Required changes:**


imorland (changes requested):
The implementation is clean and closely mirrors `FollowAfterReply`. A few things to address:

**Default preference should be `false`**

```php
->registerPreference('followAfterCreate', 'boolval', false)
```

This is an opt-in behaviour — existing users who've never saved this preference would have it silently enabled after upgrading. Please change the default to `false`.

**Tests needed**

`FollowAfterReply` has integration test coverage — `FollowAfterCreate` should too. At minimum: a user with `followAfterCreate = true` ends up subscribed after starting a discussion, and a user with `followAfterCreate = false` does not.

**`followAfterCreateLoading` not declared on the component**

This works via dynamic properties the same way `followAfterReplyLoading` does, but it should be explicitly declared on `SettingsPage` alongside the existing loading flags.

**Locale string tense**

`follow_after_create_label` reads "discussions that I created" (past tense). Should match the existing pattern: "discussions that I create".

**StyleCI**

StyleCI is flagging issues — please address before we can merge.

imorland:
Note: this is marked for beta.8 — feature cut-off is midday CET this Sunday. If the above isn't addressed in time it will have to wait until Flarum 2.1.

huoxin233:
All fixed!
Added tests to check FollowAfterReply also

imorland:
Still a failure on the JS - rebasing this from `2.x` might help here...

huoxin233:
Should pass the checks now, the errors are coming from other extension

imorland (approved):

PR #4486, opened 2026-03-22
[2.x] fix: use SlugManager in AuthorFilter to support custom slug drivers

### LINKED ISSUE #4360: [1.x & partially 2.x] Author Filter Incompatible in some cases
#### Flarum 1.x
In Flarum 1.x only the username can be used for filtering discussions and posts. This might clash when the slug driver is set to ID or another custom driver. Custom implementations would have to always use the username for filtering, which might introduce inconsistencies UX wise.

#### Flarum 2.x
In Flarum 2.x, id's and usernames can be used for filtering. Support for filtering by ID's was added in https://github.com/flarum/framework/pull/4019. But that implementation doesn't support other custom user slug drivers.

#### Flarum 1.x

1. Set User Slug Driver to ID
2. Create at least one discussion with the admin (or any other) user
3. Hit http://flarum.localhost/api/discussions?filter[author]=1

#### Flarum 2.x

1. Set User Slug Driver to a custom one
2. Create at least one discussion with the admin (or any other) user
3. Hit http://flarum.localhost/api/discussions?filter[author]=CUSTOMSLUGSCHEMA

#### Flarum 1.x

When hitting `/api/discussions?filter[author]=1`, discussions are returned when user slug driver set to ID

#### Flarum 2.x
When hitting `/api/discussions?filter[author]=CUSTOMSLUGSCHEMA`, discussions are returned when the custom slug driver is selected

### Screenshots

_No response_

### Environment

- Flarum version: 1.8.13 & 2.0.0-beta.6
- Website URL: https://demo.flarum.site/ and https://nightly.flarum.site/
- PHP version: x.y.z

_No response_

### Possible Solution

At least for 2.x, I think we should use the `SlugDriverInterface` to in the end get the ID's of users for filtering. This  should future proof the filter for any custom user slug drivers

For 1.x I'm concerned that functionally wise it would be too much of a breaking change in the public API. Integrations might expect that filtering by user names always works, even when the slug driver is set to ID

### Additional Context

Stumbled across the issue whilst working on https://github.com/glowingblue/flarum-ext-author-filter/issues/1

imorland:
## Summary

- Fixes #4360
- `AuthorFilter` (discussions and posts) previously resolved users via `UserRepository::getIdsForUsernames()`, which only queries by the `username` column — ignoring the active user slug driver
- Users with `id` or `id_with_display_name` slug drivers configured could not filter by author using the expected slug format
- Both filters now use `SlugManager::forResource(User::class)->fromSlug()` so the active driver is respected; unresolvable slugs are silently skipped

## Test plan

- [ ] Existing `author_filter_works` and `author_filter_works_negated` tests still pass (username driver)
- [ ] New `author_filter_works_with_slug_driver` data-provider tests cover `username`, `id`, and `id_with_display_name` drivers for both discussions and posts
- [ ] New `author_filter_negation_works_with_id_slug_driver` tests cover negation with `id` driver for both discussions and posts
- [ ] Manual: switch slug driver in admin, filter discussions/posts by author using the appropriate slug format

### DIFF
--- a/framework/core/src/Discussion/Search/Filter/AuthorFilter.php
+++ b/framework/core/src/Discussion/Search/Filter/AuthorFilter.php
@@ -9,12 +9,14 @@
 
 namespace Flarum\Discussion\Search\Filter;
 
+use Flarum\Http\SlugManager;
 use Flarum\Search\Database\DatabaseSearchState;
 use Flarum\Search\Filter\FilterInterface;
 use Flarum\Search\SearchState;
 use Flarum\Search\ValidateFilterTrait;
-use Flarum\User\UserRepository;
+use Flarum\User\User;
 use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\ModelNotFoundException;
 
 /**
  * @implements FilterInterface<DatabaseSearchState>
@@ -24,7 +26,7 @@ class AuthorFilter implements FilterInterface
     use ValidateFilterTrait;
 
     public function __construct(
-        protected UserRepository $users
+        protected SlugManager $slugManager
     ) {
     }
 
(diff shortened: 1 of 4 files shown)


imorland:
@DavideIadeluca would this resolve the issue with the `Author Filter` extension? #4360

DavideIadeluca:
> @DavideIadeluca would this resolve the issue with the Author Filter extension? https://github.com/flarum/framework/issues/4360

Didn't test the PR but rom briefly glancing at it I think so!

PR #4510, opened 2026-03-31
[2.x] fix(core): suppress PHP warning when extension name lacks vendor/package format

imorland:
## Problem

`Extension::nameToId()` uses array destructuring on `explode('/', $name)`:

```php
[$vendor, $package] = explode('/', $name);
```

If `$name` contains no `/` (e.g. a package with a malformed or missing name), PHP emits:

```
PHP Warning: Undefined array key 1 in Extension.php on line 97
```

This was observed during `php flarum migrate` in certain environments.

## Fix

- `ExtensionManager::getExtensions()` — skip `flarum-extension` packages from `installed.json` whose name has no `/`
- `ExtensionManager::subExtensionConfsFromJson()` — skip sub-extensions whose name has no `/`  
- `Extension::nameToId()` — made robust against a missing slash so any name that does slip through no longer triggers the warning

### DIFF
--- a/framework/core/src/Extension/Extension.php
+++ b/framework/core/src/Extension/Extension.php
@@ -94,7 +94,9 @@ class Extension implements Arrayable
 
     public static function nameToId(string $name): string
     {
-        [$vendor, $package] = explode('/', $name);
+        $parts = explode('/', $name, 2);
+        $vendor = $parts[0];
+        $package = $parts[1] ?? $name;
         $package = str_replace(['flarum-ext-', 'flarum-'], '', $package);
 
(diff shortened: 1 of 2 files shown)


luceos:
With all due respect, I think absorbing packages that have a malformed composer schema should not be quietly allowed. An error is the better option here, but maybe it could have been an actual exception. 

I'm also a bit on the fence as this clearly seems to be instigated by human error. Packagist does not allow you to submit packages or new versions that have malformed composer.json schema's. Please fix the human error and throw an actual exception instead 🙏

PR #4529, opened 2026-04-07
[2.x] [Tags] Drop legacy unused background columns from tags table

### LINKED ISSUE #4527: [1.x & 2.x] [Tags] Unused Columns in tags table
### Current Behavior

There are two columns on the tags table which appear to be unused:
* `tags.background_mode`
* `tags.background_path`

### Steps to Reproduce

1. Install Flarum and up migrate
2. Inspect the tags table 

### Expected Behavior

No unused columns are in the database

### Screenshots

_No response_

### Environment

_n.a._

_n.a._

### Possible Solution

_No response_

### Additional Context

Those columns were added 10 years ago and probably just forgotten. Or is there something I've missed which makes the existence / addition of those columns in core necessary?

Abdooo2235:
## Summary
This PR removes two legacy columns from the `tags` table that are no longer used by current tags flows:
- `background_path`
- `background_mode`

Fixes #4527.

## What changed
- Added migration using `Migration::dropColumns()` to remove both legacy columns.
- Removed `backgroundUrl` and `backgroundMode` from the tags API resource.
- Removed stale background accessors from the frontend `Tag` model.
- Removed stale references from the tag factory and model docblocks.
- Added an integration schema test asserting those columns are absent after migrations.

## Backward compatibility
- This is an intentional clean break for 2.x pre-stable, aligned with review feedback.
- Existing data in removed columns is dropped as part of schema cleanup.

## Scope note
An unrelated local change in `extensions/tags/js/tsconfig.json` was intentionally excluded from this PR to keep the fix focused.

## Testing
- Added: `extensions/tags/tests/integration/api/tags/SchemaTest.php`
- Local test execution in this environment is blocked because `phpunit` is not available (`sh: 1: phpunit: not found`).

### DIFF
--- a/extensions/tags/src/Api/Resource/TagResource.php
+++ b/extensions/tags/src/Api/Resource/TagResource.php
@@ -112,10 +112,6 @@ class TagResource extends AbstractDatabaseResource
             Schema\Boolean::make('isRestricted')
                 ->writableOnUpdate()
                 ->visible(fn (Tag $tag, FlarumContext $context) => $context->getActor()->isAdmin()),
-            Schema\Str::make('backgroundUrl')
-                ->get(fn (Tag $tag) => $tag->getAttribute('background_path')),
-            Schema\Str::make('backgroundMode')
-                ->get(fn (Tag $tag) => $tag->getAttribute('background_mode')),
             Schema\Integer::make('discussionCount'),
             Schema\Integer::make('position')
                 ->nullable(),
--- a/extensions/tags/src/Api/Resource/TagResource.php
+++ b/extensions/tags/src/Api/Resource/TagResource.php
@@ -113,8 +113,9 @@ class TagResource extends AbstractDatabaseResource
                 ->writableOnUpdate()
                 ->visible(fn (Tag $tag, FlarumContext $context) => $context->getActor()->isAdmin()),
             Schema\Str::make('backgroundUrl')
-                ->get(fn (Tag $tag) => $tag->background_path),
-            Schema\Str::make('backgroundMode'),
+                ->get(fn (Tag $tag) => $tag->getAttribute('background_path')),
+            Schema\Str::make('backgroundMode')
+                ->get(fn (Tag $tag) => $tag->getAttribute('background_mode')),
             Schema\Integer::make('discussionCount'),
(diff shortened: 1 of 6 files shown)


DavideIadeluca (commented):
Thanks for picking up the issue I opened. Have you considered using Flarum’s migration helper? The migration in the PR feels a bit more complex than necessary and could likely be simplified along these lines:

```php
use Flarum\Database\Migration;

return Migration::dropColumns('tags', [
    'background_path' => [..],
    'background_mode' => [..],
]);
```

Note that with `Migration::dropColumns()` you still need to specify the column constraints and metadata, similar to what you would define in the `up` migration.

Another thing that stood out in the PR is that the attributes are still present. If `background_mode` and `background_path` are no longer required, wouldn’t this effectively become dead code? It might be better to remove them entirely and go for a clean break.

Since Flarum 2 hasn’t been tagged as stable yet, a breaking change like this should still be acceptable without creating backwards compatibility concerns.

extensions/tags/src/Api/Resource/TagResource.php
```diff
@@ -113,8 +113,9 @@ public function fields(): array
                 ->writableOnUpdate()
                 ->visible(fn (Tag $tag, FlarumContext $context) => $context->getActor()->isAdmin()),
             Schema\Str::make('backgroundUrl')
```
luceos: This probably also needs to be removed? The attribute doesn't exist and the getter was removed.

luceos (changes requested):

Abdooo2235:
@DavideIadeluca thanks a lot for the thoughtful review and for opening this issue in the first place.

Great call on both points. I updated the PR accordingly:

- Switched the migration to use `Migration::dropColumns()` with full column definitions.
- Removed the leftover `backgroundUrl` / `backgroundMode` attributes from the API resource.
- Removed the corresponding stale accessors from the frontend `Tag` model so there’s no dead code path left.

I agree with your reasoning about 2.x still being pre-stable, so I went with the clean break approach here.

I also reran the tags integration suite after the update:
`composer test:integration` → **OK (127 tests, 213 assertions)**.

Really appreciate the guidance.

imorland (approved):

PR #4544, opened 2026-04-11
[3.x] Migrate to Intervention Image v4

olivervogel:
This PR makes the necessary changes to migrate to the just released [Intervention Image Version 4](https://github.com/Intervention/image/releases/tag/4.0.0).

Please note that the new version requires at least PHP 8.3.

I recommend a thorough review. However, I think this is a good start and takes care of most of the work for the update. 

If you have any questions, feel free to reach out to me.

PS: For your reference, here is the [upgrade guide](https://image.intervention.io/v4/getting-started/upgrade)

**Necessity**

**Confirmed**

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -141,7 +141,7 @@
         "illuminate/support": "^13.0",
         "illuminate/validation": "^13.0",
         "illuminate/view": "^13.0",
-        "intervention/image": "^3.2",
+        "intervention/image": "^4.0",
         "jenssegers/agent": "^2.6",
         "laminas/laminas-diactoros": "^3.0",
         "laminas/laminas-httphandlerrunner": "^2.6",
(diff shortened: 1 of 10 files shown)


imorland:
Thanks for this Oliver — the migration work looks good and this is something we definitely want to do.

The blocker is that Intervention Image is used directly by extensions in the ecosystem — fof/upload being a notable example. The v3 → v4 API changes mean this would be a silent breaking change for any extension that instantiates `ImageManager` or uses v3 methods directly, and we can't coordinate that at a minor version bump.

We're going to add this to the 3.0 roadmap where we can signal the breaking change properly and give extension authors time to migrate alongside us. We'll keep this PR open for reference when that work begins.

olivervogel:
> We're going to add this to the 3.0 roadmap where we can signal the breaking change properly and give extension authors time to migrate alongside us. We'll keep this PR open for reference when that work begins.

Okay, I assumed it would still work in 2.0, since it's only in beta so far.

PR #4569, opened 2026-04-14
[2.x] feat: support opcache flushing

luceos:
### Why

PHP's OPcache compiles and caches bytecode for every file it loads. When Composer installs, updates, or removes packages — whether triggered through the Flarum admin UI or directly from the CLI — the files in vendor/ change on disk. PHP-FPM workers, however, keep serving the stale cached bytecode for those paths until they are recycled. This causes class-not-found errors, method signature mismatches, or silently running old code immediately after a package operation.

### How it is solved

The fix uses a flag file (storage/cache/clear-opcache) as a signal between the process that ran Composer (CLI or queue worker) and the PHP-FPM workers that serve HTTP requests. The reason we need to push this towards HTTP is because cli has no access to the opcache. The flag is written after any Composer operation and consumed on the next incoming request.

Three components work together:

  1. flarum/composer-plugin (new package in php-packages/composer-plugin/)

  A Composer plugin that subscribes to the post-autoload-dump event. This event fires after Composer regenerates the autoloader — the point at which the vendor directory is fully updated. When triggered, the plugin writes the flag file. This covers any Composer operation run directly from the CLI (composer
   require, composer update, composer remove).

  2. ClearOPCacheAfterUpdate listener (extension manager)

  The Flarum extension manager runs Composer in-process using a temporary vendor directory (temp-vendor), which means the plugin above is not loaded during those operations. To cover this path, the extension manager listens to its own events (Installed, Updated, Removed, FlarumUpdated) and writes the same 
  flag file from PHP when any of those fire.
                                                                                                                                                                                                                                                                                                                   
  3. ClearOPCache middleware (core)

  Added early in the forum middleware stack. On each incoming HTTP request it checks whether the flag file exists. If it does, it calls opcache_reset() to clear all cached bytecode in the current FPM worker, then deletes the flag file. Subsequent workers that receive requests after the file has been removed skip the reset.

---

There are some considerations to this PR. 

- Without the composer plugin opcache will not be properly cleared when composer commands are being executed. Using the plugin signals to Flarum to clear the opcache as soon as possible.
- The Middleware is placed in core, not in extension manager, because outside interaction from cli using composer will still affect the opcache.
- The extension manager is most likely the culprit to cause stale opcache. As such this extension receives its own listener to signal an opcache clear.
- Code has been written to exit early, to reduce performance hits.
- This solution will have zero effect on scalable environments, but then again I expect no one to be running extension manager or composer commands in those environments anyway.


_Code was written 100% by my hands, summary generated with the help of Claude._ Let me know if you have any questions or things need clarification.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -110,6 +110,7 @@
         "flarum/suspend": "self.version",
         "flarum/tags": "self.version",
         "flarum/messages": "self.version",
+        "flarum/composer-plugin": "self.version",
         "flarum/phpstan": "self.version",
         "flarum/testing": "self.version"
     },
@@ -122,6 +123,7 @@
         "doctrine/dbal": "^3.6.2",
         "dragonmantank/cron-expression": "^3.3",
         "fakerphp/faker": "^1.9.1",
+        "flarum/composer-plugin": "2.x-dev",
         "flarum/json-api-server": "^0.1.0",
         "franzl/whoops-middleware": "2.0",
         "guzzlehttp/guzzle": "*",
--- a/extensions/package-manager/src/ExtensionManagerServiceProvider.php
+++ b/extensions/package-manager/src/ExtensionManagerServiceProvider.php
@@ -15,8 +15,11 @@ use Composer\Util\Platform;
 use Flarum\Extension\ExtensionManager;
 use Flarum\ExtensionManager\Composer\ComposerAdapter;
 use Flarum\ExtensionManager\Event\FlarumUpdated;
+use Flarum\ExtensionManager\Extension\Event\Installed;
+use Flarum\ExtensionManager\Extension\Event\Removed;
 use Flarum\ExtensionManager\Extension\Event\Updated;
 use Flarum\ExtensionManager\Listener\ClearCacheAfterUpdate;
+use Flarum\ExtensionManager\Listener\ClearOPCacheAfterUpdate;
 use Flarum\ExtensionManager\Listener\ReCheckForUpdates;
 use Flarum\Foundation\AbstractServiceProvider;
 use Flarum\Foundation\Paths;
@@ -100,5 +103,6 @@ class ExtensionManagerServiceProvider extends AbstractServiceProvider
 
         $events->listen(FlarumUpdated::class, ClearCacheAfterUpdate::class);
         $events->listen([FlarumUpdated::class, Updated::class], ReCheckForUpdates::class);
+        $events->listen([Installed::class, Updated::class, Removed::class, FlarumUpdated::class], ClearOPCacheAfterUpdate::class);
     }
 }
--- a/extensions/package-manager/src/Listener/ClearOPCacheAfterUpdate.php
+++ b/extensions/package-manager/src/Listener/ClearOPCacheAfterUpdate.php
@@ -1,0 +1,26 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Flarum\ExtensionManager\Listener;
+
+use Flarum\Foundation\Paths;
+use Flarum\Http\Middleware\ClearOPCache;
+
+readonly class ClearOPCacheAfterUpdate
+{
+    public function __construct(private readonly Paths $paths)
+    {
+    }
+
+    public function handle(): void
+    {
+        $path = $this->paths->storage . ClearOPCache::PATH;
+
+        if (file_exists($path) || ! function_exists('opcache_reset')) {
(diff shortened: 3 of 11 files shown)


luceos:
There's one scenario I still need to look into, and that is during installation. Related to this discussion: https://discuss.flarum.org/d/39074-php-message-php-fatal-error-on-install, people seem to be having an issue immediately after running create-project, and before the installation completed.

luceos:
Installation is now also supported and should resolve the issue the people had mentioned on discuss.

imorland:
Hey Daniel! Love the approach here — the flag file as a signal between CLI/queue and PHP-FPM workers is exactly the right way to handle this.

A few things I noticed while going through the diff:

**Middleware only on the forum stack** — `ClearOPCache` is added to `flarum.forum.middleware` but not the admin or API stacks. In practice the most common post-update flow is an admin heading straight to the admin panel, so they'd never hit the forum stack first and the cache would stay stale. Needs to land in all three (or a shared location).

**Composer plugin fallback path** — In `ComposerPlugin::flag()` the fallback when `ClearOPCache` isn't autoloaded produces `/storage/cache/clear-opcache` as an absolute path from the filesystem root, which won't resolve to anything useful. Since the plugin can't reliably depend on the Flarum autoloader being available at that point, I'd just inline the path constant as a string literal in the plugin to remove the cross-package dependency entirely.

**`file_exists` check in `ClearOPCacheAfterUpdate`** — The early return fires if the file already exists *or* if `opcache_reset` doesn't exist. The first condition seems unintentional — if the flag is already set from a previous operation that hasn't been consumed yet, that's fine, you don't need to skip. Was that meant to be just `! function_exists('opcache_reset')`?

Minor ones: the `while(true)` path-walking fallback in `ClearOPCache::path()` could use a depth limit or a note about its assumptions, and silent `@file_put_contents` failures in the listener/plugin mean a permissions problem would go completely unnoticed.

None of this is hard to fix — the fundamentals are solid!

luceos:
> **Middleware only on the forum stack** — `ClearOPCache` is added to `flarum.forum.middleware` but not the admin or API stacks. In practice the most common post-update flow is an admin heading straight to the admin panel, so they'd never hit the forum stack first and the cache would stay stale. Needs to land in all three (or a shared location).

Yeah 🤦 completely forgot to add them, they were on my list. Will patch them in as well. Was hoping we would have a generic http middleware stack by now.

> **Composer plugin fallback path** — In `ComposerPlugin::flag()` the fallback when `ClearOPCache` isn't autoloaded produces `/storage/cache/clear-opcache` as an absolute path from the filesystem root, which won't resolve to anything useful. Since the plugin can't reliably depend on the Flarum autoloader being available at that point, I'd just inline the path constant as a string literal in the plugin to remove the cross-package dependency entirely.

This is not true. The `$flagFile` variable prepends the `dirname($vendorDir)` meaning one path above `vendor`.

> **`file_exists` check in `ClearOPCacheAfterUpdate`** — The early return fires if the file already exists _or_ if `opcache_reset` doesn't exist. The first condition seems unintentional — if the flag is already set from a previous operation that hasn't been consumed yet, that's fine, you don't need to skip. Was that meant to be just `! function_exists('opcache_reset')`?

The `ClearOPCacheAfterUpdate` is a listener that sets the file. If the file already exists from a previous attempt, it would cause the listener to keep overwriting the file for no reason. So the first conditional absolutely has to stay, the second one is an additional measure to prevent writing the file in the first place if opcache isn't in use.

Come to think of it, it would probably be smarter to switch those around, because the file_exists will do a filesystem interaction, whereas the function_exists is just a php check.

> Minor ones: the `while(true)` path-walking fallback in `ClearOPCache::path()` could use a depth limit or a note about its assumptions, and silent `@file_put_contents` failures in the listener/plugin mean a permissions problem would go completely unnoticed.

There's a depth limit assumption. According to the docs dirname will return `.` if it no longer can go up. But yeah, I can probably limit the path walking to up to 6 or something to reflect `workbench/framework/core/src/Http/Middleware`.

I don't think we should look at file permissions in this PR. That would be something else entirely as it opens a can of worms I'm not looking to invest time into right now 😊 

> None of this is hard to fix — the fundamentals are solid!

luceos:
I've applied the patches.
One thing to notice is that the `2.x-dev` for the plugin reference needs to be a tag in the composer.json of core, not sure how to get that right while it's still not tagged.

PR #4582, opened 2026-04-17
fix: misalignment caused by wrong placement of badges

luceos:
Bug happens when replying,where the badges shift to the left of the avatar. This is caused by a wrong implementation in the component as they are positioned elsewhere based on username, not avatar element.

This is similar to the actual post implementation. The badges are weirdly aligned from the author username, not the avatar.


imorland (changes requested):
Thanks for the PR @luceos 

Unfortunately this change breaks the mobile view:

Image

This will need an accompanying styling change as well....

imorland:
@luceos will you have time to address this ahead of rc.2 in the coming days?

luceos:
It's done @imorland. 

Only now saw the addition pr by @huoxin233, should've checked sooner 🙈

huoxin233:
> It's done @imorland.
> 
> Only now saw the addition pr by @huoxin233, should've checked sooner 🙈

image

Your one didnt cover the part for duplicate avatar when editing, so probably I will do a merge also 🫣

https://github.com/flarum/framework/pull/4610/changes#diff-589e03833214a5b7c4920f93deceb20a05fdf719f55e8eb3948ed8e03979f912R409

---

Also do yall think its fine that the text is actually not align? (when doing new post or editing)
image

imorland (approved):

PR #4589, opened 2026-04-19
fix: halt momentum scrolling so anchorScroll's scrollTop is honored on iOS Safari

### LINKED ISSUE #4587: PostStream loses scroll position during iOS Safari momentum scroll when loadPrevious() prepends posts
### Summary

On iOS Safari, when a user flick-scrolls upward inside a long discussion and `PostStream.loadPostsIfNeeded()` triggers `stream.loadPrevious()` to prepend older posts, the viewport jumps — often all the way to the top of the discussion. The faster the flick, the more severe the jump; a very slow, deliberate scroll works correctly. Desktop browsers are unaffected.

### Affected versions

- Reproduced on `2.x-dev` (current).
- **Not reproducible on `v2.0.0-beta.8`** — see "Why this only appears post-#4561" below.
- Expected to affect RC1 and all subsequent 2.x releases once the viewport change in #4561 is in.

### Steps to reproduce

1. Open a long discussion on an iPhone (Safari). The user should already have read through it so the "end" is the target post.
2. Flick-scroll upward briskly to page back through older posts.
3. Keep scrolling until the top of the currently-loaded window is about to leave the viewport and `loadPrevious()` fires.

**Observed:** the viewport jumps sharply — sometimes to the beginning of the discussion — as the prepended batch lands.

**Expected:** the viewport stays anchored on whichever posts the user was looking at (same behavior as desktop).

### Why this only appears post-#4561

#4561 ("perf: frontend performance improvements") changed the viewport meta from

```html
<meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1, minimum-scale=1">
```

to

```html
<meta name="viewport" content="width=device-width, initial-scale=1">
```

which is a correct accessibility fix — users can now pinch-zoom on mobile. On beta.8 (with `maximum-scale=1, minimum-scale=1`), iOS Safari runs a simpler scroll pipeline in which programmatic `scrollTop` writes during momentum scroll *tend* to be honored; once user-zoom is enabled, iOS's zoom-and-scroll coordination layer kicks in and programmatic `scrollTop` writes during in-flight momentum are silently dropped. This is the same well-known WebKit behavior that has bitten many virtualized-list libraries; the viewport change in #4561 merely unmasked it in Flarum.

The accessibility win from #4561 should stay; the PostStream jump needs a client-side fix so momentum-scroll pagination works regardless of zoom configuration.

### Root cause

`PostStreamState.loadPage()` wraps the prepend redraw in `anchorScroll()`:

```js
// framework/core/js/src/forum/states/PostStreamState.ts
const redraw = () => {
  if (start < this.visibleStart || end > this.visibleEnd) return;
  const anchorIndex = backwards ? this.visibleEnd - 1 : this.visibleStart;
  anchorScroll(`.PostStream-item[data-index=\"${anchorIndex}\"]`, m.redraw.sync);
};
```

`anchorScroll()` itself is:

```js
// framework/core/js/src/common/utils/anchorScroll.js
export default function anchorScroll(element, callback) {
  const $window = $(window);
  const relativeScroll = $(element).offset().top - $window.scrollTop();
  callback();
  $window.scrollTop($(element).offset().top - relativeScroll);
}
```

On **iOS Safari during momentum (inertial) scrolling with user-zoom enabled, programmatic writes to `window.scrollTop` are silently ignored** — the in-flight momentum animation wins. As a result:

1. The batch of placeholders is prepended to `.PostStream`.
2. The final `$window.scrollTop(...)` call is dropped.
3. The momentum animation continues from the *old* absolute scroll offset, which now points far earlier in the discussion because the content above the viewport has grown by ~20 items — producing the observed jump.

Related reports with the same pattern:

- [react-window #122 — Safari scroll update while momentum is ongoing](https://github.com/bvaughn/react-window/issues/122)
- [react-virtualized #59 — Scrolling not inertial on mobile Safari](https://github.com/bvaughn/react-virtualized/issues/59)
- [CodeMirror 5 #6858 — Interrupted momentum scrolls cause a scroll jump](https://github.com/codemirror/codemirror5/issues/6858)
- Writeup with a concrete workaround: https://lovemewithoutall.github.io/it/setting-scrollTop-on-momentum-scroll/

CSS scroll-anchoring (`overflow-anchor`) would normally be a safety net, but WebKit only shipped it in Safari 26 ([WebKit tracker #171099](https://bugs.webkit.org/show_bug.cgi?id=171099), [caniuse](https://caniuse.com/css-overflow-anchor)). A grep of the Flarum JS tree for `overflow-anchor` returns no matches, so there is no fallback today either.

### Suggested fix

Route the restore through `window.scrollTo(x, y)` instead of jQuery's `$window.scrollTop(y)`. `window.scrollTo` goes through a different WebKit code path that DOES halt the in-flight momentum and apply the new position, while `scrollTop` writes are silently dropped:

```diff
-  $window.scrollTop($(element).offset().top - relativeScroll);
+  window.scrollTo(window.scrollX, $(element).offset().top - relativeScroll);
```

This is a one-line surgical change with no UX or layout side effects on any platform — on non-iOS browsers it's equivalent behavior. PR to follow.

### Environment

- Flarum `2.x-dev` (staging).
- Affected device: iPhone, mobile Safari.
- Desktop Chrome, Firefox, and Safari are all fine.
- Custom extensions installed on the reporting forum do not touch the post-stream scroll path and are not involved; reproduced with the symptom caused by core behavior alone.

ekumanov:
Closes #4587.

### What

Replace jQuery's `$window.scrollTop(y)` with `window.scrollTo(window.scrollX, y)` inside `anchorScroll` so iOS Safari's in-flight momentum scroll is halted and the restore actually lands.

```diff
 export default function anchorScroll(element, callback) {
   const $window = $(window);
   const relativeScroll = $(element).offset().top - $window.scrollTop();

   callback();

-  $window.scrollTop($(element).offset().top - relativeScroll);
+  window.scrollTo(window.scrollX, $(element).offset().top - relativeScroll);
 }
```

### Why

See the linked issue for the full diagnosis. Short version: iOS Safari silently drops programmatic `scrollTop` writes while a momentum scroll is in flight. `PostStreamState.loadPage()` relies on `anchorScroll` to hold the viewport steady when 20 posts get prepended by `loadPrevious()`; when a user flick-scrolls upward in a long discussion the restore is dropped and the page visibly jumps — often all the way to the beginning of the discussion.

`window.scrollTo(x, y)` goes through a different WebKit code path that DOES halt the in-flight momentum animation and apply the new position. On non-iOS platforms it's equivalent to the previous `scrollTop` write, so this is a no-op there.

### Why this surfaced now

#4561 removed `maximum-scale=1, minimum-scale=1` from the viewport meta (correctly — it's an accessibility fix). That unmasked this latent WebKit bug: with user-zoom disabled, iOS ran a simpler scroll pipeline where the `scrollTop` writes tended to be honored; with user-zoom enabled, the zoom/scroll coordination layer drops programmatic `scrollTop` writes during in-flight momentum. The fix in this PR is compatible with both viewport configurations.

### Alternatives considered

- **`overflow: hidden` toggle on `<html>` for one tick.** Works on iOS, but briefly causes the page's overflow state to change which can interact poorly with position: sticky headers and any custom user CSS hooked off `document.documentElement.style.overflow`. Rejected in favor of the narrower `scrollTo` change.
- **CSS `overflow-anchor`.** Good defense-in-depth, but Safari only shipped it in 26; useless for the bulk of current iOS users. Could be added as a follow-up independently of this fix.
- **Debouncing `loadPrevious()` until `scrollend`.** Cleaner UX but delays the prepend and reduces the "feels instant" perception; also doesn't help if the fix needs to backport to older iOS that lacks `scrollend`.

### Scope

- Only one file changed in source (`framework/core/js/src/common/utils/anchorScroll.js`, +9 / -1 including an explanatory comment).
- `anchorScroll` has a single callsite in source: `PostStreamState.loadPage()` (confirmed with grep).
- No TS typings change; no new dependency.
- Bundled output committed in a follow-up commit, matching this repo's convention.

### Testing

- [x] `yarn build` in `framework/core/js/` succeeds.
- [x] **Real-device verified on iPhone Safari** running against a Flarum `v2.0.0-beta.8` staging forum patched to match the post-#4561 viewport. The PostStream no longer jumps during flick-scroll-driven `loadPrevious()`. Symptom-for-symptom fix of the repro from #4587.
- Desktop browsers: sanity-checked that the `scrollTo` replacement does not visibly affect normal `loadPrevious()` behavior.

Marking ready for review.


ekumanov:
Closing — I can no longer reproduce the underlying bug (#4587) on `v2.0.0-rc.1` final. See the linked issue for the non-repro analysis.

The branch and diff are preserved on my fork (`ekumanov/framework:fix/ios-momentum-scroll-anchor`) and can be revived as-is if the bug resurfaces. The fix itself is a single-line swap from `$window.scrollTop(y)` to `window.scrollTo(window.scrollX, y)` in `anchorScroll.js` — well-scoped and re-applyable in minutes.

Issue #4502, opened 2026-03-25
[2.x] Discussion list 3x slower in 2.0 beta 8 vs 1.8 — N+1 queries in tag-permission system (287 vs 60 queries)


ekumanov:
## Summary

The discussion list page is ~3x slower on Flarum 2.0.0-beta.8 compared to 1.8.14, measured on **identical hardware with the same database**. The root cause is the tag-permission system generating 287 queries (vs 60 on 1.8), with severe N+1 patterns in `groups` and `discussions` tables.

## Environment

Both instances run on the **same VPS** (2-core Intel, 4GB RAM, PHP 8.4.18, MySQL 8.4.8, OPcache+JIT, InnoDB buffer pool 512M — entire DB fits in memory).

| | Flarum 1.8.14 | Flarum 2.0.0-beta.8 |
|---|---|---|
| Extensions enabled | 50 | 32 |
| Cache driver | file | Redis |
| Database | ~2,200 discussions, ~54,000 posts, ~900 users, 6 restricted tags | same DB |

<details>
<summary>Extension lists</summary>

**1.8.14 (50 extensions):**
flarum: approval, bbcode, emoji, flags, lang-english, likes, lock, markdown, mentions, nicknames, statistics, sticky, subscriptions, suspend, tags
fof: anti-spam, ban-ips, byobu, filter, follow-tags, formatting, forum-statistics-widget, ignore-users, links, merge-discussions, nightmode, pages, polls, sitemap, upload, user-bio, user-directory
other: afrux/forum-widgets-core, afrux/online-users-widget, askvortsov/markdown-tables, askvortsov/rich-text, blomstra/database-queue, blomstra/mark-unread, blomstra/spam-prevention, clarkwinkelmann/advanced-search-highlight, clarkwinkelmann/post-stream-search, clarkwinkelmann/selective-mediaembed, davwheat/session-keepalive, ekumanov/inline-audio, ekumanov/new-posts-notice, flamarkt/backoffice, kilowhat/rich-embeds, rob006/last-post-avatar, sycho/move-posts, v17development/seo

**2.0.0-beta.8 (32 extensions):**
flarum: approval, bbcode, emoji, flags, lang-english, likes, lock, markdown, mentions, nicknames, realtime, statistics, sticky, subscriptions, suspend, tags
fof: anti-spam, byobu, follow-tags, formatting, ignore-users, links, merge-discussions, move-posts, pages, rich-text, sitemap, upload, user-bio, user-directory
[...]

PR #4649, opened 2026-05-09
[2.x] fix(tags): restore subquery for permitted tag IDs (regression from #4502)

### LINKED ISSUE #4605: Tag visibility-permission scope regresses by 30+ extra DB queries per discussion list (2.0-rc.1 vs 1.8)
## Problem
On a forum with the `flarum/tags` extension enabled, listing discussions on 2.0-rc.1 issues **dozens more SQL queries than 1.8** for the same content. We measured **120 Execute statements on `/api/discussions?page[limit]=20`** under 2.0-rc.1 vs **29 on 1.8.16** with identical schema and data — a 4.1× query-count regression. The two new query templates that account for almost all of the regression:

| Template | 2.0 calls | 1.8 calls |
|---|---:|---:|
| `SELECT permission FROM group_permission WHERE group_id IN (?)` | 36 | 1 |
| `SELECT id, is_restricted FROM tags` | 6 | 0 |

## Root cause

`Flarum\Tags\Tag::scopeWhereHasPermission` was rewritten between 1.x and 2.0:

**1.8** (`extensions/tags/src/Tag.php#L243` in flarum/framework v1.8.16): builds a single SQL subquery that resolves permitted tag IDs inline. The actor's `getPermissions()` is called once (and is User-instance-cached); no other DB hit per scope application.

**2.0-rc.1** ([extensions/tags/src/Tag.php#L233-L289](https://github.com/flarum/framework/blob/main/extensions/tags/src/Tag.php#L233)): pulls user permissions, fetches **all** tags from the DB into PHP, filters in PHP, then builds an `IN`-clause. The result is wrapped in a `WeakMap<User, array<string,mixed>>` cache (`Tag::$permittedTagIdCache`).

```php
$userCache = static::$permittedTagIdCache[$user] ?? [];
if (array_key_exists($currPermission, $userCache)) return $userCache[$currPermission];
// cache miss path:
$user->getPermissions();                                        // → 1 SELECT (per User instance)
$allTags = static::query()->select(['id', 'is_restricted'])->get();  // → 1 SELECT (per cache miss)
$permitted = $allTags->filter(...)->pluck('id')->all();
// builds IN-clause from $permitted
```

The WeakMap cache is keyed by the **User object instance**. Every fresh `User` instance loaded by Eloquent — and a discussion list loads many: `discussion.user`, `discussion.lastPostedUser`, `firstPost.user`, `lastPostedUser.groups` — gets its own cache slot. Even though most of those users share an identical group composition (e.g. `[guest, member, group6]`), each instance independently misses the cache and re-runs both queries.

We captured the parameters of the 30 `group_permission` selects in one request — they are **all `IN (2, 3, 6)`** (the actor's groups), repeated 30 times because 30 distinct User instances each independently triggered the cache miss:

```
30× group_id IN (2, 3, 6)    ← actor's groups, repeated for every fresh User instance
 3× group_id IN (2, 3)
 1× group_id IN (2, 3, 6, 7)
 1× group_id IN (2, 3, 4, 6)
 1× group_id IN (2, 3, 4)
 1× group_id IN (2)          ← guest
```

## Reproducer

Stock 2.0-rc.1 install with `flarum/tags`, ≥20 discussions across multiple tags (some restricted, some not), at least one non-trivial group:

```bash
mysql -e "SET GLOBAL general_log='ON'; SET GLOBAL general_log_file='/tmp/g.log';"
curl -s -o /dev/null 'http://your-forum/api/discussions?page%5Blimit%5D=20'
mysql -e "SET GLOBAL general_log='OFF';"
grep -c 'group_permission' /tmp/g.log   # expect ~30+
grep -c 'select `id`, `is_restricted` from `tags`' /tmp/g.log   # expect ~6
```

## Suggested fix directions

1. **Restore the 1.8 subquery approach** (preferred). The per-PHP filter is the regression — 1.8's all-in-SQL path was bounded to one query per request regardless of how many User instances exist.
2. **Re-key the cache by the group-id-set string** (e.g. `"2,3,6"` → permission list) rather than by User instance. Multiple User instances sharing groups would then share the cache. Trivial change; almost as good as #1 in practice.
3. **Preload `Permission::all()` at request boot** into a request-scoped registry. Permissions are a small table; a single load eliminates per-User-instance queries entirely.

## Impact

On a real production forum (~927 users, ~2266 discussions, ~54k posts), `/forum/` HTML render: 1.8.16 = 0.38s, 2.0-rc.1 = 1.27s on **identical hardware**. Removing this regression alone (independent of an unrelated polls extension regression filed separately at FriendsOfFlarum/polls) should drop 2.0 to ~0.85–0.90s based on the query-count delta and observed per-query cost.

Profiling was done with [`php-spx`](https://github.com/NoiseByNorthwest/php-spx) cookie-gated, plus MySQL `general_log` for query capture. Happy to share raw flame graphs if useful.

imorland:
## Summary

Fixes #4605. \`Tag::scopeWhereHasPermission\` regressed in #4502 from a SQL subquery to a PHP-side filter cached in a \`WeakMap<User, …>\`. When the actor User instance is recreated mid-request (which happens in several places), the cache misses and re-runs:

\`\`\`sql
SELECT permission FROM group_permission WHERE group_id IN (?)
SELECT id, is_restricted FROM tags
\`\`\`

On a real forum (~20 discussions, multiple authors with varied groups) this produced ~30 redundant \`group_permission\` lookups and ~6 redundant tag enumerations per \`/api/discussions\` request, contributing measurably to page-render latency (1.27s on 2.0-rc.1 vs 0.38s on 1.8.16 on identical hardware).

## Diagnosis

The reporter (#4605) instrumented production traffic and identified two redundant query templates dominating the regression:

| Template | 2.0-rc.1 | 1.8.16 |
|---|---:|---:|
| \`SELECT permission FROM group_permission WHERE group_id IN (?)\` | 36 | 1 |
| \`SELECT id, is_restricted FROM tags\` | 6 | 0 |

Both come from the cache-miss path in \`resolvePermittedTagIds\`. The WeakMap is keyed by the actor User *object*, so any flow that re-loads the actor (or compares against a fresh instance) defeats the cache.

## Fix

Restore the SQL subquery approach that 1.x used:

\`\`\`php
// tags.id IN (SELECT id FROM tags AS perm_tags WHERE …)
\`\`\`

The DB does the filtering inline. No PHP-side cache, no enumeration of the entire tags table, no per-User-instance miss path. The subquery is cheap (a small filtered select on the tags table) and runs once per visibility-scope application — exactly what 1.x did.

\`flushPermittedTagCache()\` is dropped — it was scaffolding for the WeakMap (introduced in #4502) and the only caller was the test cleaning up after it.

## Tests

**Regression guard** — \`PermissionScopeQueryCountTest\` exercises the exact bug shape: 10 distinct non-admin authors with varied group memberships each authoring a discussion, then a non-admin actor lists the discussions. Asserts the unconstrained \`select id, is_restricted from tags\` query — the signature of #4605 — runs at most once per request.

- Pre-fix: 4 such enumerations on this fixture.
- Post-fix: 0.

**Visibility correctness** — Added 3 focused tests in \`DiscussionVisibilityTest\` to pin the parent/child clause property at single-tag granularity:

- \`permission_on_child_does_not_grant_visibility_when_parent_is_off_limits\` — restricted-child + restricted-parent + perm on child only → hidden
- \`permission_on_child_grants_visibility_when_parent_is_unrestricted\` — restricted-child + unrestricted-parent + perm on child → visible
- \`root_restricted_tag_with_explicit_permission_is_visible\` — restricted-root + perm → visible (the \`orWhereNull('parent_id')\` branch)

All three pass on **both** the un-fixed and fix branches, confirming they describe properties that hold under the existing implementation as well as the new one.

**Query-trace equivalence** — Captured the visibility SQL emitted by \`/api/discussions\` against an identical fixture on (a) the un-fixed code and (b) this fix. The outer parent/child predicate structure is byte-identical, and the permitted-set definitions (\`unrestricted ∪ explicit-perm restricted\` via PHP filter on un-fixed, vs \`is_restricted = false ∪ (is_restricted = true AND id IN explicit-perms)\` via SQL subquery on fixed) are mathematically equivalent. Full traces in [the analysis comment](https://github.com/flarum/framework/pull/4649#issuecomment-4412012639).

**Suite results** — 144/144 tags integration tests pass (1 pre-existing unrelated failure on \`ShowTest::can_show_tag_with_url_decoded_utf8_slug\` is present on \`2.x\` baseline). 673/673 core integration tests pass. PHPStan clean.

## Performance impact

The reporter's measurements suggest restoring 1.x's query count drops \`/forum/\` HTML render from 1.27s to ~0.85s on their forum (independent of an unrelated polls regression). The remaining headroom belongs to other work outside this PR.

### DIFF
--- a/extensions/tags/src/Tag.php
+++ b/extensions/tags/src/Tag.php
@@ -21,6 +21,7 @@ use Illuminate\Database\Eloquent\Relations\BelongsTo;
 use Illuminate\Database\Eloquent\Relations\BelongsToMany;
 use Illuminate\Database\Eloquent\Relations\HasMany;
 use Illuminate\Database\Eloquent\Relations\HasOne;
+use Illuminate\Database\Query\Builder as QueryBuilder;
 
 /**
  * @property int $id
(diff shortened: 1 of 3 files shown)


imorland:
Update from offline review: did two extra confidence-raising checks before recommending merge.

### 1. Side-by-side query trace

Captured the visibility-relevant SQL emitted by \`/api/discussions\` as a non-admin user (with explicit perm on restricted tags 5, 8, 11) on (a) the un-fixed WeakMap code and (b) the fix branch, against an identical fixture.

**Un-fixed (current 2.x):**
\`\`\`sql
-- 3× this leaks out (the regression — WeakMap cache miss):
SELECT id, is_restricted FROM tags

-- final discussion query embeds the PHP-resolved permitted IDs inline:
WHERE tags.id IN (1,2,3,4,5,7,8,9,10,11,13)
  AND (tags.parent_id IN (1,2,3,4,5,7,8,9,10,11,13) OR tags.parent_id IS NULL)
\`\`\`

**Fixed (this PR):**
\`\`\`sql
-- the regression query is gone; permitted IDs resolved inline as a subquery:
WHERE tags.id IN (
        SELECT perm_tags.id FROM tags AS perm_tags
        WHERE (perm_tags.is_restricted = true AND perm_tags.id IN (5, 8, 11))
           OR perm_tags.is_restricted = false
      )
  AND (tags.parent_id IN (
        SELECT perm_tags.id FROM tags AS perm_tags WHERE ...
      ) OR tags.parent_id IS NULL)
\`\`\`

The two queries define the same permitted set:
- \`unrestricted tags\` ∪ \`restricted tags with explicit perm\` (PHP filter)
- \`(is_restricted = false)\` ∪ \`(is_restricted = true AND id IN (5,8,11))\` (subquery)

These are mathematically identical. The outer parent/child clause structure (\`tags.id IN ... AND (tags.parent_id IN ... OR parent_id IS NULL)\`) is byte-identical between the two implementations. The fix is structurally a faithful port of the WeakMap path back into SQL.

### 2. Explicit parent/child visibility tests

The existing visibility tests covered the parent/child case implicitly via multi-tag discussions 5 and 6 in the fixture (excluded because not all their tags are permitted). I added three focused tests that isolate the property at single-tag granularity:

- \`permission_on_child_does_not_grant_visibility_when_parent_is_off_limits\` — restricted-child + restricted-parent + perm on child only → hidden
- \`permission_on_child_grants_visibility_when_parent_is_unrestricted\` — restricted-child + unrestricted-parent + perm on child → visible
- \`root_restricted_tag_with_explicit_permission_is_visible\` — restricted-root + perm → visible (the \`orWhereNull('parent_id')\` branch)

All three pass on **both** the un-fixed and fix branches, confirming they describe properties that hold under the existing implementation and the new one.

PR #4672, opened 2026-05-23
[2.x] Add getPausedQueues() to QueueFactory for illuminate/queue v13.11+ compatibility

forumaker:
`illuminate/queue` added `getPausedQueues()` to queue drivers in November 2025 as part of the Queue Pause/Resume feature ([#57800](https://github.com/laravel/framework/pull/57800)). In v13.11, Worker.php began calling this method on all queue factories — including Flarum's custom QueueFactory, which doesn't implement it, causing the queue worker to crash with «Call to undefined method» and breaking all queued notifications and mail.

This adds a no-op implementation — returns empty array — consistent with the existing isPaused() pattern.

### Changes proposed in this pull request
Added getPausedQueues(): array method to QueueFactory with a no-op implementation — returns empty array.

### Reviewers should focus on
Whether returning an empty array is the correct behaviour, or if Flarum should implement
actual queue pausing support in the future.


imorland (approved):

imorland (changes requested):
This was targetting the `main` branch, not `2.x` so I switched it. You'll need to rebase please so this can now merge cleanly.

luceos:
These changes are not complete, the composer.json needs to specifically constraint with the right illuminate version or it will become a mess. The illuminate constraints in core and framework need to target at least `^13.11`.

forumaker:
@imorland @luceos Replaced by #4673, which is properly rebased on `2.x` and also updates the `illuminate/queue` constraint in `composer.json`

PR #4673, opened 2026-05-26
[2.x] Add getPausedQueues() to QueueFactory for illuminate/queue v13.11+ compatibility

### LINKED ISSUE #4671: queue:work crash-loops on Laravel 12: QueueFactory missing getPausedQueues() / getRestartTimestamp()
### Current Behavior

After `composer install` pulls in a Laravel 12 `illuminate/queue` (currently `^12.0`), the `flarum queue:work` worker crash-loops on every pop. Every queue-poll iteration throws:

```
Error: Call to undefined method Flarum\Queue\QueueFactory::getPausedQueues()
  in vendor/illuminate/queue/Worker.php:476
```

On our forum this manifested as 8,000+ identical error lines in the daily log and a queue worker that supervisor restarted every ~3 minutes for hours — no jobs processed, no notification emails sent during that window. Once the new `illuminate/queue` is in `composer.lock`, the worker cannot run.

### Root Cause

`Illuminate\Queue\Worker::getPausedQueues()` (added in Laravel 12) was extended in a way that delegates to the queue **manager**:

```php
// vendor/illuminate/queue/Worker.php (~L466)
protected function getPausedQueues($connectionName, $queues)
{
    if (! static::$pausable) {
        return [];
    }
    if ($this->cache === null) {
        return [];
    }
    return $this->manager->getPausedQueues($connectionName, $queues);  // ← fatal here
}
```

It is called unconditionally inside the worker's pop loop:

```php
// vendor/illuminate/queue/Worker.php (~L437)
$paused = array_flip($this->getPausedQueues($connection->getConnectionName(), $queues));
```

Flarum binds `\Illuminate\Contracts\Queue\Factory` to `Flarum\Queue\QueueFactory`, which implements only the single `connection()` method that the contract requires. The Worker's `$this->manager` is that `QueueFactory` instance — it has no `getPausedQueues()`, no `getRestartTimestamp()`. Same root cause would surface for the `$restartable` path the next time it gets hit.

### Steps to Reproduce

1. Flarum 2.0-rc.1 (or any composer state that resolves `illuminate/queue ^12.x` with the `getPausedQueues` Worker change).
2. Configure a non-trivial queue driver (we use Redis, but `database`/`sync` should reproduce the same fatal once polling fires).
3. `php flarum queue:work` (or any supervisor-managed worker) — the very first pop logs the fatal and exits, supervisor restarts, repeat.

### Expected

Worker runs without fatals against the existing `QueueFactory`. Pause/restart polling is genuinely optional in Laravel 12 — it exists for `php artisan queue:pause`, which Flarum doesn't expose.

### Suggested Fix

Use Laravel's own opt-out, `Illuminate\Queue\QueueManager::withoutInterruptionPolling()`, or equivalently flip the two static gates the Worker honours, in `Flarum\Queue\Console\WorkCommand::handle()`:

```php
public function handle()
{
    \Illuminate\Queue\Worker::$pausable = false;
    \Illuminate\Queue\Worker::$restartable = false;

    // … existing handle() body …
}
```

This is the upstream-blessed opt-out — see `QueueManager::withoutInterruptionPolling()` at `vendor/illuminate/queue/QueueManager.php:312`, which does exactly:

```php
public function withoutInterruptionPolling()
{
    Worker::$restartable = false;
    Worker::$pausable = false;
}
```

With these flags off, `Worker::getPausedQueues()` and the restart-cache check return early without ever touching `$this->manager`, so neither method needs to exist on `QueueFactory`. This also covers any future Laravel additions that gate behind `$pausable`/`$restartable`, which a per-method shim on `QueueFactory` wouldn't.

(An alternative is to shim `getPausedQueues($connection, $queues) { return []; }` on `QueueFactory`, but that's reactive — every future Laravel release that adds another manager-method behind the same flags would need a new shim.)

### Environment

- Flarum 2.0-rc.1
- PHP 8.4 on Debian 12
- `illuminate/queue` ^12 (just pulled by composer)
- Redis queue driver
- Supervisord-managed `flarum-queue` worker

### Additional Context

Discovered on production via 8,000+ `flarum.ERROR` entries in a single day's log. Workaround applied locally by setting the two statics in `Flarum\Queue\Console\WorkCommand::handle()`. Confirmed worker stable past the previous 3-minute crash interval; backlog drained.

forumaker:
Fixes #4671
Replaces #4672, which was based on the wrong branch.

illuminate/queue added getPausedQueues() to queue drivers in November 2025 as part of the Queue Pause/Resume feature — laravel/framework#57800.

In v13.11, Worker.php began calling this method on all queue factories — including Flarum's custom QueueFactory, which doesn't implement it, causing the queue worker to crash with «Call to undefined method» and breaking all queued notifications and mail silently.

- Added `getPausedQueues(): array` to `QueueFactory` with a no-op implementation, consistent with the existing `isPaused()` pattern.
- Updated `illuminate/queue` constraint to `^13.11` in `composer.json`.

Whether returning an empty array is the correct behaviour, or if Flarum should implement actual queue pausing support in the future.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -136,7 +136,7 @@
         "illuminate/filesystem": "^13.0",
         "illuminate/hashing": "^13.0",
         "illuminate/mail": "^13.0",
-        "illuminate/queue": "^13.0",
+        "illuminate/queue": "^13.11",
         "illuminate/session": "^13.0",
         "illuminate/support": "^13.0",
         "illuminate/validation": "^13.0",
--- a/framework/core/composer.json
+++ b/framework/core/composer.json
@@ -55,7 +55,7 @@
         "illuminate/filesystem": "^13.0",
         "illuminate/hashing": "^13.0",
         "illuminate/mail": "^13.0",
-        "illuminate/queue": "^13.0",
+        "illuminate/queue": "^13.11",
         "illuminate/session": "^13.0",
         "illuminate/support": "^13.0",
         "illuminate/validation": "^13.0",
--- a/framework/core/src/Queue/QueueFactory.php
+++ b/framework/core/src/Queue/QueueFactory.php
@@ -48,7 +48,7 @@ class QueueFactory implements Factory
      * Determine if a queue is paused.
      *
      * This is a no-op implementation since Flarum's simplified queue factory
-     * does not support queue pausing. Laravel 12's Worker expects this method
+     * doesn't support queue pausing. Laravel 12's Worker expects this method
      * to exist on the queue manager.
      *
      * @param string $connection
--- a/framework/core/src/Queue/QueueFactory.php
+++ b/framework/core/src/Queue/QueueFactory.php
@@ -48,7 +48,7 @@ class QueueFactory implements Factory
      * Determine if a queue is paused.
      *
      * This is a no-op implementation since Flarum's simplified queue factory
-     * doesn't support queue pausing. Laravel 12's Worker expects this method
+     * does not support queue pausing. Laravel 12's Worker expects this method
      * to exist on the queue manager.
      *
      * @param string $connection
@@ -58,5 +58,19 @@ class QueueFactory implements Factory
     public function isPaused($connection, $queue)
     {
         return false;
+    }
+
+    /**
+     * Get the list of paused queues.
+     *
+     * This is a no-op implementation since Flarum's simplified queue factory
+     * does not support queue pausing. illuminate/queue v13.11+ expects this
+     * method to exist on the queue manager.
+     *
+     * @return array
+     */
+    public function getPausedQueues(): array
+    {
+        return [];
     }
 }


huoxin233:
Should reference this issue #4671

imorland (changes requested):
Thank you for this. The `composer.json` at the root of the repo should also be updated accordingly please.

As for your question on pausing support, yes I do think this should be supported in the future - I've also been thinking about a refactor of the Flarum queue implementation recently. It might be something we can do for 2.1+ in a non-breaking way, but equally it might have to wait for 3.0, depending on all the changes required

forumaker:
@imorland  @huoxin233 Updated — also bumped `illuminate/queue` to `^13.11` in `framework/core/composer.json` and added a reference to #4671

luceos (approved):

imorland (approved):

PR #4674, opened 2026-05-28
[2.x] fix(tooltip): resolve native tooltip flash, async teardown leaks, and dynamic text

### LINKED ISSUE #4596: [1.x & 2.x] Toggle announcement button tooltip overridden by title
### Current Behavior

When using the "Show/hide announcement" button toggle on the admin announcements widget the opposite button state after a page load is not showing the correct tooltip layout. Instead it simply shows the browser "tooltip" on long hover which is coming from a wrongly set "title" attribute.

### Steps to Reproduce

1. In admin dashboard hover the mouse cursor over the eye button top right of the announcement widget => Observe the correct tooltip layout explaining the button's purpose
2. Click it once and hover again => Observe the missing tooltip. Instead after a longer hover wait you will see the browser tooltip shown. If you reload the page now the button will show the correct tooltip styling

### Expected Behavior

Hide/show announcements button should always show the same tooltip style on hover.

### Screenshots

https://github.com/user-attachments/assets/a9dc7081-56ec-4f56-8597-c65c0bf1f466

### Environment

- Flarum version: 2.0.0-rc.1
- Browser: Chrome 147

_No response_

### Possible Solution

When using the toggle button for the first time the updated button element gets a `title` attribute with the correct tooltip text.

### Additional Context

_No response_

huoxin233:
**Fixes #4596**

**`Tooltip` component (`common/components/Tooltip.tsx`):**
- Text-only changes now update the existing Bootstrap tooltip in-place (via `data-original-title` and `.tooltip-inner`) instead of destroying and recreating the entire instance, which caused an async race condition that left the element with no active tooltip.
- Tooltip is now initialized via the Bootstrap `title` config option with the DOM `title` attribute explicitly removed, preventing the
  browser's native tooltip from ever appearing as a fallback.
- After an in-place text update, `tooltip.show()` is called to force Bootstrap to recalculate positioning, preventing vertical
  misalignment when surrounding layout shifts (e.g. the announcement list expanding/collapsing).
- The `.fade` CSS class is stripped before every `destroy()` call to force synchronous cleanup, preventing stranded tooltip DOM nodes and race conditions during rapid updates.
- `checkDomNodeChanged()` now properly destroys the tooltip on the old DOM node before reassigning, fixing a pre-existing leak where swapped nodes left orphaned tooltip instances.

**`AnnouncementsWidget` (`admin/components/AnnouncementsWidget.tsx`):**
- Added `e.currentTarget.blur()` to the visibility toggle button's click handler so the tooltip dismisses on mouse-out after clicking.

For the `blur()` call in AnnouncementsWidget, I am not fully sure if this is the right approach. The issue is that after clicking the show/hide button, the browser keeps the button focused, which causes the tooltip to stay visible even after the cursor leaves (because tooltips trigger on both `hover` and `focus`). The `blur()` call fixes this, but other buttons in Flarum also has the same issue (e.g. the Tools button on top, shown in the video below).

So I am not sure if this is a intended design or this should be fix (if we are fixing that at a broader scope, then probably I should not include the fix for AnnouncementsWidget in this PR). I will leave this to the team to decide.

**Screenshot**

Before applying `blur()`:

https://github.com/user-attachments/assets/5a528a30-7f18-4710-ab7f-4af324e1c16e

After applying `blur()`:

https://github.com/user-attachments/assets/1d1454b7-5a58-43c1-9a7e-8e080da379b0

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/framework/core/js/src/admin/components/AnnouncementsWidget.tsx
+++ b/framework/core/js/src/admin/components/AnnouncementsWidget.tsx
@@ -86,7 +86,7 @@ export default class AnnouncementsWidget extends DashboardWidget {
               </Tooltip>
             )}
             <Tooltip text={app.translator.trans(this.hidden ? 'core.admin.announcements.show' : 'core.admin.announcements.hide')}>
-              <Button className="Button Button--icon" icon={this.hidden ? 'fas fa-eye' : 'fas fa-eye-slash'} onclick={() => this.toggleHidden()} />
+              <Button className="Button Button--icon" icon={this.hidden ? 'fas fa-eye' : 'fas fa-eye-slash'} onclick={(e: any) => { this.toggleHidden(); e.currentTarget.blur(); }} />
             </Tooltip>
           </div>
         </div>
--- a/framework/core/js/src/admin/components/AnnouncementsWidget.tsx
+++ b/framework/core/js/src/admin/components/AnnouncementsWidget.tsx
@@ -86,7 +86,14 @@ export default class AnnouncementsWidget extends DashboardWidget {
               </Tooltip>
             )}
             <Tooltip text={app.translator.trans(this.hidden ? 'core.admin.announcements.show' : 'core.admin.announcements.hide')}>
-              <Button className="Button Button--icon" icon={this.hidden ? 'fas fa-eye' : 'fas fa-eye-slash'} onclick={(e: any) => { this.toggleHidden(); e.currentTarget.blur(); }} />
+              <Button
+                className="Button Button--icon"
+                icon={this.hidden ? 'fas fa-eye' : 'fas fa-eye-slash'}
+                onclick={(e: any) => {
+                  this.toggleHidden();
+                  e.currentTarget.blur();
+                }}
+              />
             </Tooltip>
           </div>
         </div>
--- a/framework/core/js/src/admin/components/AnnouncementsWidget.tsx
+++ b/framework/core/js/src/admin/components/AnnouncementsWidget.tsx
@@ -86,14 +86,7 @@ export default class AnnouncementsWidget extends DashboardWidget {
               </Tooltip>
             )}
             <Tooltip text={app.translator.trans(this.hidden ? 'core.admin.announcements.show' : 'core.admin.announcements.hide')}>
-              <Button
-                className="Button Button--icon"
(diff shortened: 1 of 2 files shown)


dsevillamartin:
> For the blur() call in AnnouncementsWidget,

Wondering if it perhaps makes sense to add this to all Tooltip components, or perhaps Tooltip components with buttons inside. Would need to make sure that clicking on the tooltip itself doesn't call blur() but otherwise I think it makes sense to blur after clicking. Since this isn't just a problem with buttons that disappear but even normal ones with a tooltip (eg. clicking down on "show announcements" button but not letting go before moving away doesn't actually "click" but tooltip remains).

huoxin233:
> Wondering if it perhaps makes sense to add this to all Tooltip components,

I did a test by adding EventListener to listen for mousedown on the child element and schedule a blur on the subsequent mouseup, this will handle the the press and drag away case, but wouldnt this be a discrepancy between the buttons? As shown in the video below, button with tooltip will call blur in the press and drag away case, but normal buttons will still stay focus

https://github.com/user-attachments/assets/f4c866e9-3423-4359-9009-0bfed0dea80a

https://github.com/user-attachments/assets/c69ae795-66e8-4a23-9719-ebdd245f173e

dsevillamartin:
Looking further into it, I don't think applying `.blur()` is the correct strategy since that does more than we want. It could be replaced with a `$(this).tooltip('hide');` -- I believe that should do the trick. 

However, I just tested the latest 2.x (dfb5ef7f392803d65ca2ac3cc2711316cbed1bcf) and the behavior is different than what I remember:

https://github.com/user-attachments/assets/a601beb2-2184-4540-950e-d94ceca0833c

So perhaps we don't need to worry about blurring/hiding as much. I honestly think this might be fine - the other fixes included in this PR are still an improvement. And if we do want to keep it, I think calling `hide` makes more sense to avoid styling from the blur.

 This is what it looks like with this PR's changes (excluding the `.blur()` in AnnouncementsWidget).

https://github.com/user-attachments/assets/0ca8345a-bdb1-4373-a143-5914d2351ca6

huoxin233:
> Looking further into it, I don't think applying `.blur()` is the correct strategy since that does more than we want。

Sure, I will be reverting the `.blur()` in AnnouncementsWidget

imorland (approved):

PR #4677, opened 2026-05-31
[2.x] Implement content negotiation for forum error handling

### LINKED ISSUE #3850: LogInController doesn't handle validation exceptions properly
### Current Behavior

Logging in is the only auth route that doesn't use the API.
- Signing up is `/register`, which uses `/api/users` in the background
- Forgot password is `/api/forgot`

By default, the LogInValidator doesn't have any rules. This is on purpose, so that it can be extended. However, if you have failing rules there, the `ValidationExceptionHandler` won't return JSON since this is not an API route. Instead, it'll return a 422 with the exception stack trace (in debug mode):

```php
POST /login

Illuminate\Validation\ValidationException: The given data was invalid. in file ./vendor/flarum/core/src/Foundation/AbstractValidator.php on line 64
Stack trace:
  1. Illuminate\Validation\ValidationException->() ./vendor/flarum/core/src/Foundation/AbstractValidator.php:64
  2. Flarum\Foundation\AbstractValidator->assertValid() .vendor/flarum/core/src/Forum/Controller/LogInController.php:83
```

### Steps to Reproduce

1. Add a rule to `LogInValidator` (e.g. enable login recaptcha with `fof/recaptcha`)
2. Fail the validation rules (e.g. make request to `/login` without `g-recaptcha-response`)
3. View 422 with no JSON response

### Expected Behavior

The route should return JSON on a validation exception

### Screenshots

reCAPTCHA failure:

image
image

Authentication failure:

image

### Environment

- Flarum version: 1.8.1
- Webserver: nginx

_No response_

### Possible Solution

_No response_

### Additional Context

See https://github.com/FriendsOfFlarum/recaptcha/issues/39 and https://discuss.flarum.org/d/18399-friendsofflarum-recaptcha/173

dsevillamartin:
**Fixes #3850**

- Create a error handler for the forum that determines whether to forward to JSON or HTML view
  - Use JSON API Formatter for forum errors if the request did not explicitly state `text/html` in `Accepts` header
  - Fixes `/login` route not returning JSON from forum
  - Fixes `HighMaintenanceModeHandler` not returning JSON for requests initiated by forum (uses `isApiRequest`, and our requests do not set the header)

I added the `bitworking/mimeparse` package (last updated end of 2025) to parse more complex `Accepts` headers (eg. with quality priority). It looks like browsers basically always (wasn't able to find concrete evidence of this) have `text/html` in their Accepts. While our XHR requests have `Accepts: */*`.

However, this did make `isHtmlRequest` a weaker check that could be true at the same time as `isApiRequest` (due to the quality priorities). I think properly parsing with this package makes sense, however, perhaps `isHtmlRequest` should just be the negation of the stronger `isApiRequest` check (**made this change**)? Additionally, `isHtmlRequest` is only used in `LogInController` for the maintenance login error validaiton.

I also thought that adding this check to the forum-wide error handler made more sense than just the login route alone.
Plus, I didn't think just slapping an `Accepts` header to our JS requests was the proper solution.

**Screenshot**
image
image

**Necessity**

**Confirmed**

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -116,13 +116,14 @@
     "require": {
         "php": "^8.3",
         "ext-json": "*",
-        "fortawesome/font-awesome": "^7.0",
+        "bitworking/mimeparse": "^2.3",
         "composer/composer": "^2.7",
         "dflydev/fig-cookies": "^3.0",
         "doctrine/dbal": "^3.6.2",
         "dragonmantank/cron-expression": "^3.3",
         "fakerphp/faker": "^1.9.1",
         "flarum/json-api-server": "^0.1.0",
+        "fortawesome/font-awesome": "^7.0",
         "franzl/whoops-middleware": "2.0",
         "guzzlehttp/guzzle": "*",
         "illuminate/bus": "^13.0",
@@ -156,6 +157,7 @@
         "nelexa/zip": "^4.0.2",
         "nesbot/carbon": "^3.0",
         "nikic/fast-route": "^1.3",
+        "plesk/ratchetphp": "v1.0.4",
         "psr/http-message": "^1.1",
         "psr/http-server-handler": "^1.0.2",
         "psr/http-server-middleware": "^1.0.2",
@@ -172,8 +174,7 @@
         "symfony/postmark-mailer": "^7.0",
         "symfony/translation": "^7.0",
         "symfony/yaml": "^7.0",
-        "wikimedia/less.php": "^5.3",
-        "plesk/ratchetphp": "v1.0.4"
+        "wikimedia/less.php": "^5.3"
     },
     "require-dev": {
         "flarum/testing-tests": "*@dev",
--- a/framework/core/composer.json
+++ b/framework/core/composer.json
@@ -37,11 +37,13 @@
     },
     "require": {
         "php": "^8.3",
-        "fortawesome/font-awesome": "^7.0",
+        "bitworking/mimeparse": "^2.3",
         "dflydev/fig-cookies": "^3.0",
         "doctrine/dbal": "^3.6",
         "dragonmantank/cron-expression": "*",
         "fakerphp/faker": "^1.9.1",
+        "flarum/json-api-server": "^0.1.0",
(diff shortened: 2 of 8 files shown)


imorland:
Pushed some hardening and test coverage on top of this (`9e85a3bc9`) ahead of considering it for 2.0.0-rc.3. 👍

### Hardening `RequestUtil::getPreferredContentType()`

While adding tests I found that this helper — now reached on the **error-handling path** via `ContentNegotiationFormatter` — can blow up in two ways, turning an error response into a *second*, fatal error:

1. **`null` return → `TypeError`.** `Mimeparse::bestMatch()` is typed `?string` and returns `null` when the `Accept` header matches none of the offered types (e.g. `Accept: application/xml` or `text/plain`). The method is declared `: string`, so that's an uncaught `TypeError`. This is easy to trigger from the wild (bots, proxies, odd clients).
2. **Malformed `Accept` → warning + `UnexpectedValueException`.** A media range without a subtype makes `bestMatch()` emit an `E_WARNING` and throw.

Both now fall back to an empty string (treated as "not an API request" → HTML), so a hostile or unusual `Accept` header can't escalate. The `@` suppresses the library's warning and the `catch` swallows the throw — deliberately, since this is defensive code in the error handler.

### Tests added

- **`RequestUtilTest`** (unit) — `isApiRequest` / `isHtmlRequest` / `getPreferredContentType` across JSON:API, plain JSON, `*/*`, empty, typical-browser, unrelated, and malformed headers (explicitly pinning the never-throw and `null`→`''` behaviour, plus documenting that an empty `Accept` is treated as API).
- **`ContentNegotiationFormatterTest`** (unit) — API requests delegate to the JSON formatter, browser requests to the HTML formatter.
- **`ErrorNegotiationTest`** (integration) — a forum route that errors is negotiated through the full middleware stack: JSON for `*/*` (the #3850 regression) and `application/vnd.api+json`, HTML for browser `Accept`.

All green locally (13 unit + 3 integration) and PHPStan is clean on the changed files. This checks the "Tests have been added" box.

PR #4679, opened 2026-06-01
[2.x] fix: mobile discussion header total post count not updating on reply

### LINKED ISSUE #4602: [2.x] Mobile discussion total post count does not update on reply
### Current Behavior

When replying on mobile the total post count of the discussion in the header does not update.

### Steps to Reproduce

1. On mobile layout post a reply to a discussion
2. Observe the post count in the scrubber header => The current post # increases, but not the total count to the right

### Expected Behavior

When posting a reply the total post count should reflect accordingly.

### Screenshots

Image

### Environment

- Flarum version: 2.0.0-rc.1
- Browser: 147

_No response_

### Possible Solution

_No response_

### Additional Context

_No response_

huoxin233:
**Fixes #4602**

This PR fixes a UI regression in the `PostStreamScrubber` component where the total post count in the mobile discussion header fails to update when a user posts a reply. 

Previously, the `viewing` translation string was instantiated once as a static array of VNodes and injected into two distinct DOM locations (the dropdown header button and the scrubber handle). Because Mithril 2 does not support reusing VNode object references across multiple locations, the `vnode.dom` pointers for the header were silently overwritten by the scrubber handle. This caused subsequent `m.redraw()` cycles to update the total count and pluralization natively in the scrubber handle, but completely ignore the header button.

By wrapping the translation output in a `getViewing()` closure function and calling `{getViewing()}` in both JSX locations, we ensure fresh, independent VNodes are generated. This allows Mithril to accurately track and natively update both DOM elements during standard redraw cycles.

An alternative approach considered was: manually updating `.Scrubber-count` via jQuery inside `updateScrubberValues()` (similar to how the current scroll `index` is handled).
```
$scrubber.find('.Scrubber-count').text(formatNumber(count));
```
While this successfully forced the number to update in both locations, it was ultimately rejected because it bypassed Mithril's native rendering cycle. Bypassing Mithril for the total count is an anti-pattern and breaks the encapsulation of the UI component.

- `framework/core/js/src/forum/components/PostStreamScrubber.js`
- Confirming that the transition from a static variable to a closure function (`getViewing()`) resolves the stale DOM issue without introducing unnecessary overhead during scroll events (since the index is still updated via jQuery).

**Screenshot**

https://github.com/user-attachments/assets/cda3fc10-b437-4200-ba7e-210d21293ec9

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/framework/core/js/src/forum/components/PostStreamScrubber.js
+++ b/framework/core/js/src/forum/components/PostStreamScrubber.js
@@ -28,11 +28,12 @@ export default class PostStreamScrubber extends Component {
     const count = this.stream.count();
 
     // Index is left blank for performance reasons, it is filled in in updateScubberValues
-    const getViewing = () => app.translator.trans('core.forum.post_scrubber.viewing_text', {
-      count,
-      index: <span className="Scrubber-index"></span>,
-      formattedCount: <span className="Scrubber-count">{formatNumber(count)}</span>,
-    });
+    const getViewing = () =>
+      app.translator.trans('core.forum.post_scrubber.viewing_text', {
+        count,
+        index: <span className="Scrubber-index"></span>,
+        formattedCount: <span className="Scrubber-count">{formatNumber(count)}</span>,
+      });
 
     const unreadCount = this.stream.discussion.unreadCount();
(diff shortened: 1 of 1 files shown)


huoxin233:
Oops...
Closing this, duplicate of #4678

Issue #4137, opened 2024-12-16
[2.0.0-beta.1] Opening a discussion not always loads all posts


luceos:
### Current Behavior

When opening any discussion with icognito from the index, posts are missing within the view.

### Steps to Reproduce

1. Open incognito window
2. Open the community index
3. Open any discussion with at least 3 (or 4) posts
4. See one in loading state rendered

### Expected Behavior

The loading state post should be rendered.

### Screenshots

![image](https://github.com/user-attachments/assets/88283150-002b-42e4-8355-4df3d0bb2d1c)

### Environment

.

### Output of `php flarum info`

```
Flarum core: 2.0.0-beta.1
PHP version: 8.2.18
MySQL version: 10.11.9-MariaDB-ubu2204
Loaded extensions: Core, date, libxml, openssl, pcre, zlib, filter, hash, json, pcntl, random, Reflection, SPL, session, standard, sodium, mysqlnd, PDO, xml, calendar, ctype, curl, dom, mbstring, FFI, fileinfo, ftp, gd, gettext, iconv, intl, exif, mysqli, pdo_mysql, Phar, posix, readline, shmop, SimpleXML, sockets, sysvmsg, sysvsem, sysvshm, tokenizer, xmlreader, xmlwriter, xsl, zip, Zend OPcache
+---------------------+---------+--------+
| Flarum Extensions   |         |        |
+---------------------+---------+--------+
| ID                  | Version | Commit |
+---------------------+---------+--------+
| flarum-tags         | 0.0     |        |
| flarum-flags        | 0.0     |        |
| flarum-suspend      | 0.0     |        |
| flarum-sticky       | 0.0     |        |
| flarum-statistics   | 0.0     |        |
| flarum-mentions     | 0.0     |        |
| flarum-markdown     | 0.0     |        |
| flarum-lock         | 0.0     |        |
| flarum-likes        | 0.0     |        |
| flarum-lang-english | 0.0     |        |
| flarum-emoji        | 0.0     |        |
| flarum-bbcode       | 0.0     |        |
| flarum-approval     | 0.0     |        |
+---------------------+---------+--------+
Base URL: https://nightly.flarum.site
Installation path: /var/www/demos/nightly
Queue driver: sync
Session driver: file
Mail driver: array
Debug mode: off
```

### Possible Solution

It seems that not all posts from the relationship are loaded. Either the `near` logic is broken or something else, but the payload for the missing post is not in the api response.

PR #4703, opened 2026-06-10
[2.x] fix: avoid re-fetching the page of posts embedded in the initial page load

### LINKED ISSUE #4702: DiscussionPage re-fetches the page of posts already embedded in the initial-load payload (extra serialized API round-trip on every cold discussion view; LCP regression vs 1.x, introduced by #4156)
### Current Behavior

On a full page load of a discussion URL, `Content\Discussion` embeds the `page[near]` window of posts (including `contentHtml`) in the preloaded `apiDocument` — it needs them for the noscript/SEO content anyway — and `preloadedApiDocument()` pushes them all into the store.

Since #4156, the frontend never uses them: `DiscussionPage.show()` constructs `PostStreamState` without `includedPosts`, so the stream starts empty, `goToNumber()` finds nothing loaded, and `loadNearNumber()` fires `GET /api/posts?filter[discussion]=N&page[near]=M` — re-fetching byte-identical data that is already in the page.

That request is serialized *after* the JS boot and *before* the first post can paint, so on every cold discussion-page view the content render waits for one extra full API round trip (TTFB-bound, typically 0.5–1.5s for real visitors), and post images — frequently the LCP element — start downloading even later.

Flarum 1.x seeded the stream from the payload (`new PostStreamState(discussion, includedPosts)`). #4156 removed that seeding entirely because #4067 had made the *API* show endpoint stop returning the posts page, which broke the old extraction for in-app navigation (#4137). The removal fixed #4137 but also killed the optimization for the initial-page-load path, where the server still embeds a complete, correct window.

### Steps to Reproduce

1. Open any discussion URL in a logged-out browser with the network tab open (cold load, not in-app navigation).
2. Observe `GET /api/posts?filter[discussion]=…&page[near]=…` firing after boot, returning the same posts already present in the document payload (`apiDocument.included`).

### Measured impact

Lighthouse (mobile, simulated Slow 4G) on a production rc.3 forum, discussion page, before vs after restoring the seeding via a local patch:

| | before | after |
|---|---|---|
| posts XHR on landing | yes | no |
| simulated mobile LCP | 7.7 s | 6.2 s (median of 3) |

The same A/B in a Docker mirror of the production site (same machine, same DB, only the bundle changed) confirms the request is eliminated with no functional change. Field-data context: this extra round-trip was a major contributor to our `/d/*` URL group crossing Search Console's 4.0 s "poor mobile LCP" threshold after migrating 1.8 → 2.0 (1.x did not re-fetch).

Related but not in scope here: since #4067, in-app navigation to a discussion costs two API requests (discussion show + posts near) where 1.x cost one.

### Expected Behavior

The initial page load renders the post stream from the preloaded document without re-fetching the posts.

### Suggested Fix (tested)

Seed only on the preload path, and guard against the #4137 failure mode (non-contiguous includes) by seeding the longest run of store-loaded posts that is contiguous in `postIds()` order:

```ts
// DiscussionPage.load(), preloaded branch:
setTimeout(this.show.bind(this, preloadedDiscussion, this.preloadedNearPage(preloadedDiscussion)), 0);

// DiscussionPage.show():
show(discussion: ApiResponseSingle<Discussion>, preloadedPosts: Post[] = []): void {
  ...
  this.stream = new PostStreamState(discussion, preloadedPosts);
  ...
  const targetPost = nearParam === 'reply' ? 'reply' : nearParam && !isNaN(nearParam) ? nearParam : preloadedPosts[0]?.number() ?? 1;

/**
 * Extract the page of posts embedded in the server-preloaded discussion
 * document. Returns the longest run of posts contiguous in stream order, so
 * stray posts included via other relationships (e.g. extension includes —
 * production payloads do contain them) cannot corrupt the visible window
 * (the #4137 failure mode). API show responses (post-#4067) contain no posts
 * page, so for in-app navigation this returns [] and the stream fetches as
 * before.
 */
preloadedNearPage(discussion: ApiResponseSingle<Discussion>): Post[] {
  let best: Post[] = [];
  let current: Post[] = [];

  for (const id of discussion.postIds()) {
    const post = id ? app.store.getById<Post>('posts', id) : null;

    if (post) {
      current.push(post);
      if (current.length > best.length) best = current;
    } else {
      current = [];
    }
  }

  return best;
}
```

Validation performed (production mirror with real data, rc.3): landing on a 4-post and a 749-post discussion and on `/d/{id}/{slug}/400` deep links renders the correct window with no posts request; scrubber count correct (749); stray non-contiguous extension include correctly excluded; in-app navigation from the index still performs its fetch and renders fine (#4137 preserved); no console errors. Happy to submit this as a PR.

### Environment

- Flarum core: 2.0.0-rc.3
- PHP: 8.4.21
- MySQL: 8.4.9
- 41 extensions enabled (also reproduced with the suggested fix on a mirror of the same install)

ekumanov:
**Fixes #4702**

Since #4156, `DiscussionPage.show()` constructs `PostStreamState` without seeding it, so on every cold discussion-page load the stream re-fetches `GET /api/posts?filter[discussion]=N&page[near]=M` — data that `Content\Discussion` already embedded in the preloaded document (it needs it for the noscript content anyway) and that `preloadedApiDocument()` has already pushed into the store. The request is serialized after the JS boot and before the first post can paint, so it adds a full API round-trip (typically 0.5–1.5 s for real visitors) to the LCP critical path of every cold discussion view. 1.x rendered straight from the payload.

This PR restores the seeding, scoped so it cannot reintroduce #4137:

- `load()` passes `this.preloadedNearPage(preloadedDiscussion)` to `show()` **only on the preloaded-document path**. API show responses (post-#4067) embed no posts page, so for in-app navigation `preloadedNearPage()` returns `[]` and the stream fetches exactly as before.
- `preloadedNearPage()` returns the **longest run of store-loaded posts contiguous in `postIds()` order**. Stray posts pulled in by other relationships (e.g. extension includes — real production payloads do contain them) therefore can't corrupt `visibleStart`/`visibleEnd`, which was the #4137 failure mode that #4156 fixed by removing the old extraction.
- `show()` accepts the seed as an optional second parameter (default `[]`), so existing callers and subclasses are unaffected.
- When no `near` param is present, the pop-in target falls back to the first preloaded post's number (matching 1.x), then to 1.

- The contiguity rule in `preloadedNearPage()` as the guard against #4137.
- The `show()` signature extension (optional param) for compatibility with extensions overriding `show()`.

**Measured impact** (production rc.3 forum, 41 extensions; details in #4702):

- Docker mirror of the production site (same machine/DB, only the bundle changed): the posts request is eliminated on cold loads; no functional change.
- Production, same host, same day: Lighthouse mobile (simulated Slow 4G) LCP on a discussion page **7.7 s → 6.2 s** (median of 3); one fewer API request per cold discussion view.

**QA performed** (production mirror with real data):

- Cold landing on a 4-post and a 749-post discussion: no posts request, correct window, scrubber total correct.
- `/d/{id}/{slug}/400` deep link: window around post 400 rendered with no posts request.
- Non-contiguous extension include in the payload: correctly excluded by the contiguity rule.
- In-app navigation from the index: still fetches (discussion show + posts near) and renders, preserving #4137 behavior.
- Event posts (e.g. discussion renamed) inside the seeded window render fine; no console errors anywhere.

**Necessity**

- [x] Has the problem this PR solves been discussed prior to creating it? — #4702 (and the history in #4067 → #4137 → #4156).

**Confirmed**

- [x] Frontend changes: tested on a local Flarum installation (and a production mirror, see QA above).
- [ ] Backend changes: n/a — no backend changes.


imorland (approved):
Looks solid to me, approving.

Couple of minor things;

- Please don't commit the `dist` files in future, they can easily cause merge conflicts if the js changes on `2.x` between the time of the PR and merge. The Flarum Bot handles building at merge time.
- I would have liked to have seen a regression test/guard here. I'll follow up immediately after the merge with one this time, just something to note for the future :)

Issue #4622, opened 2026-05-04
Mention notifications and emails duplicate on rapid post-then-edit (race in queued NotificationSyncer insert path)


ekumanov:
### Current Behavior

When a `CommentPost` is created and then quickly re-saved (e.g. user fixes a typo within a second of posting, or anything else that fires `Posted` followed shortly by `Revised`), mentioned users can receive **multiple** notifications and **multiple** emails for the same `(blueprint, recipient)`.

In production data on a forum running 2.0 RC1, querying the `notifications` table for true duplicates of `userMentioned`:

```sql
SELECT user_id, from_user_id, subject_id, COUNT(*) AS dups,
       GROUP_CONCAT(id ORDER BY id),
       GROUP_CONCAT(created_at ORDER BY id),
       GROUP_CONCAT(is_deleted ORDER BY id)
FROM notifications
WHERE type = 'userMentioned'
GROUP BY user_id, from_user_id, subject_id, data
HAVING COUNT(*) > 1
ORDER BY MAX(created_at) DESC
LIMIT 5;
```

returns rows like:

```
user_id=2  from_user_id=51  subject_id=53055  dups=3  ids=105020,105025,105030  created_at=15:20:06,15:20:07,15:28:02  is_deleted=0,1,1
user_id=2  from_user_id=62  subject_id=54475  dups=2  ids=108367,108369         created_at=11:30:03,11:30:04            is_deleted=0,1
user_id=53 from_user_id=51  subject_id=53412  dups=2  ids=105817,105820         created_at=17:26:03,17:26:04            is_deleted=0,1
user_id=125 from_user_id=36 subject_id=53198  dups=2  ids=105345,105351         created_at=09:12:02,09:12:03            is_deleted=0,1
```

Note the `created_at` timestamps: each duplicate pair is ~1 second apart, all with `data IS NULL`. `Notification::scopeMatchingBlueprint` correctly returns all of them when queried after the fact — so the matching query is sound; the problem is **when** it's evaluated relative to the actual `INSERT`.

The user-visible symptom is duplicate emails: each `SendMentionsNotificationsJob` that doesn't see a prior row queues its own `SendEmailNotificationJob`, so the recipient receives N emails for one logical mention.

### Steps to Reproduce

1. With queue driver `redis` (or any non-`sync` driver), and a single queue worker, post a comment that mentions another user.
2. Within ~1 second (faster than the worker can drain the queue), open the post's edit dialog and save a small change.
[...]

PR #4715, opened 2026-06-12
[2.x] fix(realtime): push the notification record matching the blueprint, not the latest of its type

### LINKED ISSUE #4714: [2.x] flarum/realtime: notification push can broadcast a stale notification — record resolved by latest()-of-type instead of the event's blueprint
### Summary

When `flarum/realtime` broadcasts a `notification` event to a connected user, the job preparing the payload does not know which `notifications` record the event just created — it re-derives it with a "latest record of this type for this user" query. That heuristic resolves to the **wrong record** in realistic situations, and the client then renders a toast for an older, unrelated notification while the actual new one is never pushed. This matches recent user reports (e.g. on discuss.flarum.org) of the realtime notification banner "showing old notifications", sometimes suspected to be connected to backgrounding — the backgrounded device merely makes the stale record more noticeable when it resurfaces.

### Affected versions

- Present since the notification toast feature landed for rc.1 (#4473); the code is unchanged on current `2.x` HEAD.
- All supported databases. Most visible on installs running **multiple queue workers**, but reproducible on any setup (see the same-second case below).

### Root cause

`extensions/realtime/src/Push/Jobs/SendNotificationsJob.php`:

```php
$notification = Notification::query()
    ->where('user_id', $user->id)
    ->where('type', $this->blueprint::getType())
    ->latest()
    ->first();
```

The job receives the blueprint and recipients, but resolves the record with this query — no subject constraint, no id tiebreak, ordering only on `created_at`. Two independent ways this picks the wrong record:

**1. Same-second sibling.** `notifications.created_at` is a `DATETIME` (second granularity) and is not covered by any index, so `ORDER BY created_at DESC LIMIT 1` goes through a filesort whose order within equal timestamps is undefined. Two same-type notifications created within the same second (two replies in subscribed discussions, two mentions in quick succession — routine on a busy forum) sort arbitrarily; observed on MySQL 8.0 returning the **older** id. The job for the second event then re-pushes the *first* notification — the user sees a banner for something they already saw — and the second notification is never pushed at all.

**2. Concurrent queue workers.** The record itself is inserted by the alert driver's own queued job (`Flarum\Notification\Job\SendNotificationsJob`), and there is no ordering guarantee between that job and this one once more than one worker runs (or when the insert job is retried after a transient failure). When the realtime job reads before the insert commits, `latest()` resolves to the user's *previous* notification of that type — possibly hours old, e.g. one issued while their device was asleep. The stale record is pushed and toasted; the brand-new notification is silently dropped.

As a general point, `created_at` cannot serve as event identity here: `NotificationSyncer` restores existing records via un-delete without touching `created_at`, so "newest timestamp of this type" and "the record for this event" are simply different things.

### Demonstration (exact query shape, MySQL 8.0)

```sql
INSERT INTO notifications (user_id, from_user_id, type, subject_id, created_at, is_deleted, data)
VALUES (1, NULL, 'postMentioned', 101, '2026-06-12 10:00:00', 0, NULL),
       (1, NULL, 'postMentioned', 102, '2026-06-12 10:00:00', 0, NULL);

SELECT id, subject_id FROM notifications
WHERE user_id = 1 AND type = 'postMentioned'
ORDER BY created_at DESC LIMIT 1;
-- returns id 1 / subject 101 — the OLDER record.
-- EXPLAIN: key=notifications_user_id_index, Extra: Using where; Using filesort
```

So when the second mention's job runs, the payload generator serializes record 1 (`GET /api/notifications/1` as the recipient), the client's `userChannel.bind('notification', ...)` handler pushes it into the store, increments the unread badge, and toasts it — the first mention appears *again*, triggered by the second one.

### Expected behavior

The pushed payload corresponds to the notification record created (or restored) for the event being broadcast. The blueprint is already on the job, and core already has the right resolver: `Notification::matchingBlueprint($blueprint)` pins type + subject + sender + data. When the record is not visible yet (the insert job hasn't committed — the multi-worker case), a brief bounded retry is preferable to pushing a wrong record or dropping the event.

PR with the fix and regression tests follows.

ekumanov:
## Background

Closes #4714.

Users on 2.0 rc report the realtime notification banner "showing old notifications", sometimes suspected to be connected to device backgrounding. The actual mechanism is server-side and has been present since the toast feature landed for rc.1 (#4473): the push job resolves the record to broadcast with a *latest-of-this-type-for-this-user* lookup that is not tied to the event being handled.

## Root cause

`SendNotificationsJob` receives the blueprint and recipients, but re-derives the record with:

```php
$notification = Notification::query()
    ->where('user_id', $user->id)
    ->where('type', $this->blueprint::getType())
    ->latest()
    ->first();
```

No subject constraint, no id tiebreak, ordering only on `created_at` — which is second-granularity, not covered by an index (so equal timestamps sort arbitrarily; observed on MySQL 8.0 returning the *older* id), and not updated when `NotificationSyncer` restores a record. On top of that, the record is inserted by the alert driver's own queued job (`Flarum\Notification\Job\SendNotificationsJob`), with no ordering guarantee once multiple queue workers run: the realtime job can read before the insert commits and resolve the user's *previous* notification of that type — possibly hours old. Full analysis with a SQL demonstration in #4714.

Either way the client toasts whatever record was picked (and increments the badge), while the actual new notification is never pushed.

## Fix

Resolve the record from the blueprint, which pins type + subject + sender + data — `Notification::matchingBlueprint()` is exactly the resolver core's `NotificationSyncer` uses to identify a blueprint's records — with `latest('id')` as a deterministic tiebreak:

```php
$notification = Notification::matchingBlueprint($this->blueprint)
    ->where('user_id', $user->id)
    ->latest('id')
    ->first();
```

When no record is visible yet (the insert job hasn't committed — the concurrent-worker case), the job now re-dispatches itself for the affected recipients only, with a short delay and a bounded number of attempts (3 total), instead of pushing a wrong record or dropping the event:

- Recipients whose record *was* found are pushed immediately and are not part of the retry, so a retry can never double-push them.
- On the sync queue driver `later()` executes immediately; the attempt bound keeps that finite (and on sync the insert always precedes this job anyway, so the retry path is effectively never taken there).
- After the final attempt the job gives up silently — same outcome as the pre-existing `if ($notification)` guard when nothing matched.

## Not addressed here (pre-existing, separate concern)

When core syncs the same blueprint twice in rapid succession (the #4622 scenario), core dedups the *insert* at queue time, but two realtime jobs are still queued and will now both resolve the same (correct) record and push it twice. That duplicate-push of an identical record is a distinct, rarer issue and is unchanged by this PR — flagging it for completeness.

## Verification

- New integration tests in `extensions/realtime/tests/integration/push/SendNotificationsJobTest.php`:
  - `pushes_the_record_matching_the_blueprint_not_the_latest_of_its_type` — fixtures contain the blueprint's record with an *older* `created_at` plus an unrelated same-type record with a newer one; asserts the pushed `SendGeneratedPayloadJob` carries the blueprint's record.
  - `retries_when_the_record_is_not_yet_visible` — no matching record: asserts nothing is pushed and a re-dispatch with an incremented attempt is scheduled.
  - `gives_up_after_max_attempts` — at the attempt cap: asserts no push and no further re-dispatch.
- All three tests **fail on current `2.x` HEAD** without the src change (the first resolves the wrong record; the other two observe a push for an event whose own record doesn't even exist — the bug in miniature) and pass with it.
- Full realtime suite green: 14 tests, 33 assertions (1 pre-existing skip; remaining deprecation notices come from `middlewares/utils` under PHP 8.4, unrelated).
- `phpstan analyse extensions/realtime/src` (level 6): no errors.
- PHP-only change — no JS source or `dist` files touched.

### DIFF
--- a/extensions/realtime/src/Push/Jobs/SendNotificationsJob.php
+++ b/extensions/realtime/src/Push/Jobs/SendNotificationsJob.php
@@ -17,8 +17,21 @@ class SendNotificationsJob extends Job
 {
     public static ?string $onQueue = null;
 
-    public function __construct(private BlueprintInterface $blueprint, private array $recipients)
-    {
+    /**
+     * The notification record is inserted by the alert driver's own queued job
+     * (Flarum\Notification\Job\SendNotificationsJob), which this job has no
+     * ordering guarantee with on concurrent queue workers. When the record
+     * isn't visible yet, re-check a bounded number of times instead of
+     * dropping the push.
+     */
+    protected const MAX_ATTEMPTS = 3;
+    protected const RETRY_DELAY_SECONDS = 2;
+
+    public function __construct(
+        private BlueprintInterface $blueprint,
+        private array $recipients,
+        private int $attempt = 1
+    ) {
         parent::__construct();
     }
 
@@ -27,22 +40,38 @@ class SendNotificationsJob extends Job
         // Only dispatch notification jobs for users on the socket.
         $intersect = $this->connectedUsers()->intersect($this->recipients);
 
+        $missing = [];
+
         foreach ($intersect as $user) {
             if (! $user->shouldAlert($this->blueprint::getType())) {
                 continue;
             }
 
-            $notification = Notification::query()
+            // Resolve the exact record this blueprint produced for the user.
+            // A latest()-of-type lookup is not equivalent: `created_at` has
+            // second granularity and restored records keep their original
+            // timestamp, so with several notifications of the same type it
+            // can resolve to a different, older record — pushing a stale
+            // notification to the client.
+            $notification = Notification::matchingBlueprint($this->blueprint)
                 ->where('user_id', $user->id)
-                ->where('type', $this->blueprint::getType())
-                ->latest()
+                ->latest('id')
                 ->first();
 
             if ($notification) {
                 $queue->push(
                     new SendGeneratedPayloadJob('notification', $notification, $user)
                 );
+            } elseif ($this->attempt < self::MAX_ATTEMPTS) {
+                $missing[] = $user;
             }
+        }
+
+        if (count($missing)) {
+            $queue->later(
+                self::RETRY_DELAY_SECONDS,
+                new self($this->blueprint, $missing, $this->attempt + 1)
+            );
         }
     }
 }
--- a/extensions/realtime/tests/integration/push/SendNotificationsJobTest.php
+++ b/extensions/realtime/tests/integration/push/SendNotificationsJobTest.php
@@ -1,0 +1,181 @@
+<?php
+
+/*
+ * This file is part of Flarum.
+ *
+ * For detailed copyright and license information, please view the
+ * LICENSE file that was distributed with this source code.
+ */
+
+namespace Flarum\Realtime\Tests\integration\push;
+
+use Carbon\Carbon;
+use Flarum\Database\AbstractModel;
+use Flarum\Discussion\Discussion;
+use Flarum\Notification\AlertableInterface;
+use Flarum\Notification\Blueprint\BlueprintInterface;
+use Flarum\Notification\Notification;
+use Flarum\Realtime\Push\Jobs\SendGeneratedPayloadJob;
+use Flarum\Realtime\Push\Jobs\SendNotificationsJob;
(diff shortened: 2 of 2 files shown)


imorland:
Closing in favour of #4713

PR #4720, opened 2026-06-12
[2.x] fix(core): return HTTP 503 when upgrade page is shown

DavideIadeluca:
**Fixes #0000**
Previously a HTTP 200 response was returned for the upgrade page, which is semantically incorrect and might lead to false negatives in monitoring systems or in testing contexts. See:

Screenshot 2026-06-12 at 22 21 35

Return 503 which is **less** semantically incorrect than 200. not perfect but better

Looking at the history it appears like it was actually a 503  10 years ago (see https://github.com/flarum/framework/commit/23eb4c805b8c49ea9a0df3724369356b38867ac9) but at one point was acidentally removed whilst refactoring, falling back to the default 200 response

**Screenshot**

**Necessity**

**Confirmed**

**Required changes:**

### DIFF
--- a/framework/core/src/Update/Controller/IndexController.php
+++ b/framework/core/src/Update/Controller/IndexController.php
@@ -13,6 +13,8 @@ use Flarum\Foundation\Config;
 use Flarum\Http\Controller\AbstractHtmlController;
 use Illuminate\Contracts\Support\Renderable;
 use Illuminate\Contracts\View\Factory;
+use Laminas\Diactoros\Response\HtmlResponse;
+use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface as Request;
 
 class IndexController extends AbstractHtmlController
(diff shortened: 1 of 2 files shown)


imorland (changes requested):
Thank you @DavideIadeluca this is a very welcome change. I feel that `503` is _exactly_ what this falls under:

> "the server is currently unable to handle the request." MDN/RFC commentary: "server is down for maintenance or overloaded… should be used for temporary conditions."

As this has regressed before as you rightly say, I'd like to see a basic test implemented to lock in the behaviour here, please. A full end to end might be tricky, but a simple test to hit the `Update/Controller/IndexController` IMO would be sufficient.

DavideIadeluca:
> As this has regressed before as you rightly say, I'd like to see a basic test implemented to lock in the behaviour here

yep good call. added in https://github.com/flarum/framework/pull/4720/commits/83806d6f333c54b3d4d70549a2f07b747f84505f / https://github.com/flarum/framework/pull/4720/commits/3e6aebcbd2358abb070371a1b432855b383e0f9d

imorland (approved):

PR #4749, opened 2026-06-17
[2.x] fix(a11y): add accessible label to new discussion button

claudiushenrichs:
The "New Discussion" button in the main forum sidebar has a visible button text on desktop. On mobile only the icon is shown without text so the mobile button is failing the "Buttons do not have an accessible name" accessibility test. This pull request thus adds the aria-label re-using the button text so the icon only mobile button variant can be read out e.g. by screen readers.

Is it okay to re-use the button label or should a different text key be used for the accessible label

**Screenshot**
image

**Necessity**

**Confirmed**

**Required changes:**
None


imorland:
> Is it okay to re-use the button label or should a different text key be used for the accessible label

Yes, this is _generally_ the approach we take :)

imorland (approved):

PR #4777, opened 2026-06-20
[2.x] fix(tags): update tag metadata when a pending discussion is approved

### LINKED ISSUE #4776: [2.x] Tags: `discussion_count` / `last_posted_discussion_id` not updated when a pending discussion is approved
### Reproduced on

A clean Flarum **2.0.0-rc.4** install with only `flarum/tags` and `flarum/approval` enabled (plus `flarum/flags`, an approval dependency), MariaDB 11. The control case below confirms the normal path is unaffected.

### Description

When `flarum/approval` is active, content created pending approval is treated as `is_private` until approved. `flarum/tags`' `UpdateTagMetadata` correctly skips private content, so a pending discussion does not update tag metadata. But when the discussion is later **approved**, nothing recomputes the tag — so the tag's `discussion_count` and `last_posted_discussion_id` are left permanently unset/stale even though the discussion is now public.

### Steps to reproduce

1. Fresh Flarum 2.0.0-rc.4; enable Tags + Approval.
2. Create a (non-restricted) tag. Grant the Members group `startDiscussion` but **not** `discussion.startWithoutApproval`, so members' new discussions require approval.
3. As a member, start a discussion with that tag. It is created pending: `is_approved = false`, and approval marks it `is_private = true`.
4. As an admin/mod, approve it: `PATCH /api/posts/{firstPostId}` with `{ "data": { "attributes": { "isApproved": true } } }`.

### Expected

After approval, the tag's `discussion_count` and `last_posted_discussion_id` reflect the now-visible discussion.

### Actual

The discussion becomes `is_approved = true`, `is_private = false`, fully visible — but the tag's `discussion_count` is still `0` and `last_posted_discussion_id` is still `null`. The `/tags` page shows a wrong count and no "latest discussion" for that tag. Replies gated by `replyWithoutApproval` are affected the same way.

Minimal reproduction output (core + tags + approval only):

```
A) member posts   -> is_approved=false, is_private=true  -> tag count=0, last_posted_discussion_id=NULL
   admin approves -> is_approved=true,  is_private=false -> tag count=0, last_posted_discussion_id=NULL   <- bug
B) control: admin auto-approved post -> tag count=1, last_posted_discussion_id set   (normal path works)
```

### Root cause

- `Flarum\Approval\Listener\UnapproveNewContent::markUnapprovedContentAsPrivate()` — unapproved content is treated as `is_private`.
- `Flarum\Tags\Listener\UpdateTagMetadata::updateTags()` gates both the `discussion_count` change and `setLastPostedDiscussion()` behind `! $discussion->is_private`, so pending content is (correctly) skipped on `Started` / `Posted`.
- On approval, `Flarum\Approval\Listener\UpdateDiscussionAfterPostApproval` (handling `PostWasApproved`) refreshes the discussion's and author's counts, but never touches tag metadata — and `UpdateTagMetadata` does not subscribe to `PostWasApproved` / `PostWasUnapproved`.

So no code path updates tag metadata when content transitions from pending to approved.

### Suggested fix

Have `UpdateTagMetadata` subscribe to `PostWasApproved` (and `PostWasUnapproved`) and recompute the affected tags' `discussion_count` + `last_posted_discussion_id` (e.g. `refreshLastPostedDiscussion()` plus a recount), with the matching decrement on unapprove/reject.

### Related

- #3930 / #3925 corrected *when* `PostWasApproved` fires, but not tag metadata.
- #3252 fixed tag `discussion_count` on first/only-post deletion — a similar gap in the same listener.

### Versions

- Flarum core / tags / approval: 2.0.0-rc.4
- PHP 8.x, MariaDB 11

karl-bullock:
Fixes #4776

## Problem

With `flarum/approval` active, a discussion/post created pending approval is `is_private`, so `flarum/tags` correctly skips it while maintaining `discussion_count` / `last_posted_discussion_id`. But nothing refreshes those counters once the content is **approved**, so a tag is left with a permanently wrong count and stale last-posted discussion (e.g. the `/tags` tile shows the wrong number and no latest discussion).

## Fix

Reworked per review: this now lives entirely within `flarum/approval` as an optional, conditional integration with `flarum/tags`, mirroring how `flarum/mentions` refreshes its own metadata on approval.

- A new `Listener\UpdateTagMetadataAfterApproval` listens for the **typed** `PostWasApproved` event (no string-based listeners).
- It is registered only when tags is present, via `Extend\Conditional()->whenExtensionEnabled('flarum-tags', ...)`, so **`flarum/tags` is left completely untouched** and stays unaware of approval.
- On approval it refreshes the affected tags through their public model API: approving a first post makes the discussion countable (`+1`), an approved reply just refreshes the last-posted metadata. Since listener order relative to `UpdateDiscussionAfterPostApproval` (which flips `is_private`) is not guaranteed, it defers via `afterSave` while the discussion is still private.
- `flarum/approval` now dev-requires `flarum/tags` (as `flarum/mentions` does) for the integration test.

## Testing

- Added `TagMetadataTest` (integration): a pending discussion in a tag starts at `discussion_count = 0` with no last-posted; after its first post is approved the tag shows `discussion_count = 1` and points at the discussion. Passes locally.
- `composer analyse:phpstan` passes.

### DIFF
--- a/extensions/approval/composer.json
+++ b/extensions/approval/composer.json
@@ -91,6 +91,7 @@
     },
     "require-dev": {
         "flarum/audit": "*",
+        "flarum/tags": "*@dev",
         "flarum/testing": "^2.0"
     }
 }
--- a/extensions/approval/composer.json
+++ b/extensions/approval/composer.json
@@ -91,7 +91,7 @@
     },
     "require-dev": {
         "flarum/audit": "*",
-        "flarum/tags": "*@dev",
+        "flarum/tags": "*",
         "flarum/testing": "^2.0"
     }
 }
--- a/extensions/approval/extend.php
+++ b/extensions/approval/extend.php
@@ -67,6 +67,15 @@ return [
     (new Extend\ModelPrivate(CommentPost::class))
         ->checker(Listener\UnapproveNewContent::markUnapprovedContentAsPrivate(...)),
 
+    // Refresh tag metadata that flarum/tags skipped while the content was
+    // pending approval (and therefore private). Kept optional/conditional here
+    // so flarum/tags stays unaware of approval.
+    (new Extend\Conditional())
+        ->whenExtensionEnabled('flarum-tags', fn () => [
+            (new Extend\Event())
+                ->listen(PostWasApproved::class, Listener\UpdateTagMetadataAfterApproval::class),
+        ]),
+
     (new Extend\Conditional())
         ->whenExtensionEnabled('flarum-audit', fn () => [
             (new \Flarum\Audit\Extend\Audit())
--- a/extensions/approval/src/Listener/UpdateTagMetadataAfterApproval.php
+++ b/extensions/approval/src/Listener/UpdateTagMetadataAfterApproval.php
@@ -1,0 +1,50 @@
+<?php
+
+/*
+ * This file is part of Flarum.
+ *
+ * For detailed copyright and license information, please view the
+ * LICENSE file that was distributed with this source code.
+ */
+
+namespace Flarum\Approval\Listener;
+
+use Flarum\Approval\Event\PostWasApproved;
+use Flarum\Discussion\Discussion;
+
+class UpdateTagMetadataAfterApproval
+{
+    public function handle(PostWasApproved $event): void
+    {
+        $discussion = $event->post->discussion;
+
+        // flarum/tags counts a discussion only once its first post is approved;
+        // approving a later reply just needs the last-posted metadata refreshed.
+        $countable = $event->post->number === 1;
+
+        // flarum/tags skips private (pending-approval) content, so its counters
+        // are left stale after approval. The visibility change is applied
+        // separately by UpdateDiscussionAfterPostApproval and listener order is
+        // not guaranteed, so defer to the discussion's save while still private.
+        if ($discussion->is_private) {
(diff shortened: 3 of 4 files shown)


imorland (changes requested):
This does not belong in `flarum/tags`. This needs to be an optional/conditional within `flarum/approval`.

We don't use string based representations of events for listeners either, it defeats the purpose of having everything strongly typed and static analysis.

karl-bullock:
Thanks, that makes sense. I've reworked it so it lives entirely in `flarum/approval` now: a typed `PostWasApproved` listener registered under `whenExtensionEnabled('flarum-tags', ...)`, refreshing the tags through their public model API. `flarum/tags` is untouched and there are no string-based listeners. Added an integration test for the approve-then-count flow. Happy to reshape if you'd prefer it differently.

extensions/approval/composer.json
```diff
@@ -91,6 +91,7 @@
     },
     "require-dev": {
         "flarum/audit": "*",
+        "flarum/tags": "*@dev",
```
imorland:
```suggestion
        "flarum/tags": "*",
```

imorland (changes requested):

karl-bullock:
Fixed.

imorland (approved):

PR #4816, opened 2026-07-14
[2.x] Fix notifications endpoint 400 on invalid `subject.discussion` default include

### LINKED ISSUE #4815: [2.x] Notifications endpoint 400s on an invalid `subject.discussion` default include when a notification subject type has no discussion relationship
## Description

`NotificationResource` adds `subject.discussion` to the **default** include for the notifications endpoints whenever more than one notification subject type is registered:

```php
// framework/core/src/Api/Resource/NotificationResource.php
->defaultInclude(array_filter([
    'fromUser',
    'subject',
    $this->initialized && count($this->subjectTypes()) > 1
        ? 'subject.discussion'
        : null,
]))
```

The include validator (`flarum/json-api-server`, `IncludesData::validateInclude`) validates default includes too, and a nested include on the polymorphic `subject` relationship is only valid if **at least one** subject type exposes an includable `discussion` relationship. When the registered subject types are, for example, `{discussions, users}` (neither of which has a `discussion` relationship), `subject.discussion` is invalid and the request is rejected:

```json
{"errors":[{"status":"400","code":"validation_error","title":"Bad Request","detail":"Invalid include [subject.discussion]","source":{"parameter":"include"}}]}
```

In other words, the framework generates a default include that it then rejects with a 400. Because `NotificationListState` sends no explicit `include` and relies on the server default, this breaks the notifications dropdown and page for every user on the affected forum, not just direct API consumers.

## Steps to reproduce

On a forum where **no** registered notification subject type is discussion-bearing (i.e. no `Post`-subject notification type is present: mentions, subscriptions, likes, flags, etc. all disabled), register any notification type whose subject is a `User` (or any model without a `discussion` relationship). A minimal test extender is enough:

```php
use Flarum\Database\AbstractModel;
use Flarum\Extend;
use Flarum\Notification\AlertableInterface;
use Flarum\Notification\Blueprint\BlueprintInterface;
use Flarum\User\User;

class UserSubjectBlueprint implements BlueprintInterface, AlertableInterface
{
    public function getFromUser(): ?User { return null; }
    public function getSubject(): ?AbstractModel { return null; }
    public function getData(): mixed { return []; }
    public static function getType(): string { return 'userSubjectNotification'; }
    public static function getSubjectModel(): string { return User::class; }
}

// extend.php
(new Extend\Notification())->type(UserSubjectBlueprint::class, ['alert']),
```

Then, authenticated as any user:

```
GET /api/notifications
```

No data is required. The subject collection now resolves to `{discussions, users}`, so `count(subjectTypes()) > 1` is true and `subject.discussion` is added to the default include.

Confirmed behavior on a fresh `2.0.0-rc.5` install (MariaDB), varying only the registered subject types:

| Registered notification subject types | `GET /api/notifications` |
| --- | --- |
| `{discussions}` (core only) | `200` |
| `{discussions, users}` (add a User-subject type) | `400 Invalid include [subject.discussion]` |
| `{discussions, users, posts}` (also add a Post-subject type) | `200` |

## Expected behavior

`GET /api/notifications` returns `200` and the notification list loads. A default include the framework generates for itself should never be invalid.

## Actual behavior

`GET /api/notifications` returns `400` with `Invalid include [subject.discussion]`, so the notifications dropdown/page fails to load for all users on that forum.

## Why this is not exotic

`User` subjects are already a first-class pattern in the bundled extensions: `flarum/mentions` ships `UserMentionedBlueprint` (subject `User`) alongside `GroupMentionedBlueprint` (subject `Group`). Mentions only avoids the 400 because it *also* ships `PostMentionedBlueprint` (subject `Post`), and `Post` has a `discussion` relationship, which keeps `subject.discussion` valid for the whole collection. Any forum that registers a `User`-subject (or `Group`-subject) notification type without a `Post`-subject type present hits the invalid include. A default install is safe because mentions and subscriptions are enabled, so this is low-frequency, but the underlying include is still invalid by construction.

## Suggested direction

Only add `subject.discussion` to the default include when at least one subject type actually exposes an includable `discussion` relationship, rather than gating on `count($this->subjectTypes()) > 1`. Alternatively, make nested includes on a polymorphic relationship tolerant of member types that lack the nested field (apply where valid, skip where not) so a framework-generated default can never 400.

## Environment

- Flarum: reproduced on a fresh `2.0.0-rc.5` install (MariaDB); the `NotificationResource` default-include code is unchanged from `rc.4` and is identical on the current `2.x` branch.
- `flarum/json-api-server`: `0.1.2`.
- Reproduced with a minimal single-type test extender, no third-party extensions involved.

karl-bullock:
Fixes #4815

## Changes proposed in this pull request

`NotificationResource` added `subject.discussion` to the notifications endpoints' default include whenever more than one notification subject type was registered:

```php
$this->initialized && count($this->subjectTypes()) > 1
    ? 'subject.discussion'
    : null,
```

That nested include is only valid when at least one subject type actually exposes an includable `discussion` relationship. The include validator validates default includes too, so when the subject collection is, for example, `{discussions, users}` (neither of which has a `discussion` relationship), it rejects the endpoint's own default include:

```json
{"errors":[{"status":"400","code":"validation_error","title":"Bad Request","detail":"Invalid include [subject.discussion]","source":{"parameter":"include"}}]}
```

Because `NotificationListState` sends no explicit `include` and relies on the server default, this returns `400` for every user and breaks the notifications dropdown and page. `User` subjects are not exotic: `flarum/mentions` ships `UserMentionedBlueprint`; it only stays safe because it also ships `PostMentionedBlueprint`, and `Post` has a `discussion` relationship. Any forum with a `User`-subject (or `Group`-subject) notification type and no `Post`-subject type present hits this.

This PR gates the include on whether a subject type genuinely exposes an includable `discussion` relationship, instead of on the subject-type count. The include is preserved exactly where it matters (Post subjects) and dropped only when it would be invalid.

## Reviewers should focus on

- `subjectsHaveDiscussionRelationship()` resolves each subject type's resource and checks its fields for an includable `discussion` relationship. This uses `resolveFields()` (which applies field extenders), guarded by an `instanceof AbstractResource` check.

## Verified

Reproduced and fixed on a fresh `2.0.0-rc.5` install (MariaDB):

| Registered subject types | `GET /api/notifications` before | after |
| --- | --- | --- |
| `{discussions}` (core only) | 200 | 200 |
| `{discussions, users}` | **400** | 200 |
| `{discussions, users, posts}` | 200 | 200 |

Two regression tests are added to `notifications/ListTest`: one asserting the endpoint no longer 400s when a subject type has no discussion relationship, and one asserting the discussion is still eager-loaded when a discussion-bearing subject type is present. The existing notification tests still pass.

### DIFF
--- a/framework/core/src/Api/Resource/NotificationResource.php
+++ b/framework/core/src/Api/Resource/NotificationResource.php
@@ -19,6 +19,7 @@ use Flarum\Notification\NotificationRepository;
 use Illuminate\Contracts\Cache\Repository as CacheRepository;
 use Illuminate\Database\Eloquent\Builder;
 use Tobyz\JsonApiServer\Pagination\OffsetPagination;
+use Tobyz\JsonApiServer\Schema\Field\Relationship;
 
 /**
  * @extends AbstractDatabaseResource<Notification>
@@ -66,16 +67,25 @@ class NotificationResource extends AbstractDatabaseResource
 
     public function endpoints(): array
     {
+        // Eager-load the subject's discussion only when at least one subject
+        // type actually exposes a `discussion` relationship. Gating on the
+        // number of subject types instead would add an invalid `subject.discussion`
+        // default include when the extra types have no such relationship (e.g. a
+        // User-subject notification like flarum/mentions' `userMentioned` on a
+        // forum with no Post-subject notification type), which the include
+        // validator then rejects with a 400 on the endpoint's own default.
+        $defaultInclude = array_filter([
+            'fromUser',
+            'subject',
+            $this->initialized && $this->subjectsHaveDiscussionRelationship()
+                ? 'subject.discussion'
+                : null,
+        ]);
+
         return [
             Endpoint\Show::make()
                 ->authenticated()
-                ->defaultInclude(array_filter([
-                    'fromUser',
-                    'subject',
-                    $this->initialized && count($this->subjectTypes()) > 1
-                        ? 'subject.discussion'
-                        : null,
-                ])),
+                ->defaultInclude($defaultInclude),
             Endpoint\Update::make()
                 ->authenticated(),
             Endpoint\Index::make()
@@ -86,13 +96,7 @@ class NotificationResource extends AbstractDatabaseResource
                     // Invalidate new notification count cache since read_notifications_at changed
                     $this->cache->forget("user.{$actor->id}.new_notification_count");
                 })
-                ->defaultInclude(array_filter([
-                    'fromUser',
-                    'subject',
-                    $this->initialized && count($this->subjectTypes()) > 1
-                        ? 'subject.discussion'
-                        : null,
-                ]))
+                ->defaultInclude($defaultInclude)
                 ->paginate(),
         ];
     }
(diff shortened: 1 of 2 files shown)


imorland:
Superseded by #4828, which combines this with #4817 and adds tests for both. Thanks!

PR #4821, opened 2026-07-22
[2.x] fix: blank index title control when no nav item is active

### LINKED ISSUE #4819: [2.x] Mobile toolbar title is blank on tag pages when the current tag has no sidebar nav item
### Current Behavior

On phone-width viewports, the index toolbar's title dropdown (`App-titleControl`) renders blank — only the `fas fa-sort` caret shows — when viewing a tag that has no item in the index sidebar nav.

This affects any **secondary tag that isn't among the site's 3 most-discussed secondary tags**, because `addTagList` only adds the top 3 secondary tags to the nav, and the nav `SelectDropdown` has no `defaultLabel` to fall back on when none of its children is active.

Live comparison (2.0.0-rc.5) — both are secondary tags with 1 discussion each; "Sandbox" happens to rank in the top 3 secondary tags, "Uptime Monitoring" ranks 4th:

- https://linkrobins.com/forum/t/uptime-monitoring — toolbar title is blank
- https://linkrobins.com/forum/t/sandbox — toolbar shows "Sandbox"

Because the ranking depends on live discussion counts, the set of affected tags silently shifts as counts change, so the bug looks intermittent/random from an admin's point of view.

### Steps to Reproduce

1. Create 4+ secondary tags (tags without a position) with at least one discussion each.
2. On a phone-width viewport, open `/t/{slug}` for a secondary tag that is **not** among the top 3 secondary tags by discussion count.
3. The toolbar title is empty; only the sort caret renders. (The tag hero below it renders the name fine — the tag is in the store.)
4. Compare with a primary tag or a top-3 secondary tag: the name renders in the toolbar.

### Expected Behavior

The toolbar title shows the current tag's name for every tag page, regardless of whether the tag earned a spot in the sidebar nav list.

### Screenshot

Image

### Environment

- Flarum version: 2.0.0-rc.5
- Website URL: https://linkrobins.com/forum
- Webserver: nginx
- Hosting environment: vps (docker)
- PHP version: 8.3.32
- Browser: Chrome (screenshots reproduced with headless Chromium at 390×844)

### Possible Solution

The chain, from the 2.x source:

1. `SelectDropdown.getButtonContent()` uses the first active child's label, else `this.attrs.defaultLabel` ([SelectDropdown.tsx](https://github.com/flarum/framework/blob/2.x/framework/core/js/src/common/components/SelectDropdown.tsx)).
2. The nav `SelectDropdown` built in `IndexSidebar.items()` passes no `defaultLabel` ([IndexSidebar.tsx](https://github.com/flarum/framework/blob/2.x/framework/core/js/src/forum/components/IndexSidebar.tsx)).
3. `addTagList` adds all primary tags, but secondary tags are sorted by `discussionCount` and only `more.splice(0, 3)` are added ([addTagList.js](https://github.com/flarum/framework/blob/2.x/extensions/tags/js/src/forum/addTagList.js)). A tag outside that set has no nav item, so nothing is active and the label collapses to the missing `defaultLabel`.

Either of these would fix it:

- In `addTagList`, always `addTag(app.currentTag())` when the position/top-3 filters excluded it, so the current tag always has an (active) nav item; or
- Pass a `defaultLabel` to the nav `SelectDropdown` in `IndexSidebar` — e.g. `app.currentTag()?.name()`, falling back to the all-discussions label.

### Additional Context

#3773 fixed an empty-string `SelectDropdown` title previously, but not this code path — here the dropdown has children, just no active one.

karl-bullock:
Fixes #4819

### What / why

On phone-width viewports the index toolbar title (the nav `SelectDropdown`, `App-titleControl`) renders blank on tag pages whenever the current tag has no item in the index sidebar nav. In practice that means any secondary tag that is not among the 3 most discussed secondary tags, so the set of affected tags shifts as discussion counts change and the bug looks random from an admin's point of view.

### Root cause

`SelectDropdown.getButtonContent()` uses the label of the first active child and falls back to `attrs.defaultLabel`. `defaultLabel` is declared required on `ISelectDropdownAttrs`, but the nav dropdown built in `IndexSidebar.items()` never passed one. When no child is active the label collapses to nothing and only the sort caret is left.

Tags is what exposes it: `addTagList` adds primary tags plus the top 3 secondary tags by `discussionCount()`, so a tag outside those filters contributes no nav item and nothing is active.

### Fix

Two parts, following the approach agreed in review:

**Core.** `IndexSidebar` now passes a `defaultLabel` to the nav `SelectDropdown`, read from the current page with the index's own link as the fallback:

```jsx
defaultLabel={app.current.get('titleControlLabel') ?? app.translator.trans('core.forum.index.all_discussions_link')}
```

This resolves the empty-label case for any consumer, not just tags. A page that filters the discussion list without contributing a nav item can name itself by setting `titleControlLabel` on `app.current`, which is the existing channel a page uses to hand context to other components. Anything that does not falls back to "All Discussions".

**Tags.** On a tag page, set `app.current.set('titleControlLabel', tag.name())`. Because `getButtonContent()` only reaches for `defaultLabel` when no child is active, this surfaces the tag name precisely in the unlisted-tag case, while a listed tag still wins through its own active item.

The change is label-only, so nothing is added to the sidebar, and it needs no `override()`. `addTagList` is untouched.

On ordering: the tags side sets the label from an `extend` on `IndexPage.prototype.view`. That runs after the original `view()` returns, but `view()` only returns a `PageStructure` vnode holding a bound `sidebar` callback, so `IndexSidebar.view()` (and therefore `items()`) does not run until later in the diff. The label is always set before it is read. Stale labels are not a concern either, since `DefaultResolver.makeKey()` keys on route name plus route params, so any tag-to-tag or tag-to-index navigation re-creates the page and with it `app.current`.

### Testing

New `framework/core/js/tests/integration/forum/components/IndexSidebar.test.ts` covers the title control: that it renders, that it falls back to the index link when no nav item is active, that it uses `titleControlLabel` when a page sets one, and that an active nav item still beats `titleControlLabel`. The two behavioural cases fail on `2.x` without the `IndexSidebar` change and pass with it.

Also verified by hand on a 2.x-dev install with 4 secondary tags at descending discussion counts (4/3/2/1), Chromium at 390x844:

- Before: the 4th-ranked tag's page shows an empty toolbar title with only the sort caret.
- After: it shows the tag name. A top-3 tag page renders byte-identically to before, and the index still shows "All Discussions".
- At 1280px the tag page and the index render byte-identically before and after, confirming the sidebar is unchanged: the unlisted tag is still absent from it.

Client-side navigation was checked separately, since the ordering argument above only matters on a route change rather than a fresh load. Driving the router through listed tag, unlisted tag, listed tag, unlisted tag, index, unlisted tag, listed tag gives the correct title at every step, and the unlisted tag never gains a nav item.

No `js/dist` in the diff, per the contributing docs.

### DIFF
--- a/extensions/tags/js/src/forum/addTagFilter.tsx
+++ b/extensions/tags/js/src/forum/addTagFilter.tsx
@@ -87,6 +87,16 @@ export default function addTagFilter() {
     }
   });
 
+  // Name the current tag in the index title control. The nav dropdown only falls
+  // back to this label when none of its items is active, so it surfaces exactly
+  // when the current tag has no item in the sidebar nav, which is the case for
+  // any secondary tag outside the 3 most discussed ones.
+  extend(IndexPage.prototype, 'view', function () {
+    const tag = app.currentTag();
+
+    if (tag) app.current.set('titleControlLabel', tag.name());
+  });
+
   // If currently viewing a tag, restyle the 'new discussion' button to use
   // the tag's color, and disable if the user isn't allowed to edit.
   extend(IndexSidebar.prototype, 'items', function (items) {
--- a/framework/core/js/src/forum/components/IndexSidebar.tsx
+++ b/framework/core/js/src/forum/components/IndexSidebar.tsx
@@ -47,12 +47,18 @@ export default class IndexSidebar<CustomAttrs extends IndexSidebarAttrs = IndexS
       </Button>
     );
 
+    // `SelectDropdown` labels its button with the active child, and falls back to
+    // `defaultLabel` when none of them is active. That happens whenever a page
+    // filters the discussion list without contributing a nav item of its own, so
+    // such a page can name itself here by setting `titleControlLabel` on
+    // `app.current`. Otherwise we are on the unfiltered index.
     items.add(
       'nav',
       <SelectDropdown
         buttonClassName="Button"
         className="App-titleControl"
         accessibleToggleLabel={app.translator.trans('core.forum.index.toggle_sidenav_dropdown_accessible_label')}
+        defaultLabel={app.current.get('titleControlLabel') ?? app.translator.trans('core.forum.index.all_discussions_link')}
       >
         {this.navItems().toArray()}
       </SelectDropdown>
--- a/framework/core/js/tests/integration/forum/components/IndexSidebar.test.ts
+++ b/framework/core/js/tests/integration/forum/components/IndexSidebar.test.ts
@@ -1,0 +1,58 @@
+import bootstrapForum from '@flarum/jest-config/src/bootstrap/forum';
+import IndexSidebar from '../../../../src/forum/components/IndexSidebar';
+import { app } from '../../../../src/forum';
+import mq from 'mithril-query';
+import m from 'mithril';
+
+beforeAll(() => bootstrapForum());
+
+const titleControlLabel = (sidebar: any): string | undefined =>
+  sidebar.rootEl.querySelector('.App-titleControl > .Dropdown-toggle .Button-labelText')?.textContent;
+
+const navItemLabel = (sidebar: any, item: string): string | undefined =>
+  sidebar.rootEl.querySelector(`.Dropdown-menu .item-${item} .Button-labelText`)?.textContent;
+
+describe('IndexSidebar', () => {
+  beforeAll(() => app.boot());
+
+  afterEach(() => app.current.set('titleControlLabel', undefined));
+
+  test('renders', () => {
+    const sidebar = mq(IndexSidebar, {});
+
+    expect(sidebar).toHaveElement('.App-titleControl');
+  });
+
+  test('labels the title control with the index link when no nav item is active', () => {
+    const sidebar = mq(IndexSidebar, {});
+
+    expect(titleControlLabel(sidebar)).toBeTruthy();
+    expect(titleControlLabel(sidebar)).toBe(navItemLabel(sidebar, 'allDiscussions'));
+  });
+
+  test('labels the title control from app.current when a page provides a label', () => {
+    app.current.set('titleControlLabel', 'Support');
+
+    const sidebar = mq(IndexSidebar, {});
+
+    expect(titleControlLabel(sidebar)).toBe('Support');
+  });
+
+  test('prefers an active nav item over the label from app.current', () => {
+    class SidebarWithActiveItem extends IndexSidebar {
+      navItems() {
+        const items = super.navItems();
+
+        items.add('active', m('button', { active: true }, 'Active item'), 200);
+
+        return items;
+      }
+    }
+
+    app.current.set('titleControlLabel', 'Support');
+
+    const sidebar = mq(SidebarWithActiveItem, {});
+
+    expect(titleControlLabel(sidebar)).toBe('Active item');
+  });
+});


imorland (changes requested):
Thanks for the thorough writeup — the root-cause analysis here is spot on, and I verified it: `SelectDropdown.getButtonContent()` derives the label purely from the active child and the nav dropdown passes no `defaultLabel`, so an unlisted current tag → no active child → blank title. That part's not in question.

Where I'd push back is the layer this is fixed at. Adding the current tag to the nav list doesn't only restore the title — it makes the current tag **appear in the index sidebar** (desktop and mobile) whenever you're on its page, when it otherwise wouldn't be listed. That's a visible behaviour change to the sidebar beyond the reported bug, and it's the kind of thing that should be a deliberate product decision rather than a side effect of a title fix.

Digging into alternatives, they all felt like they were compensating for the same underlying thing rather than fixing it:

- **Passing a `defaultLabel` from the tags extension** — the natural "show the tag name" approach — isn't cleanly reachable: the nav `SelectDropdown` is built inline in `IndexSidebar.items()` with no hook to supply a per-page label, and `extend()` can't customise a scalar return, so it'd force an `override()`.
- **A hardcoded core `defaultLabel`** ("All Discussions") fixes the blank but shows a slightly-wrong title on tag pages, and still leaves the door open on other consumers.
- **Making the label an `ItemList`** so it's `extend()`-able is really just a single value wearing a list costume — a smell of its own.

The common thread is that they're all working around the title control having nothing to show when no item is active, rather than addressing why it ends up with nothing to show — and the tags list is just what happens to expose it.

Could you take another pass at this? The constraints I'd want a fix to meet: it shouldn't change what appears in the sidebar, it should hold for any consumer that hits the empty-label case (not just tags), and it should avoid an `override()`. Worth agreeing the approach here before you rebuild it, rather than reworking the current one.

karl-bullock:
Thanks, this is fair. I agree the sidebar behaviour change is the real problem with the current patch: adding the tag to the list to recover the title conflates two separate things.

Working within your three constraints (no sidebar change, holds for any consumer, no `override()`), I think the pieces are already in place:

- `SelectDropdown` already accepts a `defaultLabel`, and `getButtonContent()` falls back to it when no child is active. The nav dropdown in `IndexSidebar.items()` simply never passes one.
- `app.current` (the `PageState`) is the existing channel a page uses to hand context to other components (`DiscussionPage` sets `discussion`/`stream`, `UserPage` sets `user`), and tags already writes `noTagsList` to it.

So the fix I would propose:

**1. Core** — `IndexSidebar` passes a `defaultLabel` to the nav `SelectDropdown`, read from the current page with the index's own label as the fallback:

```jsx
defaultLabel={app.current.get('title') ?? app.translator.trans('core.forum.index.all_discussions_link')}
```

That resolves the empty-label case for any consumer, adds nothing to the sidebar, and needs no override.

**2. Tags** — on a tag page, `app.current.set('title', currentTag.name())`. Because `getButtonContent()` only reaches for `defaultLabel` when no child is active, this surfaces the tag name precisely in the unlisted-tag case, while a listed tag still wins through its own active item.

Net result: an unlisted tag page shows the tag name, the plain index shows "All Discussions", and any other extension caught in the same empty-label spot can set its own page title or fall back to the default. No sidebar change, no `override()`.

The one thing I would want your call on before I rebuild is the `PageState` key. I used `title` above for readability, but something more specific like `sidebarTitle` may be safer against collisions. Happy to go with whatever you prefer.

imorland:
This looks great — thanks for taking it back to the cause rather than working around it. The `defaultLabel` + `app.current` approach hits all three constraints cleanly: label-only so the sidebar is untouched, works for any consumer via the `all_discussions_link` fallback in core, and no `override()`. It also tidies up the fact that `IndexSidebar` never passed the (required) `defaultLabel` in the first place.

On the key: agreed that `title` is too generic for a shared, untyped `PageState` bag — it wouldn't collide today, but it's the kind of name something else will reach for eventually. I'd lean towards a key of `titleControlLabel` — it names what it actually feeds (the `App-titleControl` label) rather than a page concept, so there's no ambiguity about what setting it does.

Go ahead and rebuild on that. 👍

karl-bullock:
Rebuilt on `titleControlLabel` as agreed. `addTagList` is back to stock, so the whole fix is now the `defaultLabel` in `IndexSidebar` plus the tags side setting the label on `app.current`. I have rewritten the description to match.

On the three constraints:

- **No sidebar change.** The tag page and the index render byte-identically to the stock build at desktop width, and the unlisted tag stays out of the sidebar.
- **Any consumer.** The fallback lives in core and terminates at `all_discussions_link`, so nothing can land on blank, and a page can name itself by setting the key.
- **No `override()`.**

Added `IndexSidebar.test.ts` covering the fallback, the `app.current` label, and an active item still winning over it. The two behavioural cases fail on 2.x without the change.

One thing worth your eye, since it is the only call I had to make that had not already come up: the tags side sets the label from an `extend` on `IndexPage.prototype.view`. That works because `view()` only returns a `PageStructure` vnode holding a bound `sidebar` callback, so `IndexSidebar.items()` does not run until later in the diff and the label is always set before it is read. Happy to move it if you would rather it sat somewhere more explicit.

imorland (approved):

Issue #4212, opened 2025-05-03
Awaiting Approval Dropdown bug


tyler71:
### The issue

This is for discussions where the post is "awaiting approval". When someone tries to access drop down options in the UI, it goes away.

### Steps to Reproduce

1. For a awaiting approval post
2. Highlight over avatar
3. Try to navigate to dropdown (might fail)
4. Try to navigate the dropdown itself (might fail)

### Expected Behavior

The popup and dropdown shouldn't have a opacity < 1
You should be able to navigate both without it suddenly disappearing.

### Screenshots

![CopyQ weNnMh](https://github.com/user-attachments/assets/659bf248-518e-475c-8dd4-e7afa537b0c6)

https://i.imgur.com/YjGrD1z.mp4

### Environment

- Flarum version: 1.8.9
- Website URL: https://ledstrain.org
- Webserver: [caddy]
- Hosting environment: Docker Swarm / VPS
- PHP version: x.y.z
- Browser: Firefox 136

### Output of `php flarum info`

```
Flarum core: 1.8.9
PHP version: 8.1.31
MySQL version: 11.4.3-MariaDB-ubu2404
Loaded extensions: Core, date, libxml, openssl, pcre, sqlite3, zlib, ctype, curl, dom, fileinfo, filter, ftp, hash, iconv, json, mbstring, SPL, session, PDO, pdo_sqlite, standard, posix, readline, Reflection, Phar, SimpleXML, tokenizer, xml, xmlreader, xmlwriter, mysqlnd, exif, FFI, gd, gmp, intl, pdo_mysql, zip, excimer, opentelemetry, protobuf, Zend OPcache
+-------------------------------------+---------+--------+
| Flarum Extensions                   |         |        |
+-------------------------------------+---------+--------+
| ID                                  | Version | Commit |
+-------------------------------------+---------+--------+
| flarum-flags                        | v1.8.2  |        |
| flarum-suspend                      | v1.8.4  |        |
| flarum-approval                     | v1.8.2  |        |
| flarum-tags                         | v1.8.3  |        |
| fof-user-bio                        | 1.4.2   |        |
| flarum-sticky                       | v1.8.2  |        |
| flarum-lock                         | v1.8.2  |        |
| fof-pages                           | 1.0.8   |        |
| fof-byobu                           | 1.3.9   |        |
| flarum-mentions                     | v1.8.5  |        |
| flarum-subscriptions                | v1.8.1  |        |
[...]

Issue #3031, opened 2021-08-18
Post metadata becomes transparent and is unreadable when a post is pending approval


matteocontrini:
## Bug Report

**Current Behavior**

When a post is in the status of requiring approval, the post metadata modal becomes transparent and is unreadable.

**Steps to Reproduce**
1. Configure approval so that new posts require approval
2. Create a new post (with a "normal" user), so that it's not approved by default
3. Click on "x minutes ago" on the post and notice that the modal is unreadable

**Expected Behavior**

The modal's opacity should not be changed when the post is pending approval.

**Screenshots**

![image](https://user-images.githubusercontent.com/2164763/129846561-0139f86f-ced5-4c63-8c49-f71c59e0f915.png)

**Environment**
- Flarum version: 1.0.4
- Browser: Firefox 91

SychO9:
same as flarum/issue-archive#328

Issue #950, opened 2016-05-03
Avatar popup overlaps with awaiting approval post


ralphskie:
When a post is awaiting for an approval and you mouse over at the avatar, the popup div overlaps with the post.

i think its a z-index problem

please see the photo
![2016-05-03 17_37_00-new community - negros island ph](https://cloud.githubusercontent.com/assets/5318148/14979658/491d5d7e-1156-11e6-9fd1-572bdcc6388e.jpg)

tobyzerner:
Duplicate of flarum/issue-archive#328

PR #4836, opened 2026-07-25
[2.x] fix(approval): scope the unapproved fade to text, not the whole post

### LINKED ISSUE #4505: [2.x] Flagged posts cannot be edited.
### Current Behavior

When a post is flagged, it cannot be edited by anyone. 

### Steps to Reproduce

Post is flagged.
Go to the post in question. 
Click the menu to edit the post.
Entire menu is washed out, unable to click on any options, including "Edit." 

### Expected Behavior

The menu should come up as normal, not washed out and disabled options. 

### Screenshots

Image

### Environment

I see this happening on Flarum Discuss, I didn't take the time to see if it's reproduced on other communities. 

_No response_

### Possible Solution

_No response_

### Additional Context

_No response_

karl-bullock:
`.Post--unapproved` puts `opacity: 0.5` on the whole `<article>`, which gives the post its own stacking context. Everything that pops out of the post is then stuck inside it. The controls dropdown, the post-meta modal and the user card all render at half opacity, and anything that follows the post in the stream paints over them. The practical effect is that the controls menu on a pending post can't be clicked, because the clicks land on the post underneath.

To reproduce, with Approval enabled:

1. Post a reply, then set `is_approved = 0` on its row in the `posts` table
2. Make sure there is another post after it in the stream
3. Open the controls menu on the pending post

This is the same bug #4453 dealt with, moved rather than removed. Before that PR the opacity sat on `.Post-header`, which trapped the user card. Moving it up to the article freed the card, but pulled `.Post-actions` into the fade for the first time. The wider version of this has been open as flarum/issue-archive#328 since 2016, with #950 and #3031 closed as duplicates of it, and #4212 closed by #4453.

The fix mutes the header with `color` and applies the fade to the elements themselves rather than to their common ancestor, leaving out only each popup's own container. The action buttons are reached one level at a time (`.Post-actions > ul > li > .Button`, `.Post-controls > .Dropdown-toggle`) so that the menu the toggle opens keeps its own opacity: it is a sibling of the toggle inside `.Dropdown`, not a child of it.

**A pending post looks the same as it does today.** Everything visible while the menus are closed stays at half strength, including the avatar and the action buttons. Measured effective opacity on 2.0.0-rc.5, with `.Post-actions` forced visible since core hides it until hover:

| | article | avatar | body | Reply / Like | controls toggle | open controls menu |
| --- | --- | --- | --- | --- | --- | --- |
| current `2.x` | 0.5 | 0.5 | 0.5 | 0.5 | 0.5 | **0.5, and behind later posts** |
| this PR | 1 | 0.5 | 0.5 | 0.5 | 0.5 | **1** |

Checked on 2.0.0-rc.5 against a comment post, a discussion's opening post, and an unapproved event post.

1.x has the same root cause but breaks differently. The controls menu is fine there, and it's the user card and post-meta modal that suffer. The equivalent change fixes it, so I'm happy to open a 1.x PR if that's wanted.

Fixes #4505

### DIFF
--- a/extensions/approval/less/forum.less
+++ b/extensions/approval/less/forum.less
@@ -1,5 +1,16 @@
 .Post--unapproved {
-  opacity: 0.5;
+  .Post-header, .Post-header a,
+  .PostUser .PostUser-name, .PostUser .PostUser-name a {
+    color: var(--muted-more-color);
+  }
+  .Post-body,
+  .Post-footer,
+  .Post-side > *,
+  .PostUser-badges,
+  .EventPost-icon,
+  .EventPost-info {
+    opacity: 0.5;
+  }
 }
 .DiscussionListItem--unapproved {
   .DiscussionListItem--hidden();
--- a/extensions/approval/less/forum.less
+++ b/extensions/approval/less/forum.less
@@ -1,17 +1,50 @@
+// A pending post reads as de-emphasised in its entirety, as it always has. What
+// it must NOT do is carry `opacity` on the <article>: a value below 1 gives the
+// post its own stacking context, and nothing rendered inside it can escape that
+// — not with z-index, not with position: fixed. Every popup the post owns (the
+// user card and the meta dropdown in the header, the controls dropdown in the
+// actions) is drawn as part of that faded group, and the posts after it in the
+// stream paint over the top. The controls menu then cannot be clicked at all,
+// because the clicks land on the post underneath (#4505, and the same root
+// cause behind #4453, #4212, #3031, #950).
+//
+// So the fade is applied to the elements themselves rather than to their common
+// ancestor, and each popup's own container is left out of it. Every part of the
+// post a reader can see while its menus are closed is still at half strength.
 .Post--unapproved {
-  .Post-header, .Post-header a,
-  .PostUser .PostUser-name, .PostUser .PostUser-name a {
+  .Post-header,
+  .Post-header a,
+  .PostUser .PostUser-name,
+  .PostUser .PostUser-name a {
     color: var(--muted-more-color);
   }
+
   .Post-body,
   .Post-footer,
-  .Post-side > *,
   .PostUser-badges,
   .EventPost-icon,
   .EventPost-info {
     opacity: 0.5;
   }
+
+  // The avatar, not the link wrapping it: on 2.x the user card is a sibling
+  // inside the header, but keeping the fade on the image itself holds for any
+  // theme that moves the card inside that link.
+  .Post-avatar,
+  .PostUser-avatar {
+    opacity: 0.5;
+  }
+
+  // Reply, Like, and the controls toggle. Reached one level at a time so the
+  // menu the toggle opens, which is a sibling of it inside .Dropdown, keeps its
+  // own opacity. Fading .Post-actions wholesale would take the menu with it and
+  // put us back where we started.
+  .Post-actions > ul > li > .Button,
+  .Post-controls > .Dropdown-toggle {
+    opacity: 0.5;
+  }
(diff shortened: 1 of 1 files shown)


imorland (changes requested):
The stacking-context diagnosis is correct and the issue history is accurate — but I don't think this is the right fix, and it'll need reworking before it can go in.

**The uniform 50% grey is a deliberate design choice.** A pending post is meant to read as de-emphasised in its entirety — avatar, actions and all — while staying legible enough for a moderator to judge it. This PR restores the avatar, Reply and Like to full strength to free the dropdown, which isn't really a side effect — it's the mechanism of the fix, and it works by undoing that design. I'd like to keep the greying uniform.

**The `.Post--hidden` comparison doesn't quite hold.** `.Post--hidden` *collapses* the post — `display: none` on body/footer/side when `:not(.revealContent)`, header muted via `color`, `opacity` only in the revealed state. That's "hide unless revealed". An unapproved post needs to stay fully shown and simply marked pending, so reusing those selectors borrows styling built for the opposite intent — it reads as precedent but isn't quite one.

**The underlying bug is still there.** The root cause is that `opacity < 1` on the `<article>` traps every absolutely-positioned popup inside it — the controls dropdown, the user card, the post-meta modal. #4453 freed the user card and trapped controls; this frees controls at the cost of the grey. It's the same pendulum swung the other way rather than a resolution.

I think the fix wants to keep the whole-post fade and address the popups where they render: lift the controls dropdown, user card and post-meta modal out of the post's stacking context so the article can keep `opacity: 0.5` intact. That scopes the change to the popup that's open, which is the only surface the bug is actually on — and it'd properly close out the long-standing issue rather than moving it again. Happy to help think through the portal/floating side of that if it's useful.

karl-bullock:
You are right that the uniform grey should stay, and I have kept it. Pushed as e3df1b7.

The previous revision left `.Post-actions` out of the fade wholesale, which is what took Reply and Like with it. Those buttons are now faded individually, one level at a time, so only the popups' own containers are excluded: the controls menu is a *sibling* of the toggle inside `.Dropdown`, not a child, so `.Post-actions > ul > li > .Button` and `.Post-controls > .Dropdown-toggle` reach every button a reader can see while the menu is closed and none of the menu's own. Measured effective opacity on 2.0.0-rc.5, with `.Post-actions` forced visible since core hides it until hover:

| | article | avatar | body | Reply / Like | controls toggle | open menu |
| --- | --- | --- | --- | --- | --- | --- |
| current `2.x` | 0.5 | 0.5 | 0.5 | 0.5 | 0.5 | 0.5 |
| this PR | 1 | 0.5 | 0.5 | 0.5 | 0.5 | 1 |

A correction to my own description while I am here: it claimed the avatar went back to full strength. That was wrong for 2.x. The avatar sits in `.Post-side`, which the previous revision faded through its wrapper, so it was already at 0.5 and you were reading a description that misdescribed its own diff. Only Reply, Like and the toggle actually changed. Fixed in the body above.

On lifting the popups out of the stacking context instead: I do not think that one is available without core changes. `opacity` below 1 makes the element a stacking context and nothing inside can leave it, so the only escape is to leave the post in the DOM. On 2.x all three render inside `<article class="Post">` — the user card and the meta dropdown from `.Post-header` (`Comment.tsx`), the controls dropdown from `.Post-actions` in `AbstractPost.tsx`. Portaling them would mean new core infrastructure for three components, and it would still be the right call for other reasons, just a much bigger one than this bug needs. Happy to help with it as separate work if you want to go that way.

Still glad to open the 1.x equivalent once this lands.

imorland (approved):

PR #4851, opened 2026-07-29
[2.x] fix(messages): repair dialogs left pointing at a deleted first or last message

### LINKED ISSUE #4848: [2.x] `flarum/messages` null bug
### Current Behavior

i recently ran into a bug with messages.

when a message that happens to be the first or last message in a conversation gets deleted, mysql automatically sets that column to null.

because of that, if the first or last message is deleted, the conversation won't load anymore and throws an error in the console:

TypeError: Cannot read properties of null (reading 'id')
at te.content (MessageStream.tsx:67:154)
at te.view (MessageStream.tsx:59:72)
at view.l (render.js:37:16)
at render.js:462:45
at g (render.js:414:9)
at f (render.js:297:11)
at render.js:458:4
at g (render.js:411:15)
at render.js:468:9
at g (render.js:414:9)

i fixed it for now by updating the first message id to an existing message in the database, but it should be fixed in the code to prevent the same issue in the future.

### Steps to Reproduce

1. create a message (private messaging)
2. send some messages
3. delete first messages
4. reload
5. and its an infinite circle

### Expected Behavior

it should work like discussions when even a first post.

### Screenshots

_No response_

### Environment

- Flarum version: x.y.z
- PHP version: x.y.z

```
```

### Possible Solution

_No response_

### Additional Context

_No response_

karl-bullock:
Fixes #4848.

## Why

Deleting the first or last message of a conversation can leave the dialog with a null `first_message_id` or `last_message_id`. After that the conversation does not render at all: `MessageStream` reads straight through the missing relationship and throws `Cannot read properties of null (reading 'id')`, which is the error in the issue. The reply box is part of the same stream, so the one thing that repairs the pointer, posting another message, is out of reach. That is the "infinite circle" the reporter describes, and why they had to repair the row by hand.

Both columns are declared with `nullOnDelete()`, so the database sets the column to null as part of the delete. The repair in `DialogMessage`'s `deleted` handler then compares the dialog's `first_message_id` against the id of the message that was just removed, but by then that column can already be null, the comparison misses, and no repair runs.

It only works when the dialog happens to be in memory before the delete, which is why this looks intermittent. `DialogMessagePolicy::delete()` is the only thing on that path that touches `$message->dialog`, and only in the `'reply'` branch, so PHP short-circuits past it for the other values of `allow_delete_own_messages`. On a bench, deleting the same message from the same dialog:

| `allow_delete_own_messages` | result |
| --- | --- |
| `-1` (indefinitely) | `last_message_id` set to null |
| `reply` | `last_message_id` correctly moved back one message |

## What

- `DialogMessage::boot()` treats a pointer that is missing as one that needs recomputing, rather than only matching it against the deleted id, and uses two conditions instead of an if/elseif chain so both ends can be repaired in one pass. Dialogs that are already in this state are repaired the next time a message in them is deleted.
- `Dialog` casts `first_message_id` and `last_message_id` to integer, which core's `Discussion` already does for `first_post_id` and `last_post_id`. Without a cast the value arrives however the driver hands it over, while `$message->id` is an int, so a strict comparison in the handler above could miss. With the cast it means what it says. (Thanks @imorland: an earlier revision loosened the comparison to `==` instead, which worked but papered over this.)
- `DialogMessageResource::created()` set `first_message_id` to the newly created message whenever it was missing. For a new dialog that is the same thing, but for a dialog whose pointer had gone missing it recorded the newest message as the first one, leaving the stream with a "load previous messages" button that has nothing to load. It now starts from the oldest surviving message.
- The frontend read these relationships in three places without allowing for their absence (`MessageStream.content`, `MessageStream.timeGap`, and the mark-as-read button in `DialogListItem`), so a dialog in this state took the whole page down. They now go through `Dialog.messageRelationshipId()`, which returns `undefined` when the relationship is not there. With no first message there is nothing older to load, with no last message nothing newer, and marking as read is skipped rather than sending a request built from null. This matters for forums that are already affected, since their data is only repaired on the next deletion.

## Verification

`extensions/messages` integration suite: 18 passing. The four new tests fail against current `2.x` and pass with this change. The middle-message and only-message cases pass either way and are there to pin behaviour that should not change.

PHPStan level 6 clean on `extensions/messages`, `check-typings` clean, prettier clean. Source only, since the bot builds `dist` at merge.

One caveat on the cast: I could not produce a failing case for it locally. PDO on this machine returns these columns as ints under both `ATTR_EMULATE_PREPARES` settings, so the suite passes with or without it. It is there because the type should not depend on that, not because a test caught it.

Also checked by hand on 2.0.0-rc.5 with the frontend built from this branch: a dialog whose first message was deleted renders normally instead of throwing, a healthy dialog is unchanged (start-of-conversation gap present, no spurious load buttons), and `messageRelationshipId` returns the same ids the previous code read, with mark-as-read still saving.

### DIFF
--- a/extensions/messages/js/src/common/models/Dialog.ts
+++ b/extensions/messages/js/src/common/models/Dialog.ts
@@ -1,4 +1,4 @@
-import Model from 'flarum/common/Model';
+import Model, { type ModelIdentifier } from 'flarum/common/Model';
 import User from 'flarum/common/models/User';
 import DialogMessage from './DialogMessage';
 import app from 'flarum/common/app';
@@ -25,6 +25,17 @@ export default class Dialog extends Model {
   }
   lastMessage() {
     return Model.hasOne<DialogMessage>('lastMessage').call(this);
+  }
+
+  /**
+   * The id of this dialog's first or last message, when it has one.
+   *
+   * A dialog can be left without either, for instance when the message one of
+   * them pointed at was deleted, so every caller has to cope with the
+   * relationship being absent rather than reading straight through it.
+   */
+  messageRelationshipId(name: 'firstMessage' | 'lastMessage'): string | undefined {
+    return (this.data.relationships?.[name]?.data as ModelIdentifier | undefined)?.id;
   }
 
   unreadCount() {
--- a/extensions/messages/js/src/forum/components/DialogListItem.tsx
+++ b/extensions/messages/js/src/forum/components/DialogListItem.tsx
@@ -67,17 +67,19 @@ export default class DialogListItem<CustomAttrs extends IDialogListItemAttrs = I
           e.preventDefault();
           e.stopPropagation();
 
-          this.attrs.dialog
-            .save({ lastReadMessageId: (this.attrs.dialog.data.relationships?.lastMessage.data as ModelIdentifier).id })
-            .finally(() => {
-              if (this.attrs.dialog.unreadCount() === 0) {
-                app.session.user!.pushAttributes({
-                  messageCount: (app.session.user!.attribute<number>('messageCount') ?? 1) - 1,
-                });
-              }
+          const lastMessageId = this.attrs.dialog.messageRelationshipId('lastMessage');
 
-              m.redraw();
-            });
+          if (!lastMessageId) return;
+
+          this.attrs.dialog.save({ lastReadMessageId: lastMessageId }).finally(() => {
+            if (this.attrs.dialog.unreadCount() === 0) {
+              app.session.user!.pushAttributes({
+                messageCount: (app.session.user!.attribute<number>('messageCount') ?? 1) - 1,
+              });
+            }
+
+            m.redraw();
+          });
         }}
       />,
       100
--- a/extensions/messages/js/src/forum/components/MessageStream.tsx
+++ b/extensions/messages/js/src/forum/components/MessageStream.tsx
@@ -84,7 +84,13 @@ export default class MessageStream<CustomAttrs extends IDialogStreamAttrs = IDia
     const ReplyPlaceholder = this.replyPlaceholderComponent();
     const LoadingPost = this.loadingPostComponent();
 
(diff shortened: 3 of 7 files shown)


imorland (changes requested):
Nice diagnosis — the `'reply'`-branch short-circuit explanation for the intermittency is spot on, and matches what I traced through the policy.

One thing I'd like to resolve before this goes in: the switch from strict `===` to loose `==` in the `deleted` handler.

```php
if ($dialog->first_message_id === null || $dialog->first_message_id == $message->id) {
```

This works, but it papers over the underlying cause rather than fixing it. The reason `===` was unreliable is that `Dialog` doesn't cast its id columns, so `first_message_id` comes back as a driver-dependent type (string under some PDO configs, int under others) while `$message->id` is an int — so strict comparison can miss.

Core's `Discussion` model casts exactly these columns:

```php
// framework/core/src/Discussion/Discussion.php
protected $casts = [
    'first_post_id' => 'integer',
    'last_post_id'  => 'integer',
    // ...
];
```

`Dialog` has no equivalent. Could we add `'first_message_id' => 'integer'` and `'last_message_id' => 'integer'` to `Dialog::$casts` and keep `===`? That fixes the type mismatch at its source, brings the model in line with core's equivalent, and makes the comparison mean what it says. If there's a reason `==` is preferable here, could you note it in the PR description — right now the change from `===` isn't mentioned, so it's easy to miss on review.

karl-bullock:
Thanks, that is a better fix and I have made it.

`Dialog` now casts `first_message_id` and `last_message_id` to integer, matching `Discussion`, and the handler is back to `===`. I have also called the comparison out in the description, which it should have been from the start.

One caveat worth stating rather than leaving implied: I could not produce a failing case for the cast locally. PDO on this machine returns both columns as ints with `ATTR_EMULATE_PREPARES` either way, so the integration suite passes with and without it. The cast is there because the type should not depend on the driver's configuration, not because a test caught it, and the suite cannot tell the two revisions apart. If you know of a configuration where these come back as strings, I am happy to add a case that pins it.

Pushed as de2dfec.

imorland (approved):

PR #4854, opened 2026-07-29
[2.x] fix: Handle network connection loss in the frontend 

gianniguida:
## Problem

All frontend XHR goes through `Application#request()`. When a request fails at the network level — the user is offline, DNS fails, the origin is unreachable — the rejection carries `status === 0`, which `requestErrorCatch()` has no branch for. Users get the generic *"Oops! Something went wrong. Please reload the page and try again."* alert, which wrongly suggests a server-side bug and prompts bug reports for what is a client connectivity issue.

Core also makes no use of the browser's `online`/`offline` events, so users get no proactive indication that they've gone offline, and content that failed to load while offline stays missing until a full page reload.

On 2.x specifically, there is a second network dependency 1.x didn't have: **code-split chunks**. A lazy chunk that fails to load while offline currently breaks its caller with no recovery — opening a discussion leaves the page stuck on its loading skeleton (the `PostStream` import in `DiscussionPage#load` has no rejection handling and never retries), and the reply/discussion composer, log-in/sign-up modals, search modal, and lazy routes (`/settings`, `/notifications`, …) silently do nothing. Reconnecting does not recover any of these.

This is the 2.x counterpart of #4843 (1.x), discussed with @imorland. Two self-contained commits.

## What this PR does

**Commit 1 — network-loss handling in the request layer** (port of #4843):

1. **Classify network-level failures (`status === 0`)** in `requestErrorCatch`: an offline-specific message when `navigator.onLine === false`, otherwise a generic connection-problem message. The existing cross-origin special case is preserved. Parallel failures produce a single alert. Aborted requests cannot trigger this — Mithril leaves their promises unsettled.
2. **`online`/`offline` events**: going offline shows a persistent, dismissible alert; reconnecting clears connection alerts and shows a brief confirmation.
3. **Defer GETs that fail while offline**: instead of rejecting, their promises are held open and settled with the result of a retry once the `online` event fires — so content that failed to load appears automatically on reconnection, with no changes to calling code. Only GETs qualify; writes keep failing fast so nothing can be submitted twice. Deferred requests are keyed by method + URL + params, and identical ones (e.g. a widget polling every few seconds) are retried with a single request that settles every caller.

**Commit 2 — the same self-healing for lazy chunks** (2.x-only): every chunk load funnels through `ExportRegistry#loadChunk` (webpack's script loader is overridden with it), so a chunk failure that occurs while the browser is offline is now held back and retried on the `online` event. The `import()` promise stays pending in the meantime, and every call site — composer bodies, lazy routes, modals, the post stream, and extension chunks — recovers on its own. Chunk failures while the browser is online are reported unchanged.

New locale keys: `core.lib.error.network_message`, `core.lib.error.offline_message`, `core.lib.connection_restored_message` (English only).

## Behavior notes

- `errorHandler` and rejection propagation are unchanged for every status except the offline-GET deferral case; regression tests cover 422/500/cross-origin.
- A survey of core, bundled extensions, and popular FoF extensions found no GET call sites whose `errorHandler`/`.catch` logic would be adversely affected by deferral — GET failure handling in the wild is loading-flag resets, which now correctly keep spinners visible until the content self-heals.
- No `js/dist` changes included; bundles are left to CI.

## Testing

- 26 new Jest unit tests: `tests/unit/common/Application.test.ts` (classification online/offline/cross-origin, alert dedup, custom-error-handler contract, 422/500 regression guards, connectivity events, deferral + resolution on reconnect, re-deferral while still offline, retry dedup) and `tests/unit/common/ExportRegistry.test.ts` (chunk load success, online-failure passthrough, offline-failure retry on reconnect, repeated offline failures). Full suite: 259 passing.
- Manual smoke test on a local 2.x instance: offline request failure → single dedicated alert; opening a discussion / the composer while offline → waits instead of breaking; reply while offline → alert + composer content preserved; reconnect → offline alert cleared, confirmation shown, deferred content and chunks load without reload; 422/500 render as on `2.x`.

**Screenshot**
image
image

**Necessity**

**Confirmed**

### DIFF
--- a/framework/core/js/src/common/Application.tsx
+++ b/framework/core/js/src/common/Application.tsx
@@ -75,7 +75,7 @@ export type AsyncNewComponent<Comp> = () => Promise<any & { default: NewComponen
 export type RouteItem<
   Attrs extends ComponentAttrs,
   Comp extends Component<Attrs & { routeName: string }>,
-  RouteArgs extends Record<string, unknown> = {}
+  RouteArgs extends Record<string, unknown> = {},
 > = {
   /**
    * The path for your route.
@@ -111,7 +111,7 @@ export type RouteItem<
 export interface RouteResolver<
   Attrs extends ComponentAttrs,
   Comp extends Component<Attrs & { routeName: string }>,
-  RouteArgs extends Record<string, unknown> = {}
+  RouteArgs extends Record<string, unknown> = {},
 > {
   /**
    * A method which selects which component to render based on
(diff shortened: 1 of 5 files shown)


imorland (approved):
Nice work @gianniguida thank you very much for this!

PR #4880, opened 2026-08-02
[2.x] fix(realtime): honour user.viewLastSeenAt in the typing indicator

### LINKED ISSUE #4879: Realtime typing indicator ignores `user.viewLastSeenAt`, so moderators see `[Anonymous]` instead of the typing user
## Current Behavior

Core treats `user.viewLastSeenAt` ("Always view user last seen time") as the override for a user's
`discloseOnline` preference: a user who has hidden their online status still has their `lastSeenAt`
disclosed to actors holding that permission.

`UserResource`:

```php
Schema\DateTime::make('lastSeenAt')
    ->visible(fn (User $user, Context $context) =>
        $user->getPreference('discloseOnline') || $context->getActor()->can('viewLastSeenAt', $user)
    )
```

The realtime typing indicator is the one place where that override is not applied. It anonymises at
the *sender*, unconditionally, so the identity never reaches anyone — including actors who are
explicitly permitted to see through `discloseOnline`.

`extensions/realtime/js/src/forum/extend/Discussion/TypingIndicator.tsx`:

```ts
this.actorIsTyping = (): void => {
  const discloseOnline = app.session.user?.preferences()?.discloseOnline;

  app.websocket_channels.discussion?.trigger('client-typing', {
    displayName: discloseOnline ? app.session.user?.displayName() : '[anonymous]',
    discloseOnline,
    time: Date.now(),
  });
};
```

The result is that a moderator with both `discussion.flarum-realtime.view-who-types` and
`user.viewLastSeenAt` sees `[Anonymous] is typing`, while the same moderator can see that user's
last-seen time on their profile. The two disclosures are inconsistent.

This is not fixable in a theme or extension: the sender scrubs the name before it is ever
broadcast, and `private-typing={id}` has a single audience — everyone who can see the discussion
(`AuthController::typing()` checks only `Discussion::whereVisibleTo`). Simply broadcasting the real
name and filtering in the client would expose it to every reader over the websocket, so the routing
has to change server-side.

## Steps to Reproduce

1. Install Flarum 2.0 with the Realtime extension enabled and the "Typing indicator" setting on
   (default).
2. In **Admin → Permissions**, grant `discussion.flarum-realtime.view-who-types` to Moderators.
   (`user.viewLastSeenAt` is already seeded to Moderators by
   `2018_07_21_000100_seed_default_group_permissions.php`.)
3. As user A (a regular member), go to **Settings → Privacy** and disable
   "Allow others to see when I am online" (`discloseOnline`).
4. As user A, open a discussion and start typing a reply.
5. As user B (a Moderator), view the same discussion in another browser.

Observed: B sees `[Anonymous] is typing`.
Also observed: B *can* see A's last-seen time on A's profile, i.e. the same override that applies
there does not apply here.

## Expected Behavior

An actor holding `user.viewLastSeenAt` should see who is typing, even when that user has hidden
their online status — consistent with how the same permission already overrides `discloseOnline`
for `lastSeenAt`. Actors without the permission should continue to see `[Anonymous]`, and the
identity should never reach their socket.

## Environment

- Flarum version: 2.0.0-rc.5, and unchanged on current `2.x` (`2553eb514`)
- Webserver: nginx
- Hosting environment: vps
- PHP version: 8.4
- Browser: not browser-specific

The behaviour above is structural rather than environmental — the sender scrubs the display name
before the event is ever broadcast, so there is no configuration under which a permitted actor
receives it.

## Possible Solution

Happy to open a PR. Sketch, reusing two patterns already in the extension — an AuthController-gated
channel (as `private-index-typing-tag=` does) and server-side re-routing of a client event in
`Message` (as `relayIndexTyping()` / `relayComposeTyping()` do):

1. **New permission-gated channel** `private-typingIdentified={discussionId}`. The existing
   dispatcher regex `~^private-(?<subject>[a-zA-Z]+)=(?<id>[0-9]+)$~` already matches a camelCase
   subject, so this needs no regex change — only a new method:

   ```php
   protected function typingIdentified(int $id): bool
   {
       return $this->actor->hasPermission('user.viewLastSeenAt')
           && Discussion::whereVisibleTo($this->actor)->where('id', $id)->exists();
   }
   ```

   Permission evaluation happens once per subscription in a normal HTTP request with a real actor,
   so the websocket process does no permission work in the hot path.

2. **Split the relay in `Message::respond()`** when a `client-typing` event carries a
   non-disclosing sender:
   - full payload → `private-typingIdentified={id}`;
   - anonymised payload (`displayName: null`) → `private-typing={id}`, excluding the sender *and*
     every socket subscribed to the identified channel.

   The exclusion means a privileged subscriber (who is on both channels) receives exactly one
   event, so no client-side deduplication is needed. When the sender is disclosing, the current
   single broadcast is unchanged. When nobody privileged is listening the identified channel does
   not exist and behaviour is identical to today.

3. **Derive the sender's identity server-side rather than trusting the payload.** Gating a name
   behind a moderator permission makes an authority claim, so the name should not be a
   client-asserted string. `Manager` can index `socketId → userId` when a connection subscribes to
   `private-user={id}` — an authenticated identity claim, since `AuthController::user()` requires
   `actor->id === id` and subscription verifies the HMAC — and `Message` can then read
   `display_name` and `discloseOnline` from the database (short TTL cache) and overwrite both
   fields before relaying. This keeps the existing payload shape, so third-party `client-typing`
   listeners and the copy in `flarum/messages` are unaffected. Senders that cannot be identified
   (guests, or the brief window after a `forceReconnect` before the user channel re-subscribes)
   fail closed to anonymous. As a side effect this also closes the existing ability for a modified
   client to broadcast an arbitrary `displayName`.

4. **Frontend**: a `canViewHiddenTypers` boolean on `UserResource` (visible to self only, mirroring
   the existing `canViewWhoTypes` field), used to decide whether to subscribe to the identified
   channel; `TypingState.add()` anonymises only when the payload carries no name. I would render
   the disclosed name marked rather than bare — e.g. `Bob (hidden)` — so a moderator knows that
   person is invisible to everyone else.

Roughly 300 lines plus regression tests, which would follow the existing
`tests/unit/Websocket/MessageTest.php` (relay routing) and
`tests/integration/api/IndexTypingTagAuthTest.php` (channel auth by permission).

## Additional Context

- **Defaults are unchanged for regular members.** `user.viewLastSeenAt` is seeded to Moderators
  only, so on a default install this discloses nothing that is not already disclosed to that same
  group via `lastSeenAt`. I do not think it warrants a separate setting, but happy to add one
  (mirroring `index-typing-indicator-restricted`) if you would rather it were opt-in.
- **Semantics are the part worth your call**, not the mechanism: whether `user.viewLastSeenAt`
  should imply "may see who is typing while hidden", or whether typing identity deserves its own
  permission (e.g. `flarum-realtime.view-hidden-typers`). I have assumed reuse, since that is what
  admins already reach for when they want moderators to see through hidden online status, and it
  needs no migration — but I will follow whichever you prefer.
- Like the index typing dots, this lives in the relay and therefore assumes the bundled websocket
  server.
- `flarum/messages` has its own copy of the typing indicator, which currently *drops* non-disclosing
  typers entirely rather than showing `[Anonymous]`. I have deliberately left it out of scope; it
  could follow the same shape in a separate PR if this one is accepted.

ekumanov:
Fixes #4879.

Core treats `user.viewLastSeenAt` as the override for a user's `discloseOnline` preference — someone who hides their online status still has `lastSeenAt` disclosed to actors holding it (`UserResource`). The typing indicator was the one place that override wasn't applied: the display name was scrubbed at the *sender*, so it never reached anyone, and a moderator saw `[Anonymous] is typing` for a user whose last-seen time they can read on the profile page.

Scrubbing at the receiver instead isn't an option: `private-typing={id}` is subscribable by everyone who can see the discussion, so a name broadcast there is a name disclosed to all of them. So the relay splits the event by audience instead.

### How it works

When the typist is hiding their online status:

- the full payload goes to `private-typingIdentified={id}`, which only holders of the permission can subscribe to;
- an anonymised payload — no name at all — goes to the discussion channel, **skipping the identified channel's subscribers**, so a privileged viewer sees one event rather than a name *and* an `[Anonymous]` for the same person.

When the typist is disclosing, the existing single broadcast is unchanged. When nobody privileged is listening, the identified channel doesn't exist and behaviour is exactly as before.

Authorisation happens in `AuthController::typingIdentified()`, so the permission is evaluated once per subscription against a real actor in an ordinary request — the websocket server does no permission work per event, only set arithmetic over socket IDs. This follows the pattern established by `private-index-typing-tag=` in #4756, and the relay split follows `relayIndexTyping()` / `relayComposeTyping()`.

### Identity is no longer taken from the payload

Gating a name behind a permission makes an authority claim about who is typing, and a client-asserted string can't back that. Both the display name and the disclosure preference are now resolved from the identity the connection authenticated with:

- `Manager` indexes socket → user id when a connection subscribes to `private-user={id}` — an authenticated identity claim, since that channel is only signed for the matching actor and the signature is verified against the socket ID. Same reasoning `relayComposeTyping()` already relies on;
- `TypingIdentity` resolves name + `discloseOnline` from that id, cached briefly (5s, deliberately short — it bounds how long someone who has just hidden their status keeps being announced by name);
- senders that can't be identified (guests, or the brief window after a reconnect before channels are re-established) fail closed to anonymous.

As a side effect this closes the existing ability for a modified client to broadcast an arbitrary `displayName` into a discussion.

The payload shape is unchanged, so existing `client-typing` listeners keep working — including flarum/messages, whose own dialog channel this doesn't touch (covered by a test).

### Notes

- **Defaults are unchanged for regular members.** `user.viewLastSeenAt` is seeded to Moderators only, so on a default install this discloses nothing to a group that can't already read the same users' last-seen time. Happy to add a setting (mirroring `index-typing-indicator-restricted`) if you'd prefer it opt-in.
- **The semantics are your call, not the mechanism** — whether `user.viewLastSeenAt` should imply "may see who is typing while hidden", or whether this deserves its own permission (`flarum-realtime.view-hidden-typers`). I assumed reuse: it's what admins already reach for when they want moderators to see through a hidden online status, and it needs no migration. Straightforward to switch.
- A disclosed hidden user is rendered `{username} (hidden)` rather than bare, so a moderator can tell that person is invisible to everyone else.
- Like the index typing dots, this lives in the relay and assumes the bundled websocket server.
- `flarum/messages` has its own copy of the typing indicator (it currently *drops* non-disclosing typists rather than showing `[Anonymous]`). Deliberately out of scope; it could follow the same shape separately.

### Tests

- `tests/unit/Websocket/TypingRelayTest.php` — 7 tests over the split: name reaches the whole discussion when disclosing; name withheld from the discussion channel when hidden; name reaches the identified channel (addressed to it, so pusher-js routes it correctly); identified subscribers excluded from the anonymised broadcast; payload-claimed identity ignored; unidentifiable sender falls back to anonymous; typing on other channels passes through verbatim.
- `tests/integration/api/TypingIdentifiedAuthTest.php` — 5 tests over channel auth: permitted user 200; user with view-who-types but without `user.viewLastSeenAt` 403; guest 403; permission doesn't bypass discussion visibility; the ordinary typing channel is unaffected.

Realtime's full unit (41) and integration (39) suites pass locally, PHPStan is clean, and `tsc` typechecks. No `js/dist` or `dist-typings` in the commit.

### DIFF
--- a/extensions/realtime/extend.php
+++ b/extensions/realtime/extend.php
@@ -64,6 +64,19 @@ return [
 
                     return $settings->get('flarum-realtime.typing-indicator')
                         && $context->getActor()->hasPermissionLike('flarum-realtime.view-who-types');
+                }),
+
+            // Whether to also subscribe to the channel that names users who are typing
+            // while hiding their online status. The channel is authorized server-side
+            // regardless (see AuthController::typingIdentified); this only saves the
+            // client a subscription it would be refused.
+            Schema\Boolean::make('canViewHiddenTypers')
+                ->visible(fn (User $user, Context $context) => $context->getActor()->id === $user->id)
+                ->get(function (User $model, Context $context) {
+                    $settings = resolve(SettingsRepositoryInterface::class);
+
+                    return $settings->get('flarum-realtime.typing-indicator')
+                        && $context->getActor()->hasPermission('user.viewLastSeenAt');
                 })
         ]),
 
--- a/extensions/realtime/js/src/@types/shims.d.ts
+++ b/extensions/realtime/js/src/@types/shims.d.ts
@@ -11,6 +11,12 @@ declare module 'flarum/common/Application' {
       user: Channel | null;
       /** Presence channel for the currently open discussion (typing indicator). */
       discussion?: Channel;
+      /**
+       * Companion to {@link discussion}, carrying the identities of users typing
+       * while hiding their online status. Only subscribable with core's
+       * `user.viewLastSeenAt` permission.
+       */
+      discussionIdentified?: Channel;
       /** Shared public channel feeding ambient typing dots on the discussion list. */
       indexTyping?: Channel;
       /** Per-restricted-tag channels feeding typing dots for restricted discussions. */
--- a/extensions/realtime/js/src/forum/extend/Discussion/TypingIndicator.tsx
+++ b/extensions/realtime/js/src/forum/extend/Discussion/TypingIndicator.tsx
@@ -31,11 +31,10 @@ export default function (): void {
     this.discussion.typingState = new TypingState();
 
     this.actorIsTyping = (): void => {
-      const discloseOnline = app.session.user?.preferences()?.discloseOnline;
-
+      // Nothing identifying is sent: the server resolves who we are from the
+      // connection's authenticated user channel, and decides which audience may
+      // learn it. `time` drives the receiving client's expiry countdown.
       app.websocket_channels.discussion?.trigger('client-typing', {
-        displayName: discloseOnline ? app.session.user?.displayName() : '[anonymous]',
-        discloseOnline,
         time: Date.now(),
       });
     };
@@ -60,12 +59,27 @@ export default function (): void {
     }
 
     if (this.discussion) {
-      app.websocket_channels.discussion = app.websocket.subscribe('private-typing=' + m.route.param('id').match(/[0-9]+/));
+      const discussionId = m.route.param('id').match(/[0-9]+/);
+
+      app.websocket_channels.discussion = app.websocket.subscribe('private-typing=' + discussionId);
 
       if (this.discussion.attribute('canViewWhoTypes')) {
         app.websocket_channels.discussion.bind('client-typing', (data: TypingData) => {
           this.discussion.typingState?.add(data);
         });
+
+        // Users who may see through a hidden online status (core's
+        // `user.viewLastSeenAt`) also listen on the channel that names hidden
+        // typists. Subscription is authorised server-side; the attribute just
+        // avoids asking for a channel we'd be refused. Anonymised events for
+        // those same typists are not sent here, so there's nothing to de-duplicate.
+        if (app.session.user?.attribute('canViewHiddenTypers')) {
+          app.websocket_channels.discussionIdentified = app.websocket.subscribe('private-typingIdentified=' + discussionId);
+
+          app.websocket_channels.discussionIdentified.bind('client-typing', (data: TypingData) => {
+            this.discussion.typingState?.add(data);
+          });
+        }
       }
     }
   });
--- a/extensions/realtime/js/src/forum/states/TypingState.ts
+++ b/extensions/realtime/js/src/forum/states/TypingState.ts
@@ -5,7 +5,11 @@ export interface TypingUserMap {
 }
 
 export interface TypingData {
-  displayName: string;
+  /**
+   * Null when the server withheld the identity: the user is hiding their online
+   * status and we are not permitted to see through it.
+   */
+  displayName: string | null;
   discloseOnline: boolean;
   time: number;
 }
@@ -29,15 +33,21 @@ export default class TypingState {
   protected truncationTimer: ReturnType<typeof setTimeout> | null = null;
 
   /**
-   * Record an incoming typing event. When the sender has not disclosed their
-   * online status, their name is replaced with the anonymous placeholder.
+   * Record an incoming typing event.
+   *
+   * Whether we are allowed to know who is typing is decided server-side — a name
+   * only reaches us if we are entitled to it — so this just renders what arrived.
+   * A hidden user we *are* permitted to see is labelled as such, so it's clear
+   * they are invisible to everyone else.
    */
   add(data: TypingData): void {
-    if (!data.discloseOnline) {
-      data.displayName = String(app.translator.trans('flarum-realtime.forum.typing-indicator.anonymous-user'));
-    }
+    const name = data.displayName
+      ? data.discloseOnline
+        ? data.displayName
+        : String(app.translator.trans('flarum-realtime.forum.typing-indicator.hidden-user', { username: data.displayName }))
+      : String(app.translator.trans('flarum-realtime.forum.typing-indicator.anonymous-user'));
 
-    this.usersTyping[data.displayName] = data.time;
+    this.usersTyping[name] = data.time;
     m.redraw();
   }
 
--- a/extensions/realtime/js/src/forum/states/TypingState.ts
+++ b/extensions/realtime/js/src/forum/states/TypingState.ts
@@ -1,4 +1,5 @@
 import app from 'flarum/forum/app';
+import extractText from 'flarum/common/utils/extractText';
 
 export interface TypingUserMap {
   [displayName: string]: number;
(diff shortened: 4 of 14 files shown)


imorland (changes requested):
Thanks — the audience split and the identity handling both look right, and the test coverage over the privacy properties is solid.

One defect that needs fixing before merge:

**`TypingIdentity` is never bound as a singleton, so the cache never works.**

`WebsocketProvider::register()` singletons the websocket server's stateful services (`Manager`, `IndexTypingPresence`, `PresenceChannelAuthorizer`), but `TypingIdentity` isn't among them. `resolve(TypingIdentity::class)` in `Message::relayTyping()` therefore constructs a fresh instance on every typing event — the 5s cache never hits, `PRUNE_THRESHOLD` is dead code, and every ping from every typist runs a `User::find()` in the event loop, which is exactly what the class docblock says it's there to prevent.

The unit tests don't catch it because they register the stub with `$this->container->instance(...)`, which is shared — production resolution isn't.

Fix is one line in `WebsocketProvider::register()`:

```php
$this->container->singleton(TypingIdentity::class);
```

Ideally with a test that pins the binding (resolving it twice from the container and asserting the same instance), so this can't regress silently.

ekumanov:
Good catch, and thanks for pinpointing why the unit tests were blind to it — that was the part I'd have gone looking for last.

Confirmed and fixed in fcc8fdc. `TypingIdentity` is now bound alongside the other stateful websocket services in `WebsocketProvider::register()`.

On the test: I pinned all five stateful services rather than just this one, since the same omission would silently defeat any of them and the failure mode is invisible — the class keeps working, just with its cache permanently empty. It's a data-provider test asserting each resolves to the same instance twice from the real container. I checked it actually fails with the binding removed (only the typing-identity case, the other four stay green) rather than trusting that it would.

Also re-ran the end-to-end check against a live daemon after the change — the audience split and the identity resolution still behave correctly, so the binding didn't shift any observable behaviour beyond making the cache real.

imorland (approved):
The singleton binding is right, and the test is better than what I suggested — parameterising across every stateful websocket service rather than just `TypingIdentity` means the next service added to `WebsocketProvider` is covered by construction, and the docblock explains precisely why unit tests couldn't have caught it (they inject their own shared stub).

Verified: the new test fails on exactly the `typing identity cache` case with the binding removed, and only that case. Realtime suites green — 41 unit, 44 integration.

Everything from my earlier review still stands: the audience split, the identity resolution from the authenticated connection, and the fail-closed behaviour for unidentifiable senders. Nice work.

PR #4920, opened 2026-08-11
[2.x] fix(phpstan): recognize conditional model casts

### LINKED ISSUE #4808: `Flarum\Extend\Model::cast` not working when used inside conditional extender
### Current Behavior

When `Flarum\Extend\Model::cast` is used to add a custom attribute cast inside a `Flarum\Extend\Conditional` extender, Static Code Analysis doesn't pick up on it

### Steps to Reproduce

Add a cast in extend.php:

```php
 (new Extend\Conditional())
        ->whenExtensionEnabled('acme-foobar', fn () => [
            (new Extend\Model(Foo::class))
                ->cast('bar', 'boolean'),
        ]),
```

And try to use it somewhere:

```php
$foo->bar = true
// Access to an undefined property ACME\Foobar\Foo::$bar
```

### Expected Behavior

Static code analysis should pick up on casts inside the conditional extender or `cast()` should throw an error when used inside a conditional extender to prevent it unexpectedly not working

### Screenshots

_No response_

### Environment

- Flarum version: 1.8.16
- Website URL: n.a.
- Webserver: n.a.
- Hosting environment: n.a.
- PHP version: 8.3.31
- Browser: n.a.

_No response_

### Possible Solution

_No response_

### Additional Context

Workaround now is to just always add the casts outside of conditional extenders which has no negative sideffect on runtime logic, however the intent when reading code gets clouded.

I didn't confirm that the issue also happens on Flarum 2.x however I don't recall any changes being made in this area which could address this issue

drewmt:
## Changes

- inspect the callable arrays returned by all `Conditional` extender methods
- support both arrow functions and closure return bodies without executing them
- add end-to-end PHPStan fixtures for extension, setting, and generic conditions

## Why

`Conditional` moved from direct extender arrays to `callable|string`, but the
PHPStan resolver still looked only for a direct array on
`whenExtensionEnabled()`. Valid model casts nested in the current callable API
were therefore reported as undefined properties.

Fixes #4808

## Testing

- Failing-first fixture reproduced `property.notFound` on unchanged `2.x`
- Full `vendor/bin/phpstan analyse -c phpstan.neon --no-progress --error-format=table --memory-limit=2G`
- PHP syntax checks for the resolver and both fixtures
- `composer validate --no-check-publish` (valid; existing dependency-constraint warnings only)

### DIFF
--- a/php-packages/phpstan/src/Extender/Resolver.php
+++ b/php-packages/phpstan/src/Extender/Resolver.php
@@ -11,7 +11,10 @@ namespace Flarum\PHPStan\Extender;
 
 use Flarum\PHPStan\Extender\MethodCall as ExtenderMethodCall;
 use PhpParser\Node\Arg;
+use PhpParser\Node\Expr;
 use PhpParser\Node\Expr\Array_;
+use PhpParser\Node\Expr\ArrowFunction;
+use PhpParser\Node\Expr\Closure;
 use PhpParser\Node\Expr\MethodCall;
 use PhpParser\Node\Expr\New_;
 use PhpParser\Node\Stmt\Namespace_;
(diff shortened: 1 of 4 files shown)


imorland (approved):
This is very much appreciated, thank you @drewmt

drewmt:
Thanks a lot for the review and merge! Happy to continue with the next fix when I see the next clear gap.

PR #4952, opened 2026-08-20
[2.x] fix: populate statistics previous period

### LINKED ISSUE #4951: [1.x & 2.x] Statistics: "Previous period" graph is never populated
### Current Behavior

Hovering over the graph of the "Statistics" flarum core extension the tooltip always shows "Previous period" as empty. This is the case regardless of data point (users, discussions, posts) as well as time period selected (pre-defined, custom)

### Steps to Reproduce

1. On a community with a few weeks worth of tracked statistics data check out the "Previous period" tooltip on the graph

### Expected Behavior

The "Previous period" metric should be populated with the metric I  am looking at but just [SELECTED TIME FRAME] before the date I am looking at. So if i selected "Last 7 days" as time frame and I look at August 16th I would expected "Previous period" to show metric from August 9th.

### Screenshots

Image

### Environment

- Flarum version: 1.8.19 and 2.0.0-rc.6
- PHP version: 8.3.33

_No response_

### Possible Solution

_No response_

### Additional Context

_No response_

drewmt:
## What changed

The previous-period chart points were querying zero-length ranges because each end timestamp matched its start timestamp. This includes the selected period step in the end timestamp so every point covers the corresponding interval from the preceding period.

## Why

The Statistics chart displayed zero for the entire previous-period series even when the API response contained matching data.

Fixes #4951.

## Validation

- Statistics frontend formatting check
- TypeScript check
- Statistics production Webpack build
- Focused range regression check for the previous-period interval

### DIFF
--- a/extensions/statistics/js/src/admin/components/StatisticsWidget.tsx
+++ b/extensions/statistics/js/src/admin/components/StatisticsWidget.tsx
@@ -379,7 +379,7 @@ export default class StatisticsWidget extends DashboardWidget {
       labels.push(label);
 
       thisPeriod.push(this.getPeriodCount(this.selectedEntity, { start: i, end: i + period.step }));
-      lastPeriod.push(this.getPeriodCount(this.selectedEntity, { start: i - periodLength, end: i - periodLength }));
+      lastPeriod.push(this.getPeriodCount(this.selectedEntity, { start: i - periodLength, end: i - periodLength + period.step }));
     }
 
     if (thisPeriod.length === 0) {


imorland (approved):
Confirmed independently, and this is the right fix.

`getPeriodCount` filters on `time >= period.start && time < period.end`, so the zero-width interval matched nothing and every previous-period bucket came back 0 — populated array, right length, all zeros, which is exactly the "always empty" symptom in the issue.

Worth recording that this is a restoration rather than a new behaviour: the line was correct until #3661 dropped `+ period.step` while working on custom date ranges, so it's been broken since November 2022 on both 1.x and 2.x.

I verified against the backend too — nothing is missing server-side. `ShowStatisticsData` deliberately over-fetches for this: predefined periods start at `-2 years` ("need -12 months and period before that") and custom ranges subtract the range's own length ("we need to include an equal amount beforehand to show the data for the previous period"). The data was arriving and being discarded client-side.

Thanks for the clean diagnosis.

I'll follow up separately with two things rather than hold this up:

- A jest harness for this extension plus regression tests around the chart datasets. This line regressed silently once already, in a PR about something else, so it's worth pinning down.
- A sibling of the same mistake at `StatisticsWidget.tsx:182`, where `content()` builds the custom-period range as `start: this.customPeriod?.end!` — a zero-width interval again, so the per-entity count reads 0 whenever a custom range is selected. Out of scope here; no change needed on this PR.

